PR #724, opened 2014-03-04
On the suggestion of @zooko, note that we have not been audited

alex:


seanieb:
@zooko has there been any further discussion of an audit?  (I've searched and cant find any)

zooko:
Not that I'm aware of.

PR #1279, opened 2014-07-22
WIP: DH implementation

public:
This is an implementation of DH suitable for use in TLS. However it also leaves room for X9.42/RFC 2631 DH exchange in the future.

This flexibility will be achieved by adding a `DHExchangeAlgorithm` interface, instances of which are passed to `key.agree` to work out which key validations to apply and also by adding a new `DSAParameters` variant to hold the additional values used in X9.42 to verify public key and parameter groups.

The words "exchange" and "agree" are both used in this PR. "Agree" refers to the act of taking the received key material and computing the shared key value. "Exchange" refers to the rest of the process.

TODO:

- [x] PR for vector loader
- [x] PR for numbers
- [ ] PR for interfaces
- [ ] PR for implementation

### DIFF
--- a/cryptography/hazmat/backends/openssl/backend.py
+++ b/cryptography/hazmat/backends/openssl/backend.py
@@ -32,6 +32,9 @@ from cryptography.hazmat.backends.openssl.ciphers import (
     _AESCTRCipherContext, _CipherContext
 )
 from cryptography.hazmat.backends.openssl.cmac import _CMACContext
+from cryptography.hazmat.backends.openssl.dh import (
+    _DHParameters, _DHPrivateKey, _DHPublicKey, _DHKeyAgreementContext
+)
 from cryptography.hazmat.backends.openssl.dsa import (
     _DSAParameters, _DSAPrivateKey, _DSAPublicKey,
     _DSASignatureContext, _DSAVerificationContext
--- a/cryptography/hazmat/backends/openssl/backend.py
+++ b/cryptography/hazmat/backends/openssl/backend.py
@@ -33,7 +33,7 @@ from cryptography.hazmat.backends.openssl.ciphers import (
 )
 from cryptography.hazmat.backends.openssl.cmac import _CMACContext
 from cryptography.hazmat.backends.openssl.dh import (
-    _DHParameters, _DHPrivateKey, _DHPublicKey, _DHKeyExchangeContext
+    _DHParameters, _DHPrivateKey, _DHPublicKey
 )
 from cryptography.hazmat.backends.openssl.dsa import (
     _DSAParameters, _DSAPrivateKey, _DSAPublicKey,
@@ -1095,7 +1095,6 @@ class Backend(object):
         return _DHPrivateKey(self, dh_key_cdata)
 
     def load_dh_private_numbers(self, numbers):
-        #dh._check_dh_private_numbers(numbers)
         parameter_numbers = numbers.public_numbers.parameter_numbers
 
         dh_cdata = self._lib.DH_new()
@@ -1107,7 +1106,6 @@ class Backend(object):
         dh_cdata.pub_key = self._int_to_bn(numbers.public_numbers.y)
         dh_cdata.priv_key = self._int_to_bn(numbers.x)
 
-
         codes = self._ffi.new("int[]", 1)
         res = self._lib.DH_check(dh_cdata, codes)
         assert res == 1
@@ -1116,7 +1114,6 @@ class Backend(object):
         return _DHPrivateKey(self, dh_cdata)
 
     def load_dh_public_numbers(self, numbers):
-        #dh._check_dh_parameters(numbers.parameter_numbers)
         dh_cdata = self._lib.DH_new()
         assert dh_cdata != self._ffi.NULL
         dh_cdata = self._ffi.gc(dh_cdata, self._lib.DH_free)
@@ -1128,7 +1125,6 @@ class Backend(object):
         return _DHPublicKey(self, dh_cdata)
 
     def load_dh_parameter_numbers(self, numbers):
-        #dh._check_dh_parameters(numbers)
         dh_cdata = self._lib.DH_new()
         assert dh_cdata != self._ffi.NULL
         dh_cdata = self._ffi.gc(dh_cdata, self._lib.DH_free)
@@ -1137,7 +1133,6 @@ class Backend(object):
         dh_cdata.g = self._int_to_bn(numbers.g)
 
         return _DHParameters(self, dh_cdata)
-
 
 
 class GetCipherByName(object):
--- a/cryptography/hazmat/backends/openssl/backend.py
+++ b/cryptography/hazmat/backends/openssl/backend.py
@@ -1067,7 +1067,7 @@ class Backend(object):
         dh_param_cdata = self._ffi.gc(dh_param_cdata, self._lib.DH_free)
 
         res = self._lib.DH_generate_parameters_ex(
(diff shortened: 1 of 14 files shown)


alex:
Needs a rebase because of merge conflicts.

cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -329,6 +333,8 @@ def _unknown_error(self, error):
         )
 
     def _bn_to_int(self, bn):
+        assert bn != self._ffi.NULL
```
reaperhulk: Is this assert required for this PR or could it be split into a separate one?

cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -356,6 +362,8 @@ def _int_to_bn(self, num, bn=None):
         be discarded after use.
         """
 
+        assert (bn is None) or (bn is not None and bn != self._ffi.NULL)
```
reaperhulk: Same question: Is this assert required for this PR or could it be split into a separate one?

cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1055,6 +1063,77 @@ def _ec_key_set_public_key_affine_coordinates(self, ctx, x, y):
...
+
+        return _DHParameters(self, dh_param_cdata)
+
+    def generate_dh_private_key(self, parameters):
+        dh_key_cdata = self._lib.DH_new()
+        assert dh_key_cdata != self._ffi.NULL
+        dh_key_cdata = self._ffi.gc(dh_key_cdata, self._lib.DH_free)
+
+        dh_key_cdata.p = self._lib.BN_dup(parameters._dh_cdata.p)
+        dh_key_cdata.g = self._lib.BN_dup(parameters._dh_cdata.g)
```
reaperhulk: This section could be simplified by using `DH *DHparams_dup(DH *);` to get a new `DH *` that already has the duplicated parameters. This was a macro prior to 1.0.0 so you'll want to put it in MACROS.

cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1055,6 +1063,77 @@ def _ec_key_set_public_key_affine_coordinates(self, ctx, x, y):
...
+
+        dh_cdata.p = self._int_to_bn(parameter_numbers.p)
+        dh_cdata.g = self._int_to_bn(parameter_numbers.g)
+        dh_cdata.pub_key = self._int_to_bn(numbers.public_numbers.y)
+        dh_cdata.priv_key = self._int_to_bn(numbers.x)
+
+        codes = self._ffi.new("int[]", 1)
+        res = self._lib.DH_check(dh_cdata, codes)
+        assert res == 1
+        assert codes[0] in (0, 6, 2, 4)
```
reaperhulk: The [docs](https://www.openssl.org/docs/crypto/DH_generate_parameters.html) say:

"DH_check() validates Diffie-Hellman parameters. It checks that p is a safe prime, and that g is a suitable generator. In the case of an error, the bit flags DH_CHECK_P_NOT_SAFE_PRIME or DH_NOT_SUITABLE_GENERATOR are set in *codes. DH_UNABLE_TO_CHECK_GENERATOR is set if the generator cannot be checked, i.e. it does not equal 2 or 5."

The bit flags are in `dh.h` with these values
```c
/* DH_check error codes */
# define DH_CHECK_P_NOT_PRIME            0x01
# define DH_CHECK_P_NOT_SAFE_PRIME       0x02
# define DH_UNABLE_TO_CHECK_GENERATOR    0x04
# define DH_NOT_SUITABLE_GENERATOR       0x08
# define DH_CHECK_Q_NOT_PRIME            0x10
# define DH_CHECK_INVALID_Q_VALUE        0x20
# define DH_CHECK_INVALID_J_VALUE        0x40
```

0 definitely looks safe, but why would 2, 4, or 6 (2 | 4) be okay?

alex: Why are we using the integers in the assert, instead of binding those
values?

On Tue, Mar 3, 2015 at 6:41 AM, Paul Kehrer <notifications@github.com>
wrote:

> In cryptography/hazmat/backends/openssl/backend.py
> <https://github.com/pyca/cryptography/pull/1279#discussion_r25691575>:
>
> > +    def load_dh_private_numbers(self, numbers):
> > +        parameter_numbers = numbers.public_numbers.parameter_numbers
> > +
> > +        dh_cdata = self._lib.DH_new()
> > +        assert dh_cdata != self._ffi.NULL
> > +        dh_cdata = self._ffi.gc(dh_cdata, self._lib.DH_free)
> > +
> > +        dh_cdata.p = self._int_to_bn(parameter_numbers.p)
> > +        dh_cdata.g = self._int_to_bn(parameter_numbers.g)
> > +        dh_cdata.pub_key = self._int_to_bn(numbers.public_numbers.y)
> > +        dh_cdata.priv_key = self._int_to_bn(numbers.x)
> > +
> > +        codes = self._ffi.new("int[]", 1)
> > +        res = self._lib.DH_check(dh_cdata, codes)
> > +        assert res == 1
> > +        assert codes[0] in (0, 6, 2, 4)
>
> The docs <https://www.openssl.org/docs/crypto/DH_generate_parameters.html>
> say:
>
> "DH_check() validates Diffie-Hellman parameters. It checks that p is a
> safe prime, and that g is a suitable generator. In the case of an error,
> the bit flags DH_CHECK_P_NOT_SAFE_PRIME or DH_NOT_SUITABLE_GENERATOR are
> set in *codes. DH_UNABLE_TO_CHECK_GENERATOR is set if the generator cannot
> be checked, i.e. it does not equal 2 or 5."
>
> The bit flags are in dh.h with these values
>
> /* DH_check error codes */
> # define DH_CHECK_P_NOT_PRIME            0x01
> # define DH_CHECK_P_NOT_SAFE_PRIME       0x02
> # define DH_UNABLE_TO_CHECK_GENERATOR    0x04
> # define DH_NOT_SUITABLE_GENERATOR       0x08
> # define DH_CHECK_Q_NOT_PRIME            0x10
> # define DH_CHECK_INVALID_Q_VALUE        0x20
> # define DH_CHECK_INVALID_J_VALUE        0x40
>
> 0 definitely looks safe, but why would 2, 4, or 6 (2 | 4) be okay?
>
> —
> <https://github.com/pyca/cryptography/pull/1279/files#r25691575>.
>



-- 

cryptography/hazmat/backends/openssl/backend.py
reaperhulk: Don't use this. See #1720

cryptography/hazmat/backends/openssl/backend.py
public: Could go in another PR :)

cryptography/hazmat/backends/openssl/backend.py
public: OK :)

cryptography/hazmat/backends/openssl/backend.py
public: Probably just me being lazy many months ago. Will add them to the bindings.

markrwilliams:
@public It looks like you put a lot of good effort into the `DHBackend` implementation in this PR.  I see that the interfaces made it in through [a separate PR](https://github.com/pyca/cryptography/pull/1795).

Was there anything about the implementation here that caused you to close this?  Is it a good idea for an interested party to take this over?

PR #1316, opened 2014-08-24
DH key interfaces

public:
Adds the DH*Key interfaces and docs.

### DIFF
--- a/docs/hazmat/primitives/asymmetric/dh.rst
+++ b/docs/hazmat/primitives/asymmetric/dh.rst
@@ -6,6 +6,9 @@ Diffie-Hellman key exchange
 .. currentmodule:: cryptography.hazmat.primitives.asymmetric.dh
 
 
+Numbers
+~~~~~~~
+
 .. class:: DHPrivateNumbers(x, public_numbers)
 
     .. versionadded:: 0.8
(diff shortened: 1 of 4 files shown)


docs/hazmat/primitives/interfaces.rst
```diff
@@ -696,6 +696,115 @@ Key derivation functions
...
+        Return the numbers that make up this set of parameters.
+
+        :return: A :class:`~cryptography.hazmat.primitives.asymmetric.dh.DHParameterNumbers`.
+
+
+.. class:: DHPrivateKey
+
+    .. versionadded:: 0.8
+
+    .. attribute:: key_size()
```
reaperhulk: In #1478 we're having a discussion about our use of `key_size` where better descriptors would be appropriate. Is key size the right thing to use here?

reaperhulk:
Since we're splitting the interfaces we should move them in this PR too.

public:
Yeah, was waiting for the existing interfaces to start getting moved first.

docs/hazmat/primitives/interfaces.rst
public: Still no consensus on this so I'll leave it alone for now :)

alex:
Not sure I see why tests are failing, might need a merge/rebase now that the classes are landed.

public:
Rebased.

reaperhulk:
As of #1503 we are starting to rename our numbers interfaces to "with serialization" so we should probably start these with the new name.

reaperhulk:
This question is perhaps more properly suited for the implementation but could you show me an example of how someone will use this API to do a handshake?

public:
Well #1503 isn't merged yet and I'm not sure I like tying Numbers and Serialization together like that :)

#1279 has examples of how the API works in the tests https://github.com/pyca/cryptography/pull/1279/files#diff-e59233a048e6ea76b0f12f336aa5d109R135

alex:
#1503 is now merged :-)

Same question as @reaperhulk about what it looks like to use this in practice. Also, I'm not sure I understand what the DH exchange algorithm is?

public:
So initially I added this as a way to try and support the wide array of different DH modes out there. However going back over the NIST docs now I'm not entirely sure this API is suitable for all of the modes. (For example dhHybrid1 requires 2 sets of keys on each end.) So it probably makes sense to remove this in favour of having different `KeyExchangeContext`s for different DH variants instead of trying to jam them all through the same interface.

If someone else could have a read of the relevent bit of SP-80056B that would be helpful :)

alex:
Is there a particular section you think we should read :-)

On Sun, Mar 1, 2015 at 8:32 AM, Alex Stapleton <notifications@github.com>
wrote:

> So initially I added this as a way to try and support the wide array of
> different DH modes out there. However going back over the NIST docs now I'm
> not entirely sure this API is suitable for all of the modes. (For example
> dhHybrid1 requires 2 sets of keys on each end.) So it probably makes sense
> to remove this in favour of having different KeyExchangeContexts for
> different DH variants instead of trying to jam them all through the same
> interface.
>
> If someone else could have a read of the relevent bit of SP-80056B that
> would be helpful :)
>
> —
> <https://github.com/pyca/cryptography/pull/1316#issuecomment-76608545>.
>



-- 

public:
Sorry I meant 56A :-/ http://csrc.nist.gov/groups/ST/toolkit/documents/SP800-56Arev1_3-8-07.pdf

DH is mostly covered in section 6. Basically any of the modes that feature more than 2 key pairs are problematic for the API as it stands but ExchangeAlgorithm is suitable for stuff like the MQV agreement modes which are kind of like DH but deprecated by NIST and partended by Certicom.

reaperhulk:
So instead of a generic KeyExchangeContext (that then begets a DHKeyExchangeContext, which takes a type) you'd make a DHTLSKeyExchange context? Does that mean you'd also want a DHETLSKeyExchange context?

I took a look at SP800-56A and it reminded me that NIST has difficulty being proscriptive. There are 14 DH variants listed. That is ridiculous.

reaperhulk:
And when I say ridiculous I mean we should not strive to support them all. We should support the subset of DH key agreement variants that are actually in serious use (e.g. DH, DHE, ECDH, ECDHE in TLS) and worry about the rest when someone brings us a concrete use case.

public:
I am fine with that :) The main reason I attempted to take it into account was that I know at least 1 implementation of MQV, Crypto++ has it. I don't know any protocols that really exploit the asymmetric DH configurations though.

I'll tidy up this and backend PR.

public:
Elimated DHExchangeAlgorithm.

reaperhulk:
retest this please

PR #1390, opened 2014-10-08
[WIP]: Move to pure python code for parsing PEMs.

alex:
OMFG. Not final.

Implements:

* the core "PEM" parser logic
* PEM decryption
* Handling of the ASN.1 DER for ECDSA Private Keys

Not done:

* Parsing of a bunch of other key types
* non-md5 based PBES

Some notes:

* I don't understand the error handling around ASN.1 well enough yet, which
  makes me pretty uncomfortable given the sensitivity of this stuff.
* PyASN.1's API is kind of miserable, but it does appear to be complete
* There's some not-great error handling in (at least) the EC number loading in
  the OpenSSL backend
* Bits vs. Bytes for key_size is the worst :-(
* If we're comitted to doing this, it's going to be a decent sized project.
  We'll have to keep the old key loading logic around and port one key type at
  a time.
* I don't know if I read the RFC closely enough :-( There's all kinds of random
  notes in various places, and I assume if I ignore any one of those the
  security of the system is screwed.

FEEDBACK PLEASE!!! :-)


reaperhulk:
I think we should close this and if we choose to revisit it in the future a new PR can be opened. Let me know if you disagree @alex :)

alex:
That's fine.

On Wed, Sep 30, 2015 at 12:23 AM, Paul Kehrer <notifications@github.com>
wrote:

> I think we should close this and if we choose to revisit it in the future
> a new PR can be opened. Let me know if you disagree @alex
> <https://github.com/alex> :)
>
> —
> <https://github.com/pyca/cryptography/pull/1390#issuecomment-144277566>.
>



-- 

PR #1456, opened 2014-10-31
[WIP] Scrypt

Ayrx:
This should be a working implementation of the Scrypt primitive and a libscrypt based backend. 

Few notes:
1. The Scrypt source and paper refers to 3 parameters (N, p and r). I have named these `work_factor`, `block_size` and `parallelization_factor` respectively. We may want to debate the naming for these parameters.

2. The libscrypt in Ubuntu's repository is borked at the moment (ref: https://bugs.launchpad.net/ubuntu/+source/libscrypt/+bug/1313311), I'm not sure who to bug to get the fix in.

3. The libscrypt backend gets tested automatically when installed (refer to the tox.ini file to see what I mean), @dstufft has some views against this that we need to discuss.


reaperhulk:
We're still a long way from understanding what we want to do regarding these types of backends so I'm going to go ahead and close this for now. We can revisit when the time comes.

Issue #1522, opened 2014-12-13
Where should our backends live?


Ayrx:
There have been discussions in IRC about whether our optional backends (ala my scrypt PR in #1456) should live in the primary cryptography git repository or if it should be in a separate repository. Rather than have the discussion be lost in IRC chat logs, let's document our thoughts in a more visible place.

I know @dstufft has strong opinions on this matter so rather than paraphrasing what he said on IRC, I'll leave it to him to argue his stand. :) Anyone with opinions on this matter should voice them out as well as I think it will be a major pain to change later.

dstufft:
Ok, so I want to bring this discussion back around.

Thinking about this more, I think the only sane way to really handle this is to split out each backend into their own wholly independent package. This would mean that we'd have the following packages:

* cryptography
* cryptography-openssl
* cryptography-commoncrypto
* cryptography-vectors

Doing this would require moving each of the backends to their own import names, probably with some backwards compatibility shims in the cryptography project so that people using the bindings directly don't break (but ideally with a deprecation warning telling people to switch to using the appropriate binding project directly).

This will immediately simplify the logic around under what cases a particular backend is built (always) and we'd be able to rely directly on the packaging toolchains support for non-optional inclusion of things like a commoncrypto backend (or perhaps a Windows crypto backend).

If I recall the last time we had this discussion there was some talk about how Django's database backends work by having a few included within Django itself and then third parties can register their own. I believe that this analogy is flawed because these backends are not like the Django database backends, but are more akin to the database drivers that those backends use (like psycopg2). This is because there is practically no cost incurred to have the PostgreSQL backend installed in Django (other than minor amounts of disk space) since it is pure Python, relies on nothing in the environment unless you actually use it, and doesn't affect the output of building Django. On the other hand, the OpenSSL or CommonCrypto or Scrypt or any other backend in cryptography does not have similar properties. Installing the CommonCrypto backend requires that CommonCrypto be available and the output of the build process (which defines what cryptography "is") changes based on what the answer to that question is. In other words, it would be more like trying to include psycopg2 into Django itself.

Each of these projects needs to be, as far as packaging is concerned, completely independent. That means that they cannot share any files or resources at the filesystem level (such as the ``README.rst`` issue that @Ayrx had above). This includes things like each project having it's own tox file (because tox does not really have any means of installing multiple local packages).

[...]

PR #1583, opened 2014-12-28
Move cryptography package one level down.

### LINKED ISSUE #1473: Better handle multiple packages
Right now the cryptography repository has two packages inside of it. It has the "primary" package of cryptography which is located at `.`, and then a "secondary" package of cryptography_vectors. In #1456a third one is added which is cryptography_scrypt.

This organization is confusing, we have packages inside of packages. I think we should either decide if we want to hold multiple packages in one repository and if we do we should reorganize the repository to not do this nesting thing.

To be exact the proposal is to change the directory structure to:

```
.
├── cryptography
│   ├── setup.py
│   └── src
│       └── cryptography
├── cryptography_scrypt
│   ├── cryptography_scrypt
│   └── setup.py
└── cryptography_vectors
    ├── cryptography_vectors
    └── setup.py
```

If we're going to hold everything in a single repository. If we're not going to hold everything in a single repository then instead of having a single repository at `.` each direct descendent (cryptography, cryptography_scrypt, cryptography_vectors) would have it's own repository.

Personally I think it sort of makes sense sense to have cryptography and cryptography_vectors in the same repository (primarily because the split really only exists because of file size), but I do not think it makes sense to include other random backends in the same repository. Overall I think putting multiple things into the same repository isn't a very good way to organize things. Pip has some newish support for layouts like this, but I don't believe anything else does.

My vote is to stick to one repository per package.

Ayrx:
Moves the cryptography package one level down.

Fixes #1473.  Ref #1522.


public:
Windows supports symlinks as long as you aren't using FAT32 right?

Ayrx:
@public I have done some research and it seems like msysgit (what looks like the most popular git thing used in Windows-land?) [doesn't seem to support symlinks][0]. Looks like that's a no-go.

[0]: https://www.quora.com/How-does-git-handle-symbolic-links-on-Windows

reaperhulk:
We need a final answer on README/AUTHORS before we can merge this.

Ayrx:
@reaperhulk As far as the README goes each package will have a package-specific README once I factor out the primitives and interfaces into another package. I think there's little harm just duplicating the file at the moment.

I'll need an answer on what to do with the AUTHORS files though. :)

So I see three options.
1. Moving it back into the root and leaving it out of the sdist.
2. Autogenerate it. (Not sure how to do this though, I'll need someone to point me at an example...)
3. Some packaging voodoo I don't know about so it can stay in the root and still go into the sdist.

reaperhulk:
The release script could copy it from root probably?

PR #1592, opened 2014-12-30
Move RSA*Key interfaces to cryptography.hazmat.primitives.asymmetric.rsa

public:


alex:
this lgtm -- two thoughts, a) should we use something like http://paste.openstack.org/show/155299/ so people get warnings, b) @reaperhulk can you also take a look?

public:
I was thinking about doing something like that but was worried about messing up isinstance checks.

src/cryptography/utils.py
```diff
@@ -12,6 +12,8 @@
 # DeprecatedIn07 objects exist. This comment exists to remind developers to
 # look for them when it's time for the ninth release cycle deprecation dance.
 
+# DeprecatedIn08 objects exist. This comment exists to remind developers to
+# look for them when it's time for the tent release cycle deprecation dance.
```
reaperhulk: s/tent/tenth

reaperhulk:
I'd definitely like to raise deprecation warnings if we can. Do we know that it causes problems with `isinstance`?

alex:
Nope, that code shouldn't.

alex:
Bad news. This will need a rebase. I'll put a separate PR in to add the deprecation code.

public:
Or we can do what was done in #1595 and re-import everything in `interfaces.py`?

reaperhulk:
I'd like `cryptography.hazmat.primitives.interfaces` to ultimately go away so we do need to develop a deprecation strategy. #1595 only made the existing file easier to read rather than moving them to their ultimate home.

public:
Added deprecation stuff to this.

public:
Random PyPy/EC failure? Wat?

reaperhulk:
We've had that intermittent failure for several months. The working assumption is a (very, very hard to reliably reproduce) bug in PyPy's gc.

reaperhulk:
You've also got a pep8 problem

src/cryptography/hazmat/primitives/interfaces/__init__.py
```diff
@@ -121,70 +123,34 @@ def copy(self):
...
-    @abc.abstractproperty
-    def key_size(self):
-        """
-        The bit length of the public modulus.
-        """
+RSAPrivateKey = utils.deprecated(
+    rsa.RSAPrivateKey,
+    __name__,
+    "The RSAPrivateKey interface has moved to the cryptography.hazmat.primitives.asymmetric.rsa module",
+    DeprecationWarning
```
reaperhulk: these should be pending deprecation warnings

alex: They should really just use the `DeprecatedIn08` or whatever constants.

public: Done.

reaperhulk:
@public does this fail pypy locally for you? I'm happy that we're apparently reproducible on travis and jenkins, although it's really weird that it only happens on the OS X build on Travis and the 1.0.2 build on jenkins...

@alex is there something you'd like to see for debugging here?

alex:
Debugging the PyPy thing? Something I can stably reproduce outside of
travis :-/ Can't much file a bug without that.

public:
WTF again? I've re-run this multiple times now :(

PR #1593, opened 2014-12-30
Make the symmetric-enc example an example

greghaynes:
Making some minor tweaks to the doc example for symmetric encryption so
it is an actual, runable example.

### DIFF
--- a/docs/hazmat/primitives/symmetric-encryption.rst
+++ b/docs/hazmat/primitives/symmetric-encryption.rst
@@ -35,10 +35,12 @@ in an "encrypt-then-MAC" formulation as `described by Colin Percival`_.
 
     .. doctest::
 
+        >>> import os
         >>> from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
         >>> from cryptography.hazmat.backends import default_backend
         >>> backend = default_backend()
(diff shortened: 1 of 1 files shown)


reaperhulk:
We create a key and iv in a `testsetup` block right now. We should remove that block if we want to make this doctest more clearly runnable.

reaperhulk:
@greghaynes if you get a chance to update this PR with the fixes @alex and myself have requested we can merge it!

greghaynes:
Yep, got a bit caught up with some holiday stuff but ill try and update soon.

Thanks!

reaperhulk:
retest this please

docs/hazmat/primitives/symmetric-encryption.rst
```diff
@@ -35,9 +29,12 @@ in an "encrypt-then-MAC" formulation as `described by Colin Percival`_.
 
     .. doctest::
 
+        >>> import os
         >>> from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
         >>> from cryptography.hazmat.backends import default_backend
         >>> backend = default_backend()
+        >>> key = os.urandom(32)
+        >>> iv = os.urandom(32)
```
reaperhulk: IV should actually be 16 (the block size of AES is 16 bytes). This should fail in the `validate_for_algorithm` check, but we won't know until Travis finishes up :)

Issue #1495, opened 2014-11-25
Split up primitives/interfaces.py into multiple files


alex:
It's getting big, and badly organized. All names will need to be re-exported into `primitives/intefaces/__init__.py`.

reaperhulk:
(No longer relevant due to interface diaspora)

PR #1595, opened 2014-12-30
Start splitting out interfaces with ciphers

greghaynes:
The interfaces module is getting unwieldy so starting to split this out.
Beginning by creating a ciphers module.

This starts implementing issue #1495

### DIFF
--- a/src/cryptography/hazmat/primitives/interfaces/__init__.py
+++ b/src/cryptography/hazmat/primitives/interfaces/__init__.py
@@ -12,6 +12,7 @@ from cryptography.hazmat.primitives.interfaces.ciphers import (
     BlockCipherAlgorithm,
     CipherAlgorithm,
     Mode,
+    ModeWithAuthenticationTag,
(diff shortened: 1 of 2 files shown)


Ayrx:
:+1: This looks merge-ready to me.

reaperhulk:
Do we want to split these into submodules or move the interfaces to live near the primitives they're associated with? (For example, cipher interfaces could live somewhere in `cryptography.hazmat.primitives.ciphers`)

alex:
I think long term we want to move them to by near with the primtivies
they're for, however, we can move them to `interfaces/ciphers.py` and
re-export them in `__init__.py` without leading to any backwards
incompatibility or churn. (That is to say, we don't deprecate the old
locations).

reaperhulk:
I'm okay with that. Thanks for the contribution @greghaynes. I'll merge now. Feel free to send a PR adding your name to the `AUTHORS.rst` file.

PR #1600, opened 2015-01-02
Added a utility for deprecating a module attribute

alex:


alex:
I'm not sure how to test this. If I put `X = deprecated(...)` in a test module, `py.test` doesn't seem to be able to pick it up, and I'm not sure why.

public:
These tests are a lot more complex than mine were. Do we care?

alex:
Not sure!

glyph:
This looks reasonably good to me.  It should be noted that this is more or less a clone of https://github.com/twisted/twisted/blob/trunk/twisted/python/deprecate.py and we should probably factor all this out into a library.

PR #1604, opened 2015-01-02
Add deprecated_name helpers

public:


reaperhulk:
pep8 failure (trailing newlines in `tests/deprecated.py`)

alex:
This is the same as my PR, but with different tests right?

public:
Totally missed your PR :-)

alex:
My PR is #1600 for reference.

Issue #1573, opened 2014-12-23
Support loading of DER private/public keys


reaperhulk:
We've implemented PEM loading, now it's time for DER.

refs #614

- [x] Add DER encoded keys (#1607)
- [x] Add backend interface (#1608)
- [x] Add multibackend support (#1609)
- [ ] Add OpenSSL support + tests

alex:
This is done.

PR #1608, opened 2015-01-04
add DER backend interfaces

reaperhulk:
Backend interface for loading asymmetric DER encoded keys. We need to be able to load private and public keys with the following types:

**Private Key Types**
* PKCS8 unencrypted
* PKCS8 encrypted
* traditional OpenSSL unencrypted (the encrypted form of this is a PEM specific thing so it doesn't exist for DER)

**Public Key Types**
* PKCS8 unencrypted
* traditional OpenSSL unencrypted

It turns out to be pretty easy to make this interface the same as PEM when implementing it so here we go.

refs #1573

### DIFF
--- a/src/cryptography/hazmat/backends/interfaces.py
+++ b/src/cryptography/hazmat/backends/interfaces.py
@@ -233,6 +233,22 @@ class PEMSerializationBackend(object):
 
 
 @six.add_metaclass(abc.ABCMeta)
+class DERSerializationBackend(object):
+    @abc.abstractmethod
+    def load_der_private_key(self, data, password):
+        """
+        Loads a priate key from DER encoded data. Uses the provided password
+        if the data is encrypted.
+        """
+
+    @abc.abstractmethod
+    def load_der_public_key(self, data):
+        """
+        Loads a public key from DER encoded data.
+        """
+
+
+@six.add_metaclass(abc.ABCMeta)
 class X509Backend(object):
     @abc.abstractmethod
     def load_pem_x509_certificate(self, data):
--- a/src/cryptography/hazmat/backends/interfaces.py
+++ b/src/cryptography/hazmat/backends/interfaces.py
@@ -237,7 +237,7 @@ class DERSerializationBackend(object):
     @abc.abstractmethod
     def load_der_private_key(self, data, password):
         """
-        Loads a priate key from DER encoded data. Uses the provided password
+        Loads a private key from DER encoded data. Uses the provided password
         if the data is encrypted.
         """
 
--- a/docs/hazmat/backends/interfaces.rst
+++ b/docs/hazmat/backends/interfaces.rst
@@ -468,6 +468,30 @@ A specific ``backend`` may provide one or more of these interfaces.
             serialized data contains.
         :raises ValueError: If the data could not be deserialized.
 
+.. class:: DERSerializationBackend
+
+    .. versionadded:: 0.8
+
+    A backend with methods for working with DER encoded keys.
+
+    .. method:: load_der_private_key(data, password)
+
+        :param bytes data: DER data to load.
+        :param bytes password: The password to use if the data is encrypted.
+            Should be ``None`` if the data is not encrypted.
+        :return: A new instance of the appropriate type of private key that the
+            serialized data contains.
+        :raises ValueError: If the data could not be deserialized.
+        :raises cryptography.exceptions.UnsupportedAlgorithm: If the data is
+            encrypted with an unsupported algorithm.
+
+    .. method:: load_der_public_key(data)
+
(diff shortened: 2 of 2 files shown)


alex:
Is it possible to detect the difference between the different DER blobs? If so, how?

reaperhulk:
No, but there's only two things to try so you can pretty much do exactly this:

```python
    def load_der_private_key(self, data, password):
        mem_bio = self._bytes_to_bio(data)
        key = self._lib.d2i_PrivateKey_bio(mem_bio.bio, self._ffi.NULL)
        if key != self._ffi.NULL:
            return self._evp_pkey_to_private_key(key)
        else:
            return self._load_key(
                self._lib.d2i_PKCS8PrivateKey_bio,
                self._evp_pkey_to_private_key,
                data,
                password,
            )
```

d2i_PrivateKey_bio is likely to add errors to the stack so the code above needs to be updated to consume those, but otherwise it should be fine.

You could actually streamline this code a bit more and have it not even try to load the traditional key if a password is provided (as only PKCS8 can have a password when DER encoded)

alex:
Is this safe?

reaperhulk:
I'm not aware of any security considerations with this approach (cc @lvh). Security.framework essentially does this when loading as you can provide it no information about what the key is and it still figures out how to load it.

alex:
@reaperhulk A follow question, assuming it's safe to randomly guess at what the contents of a DER encoded blob are. *And* it's possible to distinguish DER from PEM (which I don't think is controversial), should we have a unified API for loading them?

reaperhulk:
I actually vaguely recall proposing a unified API. Assuming @lvh agrees that it's okay to do this I'd think that'd be fine (`load_private_key(data, password, backend)` is nice and simple). Would we want to deprecate `PEMSerializationBackend`? Is there any advantage to retaining the separate methods as well?

ronf:
I like the idea of a unified API for loading any format. It's pretty easy to distinguish PEM vs. DER vs. other variants like OpenSSH or RFC4716 from one another. It's a bit harder to distinguish PKCS8 vs. PKCS1, but it's possible. The hardest one is distinguishing the specific key type in the PKCS1 DER case (RSA vs. DSA vs. EC) since there's absolutely no metadata to help you out, but you can basically look at the exact structure of the ASN.1 to determine the key type. I did this in my AsyncSSH implementation, where i have import_private_key(data, passphrase) and import_public_key(data) functions that will handle all of these formats.

reaperhulk:
So in the interest of moving this forward...do we want to keep these independent APIs and then build a unified one on top or replace the interfaces with a single unified one?

src/cryptography/hazmat/backends/interfaces.py
```diff
@@ -233,6 +233,22 @@ def load_pem_public_key(self, data):
 
 
 @six.add_metaclass(abc.ABCMeta)
+class DERSerializationBackend(object):
+    @abc.abstractmethod
+    def load_der_private_key(self, data, password):
+        """
+        Loads a priate key from DER encoded data. Uses the provided password
```
public: Spelling.

public:
If you parsed the ASN.1 yourself would you be able to do something cleaner than trying both and looking at the errors?

public:
DER has a special place in the security of X.509 not sure how comfortable I am with adding lots of magic around it.

ronf:
Parsing the ASN.1 yourself would definitely let you determine the format of the key and know exactly which decode to perform. However, it does require very detailed knowledge of the different key types, and it felt kind of ugly when I explicitly coded that in AsyncSSH.

To give a feel for this, here's what I look for for the PKCS#1 key format right now:

RSA private: A sequence of all integers, with 9 or more elements (though in fact only 5 of the values are needed)
RSA public: a sequence of exactly two integers
DSA private: A sequence of exactly 6 integers, with the first integer being 0
DSA public: A sequence of exactly 4 integers

The EC case is even more complicated, involving a sequence with a mix of an integer, byte array, and tagged DER objects containing OIDs and bit strings.

reaperhulk:
After some discussion in IRC and some hunting in the OpenSSL codebase we've found a place where OpenSSL does this exact sort of introspection: https://github.com/openssl/openssl/blob/master/crypto/asn1/d2i_pr.c#L130. Leaving the link here for historical purposes :)

alex:
Discussion between @reaperhulk and myself on IRC:

* Both OpenSSL and CommonCrypto appear to have functionality for "guessing" at DER encoded contents -- this supports the position that they're distinguishable unambiguously.
* Separate interface for this, if we to consider a single unified "load a key, I have no earthly idea if it's SSH or PEM or DER" we can do that later,
* This PR is therefore ok.

PR #1612, opened 2015-01-08
X509 distinguished name parsing support in the OpenSSL backend

### LINKED ISSUE #1569: Add support for parsing distinguished names in X.509
We need to be able to parse subject and issuer DNs!

reaperhulk:
Support distiguished names (issuer/subject). Work items:

- [x] Test all supported OIDs (subject & issuer)
- [x] Test unsupported OID (subject & issuer)
- [x] Test attribute values outside ASCII charset
- [x] Add changelog entry
- [x] Add examples to docs

Fixes #1569, depends on #1632, #1645, and #1656

### DIFF
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -195,12 +195,13 @@ X.509 Certificate Object
 
     .. doctest::
 
-        >>> assert len(cert.subject) == 3
-        >>> attributes = []
-        >>> for attribute in cert.subject:
-        ...     attributes.append(attribute)
-        >>> len(attributes)
+        >>> len(cert.subject)
         3
+        >>> for attribute in cert.subject:
+        ...     print(attribute)
+        <NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.6, name=countryName)>, value=u'US')>
+        <NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.10, name=organizationName)>, value=u'Test Certificates 2011')>
+        <NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.3, name=commonName)>, value=u'Good CA')>
 
     .. method:: get_attributes_for_oid(oid)
 
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -166,6 +166,143 @@ X.509 Certificate Object
             >>> cert.not_valid_after
             datetime.datetime(2030, 12, 31, 8, 30)
 
+    .. attribute:: issuer
+
+        .. versionadded:: 0.8
+
+        :type: :class:`Name`
+
+        The :class:`Name` of the issuer.
+
+    .. attribute:: subject
+
+        .. versionadded:: 0.8
+
+        :type: :class:`Name`
+
+        The :class:`Name` of the subject.
+
+
+.. class:: Name
+
+    .. versionadded:: 0.8
+
+    An X509 Name is an ordered list of attributes. The entire list can be
+    obtained with :attr:`attributes` or you can use the helper properties to
+    obtain the specific type you want. Names are sometimes represented as a
+    slash or comma delimited string (e.g. ``/CN=mydomain.com/O=My Org/C=US``).
+
+    .. attribute:: attributes
+
+        :type: :class:`list`
+
+        A list of all the :class:`NameAttribute` objects.
+
+        .. doctest::
+
+            >>> len(cert.subject.attributes)
+            3
+
+    .. attribute:: country_name
+
+        :type: :class:`list`
+
+        A list of country name :class:`NameAttribute` objects.
+
+        .. doctest::
+
+            >>> cert.subject.country_name == [
+            ...    x509.NameAttribute(
+            ...        x509.OID_COUNTRY_NAME,
+            ...        'US'
+            ...    )
+            ... ]
+            True
+
+    .. attribute:: organization_name
+
+        :type: :class:`list`
+
+        A list of organization name :class:`NameAttribute` objects.
+
+    .. attribute:: organizational_unit_name
+
+        :type: :class:`list`
+
+        A list of organizational unit name :class:`NameAttribute` objects.
+
+    .. attribute:: dn_qualifier
+
+        :type: :class:`list`
+
+        A list of DN qualifier :class:`NameAttribute` objects.
+
+    .. attribute:: state_or_province_name
+
+        :type: :class:`list`
+
+        A list of state or province name :class:`NameAttribute` objects.
+
+    .. attribute:: common_name
+
+        :type: :class:`list`
+
+        A list of common name :class:`NameAttribute` objects.
+
+    .. attribute:: serial_number
+
+        :type: :class:`list`
+
+        A list of serial number :class:`NameAttribute` objects. This is not the
+        same as the certificate's serial number.
+
+    .. attribute:: locality_name
+
+        :type: :class:`list`
+
+        A list of locality name :class:`NameAttribute` objects.
+
+    .. attribute:: title
+
+        :type: :class:`list`
+
+        A list of title :class:`NameAttribute` objects.
+
+    .. attribute:: surname
+
+        :type: :class:`list`
+
+        A list of surname :class:`NameAttribute` objects.
+
+    .. attribute:: given_name
+
+        :type: :class:`list`
+
+        A list of given name :class:`NameAttribute` objects.
+
+    .. attribute:: pseudonym
+
+        :type: :class:`list`
+
+        A list of pseudonym :class:`NameAttribute` objects.
+
+    .. attribute:: generation_qualifier
+
+        :type: :class:`list`
+
+        A list of generation qualifier :class:`NameAttribute` objects.
+
+    .. attribute:: domain_component
+
+        :type: :class:`list`
+
+        A list of domain component :class:`NameAttribute` objects.
+
+    .. attribute:: email_address
+
+        :type: :class:`list`
+
+        A list of email address :class:`NameAttribute` objects.
 
 .. class:: Version
 
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -188,7 +188,7 @@ X.509 Certificate Object
     .. versionadded:: 0.8
 
     An X509 Name is an ordered list of attributes. The object is iterable to
-    get every attribute or you can use the helper properties to
+    get every attribute or you can use :meth:`Name.get_attributes_for_oid` to
     obtain the specific type you want. Names are sometimes represented as a
     slash or comma delimited string (e.g. ``/CN=mydomain.com/O=My Org/C=US`` or
     ``CN=mydomain.com, O=My Org, C=US``).
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -187,21 +187,20 @@ X.509 Certificate Object
 
     .. versionadded:: 0.8
 
-    An X509 Name is an ordered list of attributes. The entire list can be
-    obtained with :attr:`attributes` or you can use the helper properties to
+    An X509 Name is an ordered list of attributes. The object is iterable to
+    get every attribute or you can use the helper properties to
     obtain the specific type you want. Names are sometimes represented as a
-    slash or comma delimited string (e.g. ``/CN=mydomain.com/O=My Org/C=US``).
+    slash or comma delimited string (e.g. ``/CN=mydomain.com/O=My Org/C=US`` or
+    ``CN=mydomain.com, O=My Org, C=US``).
 
-    .. attribute:: attributes
+    .. doctest::
 
-        :type: :class:`list`
-
-        A list of all the :class:`NameAttribute` objects.
-
-        .. doctest::
-
-            >>> len(cert.subject.attributes)
-            3
+        >>> assert len(cert.subject) == 3
+        >>> attributes = []
+        >>> for attribute in cert.subject:
+        ...     attributes.append(attribute)
+        >>> len(attributes)
+        3
 
     .. method:: get_attributes_for_oid(oid)
 
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -203,106 +203,17 @@ X.509 Certificate Object
             >>> len(cert.subject.attributes)
             3
 
-    .. attribute:: country_name
+    .. method:: get_attributes_for_oid(oid)
 
-        :type: :class:`list`
+        :param oid: An :class:`ObjectIdentifier` instance.
 
-        A list of country name :class:`NameAttribute` objects.
+        :returns: A list of :class:`NameAttribute` instances that match the
+            OID provided. If nothing matches an empty list will be returned.
 
         .. doctest::
 
-            >>> cert.subject.country_name == [
-            ...    x509.NameAttribute(
-            ...        x509.OID_COUNTRY_NAME,
-            ...        'US'
-            ...    )
-            ... ]
-            True
-
-    .. attribute:: organization_name
-
-        :type: :class:`list`
-
-        A list of organization name :class:`NameAttribute` objects.
-
-    .. attribute:: organizational_unit_name
-
-        :type: :class:`list`
-
-        A list of organizational unit name :class:`NameAttribute` objects.
-
-    .. attribute:: dn_qualifier
-
-        :type: :class:`list`
-
-        A list of DN qualifier :class:`NameAttribute` objects.
-
-    .. attribute:: state_or_province_name
-
-        :type: :class:`list`
-
-        A list of state or province name :class:`NameAttribute` objects.
-
-    .. attribute:: common_name
-
-        :type: :class:`list`
-
-        A list of common name :class:`NameAttribute` objects.
-
-    .. attribute:: serial_number
-
-        :type: :class:`list`
-
-        A list of serial number :class:`NameAttribute` objects. This is not the
-        same as the certificate's serial number.
-
-    .. attribute:: locality_name
-
-        :type: :class:`list`
-
-        A list of locality name :class:`NameAttribute` objects.
-
-    .. attribute:: title
-
-        :type: :class:`list`
-
-        A list of title :class:`NameAttribute` objects.
-
-    .. attribute:: surname
-
-        :type: :class:`list`
-
-        A list of surname :class:`NameAttribute` objects.
-
-    .. attribute:: given_name
-
-        :type: :class:`list`
-
-        A list of given name :class:`NameAttribute` objects.
-
-    .. attribute:: pseudonym
-
-        :type: :class:`list`
-
-        A list of pseudonym :class:`NameAttribute` objects.
-
-    .. attribute:: generation_qualifier
-
-        :type: :class:`list`
-
-        A list of generation qualifier :class:`NameAttribute` objects.
-
-    .. attribute:: domain_component
-
-        :type: :class:`list`
-
-        A list of domain component :class:`NameAttribute` objects.
-
-    .. attribute:: email_address
-
-        :type: :class:`list`
-
-        A list of email address :class:`NameAttribute` objects.
+            >>> cert.subject.get_attributes_for_oid(x509.OID_COMMON_NAME)
+            [<NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.3, name=commonName)>, value=u'Good CA')>]
 
 .. class:: Version
 
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -91,3 +91,45 @@ class _Certificate(object):
             )
         ).decode("ascii")
         return datetime.datetime.strptime(time, "%Y%m%d%H%M%SZ")
+
+    @property
+    def issuer(self):
+        issuer = self._backend._lib.X509_get_issuer_name(self._x509)
+        assert issuer != self._backend._ffi.NULL
+        return self._build_x509_name(issuer)
+
+    @property
+    def subject(self):
+        subject = self._backend._lib.X509_get_subject_name(self._x509)
+        assert subject != self._backend._ffi.NULL
+        return self._build_x509_name(subject)
+
+    def _build_x509_name(self, x509_name):
+        count = self._backend._lib.X509_NAME_entry_count(x509_name)
+        attributes = []
+        for x in range(0, count):
+            entry = self._backend._lib.X509_NAME_get_entry(x509_name, x)
+            obj = self._backend._lib.X509_NAME_ENTRY_get_object(entry)
+            assert obj != self._backend._ffi.NULL
+            data = self._backend._lib.X509_NAME_ENTRY_get_data(entry)
+            assert data != self._backend._ffi.NULL
+            buf = self._backend._ffi.new("unsigned char **")
+            res = self._backend._lib.ASN1_STRING_to_UTF8(buf, data)
+            assert buf[0] != self._backend._ffi.NULL
+            buf = self._backend._ffi.gc(
+                buf, lambda buf: self._backend._lib.OPENSSL_free(buf[0])
+            )
+            value = self._backend._ffi.buffer(buf[0], res)[:].decode('utf8')
+            buf_len = 50
+            buf = self._backend._ffi.new("char[]", buf_len)
+            res = self._backend._lib.OBJ_obj2txt(buf, buf_len, obj, 1)
+            assert res > 0
+            oid = self._backend._ffi.buffer(buf, res)[:].decode()
+
+            attributes.append(
+                x509.NameAttribute(
+                    x509.ObjectIdentifier(oid), value
+                )
+            )
+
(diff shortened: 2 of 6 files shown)


docs/x509.rst
```diff
@@ -166,6 +166,155 @@ X.509 Certificate Object
...
+
+        :type: str
+
+        The attribute OID (e.g. ``2.5.4.3``)
+
+    .. attribute:: name
+
+        :type: str
+
+        The name of the OID (e.g. ``commonName``)
```
alex: I think you should use quotes both here and in `oid` to make it clear they're strings.

alex: Also, what even is a str; does this mean they are unicode on python3?

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -91,3 +110,46 @@ def _parse_asn1_time(self, asn1_time):
...
+            )
+            value = self._backend._ffi.buffer(buf[0], res)[:].decode('utf8')
+            buf_len = 50
+            buf = self._backend._ffi.new("char[]", buf_len)
+            res = self._backend._lib.OBJ_obj2txt(buf, buf_len, obj, 1)
+            assert res > 0
+            oid = self._backend._ffi.buffer(buf, res)[:].decode()
+            try:
+                attribute_name = _OID_TO_NAME[oid]
+            except AttributeError:
```
alex: This is definitely wrong, for one, that'll raise a `KeyError`.

reaperhulk: :disappointed:

docs/x509.rst
public: `six.text_type`?

alex:
Ready for a rebase :-)

src/cryptography/x509.py
```diff
@@ -29,6 +32,92 @@ def __init__(self, msg, parsed_version):
         self.parsed_version = parsed_version
 
 
+class UnknownAttribute(Exception):
+    pass
+
+
+Attribute = collections.namedtuple("Attribute", ["oid", "name", "value"])
```
alex: Let's use a "normal" class instead of a namedtuple; I dislike them for public APIs because it's a very leak abstraction; e.g. `Attribute("a", "b", "c") == ("a", "b", "c")` or any other namedtuple type.

alex:
Merge conflict :-(

alex:
retest this please

alex:
So, a few high level issues I want to talk about before I review the patch:

* OIDs. These are probably going to pop up in quite a few places. Do we want to just use the dot-separated string representation for them, or should they have their own type?
* `Name` and its attributes, specifically, they're all lists, which are mutable, the idea of this class is to be immutable, purely data. I don't want to give people the idea they can `.append()` or whatever to alter them. Thoughts? One approach would be to, instead of caching them all, recompute them on access?
* Are all the OID names based on their entry in the OID registry?

reaperhulk:
We are definitely going to end up with a lot of OIDs. There are two forms people typically want them in: dot separated string and binary. Binary is unusual but if we want to support it in the future we should go ahead and make an OID object.

How about returning tuples instead of lists?

The OID names are based on the names defined in the OID registry, yes. There are quite a few of them though so any review should check me on that.

alex:
I suspect the other form someone might want an OID in is a sequence of
integers?

I dislike tuples because they're semantically incorrect: a tuple is a
record (in the RDBMS sense), not a sequence of homogenous items.

reaperhulk:
Would be nice if Python had an actual immutable list ;)

reaperhulk:
Right now the `Name` object is backend independent. If we choose to re-parse on access we would need to make it backend specific. Pretty ugly.

alex:
Nope! It has the full list of attributes, they just because `[a for a in
attributes if a.oid == "I am a super cool oid"]`

reaperhulk:
Ah, so always iterate a copy. Got it. Feels pretty inelegant but definitely would stop mutation :)

reaperhulk:
Do people think we should have a docs example for accessing each property on `Name` or is one sufficient?

docs/x509.rst
```diff
@@ -166,6 +166,180 @@ X.509 Certificate Object
...
+
+
+.. class:: Name
+
+    .. versionadded:: 0.8
+
+    An X509 Name is an ordered list of attributes. The entire list can be
+    obtained with :attr:`attributes` or you can use the helper properties to
+    obtain the specific type you want.
+
```
alex: Should we mention that they're sometimes rendered with the `CN=abc/OU=abc/O=whatever` format?

src/cryptography/x509.py
```diff
@@ -29,6 +31,147 @@ def __init__(self, msg, parsed_version):
...
+        )
+
+    def __ne__(self, other):
+        return not self == other
+
+
+class ObjectIdentifier(object):
+    def __init__(self, oid, name):
+        self._value = oid
+        self._name = name
```
alex: Should this fill in name automatically if it's in `_OID_NAME` dict?

src/cryptography/x509.py
```diff
@@ -29,6 +31,147 @@ def __init__(self, msg, parsed_version):
...
+
+    value = utils.read_only_property("_value")
+    name = utils.read_only_property("_name")
+
+
+class Name(object):
+    def __init__(self, attributes):
+        self._attributes = attributes
+
+    def _build_list(self, oid):
```
alex: This is not a great name for a function; it doesn't say what it does at all.

src/cryptography/x509.py
```diff
@@ -29,6 +31,147 @@ def __init__(self, msg, parsed_version):
...
+
+    @property
+    def email_address(self):
+        return self._build_list(
+            ObjectIdentifier("1.2.840.113549.1.9.1", "emailAddress")
+        )
+
+    @property
+    def attributes(self):
+        return [i for i in self._attributes]
```
alex: `self._attributes[:]`

src/cryptography/x509.py
reaperhulk: If we choose to do that we'd need to handle these cases:

* Known OID (name not provided)
* Known OID (name provided and matches)
* Known OID (name provided and does not match)
* Unknown OID (name provided)
* Unknown OID (name not provided)

Any others/

alex: How about an intermediary state:

* `name` isn't part of the public API for an OID.
* `repr()` of an OID contains the name if it's known
* Constructor for OIDs isn't public so we can punt on this :-)

On Tue, Jan 27, 2015 at 9:50 PM, Paul Kehrer <notifications@github.com>
wrote:

> In src/cryptography/x509.py
> <https://github.com/pyca/cryptography/pull/1612#discussion_r23662579>:
>
> > +        if not isinstance(other, Attribute):
> > +            return NotImplemented
> > +
> > +        return (
> > +            self.oid == other.oid and
> > +            self.value == other.value
> > +        )
> > +
> > +    def __ne__(self, other):
> > +        return not self == other
> > +
> > +
> > +class ObjectIdentifier(object):
> > +    def __init__(self, oid, name):
> > +        self._value = oid
> > +        self._name = name
>
> If we choose to do that we'd need to handle these cases:
>
>    - Known OID (name not provided)
>    - Known OID (name provided and matches)
>    - Known OID (name provided and does not match)
>    - Unknown OID (name provided)
>    - Unknown OID (name not provided)
>
> Any others/
>
> —
> <https://github.com/pyca/cryptography/pull/1612/files#r23662579>.
>



-- 

alex: And (optionally) some public constants for known oids?

On Tue, Jan 27, 2015 at 9:56 PM, Alex Gaynor <alex.gaynor@gmail.com> wrote:

> How about an intermediary state:
>
> * `name` isn't part of the public API for an OID.
> * `repr()` of an OID contains the name if it's known
> * Constructor for OIDs isn't public so we can punt on this :-)
>
> On Tue, Jan 27, 2015 at 9:50 PM, Paul Kehrer <notifications@github.com>
> wrote:
>
>> In src/cryptography/x509.py
>> <https://github.com/pyca/cryptography/pull/1612#discussion_r23662579>:
>>
>> > +        if not isinstance(other, Attribute):
>> > +            return NotImplemented
>> > +
>> > +        return (
>> > +            self.oid == other.oid and
>> > +            self.value == other.value
>> > +        )
>> > +
>> > +    def __ne__(self, other):
>> > +        return not self == other
>> > +
>> > +
>> > +class ObjectIdentifier(object):
>> > +    def __init__(self, oid, name):
>> > +        self._value = oid
>> > +        self._name = name
>>
>> If we choose to do that we'd need to handle these cases:
>>
>>    - Known OID (name not provided)
>>    - Known OID (name provided and matches)
>>    - Known OID (name provided and does not match)
>>    - Unknown OID (name provided)
>>    - Unknown OID (name not provided)
>>
>> Any others/
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/pyca/cryptography/pull/1612/files#r23662579>.
>>
>
>
>
> --
> "I disapprove of what you say, but I will defend to the death your right
> to say it." -- Evelyn Beatrice Hall (summarizing Voltaire)
> "The people's good is the highest law." -- Cicero
> GPG Key fingerprint: 125F 5C67 DFE9 4084
>



-- 

reaperhulk: If we remove name as a public property then I'd definitely say we need a set of constants (e.g. `OID_COMMON_NAME`) so users can do equality checks. I'll add a repr and OID constants and figure out some way to document them.

reaperhulk:
I am going to split this PR today. Should at least be able to pull out the Attribute/ObjectIdentifier classes along with those docs so we can shrink this from its current huge size.

alex:
Awesome <3

On Thu, Jan 29, 2015 at 11:11 AM, Paul Kehrer <notifications@github.com>
wrote:

> I am going to split this PR today. Should at least be able to pull out the
> Attribute/ObjectIdentifier classes along with those docs so we can shrink
> this from its current huge size.
>
> —
> <https://github.com/pyca/cryptography/pull/1612#issuecomment-72051726>.
>



-- 

alex:
This can be rebased now.

docs/x509.rst
```diff
@@ -166,6 +166,143 @@ X.509 Certificate Object
...
+        :type: :class:`list`
+
+        A list of all the :class:`NameAttribute` objects.
+
+        .. doctest::
+
+            >>> len(cert.subject.attributes)
+            3
+
+    .. attribute:: country_name
```
alex: This, and other attributes are singulars, where it seems like they should be plurals? They're returning a list after all.

docs/x509.rst
```diff
@@ -166,6 +166,143 @@ X.509 Certificate Object
...
+
+        :type: :class:`list`
+
+        A list of country name :class:`NameAttribute` objects.
+
+        .. doctest::
+
+            >>> cert.subject.country_name == [
```
alex: This has a reasonable repr, any reason not to write this as:

```rst
>>> cert.subject.country_name
[<NameAttribute whatever the hell goes here>]
```

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -91,3 +91,45 @@ def _parse_asn1_time(self, asn1_time):
...
+    @property
+    def subject(self):
+        subject = self._backend._lib.X509_get_subject_name(self._x509)
+        assert subject != self._backend._ffi.NULL
+        return self._build_x509_name(subject)
+
+    def _build_x509_name(self, x509_name):
+        count = self._backend._lib.X509_NAME_entry_count(x509_name)
+        attributes = []
+        for x in range(0, count):
```
alex: This can just be `range(count)`, no need for the `0`.

src/cryptography/x509.py
```diff
@@ -98,6 +98,87 @@ def __repr__(self):
...
+    def domain_component(self):
+        return self._filter_attr_list(OID_DOMAIN_COMPONENT)
+
+    @property
+    def email_address(self):
+        return self._filter_attr_list(OID_EMAIL_ADDRESS)
+
+    @property
+    def attributes(self):
+        return self._attributes[:]
```
alex: So, thinking outloud here, is there a reason we really need to have all these properties? Is `cert.issuer.common_names` significantly better than `cert.issuer.get_attributes_for_oid(OID_COMMON_NAME)`?

(Thinking outloud, I'm not sure!)

reaperhulk: We've had some conversations in the past about whether `get_extension_for_oid` might be a good API for the extensions. The large volume of convenience properties (many of which are unlikely to be used) and lack of a public way to filter directly by a specific OID that we don't know about (`_filter_attr_list` is private) both make me inclined to say this is a good idea.

alex: Sorry, which direction is a good idea?

On Wed, Feb 11, 2015 at 12:39 PM, Paul Kehrer <notifications@github.com>
wrote:

> In src/cryptography/x509.py
> <https://github.com/pyca/cryptography/pull/1612#discussion_r24520351>:
>
> > +
> > +    @property
> > +    def pseudonym(self):
> > +        return self._filter_attr_list(OID_PSEUDONYM)
> > +
> > +    @property
> > +    def domain_component(self):
> > +        return self._filter_attr_list(OID_DOMAIN_COMPONENT)
> > +
> > +    @property
> > +    def email_address(self):
> > +        return self._filter_attr_list(OID_EMAIL_ADDRESS)
> > +
> > +    @property
> > +    def attributes(self):
> > +        return self._attributes[:]
>
> We've had some conversations in the past about whether
> get_extension_for_oid might be a good API for the extensions. The large
> volume of convenience properties (many of which are unlikely to be used)
> and lack of a public way to filter directly by a specific OID that we don't
> know about (_filter_attr_list is private) both make me inclined to say
> this is a good idea.
>
> —
> <https://github.com/pyca/cryptography/pull/1612/files#r24520351>.
>



-- 

reaperhulk: `cert.issuer.get_attributes_for_oid(OID_COMMON_NAME)` is the good idea.

reaperhulk: I'll revise the PR with that and we can see if we like it. Can always revert if we don't.

alex: Sounds like a plan! I think starting with just attributes_for_oid, and
adding things back as we understand the ergonomics is a good idea.

On Wed, Feb 11, 2015 at 3:27 PM, Paul Kehrer <notifications@github.com>
wrote:

> In src/cryptography/x509.py
> <https://github.com/pyca/cryptography/pull/1612#discussion_r24535007>:
>
> > +
> > +    @property
> > +    def pseudonym(self):
> > +        return self._filter_attr_list(OID_PSEUDONYM)
> > +
> > +    @property
> > +    def domain_component(self):
> > +        return self._filter_attr_list(OID_DOMAIN_COMPONENT)
> > +
> > +    @property
> > +    def email_address(self):
> > +        return self._filter_attr_list(OID_EMAIL_ADDRESS)
> > +
> > +    @property
> > +    def attributes(self):
> > +        return self._attributes[:]
>
> I'll revise the PR with that and we can see if we like it. Can always
> revert if we don't.
>
> —
> <https://github.com/pyca/cryptography/pull/1612/files#r24535007>.
>



-- 

docs/x509.rst
reaperhulk: We don't have a repr for NameAttribute, just ObjectIdentifier. We could add one to NameAttribute, although we'd either need to return it as unicode or decode all values to ascii in the repr.

alex: Oh. Something to think about

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -91,3 +91,45 @@ def _parse_asn1_time(self, asn1_time):
...
+            data = self._backend._lib.X509_NAME_ENTRY_get_data(entry)
+            assert data != self._backend._ffi.NULL
+            buf = self._backend._ffi.new("unsigned char **")
+            res = self._backend._lib.ASN1_STRING_to_UTF8(buf, data)
+            assert buf[0] != self._backend._ffi.NULL
+            buf = self._backend._ffi.gc(
+                buf, lambda buf: self._backend._lib.OPENSSL_free(buf[0])
+            )
+            value = self._backend._ffi.buffer(buf[0], res)[:].decode('utf8')
+            buf_len = 50
```
alex: Is this guarnteed to be enough? Is there a reasonable way to introspect the needed length?

tests/test_x509.py
```diff
@@ -55,6 +55,171 @@ def test_load_der_cert(self, backend):
...
+        assert cert.subject.get_attributes_for_oid(x509.OID_COMMON_NAME) == [
+            x509.NameAttribute(
+                x509.OID_COMMON_NAME,
+                b'We heart UTF8!\xe2\x84\xa2'.decode('utf8')
+            )
+        ]
+        assert cert.issuer.get_attributes_for_oid(x509.OID_COMMON_NAME) == [
+            x509.NameAttribute(
+                x509.OID_COMMON_NAME,
+                b'We heart UTF8!\xe2\x84\xa2'.decode('utf8')
```
alex: Is there not a way we can write these without the explicit decode? Maybe using `six.u`?

src/cryptography/x509.py
```diff
@@ -104,6 +104,27 @@ def __repr__(self):
...
+class Name(object):
+    def __init__(self, attributes):
+        self._attributes = attributes
+
+    def get_attributes_for_oid(self, oid):
+        return [i for i in self._attributes if i.oid == oid]
+
+    @property
+    def attributes(self):
+        return self._attributes[:]
```
alex: What if instead of a `.attributes`, we made `Name` objects be iterable; `for x in subject.issuer:`, does that make sense, or no?

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: The assertion will catch anything that exceeds this limit (and only OIDs way outside the norm would do that) but we can call `int ASN1_STRING_length(ASN1_STRING *x);` instead.

src/cryptography/x509.py
reaperhulk: There aren't alternate representations of this data (other than filtering it via `get_attributes_for_oid` or similar) so I suppose a property isn't required. No objection.

tests/test_x509.py
reaperhulk: `six.u` does appear to work.

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: I'm mistaken. Since this is actually an ASN1_OBJECT (not a string) we can't use that function. `OBJ_obj2txt` converts it to the dotted string form, but there doesn't appear to be a way to obtain the number of bytes this will actually consume.

alex: Perhaps `X509_NAME_MAXLEN` is what we should be using?

On Fri, Feb 13, 2015 at 6:47 PM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/1612#discussion_r24709108>:
>
> > +        count = self._backend._lib.X509_NAME_entry_count(x509_name)
> > +        attributes = []
> > +        for x in range(count):
> > +            entry = self._backend._lib.X509_NAME_get_entry(x509_name, x)
> > +            obj = self._backend._lib.X509_NAME_ENTRY_get_object(entry)
> > +            assert obj != self._backend._ffi.NULL
> > +            data = self._backend._lib.X509_NAME_ENTRY_get_data(entry)
> > +            assert data != self._backend._ffi.NULL
> > +            buf = self._backend._ffi.new("unsigned char **")
> > +            res = self._backend._lib.ASN1_STRING_to_UTF8(buf, data)
> > +            assert buf[0] != self._backend._ffi.NULL
> > +            buf = self._backend._ffi.gc(
> > +                buf, lambda buf: self._backend._lib.OPENSSL_free(buf[0])
> > +            )
> > +            value = self._backend._ffi.buffer(buf[0], res)[:].decode('utf8')
> > +            buf_len = 50
>
> I'm mistaken. Since this is actually an ASN1_OBJECT (not a string) we
> can't use that function. OBJ_obj2txt converts it to the dotted string
> form, but there doesn't appear to be a way to obtain the number of bytes
> this will actually consume.
>
> —
> <https://github.com/pyca/cryptography/pull/1612/files#r24709108>.
>



-- 

alex:
Oh whoops, that's just hte name of some variable cpython and pypy have,
they both use 256 FWIW.

On Fri, Feb 13, 2015 at 7:03 PM, jenkins-cryptography <
notifications@github.com> wrote:

> Test PASSed.
> Refer to this link for build results:
> https://jenkins.cryptography.io/job/cryptography-pr-experimental/2937/
>
> —
> <https://github.com/pyca/cryptography/pull/1612#issuecomment-74353409>.
>



-- 

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: `X509_NAME_MAXLEN` appears to be a constant defined in the CPython source code (value is 256) and is not present in OpenSSL. I'm a bit confused why that was chosen since X.509 Names can be way more than 256 bytes (see our "all_supported_names.pem" cert). It also uses X509_NAME_oneline (https://www.openssl.org/docs/crypto/X509_NAME_print_ex.html) so it looks like it just truncates any name > 256 bytes.

alex: yeah, 256 should be totally reasonable for the dotted string though

docs/x509.rst
```diff
@@ -166,6 +166,53 @@ X.509 Certificate Object
...
+    ``CN=mydomain.com, O=My Org, C=US``).
+
+    .. doctest::
+
+        >>> assert len(cert.subject) == 3
+        >>> attributes = []
+        >>> for attribute in cert.subject:
+        ...     attributes.append(attribute)
+        >>> len(attributes)
+        3
```
alex: Is there any reason not to write this example as:

```pycon
>>> len(cert.subjects)
3
>>> for attribute in cert.subject:
...       print attribute
```

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -91,3 +91,47 @@ def _parse_asn1_time(self, asn1_time):
...
+        count = self._backend._lib.X509_NAME_entry_count(x509_name)
+        attributes = []
+        for x in range(count):
+            entry = self._backend._lib.X509_NAME_get_entry(x509_name, x)
+            obj = self._backend._lib.X509_NAME_ENTRY_get_object(entry)
+            assert obj != self._backend._ffi.NULL
+            data = self._backend._lib.X509_NAME_ENTRY_get_data(entry)
+            assert data != self._backend._ffi.NULL
+            buf = self._backend._ffi.new("unsigned char **")
+            res = self._backend._lib.ASN1_STRING_to_UTF8(buf, data)
```
alex: `assert res >= 0`?

src/cryptography/x509.py
```diff
@@ -104,6 +104,29 @@ def __repr__(self):
...
+        if not isinstance(other, Name):
+            return NotImplemented
+
+        return self._attributes == other._attributes
+
+    def __ne__(self, other):
+        return not self == other
+
+    def __iter__(self):
+        return iter(self._attributes[:])
```
alex: You can drop the copy now.

tests/test_x509.py
```diff
@@ -55,6 +57,183 @@ def test_load_der_cert(self, backend):
...
+            ),
+            x509.load_der_x509_certificate,
+            backend
+        )
+        issuer = cert.issuer
+        assert isinstance(issuer, x509.Name)
+        attributes = []
+        for attrs in issuer:
+            attributes.append(attrs)
+        assert attributes == [
```
alex: Simplifies to `list(issuer)`

alex: Same for all of them.

alex:
Almost there!

docs/x509.rst
reaperhulk: doctest still confuses me and I didn't realize I could print and still do checks against the print. What a weird tool.

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: Done.

docs/x509.rst
```diff
@@ -166,6 +166,54 @@ X.509 Certificate Object
...
+        The :class:`Name` of the subject.
+
+
+.. class:: Name
+
+    .. versionadded:: 0.8
+
+    An X509 Name is an ordered list of attributes. The object is iterable to
+    get every attribute or you can use the helper properties to
+    obtain the specific type you want. Names are sometimes represented as a
```
alex: This sentence is now out of date

alex:
Besides this one sentence, this is ready for merge.

PR #1613, opened 2015-01-08
Bump some years that were missed.

alex:


reaperhulk:
I have a feeling we may need to change the language on these docs that specify a year but for now let's merge. :)

alex:
@reaperhulk yeah, I tried to find a way to make it sound good but not have a year, but couldn't figure it out. If you have language suggestions I'm happy to review :cake:

PR #1615, opened 2015-01-08
Fixed #1614 -- don't cause warnings with the latest setuptools

### LINKED ISSUE #1614: Backend loading code produces a warning with the latest setuptools
The use `load(require=False)` (specifically the `require` kwarg) is deprecated. /cc @dstufft

alex:


reaperhulk:
This change terrifies me in that I feel like we're going to forget why it's like this and it will become Code That Must Not Be Changed™. Could we add a comment about why it's like this and maybe some info about when it might theoretically be safe to remove it?

alex:
Sure, sounds good.

PR #1616, opened 2015-01-09
There's no reason for this text to be monospace

alex:


alex:
I'm not sure, I sent this because the section looked weird in monospace,
personally I'd rather have the words added than the monospace, but I don't
feel strongly.

reaperhulk:
I'm going to close this as wontfix for the moment. I'm +1 on anything we can come up with that allows us to stop using monospace without having to add non-words to the wordlist though.

PR #1617, opened 2015-01-10
Fixed the paths in the go cast verifier.

alex:
Also made some changes to appease go vet and other linters.


alex:
We should probably be running this as a part of our CI, it's been broken for quiet a while.

PR #1618, opened 2015-01-10
Run our vector verification go program

alex:


alex:
I suspect this is going to take a few tries to get everything happy on travis

Ayrx:
This would mean that golang is now a requirement for development? We should add that to the docs then.

alex:
I think travis is configured right, but this is blocked on #1617

reaperhulk:
This seems like a better candidate for a jenkins job than travis to me (since it doesn't affect coverage)

Ayrx:
Jenkins job sounds good.

alex:
@reaperhulk So, if we run this under jenkins, do you still want it in the `tox.ini`?

(It's slightly awkward there, `tox` really wants what you run to be a Python program)

reaperhulk:
I'd be okay with it not being in tox if the tox invocation is weird (which makes sense)

alex:
You can see what it looks like; note that if you run it you get some
warnings because the `go` binary lives outside of the virtualenv; I can't
find a way to make it go away.

public:
I also think it would be better if this was not done in Tox.

alex:
Ok, I'm going to close this then.

PR #1619, opened 2015-01-11
Add support for querying the negotiated TLS/SSL version.

richmoore:
This request makes it possible to see which version of TLS you actually end up with when the connection is negotiated.


Ayrx:
ok to test

alex:
ok to test

PR #1622, opened 2015-01-14
Allow building against openssl configured with no-comp

schlenk:
This patch allows building against an openssl version that was
configured with the no-comp switch, to disable compression (to mitigate
CRIME). (see Issue #1622).

This checks for OPENSSL_NO_COMP and does the appropriate things.

If compression is enabled, the new Cryptography_HAS_COMPRESSION flag is
set.

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/ssl.py
+++ b/src/cryptography/hazmat/bindings/openssl/ssl.py
@@ -32,6 +32,7 @@ static const long Cryptography_HAS_RELEASE_BUFFERS;
  * supported
  */
 static const long Cryptography_HAS_OP_NO_COMPRESSION;
+static const long Cryptography_HAS_COMPRESSION;
 
 static const long Cryptography_HAS_SSL_OP_MSIE_SSLV2_RSA_PADDING;
 static const long Cryptography_HAS_SSL_SET_SSL_CTX;
@@ -189,9 +190,6 @@ int SSL_shutdown(SSL *);
 const char *SSL_get_cipher_list(const SSL *, int);
 Cryptography_STACK_OF_SSL_CIPHER *SSL_get_ciphers(const SSL *);
 
-const COMP_METHOD *SSL_get_current_compression(SSL *);
-const COMP_METHOD *SSL_get_current_expansion(SSL *);
-const char *SSL_COMP_get_name(const COMP_METHOD *);
 
 /*  context */
 void SSL_CTX_free(SSL_CTX *);
@@ -420,6 +418,18 @@ static const long Cryptography_HAS_OP_NO_COMPRESSION = 1;
 #else
 static const long Cryptography_HAS_OP_NO_COMPRESSION = 0;
 const long SSL_OP_NO_COMPRESSION = 0;
+#endif
+
+#ifdef OPENSSL_NO_COMP
+static const long Cryptography_HAS_COMPRESSION = 0;
+const void *SSL_get_current_compression(SSL *s);
+const void *SSL_get_current_expansion(SSL *s);
+const char *SSL_COMP_get_name(const void *comp);
+#else
+static const long Cryptography_HAS_COMPRESSION = 1;
+const COMP_METHOD *SSL_get_current_compression(SSL *);
+const COMP_METHOD *SSL_get_current_expansion(SSL *);
+const char *SSL_COMP_get_name(const COMP_METHOD *);
 #endif
 
 #ifdef SSL_OP_NO_TLSv1_1


schlenk:
The correct issue number is 1622.

Ayrx:
ok to test

alex:
This also needs to be added to the `CONDITIONAL_NAMES` `dict` in that file.

schlenk:
I get errors on import if i add it to CONDITIONAL_NAMES, the delattr() fails if i add it. Not sure why.

alex:
@schlenk ah, you need to have the declarations in the `MACROS` section; take a look at the places you see `TLSv1_1_method` for example.

schlenk:
@alex That does not work either, because openssl/ssl.h declares prototypes for the functions in both cases, e.g. for OPENSSL_NO_COMP defined and undefined, unlike it does for the SSLv3_methods for example. Compare:
```
#ifndef OPENSSL_NO_SSL2
const SSL_METHOD *SSLv2_method(void);		/* SSLv2 */
const SSL_METHOD *SSLv2_server_method(void);	/* SSLv2 */
const SSL_METHOD *SSLv2_client_method(void);	/* SSLv2 */
#endif

#ifndef OPENSSL_NO_SSL3_METHOD
const SSL_METHOD *SSLv3_method(void);		/* SSLv3 */
const SSL_METHOD *SSLv3_server_method(void);	/* SSLv3 */
const SSL_METHOD *SSLv3_client_method(void);	/* SSLv3 */
#endif
```
with 
```
#ifndef OPENSSL_NO_COMP
const COMP_METHOD *SSL_get_current_compression(SSL *s);
const COMP_METHOD *SSL_get_current_expansion(SSL *s);
const char *SSL_COMP_get_name(const COMP_METHOD *comp);
STACK_OF(SSL_COMP) *SSL_COMP_get_compression_methods(void);
int SSL_COMP_add_compression_method(int id,COMP_METHOD *cm);
#else
const void *SSL_get_current_compression(SSL *s);
const void *SSL_get_current_expansion(SSL *s);
const char *SSL_COMP_get_name(const void *comp);
void *SSL_COMP_get_compression_methods(void);
int SSL_COMP_add_compression_method(int id,void *cm);
#endif
```
So maybe those need some different handling here?

reaperhulk:
Yeah this is a different case than our normal conditional bindings in that OpenSSL has chosen to alter the function signatures to return void pointers rather than remove the functions in their entirety. I would have expected `COMP_METHOD` to be undefined when `OPENSSL_NO_COMP` is defined though. Which would necessitate adding a typedef in `CUSTOMIZATIONS`.

This works properly on your OpenSSL?

schlenk:
COMP_METHOD gets defined to void in openssl/ssl.h.

The exact error i get is:
  File "[snip]\site-packages\pyopenssl-0.14-py2.7.egg\OpenSSL\rand.py", line 11, in <module>
    from OpenSSL._util import (
  File "[snip]\site-packages\pyopenssl-0.14-py2.7.egg\OpenSSL\_util.py", line 4, in <module>
    binding = Binding()
  File "[snip]\site-packages\cryptography-0.8.dev1-py2.7-win32.egg\cryptography\hazmat\bindin
gs\openssl\binding.py", line 113, in __init__
    self._ensure_ffi_initialized()
  File "[snip]\site-packages\cryptography-0.8.dev1-py2.7-win32.egg\cryptography\hazmat\bindin
gs\openssl\binding.py", line 123, in _ensure_ffi_initialized
    cls._modules,
  File "[snip]\site-packages\cryptography-0.8.dev1-py2.7-win32.egg\cryptography\hazmat\bindin
gs\utils.py", line 39, in load_library_for_binding
    delattr(lib, name)
AttributeError: SSL_get_current_compression

Seems the ffid Lib doesn't contain the function anyway, so no need to remove it afterwards.

And yes, this works, i can connect to a Twisted via openssl s_client or a browser using this version of cryptography (via pyopenssl).

schlenk:
@reaperhulk Seems there is something strange with the DLL i use (or openssl), as it does not export the SSL_get_current_compression function either, even if it is listed in openssl/ssl.h. Will retry with a static lib and see what it gives.

reaperhulk:
@schlenk When you put them in `CUSTOMIZATIONS` they're no longer exported as functions to the cffi object (that only happens to things defined in `FUNCTIONS` and `MACROS`). Given the difference in the  functions I think we might be able to do something pretty simple here. Let me put up a branch for you to test against in a moment.

schlenk:
Just checked the build with and without 'no-comp' configure switch on windows (using MSVC9).

If i configure with 'no-comp', the symbols for compression are elided from the generated ms/ssleay32.def which defines if they should be included in the DLL. So the ssleay32.dll never contains the SSL_get_current_compression() functions even if they are declared in the header.

Would call openssl/ssl.h's declaration kind of stupid on windows for that case, as it suggests you might link against those functions while they are simply not present at all in the libraries.

reaperhulk:
Yeah that's...not good (and also means my current branch won't work). I wonder if those symbols exist in the OS X/linux objects. Naturally adding `no-comp` is causing a compilation failure for me at the moment on OS X so I'm having trouble determining. I'll build it on linux... If it's consistent then we can still handle this relatively easily.

reaperhulk:
Compiling with `./config no-comp shared`

```
root@precise64:~/openssl-1.0.1k# nm -D libssl.so | grep COMP
0000000000044a20 T SSL_COMP_add_compression_method
0000000000044a10 T SSL_COMP_get_compression_methods
0000000000044a30 T SSL_COMP_get_name
```

So that means we need to handle three cases to have this work properly:

* `OPENSSL_NO_COMP` is not defined.
* `OPENSSL_NO_COMP` is defined and the alternate functions are exported (linux, presumably OS X)
* `OPENSSL_NO_COMP` is defined and the alternate functions are *not* exported (Windows).

Totally ridiculous. I think I have an idea on how to handle this, but it's going to require OS detection in `CUSTOMIZATIONS` (ugh).

schlenk:
Maybe a simple hack would be enough: Do not simply delattr() in utils, but check first via hasattr() if the function exists, and if not, simply do nothing.

reaperhulk:
This digs a bit deeper into the guts of cffi than you maybe are interested in, but cffi learns what it should expose as a function through the `FUNCTIONS` and `MACROS` sections we define (Those are fed to `ffi.cdef`) while `CUSTOMIZATIONS` are fed to the verifier along with some other code. This means anything declared in `FUNCTIONS` or `MACROS` **must** be available. The hacky workaround we have to use to avoid that is to use the preprocessor (which is available in `CUSTOMIZATIONS`) to define functions when we know they don't actually exist. In this case it appears we'll need to basically do something like:

```
#if defined(OPENSSL_NO_COMP) && defined(_WIN32)
const COMP_METHOD (*SSL_get_current_compression)(SSL *) = NULL;
// etc
#endif
```

I'm leaning towards removing these functions even on the linux/OS X side if `OPENSSL_NO_COMP` is defined.

schlenk:
As the functions do not do anything useful (return NULL in all cases), even if defined, that sounds, no use in exporting a function that does nothing.

reaperhulk:
@schlenk mind trying https://github.com/reaperhulk/cryptography/tree/experiment-no-comp and letting me know if it works? I'm guessing this isn't going to work first try, but let's see.

schlenk:
@reaperhulk I'll give it a try.

schlenk:
It works with a little tweak to your patch:
```
- const COMP_METHOD (*SSL_get_current_compression)(SSL *) = NULL;
- const COMP_METHOD (*SSL_get_current_expansion)(SSL *) = NULL;
- const char (*SSL_COMP_get_name)(const COMP_METHOD *) = NULL;
+ const COMP_METHOD* SSL_get_current_compression(SSL * x) { return NULL;};
+ const COMP_METHOD* SSL_get_current_expansion(SSL * x) { return NULL;};
+ const char* SSL_COMP_get_name(const COMP_METHOD * x) { return NULL;};
```
It really wants the implementation, not just some NULL function pointer.

reaperhulk:
Hmm, thanks for checking @schlenk. Not sure why the null function pointers don't work (@alex any ideas?).

I'd like to confirm this works on Linux/OS X before moving forward, but hopefully this is (most of) the solution...

reaperhulk:
My branch works on Linux with one other tweak but I think we need to understand why null function pointers don't work on Windows (if only because we do that in many places in our bindings right now...does that mean if those conditional paths are triggered on Windows the compile fails?)

schlenk:
It should work for the other conditional stuff (also disabled dtls, ssl2 and ssl3 and some algorithms like IDEA), maybe i can see whats up on monday when back at work (and testing through the next iteration of openssl brokenness with 1.0.1l.

reaperhulk:
@schlenk thanks! I updated my branch with support for linux (although it's still using the null function pointers until we have a good answer about why that's causing trouble on Windows for this while it works for the other functions we remove). Hopefully we can get a fix landed in master next week.

reaperhulk:
@schlenk did you have a chance to take a look at this?

schlenk:
Sadly not yet, being swamped with other work currently. :-(

schlenk:
@reaperhulk Still fails to compile for your experiment-no-comp branch. 

d:\code\repos\cryptoreaper\cryptography>d:\code\trunk\win32\debug\img\python setup.py build_ext
running build_ext
building '_Cryptography_cffi_343253bex3a1aa2c0' extension
c:\Program Files (x86)\Microsoft Visual Studio 9.0\VC\BIN\cl.exe /c /nologo /Ox /MD /W3 /GS- /DNDEBUG -Id:\code\trunk\wi
n32\debug\img\include -Id:\code\trunk\win32\debug\img\PC /Tcsrc\cryptography\hazmat\bindings\__pycache__\_Cryptography_c
ffi_343253bex3a1aa2c0.c /Fobuild\temp.win32-2.7\Release\src\cryptography\hazmat\bindings\__pycache__\_Cryptography_cffi_
343253bex3a1aa2c0.obj
_Cryptography_cffi_343253bex3a1aa2c0.c
src\cryptography\hazmat\bindings\__pycache__\_Cryptography_cffi_343253bex3a1aa2c0.c(2116) : error C2365: "SSL_get_curren
t_compression": Erneute Definition; vorherige Definition war "Funktion".
        d:\code\trunk\win32\debug\img\include\openssl/ssl.h(2544): Siehe Deklaration von 'SSL_get_current_compression'
src\cryptography\hazmat\bindings\__pycache__\_Cryptography_cffi_343253bex3a1aa2c0.c(2117) : error C2365: "SSL_get_curren
t_expansion": Erneute Definition; vorherige Definition war "Funktion".
        d:\code\trunk\win32\debug\img\include\openssl/ssl.h(2545): Siehe Deklaration von 'SSL_get_current_expansion'
src\cryptography\hazmat\bindings\__pycache__\_Cryptography_cffi_343253bex3a1aa2c0.c(2118) : error C2365: "SSL_COMP_get_n
ame": Erneute Definition; vorherige Definition war "Funktion".
        d:\code\trunk\win32\debug\img\include\openssl/ssl.h(2546): Siehe Deklaration von 'SSL_COMP_get_name'
src\cryptography\hazmat\bindings\__pycache__\_Cryptography_cffi_343253bex3a1aa2c0.c(2477) : warning C4048: Unterschiedli
che Arraygrößen: 'unsigned char (*)[]' und 'unsigned char (*)[32]'
src\cryptography\hazmat\bindings\__pycache__\_Cryptography_cffi_343253bex3a1aa2c0.c(2478) : warning C4048: Unterschiedli
che Arraygrößen: 'unsigned char (*)[]' und 'unsigned char (*)[32]'
src\cryptography\hazmat\bindings\__pycache__\_Cryptography_cffi_343253bex3a1aa2c0.c(2502) : warning C4048: Unterschiedli
che Arraygrößen: 'unsigned char (*)[]' und 'unsigned char (*)[48]'
error: command 'c:\\Program Files (x86)\\Microsoft Visual Studio 9.0\\VC\\BIN\\cl.exe' failed with exit status 2

schlenk:
My guess at why this fails.

There are usually two forms, how openssl/ssl.h treats those optional methods:
1. Macros that defer to some other function
2. Leaving out the function declaration via ifdef

Both have the effect to not create an entry in the namespace of functions and variables, so the type of the name is still undeclared when the cffi code compiles.

But for the NO_COMP stuff, there IS a declaration, so once you include 'ssl.h' the type is fixed to function. Using the = NULL thing requires the type of the declaration to be variable, so this fails. If they hadn't messed up the exports, this would be a non issue, as the functions are actually defined in the lib, so cryptography would not need to define anything and just link the empty functions.

But as the type is declared, you cannot assign to it. You can only provide a definition that matches the one already there (but hidden).

reaperhulk:
@schlenk that makes sense to me. /cc @alex to get another opinion before we revise this PR. Alex, the proposed fix here would change the conditional declaration from a pointer assignment (which we use everywhere else) to a full function.

alex:
@reaperhulk I'm not sure I follow what the proposal is here; maybe let's just do a patch and review that, unless there's complexity here I don't appreciate?

reaperhulk:
Yeah sorry I didn't look at what my change here was before asking you to check the approach. I'll revise this tomorrow with the fix I know worked for @schlenk and we can review it then.

reaperhulk:
I'm going to close this in favor of #1675

PR #1631, opened 2015-01-16
switch around the order of our travis tests and enable fast failure

reaperhulk:
An experiment to see if it makes sense to do fast failure on Travis. This reorders our tests to do docs, pep8, and py3pep8 first, then py27/py34 with OpenSSL 1.0.1 and 0.9.8 on linux. Ideally this would catch most failures early and waste less time on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,10 +60,10 @@ matrix:
 
         - os: linux
           env: TOXENV=docs
-          compiler: clang
+          compiler: gcc
         - os: linux
           env: TOXENV=pep8
-          compiler: clang
+          compiler: gcc
         - os: linux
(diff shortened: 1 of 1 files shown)


alex:
How does `fast_finish` interact with retries? We have a non-trivial failure rate due to ephemeral network failures.

reaperhulk:
Not sure. I believe if the build aborts (grey on Travis) it should be okay, but a perceived failure (red) will probably cause it to fast fail?

alex:
Sometimes we get a red fail because it's `tox` that hits the ephemeral
error. :-/

reaperhulk:
I'm going to close this for now as alex brought up a good point.

PR #1634, opened 2015-01-18
Move DSA* interfaces to interfaces.dsa module

public:
I bet this breaks the EC tests too.


reaperhulk:
*edits the title so it says interfaces.dsa*

public:
Looks like it should go green soon.

reaperhulk:
Docs failure

reaperhulk:
I think the final home for these should be `cryptography.hazmat.primitives.asymmetric.dsa` (as we're trying to do for RSA in #1592). The `cryptography.hazmat.primitives.interfaces.dsa` was only ever meant to be a temporary thing.

public:
Yep. Will move them over.

reaperhulk:
(Don't forget to move the interface docs while you're at it!)

docs/hazmat/backends/interfaces.rst
```diff
@@ -306,17 +306,17 @@ A specific ``backend`` may provide one or more of these interfaces.
             restricted to only the 1024-bit keys specified in FIPS 186-2.
 
         :return: A new instance of a
-            :class:`~cryptography.hazmat.primitives.interfaces.DSAParameters`
+            :class:`~cryptography.hazmat.primitives.dsa.DSAParameters`
```
reaperhulk: forgot asymmetric in the module path (for all of them)

reaperhulk:
There's also a pep8 issue, but once that and the paths are fixed I'll do a final review and merge.

public:
I'm doing very well at this :|

Issue #928, opened 2014-04-16
Test against the LibreSSL fork of OpenSSL?


public:
http://www.libressl.org/

http://opensslrampage.org/

Sp1l:
Just submitted https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=196827 to FreeBSD ports.
2. Removes EGD (Perl Entropy Gathering Daemon) support. This hasn't been needed on FreeBSD since FreeBSD 4.2
3. Disable compression conditionally using OPENSSL_NO_COMP
4. Check features, not version for x509_vfy
Seems to be working fine for me! Thanks to https://github.com/gentoo/libressl/tree/master/dev-python/cryptography

reaperhulk:
We do this now.

PR #1636, opened 2015-01-20
Remove EGD

Sp1l:
EGD was only necessary for some commercial UNIX systems, versions that
needed
it all reached end of life.
EGD needed until        OS release date
IRIX 6.5.19   feb 2003
Solaris 2.6                     jul 1997
AIX     5.2                     oct 2002
Tru64   5.1B                    sep 2002
HP-UX   11i v2                  sep 2003
https://en.wikipedia.org/wiki//dev/random#EGD_as_an_alternative


public:
ok to test

public:
I am OK with this.

Ayrx:
This is technically a backwards compatibility break, no?

Sp1l:
Requires pyOpenSSL to remove EGD as well is what I was told on #dev-cryptography

public:
https://github.com/pyca/pyopenssl/pull/187 in particular :)

Ayrx:
I'd make a note of it in the CHANGELOG as well.

Ayrx:
Otherwise +1 because EGD is completely stupid.

Sp1l:
Next point versions of Python won't always have EGD support. EGD support is built conditionally depending on the build system's support of EGD. This already breaks building with LibreSSL. See https://github.com/python/cpython/commit/e3ec96251b1b0d4afabcc16711e9914b13f8b303 and https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=196827 and https://github.com/pyca/cryptography/issues/928

Sp1l:
This seems to work as well... Don't know if this a clean way and if it adheres to your coding standards.

reaperhulk:
We're going to need to coordinate with pyOpenSSL to determine how we want to proceed here. I'm -1 on merging anything that does conditional binding for Libre inside our OpenSSL backend at this time (I still think we're going to need to fork the backends due to the ENGINE and other changes) but I'm willing to remove these bindings if there's a solution on the pyOpenSSL side.

public:
Do we just need PyOpenSSL to implement a shim that returns the right type of error?

Sp1l:
pyOpenSSL implements 3 egd functions. I want to try doing an import of EGD from OpenSSL and then make the def egd-function's conditional.
try:
     from OpenSSL import RAND_egd
except ImportError:
     set somevariable

if not somevariable
   def egd( etc.

reaperhulk:
Presumably you'll want to implement methods in pyOpenSSL such that the absence of EGD will not cause failures for any software that happens to "use" it (It will just noop and raise deprecation warnings).

If that path is blessed by @exarkun then I'm happy to merge a PR that outright removes the bindings (e.g. your first commit on this PR).

exarkun:
> Presumably you'll want to implement methods in pyOpenSSL such that the absence of EGD will not cause failures for any software that happens to "use" it (It will just noop and raise deprecation warnings).

It would be nice if current versions of pyopenssl didn't suddenly break against the next release of cryptography, though.  pyopenssl trustingly declares a `>=` cryptography dependency - so new installs are going to get the newest cryptography.  I'm happy for cryptography to eventually delete these bindings but please wait until there's at least one new release of pyopenssl that doesn't depend on them.

reaperhulk:
We definitely won't strip these bindings until there's a pyOpenSSL that doesn't require them. Possibly 2 or more releases (depending on your planned 2015 release cadence :grinning: )

public:
I think a noop random bytes function is probably a bad idea. I think it needs to signal to the caller that EGD isn't available by returning `-1`.

reaperhulk:
Closing in favor of #1679.

PR #1644, opened 2015-01-29
Multiple invocations of setup.py build_ext shouldn't needlessly rebuild

dstufft:
cffi doesn't actually determine if it needs to write out the source file or not. Instead it just records if it has already written out the source file in *this* invocation of the program. This means that each invocation of ``setup.py build_ext`` will write out a new ``.c`` file which will cause distutils to think it needs to rebuild the software because the mtimes are different.

### DIFF
--- a/src/cryptography/hazmat/bindings/utils.py
+++ b/src/cryptography/hazmat/bindings/utils.py
@@ -5,11 +5,47 @@
 from __future__ import absolute_import, division, print_function
 
 import binascii
+import os.path
 import sys
 import threading
 
 from cffi import FFI
-from cffi.verifier import Verifier
+from cffi.verifier import Verifier as _Verifier, _ensure_dir
+import six
+
+
+class Verifier(_Verifier):
+
+    def _write_source(self, file=None):
+        # Write our source file to an in memory file.
+        self._vengine._f = six.StringIO()
+        try:
+            self._vengine.write_source_to_f()
+        finally:
+            source_data = self._vengine._f.getvalue()
+            del self._vengine._f
+
+        # Determine if this matches the current file
+        if file is None and os.path.exists(self.sourcefilename):
+            with open(self.sourcefilename, "r") as fp:
+                needs_written = not (fp.read() == source_data)
+        else:
+            needs_written = True
+
+        # Actually write the file out if it doesn't match
+        must_close = (file is None)
+        if needs_written:
+            if must_close:
+                _ensure_dir(self.sourcefilename)
+                file = open(self.sourcefilename, "w")
+            try:
+                file.write(source_data)
+            finally:
+                if must_close:
+                    file.close()
+
+        if must_close:
+            self._has_source = True
 
 
 class LazyLibrary(object):


dstufft:
I'm trying this one instead.

koobs:
Works for me! 

@dstufft This might be a REALLY good opportunity to make distutils (setuptools, whatever) spit out a warning when files exist in a build run, but stamps have changed, or whatever condition that causes another build run to execute while the target files already exist is matched. 

This will highlight for users/developers of any projects that use C extensions if their implicit expectations aren't met, and that possibly cause heisenbug build issues downstream as well. It would certainly have helped me (and us) pinpoint quickly what the issue was.

dstufft:
Ok, so this fixes things for @koobs. There is a PR to cffi that does this located at https://bitbucket.org/cffi/cffi/pull-request/56/actually-check-if-the-file-exists-rather/diff. So the question is then do we want to just wait for cffi to fix this and bump our minimum requirement or do we want to fix this ourselves while we wait for cffi to fix it?

I think @alex had an opinion.

alex:
How about a middle ground: wait for cffi to land the patch, then ask how
long until the next release, if it's non-trivial then backport it ourselves?

On Thu, Jan 29, 2015 at 10:06 PM, Donald Stufft <notifications@github.com>
wrote:

> Ok, so this fixes things for @koobs <https://github.com/koobs>. There is
> a PR to cffi that does this located at
> https://bitbucket.org/cffi/cffi/pull-request/56/actually-check-if-the-file-exists-rather/diff.
> So the question is then do we want to just wait for cffi to fix this and
> bump our minimum requirement or do we want to fix this ourselves while we
> wait for cffi to fix it?
>
> I think @alex <https://github.com/alex> had an opinion.
>
> —
> <https://github.com/pyca/cryptography/pull/1644#issuecomment-72147030>.
>



-- 

koobs:
@dstufft If your upstream patch is well QA'd, solved the problem, and is very very unlikely not to get rejected/modified before being merged, I can try backport it into cffi in FreeBSD ports, which solves the issue in the short term for us (consumer).

dstufft:
I have no idea if it's going to get accepted or not. I hope so obviously, but I'm not a cffi developer and they didn't say yet.

dstufft:
The cffi patch merged and arigato says he expects it to be released soon. Unless we have a reason to not want to up our minimum cffi version I think we can close this?

public:
Updating our CFFI dep sounds fine to me.

alex:
I don't think we can update our cffi dep, it causes weird problems for
PyPy: nonetheless, I don't think it's a problem, we have `cffi>=0.8` in
there, so any new cffi users will get a new version automatically.

On Sat, Jan 31, 2015 at 11:46 AM, Donald Stufft <notifications@github.com>
wrote:

> The cffi patch merged and arigato says he expects it to be released soon.
> Unless we have a reason to not want to up our minimum cffi version I think
> we can close this?
>
> —
> <https://github.com/pyca/cryptography/pull/1644#issuecomment-72325327>.
>



-- 

reaperhulk:
I'm going to go ahead and close this, although the problem is not resolved until 0.8.7 is released.

PR #1654, opened 2015-02-12
make our docs job nitpicky and fix every broken link

reaperhulk:
By enabling `nitpicky` mode we can have our docs fail on any link that doesn't work. Turning this on revealed quite a few link mistakes. These are all now fixed.

One caveat of this change is that the changelog will need to be altered on occasion if we deprecate/remove something we linked.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -100,11 +100,8 @@ Changelog
   ``load_elliptic_curve_private_numbers`` and
   ``load_elliptic_curve_public_numbers`` on
   :class:`~cryptography.hazmat.backends.interfaces.EllipticCurveBackend`.
-* Added
-  :class:`~cryptography.hazmat.primitives.interfaces.EllipticCurvePrivateKeyWithNumbers`
-  and
-  :class:`~cryptography.hazmat.primitives.interfaces.EllipticCurvePublicKeyWithNumbers`
-  support.
+* Added ``EllipticCurvePrivateKeyWithNumbers`` and
+  ``EllipticCurvePublicKeyWithNumbers`` support.
 * Work around three GCM related bugs in CommonCrypto and OpenSSL.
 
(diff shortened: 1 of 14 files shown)


CHANGELOG.rst
```diff
@@ -164,36 +159,24 @@ Changelog
...
-* Deprecated
-  :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPublicKey` in favor
-  of backend specific providers of the
-  :class:`~cryptography.hazmat.primitives.interfaces.DSAPublicKey` interface.
-* Deprecated
-  :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAParameters` in
-  favor of backend specific providers of the
-  :class:`~cryptography.hazmat.primitives.interfaces.DSAParameters` interface.
+* Deprecated the concrete ``RSAPrivateKey`` class in favor of backend
+  specific providers of the ``RSAPrivateKey`` interface.
```
alex: Oof, these get hard to read, should we write out the full path name, even though it's kind of verbose?

alex:
Besides that comment, this looks great to me. thanks for taking on this work!

CHANGELOG.rst
reaperhulk: When we deprecated the concrete `cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKey` class we said you should use the `cryptography.hazmat.primitives.interfaces.RSAPrivateKey` interface. We have subsequently moved that interface so it's located at... `cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKey` (aka exactly where the concrete class used to live). Maybe the solution to this is to make the second definition point to `cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKey` as the proper interface?

alex:
Merge conflict :-( Please rebase.

reaperhulk:
awwwwww crap :)

PR #1655, opened 2015-02-12
remove unused X509 certificate interface

reaperhulk:
We merged this interface and then changed our approach and merged a cryptography.x509.Certificate interface instead. This one was never directly used.


alex:
nice catch!

PR #1656, opened 2015-02-12
add repr for x509.NameAttribute

reaperhulk:


alex:
this lgtm, will merge when tests pass

PR #1657, opened 2015-02-13
Use readme to ensure the long description renders on PyPI

dstufft:


alex:
Does `readme` really install a `setup.py` command that's as vaguely named as "check"?

dstufft:
Sort of. Check is a distutils command which already attempted to do this but it used different code then pypi does. So readme subclasses the check command and overrides the rst parsing.

alex:
Is this going to break if something else is installed that also installs a
check command?

On Thu, Feb 12, 2015 at 8:48 PM, Donald Stufft <notifications@github.com>
wrote:

> Sort of. Check is a distutils command which already attempted to do this
> but it used different code then pypi does. So readme subclasses the check
> command and overrides the rst parsing.
>
>
> > On Feb 12, 2015, at 9:42 PM, Alex Gaynor <notifications@github.com>
> wrote:
> >
> > Does readme really install a setup.py command that's as vaguely named as
> "check"?
>
> >
> > —
> > Reply to this email directly or view it on GitHub.
> >
>
> —
> <https://github.com/pyca/cryptography/pull/1657#issuecomment-74196476>.
>



-- 

dstufft:
Most likely yes. The same would be true of any command name because of the way that setuptools entrypoints work. I can add a second, readme specific, command though if it'd make you feel better.

alex:
I guess it's fine.

On Thu, Feb 12, 2015 at 9:08 PM, Donald Stufft <notifications@github.com>
wrote:

> Most likely yes. The same would be true of any command name because of the
> way that setuptools entrypoints work. I can add a second, readme specific,
> command though if it'd make you feel better.
>
> —
> <https://github.com/pyca/cryptography/pull/1657#issuecomment-74198051>.
>



-- 

dstufft:
For the record this'll also check other things besides the ``long_description``. It'll check that it's not missing any metadata that distutils considers mandatory (where mandatory in distutils mind means if you don't include it we'll substitute the value ``UNKNOWN``). It's unlikely that we'll ever regress to the point we don't have a version number or url or something for cryptography though.

PR #1661, opened 2015-02-14
tox -e docs is broken

Ayrx:
I updated my `master`, ran tox and found that `tox -e docs` is broken. Here is the error log: https://gist.github.com/Ayrx/d89f0b519d3e5f2d6a82

Swapping the install order of the dependencies fixes the issue for me. Not sure why things were broken though.

/cc @dstufft

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -21,10 +21,10 @@ commands =
 deps =
     doc8
     pyenchant
-    readme>=0.5.1
     sphinx
     sphinx_rtd_theme
     sphinxcontrib-spelling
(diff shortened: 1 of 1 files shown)


alex:
hmm, so I'm not able to reproduce this problem. If you delete your `.tox/` does it still happen on the next invocation?

dstufft:
It's most likely an issue that happens during an upgrade of setuptools.

Ayrx:
Looks like @dstufft got it right. This is probably a non-issue. :)

Issue #1286, opened 2014-07-27
Expand WithNumbers Classes


reaperhulk:
From the mailing list:

```
As I worked with these asymmetric key classes, I also noticed a bit of 
an inconsistency in the API used to access the various raw numbers that 
go into making up the keys. The current API uses a mixture of method 
calls and properties, and provides different ways to access some of the 
values depending on whether you are working with public or private keys. 
I think it would be good to make this a bit more consistent, at least about 
method calls vs. properties, but also ideally to provide a common way to 
access all the values common to public & private keys using the same 
API, with additional API calls for private keys to access values only 
applicable there.

As an example of what I’m talking about, here are the calls needed to 
access the parameters of a DSA public key today:

p: key.parameters().parameter_numbers().p
q: key.parameters().parameter_numbers().q
g: key.parameters().parameter_numbers().g
y: key.public_numbers().y

Note how in one case you can get to the public numbers directly, whereas 
in the other you have to go through an intermediate object first. It would be 
nice there to have parameter_numbers() available directly from the key 
similar to public_numbers(), at least as a convenience function. I would 
keep the existing parameters() function as well, as it has value to extract 
the parameters as a group to use in creating other keys. However, 
allowing access either that way or directly to the numbers object would 
be nice.
```

To resolve the first issue we should consider adding the other `*_numbers` methods to the `WithNumbers` interface and implement them in the OpenSSL backend. Any private key has access to the public numbers so this would be an expansion of the API.

reaperhulk:
Closing for now (see #1662 for more about why this was declined)

Issue #1339, opened 2014-09-20
Make (RSA/DSA)PrivateKey implement the PublicKey interface


singpolyma:
Instead of just providing a method to get a public key.  Would allow for much more polymorphism.

reaperhulk:
At the moment there is a broad feeling among core team members that it's better to be explicit on this front and require the developer to acknowledge they're operating on the public components of a private key, so we're going to decline this.

Issue #1288, opened 2014-07-27
Attribute/method inconsistency on numbers objects


reaperhulk:
From the mailing list:

```
y: key.private_numbers().public_numbers.y
```

reaperhulk:
Per discussion in IRC this perceived inconsistency is because by convention cryptography makes things that require computation methods and things that do not properties.

PR #1662, opened 2015-02-14
expand withnumbers refs #1286

reaperhulk:
- [x] Do this for elliptic curve
- [x] Add changelog entry

This is a PR to jump start the discussion of #1286. Fixing method/property consistency may be tricky (is there a way to have a method/property with the same name on a class?).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -71,6 +71,13 @@ Changelog
   :mod:`~cryptography.hazmat.primitives.asymmetric.rsa`.
 * Added support for parsing X.509 names. See the
   :doc:`X.509 documentation</x509>` for more information.
+* Added
+  :meth:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKeyWithNumbers.public_numbers`,
+  :meth:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPrivateKeyWithNumbers.public_numbers`,
+  :meth:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPublicKeyWithNumbers.parameter_numbers`,
+  and
+  :meth:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePrivateKeyWithNumbers.public_numbers`
+  as convenience methods.
 * Fixed building against LibreSSL, a compile-time substitute for OpenSSL.
 
 0.7.2 - 2015-01-16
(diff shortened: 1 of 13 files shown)


reaperhulk:
Interested in opinions on methods vs properties (and how we might deprecate to be consistent). cc @public @alex @ronf

alex:
The reason they are "incosistent" right now is a conceptual difference: Key
-> Numbers potentially involves computation, so it's a method,
PrivateNumbers -> PublicNumbers is just returning the subset of data, so
it's a property.

On Mon, Feb 16, 2015 at 12:22 PM, Paul Kehrer <notifications@github.com>
wrote:

> Interested in opinions on methods vs properties (and how we might
> deprecate to be consistent). cc @public <https://github.com/public> @alex
> <https://github.com/alex> @ronf <https://github.com/ronf>
>
> —
> <https://github.com/pyca/cryptography/pull/1662#issuecomment-74543065>.
>



-- 

reaperhulk:
Okay, so we can leave the method/property distinction in place. Is there an objection to adding these convenience methods?

alex:
Not sure how I feel about them. I guess it's a fairly explicit transition
from private -> public.

On Tue, Feb 17, 2015 at 12:10 PM, Paul Kehrer <notifications@github.com>
wrote:

> Okay, so we can leave the method/property distinction in place. Is there
> an objection to adding these convenience methods?
>
> —
> <https://github.com/pyca/cryptography/pull/1662#issuecomment-74706215>.
>



-- 

public:
I quite like that `key.public_key().public_numbers()` makes it quite obvious that `key` is a private key currently.

If private keys were interoperable as public keys it would be easier to send private data to a function that was only ever really intended with public data in mind. I'm not sure how much of a footgun this is in practice though.

alex:
That strict distinction was definitely my motivation when I originally
worked on this stuff.

On Mon, Feb 23, 2015 at 9:23 AM, Alex Stapleton <notifications@github.com>
wrote:

> I quite like that key.public_key().public_numbers() makes it quite
> obvious that key is a private key currently.
>
> If private keys were interoperable as public keys it would be easier to
> send private data to a function that was only ever really intended with
> public data in mind. I'm not sure how much of a footgun this is in practice
> though.
>
> —
> <https://github.com/pyca/cryptography/pull/1662#issuecomment-75588421>.
>



-- 

reaperhulk:
I am neutral about this so if we want to say "no we don't want these convenience methods because they hide some of the subtlety" that's fine. These ideas touch #1286, #1288, and probably #1339 so we can probably close all 3 if we are happy with the current layout.

alex:
This either needs a rebase or be closed.

reaperhulk:
Okay, consensus for now is that this is working as designed. We want the verbosity to make it clear to the user what exactly is going on (obtaining public numbers from a private key for example).

PR #1673, opened 2015-02-18
Fix (most) problems with LibreSSL

AnchorCat:
These four commits fix most, but not all of the problems with LibreSSL. On my OpenBSD -current (as of 2015-02-13) system, the cryptography test suite passes with the following patches applied:

 * All in this pull request.
 * Removal of cryptodev (I'm not as confident in that patch as I am in this pull request, so I will be submitting it separately): https://github.com/anchor/cryptography/commit/27c62f5a5a0def2a82ca98eafcfb51a2547c7ff5
 * @Sp1l's patch to remove EGD: https://github.com/Sp1l/cryptography/commit/7c1874d5b94e931b7520151c6f59557036e2e889
  * Note that it does **not** work with all patches from the [pull request](https://github.com/pyca/cryptography/pull/1636) relating to EGD's removal, because that commit has been reverted and replaced with something that still assumes EGD's existence.

I have tried to provide sufficient explanation of what each commit is doing in the individual commit messages, but please let me know if clarification is needed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -68,6 +68,7 @@ Changelog
   :mod:`~cryptography.hazmat.primitives.asymmetric.rsa`.
 * Added support for parsing X.509 names. See the
   :doc:`X.509 documentation</x509>` for more information.
+* Fixed building against LibreSSL, a compile-time substitute for OpenSSL.
 
 0.7.2 - 2015-01-16
 ~~~~~~~~~~~~~~~~~~
--- a/src/cryptography/hazmat/bindings/openssl/ssl.py
+++ b/src/cryptography/hazmat/bindings/openssl/ssl.py
@@ -7,6 +7,14 @@ from __future__ import absolute_import, division, print_function
 INCLUDES = """
 #include <openssl/ssl.h>
 
+/* LibreSSL has removed support for compression, and with it the
+ * COMP_METHOD use in ssl.h. This is a hack to make the function types
+ * in this code match those in ssl.h.
+ */
+#ifdef LIBRESSL_VERSION_NUMBER
+#define COMP_METHOD void
+#endif
+
 typedef STACK_OF(SSL_CIPHER) Cryptography_STACK_OF_SSL_CIPHER;
 """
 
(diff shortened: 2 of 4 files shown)


alex:
ok to test

AnchorCat:
I've submitted the cryptodev patch (referenced in my initial comment) as https://github.com/pyca/cryptography/pull/1674.

Ayrx:
I'm gonna have to say -1 on landing this as I agree with @reaperhulk that LibreSSL should be supported as a different backend since it seems like the LibreSSL team has no intention of preserving compatibility with OpenSSL. Supporting LibreSSL in this manner will probably get increasingly messy.

reaperhulk:
@AnchorCat did Libre add dynamic engine support back in? I thought previous releases had completely removed the engine subsystem?

AnchorCat:
@Ayrx: Please forgive my noob question, as I'm not a Python programmer, but can that be done in such a way as to reuse the parts which are common (which is currently most of the openssl backend)? It seems equally messy to maintain a copy of all the backend code.

@reaperhulk: Referencing the OpenBSD CVS version of [engine.h](http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/lib/libssl/src/crypto/engine/engine.h?annotate=1.30), the declarations of `ENGINE_load_openssl` and `ENGINE_load_dynamic` (lines 319-321) haven't been changed since revision 1.2, in 2002. Unless I'm missing something (which is entirely possible, since I only started looking at this problem yesterday), it looks like the engine subsystem was never removed.

reaperhulk:
Hmm, interesting. We've had some other people come through asking for Libre fixes in the past and I gathered the impression that engine support was entirely missing. Clearly that's not the case.

Re: alternate backends, the long term concern is that LibreSSL, while **close** to OpenSSL, appears to be slowly diverging and these conditional checks may ultimately prove to be a burden above and beyond maintaining a (very similar) second backend. Unfortunately right now the alternate backend system isn't really mature enough for use so that's not a serious option.

Given the above I'm :+1: on this for now. I think I'm willing to add a LibreSSL builder to prevent regression, but I don't think I want to officially state we support LibreSSL until either a) we've got the tooling in place to support it as a separate backend or b) we're confident they're not going to remove/break any additional APIs.

Definitely interested in opinions from other team members here.

alex:
One downside to a seperate binding/backend is that things that are "OpenSSL" specific, such as pyOpenSSL won't work with it.

I'm broadly happy with the approach here, but strongly agree that we need CI.

Ayrx:
Ok, I guess this isn't *that* messy at the moment and if things start to get hairy we can always refactor. :)

We still need to discuss removing cryptodev (#1674) and EGD though, those are technically backward compatibility breaks and pyopenssl may be relying on them existing.

reaperhulk:
Closing in favor of #1679.

PR #1674, opened 2015-02-18
Remove cryptodev engine support

AnchorCat:
This was removed from LibreSSL, but is still present in OpenSSL. It
can't be wrapped in an #ifndef, because cffi doesn't know how to handle
that case.

I'm actually not sure if this is the best possible solution, but I'm not
familiar enough with cffi to make that call myself. The cryptography
test suite still passes with this patch applied, which is why I've
submitted it anyway; if this is the wrong fix, please advise as to a
better approach and I'll revise it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -68,7 +68,7 @@ Changelog
   :mod:`~cryptography.hazmat.primitives.asymmetric.rsa`.
 * Added support for parsing X.509 names. See the
   :doc:`X.509 documentation</x509>` for more information.
-* Removed support for OpenSSL's cryptodev engine.
+* Removed support for OpenSSL's ``cryptodev`` engine.
 
 0.7.2 - 2015-01-16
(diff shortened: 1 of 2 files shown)


alex:
ok to test

CHANGELOG.rst
```diff
@@ -68,6 +68,7 @@ Changelog
   :mod:`~cryptography.hazmat.primitives.asymmetric.rsa`.
 * Added support for parsing X.509 names. See the
   :doc:`X.509 documentation</x509>` for more information.
+* Removed support for OpenSSL's cryptodev engine.
```
reaperhulk: To prevent the docs failure just wrap cryptodev in double backticks.

AnchorCat:
@reaperhulk: Done, thanks for the tip.

alex:
I think we want to make this a conditional definition.

reaperhulk:
Closing in favor of #1679.

PR #1675, opened 2015-02-18
support OpenSSL built with no-comp

reaperhulk:
refs #1622 

This PR should fix building OpenSSL with compression disabled (no-comp) on all platforms. On Windows we (apparently) have to declare the functions before we remove them via `CONDITIONAL_NAMES`. For more information about why see: https://github.com/pyca/cryptography/pull/1622#issuecomment-72643336

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/ssl.py
+++ b/src/cryptography/hazmat/bindings/openssl/ssl.py
@@ -546,13 +546,40 @@ static const long Cryptography_HAS_ALPN = 0;
 #else
 static const long Cryptography_HAS_ALPN = 1;
 #endif
+/* Gather 'round children and I shall tell you a story. A story of how OpenSSL
+   has a feature flag called OPENSSL_NO_COMP. When this flag is defined
+   SSL/TLS compression is disabled. However, rather than not define the symbols
+   for a few functions that no longer have any utility OpenSSL decided to
+   change their signature. This in and of itself was not a serious problem,
+   but the Windows build decided to not actually export those symbols. Thanks
+   to those decisions and cffi's inflexibility here's what we're about to do.
+   If OPENSSL_NO_COMP is defined and we're on Windows we will provide a NULL
+   implementation of the functions. Otherwise we'll just detect that
+   compression is disabled and set Cryptography_HAS_COMPRESSION so we can
+   remove the functions from the cffi lib object via CONDITIONAL_NAMES.
+
+   LibreSSL also doesn't have compression.
+*/
 #if defined(OPENSSL_NO_COMP) || defined(LIBRESSL_VERSION_NUMBER)
 static const long Cryptography_HAS_COMPRESSION = 0;
 typedef void COMP_METHOD;
+/* On Windows we need to define these functions so we can remove them
+   conditionally. This is necessary because Windows OpenSSL messes up the
+   exports of these functions. See #1622 for more details. */
+#if defined(_WIN32)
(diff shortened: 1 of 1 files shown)


reaperhulk:
@schlenk when you get a chance could you confirm this works for you?

alex:
I don't understand why Windows is special here, and the comment you linked doesn't explain it IMO.

schlenk:
Doesn't compile when OPENSSL_NO_COMP is defined (you missed the parameter names in the lines 569, 572, 575) of ssl.py.

schlenk:
Once those are fixed, it compiles and can be used, tests run. (64-Bit Version, will test 32-Bit at work).

alex:
This needs a rebease.

reaperhulk:
@schlenk Do you happen to have an installer for an OpenSSL with no-comp on Windows? Or, failing that, a simple way to compile OpenSSL on Windows? I've never actually done it :)

schlenk:
@reaperhulk Compiling OpenSSL on Windows is actually easy enough 

Open a cmd shell and run vcvarsall.bat to get the compiler path etc.
```
perl Configure --prefix=d:\code\openssl no-comp no-ssl2 no-ssl3 VC-WIN32
ms\do_ms.bat    (or ms\do_nasm.bat if you have the NASM assembler installed and need the speed)
nmake -f ms\ntdll.mak   (or nmake -f ms\nt.mak if you only want the static libs, the issue might only appear with the dynamic lib, not sure.)
```
there is install target, but the libs are there (but not with the 'mt' suffixes expected by the cryptography build, so just rename the .lib files), see the out32 and inc32 folders.

schlenk:
Ok. Builds now (on Windows x64_64, Python 2.7.9rc1_x64 from python.org, openssl-1.0.2 with no-comp) and tests run without failures.

schlenk:
Ping. Works for me now. Anything I could do to get this one merged?

alex:
So the UNIX bits here look ok to me, but I still don't understand why Windows is special?

alex:
Could we start with a version that just added the `defined(OPENSSL_NO_COMP)` to the thing

schlenk:
Some more information:
The windows build works fine without the extra `#ifdef (_WIN32)` stuff in the patch, if openssl is linked statically. That makes sense, as the broken .def file from openssl which hides the symbols is just used for building the DLL version.
So for static linking, the windows code is 100% identical to the Linux version with OPENSSL_NO_COMP. Only for dynamic linking there is a difference due to the weird way openssl builds but hides those symbols which lets the MS linker complain. In fact, the 'static link' case breaks with the `#ifdef (_WIN32)` part included, while 'dynamic' works.

Not nice. Guess one could remove the #ifdef (_WIN32) block from the patch and at least fix the build for the (probably more common) static linking case. No idea how to get both the static and dynamic version to work easily.

alex:
needs a rebase now.

reaperhulk:
This has been rebased, but if the current approach doesn't work for both dynamic/static I'm very much against it being merged.

alex:
Agreed.

schlenk:
Thx for merging #1824. That at least fixes the problem for the most common cases (including static libssl on Windows). 
Anyone using the .dll version has probably built it anyway, so could they could patch the crappy .def file too, which would make this a non-issue. (and maybe the openssl guys fix the interface for the future).

reaperhulk:
I think I'm going to close this for now. We can reopen if necessary, but hopefully it'll just get fixed upstream.

Issue #1508, opened 2014-12-04
Consider moving interface definitions out of 'interfaces' module


alex:
Why aren't RSAPrivateKey/RSAPublicKey just in the `rsa` module?

At a minimum, i think we should consider moving the docs.

reaperhulk:
Anybody have an opinion on where `AsymmetricPadding`, `AsymmetricVerificationContext`, and `AsymmetricSignatureContext` should live? Those interfaces, along with `KeyDerivationFunction` and `MACContext` are all that are left once the cipher PR merges.

alex:
Not sure what to do with `KDF` and `MACContext`, for the Asymmetric ones, as a first pass we could put them in `hazmat.primtives.asymmetric`?

reaperhulk:
We can move kdf to `cryptography.hazmat.primitives.kdf.KeyDerivationFunction`. Still not sure what to do about `MACContext`.

reaperhulk:
One possible solution would be to move hmac/cmac to `cryptography.hazmat.primitives.mac` and deprecate their current module path. Then we'd have a place for it, but at the cost of deprecating two more things.

public:
I think we should move the mac stuff too.

reaperhulk:
The only item left is `MACContext` and we're not going to move that for now (per discussion in IRC).

PR #1677, opened 2015-02-18
move hmac/cmac to mac module, move MACContext to mac module

reaperhulk:
This is a potential solution to moving `MACContext`. It deprecates and moves `hmac` and `cmac` into a new `mac` module. Advantage: Completes our interface diaspora. Disadvantage: deprecates and moves `hmac` and `cmac`.

refs #1508

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,12 @@ Changelog
   now load elliptic curve public keys.
 * Added
   :func:`~cryptography.hazmat.primitives.asymmetric.rsa.rsa_recover_prime_factors`
+* :mod:`cryptography.hazmat.primitives.mac.hmac` and
+  :mod:`cryptography.hazmat.primitives.mac.cmac` were moved to the
+  :mod:`~cryptography.hazmat.primitives.mac` module.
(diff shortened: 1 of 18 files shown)


alex:
I'm not wild about this.

While moving interfaces probably doesn't affect most people, moving these will effect many users.

reaperhulk:
I would prefer not to move these modules either. We could add a `mac` module that only exports the `MACContext` and not move `hmac` and `cmac` (or move them but not deprecate the original module path? Not sure I like that idea).

alex:
I think maybe the right way to go is to just leave it for now, it's
imperfect, that's ok.

On Thu, Feb 19, 2015 at 3:54 PM, Paul Kehrer <notifications@github.com>
wrote:

> I would prefer not to move these modules either. We could add a mac
> module that only exports the MACContext and not move hmac and cmac (or
> move them but not deprecate the original module path? Not sure I like that
> idea).
>
> —
> <https://github.com/pyca/cryptography/pull/1677#issuecomment-75133905>.
>



-- 

reaperhulk:
After discussion in IRC it was decided to leave the MACContext where it is for now.

PR #1678, opened 2015-02-19
use pip install --user on travis

reaperhulk:
This attempts to use the Travis cache + wheels to speed up our builds a bit. Also dropped the use of sudo in favor of `pip install --user`.

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -103,7 +103,7 @@ else
     esac
 fi
 
-sudo pip install virtualenv
+pip install --user virtualenv
 virtualenv ~/.venv
 source ~/.venv/bin/activate
 pip install tox coveralls


alex:
Travis's cache is only available with docker or paid plans, so this doesn't do anything.

Ayrx:
`pip install --user` sounds like a good change though.

reaperhulk:
Oops, didn't realize it was docker only. Nothing can ever speed up our builds :cry:

reaperhulk:
Okay, altered track to just use `pip install --user`

alex:
`pip install --user` doesn't install stuff in a place that's actually on `$PATH`. I think we should probaly just close this PR out.

reaperhulk:
Yeah this isn't worth it for now.

PR #1680, opened 2015-02-19
Simplify x509.Name.get_attributes_for_oid

alex:


reaperhulk:
(coveralls is useless garbage that claims it'll use the status API but does not)

PR #1683, opened 2015-02-20
Tell py.test to only run the tests from the local tests dir.

alex:
Right now if you have a virtualenv or something in the same directory it will recurse into it to run tests.


reaperhulk:
:+1:

PR #1684, opened 2015-02-20
we don't test on freebsd 9.x any more

reaperhulk:


alex:
I think this deserves a CHANGELOG entry

PR #1692, opened 2015-02-23
remove IBM link and replace with smartbear PDF link

reaperhulk:


alex:
Note for future people: the IBM link does not respond to HEAD requests and so it causes problems for the linkcheck builder.

PR #1695, opened 2015-02-24
Remove the CFFI requirement, which breaks pip install -U on PyPy

Julian:
PyPy ships with CFFI, but pip install -U will upgrade it if a newer CFFI is available, which will break the CFFI installation (and PyPy itself).

I haven't tested this, this PR is half a reminder and half confirmation from @dstufft that this is the solution he was proposing.

CC @alex

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,6 @@ SIX_DEPENDENCY = "six>=1.4.1"
 VECTORS_DEPENDENCY = "cryptography_vectors=={0}".format(about['__version__'])
 
 requirements = [
-    CFFI_DEPENDENCY,
     "pyasn1",
     SIX_DEPENDENCY,
     SETUPTOOLS_DEPENDENCY
@@ -43,6 +42,9 @@ requirements = [
 
 if sys.version_info < (3, 4):
     requirements.append("enum34")
+
+if platform.python_implementation() != "PyPy":
+    requirements.append(CFFI_DEPENDENCY)
 
 # If you add a new dep here you probably need to add it in the tox.ini as well
 test_requirements = [


dstufft:
The setuptools dependency is there because cryptography depends on setuptools at runtime in order to discover backends.

dstufft:
Also, you're going to need to duplicate the entire dependency list inside of ``setup.cfg``, it's all or nothing.

Julian:
Ah, OK, re-added (setuptools).

That seems... unfortunate about the DRY -- I guess I'll wait for some confirmation or consensus then that you guys want to go that route?

alex:
add to whitelist

Julian:
Further thoughts?

reaperhulk:
I'd like to see this in 0.8 but I'm not qualified to judge this. The change to setup.py looks :+1: to me but I don't fully understand the need for the information in setup.cfg.

public:
I guess we defer to @dstufft on this? :)

dstufft:
It should work, duplicating information is because ``bdist_wheel`` will use ``install_requires`` to determine what the requirements of the Wheel will be, unless there's a setup.cfg which will override that. It's the only way to conditionally depend on cffi like we do really.

reaperhulk:
Is that also true for enum34 then?

dstufft:
Yes, it "works" because our wheels are Python version specific so we have to generate a wheel for each version anyways.

reaperhulk:
This problem is going to affect 100% of projects using cffi, correct? Is it really practical to have every cffi project know to do this? Seems like a crappy experience for pypy users...

alex:
@dstufft given that it sounds like we could omit the `setup.cfg`, as we have for `enum34`?

dstufft:
Um, yes. But it'll also mean that the dependency information that shows up on PyPI depends on which wheel is uploaded, uh either first or last. If I recall we accepted that for enum34, not sure if want to accept it for CFFI.

alex:
@dstufft do people actually use the dependency info on pypi? Most packages don't have it from what I've seen.

dstufft:
Currently it's mainly for informational display.

alex:
@dstufft so would you be ok with the `setup.py` only version of this change?

dstufft:
It'll be functional so yea.

alex:
Cool. @Julian you want to update the patch?

Julian:
Updated.

reaperhulk:
:+1: if it passes tests we'll merge for the 0.8 release.

PR #1708, opened 2015-03-01
move private_bytes to a shared method so we can reuse it

reaperhulk:
This will make things much simpler for EC and DSA.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -36,7 +36,7 @@ from cryptography.hazmat.backends.openssl.rsa import (
 )
 from cryptography.hazmat.backends.openssl.x509 import _Certificate
 from cryptography.hazmat.bindings.openssl.binding import Binding
-from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import dsa, ec, rsa
 from cryptography.hazmat.primitives.asymmetric.padding import (
     MGF1, OAEP, PKCS1v15, PSS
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1093,6 +1093,65 @@ def _ec_key_set_public_key_affine_coordinates(self, ctx, x, y):
 
         return ctx
 
+    def _private_bytes(self, encoding, format, encryption_algorithm,
```
alex: I think this method name should be "private_key_bytes", so we have context.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1093,6 +1093,65 @@ def _ec_key_set_public_key_affine_coordinates(self, ctx, x, y):
...
+
+        # This is a temporary check until we land DER serialization.
+        if encoding is not serialization.Encoding.PEM:
+            raise ValueError("Only PEM encoding is supported by this backend")
+
+        if format is serialization.Format.PKCS8:
+            write_bio = self._lib.PEM_write_bio_PKCS8PrivateKey
+            key = evp_pkey
+        elif format is serialization.Format.TraditionalOpenSSL:
+            write_bio = pem_write_func
```
alex: `pem_write_func` isn't a great name for what this is, it's not like PKCS8 isn't pem.

reaperhulk: Could be "traditional_write_func".

alex: Sounds good.

alex:
lgtm

PR #1709, opened 2015-03-01
Serialize EC private keys

reaperhulk:


src/cryptography/hazmat/backends/openssl/ec.py
```diff
@@ -200,6 +200,23 @@ def private_numbers(self):
             public_numbers=self.public_key().public_numbers()
         )
 
+    def private_bytes(self, encoding, format, encryption_algorithm):
+        evp_pkey = self._backend._lib.EVP_PKEY_new()
+        assert evp_pkey != self._backend._ffi.NULL
+        evp_pkey = self._backend._ffi.gc(
+            evp_pkey, self._backend._lib.EVP_PKEY_free
+        )
+        res = self._backend._lib.EVP_PKEY_set1_EC_KEY(evp_pkey, self._ec_key)
```
alex: Ugh, why do we have an `EVP_PKEY` on RSA but not EC?

reaperhulk: This is a question dreid asked awhile back in an issue I can't find at the moment. The short answer is "we didn't need it until now". We can hoist it up into the init now though if we want.

alex: I'd kind of like to see what that patch looks like if you don't mind

alex:
This LGTM, as with most substantitve things, I'd love another pair of eyes (cc: @public).

src/cryptography/hazmat/backends/openssl/ec.py
public: I found that the EVP APIs get in the way and are often less obvious than the low level APIs for the kind of stuff we do but maybe I was insufficiently good at OpenSSL back then.

alex: I don't think we have a choice about using EVP for PKCS8.

On Mon, Mar 2, 2015 at 7:39 AM, Alex Stapleton <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/1709#discussion_r25607137>:
>
> > @@ -200,6 +200,23 @@ def private_numbers(self):
> >              public_numbers=self.public_key().public_numbers()
> >          )
> >
> > +    def private_bytes(self, encoding, format, encryption_algorithm):
> > +        evp_pkey = self._backend._lib.EVP_PKEY_new()
> > +        assert evp_pkey != self._backend._ffi.NULL
> > +        evp_pkey = self._backend._ffi.gc(
> > +            evp_pkey, self._backend._lib.EVP_PKEY_free
> > +        )
> > +        res = self._backend._lib.EVP_PKEY_set1_EC_KEY(evp_pkey, self._ec_key)
>
> I found that the EVP APIs get in the way and are often less obvious than
> the low level APIs for the kind of stuff we do but maybe I was
> insufficiently good at OpenSSL back then.
>
> —
> <https://github.com/pyca/cryptography/pull/1709/files#r25607137>.
>



-- 

public: Using the OpenSSL API, maybe not, but I think we generally do need access to the low level key objects for the other operations we implement.

public:
LGTM

alex:
Needs a CHANGELOG and then I'll merge

PR #1706, opened 2015-02-28
Serialize RSA public keys

reaperhulk:
Outstanding work:
- [x] Discuss `PublicFormat` naming.
- [x] Implement "bare" public key support (refs #1705)
- [x] rebase after #1711 lands

PR #1711, opened 2015-03-02
s/Format/PrivateFormat

reaperhulk:
Public formats are sufficiently different from private formats that we should have them be different enums. This renames Format to PrivateFormat to make way for a PublicFormat in the near future (see: #1706).


alex:
lgtm

PR #1713, opened 2015-03-02
Update OpenSSL test vector link to point at Github repository

reaperhulk:
Fixes linkcheck (plus Github is more reliable than OpenSSL's gitweb anyway).


alex:
Noting that the sha1 of the commit changed, but it isn't particularly material because that file has only one revision ever.

Issue #1705, opened 2015-02-28
Support bare public keys?


reaperhulk:
Right now the OpenSSL backend can import PKCS8 public keys (and probably SubjectPublicKeyInfo wrapped keys, although we should verify that) but fails on "bare" RSA public keys (e.g. items written with `PEM_write_BIO_RSAPublicKey`). This is apparently a thing some people expect to be able to read (see: http://stackoverflow.com/questions/10569189/how-to-read-a-rsa-public-key-in-pem-pkcs1-format). Should we support it? And, relatedly, should we support writing this format with `as_bytes`?

I haven't researched EC or DSA so they may be subject to similar weirdness.

alex:
This seems reasonable to support.

ronf:
Both DSA and RSA support PKCS#1 format for both private and public keys, and such keys still seem pretty common. EC supports PKCS#1 private keys, but I haven't been able to find a specification for PKCS#1 public keys there (only PKCS#8, which supports both private & public, and other formats like OpenSSH and RFC 4716 which only seem to support public keys).

public:
Previously. https://github.com/pyca/cryptography/pull/1233#issuecomment-55365638

reaperhulk:
The naming problems in this space are manifold... It appears PKCS8 is not a public key format. PEM public keys with `-----BEGIN PUBLIC KEY-----` armoring are more properly called X.509 subjectPublicKeyInfo structures that contain PKCS1 payloads. So the two primary public key formats are "subjectPublicKeyInfo" and "PKCS1" even though the latter is contained within the former.

RSA PKCS1 public keys are relatively easy to create with the command line. DSA keys also exist (see @public's comment), and I'll need to conduct a test with `PEM_write_bio_EC_PUBKEY` to see if it generates PKCS1 or subjectPublicKeyInfo (which would give it the same output as PEM_write_bio_PUBKEY).

reaperhulk:
`PEM_write_bio_EC_PUBKEY` does a `subjectPublicKeyInfo` so it's possible that EC "bare" public keys don't exist/aren't used (at least by OpenSSL). Evidence otherwise appreciated.

reaperhulk:
`PEM_read_bio_DSAPublicKey` is not defined, so in practice loading a PKCS1 only DSA public key requires doing something like:

```c
# define PEM_read_bio_DSAPublicKey(bp,x,cb,u) (DSA *)PEM_ASN1_read_bio( \
[...]

PR #1714, opened 2015-03-03
add RSA/DSA PKCS1 public keys

reaperhulk:
`rsa.pub.pem` is PEM PKCS1 formatted public key converted from the subjectPublicKeyInfo version of the public key (`asymmetric/PKCS8/unenc-rsa-pkcs8.pub.pem`).
`dsa.pub.pem` is the DSA key from the ruby OpenSSL tests.

Refs #1705 

These will be used to test expanded public key loading support (well, the RSA one will at least).


reaperhulk:
To validate these vectors do the following (all file paths below assume you're in the root of a cryptography checkout):

* clone cryptography
* `git fetch origin +refs/pull/1714/merge:`
* `git checkout -qf FETCH_HEAD`
* For DSA check that the link in the docs to the ruby tests shows the same key as the one added.
* For RSA PEM parse it using openssl (OS X system OpenSSL won't work, use homebrew) `/usr/local/opt/openssl/bin/openssl rsa -RSAPublicKey_in -in vectors/cryptography_vectors/asymmetric/public/PKCS1/rsa.pub.pem -text`. This will output some data that will show a 1024-bit modulus and also a PEM public key that is identical to `vectors/cryptography_vectors/asymmetric/PKCS8/unenc-rsa-pkcs8.pub.pem`
* Repeat for the RSA DER: `/usr/local/opt/openssl/bin/openssl rsa -inform der -RSAPublicKey_in -in vectors/cryptography_vectors/asymmetric/public/PKCS1/rsa.pub.der -text`

public:
Verified. :+1:

PR #1721, opened 2015-03-05
WIP: Start of OpenSSL DH backend implementation

public:


palaviv:
Is there any plan to continue the work on this PR?
@public would it be OK if I will try to make a new attempt at implementing this?

PR #1723, opened 2015-03-05
Stricter asserts on _bn_to_int and _int_to_bn

public:
Helped me catch my terrible programming when working on DH

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -320,6 +320,7 @@ class Backend(object):
         )
 
     def _bn_to_int(self, bn):
+        assert bn != self._ffi.NULL
         if six.PY3:
             # Python 3 has constant time from_bytes, so use that.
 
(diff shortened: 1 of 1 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -346,6 +347,7 @@ def _int_to_bn(self, num, bn=None):
         ownership of the object). Be sure to register it for GC if it will
         be discarded after use.
         """
+        assert (bn is None) or (bn is not None and bn != self._ffi.NULL)
```
alex: I think this simplifies to `assert bn is None or bn != self._ffi.NULL`? The second `bn is not None` is a tautology.

public: `bn = None` means that `bn` is immediately set to `NULL`. However if a non-None value is provided then write the value to that instead.

This assert removes the ambiguity of the `_int_to_bn(0, NULL)` vs `_int_to_bn(0)`.

alex: Right, I don't think the line I proposed is semantically different from what you have, just a bit shorter.

public: You are correct :)

public:
Updated.

reaperhulk:
:+1: will merge once tests pass. C'mon travis.

PR #1738, opened 2015-03-09
Do the deprecation dance

reaperhulk:


alex:
Looks like there's still some references to `CMACContext` in `docs/hazmat/backends/interfaces.rst` and `docs/hazmat/primitives/interfaces.rst`

PR #1747, opened 2015-03-11
remove unneeded deprecated interface registration

reaperhulk:


alex:
`DSAPublicKey` and `ECPublicKey` need to be updated to serialization

Issue #1753, opened 2015-03-12
DER serialization of private keys


reaperhulk:
Tracking issue for landing support for this in the OpenSSL backend:

- [x] RSA
- [x] DSA
- [x] Elliptic curve

PR #1754, opened 2015-03-13
support RSA DER private key serialization

reaperhulk:
Support RSA + the framework to support EC/DSA (future PRs).

refs #1753

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -1133,24 +1133,6 @@ class Backend(object):
                 "format must be an item from the PrivateFormat enum"
             )
 
-        # This is a temporary check until we land DER serialization.
-        if encoding is not serialization.Encoding.PEM:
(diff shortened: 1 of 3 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1191,6 +1204,19 @@ def _private_key_bytes(self, encoding, format, encryption_algorithm,
...
+    def _private_key_bytes_traditional_der(self, type, cdata):
+        if type == self._lib.EVP_PKEY_RSA:
+            write_bio = self._lib.i2d_RSAPrivateKey_bio
+        else:
+            raise TypeError(
+                "Only RSA keys are supported for DER serialization"
+            )
+
+        bio = self._create_mem_bio()
+        res = write_bio(bio, cdata)
```
alex: I assume the `write_bio` indrection here is because there'll be more branches on that `if` shortly?

reaperhulk: yep

PR #1759, opened 2015-03-15
add keywrap test vectors from NIST

reaperhulk:
From http://csrc.nist.gov/groups/STM/cavp/documents/mac/kwtestvectors.zip


Ayrx:
That's definitely gonna bloat our vectors package. :P

chelseawinfree:
I have verified these vectors via diff on command line.

I took the zip file and compared the files there to the branch for this request. No changes were determined via diff.

PR #1772, opened 2015-03-23
Attempt at simplifying the repr code.

alex:


alex:
I think this is a bit of an improvement.

PR #1773, opened 2015-03-23
Property based testing with Hypothesis

Ayrx:
So I have been playing around with [Hypothesis](http://hypothesis.readthedocs.org/en/latest/#where-hypothesis-comes-from) a bit and it seems like a very useful addition to our test suite.

This PR sets up the groundwork needed to start using Hypothesis in the cryptography test suite, as well as a single test (round trip encrypt-decrypt for Fernet) to start off with. 

Hypothesis currently does not (and probably does not plan to) support Python 2.6. The approach I have taken is to not install Hypothesis in the Python 2.6 tox target and use `pytest.importorskip` to skip the test if the import fails.

Thoughts?


public:
I think this is a useful thing to try and integrate too but I've mostly had trouble working out how to actually do that so far. e.g. to test ASN.1 based stuff properly we might need to write our own test data generator because plain strings don't really cut it.

I think this should probably be a separate `tox` target or something so we can run it separately from the main suite.

setup.py
```diff
@@ -53,6 +53,10 @@
     "iso8601",
 ]
 
+if sys.version_info[:2] == (2, 6):
```
public: I think this is backwards?

Ayrx: Ooops, yes. :)

DRMacIver: Isn't this test inverted? I think you want to add it when it's *not* python 2.6.

DRMacIver: Oops. Race condition in commenting :-)

Ayrx:
retest this please

Ayrx:
@koobs Seems like our FreeBSD builders have issue with the sqlite3 dependency. Any ideas on how to go about fixing that?

Ayrx:
retest this please

Ayrx:
Tracked down the Windows failure. See DRMacIver/hypothesis#44. The FreeBSD failure is a problem with our builders and has to be fixed by @reaperhulk. :)

reaperhulk:
Where do we stand on this @ayrx? Is it worth pursuing still or should we drop it for now?

Ayrx:
@reaperhulk I *want* this to go through but I can't figure out a fix for the Windows failure. :( I hate unicode.

Probably close this unless someone can figure that problem out.

DRMacIver:
I thought that the windows problem had been solved as you just needing to convert PYTHONPATH to a str before passing it to a subprocess? Did that not work?

DRMacIver:
I'd left this alone because I assumed you'd solved the problem, but the above commit should work around your issue (I still think your code is wrong, but it's the sort of code that is very hard to get right cross-platform). I'm planning to do a point release today that will include it.

However this has been open long enough that the API is moved under you and this PR will now emit deprecation warnings. :-) You need to change this code to use 'from hypothesis import binary' and 'binary()' instead of 'bytes'.

Ayrx:
jenkins retest this please

reaperhulk:
jenkins, retest this please

reaperhulk:
jenkins, retest this please

Ayrx:
Ok, looks like this PR is all good to go thanks to @DRMacIver! 

Are there any feedbacks I need to address? @alex @public @reaperhulk

reaperhulk:
How do you envision us expanding our use of hypothesis going forward?

Ayrx:
We can definitely expand this to most if not all of our public interfaces at the very least.

reaperhulk:
This is a hypothesis ignorant question, but what kind of performance should we expect from tests like this? Can we keep them in our standard test suite without worrying about it?

DRMacIver:
Hypothesis itself is pretty fast, particularly for simple things like generating byte sequences. It's not *outrageously* fast, but generally it's faster than the tests you run. If that's not the case then I will usually consider it a performance bug in Hypothesis. 

It *does* run those tests 200 times by default though, so if your tests you use Hypothesis with are not themselves reasonably fast then the tests you add using Hypothesis can be on the slow side. If this starts to present a problem you can turn the numbers down, focus on the speed of the the tests you're running about it, or nag me with specific Hypothesis performance bugs as you find them. :-)

reaperhulk:
Okay, since this adds a new test dep I'd like to hear from @alex before we merge, but I'm :+1:. Looking forward to seeing this as it expands to more of our public interfaces.

tests/test_with_hypothesis.py
```diff
@@ -0,0 +1,15 @@
+from hypothesis.strategies import binary
+
+import pytest
+
+from cryptography.fernet import Fernet
+
+
+hypothesis = pytest.importorskip("hypothesis")
```
alex: When would this import fail, but the first one succeed?

tests/test_with_hypothesis.py
```diff
@@ -0,0 +1,15 @@
+from hypothesis.strategies import binary
```
alex: I think this path should be `tests/hypothesis/test_fernet.py` or something

alex:
@reaperhulk left a couple of comments, looks ok to me.

tests/test_with_hypothesis.py
Ayrx: I think that's a relic from when hypothesis did not have 2.6 support. I'll
remove that when I get to a computer.

reaperhulk: If it has 2.6 support then the setup.py requirement needs to change as well.

Ayrx:
@reaperhulk Are there any other issues I missed out addressing?

reaperhulk:
Nope, looks like you still need to rename the test file and put it under a hypothesis subdir, but then we're good.

Ayrx:
@reaperhulk All done.

Issue #1301, opened 2014-08-11
Move OpenSSL engine into standalone .c and .h files


alex:
Right now it's in a Python string, that's pretty crummy.

PR #1774, opened 2015-03-23
Moved OpenSSL engine into standalone c and h files

chelseawinfree:
References #1301


reaperhulk:
add to whitelist

reaperhulk:
I can't merge this because I paired with @chelseawinfree for this work.

alex:
@reaperhulk thanks for the heads up.

I'll review after the Bulls game.

src/cryptography/hazmat/bindings/openssl/osrandom_engine.py
```diff
@@ -4,14 +4,12 @@
...
-#else
-#include <fcntl.h>
-#include <unistd.h>
-#endif
-"""
+import os
+
+with open(os.path.join(
+    os.path.dirname(__file__), "src/osrandom_engine.h")
+) as f:
```
alex: Slight nitpick, but can we change how this is written so that the parens are aligned:

```
with open(os.path.join(
    os.path.dirname(__file__), "src/osrandom_engine.h"
)) as f:
```

src/cryptography/hazmat/bindings/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,167 @@
...
+    if (!ENGINE_add(e)) {
+        ENGINE_free(e);
+        return 0;
+    }
+    if (!ENGINE_free(e)) {
+        return 0;
+    }
+
+    return 1;
+}
```
alex: Can you add trailing new lines to this and the `.h` file so they match the other source files.

alex:
Overall this looks great!

I haven't actually verified that the moved code is identical though -- can anyone think of a relatively easy way to do that?

Also, is `#if defined(_WIN32)` the recommend way to check for windows?

reaperhulk:
`_WIN32` seems to be the common way. MSFT has a list of defined macros here: https://msdn.microsoft.com/en-us/library/b0084kay(VS.80).aspx

CPython, OpenSSL, zlib, et al also use the `#if defined(_WIN32)` check.

alex:
Will that be correct on cygwin? (Hell which even /is/ correct on Cygwin)

On Mon, Mar 23, 2015 at 10:55 PM, Paul Kehrer <notifications@github.com>
wrote:

> _WIN32 seems to be the common way. MSFT has a list of defined macros
> here: https://msdn.microsoft.com/en-us/library/b0084kay(VS.80).aspx
>
> CPython, OpenSSL, zlib, et al also use the #if defined(_WIN32) check.
>
> —
> <https://github.com/pyca/cryptography/pull/1774#issuecomment-85317414>.
>



-- 

reaperhulk:
I believe you have to check for `__CYGWIN__`? But I don't know what you'd want to do there...maybe POSIX would work...?

alex:
@reaperhulk We want to maintain the same behavior as now, which is that cygwin uses the POSIX (`/dev/urandom`) path -- does this do that?

reaperhulk:
Yep, this code is functionally identical to the previous. We just did the string interpolation and then dumped it into a separate file.

reaperhulk:
@chelseawinfree feel free to send a PR adding yourself to `AUTHORS` if you'd like. Thanks!

alex:
Please do!

On Tue, Mar 24, 2015 at 6:54 PM, Paul Kehrer <notifications@github.com>
wrote:

> @chelseawinfree <https://github.com/chelseawinfree> feel free to send a
> PR adding yourself to AUTHORS if you'd like. Thanks!
>
> —
> <https://github.com/pyca/cryptography/pull/1774#issuecomment-85726197>.
>



-- 

mboersma:
Note that this change in cryptography 0.9 breaks its use under [PyInstaller](https://github.com/pyinstaller/pyinstaller), even the tip version. Its heuristics for what supporting files must be included to use cryptography fail now, but worked just fine until version 0.8.2:
```
 File "/var/lib/jenkins/jobs/test-ec2/workspace@2/src/github.com/deis/deis/client/build/deis/out00-PYZ.pyz/cryptography.hazmat.bindings.openssl.osrandom_engine", line 10, in <module>
IOError: [Errno 2] No such file or directory: '/tmp/_MEIfxEnRz/cryptography/hazmat/bindings/openssl/src/osrandom_engine.h'
error at command wait
```

reaperhulk:
@mboersma ugh. cffi 1.0 is going to change everything around this though (and should improve life for pyinstaller, cx_freeze, etc) so for the moment the workaround may need to be pin to 0.8.2. I am hopeful that the next release of cryptography will be using cffi 1.0.

PR #1769, opened 2015-03-21
basicConstraints support for OpenSSL X509 backend

reaperhulk:
The first extension.

depends on #1768 
refs #1743

PR #1780, opened 2015-03-25
allow zero value bn using _bn_to_int in the openssl backend

reaperhulk:
BNs can be zero (for example, path_length in basic constraints). Zero worked for py2 but failed on the py3 path.

This is needed for #1769


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -327,7 +327,7 @@ def _bn_to_int(self, bn):
             bn_num_bytes = (self._lib.BN_num_bits(bn) + 7) // 8
             bin_ptr = self._ffi.new("unsigned char[]", bn_num_bytes)
             bin_len = self._lib.BN_bn2bin(bn, bin_ptr)
-            assert bin_len > 0
+            assert bin_len >= 0  # 0 length means the BN has value 0
```
alex: I think the style we use is comments above the line of code, not next to it.

public:
That makes sense.

Issue #1743, opened 2015-03-11
X509 Certificate Extensions


reaperhulk:
This is a tracking issue for landing support for all extensions in [RFC 5280](http://tools.ietf.org/html/rfc5280).

Must support:
- [x] Authority Key Identifier
- [x] Subject Key Identifier
- [x] Key Usage
- [x] Subject Alternative Name
- [x] Basic Constraints
- [x] Extended Key Usage
- [x] CRL Distribution Points
- [x] Authority Information Access
- [x] Certificate Policies

kevgliss:
+1 For this I have almost completed a port of my app off of M2Crypto onto this library. But I am still missing he ability to grab data from the subjectAltName extension.

schlenk:
Maybe add Certificate to the issue title, as your list misses the X509 CRL extensions. But in general a pretty good thing and probably much nicer than having to use pyasn1 to parse all those by hand.

reaperhulk:
Good call @schlenk, I've added certificate to the title to prevent confusion. CRL extensions will come later :)

PR #1783, opened 2015-03-25
Add GeneralName and SubjectAlternativeName classes

reaperhulk:
refs #1743

### DIFF
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -399,6 +399,80 @@ X.509 CSR (Certificate Signing Request) Object
 
         The dotted string value of the OID (e.g. ``"2.5.4.3"``)
 
+.. _general_name_classes:
+
+General Name Classes
+~~~~~~~~~~~~~~~~~~~~
+
+.. class:: GeneralName
+
+    .. versionadded:: 0.9
+
+    This is the generic interface that all the following classes are registered
+    against.
+
+.. class:: RFC822Name
+
+    .. versionadded:: 0.9
+
+    This corresponds to an email address. For example, ``user@example.com``.
+
+    .. attribute:: value
+
+        :type: :term:`text`
+
+.. class:: DNSName
+
+    .. versionadded:: 0.9
+
+    This corresponds to a domain name. For example, ``cryptography.io``.
+
+    .. attribute:: value
+
+        :type: :term:`text`
+
+.. class:: DirectoryName
+
+    .. versionadded:: 0.9
+
+    This corresponds to a directory name.
+
+    .. attribute:: value
+
+        :type: :class:`Name`
+
+.. class:: UniformResourceIdentifier
+
+    .. versionadded:: 0.9
+
+    This corresponds to a uniform resource identifier.  For example,
+    ``https://cryptography.io``.
+
+    .. attribute:: value
+
+        :type: :term:`text`
+
+.. class:: IPAddress
+
+    .. versionadded:: 0.9
+
+    This corresponds to an IP address.
+
+    .. attribute:: value
+
+        :type: :class:`~ipaddress.IPv4Address` or
+            :class:`~ipaddress.IPv6Address`.
+
+.. class:: RegisteredID
+
+    .. versionadded:: 0.9
+
+    This corresponds to a registered ID.
+
+    .. attribute:: value
+
+        :type: :class:`ObjectIdentifier`
+
 X.509 Extensions
 ~~~~~~~~~~~~~~~~
 
@@ -590,6 +664,22 @@ X.509 Extensions
         :type: bytes
 
         The binary value of the identifier.
+
+.. class:: SubjectAlternativeName
+
+    .. versionadded:: 0.9
+
+    Subject alternative name is an X.509 extension that provides a list of
+    :ref:`general name <general_name_classes>` instances that provide a set
+    of identities for which the certificate is valid. The object is iterable to
+    get every element.
+
+    .. method:: get_values_for_type(type)
+
+        :param type: A :class:`GeneralName` provider. This is one of the
+            :ref:`general name classes <general_name_classes>`.
+
+        :returns: A list of values extracted from the matched general names.
 
 
 Object Identifiers
--- a/src/cryptography/x509.py
+++ b/src/cryptography/x509.py
@@ -5,6 +5,7 @@
 from __future__ import absolute_import, division, print_function
 
 import abc
+import ipaddress
 from enum import Enum
 
 import six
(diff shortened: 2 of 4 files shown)


docs/x509.rst
```diff
@@ -273,6 +273,70 @@ X.509 Certificate Object
...
+
+    A general name is a paired type and value.
+
+    .. attribute:: type
+
+        :type: :class:`GeneralNameType`
+
+    .. attribute:: value
+
+        The value of the GeneralName. The type of this data is defined by the
```
reaperhulk: This isn't a great way to explain what Python type the value will have. Ideas?

docs/x509.rst
```diff
@@ -327,6 +391,21 @@ X.509 Extensions
...
 
+.. class:: SubjectAlternativeName
+
+    .. versionadded:: 0.9
+
+    Subject alternative name is an X.509 extension type that provides a list of
+    :class:`GeneralName` objects that provide identities bound to the
+    certificate. The object is iterable to get every :class:`GeneralName`.
+
+    .. method:: get_values_for_type(type)
```
reaperhulk: This method would not return a list of GeneralName objects. If we want that we could also add a `get_general_names_for_type`.

tests/test_x509_ext.py
```diff
@@ -55,3 +55,104 @@ def test_repr(self):
...
+    def test_iter_names(self):
+        gns = [
+            x509.GeneralName(x509.GeneralNameType.dns_name, "cryptography.io"),
+            x509.GeneralName(
+                x509.GeneralNameType.dns_name, "www.cryptography.io"
+            ),
+        ]
+        san = x509.SubjectAlternativeName(gns)
+        assert len(san) == len(gns)
+        # This assert sucks. It just does an identity test so __eq__ never gets
```
reaperhulk: My brain is fried so this stayed in this initial PR. Suggestions on how to do a better test here welcome.

reaperhulk:
Landing this PR will be very useful as GeneralName is used in many extensions (including AuthorityKeyIdentifier, AuthorityInformationAccess, CRLDistributionPoints, SubjectAlternativeName, IssuerAlternativeName, SubjectInformationAccess, and NameConstraints).

docs/x509.rst
```diff
@@ -397,6 +397,70 @@ X.509 CSR (Certificate Signing Request) Object
...
+
+    .. attribute:: uniform_resource_identifier
+
+        This type corresponds to a uniform resource identifier.  For example,
+        ``https://cryptography.io``. Values of this type will be strings.
+
+    .. attribute:: ip_address
+
+        This type corresponds to an IP address. Values of this type will be
+        strings.
```
alex: I thought we were going to use the ipaddr module here?

src/cryptography/x509.py
```diff
@@ -48,6 +49,7 @@
     "1.3.6.1.5.5.7.3.4": "emailProtection",
     "1.3.6.1.5.5.7.3.8": "timeStamping",
     "1.3.6.1.5.5.7.3.9": "OCSPSigning",
+    "2.5.29.17": "subjectAltName",
```
alex: oid-info tells me that 2.5.29.7 is also SAN, this the right SAN?

src/cryptography/x509.py
```diff
@@ -249,6 +264,96 @@ def __repr__(self):
...
+        if (
+            type in self._gn_str_types and not
+            isinstance(value, six.string_types)
+        ):
+            raise TypeError(
+                "value must be a string when type is {0}".format(type)
+            )
+
+        if (
+            type is GeneralNameType.directory_name and not
```
alex: Please move the not to teh next line

src/cryptography/x509.py
```diff
@@ -249,6 +264,96 @@ def __repr__(self):
...
+
+        if type in [
+            GeneralNameType.other_name,
+            GeneralNameType.x400_address,
+            GeneralNameType.edi_party_name
+        ]:
+            raise ValueError("{0} is not a supported type".format(type))
+
+        if (
+            type in self._gn_str_types and not
```
alex: Same up here

src/cryptography/x509.py
```diff
@@ -249,6 +264,96 @@ def __repr__(self):
...
+            isinstance(value, six.string_types)
```
alex: Do we really want to allow both text and bytes on py2?

reaperhulk: I'm not sure. @glyph has stated to me in the past that he strongly considers domain names, et al to be strings and not bytes. They are, however, a strict subset of ascii (a-z, A-Z, 0-9, and hyphen). It seems odd to require u"mydomain.com" on Python 2. I chose string_types because it allows `str` on both Py2 and Py3, but you're right that it also allows `b'string'` on Py2 and not on Py3. Maybe the lesser of two evils is to just say you pass unicode (and potentially do charset validation in the future when we allow creation)?

alex: Are they a strict subset of ascii? What format are they in here? Unicode
domain names are definitely a thing, see punycode and such.

On Tue, Mar 31, 2015 at 9:35 PM, Paul Kehrer <notifications@github.com>
wrote:

> In src/cryptography/x509.py
> <https://github.com/pyca/cryptography/pull/1783#discussion_r27539037>:
>
> > +    ]
> > +
> > +    def __init__(self, type, value):
> > +        if not isinstance(type, GeneralNameType):
> > +            raise TypeError("type must be a GeneralNameType")
> > +
> > +        if type in [
> > +            GeneralNameType.other_name,
> > +            GeneralNameType.x400_address,
> > +            GeneralNameType.edi_party_name
> > +        ]:
> > +            raise ValueError("{0} is not a supported type".format(type))
> > +
> > +        if (
> > +            type in self._gn_str_types and not
> > +            isinstance(value, six.string_types)
>
> I'm not sure. @glyph <https://github.com/glyph> has stated to me in the
> past that he strongly considers domain names, et al to be strings and not
> bytes. They are, however, a strict subset of ascii (a-z, A-Z, 0-9, and
> hyphen). It seems odd to require u"mydomain.com" on Python 2. I chose
> string_types because it allows str on both Py2 and Py3, but you're right
> that it also allows b'string' on Py2 and not on Py3. Maybe the lesser of
> two evils is to just say you pass unicode (and potentially do charset
> validation in the future when we allow creation)?
>
> —
> <https://github.com/pyca/cryptography/pull/1783/files#r27539037>.
>



-- 

reaperhulk: IDNA domains are encoded in the certificate as that strict subset. For example, Unicode biohazard (or radiation I can never remember) is xn--k4h. That said, we could potentially choose to parse that data and return the Unicode rather than the decomposed version that's actually present in the certificate... Is there a decent IDNA parser library for Python?

alex: The standard lib has it `.decode("idna")`

On Tue, Mar 31, 2015 at 9:46 PM, Paul Kehrer <notifications@github.com>
wrote:

> In src/cryptography/x509.py
> <https://github.com/pyca/cryptography/pull/1783#discussion_r27539362>:
>
> > +    ]
> > +
> > +    def __init__(self, type, value):
> > +        if not isinstance(type, GeneralNameType):
> > +            raise TypeError("type must be a GeneralNameType")
> > +
> > +        if type in [
> > +            GeneralNameType.other_name,
> > +            GeneralNameType.x400_address,
> > +            GeneralNameType.edi_party_name
> > +        ]:
> > +            raise ValueError("{0} is not a supported type".format(type))
> > +
> > +        if (
> > +            type in self._gn_str_types and not
> > +            isinstance(value, six.string_types)
>
> IDNA domains are encoded in the certificate as that strict subset. For
> example, Unicode biohazard (or radiation I can never remember) is xn--k4h.
> That said, we could potentially choose to parse that data and return the
> Unicode rather than the decomposed version that's actually present in the
> certificate... Is there a decent IDNA parser library for Python?
> … <#14c72a97a31a3306_>
>  On Mar 31, 2015, at 8:41 PM, Alex Gaynor <notifications@github.com>
> wrote: In src/cryptography/x509.py: > + ] > + > + def __init__(self, type,
> value): > + if not isinstance(type, GeneralNameType): > + raise
> TypeError("type must be a GeneralNameType") > + > + if type in [ > +
> GeneralNameType.other_name, > + GeneralNameType.x400_address, > +
> GeneralNameType.edi_party_name > + ]: > + raise ValueError("{0} is not a
> supported type".format(type)) > + > + if ( > + type in self._gn_str_types
> and not > + isinstance(value, six.string_types) Are they a strict subset of
> ascii? What format are they in here? Unicode domain names are definitely a
> thing, see punycode and such. … — Reply to this email directly or view it
> on GitHub.
>
> —
> <https://github.com/pyca/cryptography/pull/1783/files#r27539362>.
>



-- 

reaperhulk: I did not know that...let me modify this PR a little bit to support IDNA domains then.

glyph: I don't see "idna" anywhere in this PR's changed files, so I'm assuming that while the hunk may be outdated, the sentiment isn't.

You should know that the stdlib "idna" encoding is somewhat out of date, and according to @hynek you should prefer [the separately-distributed `idna` package](https://pypi.python.org/pypi/idna).

reaperhulk: The change for this PR is just to disallow bytes types in their entirety. The PR that actually implements the OpenSSL backend support for this will add the IDNA parsing. Unfortunately it looks like we're going to need to add Python 3.2 support to that package if we want to use it, as it currently only supports 3.3+, but I'll look into it.

glyph:
Since my name came up, I should say that the *reason* I strongly consider domain names to be text and not bytes is that there are numerous places where it is critical, both for security and functionality, for the octets which comprise a domain name to turn into letters and then have photons bounce off the letters and into a human eyeball.

IDNA is a bit weird in that they are text *encoded as other text* which happens to be ASCII and therefore representable as bytes in a canonical encoding; in other words, sometimes you get the `ʕ•͡ᴥ•ʔ.com` and sometimes the `xn--hpab62dq83fiedca.com` gets you.  But either is a valid representation as text which a human may need to consume in some context.  (Actually it isn't; this domain is an example of a thing that stdlib `.encode("idna")` will accept but the newer `import idna` will reject as invalid.)

I think that the direction the branch is currently going, just require unicode everywhere and reject bytes, is the right call.  If py2 users don't want to do `u"..."` then they can always `from __future__ import unicode_literals` :).

alex:
These needs a rebase.

alex:
This needs a rebase, no longer merge clean.

docs/x509.rst
```diff
@@ -399,6 +399,72 @@ X.509 CSR (Certificate Signing Request) Object
...
+
+        This is a rarely used type and is not currently supported.
+
+    .. attribute:: x400_address
+
+        This is a rarely used type and is not currently supported.
+
+    .. attribute:: edi_party_name
+
+        This is a rarely used type and is not currently supported.
```
alex: Is there any reason to even expose these unsupported types?

docs/x509.rst
```diff
@@ -591,6 +657,21 @@ X.509 Extensions
 
         The binary value of the identifier.
 
+.. class:: SubjectAlternativeName
+
+    .. versionadded:: 0.9
+
+    Subject alternative name is an X.509 extension type that provides a list of
+    :class:`GeneralName` objects that provide identities bound to the
+    certificate. The object is iterable to get every :class:`GeneralName`.
```
alex: This paragraph feels pretty incomprehensible to me :-/

src/cryptography/x509.py
```diff
@@ -387,6 +400,101 @@ def __ne__(self, other):
         return not self == other
 
 
+class GeneralName(object):
+    _gn_str_types = [
```
alex: s/str/text/

alex:
The docs could use some work in terms of explaining how stuff actually works. e.g. what does a certificate with a SAN for a simple domain look like.

Also, in terms of the class structure:

In Bagel you'd model this as a single enum with a few choices which contain stuff:

```
enum type GeneralName
     case RFC822Name(Text)
     case DNSName(Text)
     case IPAddress(net.IP)
     case DirectoryName(x509.Name)
```

is there any reasonable way to get closer to that representation in Python?

reaperhulk:
We could define a `GeneralName` interface and then make separate classes for each name type we support? So we'd end up with classes like RFC822Name, DNSName, IPAddress, etc but we can check if they're legal GeneralName objects with an isinstance check.

alex:
That seems like it might be slightly cleaner?

On Tue, Apr 14, 2015 at 9:24 AM, Paul Kehrer <notifications@github.com>
wrote:

> We could define a GeneralName interface and then make separate classes
> for each name type we support? So we'd end up with classes like RFC822Name,
> DNSName, IPAddress, etc but we can check if they're legal GeneralName
> objects with an isinstance check.
>
> —
> <https://github.com/pyca/cryptography/pull/1783#issuecomment-92830365>.
>



-- 

Issue #1790, opened 2015-03-28
Test failures due to \r\n line breaks on Windows.


Ayrx:
Test failures when running `tox -e py34 -- tests\hazmat\primitives\test_rsa.py` Environment is Windows 8.1 testing against `master`. I'm worried about why this wasn't caught by Jenkins. 

```
================================== FAILURES ===================================
 TestRSAPrivateKeySerialization.test_private_bytes_traditional_openssl_unencrypted[backend0-asymmetric\Traditional_OpenSSL_Serialization\testrsa.pem-encoding0-loader_func0]

self = <tests.hazmat.primitives.test_rsa.TestRSAPrivateKeySerialization object at 0x06C900B0>
backend = <cryptography.hazmat.backends.openssl.backend.Backend object at 0x03A98330>
key_path = 'asymmetric\\Traditional_OpenSSL_Serialization\\testrsa.pem'
encoding = <Encoding.PEM: 'PEM'>
loader_func = <function load_pem_private_key at 0x04343C90>

    @pytest.mark.parametrize(
        ("key_path", "encoding", "loader_func"),
        [
            [
                os.path.join(
                    "asymmetric",
                    "Traditional_OpenSSL_Serialization",
                    "testrsa.pem"
                ),
                serialization.Encoding.PEM,
                serialization.load_pem_private_key
            ],
            [
                os.path.join("asymmetric", "DER_Serialization", "testrsa.der"),
                serialization.Encoding.DER,
                serialization.load_der_private_key
            ],
        ]
    )
    def test_private_bytes_traditional_openssl_unencrypted(
        self, backend, key_path, encoding, loader_func
    ):
        key_bytes = load_vectors_from_file(
            key_path, lambda pemfile: pemfile.read(), mode="rb"
        )
        key = loader_func(key_bytes, None, backend)
        serialized = key.private_bytes(
            encoding,
            serialization.PrivateFormat.TraditionalOpenSSL,
            serialization.NoEncryption()
        )
>       assert serialized == key_bytes
E       assert b'-----BEGIN ...TE KEY-----\n' == b'-----BEGIN R... KEY-----\r\n'
E         At index 31 diff: 10 != 13
E         Right contains more items, first extra item: 69
E         Use -v to get the full diff

tests\hazmat\primitives\test_rsa.py:1877: AssertionError
[...]

PR #1792, opened 2015-03-28
add strip to some tests because we don't care about newlines

reaperhulk:


alex:
I'm not convinced strip is correct here, I feel like we should make a guarntee about what the backend emits in terms of newlines.

reaperhulk:
I'd be willing to strip in the backend such that our "guarantee" is that backends don't return leading or trailing newlines.

public:
Should it not be the other way around? That we should make sure backends always return PEM with a trailing new line?

alex:
I think I agree, returning a result with a single trailing `\n` makes sense.

On Sat, Mar 28, 2015 at 2:26 PM, Alex Stapleton <notifications@github.com>
wrote:

> Should it not be the other way around? That we should make sure backends
> always return PEM with a trailing new line?
>
> —
> <https://github.com/pyca/cryptography/pull/1792#issuecomment-87278988>.
>



-- 

reaperhulk:
I'm interested in why it should matter to people if there's a trailing newline or not? Is the idea that people are going to write multiple keys/certs to the same output file? There are several software packages that do ask you to concat cert+key I guess.

alex:
Concatenating certs, placing them in plaintext formatted media, hell just
printing it out to stdout.

On Thu, Apr 2, 2015 at 8:32 AM, Paul Kehrer <notifications@github.com>
wrote:

> I'm interested in why it should matter to people if there's a trailing
> newline or not? Is the idea that people are going to write multiple
> keys/certs to the same output file? There are several software packages
> that do ask you to concat cert+key I guess.
>
> —
> <https://github.com/pyca/cryptography/pull/1792#issuecomment-88898405>.
>



-- 

public:
It's not entirely clear in the RFC but it does seem to imply that stuff should be new line terminated to me.

schlenk:
This is similar to issue #1790. I think the fix in PR #1817 is better way to handle this case.

PR #1793, opened 2015-03-28
add x509 extensions class and basic tests (no extensions supported)

reaperhulk:
Had to remove ``get_extension_for_oid`` since it's impossible to fully cover that method without a supported extension. Also had to move ``seen_oids`` to an inappropriate place temporarily, but it will be moved back in the next PR.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -14,6 +14,7 @@
 from __future__ import absolute_import, division, print_function
 
 import datetime
+import warnings
 
 from cryptography import utils, x509
 from cryptography.exceptions import UnsupportedAlgorithm
@@ -151,3 +152,38 @@ class _Certificate(object):
             raise UnsupportedAlgorithm(
                 "Signature algorithm OID:{0} not recognized".format(oid)
             )
+
+    @property
+    def extensions(self):
+        extensions = []
+        seen_oids = set()
+        extcount = self._backend._lib.X509_get_ext_count(self._x509)
+        for i in range(0, extcount):
+            ext = self._backend._lib.X509_get_ext(self._x509, i)
+            assert ext != self._backend._ffi.NULL
+            crit = self._backend._lib.X509_EXTENSION_get_critical(ext)
+            critical = crit == 1
+            oid = x509.ObjectIdentifier(_obj2txt(self._backend, ext.object))
+            if oid in seen_oids:
+                raise x509.DuplicateExtension(
+                    "Duplicate {0} extension found".format(oid), oid
+                )
+            elif oid == x509.OID_BASIC_CONSTRAINTS and critical:
+                # TODO: remove this obviously.
+                warnings.warn(
+                    "Extension support is not fully implemented. A basic "
+                    "constraints extension with the critical flag was seen and"
+                    " IGNORED."
+                )
+            elif critical:
+                raise x509.UnsupportedExtension(
+                    "{0} is not currently supported".format(oid), oid
+                )
+            else:
+                # Unsupported non-critical extension, silently skipping for now
+                seen_oids.add(oid)
+                continue
+
+            seen_oids.add(oid)
+
+        return x509.Extensions(extensions)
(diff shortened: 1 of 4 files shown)


src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -151,3 +152,38 @@ def signature_hash_algorithm(self):
...
+                )
+            elif oid == x509.OID_BASIC_CONSTRAINTS and critical:
+                # TODO: remove this obviously.
+                warnings.warn(
+                    "Extension support is not fully implemented. A basic "
+                    "constraints extension with the critical flag was seen and"
+                    " IGNORED."
+                )
+                # This will move to the else block in the next PR
+                seen_oids.add(oid)
```
alex: I think we want to drop the continue from here, and below, and put stuff in `seen_oids` -- we should raise DuplicateExtension if ignored extensions are duplicated as well.

Does that make sense?

reaperhulk: That's fine for this PR, but I'll have to add continue back in next PR since we'll be actually building an extension in this loop at that point.

alex: k

On Sat, Mar 28, 2015 at 11:55 AM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/1793#discussion_r27345980>:
>
> > +            crit = self._backend._lib.X509_EXTENSION_get_critical(ext)
> > +            critical = crit == 1
> > +            oid = x509.ObjectIdentifier(_obj2txt(self._backend, ext.object))
> > +            if oid in seen_oids:
> > +                raise x509.DuplicateExtension(
> > +                    "Duplicate {0} extension found".format(oid)
> > +                )
> > +            elif oid == x509.OID_BASIC_CONSTRAINTS and critical:
> > +                # TODO: remove this obviously.
> > +                warnings.warn(
> > +                    "Extension support is not fully implemented. A basic "
> > +                    "constraints extension with the critical flag was seen and"
> > +                    " IGNORED."
> > +                )
> > +                # This will move to the else block in the next PR
> > +                seen_oids.add(oid)
>
> That's fine for this PR, but I'll have to add continue back in next PR
> since we'll be actually building an extension in this loop at that point.
>
> —
> <https://github.com/pyca/cryptography/pull/1793/files#r27345980>.
>



-- 

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -151,3 +152,38 @@ def signature_hash_algorithm(self):
...
+                    " IGNORED."
+                )
+                # This will move to the else block in the next PR
+            elif critical:
+                raise x509.UnsupportedExtension(
+                    "{0} is not currently supported".format(oid)
+                )
+            else:
+                # Unsupported non-critical extension, silently skipping for now
+                continue  # pragma: no cover
```
alex: This `else` should be dropped right now -- we want the `oid` to end up in `seen_oids` even if we're ignoring it.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -151,3 +152,38 @@ def signature_hash_algorithm(self):
...
+            elif oid == x509.OID_BASIC_CONSTRAINTS and critical:
+                # TODO: remove this obviously.
+                warnings.warn(
+                    "Extension support is not fully implemented. A basic "
+                    "constraints extension with the critical flag was seen and"
+                    " IGNORED."
+                )
+            elif critical:
+                raise x509.UnsupportedExtension(
+                    "{0} is not currently supported".format(oid)
```
alex: Oof, one more thing: can we put the `oid` on `UnsupportedExtension` as structured data.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -151,3 +152,38 @@ def signature_hash_algorithm(self):
...
+        for i in range(0, extcount):
+            ext = self._backend._lib.X509_get_ext(self._x509, i)
+            assert ext != self._backend._ffi.NULL
+            crit = self._backend._lib.X509_EXTENSION_get_critical(ext)
+            critical = crit == 1
+            oid = x509.ObjectIdentifier(_obj2txt(self._backend, ext.object))
+            if oid in seen_oids:
+                raise x509.DuplicateExtension(
+                    "Duplicate {0} extension found".format(oid)
+                )
```
alex: Same here.

PR #1794, opened 2015-03-28
[WIP] AES keywrap support

reaperhulk:
This is being put up for general review of approach. multibackend support is missing and it may be worth implementing EVP keywrap support for 1.0.2. Additionally, we should decide if the supported check on the backend makes sense. Unfortunately OpenSSL didn't add key wrap support until 0.9.8h, which CentOS 5 does not have.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,9 @@ Changelog
   :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePrivateKeyWithSerialization`.
 * Add support for parsing X.509 certificate signing requests (CSRs) with
   :func:`~cryptography.x509.load_pem_x509_csr`.
+* Add support for AES key wrapping with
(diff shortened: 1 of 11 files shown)


reaperhulk:
It appears OpenSSL < 0.9.8p has a bug in its keywrap implementation. The commits that resolved this are here: https://github.com/openssl/openssl/commit/ae3b60ba99033ff06da382b764a9f50134e30806 and https://github.com/openssl/openssl/commit/78dcaa06094b68a21bf390abac91958a8522a220

reaperhulk:
This is stalled while I consider how to handle RHEL < 6.5 having an OpenSSL that claims to be 1.0.0b in the version number (not the version text!), but still appears to have the bug.

reaperhulk:
I think the answer here is probably to just implement this in Python (or not support it). The AES encryption/decryption operation can still be performed using the backend but we would pass the IV and verify the results. I'll put up a WIP version of this so we can review the approach and decide if we're comfortable with it.

reaperhulk:
closing in favor of #1970 at least for now.

PR #1320, opened 2014-08-25
DH backend interfaces

public:
DHBackend and docs.

PR #1727, opened 2015-03-06
DH key interfaces

public:
Like #1316 but minus the stuff now in #1726

PR #1795, opened 2015-03-28
DH backend interfaces

public:
#1320 was getting a bit stale.


reaperhulk:
You've got a docs failure here and unfortunately also need to rebase. :cry:

public:
Yeah tests will fail until #1727 is merged.

reaperhulk:
docs bug:

```
None:None: WARNING: py:class reference target not found: cryptography.hazmat.primitives.interfaces.DHParameters
```

PR #1796, opened 2015-03-29
Minor refactorings

rev112:
Just a few cleanups and renamings.

Summary of changes:
- Remove redundant parentheses in `openssl/backend.py` 	
- Simplify chained comparison in `openssl/utils.py`
- Rename lambda argument in `openssl/x509.py`
- Rename 'type' variable which shadows built-in name 'type' in `openssl/backend.py`


reaperhulk:
ok to test

PR #1798, opened 2015-03-30
Emit a warning on Python 2.6

alex:


reaperhulk:
I think this needs a changelog entry?

alex:
@reaperhulk done.

PR #1802, opened 2015-03-30
add KeyUsage extension

reaperhulk:
refs #1743

### DIFF
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -447,6 +447,99 @@ X.509 Extensions
 
         Returns an instance of the extension type corresponding to the OID.
 
+.. class:: KeyUsage
+
+    .. versionadded:: 0.9
+
+    The key usage extension defines the purpose of the key contained in the
+    certificate.  The usage restriction might be employed when a key that could
+    be used for more than one operation is to be restricted. It corresponds to
+    :data:`OID_KEY_USAGE`.
+
+    .. attribute:: digital_signature
+
+        :type: bool
+
+        This is asserted when the subject public key is used for verifying
+        digital signatures, other than signatures on certificates
+        (``key_cert_sign``) and CRLs (``crl_sign``).
+
+    .. attribute:: content_commitment
+
+        :type: bool
+
+        This is asserted when the subject public key is used for verifying
+        digital signatures, other than signatures on certificates
+        (``key_cert_sign``) and CRLs (``crl_sign``). It is used to provide a
+        non-repudiation service that protects against the signing entity
+        falsely denying some action. In the case of later conflict, a
+        reliable third party may determine the authenticity of the signed
+        data. This was called ``non_repudiation`` in older revisions of the
+        X.509 specification.
+
+    .. attribute:: key_encipherment
+
+        :type: bool
+
+        This is asserted when the subject public key is used for enciphering
+        private or secret keys.
+
+    .. attribute:: data_encipherment
+
+        :type: bool
+
+        This is asserted when the subject public key is used for directly
+        enciphering raw user data without the use of an intermediate symmetric
+        cipher.
+
+    .. attribute:: key_agreement
+
+        :type: bool
+
+        This is asserted when the subject public key is used for key agreement.
+        For example, when a Diffie-Hellman key is to be used for key
+        management, then this bit is set.
+
+    .. attribute:: key_cert_sign
+
+        :type: bool
+
+        This is asserted when the subject public key is used for verifying
+        signatures on public key certificates. If this bit is asserted then
+        ``ca`` must be true in the :class:`BasicConstraints` extension.
+
+    .. attribute:: crl_sign
+
+        :type: bool
+
+        This is asserted when the subject public key is used for verifying
+        signatures on certificate revocation lists.
+
+    .. attribute:: encipher_only
+
+        :type: bool
+
+        The meaning of this bit is undefined in the absence of the
+        ``key_agreement`` bit.  When this bit is asserted and the
+        ``key_agreement`` bit is also set, the subject public key may be
+        used only for enciphering data while performing key agreement.
+
+        :raises ValueError: This is raised if accessed when ``key_agreement``
+            is false.
+
+    .. attribute:: decipher_only
+
+        :type: bool
+
+        The meaning of this bit is undefined in the absence of the
+        ``key_agreement`` bit.  When this bit is asserted and the
+        ``key_agreement`` bit is also set, the subject public key may be
+        used only for deciphering data while performing key agreement.
+
+        :raises ValueError: This is raised if accessed when ``key_agreement``
+            is false.
+
+
 .. class:: BasicConstraints
 
     .. versionadded:: 0.9
@@ -686,6 +779,11 @@ Extension OIDs
 
     Corresponds to the dotted string ``"2.5.29.19"``. The identifier for the
     :class:`BasicConstraints` extension type.
+
+.. data:: OID_KEY_USAGE
+
+    Corresponds to the dotted string ``"2.5.29.15"``. The identifier for the
+    :class:`KeyUsage` extension type.
 
 
 Exceptions
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -460,7 +460,7 @@ X.509 Extensions
 
         :type: bool
 
-        This is asserted when the subject public key is used for verifying
+        This purpose is set to true when the subject public key is used for verifying
         digital signatures, other than signatures on certificates
         (``key_cert_sign``) and CRLs (``crl_sign``).
 
@@ -468,7 +468,7 @@ X.509 Extensions
 
         :type: bool
 
-        This is asserted when the subject public key is used for verifying
+        This purpose is set to true when the subject public key is used for verifying
         digital signatures, other than signatures on certificates
         (``key_cert_sign``) and CRLs (``crl_sign``). It is used to provide a
         non-repudiation service that protects against the signing entity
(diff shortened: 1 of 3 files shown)


src/cryptography/x509.py
```diff
@@ -237,6 +237,57 @@ def __repr__(self):
                 "path_length={0.path_length})>").format(self)
 
 
+class KeyUsage(object):
+    def __init__(self, usages):
```
jmvrbanac: Is passing in a list of bools really the best option here? I wonder if defaulting self._usages to a list filled with False would be the cleaner option here.

docs/x509.rst
```diff
@@ -431,6 +431,99 @@ X.509 Extensions
...
+        used only for enciphering data while performing key agreement.
+
+    .. attribute:: decipher_only
+
+        :type: bool
+
+        The meaning of this bit is undefined in the absence of the
+        ``key_agreement`` bit.  When this bit is asserted and the
+        ``key_agreement`` bit is also set, the subject public key may be
+        used only for deciphering data while performing key agreement.
```
alex: These should both raise exceptions when accessed without `key_agreement`.

docs/x509.rst
```diff
@@ -431,6 +431,99 @@ X.509 Extensions
...
 
+.. class:: KeyUsage
+
+    .. versionadded:: 0.9
+
+    The key usage extension defines the purpose of the key contained in the
+    certificate.  The usage restriction might be employed when a key that could
+    be used for more than one operation is to be restricted. It corresponds to
+    :data:`OID_KEY_USAGE`.
+
```
alex: It seems like we should also have a way to enumerate things?

src/cryptography/x509.py
alex: Gosh, my instinct is towards a bitfield or something. Not a fan of the list of bools with magic indices.

alex: Is there any reason not to just have the constructor take 9 params?

docs/x509.rst
reaperhulk: I was discussing this with @jmvrbanac today. Iteration of the properties is certainly doable but I've struggled to come up with a use case where it would actually be helpful?

src/cryptography/x509.py
reaperhulk: No, it can definitely take 9 params and just make them kwargs. That actually sounds like a better solution than having setters.

reaperhulk: (Of note, the indices of the list match the bitmask index values in RFC 5280, although that's certainly just an implementation detail)

docs/x509.rst
reaperhulk: Done.

reaperhulk: Oop, to be clear, the exception is raised during creation of the KeyUsage object, not on access of the property.

src/cryptography/x509.py
```diff
@@ -237,6 +237,40 @@ def __repr__(self):
...
+        self._key_agreement = key_agreement
+        self._key_cert_sign = key_cert_sign
+        self._crl_sign = crl_sign
+        self._encipher_only = encipher_only
+        self._decipher_only = decipher_only
+
+    digital_signature = utils.read_only_property("_digital_signature")
+    non_repudiation = utils.read_only_property("_non_repudiation")
+    # Later versions of X509 have renamed non_repudiation to
+    # content_commitment
```
alex: Is there a reason we need both names?

src/cryptography/x509.py
```diff
@@ -237,6 +237,40 @@ def __repr__(self):
...
+    content_commitment = utils.read_only_property("_non_repudiation")
+    key_encipherment = utils.read_only_property("_key_encipherment")
+    data_encipherment = utils.read_only_property("_data_encipherment")
+    key_agreement = utils.read_only_property("_key_agreement")
+    key_cert_sign = utils.read_only_property("_key_cert_sign")
+    crl_sign = utils.read_only_property("_crl_sign")
+    encipher_only = utils.read_only_property("_encipher_only")
+    decipher_only = utils.read_only_property("_decipher_only")
```
alex: Is it really correct to return `False` here if `key_agreement` is set to `False`?

src/cryptography/x509.py
reaperhulk: While `content_commitment` is the "new" name, everyone still knows it by the old one so it'd be difficult to not support it.

alex: What's difficult about it? You write a computer program, you try to find
the name, you search the docs and it says "This used to be called blah blah
blah. under the new name. It's not like we have legacy code to support.

On Tue, Mar 31, 2015 at 8:55 PM, Paul Kehrer <notifications@github.com>
wrote:

> In src/cryptography/x509.py
> <https://github.com/pyca/cryptography/pull/1802#discussion_r27537415>:
>
> > +            )
> > +
> > +        self._digital_signature = digital_signature
> > +        self._non_repudiation = non_repudiation
> > +        self._key_encipherment = key_encipherment
> > +        self._data_encipherment = data_encipherment
> > +        self._key_agreement = key_agreement
> > +        self._key_cert_sign = key_cert_sign
> > +        self._crl_sign = crl_sign
> > +        self._encipher_only = encipher_only
> > +        self._decipher_only = decipher_only
> > +
> > +    digital_signature = utils.read_only_property("_digital_signature")
> > +    non_repudiation = utils.read_only_property("_non_repudiation")
> > +    # Later versions of X509 have renamed non_repudiation to
> > +    # content_commitment
>
> While content_commitment is the "new" name, everyone still knows it by
> the old one so it'd be difficult to not support it.
>
> —
> <https://github.com/pyca/cryptography/pull/1802/files#r27537415>.
>



-- 

reaperhulk: Fair enough, I can make it only content commitment.

src/cryptography/x509.py
reaperhulk: Would you prefer it to raise an error if accessed or something? The statement that this value is "undefined" when `key_agreement` is false in the RFC is really annoying...

alex: Yeah, I'd prefer raising an error, we shouldn't be exposing "undefined"
behavior to Python.

On Tue, Mar 31, 2015 at 9:00 PM, Paul Kehrer <notifications@github.com>
wrote:

> In src/cryptography/x509.py
> <https://github.com/pyca/cryptography/pull/1802#discussion_r27537627>:
>
> > +        self._crl_sign = crl_sign
> > +        self._encipher_only = encipher_only
> > +        self._decipher_only = decipher_only
> > +
> > +    digital_signature = utils.read_only_property("_digital_signature")
> > +    non_repudiation = utils.read_only_property("_non_repudiation")
> > +    # Later versions of X509 have renamed non_repudiation to
> > +    # content_commitment
> > +    content_commitment = utils.read_only_property("_non_repudiation")
> > +    key_encipherment = utils.read_only_property("_key_encipherment")
> > +    data_encipherment = utils.read_only_property("_data_encipherment")
> > +    key_agreement = utils.read_only_property("_key_agreement")
> > +    key_cert_sign = utils.read_only_property("_key_cert_sign")
> > +    crl_sign = utils.read_only_property("_crl_sign")
> > +    encipher_only = utils.read_only_property("_encipher_only")
> > +    decipher_only = utils.read_only_property("_decipher_only")
>
> Would you prefer it to raise an error if accessed or something? The
> statement that this value is "undefined" when key_agreement is false in
> the RFC is really annoying...
>
> —
> <https://github.com/pyca/cryptography/pull/1802/files#r27537627>.
>



-- 

src/cryptography/x509.py
```diff
@@ -266,6 +266,54 @@ def __repr__(self):
...
+    @property
+    def decipher_only(self):
+        if not self.key_agreement:
+            raise ValueError(
+                "decipher_only is undefined unless key_agreement is true"
+            )
+        else:
+            return self._decipher_only
+
+
```
alex: Too many blank lines.

reaperhulk: This is fixed.

docs/x509.rst
```diff
@@ -447,6 +447,99 @@ X.509 Extensions
...
+    .. attribute:: digital_signature
+
+        :type: bool
+
+        This is asserted when the subject public key is used for verifying
```
alex: I find the "this is asserted when" language pretty confusing, it feels unnescesarily legalistic.

reaperhulk: This was lifted pretty much straight from the RFC, so not surprising that it doesn't feel good :grinning:

How about "This purpose is set to True when" or some variant of that?

docs/x509.rst
```diff
@@ -447,6 +447,99 @@ X.509 Extensions
...
+        enciphering raw user data without the use of an intermediate symmetric
+        cipher.
+
+    .. attribute:: key_agreement
+
+        :type: bool
+
+        This is asserted when the subject public key is used for key agreement.
+        For example, when a Diffie-Hellman key is to be used for key
+        management, then this bit is set.
```
alex: Probably we don't need to describe it as a "bit"

docs/x509.rst
alex: Much better :-)

On Wed, Apr 1, 2015 at 11:26 PM, Paul Kehrer <notifications@github.com>
wrote:

> In docs/x509.rst
> <https://github.com/pyca/cryptography/pull/1802#discussion_r27630536>:
>
> > @@ -447,6 +447,99 @@ X.509 Extensions
> >
> >          Returns an instance of the extension type corresponding to the OID.
> >
> > +.. class:: KeyUsage
> > +
> > +
> > +    The key usage extension defines the purpose of the key contained in the
> > +    certificate.  The usage restriction might be employed when a key that could
> > +    be used for more than one operation is to be restricted. It corresponds to
> > +    :data:`OID_KEY_USAGE`.
> > +
> > +    .. attribute:: digital_signature
> > +
> > +        :type: bool
> > +
> > +        This is asserted when the subject public key is used for verifying
>
> This was lifted pretty much straight from the RFC, so not surprising that
> it doesn't feel good [image: :grinning:]
>
> How about "This purpose is set to True when" or some variant of that?
>
> —
> <https://github.com/pyca/cryptography/pull/1802/files#r27630536>.
>



-- 

docs/x509.rst
```diff
@@ -447,6 +447,99 @@ X.509 Extensions
...
+
+        This is asserted when the subject public key is used for verifying
+        signatures on certificate revocation lists.
+
+    .. attribute:: encipher_only
+
+        :type: bool
+
+        The meaning of this bit is undefined in the absence of the
+        ``key_agreement`` bit.  When this bit is asserted and the
```
alex: I don't think we need to say "is undefined" anymore.

docs/x509.rst
```diff
@@ -447,6 +447,98 @@ X.509 Extensions
 
         Returns an instance of the extension type corresponding to the OID.
 
+.. class:: KeyUsage
+
+    .. versionadded:: 0.9
+
+    The key usage extension defines the purpose of the key contained in the
+    certificate.  The usage restriction might be employed when a key that could
```
public: `might be`? So theres no requirement to actually enforce these?

reaperhulk: This language is actually lifted directly from the RFC. In practice certificates with key usages that don't match their EKUs are not unusual and as a result I'm unsure if, at least in the presence of an EKU, any software cares about purposes like digitalSignature, etc.

PR #1806, opened 2015-03-31
support the EKU OIDs specified in RFC 5280 section 4.2.1.12

reaperhulk:
These OIDs will be used in the extended key usage extension implementation. They are the set defined here: http://tools.ietf.org/html/rfc5280#section-4.2.1.12

refs #1743


alex:
lgtm

PR #1814, opened 2015-04-03
add repr to KeyUsage

reaperhulk:
needed because we print extensions in the docs, refs #1743


alex:
Something about the `N/A` in the repr makes me feel weird. Maybe I should just get over that, we can always change it later since reprs aren't considered stable.

alex:
Since this is becoming a long running blocker, I'm going to merge, even though I don't love the repr, and file a ticket about rethinking it.

PR #1817, opened 2015-04-03
Fix Issue #1790

schlenk:
This fixes issue #1790.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ src/cryptography/_Cryptography_cffi_*
 *.egg
 .eggs/
 *.py[co]
+setup.cfg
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -11,3 +11,4 @@ src/cryptography/_Cryptography_cffi_*
 *.egg
 .eggs/
 *.py[co]
+setup.cfg
```
Ayrx: Why has this line been added?

schlenk: Ah, oversight. Need it to build on windows with non standard Openssl install (x64), otherwise the compile blows up due to wrong pathes. I'll remove it from this PR.

public:
The PEM RFC specifically says line endings should be `<CR><LF>`. Do we care?

schlenk:
It should probably do what OpenSSL command line tools do. The RFC only really specifies the PEM messages, not the openssl extensions for keys, certificates, crls etc.

reaperhulk:
The parent issue for this has been closed as it was due git line ending translation rather than any bug.

Issue #1701, opened 2015-02-25
Lack of wrapped APIs for certificate verification


schlenk:
The openssl backend lib object provides most pieces needed for basic certificate verification with  `X509_verify_cert()` but lacks just a few APIs to allow checking CRLs too.

* One needs a wrapped `X509_STORE_CTX*`, can be done, as all needed APIs are there.
* One needs a wrapped `X509_STORE*`, can be done (in fact is done in PyOpenSSL, would just need to be moved over to cryptography), some basic APIs are there, but the `X509_STORE_add_crl()` function is sadly missing, and the so CRL checks are not possible.

An alternative would be a wrapper for `X509_STORE_set_lookup_crls_cb` to set the lookup callback instead of directly adding a fixed list of CRLs.

In addition the whole `X509_LOOKUP` API would be great to have too. It would allow hooking up callbacks for loading certificates and CRLs to handle other trust stores than plain files/symlinks in directories.

schlenk:
Just looked again, one could use CRLs, but one critical piece is missing (if one looks at the verify app in openssl), there is no wrapping for `STACK_OF(X509_CRL)` which is consumed by the already wrapped ` X509_STORE_CTX_set0_crls()` but there is no function that creates any of this structures. So the minimum to allow CRLs to be used in verification would be functions to create a `STACK_OF(X509_CRL)` from the already existing `X509_CRL` things.

alex:
If I understand this ticket correctly, it's resolved by adding the bindings in #1819. If I misunderstood the scope of this, leave a comment and we can reopen.

PR #1819, opened 2015-04-03
Add wrappers for STACK_OF(X509_CRL) manipulation

schlenk:
This is useful for PyOpenssl to implement CRL checking
in the new X509StoreContext.verify_certificate() API to add
CRLs to the check via X509_STORE_CTX_set0_crls() which is already wrapped.

Not sure about tests needed, as i didn't see any tests for the other sk_* methods anywhere.

This helps with #1701, so one could write an app similar to `openssl verify` with the `-crlfile` option.


reaperhulk:
add to whitelist

alex:
You don't need to worry about tests for bindings, the CI will make sure they compile ok

reaperhulk:
For bindings we don't require tests since they're just exposing the underlying C function/macro. We do require that the build be warnings clean against our CI though (we build with -Werror on our jenkins builders) so that will be the test.

PR #1823, opened 2015-04-04
add subjectkeyidentifier support

reaperhulk:
When we originally discussed this the plan was to try using just a bytestring to represent the SKI extension. Unfortunately, we failed to consider that the Extension class repr will fail when a value that contains non-ascii characters is present. We could fix this by checking the Extension value to see if it's a bytestring and then hexadecimal encoding it or we could create a small extension class. I chose the latter approach as it gives us better consistency with our other extensions (as well as making documentation simpler since it's just another class to document).

refs #1743

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -170,6 +170,8 @@ class _Certificate(object):
                 )
             elif oid == x509.OID_BASIC_CONSTRAINTS:
                 value = self._build_basic_constraints(ext)
+            elif oid == x509.OID_SUBJECT_KEY_IDENTIFIER:
+                value = self._build_subject_key_identifier(ext)
             elif oid == x509.OID_KEY_USAGE and critical:
                 # TODO: remove this obviously.
                 warnings.warn(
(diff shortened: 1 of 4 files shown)


alex:
This needs a rebase.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -217,6 +219,16 @@ def _build_basic_constraints(self, ext):
...
         return x509.BasicConstraints(ca, path_length)
 
+    def _build_subject_key_identifier(self, ext):
+        asn1_string = self._backend._lib.X509V3_EXT_d2i(ext)
+        assert asn1_string != self._backend._ffi.NULL
+        asn1_string = self._backend._ffi.cast(
+            "ASN1_OCTET_STRING *", asn1_string
+        )
+        return x509.SubjectKeyIdentifier(
+            self._backend._ffi.buffer(asn1_string.data, asn1_string.length)[:]
```
public: `[:]` is going to copy the buffer. Do you not need to free the original data or is that handled by the extension instance?

public:
Sounds sensible.

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: Good question. I'll have to look at the impl.

reaperhulk: Looks like a leak. Fixed.

src/cryptography/x509.py
```diff
@@ -5,6 +5,7 @@
 from __future__ import absolute_import, division, print_function
 
 import abc
+import binascii
```
alex: unused

PR #1824, opened 2015-04-04
support openssl no-comp on linux/os x

reaperhulk:
Let's land the easy part of the fix discussed in #1675 with this PR.


alex:
lgtm, thanks

PR #1833, opened 2015-04-11
add repr for x509 names

reaperhulk:


src/cryptography/x509.py
```diff
@@ -190,6 +190,9 @@ def __iter__(self):
     def __len__(self):
         return len(self._attributes)
 
+    def __repr__(self):
+        return repr(self._attributes)
```
alex: Probably it should be `"<Name({0!r})>".format(self._attributes)` or so?

reaperhulk: haha, yeah I was fixing that as soon as I uploaded it. I should swap it to 0!r though. Will do that momentarily.

tests/test_x509.py
```diff
@@ -607,3 +607,16 @@ def test_ne(self):
         ])
         assert name1 != name2
         assert name1 != object()
+
+    def test_repr(self):
+        name = x509.Name([
```
johnwlockwood: Please add a docstring.

PR #1836, opened 2015-04-12
DH examples from https://tools.ietf.org/rfc/rfc5114.txt

public:
A.{4,5,6,7,8} are omitted as they are ECDH examples.

https://tools.ietf.org/rfc/rfc5114.txt

```
   The test data in this appendix is a protocol-independent subset of
   the test data in [EX80056A].  In the test data for the three modular
   exponentiation groups, we use the following notation:

      xA: The secret key of party A

      yA: The public key of party A

      xB: The secret key of party B

      yB: The public key of party B

      Z: The shared secret that results from the Diffie-Hellman
         computation
```


public:
I'll be using these instead of the NIST vectors in #1721

reaperhulk:
Needs documentation in `test-vectors.rst`

alex:
I think historically we've reformatted IETF vectors?

reaperhulk:
We have if we needed to for our current parsers to handle them. @public do you have to write a new vector loader for this?

public:
I can convert them to something our existing loaders can cope with.

reaperhulk:
Yeah let's do that.

public:
Converted for use with the KASVS DH loader.

vectors/cryptography_vectors/asymmetric/DH/RFC5114.txt
```diff
@@ -0,0 +1,44 @@
+[A.1.  1024-bit MODP Group with 160-bit Prime Order Subgroup]
+P = B10B8F96A080E01DDE92DE5EAE5D54EC52C99FBCFB06A3C69A6A9DCA52D23B616073E28675A23D189838EF1E2EE652C013ECB4AEA906112324975C3CD49B83BFACCBDD7D90C4BD7098488E9C219A73724EFFD6FAE5644738FAA31A4FF55BCCC0A151AF5F0DC8B4BD45BF37DF365C1A65E68CFDA76D4DA708DF1FB2BC2E4A4371
+Q = F518AA8781A8DF278ABA4E7D64B7CB9D49462353
+G = A4D1CBD5C3FD34126765A442EFB99905F8104DD258AC507FD6406CFF14266D31266FEA1E5C41564B777E690F5504F213160217B4B01B886A5E91547F9E2749F4D7FBD7D3B9A92EE1909D0D2263F80A76A6A24C087A091F531DBF0A0169B6A28AD662A4D18E73AFA32D779D5918
```
reaperhulk: This doesn't look right? Shouldn't this be `A4D1CBD5C3FD34126765A442EFB99905F8104DD258AC507FD6406CFF14266D31266FEA1E5C41564B777E690F5504F213160217B4B01B886A5E91547F9E2749F4D7FBD7D3B9A92EE1909D0D2263F80A76A6A24C087A091F531DBF0A0169B6A28AD662A4D18E73AFA32D779D5918D08BC8858F4DCEF97C2A24855E6EEB22B3B2E5`?

public: Just testing you ;)

public: Fixed.

chelseawinfree:
Verified these as true compared to the rfc listed.

PR #1837, opened 2015-04-13
Remove unneeded EC digest length check

reaperhulk:
We wrote truncation support for OpenSSL versions that don't support it automatically. So...I think this is safe. Jenkins/Travis will tell me if I'm right!


alex:
Tests appear to be failing.

reaperhulk:
It would appear we do need that. I'll investigate a bit more tomorrow to understand why past us knew this fact but closing for now.

PR #1842, opened 2015-04-13
Twofactor invalid token

### LINKED ISSUE #1818: InvalidToken exception appears twice
The InvalidToken exception is defined twice: in fernet.py and in exceptions.py. Just following the docs, it's very easy to catch the wrong one, and it's not clear there's any utility in having both of them.

sbuss:
Fixes #1818

This moves `cryptography.exceptions.InvalidToken` to `cryptography.hazmat.primitives.twofactor.InvalidToken`, since that exception is intended for (and only used by) the twofactor module. The previous location now uses `utils.deprecated` for backwards compatibility.

@reaperhulk wrote:

> `The cryptography.exceptions.InvalidToken` is meant for HOTP/TOTP, but since it's the same class name as `cryptography.fernet.InvalidToken` this is super confusing [...] maybe we need to move this (and consider renaming it?) and deprecate the old location.

I think keeping the name the same is ok since we're now namespacing it properly, though it does remain slightly confusing.

### DIFF
--- a/src/cryptography/exceptions.py
+++ b/src/cryptography/exceptions.py
@@ -6,6 +6,9 @@ from __future__ import absolute_import, division, print_function
 
 from enum import Enum
 
+from cryptography import utils
+from cryptography.hazmat.primitives import twofactor
+
 
 class _Reasons(Enum):
     BACKEND_MISSING_INTERFACE = 0
@@ -53,5 +56,12 @@ class InvalidKey(Exception):
     pass
 
 
-class InvalidToken(Exception):
-    pass
+InvalidToken = utils.deprecated(
+    twofactor.InvalidToken,
+    __name__,
+    (
+        "The InvalidToken exception has moved to the "
+        "cryptography.hazmat.primitives.twofactor module"
+    ),
+    utils.DeprecatedIn08
+)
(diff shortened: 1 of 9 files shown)


src/cryptography/exceptions.py
```diff
@@ -53,5 +56,12 @@ class InvalidKey(Exception):
...
-class InvalidToken(Exception):
-    pass
+InvalidToken = utils.deprecated(
+    twofactor.InvalidToken,
+    __name__,
+    (
+        "The InvalidToken exception has moved to the "
+        "cryptography.hazmat.primitives.twofactor module"
+    ),
+    utils.DeprecatedIn08
```
reaperhulk: This should be DeprecatedIn09 (which means you'll need to add that to the utils module with value `PendingDeprecationWarning`)

reaperhulk:
add to whitelist

src/cryptography/exceptions.py
sbuss: Thanks, I was unsure about this. Fixed.

alex:
:-( I think you'll need to merge/rebase master, I think travis got into a weird place

Ayrx:
LGTM. :+1:

src/cryptography/utils.py
```diff
@@ -11,6 +11,7 @@
 
 
 DeprecatedIn08 = DeprecationWarning
+DeprecatedIn09 = DeprecationWarning
```
reaperhulk: This should be `PendingDeprecationWarning`

reaperhulk:
This also needs a changelog entry. I'm going to steal this branch and add the fixes.

sbuss:
Closing, superseded by #1848 - thanks for the additions @reaperhulk.

Issue #1838, opened 2015-04-13
decode_rfc6979_signature fails with some input


alex:
Here's a failing test:

```diff
diff --git a/tests/hazmat/primitives/test_asym_utils.py b/tests/hazmat/primitives/test_asym_utils.py
index bf55bad..679464a 100644
--- a/tests/hazmat/primitives/test_asym_utils.py
+++ b/tests/hazmat/primitives/test_asym_utils.py
@@ -63,3 +63,10 @@ def test_decode_rfc6979_invalid_asn1():
         # This byte sequence has an invalid ASN.1 sequence length as well as
         # an invalid integer length for the second integer.
         decode_rfc6979_signature(b"0\x07\x02\x01\x01\x02\x02\x01")
+
+    with pytest.raises(ValueError):
+        decode_rfc6979_signature(
+            b" ,\x02\x14\x1b\x92\xd2_!\x81>[F\xfe\xf97\x08~\x9de\x85A\xa3\xeb"
+            b"\x02\x14\x08-\xd8\xb9\xbf\xb3M\xc6)rU\xc3+\xa6\xbc\x8c\x80V\x06"
+            b"\xe5"
+        )
```

I can't tell if this is our bug, or a pyasn.1 bug.

Discovered using afl-fuzz.

alex:
More minimal test case:

```diff
diff --git a/tests/hazmat/primitives/test_asym_utils.py b/tests/hazmat/primitives/test_asym_utils.py
index bf55bad..c111708 100644
--- a/tests/hazmat/primitives/test_asym_utils.py
+++ b/tests/hazmat/primitives/test_asym_utils.py
@@ -63,3 +63,6 @@ def test_decode_rfc6979_invalid_asn1():
         # This byte sequence has an invalid ASN.1 sequence length as well as
         # an invalid integer length for the second integer.
         decode_rfc6979_signature(b"0\x07\x02\x01\x01\x02\x02\x01")
+
+    with pytest.raises(ValueError):
+        decode_rfc6979_signature(b"\x00\x81\x00")
```

alex:
Another bug, this one is definitely pyasn1:

```
'0,\x02\x14\x1b\x92\xd2_!\x81>[F\xfe\xf97\x08~\x9de\x85A\xa3\xeb\x02\x0f\x08-\xd8\xb9\xbf\xb3M\xc6)rU\xc3+\xa6\xe8\x03\x80\x02\x00\xe5'
```

is the input

[...]

PR #1843, opened 2015-04-13
Work around pyasn1's willingness to return endOfOctets in DER parsing

geofft:
See #1838 for discussion.

### DIFF
--- a/src/cryptography/hazmat/primitives/asymmetric/utils.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/utils.py
@@ -4,6 +4,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+from pyasn1.codec.ber import eoo
 from pyasn1.codec.der import decoder, encoder
 from pyasn1.error import PyAsn1Error
 from pyasn1.type import namedtype, univ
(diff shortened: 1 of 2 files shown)


reaperhulk:
add to whitelist

src/cryptography/hazmat/primitives/asymmetric/utils.py
```diff
@@ -28,6 +29,9 @@ def decode_rfc6979_signature(signature):
         raise ValueError(
             "The signature contains bytes after the end of the ASN.1 sequence."
         )
+    if data == eoo.endOfOctets:
```
alex: Can you add a comment about the fact that this occurs due to a PyASN1 bug, and can be removed in a hypothetical future where eoo is handled correctly?

geofft: Done. It's harmless, so unless you'd prefer to include text about removing it, it seems more robust to keep it indefinitely. If you want to add something about removing it, I can add a comment after I file a bug (so we can link to the bug).

alex: Hmm? I see you force pushed, but the contents looks like the same.

Issue #1821, opened 2015-04-03
AES-GCM Exception Encrypting Large Files


vel0city:
Using AES-GCM, I encounter an exception instead of an error message for the 2^39 - 256 byte limit.

    File "aes-gcm-test.py", line 25, in main
        encryptor.update(plaintext)
    File "...\hazmat\primitives\ciphers\base.py", line 159, in update
        return self._ctx.update(data)
    File "...\hazmat\backends\openssl\ciphers.py", line 118, in update
        assert res != 0
    AssertionError

I've been able to make this exception happen with the following code:

```
import os
from cryptography.hazmat.backends import default_backend
from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes

def main():
    # Use AES-GCM
    encryptor = Cipher(algorithms.AES(os.urandom(16)), modes.GCM(os.urandom(12)), backend=default_backend()).encryptor()

    # 10MB of data
    plaintext = os.urandom(10000000)
    i = 0
    try:
        while True:
            encryptor.update(plaintext)
            i += 1
    except:
        print('{0}0MB encrypted before exception.'.format(i))
        raise

if __name__ == '__main__':
    main()
```

alex:
a) If you're on OS X remember to make sure to use the OpenSSL backend to test thsi
b) It takes quite a while to encrypt enough data to hit the limit, not sure what the best testing strategy is.

reaperhulk:
OpenSSL unfortunately doesn't give anything other than a 0 response code (no error on the stack). Looks like we're going to need to track the bytes encrypted. We already do this in parts of the CC backend so not too difficult. I'll put up a PR tomorrow for this.

reaperhulk:
Anybody have an opinion what the exception should be?

alex:
Value error seems on

sbuss:
> Using AES-GCM, I encounter an exception instead of an error message for the 2^39 - 256 byte limit.

Note that according to the [NIST GCM spec](http://csrc.nist.gov/groups/ST/toolkit/BCM/documents/proposedmodes/gcm/gcm-spec.pdf) this is actually `2^39 - 256` *bits*, not bytes:

> A plaintext P, which can have any number of bits between 0 and 2^39 − 256. 

Also the spec limits AAD to 2^64 bits, so we need to track that as well:

> Additional authenticated data (AAD), which is denoted as A. This data is authenticated, but
not encrypted, and can have any number of bits between 0 and 2^64.

[...]

PR #1844, opened 2015-04-13
[WIP] Add GcmSizeValidator and a few tests.

sbuss:
Please advise on my design here. I think this is super weird and kind of ugly. I'm out of my normal skillset with the object composition used in `cryptography`, so I'm sure I made some problems.

I'd also like a tip on testing the `update_and_validate_plaintext` and `update_aad`, since they can only trigger the `ValueError` when given extremely large bytestrings as input.

CC @reaperhulk & thanks for the help with this today at the PyCon2015 sprint.

Potential solution for #1821

### DIFF
--- a/src/cryptography/hazmat/backends/commoncrypto/ciphers.py
+++ b/src/cryptography/hazmat/backends/commoncrypto/ciphers.py
@@ -8,7 +8,7 @@ from cryptography import utils
 from cryptography.exceptions import (
     InvalidTag, UnsupportedAlgorithm, _Reasons
 )
-from cryptography.hazmat.backends.utils import GcmSizeValidator
+from cryptography.hazmat.backends.utils import _GCMSizeValidator
 from cryptography.hazmat.primitives import ciphers, constant_time
 from cryptography.hazmat.primitives.ciphers import modes
 from cryptography.hazmat.primitives.ciphers.modes import (
@@ -112,7 +112,7 @@ class _GCMCipherContext(object):
         self._mode = mode
         self._operation = operation
         self._tag = None
-        self._gcm_size_validator = GcmSizeValidator()
+        self._gcm_size_validator = _GCMSizeValidator()
 
         registry = self._backend._cipher_registry
         try:
--- a/src/cryptography/hazmat/backends/commoncrypto/ciphers.py
+++ b/src/cryptography/hazmat/backends/commoncrypto/ciphers.py
@@ -8,6 +8,7 @@ from cryptography import utils
 from cryptography.exceptions import (
     InvalidTag, UnsupportedAlgorithm, _Reasons
 )
+from cryptography.hazmat.backends.utils import GcmSizeValidator
 from cryptography.hazmat.primitives import ciphers, constant_time
 from cryptography.hazmat.primitives.ciphers import modes
 from cryptography.hazmat.primitives.ciphers.modes import (
@@ -111,6 +112,7 @@ class _GCMCipherContext(object):
         self._mode = mode
         self._operation = operation
         self._tag = None
+        self._gcm_size_validator = GcmSizeValidator()
 
         registry = self._backend._cipher_registry
         try:
@@ -151,6 +153,7 @@ class _GCMCipherContext(object):
         self.authenticate_additional_data(b"")
 
     def update(self, data):
+        self._gcm_size_validator.update_and_validate_plaintext(data)
         buf = self._backend._ffi.new("unsigned char[]", len(data))
         args = (self._ctx[0], data, len(data), buf)
         if self._operation == self._backend._lib.kCCEncrypt:
@@ -185,6 +188,7 @@ class _GCMCipherContext(object):
         return b""
 
     def authenticate_additional_data(self, data):
+        self._gcm_size_validator.validate_aad(data)
         res = self._backend._lib.CCCryptorGCMAddAAD(
             self._ctx[0], data, len(data)
         )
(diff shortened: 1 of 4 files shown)


src/cryptography/hazmat/backends/utils.py
```diff
@@ -0,0 +1,38 @@
...
+        if (self._plaintext_len < 0 or
+                self._plaintext_len > self._PLAINTEXT_BIT_LIMIT):
+            raise ValueError("Exceeded GCM mode plaintext bit limit.")
+
+    def validate_aad(self, data):
+        return self.validate_aad_len(len(data) * 8)
+
+    def validate_aad_len(self, datalen):
+        if datalen < 0 or datalen > self._AAD_BIT_LIMIT:
+            raise ValueError("Exceeded GCM mode AAD bit limit.")
```
alex: Why do the plaintext and aad have different naming and implementation strategies?

src/cryptography/hazmat/backends/commoncrypto/ciphers.py
```diff
@@ -8,6 +8,7 @@
 from cryptography.exceptions import (
     InvalidTag, UnsupportedAlgorithm, _Reasons
 )
+from cryptography.hazmat.backends.utils import GcmSizeValidator
```
reaperhulk: Our style guide wants initialisms capitalized. e.g. `GCMSizeValidator`

src/cryptography/hazmat/backends/utils.py
```diff
@@ -0,0 +1,38 @@
...
+        self._plaintext_len += len(data) * 8
+        self.validate_plaintext_len()
+
+    def validate_plaintext_len(self):
+        # Technically a size of 0 is invalid, but there are a couple hacks
+        # around some things that call update(b"") on a newly initialized
+        # CipherContext, and we can't interfere with that.
+        # Also size 0 plaintext shouldn't produce any output so I don't
+        # think we need to worry about it.
+        if (self._plaintext_len < 0 or
```
alex: How would `self._palintext_len` ever get to be negative?

src/cryptography/hazmat/backends/utils.py
```diff
@@ -0,0 +1,38 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+from __future__ import absolute_import, division, print_function
+
+
+class GcmSizeValidator(object):
```
alex: The name should probably be `GCMSizeValidator`

alex:
I'm not sure how I feel about this API, but I left some comments for now.

Ayrx:
If this `GCMSizeValidator` class is going to be part of the public interface there should be documentation, otherwise it should probably be named `_GCMSizeValidator`.

src/cryptography/hazmat/backends/commoncrypto/ciphers.py
sbuss: Opps, I read the style guidelines but flipped a bit in my head for "A class should be named HTTPClient, not HttpClient."

src/cryptography/hazmat/backends/utils.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    GCM may only encrypt up to 2**39 - 256 bits of plaintext, so we
+    must track the number of bytes we see.
+    """
+    _PLAINTEXT_BIT_LIMIT = 2 ** 39 - 256
+    _AAD_BIT_LIMIT = 2 ** 64 - 1
```
sbuss: @alex Does `cryptography` have a preferred position on bits vs bytes? This is a max bits value, and as such I multiply `len` by 8 below. I think this may have been a bad choice because it makes it easy to improperly check lengths, but I wanted to keep the values as close to the NIST spec as possible. (I'd prefer to change it)

src/cryptography/hazmat/backends/utils.py
sbuss: It couldn't, assuming people are not malicious. If you think this is too defensive I can just check the max value.

src/cryptography/hazmat/backends/utils.py
sbuss: Yeah I don't like where this ended up.

My first instinct was to do something like

```python
class GCMSizeValidatorMixin(object):
    def __init__(self):
        self._plaintext_size = 0
        self._aad_size = 0

    def update(self, data):
        self._plaintext_size += len(data)
        self.validate_plaintext_len()
        return super(GCMSizeValidatorMixin, self).update(data)

    def authenticate_additional_data(self, data):
        self._aad_size = len(data)
        self.validate_aad_len()
        return super(GCMSizeValidatorMixin, self).authenticate_additional_data(data)

    def validate_plaintext_len(self):
        ...

    def validate_aad_len(self):
        ...
```

which I think is much cleaner, but does not fit the model `cryptography` is using.

I tried to map this thinking to `cryptography` but obviously it ended up messy and inconsistent. I'll attempt another design and ping you when it's up.

src/cryptography/hazmat/backends/utils.py
reaperhulk: I'd prefer these numbers to be bytes. The AAD bit limit isn't 0 mod 8 so it can't be perfectly represented without bits, but our API requires data be passed as bytes so it'd be reasonable to say it has to be > ((2 ** 64) / 8))

alex:
Fixed by #2093.

Issue #1809, opened 2015-04-01
Drop support for Python 3.2


dstufft:
In #1783 @reaperhulk made a commit where he had to use ``six.u("")`` instead of ``u""`` because Python 3.2 does not support the ``u`` prefix on strings. We recently had a discussion about dropping Python 2.6 because it was no longer supported by Python upstream. However I think a more pertinent platform to drop would be Python 3.2.

Unlike 2.6, 3.2 is still technically supported by Python upstream (for just under another year, until February 2016) however as of the last number I looked at Python 3.2 has almost zero usage both across all of PyPI and specifically for cryptography. My numbers for this are months old at this point, however I would assume that with the release of Python 3.4 that the 3.2 numbers would get smaller and smaller and not grow. Also it appears that in at least one case (#1793) supporting Python 3.2 is forcing the use of less than optimal patterns (such as calling a function to do unicode instead of using the ``u`` prefix).

Here's the graphs I happen to have (2014):

#### All of PyPI

![stacked-py-pct](https://cloud.githubusercontent.com/assets/145979/6934848/a48b26fa-d809-11e4-9203-29ad347e140a.png)

#### Cryptography

![stacked-py-pct](https://cloud.githubusercontent.com/assets/145979/6934852/b7713f3e-d809-11e4-85b4-0ae648c5e515.png)

*Note: These graphs filter out any version that has less than some average number of downloads, so that accounts for any missing versions.*

reaperhulk:
If we choose to use the IDNA 2008 library https://github.com/kjd/idna we're also going to have to patch it for Python 3.2 support. I've actually done this here: https://github.com/reaperhulk/idna/commit/af1d53fefbe4d2252d38b54c7e68a7b60b4c9cf1, but as you can see it's not the prettiest modification to make to the codebase.

public:
I am colour blind but I assume that 3.2 is the line just about 2.7 in the first graph?

dstufft:
Ah, yes sorry my graphs aren't particularly color blind friendly :(

Yes, 3.2 is that line, it's a very tiny sliver of a line that you can barely see.

reaperhulk:
3.2 ships as the Python 3 in Ubuntu 12.04 but anybody who cares can get 3.4 from deadsnakes. I think I'm in favor of officially deprecating 3.2 support (and probably dropping it soon thereafter as we start needing the `idna` library for IDNA 2008)

Lukasa:
[...]

PR #1846, opened 2015-04-14
Drop support for Python 3.2

reaperhulk:
refs #1809

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,6 @@ matrix:
           env: TOXENV=py26
         - python: 2.7
           env: TOXENV=py27
-        - python: 3.2
-          env: TOXENV=py32
         - python: 3.3
           env: TOXENV=py33
         - python: 3.4
@@ -27,14 +25,6 @@ matrix:
                       - libssl-dev/lucid
         - python: 2.7
           env: TOXENV=py27 OPENSSL=0.9.8
-          addons:
-              apt:
-                  sources:
-                      - lucid
-                  packages:
-                      - libssl-dev/lucid
-        - python: 3.2
-          env: TOXENV=py32 OPENSSL=0.9.8
           addons:
               apt:
                   sources:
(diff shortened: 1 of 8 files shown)


hynek:
:+1:

alex:
`setup.py` also needs to be updated

alex:
Do we really want to do this with no deprecation period?

dstufft:
I don't really think one is worth it, practically speaking zero people were using cryptography from Python 3.2 a few months ago and I have serious doubts that it saw a big rise in 3.2 users since that time. Is a deprecation warning for something that is actively holding back progress that (almost?) nobody is even going to encounter a useful deprecation warning?

If you want more numbers before merging this, I am redoing all of my graphs and I can generate cryptography specific ones again.

alex:
If you could show the updated numbers, demonstrating that it's really
negligble, that'd be great.

On Tue, Apr 14, 2015 at 9:20 PM, Donald Stufft <notifications@github.com>
wrote:

> I don't really think one is worth it, practically speaking zero people
> were using cryptography from Python 3.2 a few months ago and I have serious
> doubts that it saw a big rise in 3.2 users since that time. Is a
> deprecation warning for something that is actively holding back progress
> that (almost?) nobody is even going to encounter a useful deprecation
> warning?
>
> If you want more numbers before merging this, I am redoing all of my
> graphs and I can generate cryptography specific ones again.
>
> —
> <https://github.com/pyca/cryptography/pull/1846#issuecomment-93133705>.
>



-- 

dstufft:
Sure!

dstufft:
I have graphs now, these are specifically for cryptography so it should have relevant data. Both graphs represent the same data, the second one just zooms in on only Python 3.

stacked-py-pct

stacked-py3-pct

alex:
Ok, I'm on board. Thanks for putting this together @dstufft !

alex:
@reaperhulk Can you please also remove the relevant code from `install.sh`?

alex:
Also, README, installation.rst, and docs/development/getting-started.rst

reaperhulk:
retest this please

PR #1847, opened 2015-04-14
[WIP] Aes gcm bit limit validator

sbuss:
@alex CC @reaperhulk This is an alternative approach to #1844, which is maybe a bit better designed and more generic.

If you think this approach is better then I will loop back with some documentation and more tests.

Potential solution for #1821

### DIFF
--- a/src/cryptography/hazmat/backends/commoncrypto/ciphers.py
+++ b/src/cryptography/hazmat/backends/commoncrypto/ciphers.py
@@ -113,11 +113,11 @@ class _GCMCipherContext(object):
         self._operation = operation
         self._tag = None
         self._plaintext_size_validator = _SizeValidator(
-            max_bits=mode.plaintext_bit_limit,
+            max_bytes=mode.plaintext_byte_limit,
             label="%s mode plaintext" % mode.name
         )
         self._aad_size_validator = _SizeValidator(
-            max_bits=mode.aad_bit_limit,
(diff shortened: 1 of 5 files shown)


src/cryptography/hazmat/backends/commoncrypto/ciphers.py
```diff
@@ -151,6 +160,8 @@ def __init__(self, backend, cipher, mode, operation):
         self.authenticate_additional_data(b"")
 
     def update(self, data):
+        self._plaintext_size_validator.update(data)
+        self._plaintext_size_validator.validate()
```
public: If we always call `validate` right after `update` why do we need that part of the API at all?

sbuss: I didn't think it was right for `validate` to increment a counter, such that calling `validate(data)` was not idempotent. At the same time, just calling `update` doesn't imply a validation occurs. `update_and_validate` is an acceptable compromise, though.

sbuss:
@reaperhulk I like this approach more than #1844, but I definitely understand not wanting to add a more general solution like this when the GCM mode is the only thing which currently requires these kinds of validations. If you think other modes might require similar validations in the future I think we should go with this one.

I'll add more tests and docs to whichever you think is better.

public:
retest this please

alex:
Fixed by #2093.

PR #1849, opened 2015-04-15
LibreSSL added ALPN support in 2.1.3

Sp1l:
This fixes build issues on FreeBSD

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/ssl.py
+++ b/src/cryptography/hazmat/bindings/openssl/ssl.py
@@ -526,7 +526,7 @@ static const long Cryptography_HAS_NEXTPROTONEG = 1;
 #endif
 
 /* ALPN was added in OpenSSL 1.0.2. */
-#if OPENSSL_VERSION_NUMBER < 0x10002001L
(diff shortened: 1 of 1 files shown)


alex:
ok to test

reaperhulk:
@Sp1l could you give a bit more detail on this? Why does this fix a problem? LibreSSL declares both OPENSSL_VERSION_NUMBER and LIBRESSL_VERSION_NUMBER as 0x20000000L, so what's the scenario where OPENSSL_VERSION_NUMBER is < 0x10002001L but LIBRESSL_VERSION_NUMBER is defined?

Sp1l:
@reaperhulk The maintainer of the port decided to modify the OPENSSL_VERSION_NUMBER to 0x1000107fL in the port to make more ports compile without errors https://svnweb.freebsd.org/ports/head/security/libressl/files/patch-include-openssl-opensslv.h?revision=381701&view=markup
This obviously comes at the cost of having to maintain checks for features that were added after 1.0.1g.

reaperhulk:
I'm extremely conflicted here. Having a distribution make a change like this is a very, very bad thing to do and I don't want to encourage it (even though this is almost certainly what LibreSSL *should* have done). I guess we'll merge this, but since we don't directly test it this will continue to be problematic for the FreeBSD port.

PR #1854, opened 2015-04-20
IDNA and edge case SAN test vectors

reaperhulk:
refs #1743


chelseawinfree:
Awesome! verified the vectors using openssl and an IDNA decoder to learn a bit of Russian. :+1:

PR #1855, opened 2015-04-20
Support Subject Alternative Name in the OpenSSL backend

reaperhulk:
This PR adds [idna](https://github.com/kjd/idna) as a dependency.

One item to discuss here is whether we want to actually represent UniformResourceIdentifiers as strings or if we should document the value as a ParseResult from urlparse.

requires #1854 
refs #1743

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -15,6 +15,8 @@ from __future__ import absolute_import, division, print_function
 
 import datetime
 
+import idna
+
 from cryptography import utils, x509
 from cryptography.exceptions import UnsupportedAlgorithm
 from cryptography.hazmat.primitives import hashes
@@ -55,6 +57,14 @@ def _build_x509_name(backend, x509_name):
         )
 
     return x509.Name(attributes)
+
+
+def _build_general_name(backend, gn):
+    if gn.type == backend._lib.GEN_DNS:
+        data = backend._ffi.buffer(
+            gn.d.dNSName.data, gn.d.dNSName.length
+        )[:].decode("ascii")
+        return x509.DNSName(idna.decode(data))
 
 
 @utils.register_interface(x509.Certificate)
@@ -173,6 +183,8 @@ class _Certificate(object):
                 value = self._build_subject_key_identifier(ext)
             elif oid == x509.OID_KEY_USAGE:
                 value = self._build_key_usage(ext)
+            elif oid == x509.OID_SUBJECT_ALTERNATIVE_NAME:
+                value = self._build_subject_alt_name(ext)
             elif critical:
                 raise x509.UnsupportedExtension(
                     "{0} is not currently supported".format(oid), oid
@@ -253,6 +265,24 @@ class _Certificate(object):
             encipher_only,
             decipher_only
         )
+
+    def _build_subject_alt_name(self, ext):
+        gns = self._backend._ffi.cast(
+            "GENERAL_NAMES *", self._backend._lib.X509V3_EXT_d2i(ext)
+        )
+        assert gns != self._backend._ffi.NULL
+        gns = self._backend._ffi.gc(gns, self._backend._lib.GENERAL_NAMES_free)
+        num = self._backend._lib.sk_GENERAL_NAME_num(gns)
+        general_names = []
+
+        for i in range(0, num):
+            gn = self._backend._lib.sk_GENERAL_NAME_value(gns, i)
+            assert gn != self._backend._ffi.NULL
+            value = _build_general_name(self._backend, gn)
+
+            general_names.append(value)
+
+        return x509.SubjectAlternativeName(general_names)
 
 
 @utils.register_interface(x509.CertificateSigningRequest)
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -61,9 +61,7 @@ def _build_x509_name(backend, x509_name):
 
 def _build_general_name(backend, gn):
     if gn.type == backend._lib.GEN_DNS:
-        data = backend._ffi.buffer(
-            gn.d.dNSName.data, gn.d.dNSName.length
-        )[:].decode("ascii")
+        data = backend._ffi.buffer(gn.d.dNSName.data, gn.d.dNSName.length)[:]
         return x509.DNSName(idna.decode(data))
 
 
@@ -275,7 +273,7 @@ class _Certificate(object):
         num = self._backend._lib.sk_GENERAL_NAME_num(gns)
         general_names = []
 
-        for i in range(0, num):
+        for i in range(num):
             gn = self._backend._lib.sk_GENERAL_NAME_value(gns, i)
             assert gn != self._backend._ffi.NULL
             value = _build_general_name(self._backend, gn)
--- a/src/cryptography/x509.py
+++ b/src/cryptography/x509.py
@@ -70,6 +70,19 @@ _OID_NAMES = {
 }
 
 
+_GENERAL_NAMES = {
+    0: "otherName",
+    1: "rfc822Name",
+    2: "dNSName",
+    3: "x400Address",
+    4: "directoryName",
+    5: "ediPartyName",
+    6: "uniformResourceIdentifier",
+    7: "iPAddress",
+    8: "registeredID",
+}
+
+
 class Version(Enum):
     v1 = 0
     v3 = 2
@@ -113,6 +126,10 @@ class ExtensionNotFound(Exception):
     def __init__(self, msg, oid):
         super(ExtensionNotFound, self).__init__(msg)
         self.oid = oid
+
+
+class UnsupportedGeneralNameType(Exception):
+    pass
 
 
 class NameAttribute(object):
--- a/src/cryptography/x509.py
+++ b/src/cryptography/x509.py
@@ -70,19 +70,6 @@ _OID_NAMES = {
 }
 
 
-_GENERAL_NAMES = {
-    0: "otherName",
-    1: "rfc822Name",
-    2: "dNSName",
-    3: "x400Address",
-    4: "directoryName",
-    5: "ediPartyName",
-    6: "uniformResourceIdentifier",
-    7: "iPAddress",
-    8: "registeredID",
-}
-
-
 class Version(Enum):
     v1 = 0
     v3 = 2
@@ -126,10 +113,6 @@ class ExtensionNotFound(Exception):
     def __init__(self, msg, oid):
         super(ExtensionNotFound, self).__init__(msg)
         self.oid = oid
-
-
-class UnsupportedGeneralNameType(Exception):
-    pass
 
 
 class NameAttribute(object):
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,7 @@ with open(os.path.join(src_dir, "cryptography", "__about__.py")) as f:
 VECTORS_DEPENDENCY = "cryptography_vectors=={0}".format(about['__version__'])
 
 requirements = [
+    "idna",
     "pyasn1",
     "six>=1.4.1",
     "setuptools"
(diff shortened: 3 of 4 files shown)


src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -57,6 +63,89 @@ def _build_x509_name(backend, x509_name):
     return x509.Name(attributes)
 
 
+def _build_general_name(backend, gn):
+    # dNSName
+    if gn.type == 2:
```
alex: Can we not bind the constants from OpenSSL for this?

reaperhulk: Good question... I don't know where those constants would live in OpenSSL. It seems likely that they do exist though...

alex:
Please rebase at your convenience.

alex:
retest this please

src/cryptography/hazmat/backends/openssl/x509.py
alex: They're in `include/openssl/x509v3.h`, I'll send a PR.

alex: Wait. These are already bound in cryptography:

```

static const int GEN_OTHERNAME;
static const int GEN_EMAIL;
static const int GEN_X400;
static const int GEN_DNS;
static const int GEN_URI;
static const int GEN_DIRNAME;
static const int GEN_EDIPARTY;
static const int GEN_IPADD;
static const int GEN_RID;
```

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -57,6 +63,89 @@ def _build_x509_name(backend, x509_name):
...
+        value = x509.RegisteredID(x509.ObjectIdentifier(oid))
+    else:
+        # otherName, x400Address or ediPartyName
+        raise x509.UnsupportedGeneralNameType(
+            "{0} is not a supported type".format(
+                x509._GENERAL_NAMES[gn.type]
+            )
+        )
+
+    return value
```
alex: We don't actually do anything at the end, I think we should just return the value right away in each of these branches.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -57,6 +63,89 @@ def _build_x509_name(backend, x509_name):
...
+    # uniformResourceIdentifier
+    elif gn.type == 6:
+        data = backend._ffi.buffer(
+            gn.d.uniformResourceIdentifier.data,
+            gn.d.uniformResourceIdentifier.length
+        )[:].decode("ascii")
+        parsed = urllib_parse.urlparse(data)
+        hostname = idna.decode(parsed.hostname)
+        if parsed.port:
+            netloc = hostname + u":" + str(parsed.port)
```
alex: This should be `six.text_type`, not `str` (I don't think it can ever actually fail, since port is always an integer? but it'll be much cleaner)

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -254,6 +345,25 @@ def _build_key_usage(self, ext):
             decipher_only
         )
 
+    def _build_subject_alt_name(self, ext):
+        gns = self._backend._ffi.cast(
+            "GENERAL_NAMES *", self._backend._lib.X509V3_EXT_d2i(ext)
+        )
+        assert gns != self._backend._ffi.NULL
+        # GCing this should be safe. See line 964 of v3_utl.c
```
alex: Is there a better way to reference this, seems basically gaurnteed to get out of date.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -57,6 +63,89 @@ def _build_x509_name(backend, x509_name):
...
+        )
+    # registeredID
+    elif gn.type == 8:
+        oid = _obj2txt(backend, gn.d.registeredID)
+        value = x509.RegisteredID(x509.ObjectIdentifier(oid))
+    else:
+        # otherName, x400Address or ediPartyName
+        raise x509.UnsupportedGeneralNameType(
+            "{0} is not a supported type".format(
+                x509._GENERAL_NAMES[gn.type]
```
alex: Does OpenSSL actually validate that `gn.type` is a real thing?

reaperhulk: Probably not. I could do .get on this to handle that case and just have it return the number then?

alex: I was thinking just drop the text, number only. Simplicity or something.

On Mon, Apr 20, 2015 at 10:13 PM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/1855#discussion_r28745483>:
>
> > +        value = x509.IPAddress(
> > +            ipaddress.ip_address(
> > +                backend._ffi.buffer(
> > +                    gn.d.iPAddress.data, gn.d.iPAddress.length
> > +                )[:]
> > +            )
> > +        )
> > +    # registeredID
> > +    elif gn.type == 8:
> > +        oid = _obj2txt(backend, gn.d.registeredID)
> > +        value = x509.RegisteredID(x509.ObjectIdentifier(oid))
> > +    else:
> > +        # otherName, x400Address or ediPartyName
> > +        raise x509.UnsupportedGeneralNameType(
> > +            "{0} is not a supported type".format(
> > +                x509._GENERAL_NAMES[gn.type]
>
> Probably not. I could do .get on this to handle that case and just have it
> return the number then?
>
> —
> <https://github.com/pyca/cryptography/pull/1855/files#r28745483>.
>



-- 

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: I didn't actually want this one to be in the PR at all. That was a reference to myself. Gone now :)

alex: lol

On Mon, Apr 20, 2015 at 10:21 PM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/1855#discussion_r28745721>:
>
> > @@ -254,6 +345,25 @@ def _build_key_usage(self, ext):
> >              decipher_only
> >          )
> >
> > +    def _build_subject_alt_name(self, ext):
> > +        gns = self._backend._ffi.cast(
> > +            "GENERAL_NAMES *", self._backend._lib.X509V3_EXT_d2i(ext)
> > +        )
> > +        assert gns != self._backend._ffi.NULL
> > +        # GCing this should be safe. See line 964 of v3_utl.c
>
> I didn't actually want this one to be in the PR at all. That was a
> reference to myself. Gone now :)
>
> —
> <https://github.com/pyca/cryptography/pull/1855/files#r28745721>.
>



-- 

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -57,6 +65,82 @@ def _build_x509_name(backend, x509_name):
...
+            return x509.RFC822Name(data)
+    elif gn.type == backend._lib.GEN_IPADD:
+        return x509.IPAddress(
+            ipaddress.ip_address(
+                backend._ffi.buffer(
+                    gn.d.iPAddress.data, gn.d.iPAddress.length
+                )[:]
+            )
+        )
+    # registeredID
```
alex: Can this be dropped?

alex:
Do you think it'd be possible to do the different GeneralName types in different PRs?

reaperhulk:
Yes it should be possible to split them into multiple PRs. Will require a few contortions (temporarily `pass`ing on things we don't handle yet) but I can split it into 3 or 4 PRs probably.

alex:
Not raising an unsupport exception?

On Mon, Apr 20, 2015 at 10:37 PM, Paul Kehrer <notifications@github.com>
wrote:

> Yes it should be possible to split them into multiple PRs. Will require a
> few contortions (temporarily passing on things we don't handle yet) but I
> can split it into 3 or 4 PRs probably.
>
> —
> <https://github.com/pyca/cryptography/pull/1855#issuecomment-94618847>.
>



-- 

reaperhulk:
For the vectors that contain more than one type (and that's most of them) we'd be unable to actually check the types we do support if we raised an exception.

alex:
We'd have to introduce the vectors one at a time in that case, right?

On Mon, Apr 20, 2015 at 10:39 PM, Paul Kehrer <notifications@github.com>
wrote:

> For the vectors that contain more than one type (and that's most of them)
> we'd be unable to actually check the types we do support if we raised an
> exception.
>
> —
> <https://github.com/pyca/cryptography/pull/1855#issuecomment-94619331>.
>



-- 

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -254,6 +266,24 @@ def _build_key_usage(self, ext):
...
+    def _build_subject_alt_name(self, ext):
+        gns = self._backend._ffi.cast(
+            "GENERAL_NAMES *", self._backend._lib.X509V3_EXT_d2i(ext)
+        )
+        assert gns != self._backend._ffi.NULL
+        gns = self._backend._ffi.gc(gns, self._backend._lib.GENERAL_NAMES_free)
+        num = self._backend._lib.sk_GENERAL_NAME_num(gns)
+        general_names = []
+
+        for i in range(0, num):
```
alex: You can drop the `0` here.

src/cryptography/x509.py
```diff
@@ -70,6 +70,19 @@
 }
 
 
+_GENERAL_NAMES = {
```
alex: This is unused, and can be dropped, for now :-) (Yay smaller patches)

src/cryptography/x509.py
```diff
@@ -115,6 +128,10 @@ def __init__(self, msg, oid):
         self.oid = oid
 
 
+class UnsupportedGeneralNameType(Exception):
```
alex: Same here.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -57,6 +59,14 @@ def _build_x509_name(backend, x509_name):
     return x509.Name(attributes)
 
 
+def _build_general_name(backend, gn):
+    if gn.type == backend._lib.GEN_DNS:
+        data = backend._ffi.buffer(
+            gn.d.dNSName.data, gn.d.dNSName.length
+        )[:].decode("ascii")
+        return x509.DNSName(idna.decode(data))
```
alex: Ok, allow me to walk through this:

* We get a bytes object out of OpenSSL
* We ASCII decode that -> the only things stored in X.509 certs are ascii strings which have been IDNA encoded. We now have a unicode text object containing something IDNA encoded
* We decode *again*, which makes sense because idna is a text to text encoding
* In the end we have a unicode string which expresses all parts of the codepoint space that are relevant to IDNs

Confirm?

(Also, can someone who makes poor life decisions, such as @glyph, @exarkun, or @hynek also confirm :-))

hynek: `idna.decode` takes bytes so the double decodes are superfluous.

reaperhulk: Yep, that's right.

alex: So we can drop the `.decode("ascii")` then?

On Tue, Apr 21, 2015 at 9:01 AM, Hynek Schlawack <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/1855#discussion_r28774817>:
>
> > @@ -57,6 +59,14 @@ def _build_x509_name(backend, x509_name):
> >      return x509.Name(attributes)
> >
> >
> > +def _build_general_name(backend, gn):
> > +    if gn.type == backend._lib.GEN_DNS:
> > +        data = backend._ffi.buffer(
> > +            gn.d.dNSName.data, gn.d.dNSName.length
> > +        )[:].decode("ascii")
> > +        return x509.DNSName(idna.decode(data))
>
> idna.decode takes bytes so the double decodes are superfluous.
>
> —
> <https://github.com/pyca/cryptography/pull/1855/files#r28774817>.
>



-- 

reaperhulk: Looks like it. Pushing update momentarily with review changes.

PR #1856, opened 2015-04-21
UnsupportedExtension raises when accessing cert.extensions

reaperhulk:
This is raised when initially parsing extensions, not when calling `get_extension_for_oid`.


alex:
lgtm

PR #1859, opened 2015-04-22
SAN unsupported type

reaperhulk:
Handle unsupported general name types in the openssl backend

refs #1743

### DIFF
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -964,6 +964,11 @@ Exceptions
 
         Returns the OID.
 
+.. class:: UnsupportedGeneralNameType
+
+    This is raised when a certificate contains an unsupported general name
+    type in an extension.
+
 
 .. _`public key infrastructure`: https://en.wikipedia.org/wiki/Public_key_infrastructure
 .. _`TLS`: https://en.wikipedia.org/wiki/Transport_Layer_Security
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -974,9 +974,13 @@ Exceptions
 
     .. attribute:: type
 
-        :type: :term:`text` or int
+        :type: int
+
+        The integer value of the unsupported type. The complete list of
+        types can be found in `RFC 5280 section 4.2.1.6`_.
 
 
 .. _`public key infrastructure`: https://en.wikipedia.org/wiki/Public_key_infrastructure
 .. _`TLS`: https://en.wikipedia.org/wiki/Transport_Layer_Security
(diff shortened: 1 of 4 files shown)


docs/x509.rst
```diff
@@ -964,6 +967,15 @@ Exceptions
...
         Returns the OID.
 
+.. class:: UnsupportedGeneralNameType
+
+    This is raised when a certificate contains an unsupported general name
+    type in an extension.
+
+    .. attribute:: type
+
+        :type: :term:`text` or int
```
alex: Oof, not really a fan of this. This is persuasive to me that we should just use ints everywhere.

alex: Also, just so I'm clear, the integer values are something out of the RFC, not OpenSSL's invention, right?

reaperhulk: The integer value is standard, yes. It's also explicitly restricted in the RFC to the items we know about in the dict I made. A malicious ASN.1 payload could be constructed that didn't use a valid int though.

alex: I think we should assume that x509.py (et all) should be able to cope with
effectively arbitrary data.

On Wed, Apr 22, 2015 at 8:23 AM, Paul Kehrer <notifications@github.com>
wrote:

> In docs/x509.rst
> <https://github.com/pyca/cryptography/pull/1859#discussion_r28866752>:
>
> > @@ -964,6 +967,15 @@ Exceptions
> >
> >          Returns the OID.
> >
> > +.. class:: UnsupportedGeneralNameType
> > +
> > +    This is raised when a certificate contains an unsupported general name
> > +    type in an extension.
> > +
> > +    .. attribute:: type
> > +
> > +        :type: :term:`text` or int
>
> The integer value is standard, yes. It's also explicitly restricted in the
> RFC to the items we know about in the dict I made. A malicious ASN.1
> payload could be constructed that didn't use a valid int though.
>
> —
> <https://github.com/pyca/cryptography/pull/1859/files#r28866752>.
>



-- 

tests/test_x509_ext.py
```diff
@@ -757,3 +757,16 @@ def test_dns_name(self, backend):
...
+            os.path.join(
+                "x509", "custom", "san_other_name.pem"
+            ),
+            x509.load_pem_x509_certificate,
+            backend
+        )
+        with pytest.raises(x509.UnsupportedGeneralNameType) as exc:
+            cert.extensions
+
+        assert exc.value.type == "otherName"
```
alex: This needs to be updated.

PR #1862, opened 2015-04-22
add URI general name support

reaperhulk:
refs #1743

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -17,6 +17,10 @@ import datetime
 
 import idna
 
+import six
+
+from six.moves import urllib_parse
+
 from cryptography import utils, x509
 from cryptography.exceptions import UnsupportedAlgorithm
 from cryptography.hazmat.primitives import hashes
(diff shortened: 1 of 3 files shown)


src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -63,6 +67,30 @@ def _build_general_name(backend, gn):
...
+        # semantically indistinguishable from the original but is not
+        # guaranteed to be exactly the same.
+        uri = urllib_parse.urlunparse((
+            parsed.scheme,
+            netloc,
+            parsed.path,
+            parsed.params,
+            parsed.query,
+            parsed.fragment
+        ))
```
alex: It's valid for URIs to have all this crap? Is it also valid *not* to have stuff? `urllib_parse.urlparse` happily handles input like `"abc"`?

Also CC'ing @hynek and @glyph for a review.

hynek: I seem to remember that urlparse is not very useful for parsing general URIs.  E.g. `sip:1-999-123-4567@voip-provider.example.net` won’t be properly parsed.

reaperhulk: urlparse requires // so if there's a valid URI without // it'll dump everything into the "path". So the failure mode will be that the data passes through without being parsed (and any IDNA won't be decoded).

alex: Is that a problem?

On Thu, Apr 23, 2015 at 8:30 AM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/1862#discussion_r28957538>:
>
> > +        if parsed.port:
> > +            netloc = hostname + u":" + six.text_type(parsed.port)
> > +        else:
> > +            netloc = hostname
> > +
> > +        # Note that building a URL in this fashion means it should be
> > +        # semantically indistinguishable from the original but is not
> > +        # guaranteed to be exactly the same.
> > +        uri = urllib_parse.urlunparse((
> > +            parsed.scheme,
> > +            netloc,
> > +            parsed.path,
> > +            parsed.params,
> > +            parsed.query,
> > +            parsed.fragment
> > +        ))
>
> urlparse requires // so if there's a valid URI without // it'll dump
> everything into the "path". So the failure mode will be that the data
> passes through without being parsed (and any IDNA won't be decoded).
>
> —
> <https://github.com/pyca/cryptography/pull/1862/files#r28957538>.
>



-- 

reaperhulk: It's better to pass it through than fail completely, but it'd be better to be able to consistently decode IDNA. Maybe instead of using urlparse we should just look for "xn--[A-Za-z0-9]+" strings, attempt to IDNA decode them, and then build a new string (and if we can't IDNA decode just pass through).

reaperhulk:
Put up a new approach based on a simple regex. If people like this approach we'll need to add a test vector (or maybe just a regular test) to exercise the UnicodeError path.

alex:
The regex approach scares me more I think.

hynek:
IIRC I do it completely by hand in service_identity.

alex:
I looked at `service_identity`'s handling of URI, and I don't really understand it, but I didn't cross reference with the RFC, I just decided that the whole thing is stupid.

hynek:
It is.  For my needs, it’s okay to just split it up and compare protocol and hostname.  This is a different case tho…

I’ve looked into all URI parsing libs I found but none supported non-// URIs. :(

PR #1865, opened 2015-04-23
some initial OCSP bindings

### LINKED ISSUE #1864: Update openssl bindings to allow server side OCSP stapling
This is the server side for #1863 

I could not find public docs

I found code in OpenSSL sample server

https://github.com/openssl/openssl/blob/master/apps/s_server.c#L2131

```
SSL_CTX_set_tlsext_status_cb(ctx, cert_status_cb);

# callback signature
static int cert_status_cb(SSL *s, void *arg);

# Set response in callback based on OCSP response
rspderlen = i2d_OCSP_RESPONSE(resp, &rspder);
SSL_set_tlsext_status_ocsp_resp(s, rspder, rspderlen);

```

reaperhulk:
Server side bindings. Fixes #1864 

This will probably fail CI initially because these bindings are not guaranteed to be universally available

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/ssl.py
+++ b/src/cryptography/hazmat/bindings/openssl/ssl.py
@@ -20,6 +20,8 @@ static const long Cryptography_HAS_TLSv1_1;
 static const long Cryptography_HAS_TLSv1_2;
 static const long Cryptography_HAS_SECURE_RENEGOTIATION;
 static const long Cryptography_HAS_COMPRESSION;
+static const long Cryptography_HAS_TLSEXT_STATUS_REQ_CB;
+static const long Cryptography_HAS_STATUS_REQ_OCSP_RESP;
 
 /* Internally invented symbol to tell us if SNI is supported */
 static const long Cryptography_HAS_TLSEXT_HOSTNAME;
@@ -315,6 +317,12 @@ void SSL_CTX_set_tlsext_servername_callback(
     SSL_CTX *,
     int (*)(const SSL *, int *, void *));
 
+/* These were added in OpenSSL 0.9.8h, but since version testing in OpenSSL
+   is fraught with peril thanks to OS distributions we check some constants
+   to determine if they are supported or not */
+long SSL_set_tlsext_status_ocsp_resp(SSL *, unsigned char *, int);
+long SSL_CTX_set_tlsext_status_cb(SSL_CTX *, int(*)(SSL *, void *));
+
 long SSL_session_reused(SSL *);
 
 /* The following were macros in 0.9.8e. Once we drop support for RHEL/CentOS 5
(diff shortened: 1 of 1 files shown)


src/cryptography/hazmat/bindings/openssl/ssl.py
```diff
@@ -315,6 +317,11 @@
     SSL_CTX *,
     int (*)(const SSL *, int *, void *));
 
+/* These are not present in older versions of OpenSSL so we test for
```
alex: Can you specify which version of OpenSSL this was added in?

src/cryptography/hazmat/bindings/openssl/ssl.py
```diff
@@ -410,6 +417,22 @@
...
+
+#ifdef SSL_CTRL_SET_TLSEXT_STATUS_REQ_OCSP_RESP
+static const long Cryptography_HAS_STATUS_REQ_OCSP_RESP = 1;
+#else
+static const long Cryptography_HAS_STATUS_REQ_OCSP_RESP = 0;
+long (*SSL_set_tlsext_status_ocsp_resp)(SSL *, unsigned char *, int) = NULL;
+#endif
+
+
+
```
alex: This seems like a not of newlines.

src/cryptography/hazmat/bindings/openssl/ocsp.py
```diff
@@ -0,0 +1,31 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+from __future__ import absolute_import, division, print_function
+
+INCLUDES = """
+/* Windows be crazy */
```
adiroiban: This comment does not tell me nothing new.
Maybe it should be left out or updated to describe the original error which required this change.
Thanks!

But... do we need binding for OCSP ?

OSCP is just the C representation of DER encoded OSCP objects.

I think that we can use asn1  to decode/encode this in python.

What do you think?

adiroiban:
Thanks for working on this.. as I start I only care about TLS extensions callbacks for client/server side.

I don't know if we want OSCP implementation at this momennt.

There are plan to add it in stdlib http://bugs.python.org/issue17123 ...  but it might take a while... but it can also be implemented in Python code outside of stdlib and without using pyopenssl

Thanks again!

alex:
Can we try this without `oscp.py` since that seems to be the blocker?

src/cryptography/hazmat/bindings/openssl/ocsp.py
reaperhulk: The commit msg for this was "despairing windows test (this is not what we should merge)". It was made to test a Windows fix via the CI system (since that is the easiest way to test compilation across all our supported platforms) and was never intended to merge.

We may eventually need OCSP bindings but if you're not interested in them for now I'll pull them out of this PR.

Issue #1109, opened 2014-06-07
Add more curves


public:
Currently we support the NIST curves and nothing else. OpenSSL supports quite a lot of other curves though. Some of the are probably useless but it would probably be good to at least get the other SECG ones wired up.

secp256k1 is probably the only non-NIST non-DJB curve with any widespread usage. It's used by Bitcoin.

shea256:
I would love to see support for `secp256k1`. Any plans on incorporating it?

reaperhulk:
@shea256 do you know of any other sources of test vectors besides what @public linked? Adding secp256k1 is easy, but we'd like to have a good set of reference tests.

shea256:
Here are a couple more sources:
https://github.com/richardkiss/pycoin/blob/master/tests/ecdsa_test.py
https://github.com/vbuterin/pybitcointools/blob/master/test.py

reaperhulk:
@shea256: secp256k1 support has been merged in #1913. It will be in the next release.

shea256:
Awesome! Excited to see it in action.

reaperhulk:
Closing this as we should have new issues when people request new curves.

PR #1867, opened 2015-04-25
Script for generating SECP256K1 vectors

public:
For #1109.

This loads the FIPS 186-3 SigGen.txt for ECDSA and generates an SECP256K1 signature for every hash and message pair.

Currently this verifies the vectors it's generating using our ECDSA API, but that API doesn't actually support SECP256K1 right now so this PR also includes code which enables that.

### DIFF
--- a/docs/development/custom-vectors/secp256k1/generate_secp256k1.py
+++ b/docs/development/custom-vectors/secp256k1/generate_secp256k1.py
@@ -1,0 +1,89 @@
+from __future__ import absolute_import, print_function
+
+import hashlib
+import os
+from binascii import hexlify
+from collections import defaultdict
+
+from ecdsa import SECP256k1, SigningKey
+from ecdsa.util import sigdecode_der, sigencode_der
+
+from cryptography_vectors import open_vector_file
+
+from tests.utils import (
+    load_fips_ecdsa_signing_vectors, load_vectors_from_file
+)
+
+HASHLIB_HASH_TYPES = {
+    "SHA-1": hashlib.sha1,
+    "SHA-224": hashlib.sha224,
+    "SHA-256": hashlib.sha256,
+    "SHA-384": hashlib.sha384,
+    "SHA-512": hashlib.sha512,
+}
+
+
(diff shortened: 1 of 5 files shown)


docs/development/custom-vectors/secp256k1/generate_secp256k1.py
```diff
@@ -0,0 +1,124 @@
...
+dest_path = os.path.join("asymmetric", "ECDSA", "SECP256K1", "SigGen.txt")
+
+fips_vectors = load_vectors_from_file(
+    source_path,
+    load_fips_ecdsa_signing_vectors
+)
+
+write_file(
+    build_vectors(fips_vectors),
+    open_vector_file(dest_path, "w")
```
alex: Please close this file explcitly (with statement)

alex:
This definitely needs docs :-)

public:
Are we OK with the general approach here?

reaperhulk:
Do we need to do this? It looks like there are some vectors available already based on #1109.

public:
Sadly none of the existing vectors include the `r` and `s` values we expose in our API, just the internal `k` value.

reaperhulk:
Okay, makes sense. When we've done generation in the past (e.g. #626) we've added a new section to the docs (https://cryptography.io/en/latest/development/test-vectors/#custom-symmetric-vectors) that inlines the generation script as well as adding the verification code and the vectors themselves.

I'd also prefer that the actual code to enable the curve not land in this PR. That does mean that the generation script won't work immediately, but the code to enable the curve should only be added with the tests.

public:
OK as long as we are fine with the idea I'll do some more work on this :)

public:
Updated. Should I remove the SECP256K1 curve definition I added?

reaperhulk:
@public yeah let's remove it until we land support. Looks like you've got a docs failure to fix still.

reaperhulk:
I've verified these scripts generate as expected (using the python-ecdsa package) and they are verified using OpenSSL. Will merge once tests pass.

PR #1869, opened 2015-04-27
Added support for ssl_get_version

elitest:
This adds support for binding [SSL_get_version](https://www.openssl.org/docs/ssl/SSL_get_version.html) so that I can finish up pyca/pyopenssl#184.  SSL_get_version is supperior to SSL_version(already implemented), in that it returns a string(e.g. TLSv1) instead of hex(e.g. 0x769).

My first PR to cryptography and I'm new to the concepts of ffi, so please be extra skeptical.


alex:
ok to test

alex:
Thanks!

PR #1872, opened 2015-04-28
add support for ipaddress to general name parsing

reaperhulk:
Update: PR only covers ip address now :)

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -14,6 +14,7 @@
 from __future__ import absolute_import, division, print_function
 
 import datetime
+import ipaddress
 
 import idna
 
@@ -94,6 +95,14 @@ def _build_general_name(backend, gn):
     elif gn.type == backend._lib.GEN_RID:
         oid = _obj2txt(backend, gn.d.registeredID)
         return x509.RegisteredID(x509.ObjectIdentifier(oid))
+    elif gn.type == backend._lib.GEN_IPADD:
+        return x509.IPAddress(
+            ipaddress.ip_address(
+                backend._ffi.buffer(
+                    gn.d.iPAddress.data, gn.d.iPAddress.length
+                )[:]
+            )
+        )
     else:
         # otherName, x400Address or ediPartyName
         raise x509.UnsupportedGeneralNameType(
--- a/tests/test_x509_ext.py
+++ b/tests/test_x509_ext.py
@@ -809,3 +809,25 @@ class TestRSASubjectAlternativeNameExtension(object):
             u"lo",
             u"http://someregulardomain.com",
         ]
+
+    def test_ipaddress(self, backend):
+        cert = _load_cert(
+            os.path.join(
+                "x509", "custom", "san_ipaddr.pem"
+            ),
+            x509.load_pem_x509_certificate,
+            backend
+        )
+        ext = cert.extensions.get_extension_for_oid(
+            x509.OID_SUBJECT_ALTERNATIVE_NAME
+        )
+        assert ext is not None
+        assert ext.critical is False
+
+        san = ext.value
+
+        ip = san.get_values_for_type(x509.IPAddress)
+        assert [
+            ipaddress.ip_address(u"127.0.0.1"),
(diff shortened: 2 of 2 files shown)


src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -94,6 +96,39 @@ def _build_general_name(backend, gn):
...
         return x509.RegisteredID(x509.ObjectIdentifier(oid))
+    elif gn.type == backend._lib.GEN_DIRNAME:
+        return x509.DirectoryName(
+            _build_x509_name(backend, gn.d.directoryName)
+        )
+    elif gn.type == backend._lib.GEN_EMAIL:
+        data = backend._ffi.buffer(
+            gn.d.rfc822Name.data, gn.d.rfc822Name.length
+        )[:].decode("ascii")
+        parsed = parseaddr(data)
```
alex: Please use tuple unpacking here, rather than `[1]`.

Also, is it really correct to handle `<>` formatted addresses here? I've only ever seen an RFC822Name with a "normal" email addr.

reaperhulk: I couldn't decide. The proper way to encode an rfc822name is just the bare email address, but there's nothing to stop you from doing something far more horrible. Maybe we should just say we aren't going to IDNA decode anything but the standard representation though?

alex: Maybe we should just parse error on anything else, absent actual usage in
the real world?

On Tue, Apr 28, 2015 at 7:03 PM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/1872#discussion_r29297505>:
>
> > @@ -94,6 +96,39 @@ def _build_general_name(backend, gn):
> >      elif gn.type == backend._lib.GEN_RID:
> >          oid = _obj2txt(backend, gn.d.registeredID)
> >          return x509.RegisteredID(x509.ObjectIdentifier(oid))
> > +    elif gn.type == backend._lib.GEN_DIRNAME:
> > +        return x509.DirectoryName(
> > +            _build_x509_name(backend, gn.d.directoryName)
> > +        )
> > +    elif gn.type == backend._lib.GEN_EMAIL:
> > +        data = backend._ffi.buffer(
> > +            gn.d.rfc822Name.data, gn.d.rfc822Name.length
> > +        )[:].decode("ascii")
> > +        parsed = parseaddr(data)
>
> I couldn't decide. The proper way to encode an rfc822name is just the bare
> email address, but there's nothing to stop you from doing something far
> more horrible. Maybe we should just say we aren't going to IDNA decode
> anything but the standard representation though?
>
> —
> <https://github.com/pyca/cryptography/pull/1872/files#r29297505>.
>



-- 

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -94,6 +96,39 @@ def _build_general_name(backend, gn):
     elif gn.type == backend._lib.GEN_RID:
         oid = _obj2txt(backend, gn.d.registeredID)
         return x509.RegisteredID(x509.ObjectIdentifier(oid))
+    elif gn.type == backend._lib.GEN_DIRNAME:
+        return x509.DirectoryName(
+            _build_x509_name(backend, gn.d.directoryName)
+        )
```
alex: This couldn't go into its own PR?

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -94,6 +96,39 @@ def _build_general_name(backend, gn):
...
+        else:
+            return x509.RFC822Name(data)
+    elif gn.type == backend._lib.GEN_IPADD:
+        return x509.IPAddress(
+            ipaddress.ip_address(
+                backend._ffi.buffer(
+                    gn.d.iPAddress.data, gn.d.iPAddress.length
+                )[:]
+            )
+        )
```
alex: Same here, these are both much easier to review than the RFC bit

alex:
I'd really prefer if we could land "everything but email" first, that stuff is MUCH easier to review :-)

reaperhulk:
retest this please

Issue #1870, opened 2015-04-28
Replace uses of `six.u` with the `u` prefix


alex:
Now that we dropped 3.2, we should do this to clean up the place.

(Note to whoever does this, please only do a few files at a time, reading and reviewing shorter diffs is much easier)

sigmavirus24:
:+1:

eeshangarg:
Hello!

Just thought I would leave a comment that I am working on this and will be submitting a PR as soon as I can. :-)

Thanks!
Eeshan Garg

PR #1873, opened 2015-04-29
Refs #1870 - replace six.u with the u prefix in tests/hazmat/primitives

eeshangarg:
This PR only replaces the uses of six.u with the u prefix in tests/hazmat/primitives. The remaining occurrences of six.u will be addressed in another PR (I am working on it right now, just wanted to keep the diff small so I divided it into two pull requests), which will close the issue. :-)

Thanks,
Eeshan


alex:
ok to test

PR #1874, opened 2015-04-29
Replace the remaining occurrences of six.u with the u prefix

### LINKED ISSUE #1870: Replace uses of `six.u` with the `u` prefix

eeshangarg:
Closes #1870.


alex:
add to whitelist

PR #1876, opened 2015-04-29
Add loader for ECDH test vectors

simo5:
Also fix test vector files, they had a truncated line for the EE set in
both init and resp files.

Signed-off-by: Simo Sorce <simo@redhat.com>

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -21,7 +21,8 @@ from .utils import (
     check_backend_support, load_cryptrec_vectors,
     load_fips_dsa_key_pair_vectors, load_fips_dsa_sig_vectors,
     load_fips_ecdsa_key_pair_vectors, load_fips_ecdsa_signing_vectors,
-    load_hash_vectors, load_kasvs_dh_vectors, load_nist_vectors,
+    load_hash_vectors, load_kasvs_dh_vectors,
+    load_kasvs_ecdh_vectors, load_nist_vectors,
     load_pkcs1_vectors, load_rsa_nist_vectors, load_vectors_from_file,
     raises_unsupported_algorithm, select_backends, skip_if_empty
 )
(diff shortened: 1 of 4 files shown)


alex:
add to whitelist

reaperhulk:
confirmed the vector fix against latest NIST CAVP data.

alex:
Why was our original vector wrong?

On Thu, Apr 30, 2015 at 12:34 PM, Paul Kehrer <notifications@github.com>
wrote:

> confirmed the vector fix against latest NIST CAVP data.
>
> —
> <https://github.com/pyca/cryptography/pull/1876#issuecomment-97871259>.
>



-- 

reaperhulk:
I'm unsure. I confirmed it at the time (see #1264), but it wouldn't surprise me to learn that NIST updated the files to fix this.

public:
It is rather suspicious that both files have the exact same defect.

simo5:
I couldn't find the same file on NIST websites.
I can only find files with no [EA] set anymore and they also have assitional parameters.
Where did you get the original files from ?

reaperhulk:
http://csrc.nist.gov/groups/STM/cavp/documents/keymgmt/KASTestVectorsECC2014.zip should be the original source, but yeah...the EA set appears to be missing now?

public:
The PR that added them was #879 from April 2014 but the timestamps on all the files in that zip are from December 2014 so they definitely regenerated a bunch of stuff.

simo5:
Added new test vestors from NIST that include KDF after raw ECDH vectors.
Changed slightly the vector loader to reflect be compatible with both the newer and older files and load only the fields used in #1882 for testing.
Also changed how fail/pass is recorded so tests can check exactly what should go wrong if they can/care.

reaperhulk:
The new test vectors should be put into a separate PR so we can confirm they match the NIST download and merge them separately. Thanks!

simo5:
ok #1884 it is

public:
Remove the actual vectors from this PR please.

simo5:
On Sun, 2015-05-03 at 03:30 -0700, Alex Stapleton wrote:
> Remove the actual vectors from this PR please.

I am not sure what you mean, you want me to remove the test vectors that
test the function that loads the vector files ?

-- 

reaperhulk:
This LGTM. Any outstanding comments @public?

PR #1880, opened 2015-04-30
Add X509_delete_ext binding needed by Zorp

endreszabo:
[Balabit Zorp](https://github.com/balabit/zorp) application level firewall [relies on the del_extension function](https://github.com/balabit/zorp/blob/master/pylib/Zorp/Keybridge.py#L350) to remove attributes like CRL pathes from the mimicked  certificates when doing man-in-the-middle traffic filtering.


alex:
ok to test

PR #1881, opened 2015-04-30
add support for rfc822name general names

reaperhulk:


src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -107,6 +108,22 @@ def _build_general_name(backend, gn):
...
+        )[:].decode("ascii")
+        name, address = parseaddr(data)
+        if name:
+            raise ValueError("Invalid rfc822name value")
+        else:
+            parts = address.split("@")
+            if len(parts) == 1:
+                return x509.RFC822Name(address)
+            else:
+                assert len(parts) == 2
```
alex: This assert can fail, `parseaddr` happilly handles nonsense non-email addrs.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -107,6 +108,21 @@ def _build_general_name(backend, gn):
...
+        parts = address.split("@")
+        if name or len(parts) > 2:
+            raise ValueError("Invalid rfc822name value")
+        else:
+            if len(parts) == 1:
+                return x509.RFC822Name(address)
+            else:
+                return x509.RFC822Name(
+                    parts[0] + u"@" + idna.decode(parts[1])
+                )
```
alex: This would be better represented as an elif chain, also no handling of the 0 case, and I'm not sure I understand the 1 case, more comments please :-)

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -107,6 +108,26 @@ def _build_general_name(backend, gn):
         return x509.DirectoryName(
             _build_x509_name(backend, gn.d.directoryName)
         )
+    elif gn.type == backend._lib.GEN_EMAIL:
```
public: This function is getting rather large and every branch in the main if immediately returns. Maybe it should be multiple functions?

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -107,6 +108,26 @@ def _build_general_name(backend, gn):
...
+        )[:].decode("ascii")
+        name, address = parseaddr(data)
+        parts = address.split("@")
+        if name or len(parts) > 2:
+            # parseaddr has found a name (e.g. Name <email>) or the split
+            # has found more than 2 parts (which means more than one @ sign)
+            raise ValueError("Invalid rfc822name value")
+        elif len(parts) == 1:
+            # Single label email name. This is valid for local delivery. No
+            # IDNA decoding can be done since there is no domain component.
```
alex: In the event that `data` is an empty string, `address` will be an empty string here, is that correct? I think `or not address` should go into the branch at 117

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -107,6 +108,26 @@ def _build_general_name(backend, gn):
         return x509.DirectoryName(
             _build_x509_name(backend, gn.d.directoryName)
         )
+    elif gn.type == backend._lib.GEN_EMAIL:
+        data = backend._ffi.buffer(
+            gn.d.rfc822Name.data, gn.d.rfc822Name.length
+        )[:].decode("ascii")
+        name, address = parseaddr(data)
+        parts = address.split("@")
```
alex: This should really be a `u"@"`, in practice it works fine this way, but I think it's slightly easier to read when it's explicitly unicode

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: This is the last branch to add and I wouldn't anticipate further changes for now. We can refactor if we have reason to revisit this though.

alex: This seems reasonable to me.

On Sat, May 2, 2015 at 4:15 PM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/1881#discussion_r29550321>:
>
> > @@ -107,6 +108,26 @@ def _build_general_name(backend, gn):
> >          return x509.DirectoryName(
> >              _build_x509_name(backend, gn.d.directoryName)
> >          )
> > +    elif gn.type == backend._lib.GEN_EMAIL:
>
> This is the last branch to add and I wouldn't anticipate further changes
> for now. We can refactor if we have reason to revisit this though.
>
> —
> <https://github.com/pyca/cryptography/pull/1881/files#r29550321>.
>



-- 

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: Done

tests/test_x509_ext.py
```diff
@@ -856,3 +856,91 @@ def test_dirname(self, backend):
...
+
+        san = ext.value
+
+        rfc822_name = san.get_values_for_type(x509.RFC822Name)
+        uri = san.get_values_for_type(x509.UniformResourceIdentifier)
+        dns = san.get_values_for_type(x509.DNSName)
+        ip = san.get_values_for_type(x509.IPAddress)
+        dirname = san.get_values_for_type(x509.DirectoryName)
+        assert [u"user@cryptography.io"] == rfc822_name
+        assert ["https://cryptography.io"] == uri
```
alex: Shouldn't this be a unicode string?

alex:
Besides those two things, this lgtm

PR #1882, opened 2015-04-30
Add Key Exchange Agreement Interface

simo5:
Adds Key Exchange interface
Implements ECDH KEy Exchange mechanism.
Tests ECDH w/o Key Derivation

### DIFF
--- a/docs/hazmat/primitives/key-exchange-agreements.rst
+++ b/docs/hazmat/primitives/key-exchange-agreements.rst
@@ -1,0 +1,23 @@
+.. hazmat::
+
+Key Exchange agreements
+=======================
+
+.. module:: cryptography.hazmat.primitives.asymmetric.key_exchange
+
+Key exchange agreements are cryptographic operations, like Diffie-Hellman
+key exchanges, that allow two parties to use their public-private key pairs
+to establish a shared secret key over an insecure channel. Usually the
+negotiated key is further derived before using it for symmetric operations.
+
+Interfaces
+~~~~~~~~~~
+
+.. class:: KeyExchangeContext
+
+    .. versionadded:: 1.1
+
+    .. method:: agree(public_key)
+
+        :param public_key: The peer public key, the type depends on the
+            crypto system used, for example :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey`
--- a/docs/hazmat/primitives/index.rst
+++ b/docs/hazmat/primitives/index.rst
@@ -15,3 +15,4 @@ Primitives
     constant-time
     interfaces
     twofactor
+    key-exchange-agreements
--- a/src/cryptography/hazmat/primitives/asymmetric/key_exchange.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/key_exchange.py
@@ -1,0 +1,18 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+from __future__ import absolute_import, division, print_function
+
+import abc
+
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class KeyExchangeContext(object):
+    @abc.abstractmethod
+    def agree(self, public_key):
+        """
+        Returns the agreed key material.
+        """


reaperhulk:
retest this please

simo5:
I fixed the code in the docs, all looks good on that front now

simo5:
I intend to add test based on the vectors provided by #1884 and the KDF function provided by #1898 before merging this PR.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1037,6 +1037,18 @@ def load_elliptic_curve_public_numbers(self, numbers):
 
         return _EllipticCurvePublicKey(self, ec_cdata)
 
+    def elliptic_curve_exchange_algorithm_supported(
```
public: Does this need to be separate from the regular `exchange_algorithm_supported` function?

src/cryptography/hazmat/primitives/asymmetric/ec.py
```diff
@@ -245,6 +254,14 @@ def __init__(self, algorithm):
     algorithm = utils.read_only_property("_algorithm")
 
 
+@utils.register_interface(EllipticCurveKeyExchangeAlgorithm)
+class ECDH(object):
```
public: Should this also take the curve?

src/cryptography/hazmat/backends/openssl/ec.py
```diff
@@ -148,6 +151,35 @@ def verify(self):
...
+        z_buf = self._backend._ffi.new("char[]", z_len)
+        peer_key = self._backend._lib.EC_KEY_get0_public_key(pub_key)
+
+        r = self._backend._lib.ECDH_compute_key(z_buf, z_len,
+                                                peer_key, pri_key,
+                                                self._backend._ffi.NULL)
+        assert r > 0
+        z = self._backend._ffi.buffer(z_buf)[:z_len]
+
+        if self._kdf is None:
```
public: For regular DH I was planning on leaving the derviation step out of here but could be convinced to add it back. The main issue is handling of the "None" KDF as it can make it less obvious that a particular use (rather than e.g. all use) of the API is unsafe.

src/cryptography/hazmat/primitives/asymmetric/ec.py
simo5: On Sun, 2015-05-10 at 02:02 -0700, Alex Stapleton wrote:
> > @@ -245,6 +254,14 @@ def __init__(self, algorithm):
> >      algorithm = utils.read_only_property("_algorithm")
> >  
> >  
> > +@utils.register_interface(EllipticCurveKeyExchangeAlgorithm)
> > +class ECDH(object):
> 
> Should this also take the curve?

The curve is implicit in the key used ?
Or should we explicitly provide a curve and then fail if the key has a
different curve ?

Simo.

-- 

src/cryptography/hazmat/backends/openssl/ec.py
simo5: On Sun, 2015-05-10 at 02:23 -0700, Alex Stapleton wrote:
> > +        pri_key = self._private_key._ec_key
> > +        pub_key = public_key._ec_key
> > +
> > +        group = self._backend._lib.EC_KEY_get0_group(pri_key)
> > +        z_len = (self._backend._lib.EC_GROUP_get_degree(group) + 7) // 8
> > +        assert z_len > 0
> > +        z_buf = self._backend._ffi.new("char[]", z_len)
> > +        peer_key = self._backend._lib.EC_KEY_get0_public_key(pub_key)
> > +
> > +        r = self._backend._lib.ECDH_compute_key(z_buf, z_len,
> > +                                                peer_key, pri_key,
> > +                                                self._backend._ffi.NULL)
> > +        assert r > 0
> > +        z = self._backend._ffi.buffer(z_buf)[:z_len]
> > +
> > +        if self._kdf is None:
> 
> For regular DH I was planning on leaving the derviation step out of
> here but could be convinced to add it back. The main issue is handling
> of the "None" KDF as it can make it less obvious that a particular use
> (rather than e.g. all use) of the API is unsafe.

Not sure what you mean here.
If you do not embed the derivation, then you are in the same situation,
as users can simply "forget" to apply an additional step.
Instead if the KDF is a parameter of the function, IMO, users are more
conscious of the fact a KDF is needed and will naturally tend to pass in
a valid KDF function.

Simo.

-- 

public: I think in other places we use a kind of sentinel type to make it more obvious that you are disabling some security. e.g. https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/#cryptography.hazmat.primitives.serialization.NoEncryption

Maybe we should repeat that here? cc @alex @reaperhulk

src/cryptography/hazmat/primitives/asymmetric/ec.py
public: I'm mostly worried about checking that the public key is actually on the same curve as the private key. Malicious public keys are a problem for regular DH, but I'm not sure of their effect on ECDH.

It's probably OK to extract this info from the private key though.

src/cryptography/hazmat/backends/openssl/ec.py
simo5: Ok so should we just add an ECDHnoKDF class in here that you have to use when you do not want KDF ?
I think I like the idea, opinions on the name to be used ?

simo5:
I added a ECDHnoKDF KeyDerivationFunction class, let me know what you think about it.

src/cryptography/hazmat/primitives/asymmetric/ec.py
```diff
@@ -252,6 +264,28 @@ def __init__(self, algorithm):
     algorithm = utils.read_only_property("_algorithm")
 
 
+@utils.register_interface(KeyDerivationFunction)
+class ECDHnoKDF(object):
```
public: Having spoken with @alex I think it might actually be better to just leave this functionality out entirely until we have a clear use-case for it. How do you feel about that @simo5 ? Are you aware of any cases where you really need access to the raw ECDH output?

alex: If someone really needs the raw output they can always write a NullKDF very
easily.

On Sat, May 16, 2015 at 11:38 AM, Alex Stapleton <notifications@github.com>
wrote:

> In src/cryptography/hazmat/primitives/asymmetric/ec.py
> <https://github.com/pyca/cryptography/pull/1882#discussion_r30462024>:
>
> > @@ -252,6 +264,28 @@ def __init__(self, algorithm):
> >      algorithm = utils.read_only_property("_algorithm")
> >
> >
> > +@utils.register_interface(KeyDerivationFunction)
> > +class ECDHnoKDF(object):
>
> Having spoken with @alex <https://github.com/alex> I think it might
> actually be better to just leave this functionality out entirely until we
> have a clear use-case for it. How do you feel about that @simo5
> <https://github.com/simo5> ? Are you aware of any cases where you really
> need access to the raw ECDH output?
>
> —
> <https://github.com/pyca/cryptography/pull/1882/files#r30462024>.
>



-- 

simo5: I think it is needed mostly for tests, I am not sure anyone should really use this w/o KDF, so I am ok leaving it out of the official API and just adding one in tests.

simo5:
As requested by @reaperhulk splitting this PR in 2 parts.
this part is about the Key Exchange Agremment, and has been appropriately renamed.
A new PR will be opened for the actual ECDH implementation

docs/hazmat/primitives/key-exchange-agreements.rst
```diff
@@ -0,0 +1,16 @@
...
+Key Exchange agreements
+=======================
+
+.. module:: cryptography.hazmat.primitives.asymmetric.key_exchange
+
+
+Interfaces
+~~~~~~~~~~
+
+.. class:: KeyExchangeContext
```
reaperhulk: This needs a `.. versionadded:: 1.1`

reaperhulk:
Otherwise this LGTM based on trawling through the last 9 months of conversation about this. Would like a quick confirmation from @public and @alex though.

docs/hazmat/primitives/key-exchange-agreements.rst
```diff
@@ -0,0 +1,18 @@
...
+
+    .. versionadded:: 1.1
+
+    .. method:: agree(public_key)
+
+        :param public_key: The peer public key.
```
alex: What type is this?

This should also describe what the return value is.

docs/hazmat/primitives/key-exchange-agreements.rst
```diff
@@ -0,0 +1,18 @@
+.. hazmat::
+
+Key Exchange agreements
+=======================
+
+.. module:: cryptography.hazmat.primitives.asymmetric.key_exchange
+
```
alex: Can you add a sentence or two of prose saying what is key exchange

simo5: I've added some more meat and pushed a new version.

PR #1888, opened 2015-05-02
Expose SSL_CTX_set_cert_verify_callback

Lukasa:
This is for use with HPKP, as discussed in shazow/urllib3#607.

It's my belief that this change is sufficient to move forward in PyOpenSSL, but I'm not 100% confident of that fact yet. We'll see.

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/ssl.py
+++ b/src/cryptography/hazmat/bindings/openssl/ssl.py
@@ -211,6 +211,9 @@ int SSL_CTX_use_certificate_chain_file(SSL_CTX *, const char *);
 int SSL_CTX_use_PrivateKey(SSL_CTX *, EVP_PKEY *);
 int SSL_CTX_use_PrivateKey_file(SSL_CTX *, const char *, int);
 int SSL_CTX_check_private_key(const SSL_CTX *);
+void SSL_CTX_set_cert_verify_callback(SSL_CTX *,
+                                      int (*)(X509_STORE_CTX *,void *),
+                                      void *);
 
 void SSL_CTX_set_cert_store(SSL_CTX *, X509_STORE *);
 X509_STORE *SSL_CTX_get_cert_store(const SSL_CTX *);


alex:
add to whitelist

alex:
lgtm

tiran:
Uh, you do *really* want to provide your own 509_verify_cert() function? You'd have to roll *all* checks yourself, including but not limited to all X509v3 extensions, notBefore / not After dates, policies, constraints, purpose, CRLs...

https://github.com/openssl/openssl/blob/master/crypto/x509/x509_vfy.c#L183
https://github.com/openssl/openssl/blob/master/ssl/ssl_cert.c#L685

If you just want to do additional checks then please use the verify callback. It gets called for all errors with ok=0 as well as for all certs with ok=1.

Lukasa:
@tiran No, I have every intention of calling into `X509_cert_verify`. I just want to pull some data out before we get there.

tiran:
@Lukasa I still don't understand why you need to set your own application callback. The verification callback should be sufficient enough for the task.

I'm asking because I'm planing to add a verification callback to Python's stdlib although with a slightly different argument list than PyOpenSSL.

Lukasa:
@tiran A read of the OpenSSL code does not make it even remotely clear to me that the verification callback is called in the right places to do that. It *might* be good enough, but I'm not certain of that fact.

Lukasa:
Unrelated, but PyOpenSSL already exposes a verify callback, but it uses a custom API that does not expose the X509 context object, so either way I'll have to add a second callback function to PyOpenSSL's API, which is super sad.

reaperhulk:
:+1:

alex:
(FYI I plan to merge this, and we can litigate the API on teh appropriate PyOpenSSL PR :P)

Lukasa:
Agreed, the binding itself is not really a problem. I'd definitely like more insight on the API side though.

tiran:
@Lukasa The verify_cb has two purposes:

1) It gets called in every error cases (almost). An application can overwrite the decision and ignore the error.
2) It gets called for very cert in the chain https://github.com/openssl/openssl/blob/master/crypto/x509/x509_vfy.c#L1785 after the chain has been build. The allows an application to reject an certificate.

You are right about the missing feature in PyOpenSSL. That's why I like to add a verify callback with a different set of arguments: 

   verify_callback(ctx: X509StoreContext, ok: int) -> int

X509StoreContext has a reference to the SSLSocket object (which refers to SSLContext) and additional attributes for e.g. error code, error string, depths, current cert, issuer cert and the complete chain including the root CA cert.

Lukasa:
@tiran I actively want the X509StoreContext, because I want to try to determine the entire trust chain *including the anchor*, which `get_peer_cert_chain()` does not give me. Unfortunately, your assessment suggests that I actually need to be super careful with this callback because I want to save out:

1. The complete cert chain, and
2. The actual cert chain that was accepted.

tiran:
@Lukasa That's right. During verification you can access the entire chain including the trust root. Outside OpenSSL just exposes the leaf cert and the intermediate chain (if provided by the server). But you could still look up the trust roots in the SSLContext object yourself. My branch for Python 3.5 stdlib exposes this feature.

import ssl

context = ssl.create_default_context()

def get_server_certificate(addr):
    host, port = addr
    with ssl.create_connection(addr) as sock:
        with context.wrap_socket(sock, server_hostname=host) as sslsock:
            return sslsock.get_peer_cert_chain()

chain = get_server_certificate(('www.google.com', 443))
print('Chain certificates:', [s.get_subject_name(True) for s in chain])
print('Root from context:', context.lookup_issuer(chain[-1]).get_subject_name(True))

Output:
Chain certificates: ['/C=US/ST=California/L=Mountain View/O=Google Inc/CN=www.google.com', '/C=US/O=Google Inc/CN=Google Internet Authority G2', '/C=US/O=GeoTrust Inc./CN=GeoTrust Global CA']
Root from context: /C=US/O=Equifax/OU=Equifax Secure Certificate Authority

tiran:
Anyway the verification callback gets to see *every* cert in the whole chain including trust root. We just have to convince @hynek to modify the arguments or create a second kind of callback with a X509StoreContext argument.

Lukasa:
There's an alternative, which is that I can simply replicate your logic for `lookup_issuer` in urllib3 directly, or indeed in cryptography's X509 module (hey @reaperhulk!):

```c
static PyObject *
lookup_issuer(PySSLContext *self, PyObject *args, PyObject *kwds)
{
    char *kwlist[] = {"cert", NULL};
    X509_STORE_CTX *store_ctx;
    X509 *issuer_cert;
    PyObject *subject = NULL;
    PyObject *issuer = NULL;
    int ret;

    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O!:lookup_issuer",
                                     kwlist, &PyX509Cert_Type, &subject)) {
        return NULL;
    }
    if ((store_ctx = X509_STORE_CTX_new()) == NULL) {
        _setSSLError(NULL, 0, __FILE__, __LINE__);
        return NULL;
    }
    if (!X509_STORE_CTX_init(store_ctx,
                             SSL_CTX_get_cert_store(self->ctx),
                             NULL, NULL)) {
        _setX509StoreContextError(store_ctx, __FILE__, __LINE__);
        goto end;
    }

    ret = store_ctx->get_issuer(&issuer_cert,
                                store_ctx,
                                PyX509Cert_Cert(subject));
    if (ret == 1) {
        issuer = (PyObject *)newPyX509Cert(issuer_cert, 0, 0);
    }
    else if (ret == 0) {
        issuer = Py_None;
        Py_INCREF(issuer);
    }
    else {
        _setX509StoreContextError(store_ctx, __FILE__, __LINE__);
    }
  end:
    X509_STORE_CTX_free(store_ctx);
    return issuer;
}
```

Looks like the relevant code in cryptography wouldn't be too tricky (I haven't checked if the relevant bindings exist yet, but it should be easy enough to add them).

reaperhulk:
We're pretty early days for the x509 API for cryptography, but this feature does sound like it belongs there (probably in a hazmat component, which we haven't developed yet).

Lukasa:
So that raises a question, I think: should I go ahead and implement in PyOpenSSL, or possibly directly in urllib3, code that more-or-less exactly matches Christian's code?

PR #1889, opened 2015-05-02
Opt in to travis's new OS X infrastructure which should be faster

alex:


public:
Restarted the travis build.

alex:
@henrikhodne tells me there's a bug that makes using this infra fail sometimes which will be fixed early next week, so we shoudl consider this on hold until then :-)

alex:
So, there's a few problems:

* Python 2.6 and 2.7 (where we use the Pythons included in the system) for some reason don't have a pip? /cc @dstufft 
* Using the system OpenSSL fails with something inside of `test_rsa.py`. I don't remember what caused this, something about a difference between the headers and the shared library on the system? /cc @reaperhulk

reaperhulk:
Yep, the error conditions changed in OpenSSL 0.9.8zc+. Apple updated their OpenSSL but didn't update the headers for the latest upgrade. This was filed upstream with Apple but it's unclear if they plan to actually fix it.

reaperhulk:
We are running get-pip with --user but then running ``pip install --user virtualenv`` after that without adding it to the path. We should probably be doing ``python -m pip install --user virtualenv`` right?

alex:
@reaperhulk good call on the `python -m pip` bit.

What do we do about the test_rsa.py failures? I'm a bit confused about what environment triggers them (e.g. I don't see them locally).

henrikhodne:
By the way, we have fixed the bugs we have found with the infrastructure, so the builds should be stable now. Anything I can do to the build image to help resolve these issues?

We have OpenSSL installed with `brew` as well, but it is not linked in by default.

reaperhulk:
@alex I think they actually fixed this bug as of the latest xcode CLI tools... When I check out /usr/include/openssl/opensslv.h I see 0.9.8zd now, where before it was za (even when openssl was zc).

@henrikhodne Based on the name of this image I assume it's Xcode 6.2. Is there a 6.3.1 image in the works?

henrikhodne:
> @henrikhodne Based on the name of this image I assume it's Xcode 6.2. Is there a 6.3.1 image in the works?

Yep, just use `osx_image: beta-xcode6.3` instead (it’s 6.3.1, not 6.3, despite the name).

alex:
Oh, awesome! Thanks.

On Mon, May 18, 2015 at 6:23 PM, Henrik Hodne <notifications@github.com>
wrote:

> @henrikhodne <https://github.com/henrikhodne> Based on the name of this
> image I assume it's Xcode 6.2. Is there a 6.3.1 image in the works?
>
> Yep, just use osx_image: beta-xcode6.3 instead (it’s 6.3.1, not 6.3,
> despite the name).
>
> —
> <https://github.com/pyca/cryptography/pull/1889#issuecomment-103233272>.
>



-- 

PR #1891, opened 2015-05-03
Extended key usage support for the openssl backend

reaperhulk:
refs #1743


src/cryptography/hazmat/bindings/openssl/x509.py
```diff
@@ -17,6 +17,7 @@
 typedef STACK_OF(X509) Cryptography_STACK_OF_X509;
 typedef STACK_OF(X509_CRL) Cryptography_STACK_OF_X509_CRL;
 typedef STACK_OF(X509_REVOKED) Cryptography_STACK_OF_X509_REVOKED;
+typedef STACK_OF(ASN1_OBJECT) Cryptography_STACK_OF_ASN1_OBJECT;
```
alex: This is defined elsewhere.

reaperhulk: ah it's in X509_vfy damn it I thought I looked for that.

PR #1892, opened 2015-05-03
add authority information access classes

reaperhulk:
refs #1743

### DIFF
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -736,11 +736,18 @@ X.509 Extensions
 
         :type: :class:`ObjectIdentifier`
 
-        Either :data:`OID_OCSP` or :data:`OID_CA_ISSUERS`
+        The access method defines what the ``access_location`` means. It must
+        be either :data:`OID_OCSP` or :data:`OID_CA_ISSUERS`. If it is
+        :data:`OID_OCSP` the access location will be where to obtain OCSP
+        information for the certificate. If it is :data:`OID_CA_ISSUERS` the
+        access location will provide additional information about the issuing
+        certificate.
 
     .. attribute:: access_location
 
         :type: :class:`GeneralName`
+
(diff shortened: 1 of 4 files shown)


docs/x509.rst
```diff
@@ -719,6 +719,29 @@ X.509 Extensions
...
+
+    .. attribute:: access_method
+
+        :type: :class:`ObjectIdentifier`
+
+        Either :data:`OID_OCSP` or :data:`OID_CA_ISSUERS`
+
+    .. attribute:: access_location
+
+        :type: :class:`GeneralName`
```
alex: Description of what this value means, same with method.

src/cryptography/x509.py
```diff
@@ -67,6 +67,8 @@
     "1.3.6.1.5.5.7.1.1": "authorityInfoAccess",
     "1.3.6.1.5.5.7.1.11": "subjectInfoAccess",
     "1.3.6.1.5.5.7.48.1.5": "OCSPNoCheck",
+    "1.3.6.1.5.5.7.48.2": "caIssuers",
+    "1.3.6.1.5.5.7.48.1": "OCSP",
```
alex: These should probably be in the opposite order.

src/cryptography/x509.py
```diff
@@ -394,6 +396,68 @@ def __repr__(self):
...
+        return self._descriptions == other._descriptions
+
+    def __ne__(self, other):
+        return not self == other
+
+
+class AccessDescription(object):
+    def __init__(self, access_method, access_location):
+        if not (access_method == OID_OCSP or access_method == OID_CA_ISSUERS):
+            raise TypeError("access_method must be OID_OCSP or OID_CA_ISSUERS")
```
alex: `ValueError` I guess?

PR #1893, opened 2015-05-03
add example to show how to get DNSNames from SAN extension

reaperhulk:


docs/x509.rst
```diff
@@ -718,6 +754,17 @@ X.509 Extensions
 
         :returns: A list of values extracted from the matched general names.
 
+        .. doctest::
+
+            >>> from cryptography import x509
+            >>> from cryptography.hazmat.backends import default_backend
+            >>> from cryptography.hazmat.primitives import hashes
+            >>> cert = x509.load_pem_x509_certificate(cryptography_cert_pem, default_backend())
+            >>> ext = cert.extensions.get_extension_for_oid(x509.OID_SUBJECT_ALTERNATIVE_NAME)
```
alex: Let's add some comments: `# Get the subjectAltName extension from the certificate`

docs/x509.rst
```diff
@@ -718,6 +754,17 @@ X.509 Extensions
...
+            >>> san = ext.value
```
alex: I don't think this needs it's own local var, it makes stuff look more verbose than it is

docs/x509.rst
```diff
@@ -718,6 +754,17 @@ X.509 Extensions
...
+            >>> san.get_values_for_type(x509.DNSName)
```
alex: Comment: `# Get the dNSName entries from the SAN extension`

PR #1884, opened 2015-05-01
New ECDH vectors

simo5:
Add new Vectors from NIST that include a DKM value to verify the Key Derivation Aftrer ECDH Key Exchange

PR #1898, opened 2015-05-03
Add ConcatKDF from NIST SP 800-56A

simo5:
This KDF function is a prerequisite to implememnt ECDH-ES Key exchanges that use the ConcatKDF function as defined in 800-56A

### DIFF
--- a/docs/hazmat/primitives/key-derivation-functions.rst
+++ b/docs/hazmat/primitives/key-derivation-functions.rst
@@ -324,6 +324,192 @@ Different KDFs are suitable for different tasks such as:
         ``key_material`` generates the same key as the ``expected_key``, and
         raises an exception if they do not match.
 
+.. currentmodule:: cryptography.hazmat.primitives.kdf.concatkdf
+
+.. class:: ConcatKDFHash(algorithm, length, otherinfo, backend)
+
+    .. versionadded:: 1.0
+
+    ConcatKDFHash (Concatenation Key Derivation Function) is defined by the
+    NIST Special Publication `NIST SP 800-56Ar2`_ document, to be used to
+    derive keys for use after a Key Exchange negotiation operation.
+
+    .. warning::
+
+        ConcatKDFHash should not be used for password storage.
+
+    .. doctest::
+
+        >>> import os
+        >>> from cryptography.hazmat.primitives import hashes
+        >>> from cryptography.hazmat.primitives.kdf.concatkdf import ConcatKDFHash
+        >>> from cryptography.hazmat.backends import default_backend
+        >>> backend = default_backend()
+        >>> salt = os.urandom(16)
+        >>> otherinfo = b"concatkdf-example"
+        >>> ckdf = ConcatKDFHash(
+        ...     algorithm=hashes.SHA256(),
+        ...     length=256,
+        ...     otherinfo=otherinfo,
+        ...     backend=backend
+        ... )
+        >>> key = ckdf.derive(b"input key")
+        >>> ckdf = ConcatKDFHash(
+        ...     algorithm=hashes.SHA256(),
+        ...     length=256,
+        ...     otherinfo=otherinfo,
+        ...     backend=backend
+        ... )
+        >>> ckdf.verify(b"input key", key)
+
+    :param algorithm: An instance of a
+        :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm`
+        provider
+
+    :param int length: The desired length of the derived key in bytes.
+        Maximum is ``hashlen * (2^32 -1)``.
+
+    :param bytes otherinfo: Application specific context information.
+        If ``None`` is explicitly passed an empty byte string will be used.
+
+    :param backend: A
+        :class:`~cryptography.hazmat.backends.interfaces.HashBackend`
+        provider.
+
+    :raises cryptography.exceptions.UnsupportedAlgorithm: This is raised
+        if the provided ``backend`` does not implement
+        :class:`~cryptography.hazmat.backends.interfaces.HashBackend`
+
+    :raises TypeError: This exception is raised if ``otherinfo`` is not
+        ``bytes``.
+
+    .. method:: derive(key_material)
+
+        :param bytes key_material: The input key material.
+        :return bytes: The derived key.
+        :raises TypeError: This exception is raised if ``key_material`` is
+                            not ``bytes``.
+
+        Derives a new key from the input key material by performing both the
+        extract and expand operations.
+
+    .. method:: verify(key_material, expected_key)
+
+        :param bytes key_material: The input key material. This is the same as
+                                   ``key_material`` in :meth:`derive`.
+        :param bytes expected_key: The expected result of deriving a new key,
+                                   this is the same as the return value of
+                                   :meth:`derive`.
+        :raises cryptography.exceptions.InvalidKey: This is raised when the
+                                                    derived key does not match
+                                                    the expected key.
+        :raises cryptography.exceptions.AlreadyFinalized: This is raised when
+                                                          :meth:`derive` or
+                                                          :meth:`verify` is
+                                                          called more than
+                                                          once.
+
+        This checks whether deriving a new key from the supplied
+        ``key_material`` generates the same key as the ``expected_key``, and
+        raises an exception if they do not match.
+
+
+.. class:: ConcatKDFHMAC(algorithm, length, salt, otherinfo, backend)
+
+    .. versionadded:: 1.0
+
+    Similar to ConcatKFDHash but uses an HMAC function instead.
+
+    .. warning::
+
+        ConcatKDFHMAC should not be used for password storage.
+
+    .. doctest::
+
+        >>> import os
+        >>> from cryptography.hazmat.primitives import hashes
+        >>> from cryptography.hazmat.primitives.kdf.concatkdf import ConcatKDFHMAC
+        >>> from cryptography.hazmat.backends import default_backend
+        >>> backend = default_backend()
+        >>> salt = os.urandom(16)
+        >>> otherinfo = b"concatkdf-example"
+        >>> ckdf = ConcatKDFHMAC(
+        ...     algorithm=hashes.SHA256(),
+        ...     length=256,
+        ...     salt=salt,
+        ...     otherinfo=otherinfo,
+        ...     backend=backend
+        ... )
+        >>> key = ckdf.derive(b"input key")
+        >>> ckdf = ConcatKDFHMAC(
+        ...     algorithm=hashes.SHA256(),
+        ...     length=256,
+        ...     salt=salt,
+        ...     otherinfo=otherinfo,
+        ...     backend=backend
+        ... )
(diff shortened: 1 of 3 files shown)


src/cryptography/hazmat/primitives/kdf/concatkdf.py
```diff
@@ -0,0 +1,94 @@
...
+            )
+
+        self._algorithm = algorithm
+
+        max_length = algorithm.digest_size * (2 ** 32 - 1)
+        if length > max_length:
+            raise ValueError(
+                "Can not derive keys larger than {0} bits.".format(
+                    max_length
+                ))
```
Ayrx: Style nit - the second ")" should be on the next line.

src/cryptography/hazmat/primitives/kdf/concatkdf.py
```diff
@@ -0,0 +1,94 @@
...
+        self._length = length
+
+        if not isinstance(salt, bytes) and salt is not None:
```
Ayrx: Is the `salt is not None` bit necessary? `isinstance(None, bytes)` is always False.

src/cryptography/hazmat/primitives/kdf/concatkdf.py
```diff
@@ -0,0 +1,94 @@
...
+            raise TypeError("salt must be bytes.")
+        if salt is None:
+            salt = b"\x00" * algorithm.block_size
+        self._salt = salt
+
+        if not isinstance(otherinfo, bytes) and otherinfo is not None:
```
Ayrx: Same as above.

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -324,6 +324,110 @@ Different KDFs are suitable for different tasks such as:
         ``key_material`` generates the same key as the ``expected_key``, and
         raises an exception if they do not match.
 
+.. currentmodule:: cryptography.hazmat.primitives.kdf.concatkdf
+
+.. class:: ConcatKDF(algorithm, length, otherinfo, backend)
+
+    .. versionadded:: 0.9
+
+    `ConcatKDF`_ (Concatenation Key Derivation Function) is defined by the NIST
```
Ayrx: This is the bit causing the test failure. Is ConcatKDF supposed to link to anything?

simo5: I think I mean tot link to 800-56Ar2 ... will fix this

src/cryptography/hazmat/primitives/kdf/concatkdf.py
simo5: All the above have been copied straight from the hkdf.py code, to maintain the same style. So I do not think I should change them now.

Ayrx: Oh, looks like I made a mistake here. This only raises the exception if the salt isn't `bytes` and isn't None. Feel free to disregard. :)

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -324,6 +324,110 @@ Different KDFs are suitable for different tasks such as:
...
+
+    .. warning::
+
+        ConcatKDF should not be used for password storage.
+
+    .. doctest::
+
+        >>> import os
+        >>> from cryptography.hazmat.primitives import hashes
+        >>> from cryptography.hazmat.primitives.kdf.contcatkdf import ConcatKDF
```
Ayrx: "contcatkdf" Typo here.

simo5: ----- Original Message -----
> From: "Terry Chia" <notifications@github.com>
> To: "pyca/cryptography" <cryptography@noreply.github.com>
> Cc: "simo5" <simo@fedoraproject.org>
> Sent: Tuesday, May 5, 2015 9:32:49 AM
> Subject: Re: [cryptography] Add ConcatKDF from NIST SP 800-56A (#1898)
> 
> > +
> > +
> > +    ConcatKDF (Concatenation Key Derivation Function) is defined by the
> > NIST
> > +    Special Publication `NIST SP 800-56Ar2`_ document, to be used to
> > derive
> > +    keys for use after a Key Exchange negotiation operation.
> > +
> > +    .. warning::
> > +
> > +        ConcatKDF should not be used for password storage.
> > +
> > +    .. doctest::
> > +
> > +        >>> import os
> > +        >>> from cryptography.hazmat.primitives import hashes
> > +        >>> from cryptography.hazmat.primitives.kdf.contcatkdf import
> > ConcatKDF
> 
> "contcatkdf" Typo here.

Fixed and pushing.

Simo.

-- 
Simo Sorce * Red Hat, Inc. * New York

reaperhulk:
@Ayrx where are we with this PR? Have you taken a look at the algorithm itself?

simo5:
Just FYI: the test vector I am using in the basic unit tests has been taken straight from one of the NIST vectors merged in #1884

Ayrx:
I have taken a cursory look at the algorithm and it seems fine. I'll want
to do one more look before giving it my +1 to make sure I didn't miss out
anything.

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -372,6 +478,7 @@ Interface
 
 
 .. _`NIST SP 800-132`: http://csrc.nist.gov/publications/nistpubs/800-132/nist-sp800-132.pdf
+.. _`NIST SP 800-56Ar2`: http://dx.doi.org/10.6028/NIST.SP.800-56Ar2
```
Ayrx: This URL actually `303` redirects me to http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar2.pdf. Is there any reason not to link to this URL instead?

simo5: No good reason, fixed and pushed a new patch.

Ayrx:
Ok, the implementation itself seems fine to me. I'm not sure about the API design though. I'd personally prefer to pass in a `Hash()` or `HMAC()` object instead of a boolean flag or even separate classes for the two. @reaperhulk @alex will probably have more opinions on API design and I defer to their expertise in the matter. :)

public:
Do any protocols actually use this KDF?

Also I agree with @Ayrx.

simo5:
On Sun, 2015-05-10 at 01:54 -0700, Alex Stapleton wrote:
> Do any protocols actually use this KDF?

ECDH Key Exchanges do.

> Also I agree with @Ayrx.

Please provide a tentative signature.

Simo.

-- 

public:
They don't *all* use this KDF do they? e.g. TLS seems to use it's regular PRF even with ECDH exchanges.

I suggest removing the option to run this without an HMAC entirely for now unless there's a particular application in mind which relies on that?

simo5:
The option to use it w/o HMAC is necessary as the test Vectors run w/o HMAC.
I think I'll need the version w/o HMAC in order to implement the ECDH-ES algorithm defined in http://tools.ietf.org/html/draft-ietf-jose-json-web-algorithms-24#section-4.6 which is why I am implementing ConcatKDF at all.

reaperhulk:
@public Since you're most familiar with this where do we stand now? Do we want to do a NoHMAC placeholder object here?

reaperhulk:
@simo5 I think the way forward here is to the ConcatKDF class into two classes (ConcatKDFHMAC and ConcatKDFHash).

The signatures could look like:

```python
ConcatKDFHash(
     algorithm=hashes.SHA256(),
     length=256,
     otherinfo=otherinfo,
     backend=backend
)
ConcatKDFHMAC(
     algorithm=hashes.SHA256(),
     length=256,
     salt=salt,
     otherinfo=otherinfo,
     backend=backend
)
```
Much of the logic will be shared but it makes the interfaces for instantiating these things much cleaner IMO.

Ayrx:
+1, I think that's the cleanest interface.

public:
+1 on the suggestion from @reaperhulk

simo5:
Interfaces changed as suggested.

simo5:
All docs and pep8 errors should be fixed too now.

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -324,6 +324,192 @@ Different KDFs are suitable for different tasks such as:
...
+
+    :param algorithm: An instance of a
+        :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm`
+        provider
+
+    :param int length: The desired length of the derived key in bytes. Maximum
+        is ``hashlen * (2^32 -1)``.
+
+    :param bytes salt: A salt. Randomizes the KDF's output. Optional, and used
+        only if ``usehmac`` is ``True``, but highly recommended. Ideally as
```
reaperhulk: This needs to be updated for the new API

src/cryptography/hazmat/primitives/kdf/concatkdf.py
```diff
@@ -0,0 +1,125 @@
...
+
+        if not isinstance(backend, HashBackend):
+            raise UnsupportedAlgorithm(
+                "Backend object does not implement HashBackend.",
+                _Reasons.BACKEND_MISSING_INTERFACE
+            )
+        self._backend = backend
+        self._used = False
+
+    def _hash(self):
```
reaperhulk: Let's just inline this inside derive now.

src/cryptography/hazmat/primitives/kdf/concatkdf.py
```diff
@@ -0,0 +1,125 @@
...
+
+        if not isinstance(backend, HMACBackend):
+            raise UnsupportedAlgorithm(
+                "Backend object does not implement HMACBackend.",
+                _Reasons.BACKEND_MISSING_INTERFACE
+            )
+        self._backend = backend
+        self._used = False
+
+    def _hmac(self):
```
reaperhulk: Let's just inline this inside derive now.

src/cryptography/hazmat/primitives/kdf/concatkdf.py
simo5: Inline how ?

reaperhulk: Instead of passing `self._hash` to `_concatkdf_derive` we can just pass `hashes.Hash(self._algorithm, self._backend)` since self._hash isn't invoked anywhere else. This does call it and pass the context rather than a callable, but it doesn't look like `_concatkdf_derive` needs a callable since it just invokes it and uses the context generated.

simo5: No, we can't pass hashes.Hash() .. we need  to instantiate a new Hash/HMAC object on every cycle in the 'while' loop, that's why I am passing it as a function pointer instead of a class instance.

reaperhulk: Whoops, yeah, I read this wrong. Nevermind :)

Issue #1863, opened 2015-04-23
Update openssl bindings to allow client side OCSP stapling


adiroiban:
I am looking at implementing OCSP stapling for Twisted  (https://twistedmatrix.com/trac/ticket/6802) so I am documenting the work items I have identified so far

Docs: 
* https://www.openssl.org/docs/ssl/SSL_CTX_set_client_cert_cb.html
* I could not find the public docs for SSL_set_tlsext_status_type

OpenSSL client sample code: https://github.com/openssl/openssl/blob/master/apps/s_client.c#L1530

```
# Set extension and set callback 
SSL_set_tlsext_status_type(con, TLSEXT_STATUSTYPE_ocsp);
SSL_CTX_set_tlsext_status_cb(ctx, ocsp_resp_cb);
SSL_CTX_set_tlsext_status_arg(ctx, bio_c_out);

static int ocsp_resp_cb(SSL *s, void *arg)
{
    const unsigned char *p;
    int len;
    OCSP_RESPONSE *rsp;
    len = SSL_get_tlsext_status_ocsp_resp(s, &p);
    BIO_puts(arg, "OCSP response: ");
    if (!p) {
        BIO_puts(arg, "no response sent\n");
        return 1;
    }
    rsp = d2i_OCSP_RESPONSE(NULL, &p, len);
    if (!rsp) {
        BIO_puts(arg, "response parse error\n");
        BIO_dump_indent(arg, (char *)p, len, 4);
        return 0;
    }
    BIO_puts(arg, "\n======================================\n");
    OCSP_RESPONSE_print(arg, rsp, 0);
    BIO_puts(arg, "======================================\n");
    OCSP_RESPONSE_free(rsp);
    return 1;
}
```

alex:
Missing things:

- [ ] `SSL_set_tlsext_status_type`
- [ ] `SSL_CTX_set_tlsext_status_cb`
- [ ] `SSL_CTX_set_tlsext_status_arg`
- [ ] `OCSP_RESPONSE`
- [ ] `SSL_get_tlsext_status_ocsp_resp`
- [ ] `d2i_OCSP_RESPONSE`
- [ ] `OCSP_RESPONSE_print`
- [ ] `OCSP_RESPONSE_free`

adiroiban:
I am not sure if we need OCSP_RESPONSE_print ... I think that we will need to decode the DER content and construct from in an OCSP response object.

[...]

PR #1905, opened 2015-05-05
Add OpenSSL binding for SSL_set_tlsext_status_type

eeshangarg:
Refs #1863. :-)

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/ssl.py
+++ b/src/cryptography/hazmat/bindings/openssl/ssl.py
@@ -22,6 +22,7 @@ static const long Cryptography_HAS_SECURE_RENEGOTIATION;
 static const long Cryptography_HAS_COMPRESSION;
 static const long Cryptography_HAS_TLSEXT_STATUS_REQ_CB;
 static const long Cryptography_HAS_STATUS_REQ_OCSP_RESP;
+static const long Cryptography_HAS_TLSEXT_STATUS_REQ_TYPE;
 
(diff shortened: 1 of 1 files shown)


src/cryptography/hazmat/bindings/openssl/ssl.py
```diff
@@ -438,6 +440,13 @@
 long (*SSL_get_tlsext_status_ocsp_resp)(SSL *, const unsigned char **) = NULL;
 #endif
 
+#ifdef SSL_CTRL_SET_TLSEXT_STATUS_REQ_TYPE
```
reaperhulk: Could you put a comment that explains what version this was added. It will help us in the future.

eeshangarg: Well, as you mentioned on IRC, ```SSL_set_tlsext_status_type``` was added in OpenSSL 0.9.8h and I checked and ```SSL_CTX_set_tlsext_status_cb``` was also added in 0.9.8h, so I don't think we need to add a separate comment as [this comment](https://github.com/eeshangarg/cryptography/blob/add-openssl-binding-for-SSL_set_tlsext_status_type/src/cryptography/hazmat/bindings/openssl/ssl.py#L325) already says that the four macros following it were added in 0.9.8h. :-)

Thanks a lot for your help and guidance! Let me know if this needs further work!

reaperhulk: Agreed, thanks eeshan.

PR #1853, opened 2015-04-17
certificate policies extension support

reaperhulk:
Adds a bunch of ancillary classes to support this.

refs #1743

PR #1906, opened 2015-05-06
add CRLDistributionPoints and associated classes

reaperhulk:
Much like certificate policies, this has one property (distribution_point on DistributionPoint) that can be 2 different types. If we decide to do this another way (suggestions?) we can change it on both PRs.

refs #1743

### DIFF
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -835,7 +835,7 @@ X.509 Extensions
 
     .. attribute:: reasons
 
-        :type: list of :class:`ReasonFlags` or None
+        :type: frozenset of :class:`ReasonFlags` or None
 
         The reasons a given distribution point may be used for when performing
         revocation checks.
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -781,6 +781,8 @@ X.509 Extensions
 
 .. class:: AccessDescription
 
+    .. versionadded:: 0.9
+
     .. attribute:: access_method
 
         :type: :class:`ObjectIdentifier`
@@ -797,6 +799,76 @@ X.509 Extensions
         :type: :class:`GeneralName`
 
         Where to access the information defined by the access method.
+
+.. class:: CRLDistributionPoints
+
+    .. versionadded:: 0.9
+
+    The CRL distribution points extension identifies how CRL information is
+    obtained. It is an iterable, containing one or more
+    :class:`DistributionPoint` instances.
+
+.. class:: DistributionPoint
+
+    .. versionadded:: 0.9
+
+    .. attribute:: distribution_point
+
+        :type: list of :class:`GeneralName` instances, :class:`Name`, or None
+
+        This field describes methods to retrieve the CRL.
+
+    .. attribute:: crl_issuer
+
+        :type: list of :class:`GeneralName` instances or None
+
+        Information about the issuer of the CRL.
+
+    .. attribute:: reasons
+
+        :type: :class:`ReasonFlags` or None
+
+        The reasons a given distribution point may be used for when performing
+        revocation checks.
+
+.. class:: ReasonFlags
+
+    .. versionadded:: 0.9
+
+    This class holds reasons a distribution point may be used for when
+    performing revocation checks.
+
+    .. attribute:: key_compromise
+
+        :type: bool
+
+    .. attribute:: ca_compromise
+
+        :type: bool
+
+    .. attribute:: affiliation_changed
+
+        :type: bool
+
+    .. attribute:: superseded
+
+        :type: bool
+
+    .. attribute:: cessation_of_operation
+
+        :type: bool
+
+    .. attribute:: certificate_hold
+
+        :type: bool
+
+    .. attribute:: privilege_withdrawn
+
+        :type: bool
+
+    .. attribute:: aa_compromise
+
+        :type: bool
 
 Object Identifiers
 ~~~~~~~~~~~~~~~~~~
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -812,11 +812,19 @@ X.509 Extensions
 
     .. versionadded:: 0.9
 
-    .. attribute:: distribution_point
+    .. attribute:: full_name
 
-        :type: list of :class:`GeneralName` instances, :class:`Name`, or None
+        :type: list of :class:`GeneralName` instances or None
 
-        This field describes methods to retrieve the CRL.
+        This field describes methods to retrieve the CRL. If this is not None
+        then ``relative_name`` must be None.
+
+    .. attribute:: relative_name
+
+        :type: :class:`Name` or None
+
+        This field describes methods to retrieve the CRL relative to the CRL
+        issuer. If this is not None then ``full_name`` must be None.
 
     .. attribute:: crl_issuer
 
@@ -826,7 +834,7 @@ X.509 Extensions
 
     .. attribute:: reasons
 
-        :type: :class:`ReasonFlags` or None
+        :type: list of :class:`ReasonFlags` or None
 
         The reasons a given distribution point may be used for when performing
         revocation checks.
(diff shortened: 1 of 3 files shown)


schlenk:
Looks good.

docs/x509.rst
```diff
@@ -798,6 +800,76 @@ X.509 Extensions
...
+    obtained. It is an iterable, containing one or more
+    :class:`DistributionPoint` instances.
+
+.. class:: DistributionPoint
+
+    .. versionadded:: 0.9
+
+    .. attribute:: distribution_point
+
+        :type: list of :class:`GeneralName` instances, :class:`Name`, or None
```
alex: Oof, is there really no better way to represent this.

docs/x509.rst
```diff
@@ -798,6 +800,76 @@ X.509 Extensions
...
+        The reasons a given distribution point may be used for when performing
+        revocation checks.
+
+.. class:: ReasonFlags
+
+    .. versionadded:: 0.9
+
+    This class holds reasons a distribution point may be used for when
+    performing revocation checks.
+
```
alex: These each need some words describing them.

docs/x509.rst
reaperhulk: We have the same "CHOICE" definition here as in PolicyQualifierInfo (in #1853). The definition of DistributionPoint looks like this:

```
   DistributionPoint ::= SEQUENCE {
        distributionPoint       [0]     DistributionPointName OPTIONAL,
        reasons                 [1]     ReasonFlags OPTIONAL,
        cRLIssuer               [2]     GeneralNames OPTIONAL }

   DistributionPointName ::= CHOICE {
        fullName                [0]     GeneralNames,
        nameRelativeToCRLIssuer [1]     RelativeDistinguishedName }
```

We could, instead of calling the attribute distribution_point, separate them into full_name and name_relative_to_crl_issuer. Then at most one of them could be not None.

etrauschke:
Wow, I didn't notice that the values aren't in sync at all. I always thought we should still have the numerical value in this class because the string representation could potentially be (slightly) different between x509 implementations. Your definitions are also different than mine, I was looking at https://www.ietf.org/rfc/rfc2459 which actually show both matching much better:
```
ReasonFlags ::= BIT STRING {
        unused                  (0),
        keyCompromise           (1),
        cACompromise            (2),
        affiliationChanged      (3),
        superseded              (4),
        cessationOfOperation    (5),
        certificateHold         (6) }
```
```
CRLReason ::= ENUMERATED {
     unspecified             (0),
     keyCompromise           (1),
     cACompromise            (2),
     affiliationChanged      (3),
     superseded              (4),
     cessationOfOperation    (5),
     certificateHold         (6),
     removeFromCRL           (8) }
```

reaperhulk:
Unfortunately 2459 is obsoleted by 3280 which was, in turn, obsoleted by 5280. http://tools.ietf.org/html/rfc5280#section-5.3.1 and http://tools.ietf.org/html/rfc5280#section-4.2.1.13 have the current definitions. Frustrating isn't it? :)

etrauschke:
Yep, I wonder which moron decided that it is a good idea to have these two list to be out of sync. I guess the correct way of dealing with it is to have two completely different classes which both have their own definitions (including the numerical value). This is ugly but it follows what the "standard" is doing.

I wonder how this is supposed to work anyway. If I select a distribution point which should support privilegeWithdrawn I test if 7 is part of the reason flags. But then I will never get a revoked certificate for reason 7.

reaperhulk:
I don't think we really need to test for the numbers as we can map them in the backend to the enumeration. In the latest commit I've created an enumeration (ReasonFlags, although maybe we should just call it CRLReason now?) and modified DistributionPoint to require a list of those flags (with unspecified and remove_from_crl being disallowed).

I also modified the distribution_point attribute and replaced it with the full_name/relative_name approach (along with some code to check that at most one is not None).

docs/x509.rst
```diff
@@ -798,6 +800,97 @@ X.509 Extensions
...
+        This field describes methods to retrieve the CRL. If this is not None
+        then ``relative_name`` must be None.
+
+    .. attribute:: relative_name
+
+        :type: :class:`Name` or None
+
+        This field describes methods to retrieve the CRL relative to the CRL
+        issuer. If this is not None then ``full_name`` must be None.
```
alex: This would be much clearer written as "Exactly one of ``full_name`` or ``relative_name`` will be non-None" I think

docs/x509.rst
```diff
@@ -798,6 +800,97 @@ X.509 Extensions
...
+
+    .. attribute:: crl_issuer
+
+        :type: list of :class:`GeneralName` instances or None
+
+        Information about the issuer of the CRL.
+
+    .. attribute:: reasons
+
+        :type: list of :class:`ReasonFlags` or None
```
alex: Hmm, why a list of flags vs. an object with some attributes, this is inconsistent with how we did KU.

docs/x509.rst
reaperhulk: Both can be none though.

alex: "At most one of ... or ... will be non-None"?

On Sat, May 9, 2015 at 1:12 PM, Paul Kehrer <notifications@github.com>
wrote:

> In docs/x509.rst
> <https://github.com/pyca/cryptography/pull/1906#discussion_r29994448>:
>
> > +
> > +
> > +    .. attribute:: full_name
> > +
> > +        :type: list of :class:`GeneralName` instances or None
> > +
> > +        This field describes methods to retrieve the CRL. If this is not None
> > +        then ``relative_name`` must be None.
> > +
> > +    .. attribute:: relative_name
> > +
> > +        :type: :class:`Name` or None
> > +
> > +        This field describes methods to retrieve the CRL relative to the CRL
> > +        issuer. If this is not None then ``full_name`` must be None.
>
> Both can be none though.
>
> —
> <https://github.com/pyca/cryptography/pull/1906/files#r29994448>.
>



-- 

docs/x509.rst
reaperhulk: These flags are reused (mostly) as individual reasons in CRLs. So this can be a list of them but a revocation uses a single reason. That's why I changed the approach from the one we used for KU (the previous commit on this Pr had the old approach)

docs/x509.rst
reaperhulk: A+

docs/x509.rst
alex: Should this be a frozenset() instead?

reaperhulk: I'm not sure! We use list most everywhere, but this definitely doesn't care about order (since it's really a bitfield) so maybe? Do you think it makes sense?

alex: I think it does, since we want uniqueness here.

On Sat, May 9, 2015 at 6:05 PM, Paul Kehrer <notifications@github.com>
wrote:

> In docs/x509.rst
> <https://github.com/pyca/cryptography/pull/1906#discussion_r29996697>:
>
> > +    .. attribute:: relative_name
> > +
> > +        :type: :class:`Name` or None
> > +
> > +        This field describes methods to retrieve the CRL relative to the CRL
> > +        issuer. If this is not None then ``full_name`` must be None.
> > +
> > +    .. attribute:: crl_issuer
> > +
> > +        :type: list of :class:`GeneralName` instances or None
> > +
> > +        Information about the issuer of the CRL.
> > +
> > +    .. attribute:: reasons
> > +
> > +        :type: list of :class:`ReasonFlags` or None
>
> I'm not sure! We use list most everywhere, but this definitely doesn't
> care about order (since it's really a bitfield) so maybe? Do you think it
> makes sense?
>
> —
> <https://github.com/pyca/cryptography/pull/1906/files#r29996697>.
>



-- 

reaperhulk: Should we require you to pass in a frozenset or allow any iterable and just frozenset it when setting the attr?

alex: Let's require a frozenset, we can always loosen that later.

On Sat, May 9, 2015 at 8:19 PM, Paul Kehrer <notifications@github.com>
wrote:

> In docs/x509.rst
> <https://github.com/pyca/cryptography/pull/1906#discussion_r29997497>:
>
> > +    .. attribute:: relative_name
> > +
> > +        :type: :class:`Name` or None
> > +
> > +        This field describes methods to retrieve the CRL relative to the CRL
> > +        issuer. If this is not None then ``full_name`` must be None.
> > +
> > +    .. attribute:: crl_issuer
> > +
> > +        :type: list of :class:`GeneralName` instances or None
> > +
> > +        Information about the issuer of the CRL.
> > +
> > +    .. attribute:: reasons
> > +
> > +        :type: list of :class:`ReasonFlags` or None
>
> Should we require you to pass in a frozenset or allow any iterable and
> just frozenset it when setting the attr?
>
> —
> <https://github.com/pyca/cryptography/pull/1906/files#r29997497>.
>



-- 

PR #1907, opened 2015-05-06
Ecdh kdf tests

simo5:
Add separate pull request for testing the ECDH with KDF vectors as requested on #1884


simo5:
This PR depends on #1884 
We should probably hold on this until #1898 is also merged and add code to actually test the KDF derivation in the vectors.

public:
I don't think we need to wait for a new KDF land to make use of these vectors.

public:
LGTM

PR #1915, opened 2015-05-09
add support for retreiving value of received temporary key in server …

zeevt:
…key exchange message, to implement functionality like openssl's ssl_print_tmp_key

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/ec.py
+++ b/src/cryptography/hazmat/bindings/openssl/ec.py
@@ -17,6 +17,7 @@ static const int Cryptography_HAS_EC;
 static const int Cryptography_HAS_EC_1_0_1;
 static const int Cryptography_HAS_EC_NISTP_64_GCC_128;
 static const int Cryptography_HAS_EC2M;
+static const int Cryptography_HAS_EC_1_0_2;
 
 static const int OPENSSL_EC_NAMED_CURVE;
 
@@ -188,6 +189,8 @@ const EC_METHOD *EC_GFp_nistp521_method();
 const EC_METHOD *EC_GF2m_simple_method();
 
 int EC_METHOD_get_field_type(const EC_METHOD *);
+
+const char *EC_curve_nid2nist(int);
 """
 
 CUSTOMIZATIONS = """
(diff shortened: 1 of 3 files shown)


alex:
ok to test

src/cryptography/hazmat/bindings/openssl/ec.py
```diff
@@ -188,6 +189,8 @@
 const EC_METHOD *EC_GF2m_simple_method();
 
 int EC_METHOD_get_field_type(const EC_METHOD *);
+
+const char *EC_curve_nid2nist(int nid);
```
alex: The names of parameters should not be included, meaning this should be written `const char *EC_curve_nid2nist(int);`

(This applies on a few other lines)

src/cryptography/hazmat/bindings/openssl/ssl.py
```diff
@@ -591,6 +595,18 @@
 static const long Cryptography_HAS_COMPRESSION = 1;
 #endif
 
+#if defined(SSL_CTRL_GET_SERVER_TMP_KEY)
+static const long Cryptography_HAS_GET_SERVER_TMP_KEY = 1;
+#undef SSL_get_server_tmp_key
+long SSL_get_server_tmp_key(SSL *ssl, EVP_PKEY **key) {
+    return SSL_ctrl(ssl, SSL_CTRL_GET_SERVER_TMP_KEY, 0, key);
+}
```
alex: Why are you redefining the macro? You should just be able to bind it, same as the other macros.

zeevt:
Oh, it's building again before I was ready. Still working on fixing libressl compatibility.

reaperhulk:
Any reason you want nid2nist over `OBJ_nid2sn`? (Example of that being used to get the curve name here: https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/backends/openssl/ec.py#L43)

I'm fine with binding this, just curious.

zeevt:
@reaperhulk: I've implemented openssl's `ssl_print_tmp_key`[1] for pyopenssl and want cryptography to have all the symbols (if openssl 1.0.2 is used). The SECG name of the curve is enough, the NIST name is not strictly necessary, but openssl thought it would be nice to have, so I'd like to have it.

1 - https://git.openssl.org/gitweb/?p=openssl.git;a=blob;f=apps/s_cb.c;h=ee57537083afd3714f14300ec9532691cbb288ca;hb=refs/heads/OpenSSL_1_0_2-stable#l506

reaperhulk:
Sounds reasonable :)

Issue #1922, opened 2015-05-11
Add more client cert related bindings


tiran:
I like to add a few additional bindings for client cert callbacks.

* SSL_CTX_set_client_cert_cb()
* SSL_set_cert_cb() and SSL_CTX_set_cert_cb (OpenSSL 1.0.2+)
* SSL_use_*() to install cert and priv key in SSL* rather than SSL_CTX*
* ssl_client_cert_function for ENGINE API

PR #1923, opened 2015-05-11
Feature/1922 client cert cb

tiran:
Pull request for #1922

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/ssl.py
+++ b/src/cryptography/hazmat/bindings/openssl/ssl.py
@@ -45,6 +45,8 @@ static const long Cryptography_HAS_NETBSD_D1_METH;
 static const long Cryptography_HAS_NEXTPROTONEG;
 static const long Cryptography_HAS_ALPN;
 
+static const long Cryptography_HAS_102_SET_CERT_CB;
+
 static const long SSL_FILETYPE_PEM;
 static const long SSL_FILETYPE_ASN1;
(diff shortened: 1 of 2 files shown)


reaperhulk:
add to whitelist

src/cryptography/hazmat/bindings/openssl/ssl.py
```diff
@@ -602,6 +629,15 @@
 long (*SSL_get_server_tmp_key)(SSL *, EVP_PKEY **) = NULL;
 #endif
 
+/* OpenSSL 1.0.2+ cert validation callback */
+#if OPENSSL_VERSION_NUMBER < 0x10002000L
```
reaperhulk: This will fail on libressl (which stupidly defines openssl version as 0x20000000L). You'll also need to add `&& !defined(LIBRESSL_VERSION_NUMBER)`

reaperhulk:
CentOS 5 appears to dislike something: https://jenkins.cryptography.io/job/cryptography-pull-request-builder/3494/TOXENV=py27,label=centos5/console

tiran:
What OpenSSL version is available on CentOS 5?

reaperhulk:
It claims to be 0.9.8b but is actually 0.9.8e (welcome to the hell of Red Hat long term support versions).

alex:
Should we split this into a patch with the "easy" bindings, and teh hard ones? Would make it easier to review :-)

PR #1932, opened 2015-05-12
some bindings for certificate policies support in openssl

reaperhulk:
refs #1743

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/x509v3.py
+++ b/src/cryptography/hazmat/bindings/openssl/x509v3.py
@@ -128,17 +128,17 @@ typedef struct {
     ...;
 } DIST_POINT;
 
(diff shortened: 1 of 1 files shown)


src/cryptography/hazmat/bindings/openssl/x509v3.py
```diff
@@ -122,6 +128,32 @@
     ...;
 } DIST_POINT;
 
+typedef struct NOTICEREF_st {
```
alex: You can drop the struct name here I believe.

alex: Also applies to the other structs.

alex:
Besides that this looks good.

alex:
lgtm now

PR #1935, opened 2015-05-12
we also need some stack manipulation macros

reaperhulk:
And we haven't been using the pointless typedefs for other stacks, so dropping CERTIFICATEPOLICIES.

refs #1743


reaperhulk:
damn you alternate versions of openssl...looking into this shortly.

alex:
this looks ok, please rebase for coverage

PR #1938, opened 2015-05-12
Install argparse on python 2.6

alex:


alex:
Holger is fixing this in tox 2.0.1

PR #1939, opened 2015-05-12
pass through the env vars we need in tox for alternate OpenSSL builds

reaperhulk:
This is required in tox 2.0

INCLUDE and LIB are used in windows
ARCHFLAGS CFLAGS LDFLAGS are used on OS X
CFLAGS LDFLAGS LD_LIBRARY_PATH are used on linux

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,6 +9,7 @@ deps =
     pretend
     pytest
     ./vectors
+passenv = ARCHFLAGS LDFLAGS CFLAGS INCLUDE LIB LD_LIBRARY_PATH
 commands =
     # We use parallel mode and then combine here so that coverage.py will take
     # the paths like .tox/py34/lib/python3.4/site-packages/cryptography/__init__.py


tox.ini
```diff
@@ -9,6 +9,7 @@ deps =
     pretend
     pytest
     ./vectors
+passenv = ARCHFLAGS LDFLAGS CFLAGS INCLUDE LIB LD_LIBRARY_PATH
```
alex: Do we need `DYLD_LIBRARY_PATH` as well?

reaperhulk: We don't use it in any of our testing right now, so we probably can leave it out.

alex: Ok.

On Tue, May 12, 2015 at 7:45 PM, Paul Kehrer <notifications@github.com>
wrote:

> In tox.ini
> <https://github.com/pyca/cryptography/pull/1939#discussion_r30191873>:
>
> > @@ -9,6 +9,7 @@ deps =
> >      pretend
> >      pytest
> >      ./vectors
> > +passenv = ARCHFLAGS LDFLAGS CFLAGS INCLUDE LIB LD_LIBRARY_PATH
>
> We don't use it in any of our testing right now, so we probably can leave
> it out.
>
> —
> <https://github.com/pyca/cryptography/pull/1939/files#r30191873>.
>



-- 

PR #1940, opened 2015-05-13
support certificate policies in the openssl backend

reaperhulk:
refs #1743 

end game (for required extensions in 0.9)


alex:
This is breaking the docs tests.

Issue #1682, opened 2015-02-20
1.0 release


public:
http://semver.org

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

Should we bump our version number all the way to 1.0 for the next release?

Ayrx:
Do we even *want* to follow semver? Given our policy of removing deprecated stuff 3 releases later I don't really see the point.

dstufft:
To be honest, if we're not following semver we should probably just do ``YY.N`` releases or something. People expect something semver like when you're doing something that looks like ``X.Y``. On the other hand, semver really isn't mutually exclusive with removing deprecated stuff 3 releases later. Semver doesn't say how long to keep around deprecated functionality, just that when you do actually remove something that breaks backwards compatibility you need to signal it with a major version bump. It'd be perfectly reasonable to keep stuff through 3 major releases, or even just 3 feature releases and bump the version accordingly.

reaperhulk:
If we chose to do semver would we commit to supporting the last N "major" releases or would we just tell people they have to upgrade?

We do tend to deprecate things every release right now, so if we switched our version numbers would go up rapidly. Not really a big deal but something to be aware of.

I'd vote for keeping the status quo for the moment because we're still missing major chunks of functionality that I'd like for a "1.0", but I don't feel strongly about this.

Ayrx:
I am strongly -1 about attempting to support last N "major" releases. That will complicate the development process significantly imo.

Other than that I am +0 about switching to semver. It's not *my* preferred versioning scheme but I don't feel strongly about it. If we *do* switch I think it makes sense to do a 1.0 only when we are more feature-complete.

public:
@reaperhulk What features do you think we need for a 1.0? :)

reaperhulk:
Off the top of my head:

* full support for X509 certificate and request parsing (we're missing requests and extensions right now) 
* serialization of asymmetric keys (I consider at least basic support for that our primary blocker for 0.8).
* framework in place for supporting HSMs (#1506)

Other possible things I might want:

* Optional backends actually being possible (we may or may not need this for HSM support.)
* more than fernet for recipes

alex:
Personally I think we should just continue with our current numbering and
when we get to `0.10` call it `1.0`, then `1.1`, etc.

On Fri, Feb 20, 2015 at 6:48 AM, Paul Kehrer <notifications@github.com>
wrote:

> Off the top of my head:
>
[...]

PR #1949, opened 2015-05-14
open master for tenth release

reaperhulk:


alex:
Oh gosh, I guess we have to decide what we're calling the next release, I'm a fan of just letting 1.0 happen (clearly you aren't :P)

reaperhulk:
Eh, I don't think the numbering is a blocker for opening dev. We can renumber to 1.0 during the cycle if we decide to. There's some discussion in #1682.

alex:
sounds ok to me

Issue #1947, opened 2015-05-13
X509 Extensions Redux


reaperhulk:
This is a continuation of #1743 to support all X509 extensions in [RFC 5280](http://tools.ietf.org/html/rfc5280).

- [x] Issuer Alternative Name (same as SAN)
  - [x] Test vectors
  - [x] Data classes
  - [x] OpenSSL implementation
- [x] Inhibit anyPolicy 
  - [x] Test vectors
  - [x] Data classes
  - [x] OpenSSL implementation
- [x] Name Constraints 
  - [x] Test vectors
  - [x] Data classes
  - [x] OpenSSL implementation

From RFC 2560 (OCSP):
- [x] OCSP no check

There's also a set of extensions we've chosen not to implement at all until users ask for them:
- [ ] Policy Mappings 
  - [ ] Test vectors
  - [ ] Data classes
  - [ ] OpenSSL implementation
- [ ] Subject Directory Attributes 
  - [ ] Test vectors
  - [ ] Data classes
  - [ ] OpenSSL implementation
- [ ] Policy Constraints 
  - [x] Test vectors
  - [ ] Data classes
  - [ ] OpenSSL implementation
- [ ] Freshest CRL (a.k.a. Delta CRL Distribution Point)  (same as CDP)
  - [ ] Test vectors
  - [ ] Data classes
  - [ ] OpenSSL implementation
- [ ] Subject Information Access  (same as AIA)
  - [ ] Test vectors
  - [ ] Data classes
  - [ ] OpenSSL implementation

There are some other extensions (notably around CRLs) that we'll need to add as well.

alex:
@reaperhulk is there anything left on this issue that is actually a 1.0 blocker?

reaperhulk:
#2095 still required to close nameconstraints but that's it.

alex:
Done! Please file new issues any of the other extensions.

PR #1954, opened 2015-05-14
InhibitAnyPolicy class

reaperhulk:
refs #1947


alex:
I feel like the docs need work, but I'm going to land for now.

PR #1956, opened 2015-05-14
add bindings and interfaces for CRL objects

etrauschke:
This is just the OpenSSL bindings and interfaces for CRL object handling.
I also included the addition of the OIDs I'm gonna need, as well as the refactoring of SubjectAlternativeName. We talked about this by email, it seems this class doesn't do anything special so I renamed it to GeneralNames and subclassed SAN from it. This way we can use the same code for other GeneralNames objects as well.

### DIFF
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -390,16 +390,6 @@ X.509 CRL (Certificate Revocation List) Object
 
         The extensions encoded in the CRL.
 
-    .. method:: verify(public_key)
-
-        :param public_key: The
-            :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPublicKey` or
-            :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPublicKey` or
-            :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey`
-            that will be used to verify the CRL.
-
-        :return boolean: The result of the verification as boolean value.
-
 X.509 CSR (Certificate Signing Request) Object
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -454,7 +444,7 @@ X.509 CSR (Certificate Signing Request) Object
             authority.
 
 X.509 Revoked Certificate Object
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. class:: RevokedCertificate
 
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -208,7 +208,7 @@ X.509 Certificate Object
             :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm`
             that will be used to generate the fingerprint.
 
-        :return bytes: The fingerprint using the supplied hash algorithm as
+        :return bytes: The fingerprint using the supplied hash algorithm, as
             bytes.
 
         .. doctest::
@@ -335,6 +335,71 @@ X.509 Certificate Object
         :return bytes: The data that can be written to a file or sent
             over the network to be verified by clients.
 
+X.509 CRL (Certificate Revocation List) Object
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. class:: CertificateRevocationList
+
+    .. versionadded:: 1.0
+
+    .. method:: fingerprint(algorithm)
+
+        :param algorithm: The
+            :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm`
+            that will be used to generate the fingerprint.
+
+        :return bytes: The fingerprint using the supplied hash algorithm, as
+            bytes.
+
+    .. attribute:: signature_hash_algorithm
+
+        :type: :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm`
+
+        Returns the
+        :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm` which
+        was used in signing this CRL.
+
+    .. attribute:: issuer
+
+        :type: :class:`Name`
+
+        The :class:`Name` of the issuer.
+
+    .. attribute:: next_update
+
+        :type: :class:`datetime.datetime`
+
+        A naïve datetime representing when the next update to this CRL is
+        expected.
+
+    .. attribute:: last_update
+
+        :type: :class:`datetime.datetime`
+
+        A naïve datetime representing when the this CRL was last updated.
+
+    .. attribute:: revoked
+
+        :type: list of :class:`RevokedCertificate`
+
+        The revoked certificates listed in this CRL.
+
+    .. attribute:: extensions
+
+        :type: :class:`Extensions`
+
+        The extensions encoded in the CRL.
+
+    .. method:: verify(pubkey)
+
+        :param algorithm: The
+            :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPublicKey` or
+            :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPublicKey` or
+            :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey`
+            that will be used to verify the CRL.
+
+        :return boolean: The result of the verification as boolean value.
+
 X.509 CSR (Certificate Signing Request) Object
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -388,6 +453,30 @@ X.509 CSR (Certificate Signing Request) Object
             over the network to be signed by the certificate
             authority.
 
+X.509 Revoked Certificate Object
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. class:: RevokedCertificate
+
+    .. versionadded:: 1.0
+
+    .. attribute:: serial_number
+
+        :type: :class:`int`
+
+        An integer representing the serial number of the revoked certificate.
+
+    .. attribute:: revocation_date
+
+        :type: :class:`datetime.datetime`
+
+        A naïve datetime representing the date this certificates was revoked.
+
+    .. attribute:: extensions
+
+        :type: :class:`Extensions`
+
+        The extensions encoded in the revoked certificate.
 
 .. class:: Name
 
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -378,7 +378,7 @@ X.509 CRL (Certificate Revocation List) Object
 
         A naïve datetime representing when the this CRL was last updated.
 
-    .. attribute:: revoked
+    .. attribute:: revoked_certificates
 
         :type: list of :class:`RevokedCertificate`
 
@@ -390,9 +390,9 @@ X.509 CRL (Certificate Revocation List) Object
 
         The extensions encoded in the CRL.
 
-    .. method:: verify(pubkey)
+    .. method:: verify(public_key)
 
-        :param algorithm: The
+        :param public_key: The
             :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPublicKey` or
             :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPublicKey` or
             :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey`
--- a/src/cryptography/x509.py
+++ b/src/cryptography/x509.py
@@ -55,9 +55,9 @@ _OID_NAMES = {
     "2.5.29.17": "subjectAltName",
     "2.5.29.18": "issuerAltName",
     "2.5.29.19": "basicConstraints",
-    "2.5.29.21": "CRLReason",
-    "2.5.29.24": "InvalidityDate",
-    "2.5.29.29": "CertificateIssuer",
+    "2.5.29.21": "cRLReason",
+    "2.5.29.24": "invalidityDate",
+    "2.5.29.29": "certificateIssuer",
     "2.5.29.30": "nameConstraints",
     "2.5.29.31": "cRLDistributionPoints",
     "2.5.29.32": "certificatePolicies",
@@ -1250,12 +1250,6 @@ class CertificateRevocationList(object):
     def extensions(self):
         """
         Returns an Extensions object containing a list of CRL extensions.
-        """
-
-    @abc.abstractmethod
-    def verify(self, pubkey):
-        """
-        Verify this CRL against a given 'pubkey'.
         """
 
     @abc.abstractmethod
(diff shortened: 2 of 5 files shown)


etrauschke:
Not sure what happened with the tests but the failures seem to be due to some infrastructure/connection issues.

alex:
Jenkins, retest this please.

src/cryptography/x509.py
```diff
@@ -1158,3 +1225,18 @@ def signature_hash_algorithm(self):
         Returns a HashAlgorithm corresponding to the type of the digest signed
         in the certificate.
         """
+
+
+@six.add_metaclass(abc.ABCMeta)
+class RevokedCertificate(object):
```
reaperhulk: What about reason?

etrauschke: This will be part of the extension code, because that is essentially what it is. I checked with how extensions are implemented in the ABC for Certificate and it also isn't there.
So the decision would be if extensions should be a required property for the class.

I'm also planning to add little helper functions to the backend implementation (get_reason(), get_invalidity_date(), get_cert_issuer()). So this will be there, it's just not part of the required interface. But I'm fine with adding all of these to the interface specification.

reaperhulk: Ah, right, reasons are technically part of an extensions sequence. My mistake. We should add extensions itself to the interface for certificate now though, that was an oversight. I'll put a PR in.

etrauschke: Ok, I'll add the extensions property to the RevokedCertificate interface as well. However, at the moment I'm not planning to support any CRL extensions in this putback. They are fairly complicated and I personally have never needed them, but I can add them in a later pull request. 
In the meantime, I could just return an empty list if you want me to add extensions as a required property to the CRL class.

I might need to add the CRL extensions anyway to properly support the CRL.verify() functionality. As I just found out, it's not enough to compare if the issuers of cert and CRL match, sometimes the issuer is in the Revoked extension 'certificateIssuer'. However, you can only evaluate this extension if the CRL has an extensions which marks it as indirect.
So currently I punted on it and don't compare CRL issuer and cert issuer before I pass the two to X509_CRL_verify().

reaperhulk: I'm okay with incrementally adding things in future PRs so if you want to leave extensions out for the moment that's okay. We do want to keep master releasable at all times so if there's a case where it would do the "wrong" thing without supporting something we can implement a temporary exception for that.

etrauschke: Sorry, I got two things confused. This has nothing to do with verify(), this will work correctly.
What I meant was that I originally had a few additional convenience functions called is_revoked() and get_revoked() which would take a cert as an argument so you can check if a certain cert is actually revoked. Since revocation is done by serial number you have to make sure the cert actually matches the CRL, for which you compare issuers.
I'm just leaving these functions out so a user would have to do the logic themselves.

So no, the cryptography library would not be doing anything wrong, it just might not provide the interfaces you need.

reaperhulk:
Looks like we still need docs in x509.rst for CertificateRevocationList?

etrauschke:
Yes, I was mentioning this is my combined PR. I just wanted to know if there is a tool which generates the docs or at least stubs for it or if you just write it all by hand?

reaperhulk:
We write it all by hand right now

etrauschke:
Ok, let me get on that then.

etrauschke:
Rebased all my changes and synced with the current master.

etrauschke:
Apparently there is a problem with my doc test entries. I thought they were just examples and I tried to mimic what has been done for the other classes. So how do these doc tests need to be constructed?
Otherwise, I can just remove them.

etrauschke:
I think these doc tests will only work once we have the backend implementation in the source. So I removed the doc test entries for now. I will add them again in the PR for the backend code.

etrauschke:
Bump. Are there any other concerns with this PR?

docs/x509.rst
```diff
@@ -335,6 +335,71 @@ X.509 Certificate Object
...
+
+    .. attribute:: last_update
+
+        :type: :class:`datetime.datetime`
+
+        A naïve datetime representing when the this CRL was last updated.
+
+    .. attribute:: revoked
+
+        :type: :class:`RevokedCertificate`
```
reaperhulk: This is really a list of RevokedCertificate objects right?

docs/x509.rst
```diff
@@ -335,6 +335,71 @@ X.509 Certificate Object
         :return bytes: The data that can be written to a file or sent
             over the network to be verified by clients.
 
+X.509 CRL (Certificate Revocation List) Object
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. class:: CertificateRevocationList
+
+    .. versionadded:: 0.10
```
reaperhulk: We decided to call the next release 1.0 so this needs to be updated (sorry!)

docs/x509.rst
```diff
@@ -388,6 +453,30 @@ X.509 CSR (Certificate Signing Request) Object
             over the network to be signed by the certificate
             authority.
 
+X.509 Revoked Certificate Object
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. class:: RevokedCertificate
+
+    .. versionadded:: 0.10
```
reaperhulk: (1.0 as well)

src/cryptography/x509.py
```diff
@@ -976,6 +982,11 @@ def __ne__(self, other):
         return not self == other
 
 
+class SubjectAlternativeName(GeneralNames):
```
reaperhulk: I believe@alex has some thoughts on this.

reaperhulk:
A few small comments + the outstanding question to @alex about the subclass. Otherwise this looks good to me.

src/cryptography/x509.py
alex: Yup! `GeneralNames` should be a new class that `SubjectAltNames` composes around and proxies the methods, not subclasses.

docs/x509.rst
```diff
@@ -335,6 +335,71 @@ X.509 Certificate Object
...
+
+    .. method:: fingerprint(algorithm)
+
+        :param algorithm: The
+            :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm`
+            that will be used to generate the fingerprint.
+
+        :return bytes: The fingerprint using the supplied hash algorithm as
```
alex: There should be a comma before "as" here I think.

src/cryptography/x509.py
etrauschke: You mean like this:

---
class SubjectAlternativeName(object):
    def __init__(self, general_names):
        if not all(isinstance(x, GeneralName) for x in general_names):
            raise TypeError(
                "Every item in the general_names list must be an "
                "object conforming to the GeneralName interface"
            )

        self._general_names = GeneralNames(general_names)

    def __iter__(self):
        return iter(self._general_names)

    def __len__(self):
        return len(self._general_names)

    def get_values_for_type(self, type):
        return [i.value for i in self if isinstance(i, type)]

    def __repr__(self):
        return "<SubjectAlternativeName({0})>".format(self._general_names)

    def __eq__(self, other):
        if not isinstance(other, SubjectAlternativeName):
            return NotImplemented

        return self._general_names == other._general_names

    def __ne__(self, other):
        return not self == other
---

This doesn't look like it gives us any benefit but a lot of duplicate code.
The only difference between the GeneralNames class and the
SubjectAlternateName class is the name of the class, seems like the
textbook example for subclassing to me.

alex: You can write the `get_values_for_type` as `self._general_names.get_values_for_type`, but otherwise it looks good to me.

etrauschke:
Fixed all the issues in the docs but want to wait with a new commit until I hear back from Alex about the SAN issue.

src/cryptography/x509.py
alex: You can probably just simplify `__init__` to construct the `GeneralNames`, no need for the `isinstance` check, `GeneralNames.__init__` will have that.

etrauschke: Ok, understood. Regarding the test cases, I'd assume we now have to run the
whole set of tests for SAN on the GeneralNames class as well.

Also, could you explain what the benefit of this solution is compared to
subclassing?

alex: https://www.youtube.com/watch?v=3MNVP9-hglc outlines my feelings on the
matter.

On Wed, May 27, 2015 at 11:47 AM, Erik Trauschke <notifications@github.com>
wrote:

> In src/cryptography/x509.py
> <https://github.com/pyca/cryptography/pull/1956#discussion_r31148034>:
>
> > @@ -976,6 +982,11 @@ def __ne__(self, other):
> >          return not self == other
> >
> >
> > +class SubjectAlternativeName(GeneralNames):
>
> Ok, understood. Regarding the test cases, I'd assume we now have to run
> the whole set of tests for SAN on the GeneralNames class as well. Also,
> could you explain what the benefit of this solution is compared to
> subclassing?
> … <#14d960e898f81258_>
> On Wed, May 27, 2015 at 8:15 AM, Alex Gaynor <notifications@github.com>
> wrote: In src/cryptography/x509.py <
> https://github.com/pyca/cryptography/pull/1956#discussion_r31144337>: >
> @@ -976,6 +982,11 @@ def __ne__(self, other): > return not self == other >
> > > +class SubjectAlternativeName(GeneralNames): You can probably just
> simplify __init__ to construct the GeneralNames, no need for the isinstance
> check, GeneralNames.__init__ will have that. — Reply to this email directly
> or view it on GitHub <
> https://github.com/pyca/cryptography/pull/1956/files#r31144337>.
>
> —
> <https://github.com/pyca/cryptography/pull/1956/files#r31148034>.
>



-- 

etrauschke: Thanks. I won't be able to watch this right now but will check it out later.

etrauschke:
Travis issue looks like a timeout in one job. All other tests succeeded so I'd guess this is not an issue with the code itself.

docs/x509.rst
```diff
@@ -335,6 +335,71 @@ X.509 Certificate Object
...
+        A naïve datetime representing when the next update to this CRL is
+        expected.
+
+    .. attribute:: last_update
+
+        :type: :class:`datetime.datetime`
+
+        A naïve datetime representing when the this CRL was last updated.
+
+    .. attribute:: revoked
```
reaperhulk: What do you think about calling this `revoked_certificates` instead of `revoked`?

docs/x509.rst
```diff
@@ -335,6 +335,71 @@ X.509 Certificate Object
...
+
+        The revoked certificates listed in this CRL.
+
+    .. attribute:: extensions
+
+        :type: :class:`Extensions`
+
+        The extensions encoded in the CRL.
+
+    .. method:: verify(pubkey)
```
reaperhulk: Let's call this public_key

docs/x509.rst
```diff
@@ -335,6 +335,71 @@ X.509 Certificate Object
...
+
+        :param algorithm: The
```
reaperhulk: this should be `public_key` instead of `algorithm`.

reaperhulk:
Other than my small nits this LGTM now! I'd like @alex to take a look but if there are no objections I'll merge this tomorrow assuming the final comments are addressed. Thanks @etrauschke!

docs/x509.rst
etrauschke: Can do, I was using the naming in the OpenSSL structs.

etrauschke:
Thanks @reaperhulk. For the final pull request, do you need me to rebase this first?

reaperhulk:
Nope, no rebase required!

docs/x509.rst
```diff
@@ -388,6 +453,30 @@ X.509 CSR (Certificate Signing Request) Object
             over the network to be signed by the certificate
             authority.
 
+X.509 Revoked Certificate Object
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
alex: The number of squigglies should match the length of the previous lline.

src/cryptography/x509.py
```diff
@@ -55,6 +55,9 @@
     "2.5.29.17": "subjectAltName",
     "2.5.29.18": "issuerAltName",
     "2.5.29.19": "basicConstraints",
+    "2.5.29.21": "CRLReason",
+    "2.5.29.24": "InvalidityDate",
+    "2.5.29.29": "CertificateIssuer",
```
alex: I believe these two should start with a lower case character.

src/cryptography/x509.py
```diff
@@ -1175,6 +1207,71 @@ def public_bytes(self, encoding):
...
+    def extensions(self):
+        """
+        Returns an Extensions object containing a list of CRL extensions.
+        """
+
+    @abc.abstractmethod
+    def verify(self, pubkey):
+        """
+        Verify this CRL against a given 'pubkey'.
+        """
```
alex: Can we remove the `verify` method to start with -- so far we haven't added these types of methods anywhere else, and I'd like us to step back and really think about the design before we commit to an API.

alex:
Besides those items, this looks like it's basically there!

src/cryptography/x509.py
etrauschke: I'm ok with removing it from this PR but I'm gonna need this functionality for CRLs and certs. From my perspective there are two reasonable ways to verify an X509, either by supplying the public key or the issuing/intermediate certificate. Since you can get the public key from the issuing cert, I think the pubkey approach is more flexible.
The mechanism above also follows then OpenSSL workflow. Is there any specific concern about this interface?

reaperhulk: Unfortunately to properly verify a certificate or CRL you need to build a trust chain to a root/anchor, validate that none of them are expired or revoked, check that each intermediate/root is allowed to sign certificates, check that each certificate is properly signed, and potentially validate things like SCTs or signature algorithm. Complicating matters is the fact that due to cross signing many potential chains might exist and determining which one is the right one to use is a continuing problem during the 1024-bit root removal and SHA1 deprecation processes.

This process is marginally simpler in the case of CRLs where your chain is typically short, but cross signing path resolution issues can still occur (and you need to validate key usages to determine if a subordinate CRL delegate should be allowed or if it is the same as the certificate itself).

reaperhulk: Oops forgot to finish that whole thought. Ultimately the complexity of `verify` is sufficient that it needs to be its own PR (and indeed we're going to have to go through a pretty complex process of determining whether OpenSSL is internally capable of doing this right. 1.0.2 has some new capabilities in this regard but we can't rely on that being present :(

etrauschke: I think that should be part of a higher level verification process. If you don't allow verification against a single cert this would mean you can never have this work correctly without a working Internet connection for everything where you don't own the CA.
If verify always checks the whole chain, you would need to fetch the CAs and verify that they haven't been revoked or otherwise invalidated every time you want to verify a single cert.
I agree that it would be helpful to have a mechanism which helps you doing the chain verification but it shouldn't be part of every single verification process.
In our projects we do the chain verification ourselves, outside of the OpenSSL frameworks, for instance.

reaperhulk: Yeah, I definitely agree that in hazmat the core verification should not require network egress. The goal of this project is to build APIs that make it hard to shoot yourself in the foot though, so when we add verification here if it ends up being the final check (does this public key match the signature) we may want to think carefully about what we name that.

etrauschke: Yes, no argument against that here. From the original comment it just
sounded like this sort of verification wasn't considered to be doable at
all.

etrauschke:
Thanks, I'll create a new pr for the backend stuff soon.

PR #1959, opened 2015-05-15
IDNA 2003 test

reaperhulk:
This PR adds a new test vector and a test using it to confirm that IDNA 2003 encoding is disallowed.


tests/test_x509_ext.py
```diff
@@ -1299,6 +1301,17 @@ def test_rfc822name(self, backend):
...
 
+    def test_idna2003_invalid(self, backend):
+        cert = _load_cert(
+            os.path.join(
+                "x509", "custom", "san_idna2003_dnsname.pem"
+            ),
+            x509.load_pem_x509_certificate,
+            backend
+        )
+        with pytest.raises(idna.InvalidCodepoint):
```
alex: The test should match the documented behavior IMO

alex:
Can you rebase this, I think this bug has been fixed! (Probably we'll want to put some version qualifies in for the latest version, so distros don't fuck this up)

PR #1960, opened 2015-05-15
add extensions to the interface for Certificate

reaperhulk:
as noted in #1956 this was missing

### DIFF
--- a/src/cryptography/x509.py
+++ b/src/cryptography/x509.py
@@ -1125,6 +1125,12 @@ class Certificate(object):
         in the certificate.
         """
 
+    @abc.abstractproperty
+    def extensions(self):
+        """
+        Returns an Extensions object.
+        """
+
     @abc.abstractmethod
     def __eq__(self, other):
         """


alex:
Is it documented?

reaperhulk:
Yeah we documented it here: https://cryptography.io/en/latest/x509/#cryptography.x509.Certificate.extensions

src/cryptography/x509.py
```diff
@@ -1125,6 +1125,12 @@ def signature_hash_algorithm(self):
         in the certificate.
         """
 
+    @abc.abstractproperty
+    def extensions(self):
+        """
+        Returns a list of extensions
```
alex: This should be a "sequence" or something, it's not areally a list.

reaperhulk: Should probably just say an Extensions object right? :)

alex: That also works :-)

On Fri, May 15, 2015 at 12:32 PM, Paul Kehrer <notifications@github.com>
wrote:

> In src/cryptography/x509.py
> <https://github.com/pyca/cryptography/pull/1960#discussion_r30423803>:
>
> > @@ -1125,6 +1125,12 @@ def signature_hash_algorithm(self):
> >          in the certificate.
> >          """
> >
> > +    @abc.abstractproperty
> > +    def extensions(self):
> > +        """
> > +        Returns a list of extensions
>
> Should probably just say an Extensions object right? :)
>
> —
> <https://github.com/pyca/cryptography/pull/1960/files#r30423803>.
>



-- 

etrauschke:
Just mentioning that CSRs have extensions too. But I guess we can skip them for now since we don't have a backend implementation for them anyway.

reaperhulk:
Yeah we'll add them to the interface when we implement support for them (coming soon). We actually have a test that verifies interfaces so if we added it now CI would fail :)

PR #1970, opened 2015-05-17
AES keywrap support

reaperhulk:
Proposed replacement for #1794 due to backend incompatibilities.

We should discuss what risks we create by doing this (partially) in Python.

### DIFF
--- a/docs/hazmat/primitives/keywrap.rst
+++ b/docs/hazmat/primitives/keywrap.rst
@@ -15,6 +15,9 @@ protections offered by key wrapping are also offered by using authenticated
 
     .. versionadded:: 1.1
 
+    This function performs AES key wrap (without padding) as specified in
+    :rfc:`3394`.
+
     :param bytes wrapping_key: The wrapping key.
 
     :param bytes key_to_wrap: The key to wrap.
@@ -29,6 +32,9 @@ protections offered by key wrapping are also offered by using authenticated
 .. function:: aes_key_unwrap(wrapping_key, wrapped_key, backend)
 
     .. versionadded:: 1.1
+
+    This function performs AES key unwrap (without padding) as specified in
+    :rfc:`3394`.
 
     :param bytes wrapping_key: The wrapping key.
 
--- a/docs/hazmat/primitives/keywrap.rst
+++ b/docs/hazmat/primitives/keywrap.rst
@@ -1,0 +1,43 @@
+.. hazmat::
+
+.. module:: cryptography.hazmat.primitives.keywrap
+
+Key wrapping
+============
+
+Key wrapping is a cryptographic construct that uses symmetric encryption to
+encapsulate key material.
+
+.. function:: aes_key_wrap(wrapping_key, key_to_wrap, backend)
+
+    :param bytes wrapping_key: The wrapping key.
+
+    :param bytes key_to_wrap: The key to wrap.
+
+    :param backend: A
+        :class:`~cryptography.hazmat.backends.interfaces.CipherBackend`
+        provider that supports
+        :class:`~cryptography.hazmat.primitives.ciphers.algorithms.AES`.
+
+    :return bytes: The wrapped key as bytes.
+
+.. function:: aes_key_unwrap(wrapping_key, wrapped_key, backend)
+
+    :param bytes wrapping_key: The wrapping key.
+
+    :param bytes wrapped_key: The wrapped key.
+
+    :param backend: A
+        :class:`~cryptography.hazmat.backends.interfaces.CipherBackend`
+        provider that supports
+        :class:`~cryptography.hazmat.primitives.ciphers.algorithms.AES`.
+
+    :return bytes: The unwrapped key as bytes.
+
+Exceptions
+~~~~~~~~~~
+
+.. class:: InvalidUnwrap
+
+    This is raised when a wrapped key fails to unwrap. It can be caused by a
+    corrupted or invalid wrapped key or an invalid wrapping key.
--- a/docs/hazmat/primitives/keywrap.rst
+++ b/docs/hazmat/primitives/keywrap.rst
@@ -10,6 +10,8 @@ encapsulate key material.
 
 .. function:: aes_key_wrap(wrapping_key, key_to_wrap, backend)
 
+    .. versionadded:: 1.1
+
     :param bytes wrapping_key: The wrapping key.
 
     :param bytes key_to_wrap: The key to wrap.
@@ -23,6 +25,8 @@ encapsulate key material.
 
 .. function:: aes_key_unwrap(wrapping_key, wrapped_key, backend)
 
+    .. versionadded:: 1.1
+
     :param bytes wrapping_key: The wrapping key.
 
     :param bytes wrapped_key: The wrapped key.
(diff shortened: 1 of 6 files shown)


simo5:
I have very similar code in Jwcrypto that I would love to dump in favour of cryptography's implementation, is there anything blocking this PR from getting merged that needs work ?

reaperhulk:
I think the primary issue to resolve here is how comfortable we are with a Python implementation that uses the underlying AES primitives. Maybe someone like @lvh would like to weigh in on the advisability of this approach.

simo5:
What's your concern with it ? This is still Hazmat so people need to be careful no matter what.
Once OpenSSL is fixed in a way we can trust it to behave thenw e can simply gut this code and use OpenSSL directly. Do you think you'd have a different interface if it were a backend function exposed ?

lvh:
@simo5 I'm not sure that's what hazmat means. It might mean "you still need to know what you're doing", but it doesn't mean "sometimes and in ways that according to the API are supposed to be imperceptible to you, this feature will have wildly different security characteristics". That doesn't mean it's broken, but it does mean there's something to examine.

As usual, we have to start from threat models, and see how applicable they are in use cases where people are likely to want key wrapping. That's always tricky of course, since you can miss either.

Are there any use cases that we're thinking of besides the usual (fake) HSM story?

Assuming your AES implementations are good (and we are), fortunately this model doesn't have a lot of ways to mess it up. It usually falls apart because of shitty key management, not broken crypto.

simo5:
@lvh my use case is in implementing the various algo's defined in JWA and used in JWE/JWS.

JWA: https://tools.ietf.org/html/rfc7518

reaperhulk:
@lvh as simo mentioned, AESKW is part of the JW{A,E,T,everyotherletter} suite of things now, so we're starting to see cases where it's needed to implement and interoperate with those specifications.

The threat model I'm concerned with in this case is "What additional risks are we taking on by doing part of this algorithm in Python". As best I can tell, the risk is entirely predicated on how comfortable we are with having the intermediate values stored as Python byte strings.

lvh:
Okay, that's a use case :) I agree that that's the extent of the threat model, and I think it's a reasonable risk. It'd be a bit nicer if everything was in C so we didn't have to worry about when/if an object would be collected and/or overwriting that data with NULs; but I'm not sure the threat model that would help with is realistic.

reaperhulk:
Yeah, the frustrating part is that OpenSSL *does* have this as a function, but older versions have a bug and actually do the key wrap wrong. <sad trombone>

simo5:
Just an idea, would it make sense to test openssl's aeswrap at run time on first use with known values and determine what to use going forward based on the result ?
Or is it possible to check the openssl version at runtime and switch based on that ?

reaperhulk:
The best we can do is run one wrap (or unwrap) operation to determine if the underlying library works, but I'm not a big fan of having the code paths fork like that honestly.

lvh:
Yeah, me either. Fortunately, version checking is likely to work reasonably well here; I'm guessing that maybe some old versions will have the fix backported (I'm looking at you, RHEL), but newer versions are unlikely to have the shitty broken version. Using the Python version when the C version would also have worked isn't great but it's not exactly terrible either...

simo5:
That was my thinking too, yeah we probably fixed it in RHEL, but it's not a big deal if we do not use it because it fails a version check. If someone is really keen on it they can patch cryptography to relax the version check or to explicitly single out a specific version number where they know it has been fixed on their platform. Unless they do that they get the python version.

reaperhulk:
If we want to pursue doing that I'd prefer to put that in a separate PR so we can keep the review size on this one reasonable. Here's some of the code that would be required in that subsequent PR: https://github.com/reaperhulk/cryptography/compare/d0f74e1...d75c420

tests/hazmat/primitives/test_keywrap.py
```diff
@@ -0,0 +1,112 @@
...
+    )
+    def test_unwrap_invalid_key_length(self, backend):
+        with pytest.raises(ValueError):
+            keywrap.aes_key_unwrap(b"badkey", b"\x00" * 24, backend)
+
+    @pytest.mark.supported(
+        only_if=lambda backend: backend.cipher_supported(
+            algorithms.AES("\x00" * 16), modes.ECB()
+        ),
+        skip_message="Does not support AES key wrap (RFC 3394)",
```
lvh: For clarity, this checks if AES(-ECB) is supported, which is only equivalent to supporting AES key wrap because we just have a Python implementation now, right? That might be useful to specify in the error message, although I suppose it's unlikely that an implementation supports AES key wrap but not bare AES.

tests/hazmat/primitives/test_keywrap.py
```diff
@@ -0,0 +1,112 @@
...
+from ...utils import load_nist_vectors
+
+
+@pytest.mark.requires_backend_interface(interface=CipherBackend)
+class TestAESKeyWrap(object):
+    @pytest.mark.parametrize(
+        "params",
+        _load_all_params(
+            os.path.join("keywrap", "kwtestvectors"),
+            ["KW_AE_128.txt", "KW_AE_192.txt", "KW_AE_256.txt"],
```
lvh: Cool, I didn't realize these were already in the repo.

lvh:
LGTM modulo nitpick re: skip message which is up to you.

docs/hazmat/primitives/keywrap.rst
```diff
@@ -0,0 +1,50 @@
+.. hazmat::
+
+.. module:: cryptography.hazmat.primitives.keywrap
+
+Key wrapping
+============
+
+Key wrapping is a cryptographic construct that uses symmetric encryption to
+encapsulate key material.
```
alex: This needs a more verbose explanation. Why would anyone want to do that?

src/cryptography/hazmat/primitives/keywrap.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # RFC 3394 Key Wrap - 2.2.1 (index method)
+    encryptor = Cipher(AES(wrapping_key), ECB(), backend).encryptor()
+    a = b"\xa6\xa6\xa6\xa6\xa6\xa6\xa6\xa6"
+    r = [key_to_wrap[i:i + 8] for i in range(0, len(key_to_wrap), 8)]
+    n = len(r)
+    for j in range(6):
+        for i in range(n):
+            # every encryption operation is a discrete 16 byte chunk so
+            # it is safe to reuse the encryptor for the entire operation
```
lvh: This comment should probably point out that this is only true because of ECB mode, and perhaps note that the magic 16 bytes is just AES' block size.

docs/hazmat/primitives/keywrap.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Key wrapping is a cryptographic construct that uses symmetric encryption to
+encapsulate key material. Key wrapping algorithms are occasionally utilized
+to protect keys at rest or transmit them over insecure networks. Many of the
+protections offered by key wrapping are also offered by using authenticated
+:doc:`symmetric encryption </hazmat/primitives/symmetric-encryption>`.
+
+.. function:: aes_key_wrap(wrapping_key, key_to_wrap, backend)
+
+    .. versionadded:: 1.1
```
lvh: This should mention which kind of key wrapping we're doing.

lvh:
LGTM!

PR #1973, opened 2015-05-18
[WIP] ECPrivateKey from private_value

esneider:
This PR adds the class method `EllipticCurvePrivateNumbers.from_private_value_and_curve` to be able to build an `ECPrivateKey` from a given private value.

This functionality is needed to implement some private key encodings specific to Bitcoin (e.g. WIF).

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -1042,6 +1042,7 @@ class Backend(object):
 
         ec_cdata = self._lib.EC_KEY_new_by_curve_name(curve_nid)
         assert ec_cdata != self._ffi.NULL
+        ec_cdata = self._ffi.gc(ec_cdata, self._lib.EC_KEY_free)
 
         set_func, get_func, group = (
             self._ec_key_determine_group_get_set_funcs(ec_cdata)
@@ -1049,8 +1050,10 @@ class Backend(object):
 
         point = self._lib.EC_POINT_new(group)
         assert point != self._ffi.NULL
+        point = self._ffi.gc(point, self._lib.EC_POINT_free)
 
         value = self._int_to_bn(private_value)
+        value = self._ffi.gc(value, self._lib.BN_free)
 
         with self._tmp_bn_ctx() as bn_ctx:
             res = self._lib.EC_POINT_mul(group, point, value, self._ffi.NULL,
@@ -1065,10 +1068,6 @@ class Backend(object):
 
             point_x = self._bn_to_int(bn_x)
             point_y = self._bn_to_int(bn_y)
-
-        self._lib.BN_free(value)
-        self._lib.EC_POINT_free(point)
-        self._lib.EC_KEY_free(ec_cdata)
 
         return point_x, point_y
 
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -1037,6 +1037,41 @@ class Backend(object):
 
         return _EllipticCurvePublicKey(self, ec_cdata)
 
+    def derive_elliptic_curve_public_point(self, private_value, curve):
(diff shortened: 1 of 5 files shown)


reaperhulk:
jenkins, ok to test

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1037,6 +1037,41 @@ def load_elliptic_curve_public_numbers(self, numbers):
 
         return _EllipticCurvePublicKey(self, ec_cdata)
 
+    def derive_elliptic_curve_public_point(self, private_value, curve):
+        curve_nid = self._elliptic_curve_to_nid(curve)
+
+        ec_cdata = self._lib.EC_KEY_new_by_curve_name(curve_nid)
```
public: You should use `ffi.gc` on this instead of manually freeing it.

public:
Needs some tests. Maybe loading all the private keys from the NIST vectors and trying to generate their public keys?

reaperhulk:
Do we use classmethods anywhere else in our codebase? I don't think we do, so it may not be the best choice here just for consistency reasons.

public:
I think normally we use a module scope function for stuff like this yes.

esneider:
@public not being familiar with the codebase, I didn't want to write tests before getting feedback on the API: https://mail.python.org/pipermail/cryptography-dev/2015-May/000464.html

@reaperhulk true, what else do you suggest? Can you point me to a similar factory method/function in the codebase? I was following @alex sugestion, which seemed reasonable: https://mail.python.org/pipermail/cryptography-dev/2015-May/000461.html

src/cryptography/hazmat/backends/openssl/backend.py
esneider: Sure!

reaperhulk:
@esneider I apologize for keeping you waiting on this. We have historically used module scope for support functions (as well as loader functions). One example of this is `rsa_recover_prime_factors` (https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/primitives/asymmetric/rsa.py#L209). While slightly more verbose, this might work well with an approach like this:

```python


public_numbers = ec.derive_from_private_value_and_curve(private_value, curve)
```

One challenge here is that the above may actually require a backend (your current implementation does). Is there a sane way to do this in Python? Typically we're not overly worried about timing variability or performance in calculations like this (as they should be unusual), although if there's a reason why we should be let us know!

esneider:
@reaperhulk I could use python to do it, but there's some issues with that:

* The public point is computed by doing a scalar point multiplication d*G, where d is the private value and G the curve generator. Right now only the backend knows G for a given curve (it's a curve constant). It would need to be hardcoded in the python code for each supported curve, which might or might not be a good decision.

* While the scalar point multiplication in an elliptic curve field isn't that hard, it isn't trivial either. For example, in warner/python-ecdsa it's almost the entire Point class: add, mul, and double methods in  https://github.com/warner/python-ecdsa/blob/master/ecdsa/ellipticcurve.py#L63; plus the inverse_mod function in https://github.com/warner/python-ecdsa/blob/master/ecdsa/numbertheory.py#L185.

After considering this I went with the OpenSSL implementation, not because of performance, but to avoid writing all that (almost) crypto code. What do you think?

reaperhulk:
@esneider I think I'm convinced. Let's move forward with this implementation then. I think docs and tests are next :)

reaperhulk:
Closing due to lack of activity. @esneider if you still want this you can reopen. It looks like docs and tests were our primary holdup to moving forward with this previously.

ofek:
@reaperhulk What would be the best way to move forward with this? I can do tests and docs.

reaperhulk:
If you'd like to open a new PR using this as a base, then yeah, we need tests and docs. Given the length of time since we closed this PR there might also be some discussion about whether this is the approach we still want (I'm not sure -- I could see it go either way).

PR #1977, opened 2015-05-18
Retry brew update if it fails

alex:


reaperhulk:
we should look into travis_retry at some point, but lgtm.

PR #1983, opened 2015-05-19
put the 'version' field into the 'SSL' struct where it belongs; make SSL_CTX opaque.

glyph:
Currently cryptography improperly defines the `ssl_st` structure as `SSL_CTX`; this is inaccurate.  The typedef lines from `ossl_typ.h` are:

```c
typedef struct ssl_st SSL;
typedef struct ssl_ctx_st SSL_CTX;
```


reaperhulk:
add to whitelist

reaperhulk:
This was a bug introduced in this PR:

https://github.com/pyca/cryptography/commit/c4d9141ea0725bf0b893b23d9a48a1cf23455331

PR #1985, opened 2015-05-20
Docs for CTR and GCM should say that IV/nonce must be unique

### LINKED ISSUE #1971: CTR and GCM docs say nonce/iv must be random
As far as I'm aware, they actually just need to be unique, and can happily be a counter (for example).

eeshangarg:
Fixes #1971.

Hi!

The reason I escaped the whitespace character after ```**n**``` is that otherwise we would get a Sphinx build warning as explained [here] (http://blog.yjl.im/2012/02/restructuredtext-inline-markup-on.html).

Let me know what you think!

Thanks & regards,
Eeshan Garg

### DIFF
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -64,3 +64,11 @@ Glossary
     text
         This type corresponds to ``unicode`` on Python 2 and ``str`` on Python
         3.  This is equivalent to ``six.text_type``.
+
+    nonce
+        A nonce is a **n**\ umber used **once**. Nonces are used in many
+        cryptographic protocols. Generally, a nonce does not have to be secret
+        or unpredictable, but it must be unique. A nonce is often a random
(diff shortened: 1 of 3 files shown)


docs/glossary.rst
```diff
@@ -64,3 +64,11 @@ Glossary
     text
         This type corresponds to ``unicode`` on Python 2 and ``str`` on Python
         3.  This is equivalent to ``six.text_type``.
+
+    nonce
+        A nonce is a **n**\ umber used **once**. Nonces are used in many
```
reaperhulk: Looks like we need to add "Nonces" (with that capitalization) to the docs/spelling_wordlist.txt so the spellchecker understands that it is a word.

eeshangarg: I made the change. Let me know if this PR needs further work! Thanks! :-)

PR #1986, opened 2015-05-21
cffi 1.0

reaperhulk:
This is a very early patch. @dstufft I'd appreciate pointers on how to make this less crazy. Specifically I'm baffled by where these build scripts should live (and why building creates two copies of the shared objects -- one in the right place and another in the root of the repository).

The tests will also fail right now because we attempt to monkeypatch to do elliptic curve supported checks (and also pypy doesn't have cffi 1.0).


reaperhulk:
Oh, I also removed way too much in setup.py but I don't know exactly what's needed so let me know :)

setup.py
```diff
@@ -75,219 +69,6 @@ def cc_is_available():
...
-    need the equivalent for the ``install`` command to install into platlib
-    install-dir rather than purelib.
-    """
-
-    def finalize_options(self):
-        self.distribution.ext_modules = get_ext_modules()
-        install.finalize_options(self)
-
-
-class PyTest(test):
```
alex: This should probabyl stay?

reaperhulk: Yeah, definitely. It got removed as I was trying to make this all work. Hopefully @dstufft has some guidance on adding it back properly (do we need keywords with side effects still? If not, presumably we need parts of the cmd_class dict)

reaperhulk:
I've updated this PR to correct most issues. Three outstanding:

* When installing via `pip install -e` c/o/so files appear in the root of the directory as well as the so files in src/cryptography. This may be a cffi bug or a config mistake.
* We need to think about our coverage because the install code doesn't run for coverage now, which means many modules are not covered during test runs now.
* pypy will fail until 2.6.0 is released

src/cryptography/hazmat/primitives/padding.py
```diff
@@ -6,24 +6,12 @@
 
 import abc
 
-import os
+from cryptography._cryptography_padding import lib
```
alex: This import should come below six.

src/_cffi_src/utils.py
```diff
@@ -0,0 +1,84 @@
...
+    ffi = FFI()
+    ffi.cdef(cdef_source)
+    ffi.set_source(
+        so_name,
+        verify_source,
+        libraries=libraries,
+        extra_compile_args=extra_compile_args,
+        extra_link_args=extra_link_args,
+    )
+    ffi.compile()
```
dstufft: This is going to compile the module whenever the modules get imported and should just be deleted. This call also removes the ``.c`` and stuff in the ``.`` and makes everything go much faster.

dstufft: If we want to support, building the modules outside of the build system, we should add a ``if __name__ == "__main__": ffi.compile()`` to the ffi build scripts.

dstufft:
This needs to add ``_cffi_src`` and ``_cffi_src.*`` to the exclude list to the ``find_packages`` call in ``setup.py`` otherwise we're installing a ``_cffi_src`` module that is importable.

This combined with the ``ffi.compile()`` called could be the reason why you were getting weird linker errors. We would have recompiled the ``.so`` files (I think) when pytest went importing modules to try and discover test things.

dstufft:
For some reason we're still getting ``.c`` files generated inside of the ``cryptography/hazmat/bindings/__pycache__/`` directory. It's my understanding that cffi 1.0 doesn't need these so I'm not sure why we're getting them. They are also being included in the wheel file and the sdist file which is less than optimal.

Here's a list of them:
```
  1356609  02-19-15 22:36   cryptography/hazmat/bindings/__pycache__/_Cryptography_cffi_6bea6352xd5968d73.c
    11166  02-19-15 22:36   cryptography/hazmat/bindings/__pycache__/_Cryptography_cffi_a8febd48xffc7b1ce.c
   159801  02-19-15 22:36   cryptography/hazmat/bindings/__pycache__/_Cryptography_cffi_db50b3f8xb3816a41.c
  1354808  02-12-15 21:38   cryptography/hazmat/bindings/__pycache__/_Cryptography_cffi_dceb3353xf1961971.c
    11209  02-19-15 22:36   cryptography/hazmat/bindings/__pycache__/_Cryptography_cffi_dd416c1exc1767c5a.c
```

src/_cffi_src/build_openssl.py
```diff
@@ -0,0 +1,80 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+import sys
+
+from _cffi_src.utils import (
+    build_ffi_for_binding
+)
+
+from cryptography.hazmat.bindings.utils import _get_openssl_libraries
```
dstufft: The only place this function is called is from within this build script. This is also the only place that ``_cffi_src`` imports ``cryptography``. I think we should just define this function inside of this file and remove any dependency that ``_cffi_src`` has on ``cryptography``.

src/cryptography/hazmat/bindings/commoncrypto/binding.py
```diff
@@ -4,54 +4,19 @@
...
 from __future__ import absolute_import, division, print_function
 
-import threading
-
-from cryptography.hazmat.bindings.utils import (
-    build_ffi_for_binding, load_library_for_binding,
-)
+try:
+    from cryptography._cryptography_commoncrypto import ffi, lib
+except ImportError:
```
dstufft: Do we need this? Can we just not support importing ``cryptography.hazmat.bindings.commoncrypto`` if we're not on a platform that has CommonCrypto?

dstufft:
It feels kinda silly to me to have the ``build_XXX.py`` modules define a ``_build()`` function that is only ever called once... immediately after they are defined at module scope. Is there any reason not to just deindent all that code one level and just get rid of the ``_build()`` function?

reaperhulk:
I've put up a new version. This addresses most of the feedback but has two major issues:

* For some reason find_packages is not excluding _cffi_src still
* There are dozens of .c files being created (and 7 additional shared objects?!) for no apparent reason.

I still need to look at the commoncrypto feedback and figure out coverage after moving `_get_openssl_libraries` into _cffi_src as well.

dstufft:
You probably need to adjust ``MANIFEST.in`` so that instead of:

```
recursive-include src/_cffi_src/hazmat_src *.c *.h
recursive-include src/_cffi_src/openssl/src *.c *.h
```

You have

```
recursive-include src/_cffi_src *.py *.c *.h
```

reaperhulk:
jenkins, retest this please

alex:
Can I ask why the decision to move all the code for creating bindings out of the `cryptography/` package?

dstufft:
It's not part of the installed package, it's a build script.

reaperhulk:
jenkins, retest this please

reaperhulk:
jenkins, build the docs please

reaperhulk:
build the docs please

reaperhulk:
jenkins, build the docs please

reaperhulk:
jenkins retest this please

reaperhulk:
Okay I think this is ready for final review. jenkins will fail until we upgrade pypy on 10.8 and the openssl102 builders so when we're happy with this PR we should upgrade those builders, retest this, then merge.

src/_cffi_src/build_commoncrypto.py
```diff
@@ -0,0 +1,32 @@
...
+    "common_hmac",
+    "common_key_derivation",
+    "common_cryptor",
+    "common_symmetric_key_wrap",
+    "secimport",
+    "secitem",
+    "seckey",
+    "seckeychain",
+    "sectransform",
+]
```
alex: This is a bit silly, but can you move these values into the `build_ffi_for_binding` call so it looks a bit more declaraitive, and less imperative.

src/_cffi_src/build_commoncrypto.py
```diff
@@ -0,0 +1,32 @@
...
+
+ffi = build_ffi_for_binding(
+    so_name="_cryptography_commoncrypto",
```
alex: Is there a reason not to have the generated modules live inside the cryptography package?

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -4,136 +4,30 @@
...
-    elif link_type == "static" or link_type == "":
-        return ["libeay32mt", "ssleay32mt", "advapi32",
-                "crypt32", "gdi32", "user32", "ws2_32"]
-    else:
-        raise ValueError(
-            "PYCA_WINDOWS_LINK_TYPE must be 'static' or 'dynamic'"
-        )
+from cryptography._cryptography_openssl import ffi, lib
+
+res = lib.Cryptography_add_osrandom_engine()
```
alex: Is there a reason not to put this, and the lock initialization in the `Bindings` constructor, closer to what we do now?

tests/hazmat/bindings/test_commoncrypto.py
```diff
@@ -7,9 +7,13 @@
...
+try:
+    from cryptography.hazmat.bindings.commoncrypto.binding import Binding
+except ImportError:
+    Binding = None
 
 
+@pytest.mark.skipif(Binding is None, reason="CommonCrypto not available")
 @pytest.mark.skipif("commoncrypto" not in
                     [i.name for i in _available_backends()],
                     reason="CommonCrypto not available")
```
alex: Do we need both of these skips?

alex: We could also use `pytest.importorskip` here.

alex:
Could we leave the declerations of the bindings in their current places so that this diff is smaller, or is that impractical?

alex:
I note that right now we lose all coverage of the build scripts, because it's outside `cryptography/`

src/_cffi_src/build_commoncrypto.py
reaperhulk: They do. Right now to import this module you do `from cryptography._cryptography_commoncrypto import ffi, lib`. Arguably _cryptography becomes an unneeded piece. It was added partially to disambiguate in case the so doesn't get installed to the right place for <setuptools/distutils/whatever> reason.

alex: Yeah, the double cryptography seems wrong. I also feel like these should be
in hazmat/bindings/

On Sun, Jun 7, 2015 at 9:57 AM, Paul Kehrer <notifications@github.com>
wrote:

> In src/_cffi_src/build_commoncrypto.py
> <https://github.com/pyca/cryptography/pull/1986#discussion_r31877016>:
>
> > +_modules = [
> > +    "cf",
> > +    "common_digest",
> > +    "common_hmac",
> > +    "common_key_derivation",
> > +    "common_cryptor",
> > +    "common_symmetric_key_wrap",
> > +    "secimport",
> > +    "secitem",
> > +    "seckey",
> > +    "seckeychain",
> > +    "sectransform",
> > +]
> > +
> > +ffi = build_ffi_for_binding(
> > +    so_name="_cryptography_commoncrypto",
>
> They do. Right now to import this module you do from
> cryptography._cryptography_commoncrypto import ffi, lib. Arguably
> _cryptography becomes an unneeded piece. It was added partially to
> disambiguate in case the so doesn't get installed to the right place for
> reason.
>
> —
> <https://github.com/pyca/cryptography/pull/1986/files#r31877016>.
>



-- 

src/cryptography/hazmat/bindings/openssl/binding.py
reaperhulk: OpenSSL lock initialization and osrandom engine addition need to happen once and only once. Previously we accomplished this with locks, but it seems much cleaner to do it directly on import. Is there another approach you'd prefer?

alex: I think I prefer locks. Doing things at module import time is pretty gross
to me.

On Sun, Jun 7, 2015 at 10:00 AM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/1986#discussion_r31877036>:
>
> > -        return _get_windows_libraries(link_type)
> > -
> > -
> > -def _get_windows_libraries(link_type):
> > -    if link_type == "dynamic":
> > -        return ["libeay32", "ssleay32", "advapi32"]
> > -    elif link_type == "static" or link_type == "":
> > -        return ["libeay32mt", "ssleay32mt", "advapi32",
> > -                "crypt32", "gdi32", "user32", "ws2_32"]
> > -    else:
> > -        raise ValueError(
> > -            "PYCA_WINDOWS_LINK_TYPE must be 'static' or 'dynamic'"
> > -        )
> > +from cryptography._cryptography_openssl import ffi, lib
> > +
> > +res = lib.Cryptography_add_osrandom_engine()
>
> OpenSSL lock initialization and osrandom engine addition need to happen
> once and only once. Previously we accomplished this with locks, but it
> seems much cleaner to do it directly on import. Is there another approach
> you'd prefer?
>
> —
> <https://github.com/pyca/cryptography/pull/1986/files#r31877036>.
>



-- 

reaperhulk:
If we leave the binding declarations in place we'll drop significantly from 100% coverage because they are now only run/imported during the build process and not during tests (or any other import of the library past installation).

I think @dstufft has some pretty good thoughts on coverage with build scripts.

src/cryptography/hazmat/bindings/openssl/binding.py
reaperhulk: If we ever want to get rid of the `Binding` object (and I think we probably should since it doesn't really need to exist now) we'll eventually have to transition all this class global state to module global state anyway. I'm going to revert this to simplify review for now though. We can discuss deprecation/removal of `Binding` in the future.

dstufft:
Right, I don't think it makes sense to try and cover your build scripts, these are in essence no different than ``setup.py`` itself, scripts that get ran to build the project. We test them indirectly by testing the artifacts they produce, not the thing that produces the artifacts.

dstufft:
I cannot accurately describe how happy this PR makes me in simple text!

reaperhulk:
jenkins, retest this please

CHANGELOG.rst
```diff
@@ -6,6 +6,9 @@ Changelog
 
 .. note:: This version is not yet released and is under active development.
 
+* Switched to the new `cffi`_ ``set_source`` out-of-line API mode for
+  compilation. This results in significantly faster imports and lowered
+  memory consumption.
```
dstufft: I hate to trigger another test run... but I feel like this should mention that this means we no longer support PyPy releases older than 2.6 nor do we support any released version of PyPy3 and won't until one is released that supports CFFI 1.0+.

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -4,136 +4,40 @@
...
+            if not cls._lib_loaded:
+                cls._lib_loaded = True
                 res = cls.lib.Cryptography_add_osrandom_engine()
                 assert res != 0
 
     @classmethod
     def init_static_locks(cls):
         with cls._lock_init_lock:
-            cls._ensure_ffi_initialized()
-
```
dstufft: I'm not sure I know why this was here to begin with, but why are we removing it? Our ``_ensure_ffi_intialized()`` is still doing things.

PR #1989, opened 2015-05-25
Fix dual license in test_dh.py

elitest:
I saw #1988 fly by and did some grepping and noticed that this file also had the APL2 license only.  I assume that it will want to be fixed.

Cheers.
Jim


reaperhulk:
Thanks! We definitely do want this fixed. :)

PR #1990, opened 2015-05-25
Add "get_provisioning_uri" utility for HOTP/TOTP.

tonyseek:
Please review it. Thanks.

### DIFF
--- a/docs/hazmat/primitives/twofactor.rst
+++ b/docs/hazmat/primitives/twofactor.rst
@@ -193,9 +193,8 @@ Provisioning URI
 ~~~~~~~~~~~~~~~~
 
 The provisioning URI of HOTP and TOTP is not actual the part of RFC 4226 and
-RFC 6238, but a `spec of Google Authenticator <http://git.io/vkvvY>`_. It is
-widely supported by web sites and mobile applications which are using
-Two-Factor authentication.
+RFC 6238, but a `spec of Google Authenticator`_. It is widely supported by web
+sites and mobile applications which are using Two-Factor authentication.
 
 For generating a provisioning URI, you could use the ``get_provisioning_uri``
 method of HOTP/TOTP instances.
@@ -211,3 +210,5 @@ method of HOTP/TOTP instances.
 
 A common usage is encoding the provisioning URI into QR code and guiding users
 to scan it with Two-Factor authentication applications in their mobile devices.
+
+.. _`spec of Google Authenticator`: https://github.com/google/google-authenticator/wiki/Key-Uri-Format
--- a/docs/hazmat/primitives/twofactor.rst
+++ b/docs/hazmat/primitives/twofactor.rst
@@ -74,6 +74,14 @@ codes (HMAC).
         :raises cryptography.hazmat.primitives.twofactor.InvalidToken: This
              is raised when the supplied HOTP does not match the expected HOTP.
 
+    .. method:: get_provisioning_uri(account_name, counter, issuer=None)
+
+        :param str account_name: The display name of account, such as
+            ``'Alice Smith'`` or ``'alice@example.com'``.
+        :param str issuer: The optional display name of issuer.
+        :param int counter: The current value of counter.
+        :return str: An URI string.
+
 Throttling
 ~~~~~~~~~~
 
@@ -171,3 +179,33 @@ similar to the following code.
         :param int time: The time value to validate against.
         :raises cryptography.hazmat.primitives.twofactor.InvalidToken: This
              is raised when the supplied TOTP does not match the expected TOTP.
+
+    .. method:: get_provisioning_uri(account_name, issuer=None)
+
+        :param str account_name: The display name of account, such as
+            ``'Alice Smith'`` or ``'alice@example.com'``.
+        :param str issuer: The optional display name of issuer.
+        :return str: An URI string.
+
+Provisioning URI
+~~~~~~~~~~~~~~~~
+
+The provisioning URI of HOTP and TOTP is not actual the part of RFC 4226 and
+RFC 6238, but a `spec of Google Authenticator <http://git.io/vkvvY>`_. It is
+wide supported by web sites and mobile applications which using Two-Factor
+authentication.
+
+For generating a provisioning URI, you could use the ``get_provisioning_uri``
+method of HOTP/TOTP instances.
+
+.. code-block:: python
+
+    counter = 5
+    account_name = 'alice@example.com'
+    issuer_name = 'Example Inc'
+
+    hotp_uri = hotp.get_provisioning_uri(account_name, counter, issuer_name)
+    totp_uri = totp.get_provisioning_uri(account_name, issuer_name)
+
+A common usage is encoding the provisioning URI into QR code and guiding users
+to scan it with Two-Factor authentication applications in their mobile devices.
--- a/docs/hazmat/primitives/twofactor.rst
+++ b/docs/hazmat/primitives/twofactor.rst
@@ -74,11 +74,12 @@ codes (HMAC).
         :raises cryptography.hazmat.primitives.twofactor.InvalidToken: This
              is raised when the supplied HOTP does not match the expected HOTP.
 
-    .. method:: get_provisioning_uri(account_name, counter, issuer=None)
+    .. method:: get_provisioning_uri(account_name, counter, issuer)
 
         :param str account_name: The display name of account, such as
             ``'Alice Smith'`` or ``'alice@example.com'``.
-        :param str issuer: The optional display name of issuer.
+        :param issuer: The optional display name of issuer.
+        :type issuer: `string` or `None`
         :param int counter: The current value of counter.
         :return str: An URI string.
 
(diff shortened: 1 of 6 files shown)


Ayrx:
I personally do not think that this is something we should have in the library since it's a Google Authenticator specific thing (although it's probably pretty widely used at this point) and not actually part of the HOTP / TOTP standard.

alex:
Jenkins, ok to test

alex:
I'm not sure where I stand on if we should implement this, but if so, a few notes on this patch right off the bat:

* Tests are currently failing
* This should be done using a utility function shared by other HOTP and TOTP, not a mixin

tonyseek:
@Ayrx @alex Thanks for your review. I modified it into an utility function instead of mixin class. And the broken tests have been fixed.

I know that the provisioning URI is not a part of the RFC4226/RFC6238. But it is a way to let end users be able to configure two-factor authentication with their mobile devices. Personally I hope it could be included in cryptography or exists as a third-party library, not implemented many times in different applications. ;-)

reaperhulk:
jenkins, retest this please

reaperhulk:
@tonyseek do you happen to know of other places where this is used outside of Google Authenticator? I'm not opposed to adding this if it has become a de facto standard.

tonyseek:
@reaperhulk Following mobile apps have support for provisioning with QR code which included this form of URI:

- [Google Authenticator](https://github.com/google/google-authenticator)
- [Duo Security](https://www.duosecurity.com)
- [Authy](https://www.authy.com) (TOTP only)

And many sites with 2-factor authentication are providing this kind of QR code, such as Google, Dropbox, GitHub, Zoho and Slack.

reaperhulk:
Given the widespread use of this (at least as encoded in QR codes) this doesn't seem like a bad thing to add. Any other opinions? /cc @alex, @public, @ayrx (do you still feel it doesn't belong?)

src/cryptography/hazmat/primitives/twofactor/utils.py
```diff
@@ -0,0 +1,50 @@
...
+import base64
+
+from six.moves.urllib.parse import quote, urlencode
+
+
+__all__ = ['get_provisioning_uri']
+
+
+def get_provisioning_uri(otp, account_name, issuer=None, counter=None):
+    """Generates a provisioning URI which can be recognized by Two-Factor
```
reaperhulk: Assuming we decide to pull this in we put documentation separately in restructured text. You'll need to add this to the docs there.

src/cryptography/hazmat/primitives/twofactor/utils.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import unicode_literals
+
+import base64
+
+from six.moves.urllib.parse import quote, urlencode
+
+
+__all__ = ['get_provisioning_uri']
```
reaperhulk: This isn't needed (and we probably don't want to encourage `import *`)

alex: I'm ok adding support for this. I still have some feedback on the substance
of the overall patch.

On Wed, May 27, 2015 at 12:55 PM, Paul Kehrer <notifications@github.com>
wrote:

> In src/cryptography/hazmat/primitives/twofactor/utils.py
> <https://github.com/pyca/cryptography/pull/1990#discussion_r31155244>:
>
> > @@ -0,0 +1,50 @@
> > +from __future__ import unicode_literals
> > +
> > +import base64
> > +
> > +from six.moves.urllib.parse import quote, urlencode
> > +
> > +
> > +__all__ = ['get_provisioning_uri']
>
> This isn't needed (and we probably don't want to encourage import *)
>
> —
> <https://github.com/pyca/cryptography/pull/1990/files#r31155244>.
>



-- 

Ayrx:
I don't feel strongly about not having this so I'm fine with this going in.
:)

I'll do a code review when I get to a computer unless someone gets to it
first.

src/cryptography/hazmat/primitives/twofactor/utils.py
Ayrx: Agreed that this isn't needed. Should be removed.

src/cryptography/hazmat/primitives/twofactor/utils.py
```diff
@@ -0,0 +1,50 @@
...
+                :class:`cryptography.hazmat.primitives.twofactor.hotp.HOTP` or
+                :class:`cryptography.hazmat.primitives.twofactor.totp.TOTP`.
+    :param account_name: The display name of account, such as
+                         ``'Alice Smith'`` or ``'alice@example.com'``.
+    :param issuer: The display name of issuer.
+    :param counter: The current value of counter. It is required for HOTP.
+    :return: The URI string.
+    :raises RuntimeError: if counter is missing but otp type is HOTP
+    """
+    hotp = getattr(otp, '_hotp', otp)
```
Ayrx: If I'm reading this right you are attempting to determine if the otp object is an instance of HOTP and not TOTP? If so this should be rewritten as `isinstance(otp, HOTP)` instead.

Ayrx:
Overall API design feedback: I don't like that the public API is a standalone function. In particular it leads to this very ugly bit of code:

```python
if hotp is otp:
    if counter is None:
        raise RuntimeError('"counter" is required for HOTP')
   parameters.append(('counter', int(counter)))
```

Is there anything particularly bad about each class (HOTP and TOTP), having a `generate_url` method with different parameters? Common code can be factored out into a private function but the consumers of our library do not have to concern themselves with that.

reaperhulk:
:+1: on @Ayrx's comments. @tonyseek could you update the patch with `get_provisioning_uri` becoming a method on each class?

tonyseek:
Hi @Ayrx @reaperhulk @alex ,

I turned the `get_provisioning_uri` into methods of HOTP and TOTP. It is cleaner without the `if hotp is otp`. Thank you for your review. :smile: 

The description part has been moved into the `rst` docs too.

src/cryptography/hazmat/primitives/twofactor/utils.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import unicode_literals
```
reaperhulk: For files we have a licensing preamble and a specific __future__ import we want: https://cryptography.io/en/latest/development/submitting-patches/#code

src/cryptography/hazmat/primitives/twofactor/utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+import base64
+
+from six.moves.urllib.parse import quote, urlencode
+
+
+def generate_uri(hotp, type_name, account_name, issuer, extra_parameters):
```
reaperhulk: This utility function is meant for consumption by the existing classes rather than by an end user so I'd prefer it to be `_generate_uri`

src/cryptography/hazmat/primitives/twofactor/totp.py
```diff
@@ -31,3 +32,8 @@ def generate(self, time):
     def verify(self, totp, time):
         if not constant_time.bytes_eq(self.generate(time), totp):
             raise InvalidToken("Supplied TOTP value does not match.")
+
+    def get_provisioning_uri(self, account_name, issuer=None):
```
reaperhulk: In general we tend to avoid providing defaults for arguments. I think I'd prefer this just be `issuer` and then document that it's string or None. Do you have an opinion @alex?

reaperhulk:
We're getting close now, thanks for your work so far :)

tonyseek:
@reaperhulk Updated :-)

Ayrx:
@tonyseek Still missing the standard license header.

docs/hazmat/primitives/twofactor.rst
```diff
@@ -171,3 +180,34 @@ similar to the following code.
...
+        :param issuer: The optional display name of issuer.
+        :type issuer: `string` or `None`
+        :return str: An URI string.
+
+Provisioning URI
+~~~~~~~~~~~~~~~~
+
+The provisioning URI of HOTP and TOTP is not actual the part of RFC 4226 and
+RFC 6238, but a `spec of Google Authenticator <http://git.io/vkvvY>`_. It is
+wide supported by web sites and mobile applications which using Two-Factor
```
Ayrx: "widely supported", "which are using".

Ayrx:
Other than the standard license header and the small grammar fixes this LGTM. :+1:

tonyseek:
@Ayrx Fixed~

docs/hazmat/primitives/twofactor.rst
```diff
@@ -171,3 +180,34 @@ similar to the following code.
...
+            ``'Alice Smith'`` or ``'alice@example.com'``.
+        :param issuer: The optional display name of issuer.
+        :type issuer: `string` or `None`
+        :return str: An URI string.
+
+Provisioning URI
+~~~~~~~~~~~~~~~~
+
+The provisioning URI of HOTP and TOTP is not actual the part of RFC 4226 and
+RFC 6238, but a `spec of Google Authenticator <http://git.io/vkvvY>`_. It is
```
reaperhulk: Our typical model for external links is to put them at the bottom of the page like:

```
.. _`spec of Google Authenticator`: https://github.com/google/google-authenticator/wiki/Key-Uri-Format
```

Additionally we strongly prefer not to use shortener links where possible so the canonical URL is preferred.

reaperhulk:
Other than that final nit this LGTM.

tonyseek:
@reaperhulk The URL have been unshorten and moved to the page bottom.

reaperhulk:
Thank you @tonyseek! You're welcome to submit a PR to add yourself to the AUTHORS.rst file if you'd like. If you get a chance it'd also be great if you could submit a separate PR to add an entry to the CHANGELOG.rst for this new feature.

tonyseek:
@reaperhulk I will submit PR soon. Thank you!

PR #1991, opened 2015-05-26
print the version of openssl loaded by cryptography in tox

reaperhulk:
This is sometimes useful, especially in CI when debugging.


reaperhulk:
retest this please

PR #1996, opened 2015-05-29
enforce GCM AAD and ciphertext limits per NIST spec

### LINKED ISSUE #1821: AES-GCM Exception Encrypting Large Files

reaperhulk:
fixes #1821 

This is a simplified approach to properly handle NIST limits for AAD and encrypted bytes.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/ciphers.py
+++ b/src/cryptography/hazmat/backends/openssl/ciphers.py
@@ -18,6 +18,8 @@ class _CipherContext(object):
     _DECRYPT = 0
 
     def __init__(self, backend, cipher, mode, operation):
+        self._bytes_processed = 0
+        self._aad_bytes_processed = 0
         self._backend = backend
         self._cipher = cipher
         self._mode = mode
@@ -110,6 +112,17 @@ class _CipherContext(object):
         if len(data) == 0 and not isinstance(self._mode, modes.GCM):
             return b""
 
+        self._bytes_processed += len(data)
+        if (
+            isinstance(self._mode, modes.GCM) and
+            self._bytes_processed > modes.GCM._MAX_ENCRYPTED_BYTES
+        ):
+            raise ValueError(
+                "GCM has a maximum encrypted byte limit of {0}".format(
+                    modes.GCM._MAX_ENCRYPTED_BYTES
+                )
+            )
+
         buf = self._backend._ffi.new("unsigned char[]",
                                      len(data) + self._block_size - 1)
         outlen = self._backend._ffi.new("int *")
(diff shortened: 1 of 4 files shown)


src/cryptography/hazmat/primitives/ciphers/modes.py
```diff
@@ -139,6 +139,10 @@ def validate_for_algorithm(self, algorithm):
 @utils.register_interface(ModeWithAuthenticationTag)
 class GCM(object):
     name = "GCM"
+    # This is (2 ** 39) - 256 bits (see NIST GCM spec)
+    _MAX_ENCRYPTED_BYTES = 68719476704
+    # This is 2 ** 64 bits (see NIST GCM spec)
+    _MAX_AAD_BYTES = 2305843009213693952
```
alex: Any reason not to just write the constants as expressions, so we can save the comment?

alex:
I'm still not wild about this approach to the tests. Where did we ever land on writing a normal test, but only running it on one builder?

reaperhulk:
We can't really do that for the AAD limit since 2 ** 64 is a rather large number.

I've modified the tests to be more explicit in checking that the byte counts are incrementing properly prior to the exception which (hopefully) mitigates the risk here.

src/cryptography/hazmat/backends/openssl/ciphers.py
```diff
@@ -110,6 +112,17 @@ def update(self, data):
         if len(data) == 0 and not isinstance(self._mode, modes.GCM):
             return b""
 
+        self._bytes_processed += len(data)
+        if (
+            isinstance(self._mode, modes.GCM) and
```
public: It's kind of ugly that we need to add all these special if's just for GCM. I'd be interested in seeing a solution to decoupling that. I guess we already do it in a few places here though.

reaperhulk: #1844 and #1847 were more generic solutions than this, but they ended up being higher complexity than we wanted. This issue has now been outstanding for 3 months, so I'm a bit frustrated that, despite it being a user bug report, we still don't have an answer. :(

alex:
Fixed by #2093.

PR #1999, opened 2015-05-29
OpenSSL backend code for CRL interfaces

etrauschke:
This is the OpenSSL backend code for the CRL objects.


etrauschke:
Apparently there are some issues here with older (0.9.8) versions of OpenSSL which don't seem to support some of the extensions. I'll look into this next week.
Do you have some sort of #ifdef mechanism for these things?

PR #2010, opened 2015-06-06
don't double free DSA cdata when verifying things

reaperhulk:
This is already registered for GC. We should not register it again (@alex tracked this bug down, credit to him).


alex:
This LGTM. Someone else needs to review this because I wrote the patch :-)

We also need to establish if the double free actually occurs on any platforms besides PyPy 2.6.0, in which case we need to backport and issue a security release.

PR #2011, opened 2015-06-06
list package versions in tox env for easier debugging

reaperhulk:


alex:
I don't object to landing this, but really shouldn't stuff like this be in tox itself?

reaperhulk:
Maybe? Something like `print_package_versions=True` in [testenv]?

alex:
No idea, or like `tox run-some-random-commands-in-that-venv "pip freeze"`

PR #2012, opened 2015-06-06
remove pypy3 from testing until a cffi 1.0 version is available

reaperhulk:
This is required by #1986 

We should re-add pypy3 ASAP.


alex:
I think this should just be a part of #1986 -- it's very small and not independently meaningful.

(Yes, I"m doing an inverse "please split the patch")

reaperhulk:
closing to ad to #1986

PR #2025, opened 2015-06-11
0.9.2 changelog

reaperhulk:
do not merge until 1.0.2b is available from slproweb


alex:
This will need to be updated for 1.0.2c

PR #2030, opened 2015-06-13
Add flag to disable compilation of commoncrypto on OS X.

### LINKED ISSUE #1924: cannot build on OSX 10.10.3 with gcc 4.9.2
We get 

    In file included from /System/Library/Frameworks/Security.framework/Headers/SecDigestTransform.h:28:0,
                     from src/cryptography/hazmat/bindings/__pycache__/_Cryptography_cffi_2a871178xb3816a41.c:236:
    /System/Library/Frameworks/Security.framework/Headers/SecTransform.h:579:15: error: expected identifier or '(' before '^' token
     typedef void (^SecMessageBlock)(CFTypeRef message, CFErrorRef error, 
                   ^
    /System/Library/Frameworks/Security.framework/Headers/SecTransform.h:612:8: error: unknown type name 'SecMessageBlock'
            SecMessageBlock deliveryBlock) 
            ^
    error: command 'gcc' failed with exit status 1

Does it mean one is out of luck here? For it appears that the OSX header is clang-only...

Ayrx:
Fixed #1924.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,6 +91,26 @@ matrix:
         - language: generic
           os: osx
           osx_image: beta-xcode6.3
+          env: TOXENV=py26 CC=gcc
+        - language: generic
+          os: osx
+          osx_image: beta-xcode6.3
+          env: TOXENV=py27 CC=gcc
+        - language: generic
+          os: osx
+          osx_image: beta-xcode6.3
+          env: TOXENV=py33 CC=gcc
+        - language: generic
+          os: osx
+          osx_image: beta-xcode6.3
+          env: TOXENV=py34 CC=gcc
+        - language: generic
+          os: osx
+          osx_image: beta-xcode6.3
+          env: TOXENV=pypy CC=gcc
+        - language: generic
+          os: osx
+          osx_image: beta-xcode6.3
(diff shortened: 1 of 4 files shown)


Ayrx:
Let's see what our builders say. :)

reaperhulk:
I think I'd prefer testing this on a single builder since this is an edge case anyway and I don't want to increase the runtime of our tests unnecessarily.

Ayrx:
That's fair. Do you want to set it up on Jenkins or should I keep it on
Travis?

alex:
I'm not wild about adhoc flags like this; it really seems like a packaging story what better captures optional backends would be better.

Ayrx:
Fair enough, @reaperhulk do you agree? If so we can close this one. :)

reaperhulk:
I agree long term, but optional backends are still largely a figment of our imagination and I'm unconvinced we'll have them this year. I'm okay with not doing this at all, but we should close #1924 as something we're not going to deal with until optional backends are real. That's probably not a big deal since the workaround is very simple.

Ayrx:
It seems like the consensus is that we shouldn't add arbitrary flags to get around our packaging deficiencies and that this issue should be solved by getting optional backends working. Closing.

PR #2033, opened 2015-06-17
Enforce text type of NameAttribute.value

sigmavirus24:


reaperhulk:
This is going to fail in py2 tests as quite a few tests are failing to initialize using unicode values. We'll need to update the tests.

reaperhulk:
jenkins, add to whitelist

alex:
@sigmavirus24 can you add a changelog entry for this

sigmavirus24:
Changelog updated.

PR #2034, opened 2015-06-17
enable ASLR and NX on windows builds (affects 2.x only)

### LINKED ISSUE #2021: ASLR and NX support on windows
Currently the .pyd files created via cffi do not enable ASLR and NX security features on windows (like most python extensions and python 2. itself due to the outdated VS2008 toolchain used, which didn't have those as default yet). 
A good side effect of ASLR is reduced memory usage if a DLL is loaded a lot of times into different processes and would be rebased from a collision otherwise (as the build doesn't use a non standard baseaddress either currently).

The .pyd files should be built with /DYNAMICBASE and /NXCOMPAT linker flags.

reaperhulk:
An experiment to fix #2021. Not sure if this is the right place to put the args yet.

### DIFF
--- a/src/_cffi_src/build_constant_time.py
+++ b/src/_cffi_src/build_constant_time.py
@@ -5,8 +5,9 @@
 from __future__ import absolute_import, division, print_function
 
 import os
+import sys
 
-from _cffi_src.utils import build_ffi
+from _cffi_src.utils import build_ffi, extra_link_args
 
 
 with open(os.path.join(
@@ -22,5 +23,6 @@ with open(os.path.join(
 ffi = build_ffi(
     module_name="_constant_time",
     cdef_source=types,
-    verify_source=functions
+    verify_source=functions,
+    extra_link_args=extra_link_args(sys.platform),
 )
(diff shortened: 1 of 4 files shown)


reaperhulk:
This is blocked while gathering more info from the reporter in #2021.

alex:
retest this please

reaperhulk:
I have updated this PR and confirmed that the expected dynamic base and NX compatible attributes are present on Windows now.

alex:
`git grep 'NXCOMPAT'` and `git grep 'DYNAMICBASE'` don't have any results in a CPython checkout?

alex:
Ok, that's because this info is in some XML. That said, https://msdn.microsoft.com/en-us/library/ms235442.aspx and https://msdn.microsoft.com/en-us/library/bb384887.aspx indicate these are on by default?

reaperhulk:
@schlenk understands this far better than I do, but I believe these were not defaulted to on in 2008 (which is the compiler used for 2.6/2.7)

schlenk:
Yes. The default is ON since Visual Studio 2010, OFF for Visual Studio 2008. So all the Python 2.7 modules don't have it, unless explicitly specified.

Issue #2007, opened 2015-06-04
osrandom_engine should validate the fd's st_dev and st_ino


tiran:
CPython keeps a persistent fd to /dev/urandom, too. It also keeps the fd's st_dev and st_ino around and validates both before it reads from the fd. The extra check prevents tampering with the fd.

http://bugs.python.org/issue21207

reaperhulk:
:+1:

reaperhulk:
This issue is no longer relevant since we've landed #2073

PR #2035, opened 2015-06-17
improve osrandom engine

reaperhulk:
Attempts to address #2007 

See also: http://bugs.python.org/issue21207

This will require careful review /cc @tiran

### DIFF
--- a/src/_cffi_src/openssl/src/osrandom_engine.c
+++ b/src/_cffi_src/openssl/src/osrandom_engine.c
@@ -47,24 +47,36 @@ static int osrandom_rand_status(void) {
     }
 }
 #else
-static int urandom_fd = -1;
+static struct {
+    int fd;
+    dev_t st_dev;
+    ino_t st_ino;
+} urandom_cache = { -1 };
 
 static int osrandom_finish(ENGINE *e);
 
 static int osrandom_init(ENGINE *e) {
-    if (urandom_fd > -1) {
+    struct stat st;
+
+    if (urandom_cache.fd > -1) {
         return 1;
     }
-    urandom_fd = open("/dev/urandom", O_RDONLY);
-    if (urandom_fd > -1) {
-        int flags = fcntl(urandom_fd, F_GETFD);
+    urandom_cache.fd = open("/dev/urandom", O_RDONLY);
+    if (urandom_cache.fd > -1) {
+        int flags = fcntl(urandom_cache.fd, F_GETFD);
         if (flags == -1) {
             osrandom_finish(e);
             return 0;
-        } else if (fcntl(urandom_fd, F_SETFD, flags | FD_CLOEXEC) == -1) {
+        } else if (fcntl(urandom_cache.fd, F_SETFD, flags | FD_CLOEXEC) == -1) {
             osrandom_finish(e);
             return 0;
         }
+        if (fstat(urandom_cache.fd, &st)) {
+            osrandom_finish(e);
+            return 0;
+        }
+        urandom_cache.st_dev = st.st_dev;
+        urandom_cache.st_ino = st.st_ino;
         return 1;
     } else {
         return 0;
@@ -73,9 +85,24 @@ static int osrandom_init(ENGINE *e) {
 
 static int osrandom_rand_bytes(unsigned char *buffer, int size) {
     ssize_t n;
+    struct stat st;
+
+    if (fstat(urandom_cache.fd, &st) == -1 ||
+        st.st_dev != urandom_cache.st_dev ||
+        st.st_ino != urandom_cache.st_ino) {
+        /* The fd has changed since we opened it: error out. */
+        ERR_put_error(
+            ERR_LIB_RAND,
+            0,
+            ERR_R_RAND_LIB,
+            "osrandom_engine.py urandom fd changed",
+            0
+        );
+        return 0;
+    }
     while (size > 0) {
         do {
-            n = read(urandom_fd, buffer, (size_t)size);
+            n = read(urandom_cache.fd, buffer, (size_t)size);
         } while (n < 0 && errno == EINTR);
         if (n <= 0) {
             ERR_put_error(
@@ -92,9 +119,11 @@ static int osrandom_rand_bytes(unsigned char *buffer, int size) {
 static int osrandom_finish(ENGINE *e) {
     int n;
     do {
-        n = close(urandom_fd);
+        n = close(urandom_cache.fd);
     } while (n < 0 && errno == EINTR);
-    urandom_fd = -1;
+    urandom_cache.fd = -1;
+    urandom_cache.st_dev = -1;
+    urandom_cache.st_ino = -1;
     if (n < 0) {
         return 0;
     } else {
@@ -103,7 +132,9 @@ static int osrandom_finish(ENGINE *e) {
 }
 
 static int osrandom_rand_status(void) {
-    if (urandom_fd == -1) {
+    if (urandom_cache.fd == -1 ||
+        urandom_cache.st_dev == -1 ||
+        urandom_cache.st_ino == -1) {
         return 0;
     } else {
         return 1;
--- a/src/_cffi_src/openssl/src/osrandom_engine.c
+++ b/src/_cffi_src/openssl/src/osrandom_engine.c
@@ -71,9 +71,17 @@ static int osrandom_init(ENGINE *e) {
             osrandom_finish(e);
             return 0;
         }
-        if (fstat(urandom_cache.fd, &st)) {
-            osrandom_finish(e);
-            return 0;
+        errno = 0;
+        if (fstat(urandom_cache.fd, &st) == -1) {
+            /* As long as fstat failed for a reason that's not a bad file
+             * descriptor, call osrandom_finish to close the fd */
+            if (errno != EBADF) {
+                osrandom_finish(e);
+                return 0;
+            } else {
+                urandom_cache.fd = -1;
+                return 0;
+            }
         }
         urandom_cache.st_dev = st.st_dev;
         urandom_cache.st_ino = st.st_ino;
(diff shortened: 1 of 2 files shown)


src/_cffi_src/openssl/src/osrandom_engine.h
```diff
@@ -3,4 +3,7 @@
 #else
 #include <fcntl.h>
 #include <unistd.h>
+#ifdef HAVE_SYS_STAT_H
```
reaperhulk: This was pulled from the Python patch, but under what circumstances would this not be true but fstat, et al would still be available?

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -47,24 +47,35 @@ static int osrandom_rand_status(void) {
...
         if (flags == -1) {
             osrandom_finish(e);
             return 0;
-        } else if (fcntl(urandom_fd, F_SETFD, flags | FD_CLOEXEC) == -1) {
+        } else if (fcntl(urandom_cache.fd, F_SETFD, flags | FD_CLOEXEC) == -1) {
             osrandom_finish(e);
             return 0;
         }
+        if (fstat(urandom_cache.fd, &st)) {
+            return 0;
```
alex: Does this need `osrandom_finish()`?

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -73,9 +84,22 @@ static int osrandom_init(ENGINE *e) {
...
     ssize_t n;
+    struct stat st;
+
+    if (fstat(urandom_cache.fd, &st)
+        || st.st_dev != urandom_cache.st_dev
+        || st.st_ino != urandom_cache.st_ino) {
+        /* Something changed: forget the cached fd (but don't close it,
+           since it probably points to something important for some
+           third-party code). Re-initialize the engine. */
+        int res = osrandom_init(NULL);
```
alex: This won't work, init bails early if `fd` is set.

src/_cffi_src/openssl/src/osrandom_engine.h
alex: This can be dropped I think.

src/_cffi_src/openssl/src/osrandom_engine.c
alex: Would just erroring out here be more appropriate?

reaperhulk: I'd be okay with erroring out here rather than trying to recover. We could also push an error onto the OpenSSL stack so we might have a chance of figuring out why this is happening in the future when some bad software closes all fds on us.

alex: That sounds reasonable.

On Wed, Jun 17, 2015 at 11:35 AM, Paul Kehrer <notifications@github.com>
wrote:

> In src/_cffi_src/openssl/src/osrandom_engine.c
> <https://github.com/pyca/cryptography/pull/2035#discussion_r32639941>:
>
> > @@ -73,9 +84,22 @@ static int osrandom_init(ENGINE *e) {
> >
> >  static int osrandom_rand_bytes(unsigned char *buffer, int size) {
> >      ssize_t n;
> > +    struct stat st;
> > +
> > +    if (fstat(urandom_cache.fd, &st)
> > +        || st.st_dev != urandom_cache.st_dev
> > +        || st.st_ino != urandom_cache.st_ino) {
> > +        /* Something changed: forget the cached fd (but don't close it,
> > +           since it probably points to something important for some
> > +           third-party code). Re-initialize the engine. */
> > +        int res = osrandom_init(NULL);
>
> I'd be okay with erroring out here rather than trying to recover. We could
> also push an error onto the OpenSSL stack so we might have a chance of
> figuring out why this is happening in the future when some bad software
> closes all fds on us.
>
> —
> <https://github.com/pyca/cryptography/pull/2035/files#r32639941>.
>



-- 

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -73,9 +85,24 @@ static int osrandom_init(ENGINE *e) {
 
 static int osrandom_rand_bytes(unsigned char *buffer, int size) {
     ssize_t n;
+    struct stat st;
+
+    if (fstat(urandom_cache.fd, &st)
```
alex: Please write this as `== -1`, also please put the `||` on the line above, not below.

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -47,24 +47,36 @@ static int osrandom_rand_status(void) {
...
+        int flags = fcntl(urandom_cache.fd, F_GETFD);
         if (flags == -1) {
             osrandom_finish(e);
             return 0;
-        } else if (fcntl(urandom_fd, F_SETFD, flags | FD_CLOEXEC) == -1) {
+        } else if (fcntl(urandom_cache.fd, F_SETFD, flags | FD_CLOEXEC) == -1) {
             osrandom_finish(e);
             return 0;
         }
+        if (fstat(urandom_cache.fd, &st)) {
```
sigmavirus24: So `fstat` should return -1 (on error) or 0. Do we want to check for -1 explicitly? I know it's "truthy" but, it might be better to be explicit (like we are below on L90).

Also, do we always expect `orandom_finish` to be the "right thing" here? Should we be checking errno just to be safe?

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -47,24 +47,44 @@ static int osrandom_rand_status(void) {
...
+        } else if (fcntl(urandom_cache.fd, F_SETFD, flags | FD_CLOEXEC) == -1) {
             osrandom_finish(e);
             return 0;
         }
+        errno = 0;
+        if (fstat(urandom_cache.fd, &st) == -1) {
+            /* As long as fstat failed for a reason that's not a bad file
+             * descriptor, call osrandom_finish to close the fd */
+            if (errno != EBADF) {
+                osrandom_finish(e);
```
alex: I'm not sure I understand the intent of this logic here?

alex:
Can I just say this is all horrible and we should all use the relevant syscalls here.

src/_cffi_src/openssl/src/osrandom_engine.c
sigmavirus24: I think this is my fault for asking if there were any error cases where we wouldn't want to call `osrandom_finish`.

reaperhulk: Yep, @sigmavirus24 asked if calling osrandom_finish was always the right call and after looking at the fstat error codes it looks like the answer is "holy shit this should never fail, but if it does then you want to close the fd unless the error reason is a bad fd).

reaperhulk:
It really is. Future improvement: use syscalls on Linux where available (is there still no glibc release with this?).

We could use `CCGetRandomBytes` in Yosemite+ on OS X as well. `SecRandomCopyBytes` is available in 10.7+, but we would probably want to confirm that it doesn't need clang. (#1924 indicates that there's some subset of people who would be unhappy with a decision to require clang even in the openssl backend on OS X).

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -47,24 +47,44 @@ static int osrandom_rand_status(void) {
     }
 }
 #else
-static int urandom_fd = -1;
+static struct {
+    int fd;
+    dev_t st_dev;
+    ino_t st_ino;
+} urandom_cache = { -1 };
```
glyph: Shouldn't this be `{ -1, -1, -1 }`?

glyph:
For what it's worth, the implementation looks correct to me.  That said, this seems to be a rough copy/paste of the Python implementation of this functionality.  Which makes me wonder - can't this just *use* Python's implementation of this functionality?

File descriptors are often a surprisingly scarce resource, especially in things like command-line utilities that want to do things in parallel.  Of course on a server you can place some tuning burden on administrators to mess with sysctl, but for things like, let's say, a chat app, you want to be as parsimonious as possible with new FDs.  And of course one file descriptor is not a high price to pay, but if  every support library that does anything with crypto, along with python itself, is caching its own `os.urandom`, that can add up.  My interactive python environment already manages to open 26 file descriptors (I guess this would make it 27) without actually *doing* anything.  (This is bpython + twisted + pyobjc, so not just python itself.)  That's 10% of the total number of file descriptors that the process can have open at all.

So would it be possible, instead, to have the `__init__.py` here stuff a reference to `os.urandom` itself into this engine as a callback before loading openssl?  I realize that this would introduce a little overhead, but on CPython you'd be calling straight into a builtin C function and on pypy things are already fast, so it doesn't seem prohibitive.  This would eliminate the need for file-descriptor juggling and advances in the state of the art would be adopted automatically, as well as eliminating the duplicate FD.  The overhead would also disappear in any situation where you could start making the aforementioned syscalls directly.

glyph:
@alex pointed out on IRC that my update-a-callback thing is unnecessary.  Instead, the `openssl.osrandom_engine` module could `cffi.new` a `RAND_METHOD` and just populate its pointers entirely in Python, no C code required.  A C file would still be the easier way to do the Windows path here, and possibly the advanced syscalls path later (`SecRandomCopyBytes` for example), but for "fallback to `os.urandom`" you don't need to write any C code.

alex:
Windows isn't special here `os.urandom` does the right thing on Windows.

On Fri, Jun 26, 2015 at 10:29 PM, Glyph <notifications@github.com> wrote:

> @alex <https://github.com/alex> pointed out on IRC that my
> update-a-callback thing is unnecessary. Instead, the
> openssl.osrandom_engine module could cffi.new a RAND_METHOD and just
> populate its pointers entirely in Python, no C code required. A C file
> would still be the easier way to do the Windows path here, and possibly the
> advanced syscalls path later (SecRandomCopyBytes for example), but for
> "fallback to os.urandom" you don't need to write any C code.
>
> —
> <https://github.com/pyca/cryptography/pull/2035#issuecomment-115937893>.
>



-- 

reaperhulk:
I'm closing this as the Python callback model is substantially more elegant and removes our need to maintain this.

sigmavirus24:
:+1: Maintain less C code.

PR #2036, opened 2015-06-17
Added a repr() method to x509._Certificate

major:
This should make it easier to get a quick representation of the _Certificate object using print() or pprint().  I considered using something like:

    cert.subject.get_attributes_for_oid(x509.OID_COMMON_NAME)

But I wasn't sure if every certificate would have data in the commonName OID field.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -159,6 +159,9 @@ class _Certificate(object):
         self._backend = backend
         self._x509 = x509
 
+    def __repr__(self):
+        return "<Certificate(subject={0}, ...)>".format(self.subject)
+
     def __eq__(self, other):
         if not isinstance(other, x509.Certificate):
             return NotImplemented
--- a/tests/test_x509.py
+++ b/tests/test_x509.py
@@ -450,6 +450,39 @@ class TestRSACertificate(object):
         serialized = cert.public_bytes(encoding)
         assert serialized == cert_bytes
 
+    def test_certificate_repr(self, backend):
+        cert = _load_cert(
+            os.path.join(
+                "x509", "cryptography.io.pem"
+            ),
+            x509.load_pem_x509_certificate,
+            backend
+        )
+        if six.PY3:
+            assert repr(cert) == (
+                "<Certificate(subject=<Name([<NameAttribute(oid=<ObjectIdentif"
+                "ier(oid=2.5.4.11, name=organizationalUnitName)>, value='GT487"
+                "42965')>, <NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.11, "
+                "name=organizationalUnitName)>, value='See www.rapidssl.com/re"
+                "sources/cps (c)14')>, <NameAttribute(oid=<ObjectIdentifier(oi"
+                "d=2.5.4.11, name=organizationalUnitName)>, value='Domain Cont"
+                "rol Validated - RapidSSL(R)')>, <NameAttribute(oid=<ObjectIde"
+                "ntifier(oid=2.5.4.3, name=commonName)>, value='www.cryptograp"
+                "hy.io')>])>, ...)>"
+            )
+        else:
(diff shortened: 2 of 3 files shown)


reaperhulk:
Ok to test

reaperhulk:
I'll look at this more closely later today, but I think as a first pass it should use format and not % and just be Certificate instead of X509._Certificate

reaperhulk:
Jenkins, add to whitelist

major:
Good catch on those adjustments.  Here's the output of __repr__() now with one of my favorite sites as an example:

    <X509.Certificate subject:<Name([<NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.11, name=organizationalUnitName)>, value=u'Domain Control Validated')>, <NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.11, name=organizationalUnitName)>, value=u'PositiveSSL')>, <NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.3, name=commonName)>, value=u'icanhazip.com')>])>>

reaperhulk:
One more change to make it more consistent with our other reprs if you don't mind: <Certificate(subject=datahere)>

I have a bad net connection at the moment so ill just say be sure to put tests in. The repr test will need to use six.PY3 (see other tests in test_x509*) because output will be diff between py2/3

alex:
Probably it should be `<Certificate(subject=, ...)>`, (note the elippsis) or something else to indicate the repr is only showing part of the data?

major:
That sounds reasonable, @alex. I'll try to make that adjustments and figure out how to write a test for it. ;)

major:
This is my first time writing tests like these. If it fails, I'll go back in and fix it later this afternoon.

reaperhulk:
No failure :) You can reduce the amount of lines quite a bit by using a different test vector though. I'd recommend using `cryptography.io.pem` instead of the all names vector. You can load it like this:

```python
        cert = _load_cert(
            os.path.join("x509", "cryptography.io.pem"),
            x509.load_pem_x509_certificate,
            backend
        )
```

Looks like you still need to update the repr to look like `<Certificate(subject=, ...)>` per @alex's comment as well.

Thanks for your work on this!

major:
Well darn. That will teach me to write stuff early in the morning from an airport without coffee. ;)

major:
Hopefully this gets everything fixed.

reaperhulk:
@major this looks good with the exception of @alex's request that you add an ellipsis to the repr to show that it's not a complete repr (e.g. `<Certificate(subject=<somedata>, ...>`. If you can push that small change up we can get this merged :)

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -159,6 +159,9 @@ def __init__(self, backend, x509):
         self._backend = backend
         self._x509 = x509
 
+    def __repr__(self):
+        return "<Certificate(subject={0})>".format(self.subject)
```
alex: One last small thing, can you make this: `subject={0}, ...` (with a literal ellipsis), that way it's clear there's other properties

tests/test_x509.py
```diff
@@ -450,6 +450,40 @@ def test_public_bytes_match(self, cert_path, loader_func, encoding,
...
+                "42965')>, <NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.11, "
+                "name=organizationalUnitName)>, value='See www.rapidssl.com/re"
+                "sources/cps (c)14')>, <NameAttribute(oid=<ObjectIdentifier(oi"
+                "d=2.5.4.11, name=organizationalUnitName)>, value='Domain Cont"
+                "rol Validated - RapidSSL(R)')>, <NameAttribute(oid=<ObjectIde"
+                "ntifier(oid=2.5.4.3, name=commonName)>, value='www.cryptograp"
+                "hy.io')>])>, ...)>"
+            )
+        else:
+            print(cert)
```
alex: Doh, outstanding print.

PR #2038, opened 2015-06-18
Add additional X509_STORE functions

sholsapp:
I'm working on https://github.com/pyca/pyopenssl/issues/256 and can't
seem to get the X509_STORE_CTX objects to be able to find a CRL. I
suspect it is because I'm not adding the CRL to the underlying
X509_STORE object, so I'd like to add them.


alex:

PR #2044, opened 2015-06-18
Remove ext_package from setup.py.

Ayrx:
Looking at the CFFI 1.0 docs I don't think this is actually needed anymore.


Ayrx:
Looks like I was wrong. :)

Issue #1231, opened 2014-07-08
openssl backend specific keys should take an evp_pkey.


dreid:
For RSA at least we appear to take an evp_pkey, convert it to rsa_cdata, then _RSAPrivateKey makes a new evp_pkey from that.

Seems silly.

Will be easiest to refactor after the most recent deprecations are removed.

reaperhulk:
I'm going to reopen this now that we're using EVP_PKEY for serialization in EC (and soon DSA). see #1709

PR #2045, opened 2015-06-18
Adds CSR Builder (Redux of #1927)

sigmavirus24:
There's a lot of history in #1927 so go read that for review comments and such before looking over here :wink:

### DIFF
--- a/src/cryptography/hazmat/backends/interfaces.py
+++ b/src/cryptography/hazmat/backends/interfaces.py
@@ -277,7 +277,7 @@ class X509Backend(object):
     @abc.abstractmethod
     def create_x509_csr(self, builder, private_key, algorithm):
         """
-        Create and sign an X.509 CSR from a CSR buidler object.
+        Create and sign an X.509 CSR from a CSR builder object.
         """
 
 
--- a/src/cryptography/hazmat/backends/interfaces.py
+++ b/src/cryptography/hazmat/backends/interfaces.py
@@ -274,6 +274,12 @@ class X509Backend(object):
         Load an X.509 CSR from PEM encoded data.
         """
 
+    @abc.abstractmethod
+    def create_x509_csr(self, builder, private_key, algorithm):
+        """
+        Create and sign an X.509 CSR from a CSR buidler object.
+        """
+
 
 @six.add_metaclass(abc.ABCMeta)
 class DHBackend(object):
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -80,23 +80,16 @@ def _encode_asn1_str(backend, x, n):
 
 
 def _encode_name(backend, attributes):
-    resolve = {
-        x509.OID_COMMON_NAME: b'CN',
-        x509.OID_COUNTRY_NAME: b'C',
-        x509.OID_STATE_OR_PROVINCE_NAME: b'ST',
-        x509.OID_LOCALITY_NAME: b'L',
-        x509.OID_ORGANIZATION_NAME: b'O',
-        x509.OID_ORGANIZATIONAL_UNIT_NAME: b'OU',
-    }
     subject = backend._lib.X509_NAME_new()
     subject = backend._ffi.gc(subject, backend._lib.X509_NAME_free)
     for attribute in attributes:
         value = attribute.value
         if isinstance(value, six.text_type):
             value = value.encode('ascii')
-        res = backend._lib.X509_NAME_add_entry_by_txt(
+        obj = _txt2obj(backend, attribute.oid.dotted_string)
+        res = backend._lib.X509_NAME_add_entry_by_OBJ(
             subject,
-            resolve[attribute.oid],
+            obj,
             backend._lib.MBSTRING_ASC,
             value,
             -1, -1, 0,
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -90,12 +90,12 @@ def _encode_name(backend, attributes):
     subject = backend._lib.X509_NAME_new()
     subject = backend._ffi.gc(subject, backend._lib.X509_NAME_free)
     for attribute in attributes:
-        value = attribute.value.encode('ascii')
+        value = attribute.value.encode('utf8')
         obj = _txt2obj(backend, attribute.oid.dotted_string)
         res = backend._lib.X509_NAME_add_entry_by_OBJ(
             subject,
             obj,
-            backend._lib.MBSTRING_ASC,
+            backend._lib.MBSTRING_UTF8,
             value,
             -1, -1, 0,
         )
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -10,7 +10,7 @@ from contextlib import contextmanager
 
 import six
 
-from cryptography import utils
+from cryptography import utils, x509
 from cryptography.exceptions import (
     InternalError, UnsupportedAlgorithm, _Reasons
 )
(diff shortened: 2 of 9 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -686,6 +776,68 @@ def cmac_algorithm_supported(self, algorithm):
     def create_cmac_ctx(self, algorithm):
         return _CMACContext(self, algorithm)
 
+    def create_x509_csr(self, builder, private_key, algorithm):
+        # TODO: check type of private key parameter.
```
reaperhulk: Not every private_key type is guaranteed to have an `_evp_pkey` right now (and, in fact, only RSA does). There has been past discussion in #1231 about having all our OpenSSL asym objects take evp_pkey, so maybe we should do that so this doesn't need to know anything about type. Alternately this will need to check type and create an evp_pkey for DSA/EC.

Either way, we'll need at least one CSR generation for EC and DSA to confirm this works (which right now it won't).

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -56,6 +56,96 @@
...
+def _encode_name(backend, attributes):
+    subject = backend._lib.X509_NAME_new()
+    subject = backend._ffi.gc(subject, backend._lib.X509_NAME_free)
+    for attribute in attributes:
+        value = attribute.value.encode('ascii')
+        obj = _txt2obj(backend, attribute.oid.dotted_string)
+        res = backend._lib.X509_NAME_add_entry_by_OBJ(
+            subject,
+            obj,
+            backend._lib.MBSTRING_ASC,
```
reaperhulk: Pulling the discussion from #1927 over to here, we need to do some experimentation to determine whether this should be MBSTRING_UTF8 (and whether value should therefore be encoded utf8 rather than ascii).

src/cryptography/hazmat/backends/openssl/backend.py
sigmavirus24: Should we add xfail tests for EC and DSA so when that's implemented, we'll have tests that will pass without doing extra work?

reaperhulk: I wouldn't even xfail them. Let them fail CI since that will remind us that this PR isn't ready yet (and we can rebase/restart builds after the evp_pkey improvement merges)

src/cryptography/hazmat/backends/openssl/backend.py
AndreLouisCaron: @reaperhulk Would that have an impact on compatibility with older software?  Should we instead try to encode it to ASCII and then fall back to UTF-8 if there are non-ASCII characters?

src/cryptography/hazmat/backends/openssl/backend.py
AndreLouisCaron: Are you saying this PR should wait until another refactoring occurs or should we add the tests to make the EC and DSA cases work in this PR first?

reaperhulk: We should add tests for EC and DSA. Those tests will fail until the evp_pkey change is merged though.

src/cryptography/hazmat/backends/openssl/backend.py
alex: Is there specific software known to fail with UTF8?

src/cryptography/hazmat/backends/openssl/backend.py
alex: I'm also open to having some branches here, based on private key type. Which do you feel is better @reaperhulk ?

src/cryptography/x509.py
```diff
@@ -1375,3 +1375,43 @@ def extensions(self):
...
+
+    def add_extension(self, extension, critical=False):
+        """
+        Adds an X.509 extension to the certificate request.
+        """
+        if isinstance(extension, BasicConstraints):
+            extension = Extension(OID_BASIC_CONSTRAINTS, critical, extension)
+        else:
+            raise ValueError('Unsupported X.509 extension.')
+        for e in self._extensions:
```
alex: Can you add a `TODO: this is quadratic in the number of extensions` comment. We don't need to fix now, I just want it to be obvious in the ftureu

src/cryptography/x509.py
```diff
@@ -1375,3 +1375,43 @@ def extensions(self):
...
+        return CertificateSigningRequestBuilder(name, self._extensions)
+
+    def add_extension(self, extension, critical=False):
+        """
+        Adds an X.509 extension to the certificate request.
+        """
+        if isinstance(extension, BasicConstraints):
+            extension = Extension(OID_BASIC_CONSTRAINTS, critical, extension)
+        else:
+            raise ValueError('Unsupported X.509 extension.')
```
alex: `NotImplementedError` here please.

src/cryptography/x509.py
```diff
@@ -1375,3 +1375,43 @@ def extensions(self):
...
+        """
+        Sets the certificate requestor's distinguished name.
+        """
+        if not isinstance(name, Name):
+            raise TypeError('Expecting x509.Name object.')
+        if self._subject_name is not None:
+            raise ValueError('The subject name may only be set once.')
+        return CertificateSigningRequestBuilder(name, self._extensions)
+
+    def add_extension(self, extension, critical=False):
```
alex: @reaperhulk how do you feel about having a default on critical?

reaperhulk: I'd prefer we drop the default and require that users explicitly pass the boolean. Our API design has consistently favored explicit over implicit defaults and I don't see a compelling reason to change that for adding an extension (even though False will be far and away the most common value).

alex:
the `_evp_pkey` stuff is done now

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: Now that evp_pkey change is merged we can just add the tests and remove the todo. No branches needed

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: I wouldn't expect compatibility problems, no. PyOpenSSL (among many others) use this already. I also confirmed UTF8 (in my test case an emoji) is properly handled when doing:

```python
    for attribute in attributes:
        value = attribute.value.encode('utf8')
        obj = _txt2obj(backend, attribute.oid.dotted_string)
        res = backend._lib.X509_NAME_add_entry_by_OBJ(
            subject,
            obj,
            backend._lib.MBSTRING_UTF8,
            value,
            -1, -1, 0,
        )
        assert res == 1
```

We should add a test to confirm we can parse unicode back out properly, but it looks like this is the right way forward.

tests/test_x509.py
```diff
@@ -646,6 +647,117 @@ def test_public_bytes_match(self, request_path, loader_func, encoding,
...
+        builder = x509.CertificateSigningRequestBuilder().subject_name(
+            x509.Name([
+                x509.NameAttribute(x509.OID_COUNTRY_NAME, u'US'),
+                x509.NameAttribute(x509.OID_STATE_OR_PROVINCE_NAME, u'Texas'),
+                x509.NameAttribute(x509.OID_LOCALITY_NAME, u'Austin'),
+                x509.NameAttribute(x509.OID_ORGANIZATION_NAME, u'PyCA'),
+                x509.NameAttribute(x509.OID_COMMON_NAME, u'cryptography.io'),
+            ])
+        )
+        with pytest.raises(ValueError):
```
alex: This should be `NotImplementedError`

tests/test_x509.py
```diff
@@ -646,6 +647,117 @@ def test_public_bytes_match(self, request_path, loader_func, encoding,
...
+    def test_add_unsupported_extension(self, backend):
+        builder = x509.CertificateSigningRequestBuilder().subject_name(
+            x509.Name([
+                x509.NameAttribute(x509.OID_COUNTRY_NAME, u'US'),
+                x509.NameAttribute(x509.OID_STATE_OR_PROVINCE_NAME, u'Texas'),
+                x509.NameAttribute(x509.OID_LOCALITY_NAME, u'Austin'),
+                x509.NameAttribute(x509.OID_ORGANIZATION_NAME, u'PyCA'),
+                x509.NameAttribute(x509.OID_COMMON_NAME, u'cryptography.io'),
+            ])
+        )
```
alex: You can drop the subject name from this example

src/cryptography/hazmat/backends/openssl/backend.py
sigmavirus24: So we'd add these tests to `tests/hazmat/backends/test_openssl.py`, correct? We'll need one for each of `RSA`, `EC`, and `DSA`, right?

src/cryptography/hazmat/backends/openssl/backend.py
sigmavirus24: > We should add a test to confirm we can parse unicode back out properly, but it looks like this is the right way forward.

Could you provide some guidance? There aren't presently any tests in `tests/hazmat/backends/test_openssl.py` for `create_x509_csr`, and I don't see anything that parses a request as part of this, but I'm guessing that was added before this PR?

src/cryptography/hazmat/backends/openssl/backend.py
alex: No, tests to make sure all types of keys work should just be in `test_x509` -- these are tests of the common interface, not any one implementation

reaperhulk: Yep, just add decorators for requires_backend_interface so the tests only run if the given backend supports EC or DSA. Then load a DSA/EC key in the test and sign with it.

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: This test can live in test_x509 as well since it should be backend agnostic (we want to make sure any backend we implement properly handles unicode after all). You'll just create a builder and sign a request, then parse the results using `x509.load_pem_x509_csr` and confirm that the subject (which should contain some unicode characters) round tripped successfully.

alex:
So at this point the only two outstanding items are:

- Tests for DSA and EC private keys
- Encoding Names as UTF-8

@reaperhulk do you concur^ ?

Thanks so much to @sigmavirus24 and @AndreLouisCaron for all your work on this!

reaperhulk:
Yep, I'd like to give it one more review once those are addressed but it looks like all comments thus far are covered.

sigmavirus24:
With any luck the tests should now pass.

tests/hazmat/backends/test_openssl.py
```diff
@@ -453,6 +454,28 @@ class FakeAlgorithm(object):
...
+
+        with pytest.raises(NotImplementedError):
+            backend.create_x509_csr(object(), private_key, hashes.SHA1())
+
+    @pytest.mark.skipif(
+        backend._lib.OPENSSL_VERSION_NUMBER >= 0x10001000,
+        reason="Requires an older OpenSSL. Must be < 1.0.1"
+    )
+    def test_unsupported_ec_keys(self):
+        private_key = ec.generate_private_key(ec.SECT283K1(), backend)
```
reaperhulk: This also needs a `_skip_curve_unsupported`

reaperhulk:
I'd like to see a test that encodes a multi-byte unicode char (e.g. u"\U0001F37A") into a CSR, then parses the CSR (via `x509.load_pem_x509_csr`) and confirms it gets the value back out properly.

tests/hazmat/backends/test_openssl.py
reaperhulk: Also, this should probably be secp256r1 as it's a more widely supported curve across the openssl versions we support.

alex:
As a sort of general point, I think the test cases should be pruned down, to only set the attribtues that are relevant to the tests.

That said, I think this is merge-ready. @reaperhulk I'll let you double check.

alex:
Ugh, I lied: https://coveralls.io/builds/2899126 there's a coverage issue

alex:
Missing test cases:

- `CSRBuilder().subject_name(x509.Name()).subject_name(x509.Name())`
- `backend.create_x509_csr` where it raises `NotImplementedError` on unknown extension.
  - This is a place where the logic is handled by the frontend already; perhaps we can move all support OID checks to the backend?

src/cryptography/hazmat/backends/interfaces.py
```diff
@@ -274,6 +274,12 @@ def load_pem_x509_csr(self, data):
         Load an X.509 CSR from PEM encoded data.
         """
 
+    @abc.abstractmethod
+    def create_x509_csr(self, builder, private_key, algorithm):
```
reaperhulk: This needs a quick doc in the backend interfaces.

reaperhulk:
Other than my one comment this LGTM. Awesome work @AndreLouisCaron and @sigmavirus24!

PR #2047, opened 2015-06-20
create evp_pkey in constructor for DSA/EC

reaperhulk:
This relates to #1231 and #2045. It is an incremental step of instantiating a new evp_pkey inside the constructor rather than boiling the ocean of changing all invocations as is proposed in #1231.

We use `evp_pkey` objects for serialization and CSR generation (#2045).

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -388,8 +388,9 @@ class Backend(object):
             rsa_cdata, key_size, bn, self._ffi.NULL
         )
         assert res == 1
+        evp_pkey = self._rsa_cdata_to_evp_pkey(rsa_cdata)
 
-        return _RSAPrivateKey(self, rsa_cdata)
+        return _RSAPrivateKey(self, rsa_cdata, evp_pkey)
 
     def generate_rsa_parameters_supported(self, public_exponent, key_size):
         return (public_exponent >= 3 and public_exponent & 1 != 0 and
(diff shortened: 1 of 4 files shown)


alex:
So, for `_evp_pkey_to_private_key`, this results in going `EVP_PKEY *` -> `DSA *` -> `EVP_PKEY *`

reaperhulk:
@alex yep -- we do this right now in RSA. If we require evp_pkey in all constructors we'll end up adding code to create the object in 6 places (RSA/DSA/EC Public/Private numbers) and I haven't figured out how to do that without a lot of duplication yet. If we could know what kind of `CData` we created we could make a single function to do this, but I don't know if cffi has a public API for providing that info (it is available in the repr).

alex:
Why do you need to know what type of `CData` you need? Can't we just have
`_from_evp_pkey` and `_from_cdata` class methods, which construct the other
value?

On Sat, Jun 20, 2015 at 11:31 AM, Paul Kehrer <notifications@github.com>
wrote:

> @alex <https://github.com/alex> yep -- we do this right now in RSA. If we
> require evp_pkey in all constructors we'll end up adding code to create the
> object in 6 places (RSA/DSA/EC Public/Private numbers) and I haven't
> figured out how to do that without a lot of duplication yet. If we could
> know what kind of CData we created we could make a single function to do
> this, but I don't know if cffi has a public API for providing that info (it
> is available in the repr).
>
> —
> <https://github.com/pyca/cryptography/pull/2047#issuecomment-113782738>.
>



-- 

reaperhulk:
You can do that if we want 3 different methods (`evp_pkey_from_rsa_cdata`, `evp_pkey_from_dsa_cdata`, etc). We have to call the right `_set1_{DSA/RSA/EC}` method against the `evp_pkey` so we need to know the type of the cdata. So we can make 3 diff methods and just call them, we can somehow introspect the cdata to determine type, or we can explicitly pass type and just have an if/elif chain.

alex:
Yes, 3 different methods. The caller will always know which one it wants.

On Sat, Jun 20, 2015 at 11:38 AM, Paul Kehrer <notifications@github.com>
wrote:

> You can do that if we want 3 different methods (evp_pkey_from_rsa_cdata,
> evp_pkey_from_dsa_cdata, etc). We have to call the right
> _set1_{DSA/RSA/EC} method against the evp_pkey so we need to know the
> type of the cdata. So we can make 3 diff methods and just call them, we can
> somehow introspect the cdata to determine type, or we can explicitly pass
> type and just have an if/elif chain.
>
> —
> <https://github.com/pyca/cryptography/pull/2047#issuecomment-113783424>.
>



-- 

reaperhulk:
Okay, this now fully fixes #1231

alex:
lookin good

PR #2051, opened 2015-06-21
Speed up flake8 by skipping two irrelevant directories

alex:


alex:
You can verify the difference by running `flake8` with `-v` and comparing the files checked.

PR #2052, opened 2015-06-21
Add ASN.1 bindings for ASN1_TIME_set_string

sholsapp:
This will help make the API a little less annoying when dealing with
CRL re: https://github.com/pyca/pyopenssl/issues/256

### DIFF
--- a/src/_cffi_src/openssl/asn1.py
+++ b/src/_cffi_src/openssl/asn1.py
@@ -99,6 +99,7 @@ void ASN1_TIME_free(ASN1_TIME *);
 ASN1_GENERALIZEDTIME *ASN1_TIME_to_generalizedtime(ASN1_TIME *,
                                                    ASN1_GENERALIZEDTIME **);
 ASN1_TIME *ASN1_TIME_set(ASN1_TIME *, time_t);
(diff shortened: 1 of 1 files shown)


alex:

sholsapp:
Thanks. Sorry for all the individual PR... the more time I spend with CRL the more things I see that need cryptography bindings. :smile:

alex:
Ugh, I lied. It looks like this was added in OpenSSL 0.9.8k so not all supported platforms have it.

And I don't mind about all the individual PRs, makes them easier to review

sholsapp:
@alex I didn't know about the OpenSSL 0.9.8k bit. Is there a way to handle this in code, or should I find another way to do what I want without *this* particular binding?

alex:
It's possible to make the binding optional, but I don't think that'll acheive the goal you want, since then it'll just be missing

alex:
Basically, if it's possible to implement the feature without this function, that'd be better.

sholsapp:
I'll find another way to do what I want without this particular binding using pyOpenSSL's `OpenSSL.crypto._set_asn1_time` function.

I'm discarding this PR for now.

PR #2066, opened 2015-06-26
fix coverage in CSR builder

### LINKED ISSUE #2065: [BLOCKER] Coverage has dropped blow 100
Please stop landing all patches until this is resolved https://coveralls.io/r/pyca/cryptography

reaperhulk:
fixes #2065

### DIFF
--- a/tests/test_x509.py
+++ b/tests/test_x509.py
@@ -856,6 +856,20 @@ class TestCertificateSigningRequestBuilder(object):
                 critical=False,
             )
 
+    def test_add_unsupported_extension_in_backend(self, backend):
+        private_key = RSA_KEY_2048.private_key(backend)
+        builder = x509.CertificateSigningRequestBuilder()
+        builder = builder.subject_name(
+            x509.Name([
+                x509.NameAttribute(x509.OID_COUNTRY_NAME, u'US'),
+            ])
+        ).add_extension(
+            x509.SubjectAlternativeName([x509.DNSName(u"cryptography.io")]),
+            critical=False,
+        )
+        with pytest.raises(NotImplementedError):
+            builder.sign(backend, private_key, hashes.SHA256())
+
     def test_set_subject_twice(self):
         builder = x509.CertificateSigningRequestBuilder()
         builder = builder.subject_name(
--- a/tests/test_x509.py
+++ b/tests/test_x509.py
@@ -834,7 +834,7 @@ class TestCertificateSigningRequestBuilder(object):
         assert basic_constraints.value.ca is True
         assert basic_constraints.value.path_length == 2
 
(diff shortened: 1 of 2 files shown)


reaperhulk:
Coverage is back to 100% (https://coveralls.io/builds/2916623).

tests/test_x509.py
```diff
@@ -856,6 +856,38 @@ def test_add_unsupported_extension(self, backend):
...
+                x509.NameAttribute(x509.OID_COUNTRY_NAME, u'US'),
+            ]),
+            extensions=[
+                x509.Extension(
+                    x509.OID_AUTHORITY_KEY_IDENTIFIER,
+                    False,
+                    x509.AuthorityKeyIdentifier('keyid', None, None)
+                )
+            ]
+        )
```
alex: Eep, I'm not sure this is the best way to do it. The constructor is not intentended to be a public API

reaperhulk: If you have a better way to trigger the underlying error in the backend I'm happy to change it. If I don't do this I don't have a way to trigger the NotImplementedError in the backend since it will catch it inside the builder before it ever gets that far.

alex: So, a straightforward way would to be to add support for an extension to
the frontend, but not to the backend.

On Fri, Jun 26, 2015 at 7:30 PM, Paul Kehrer <notifications@github.com>
wrote:

> In tests/test_x509.py
> <https://github.com/pyca/cryptography/pull/2066#discussion_r33404877>:
>
> > @@ -856,6 +856,38 @@ def test_add_unsupported_extension(self, backend):
> >                  critical=False,
> >              )
> >
> > +    def test_add_unsupported_extension_via_init(self, backend):
> > +        private_key = RSA_KEY_2048.private_key(backend)
> > +
> > +        builder = x509.CertificateSigningRequestBuilder(
> > +            subject_name=x509.Name([
> > +                x509.NameAttribute(x509.OID_COUNTRY_NAME, u'US'),
> > +            ]),
> > +            extensions=[
> > +                x509.Extension(
> > +                    x509.OID_AUTHORITY_KEY_IDENTIFIER,
> > +                    False,
> > +                    x509.AuthorityKeyIdentifier('keyid', None, None)
> > +                )
> > +            ]
> > +        )
>
> If you have a better way to trigger the underlying error in the backend
> I'm happy to change it. If I don't do this I don't have a way to trigger
> the NotImplementedError in the backend since it will catch it inside the
> builder before it ever gets that far.
>
> —
> <https://github.com/pyca/cryptography/pull/2066/files#r33404877>.
>



-- 

reaperhulk: We'll have to keep changing the test as we add support for new extensions but we'd have to do that anyway. I'll make the change.

PR #2069, opened 2015-06-26
Make the parameter ordering in sign() consistent with other code

alex:


reaperhulk:
This will need a rebase to catch the new test added in #2066

PR #2071, opened 2015-06-27
handle wildcard DNSNames with IDNA.

### LINKED ISSUE #2054: * not allowed in general names
If you attempt to parse a DNS name from a certificate SAN that contains a wildcard (e.g. `*.cryptography.io`) you get an IDNA InvalidCodePoint when it should successfully parse.

reaperhulk:
fixes #2054

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -82,7 +82,20 @@ def _decode_general_names(backend, gns):
 def _decode_general_name(backend, gn):
     if gn.type == backend._lib.GEN_DNS:
         data = backend._ffi.buffer(gn.d.dNSName.data, gn.d.dNSName.length)[:]
-        return x509.DNSName(idna.decode(data))
+        if data.startswith(b"*."):
+            # This is a wildcard name. We need to split on period, remove the
+            # leading wildcard, IDNA decode, then re-add the wildcard
+            # Wildcard characters should always be left-most (RFC 2595
+            # section 2.4).
+            parts = data.split(b".")
+            parts.pop(0)
+            data = u"*." + idna.decode(b".".join(parts))
+        else:
+            # Not a wildcard, decode away. If the string has a * in it anywhere
+            # invalid this will raise an InvalidCodePoint
+            data = idna.decode(data)
+
+        return x509.DNSName(data)
     elif gn.type == backend._lib.GEN_URI:
         data = backend._ffi.buffer(
             gn.d.uniformResourceIdentifier.data,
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -82,7 +82,20 @@ def _decode_general_names(backend, gns):
...
         data = backend._ffi.buffer(gn.d.dNSName.data, gn.d.dNSName.length)[:]
-        return x509.DNSName(idna.decode(data))
+        if data.startswith(b"*."):
+            # This is a wildcard name. We need to split on period, remove the
+            # leading wildcard, IDNA decode, then re-add the wildcard
+            # Wildcard characters should always be left-most (RFC 2595
+            # section 2.4).
+            parts = data.split(b".")
+            parts.pop(0)
+            data = u"*." + idna.decode(b".".join(parts))
```
alex: Is there a reason not to just write this as `idna.decode(parts[2:])`

reaperhulk: No reason other than that it totally didn't occur to me. I assume you mean parts[1:] and drop the pop right?

alex: I mean, drop the split as well.

On Fri, Jun 26, 2015 at 11:25 PM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/2071#discussion_r33409091>:
>
> > @@ -82,7 +82,20 @@ def _decode_general_names(backend, gns):
> >  def _decode_general_name(backend, gn):
> >      if gn.type == backend._lib.GEN_DNS:
> >          data = backend._ffi.buffer(gn.d.dNSName.data, gn.d.dNSName.length)[:]
> > -        return x509.DNSName(idna.decode(data))
> > +        if data.startswith(b"*."):
> > +            # This is a wildcard name. We need to split on period, remove the
> > +            # leading wildcard, IDNA decode, then re-add the wildcard
> > +            # Wildcard characters should always be left-most (RFC 2595
> > +            # section 2.4).
> > +            parts = data.split(b".")
> > +            parts.pop(0)
> > +            data = u"*." + idna.decode(b".".join(parts))
>
> No reason other than that it totally didn't occur to me. I assume you mean
> parts[1:] and drop the pop right?
>
> —
> <https://github.com/pyca/cryptography/pull/2071/files#r33409091>.
>



-- 

alex: Should have been `idna.decode(data[2:])`

reaperhulk: Yeah, that's a better way to do it.

PR #2072, opened 2015-06-27
Fixed #2067 -- raise an error if a CSRbuilder doesn't hav a subject

### LINKED ISSUE #2067: Add tests for what happens when x509.CertificateSignignRequestBuilder.sign is called before subject_name()
Right now on OpenSSL it fails with a random TypeError

alex:


src/cryptography/x509.py
```diff
@@ -1486,4 +1486,6 @@ def sign(self, private_key, algorithm, backend):
         """
         Signs the request using the requestor's private key.
         """
+        if self._subject_name is None:
```
reaperhulk: Should this test `isinstance` to confirm subject name is an `x509.Name`?

alex: That sounds like the responsibility of the `subject_name` method

PR #2073, opened 2015-06-27
Replace C implementation of OS Random engine with Python one that just calls os.urandom

glyph:
This is alternate fix for #2007 that moves file descriptor shenanigans (and C code) out of cryptography entirely.  It is intended to supersede #2035 .


src/_cffi_src/openssl/engine.py
```diff
@@ -11,15 +11,24 @@
...
 static const long Cryptography_HAS_ENGINE_CRYPTODEV;
 
+struct rand_meth_st {
+    void (*seed)(const void *, int);
+    int (*bytes)(unsigned char *, int);
+    void (*cleanup)();
+    void (*add)(const void *, int, double);
+    int (*pseudorand)(unsigned char *, int);
+    int (*status)();
+};
```
alex: This can be written `typedef struct { stuff } RAND_METHOD;` so you don't need to know the name

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -32,10 +37,72 @@ def _ensure_ffi_initialized(cls):
...
+        def osrandom_rand_status():
+            return 1
+
+        @retain
+        @cls.ffi.callback("ENGINE_GEN_INT_FUNC_PTR", error=0)
+        def osrandom_init(engine):
+            return 1
+
+        @retain
+        @cls.ffi.callback("ENGINE_GEN_INT_FUNC_PTR", error=0)
```
alex: There's no need for `error` on things which can't fail

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -32,10 +37,72 @@ def _ensure_ffi_initialized(cls):
...
+        def retain(it):
+            cls._retained.append(it)
+            return it
+        method = cls.ffi.new("RAND_METHOD*")
+        retain(method)
+        method.seed = cls.ffi.NULL
+
+        @retain
+        @cls.ffi.callback("int (*)(unsigned char *, int)", error=0)
+        def osrandom_rand_bytes(buf, size):
```
alex: I believe all these callbacks can just be class methods with the decorator, which will remove the need for a keepalive list.

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -32,10 +37,72 @@ def _ensure_ffi_initialized(cls):
...
+
+        e = cls.lib.ENGINE_new()
+        result = (cls.lib.ENGINE_set_id(e, cls._osrandom_engine_id)
+                  and cls.lib.ENGINE_set_name(e, cls._osrandom_engine_name)
+                  and cls.lib.ENGINE_set_RAND(e, method)
+                  and cls.lib.ENGINE_set_init_function(e, osrandom_init)
+                  and cls.lib.ENGINE_set_finish_function(e, osrandom_finish)
+                  and cls.lib.ENGINE_add(e))
+        if not cls.lib.ENGINE_free(e):
+            return 0
```
alex: While in C we wrote code this way, in Python ca we please write it as a simple series of statements with checks abotu their results afterwords

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -32,10 +37,72 @@ def _ensure_ffi_initialized(cls):
...
+
+        @retain
+        @cls.ffi.callback("ENGINE_GEN_INT_FUNC_PTR", error=0)
+        def osrandom_init(engine):
+            return 1
+
+        @retain
+        @cls.ffi.callback("ENGINE_GEN_INT_FUNC_PTR", error=0)
+        def osrandom_finish(engine):
+            return 1
```
alex: Actually I think finish and init can be totally dropped.

src/cryptography/hazmat/bindings/openssl/binding.py
glyph: If they're class methods with a decorator, the actual bound-class-method object will still be GC'd, hence still requiring some GC trickiness.  I'm open to other suggestions though; I don't particularly like this spelling.

src/cryptography/hazmat/bindings/openssl/binding.py
glyph: Sure, but... why? It will make the code more than twice as long and it will achieve the same effect; do you find it more readable that way for some reason?

src/cryptography/hazmat/bindings/openssl/binding.py
alex: No they won't, you'll end up with a `cdata` instance as a class variable, those dont' become bound methods.

src/cryptography/hazmat/bindings/openssl/binding.py
alex: More readable, and more consistent with our other error handling practices. Also As I've indicated, since init and finish can be removed, it becomes shorter.

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -4,10 +4,69 @@
...
+        result = lib.ENGINE_set_RAND(engine, method)
+        assert result == 1
+        result = lib.ENGINE_add(engine)
+        assert result == 1
+    finally:
+        result = lib.ENGINE_free(engine)
+        assert result == 1
+    looked_up_engine = lib.ENGINE_by_id(_osrandom_engine_id)
+    assert looked_up_engine != ffi.NULL
+    return 1
```
alex: The return values aren't being used here.

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -4,10 +4,69 @@
...
+        assert result == 1
+        result = lib.ENGINE_set_RAND(engine, method)
+        assert result == 1
+        result = lib.ENGINE_add(engine)
+        assert result == 1
+    finally:
+        result = lib.ENGINE_free(engine)
+        assert result == 1
+    looked_up_engine = lib.ENGINE_by_id(_osrandom_engine_id)
+    assert looked_up_engine != ffi.NULL
```
alex: This code didn't exist in the C version?

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -4,10 +4,69 @@
...
+
+@ffi.callback("int (*)(void)")
+def osrandom_rand_status():
+    return 1
+
+
+method = ffi.new(
+    "RAND_METHOD*", dict(bytes=osrandom_rand_bytes,
+                         pseudorand=osrandom_pseudo_rand_bytes,
+                         status=osrandom_rand_status)
```
alex: This shouldn't be a global; you can return it from `_register_osrandom_engine` and place it on `cls`

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -4,10 +4,69 @@
...
 from __future__ import absolute_import, division, print_function
 
+import os
 import threading
 
 from cryptography.hazmat.bindings._openssl import ffi, lib
 
+_osrandom_engine_id = ffi.new("const char[]", b"osrandom")
+_osrandom_engine_name = ffi.new("const char[]", b"osrandom_engine")
+
```
alex: Extra newline here

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -4,10 +4,69 @@
...
+    result = os.urandom(size)
+    signed[0:size] = result
+    return 1
+
+
+@ffi.callback("int (*)(unsigned char *, int)", error=0)
+def osrandom_pseudo_rand_bytes(buf, size):
+    result = osrandom_rand_bytes(buf, size)
+    if result == 0:
+        return -1
```
alex: This line is uncovered

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -4,10 +4,69 @@
...
 import threading
 
 from cryptography.hazmat.bindings._openssl import ffi, lib
 
+_osrandom_engine_id = ffi.new("const char[]", b"osrandom")
+_osrandom_engine_name = ffi.new("const char[]", b"osrandom_engine")
+
+@ffi.callback("int (*)(unsigned char *, int)", error=0)
+def osrandom_rand_bytes(buf, size):
+    signed = ffi.cast("char*", buf)
```
alex: Space before the "*"

src/cryptography/hazmat/bindings/openssl/binding.py
glyph: There's a test which fails without them.

alex: Well, we should update the test. It was written on the assumption of a C API. Now that this code is python we can do something sane, instead of C's bonkers.

src/cryptography/hazmat/bindings/openssl/binding.py
glyph: Why not? The reasons I made it a global were:

1. @reaperhulk tells me the class is going away, so I didn't want to put more stuff on it
2. it was a global in the C code, initialized much the same way
3. `_register_osrandom_engine` already returns an integer telling you whether you registered it already or not, mirroring what the C code did previously, so having it return this would require more changes
4. it's holding references entirely to global things, so it makes sense that it live in the same scope
5. the failure mode here if registration encounters some unexpected error is that it is unused; the failure mode if I require registration to complete on the class before the reference is retained is random pointer corruption; this seemed better

So it still seems much better to me that way.  What are the arguments in *favor* of putting it on the class?

alex: Functions which operate on global state are harder to reason about.

On Sat, Jun 27, 2015 at 9:19 PM, Glyph <notifications@github.com> wrote:

> <https://github.com/pyca/cryptography/pull/2073#discussion_r33417914>:
>
> > +    result = osrandom_rand_bytes(buf, size)
> > +    if result == 0:
> > +        return -1
> > +    else:
> > +        return result
> > +
> > +
> > +@ffi.callback("int (*)(void)")
> > +def osrandom_rand_status():
> > +    return 1
> > +
> > +
> > +method = ffi.new(
> > +    "RAND_METHOD*", dict(bytes=osrandom_rand_bytes,
> > +                         pseudorand=osrandom_pseudo_rand_bytes,
> > +                         status=osrandom_rand_status)
>
> Why not? The reasons I made it a global were:
>
>    1. @reaperhulk <https://github.com/reaperhulk> tells me the class is
>    going away, so I didn't want to put more stuff on it
>    2. it was a global in the C code, initialized much the same way
>    3. _register_osrandom_engine already returns an integer telling you
>    whether you registered it already or not, mirroring what the C code did
>    previously, so having it return this would require more changes
>    4. it's holding references entirely to global things, so it makes
>    sense that it live in the same scope
>    5. the failure mode here if registration encounters some unexpected
>    error is that it is unused; the failure mode if I require registration to
>    complete on the class before the reference is retained is random pointer
>    corruption; this seemed better
>
> So it still seems much better to me that way. What are the arguments in
> *favor* of putting it on the class?
>
> —
> <https://github.com/pyca/cryptography/pull/2073/files#r33417914>.
>



-- 

src/cryptography/hazmat/bindings/openssl/binding.py
glyph: I added this as a way to fail fast if something went wrong.  When I had a typo in `_osrandom_engine_id` at one point the failure mode was completely perplexing, and it would crash when I double-registered the same thing.  But I guess there's no real reason to keep it there, no test that fails without it, so I'll remove it.

src/cryptography/hazmat/bindings/openssl/binding.py
glyph: "something sane" being "raise an exception on the second call"?

alex: sounds reasonable

On Sat, Jun 27, 2015 at 9:25 PM, Glyph <notifications@github.com> wrote:

> <https://github.com/pyca/cryptography/pull/2073#discussion_r33417937>:
>
> > +    engine = lib.ENGINE_new()
> > +    try:
> > +        result = lib.ENGINE_set_id(engine, _osrandom_engine_id)
> > +        assert result == 1
> > +        result = lib.ENGINE_set_name(engine, _osrandom_engine_name)
> > +        assert result == 1
> > +        result = lib.ENGINE_set_RAND(engine, method)
> > +        assert result == 1
> > +        result = lib.ENGINE_add(engine)
> > +        assert result == 1
> > +    finally:
> > +        result = lib.ENGINE_free(engine)
> > +        assert result == 1
> > +    looked_up_engine = lib.ENGINE_by_id(_osrandom_engine_id)
> > +    assert looked_up_engine != ffi.NULL
> > +    return 1
>
> "something sane" being "raise an exception on the second call"?
>
> —
> <https://github.com/pyca/cryptography/pull/2073/files#r33417937>.
>



-- 

src/cryptography/hazmat/bindings/openssl/binding.py
glyph: The state's still global either way though, it's just nested inside a second unnecessary global container.  For example, I could make it a singleton in a sub-sub-sub-package but the additional layers of packages or classes wouldn't make it less global.

src/cryptography/hazmat/bindings/openssl/binding.py
glyph: Any recommendations for what a good exception type would be?

alex: RuntimeError? (Let's just get something down on paper, we can argue over
that detail later)

On Sat, Jun 27, 2015 at 9:29 PM, Glyph <notifications@github.com> wrote:

> <https://github.com/pyca/cryptography/pull/2073#discussion_r33417944>:
>
> > +    engine = lib.ENGINE_new()
> > +    try:
> > +        result = lib.ENGINE_set_id(engine, _osrandom_engine_id)
> > +        assert result == 1
> > +        result = lib.ENGINE_set_name(engine, _osrandom_engine_name)
> > +        assert result == 1
> > +        result = lib.ENGINE_set_RAND(engine, method)
> > +        assert result == 1
> > +        result = lib.ENGINE_add(engine)
> > +        assert result == 1
> > +    finally:
> > +        result = lib.ENGINE_free(engine)
> > +        assert result == 1
> > +    looked_up_engine = lib.ENGINE_by_id(_osrandom_engine_id)
> > +    assert looked_up_engine != ffi.NULL
> > +    return 1
>
> Any recommendations for what a good exception type would be?
>
> —
> <https://github.com/pyca/cryptography/pull/2073/files#r33417944>.
>



-- 

src/cryptography/hazmat/bindings/openssl/binding.py
glyph: So, this is (poorly) mimicking the behavior of the C code, which I'm pretty sure was also never exercised.  I don't understand why the FIPS thing does what it does, and I'm inclined to just make `osrandom_rand_bytes` be `method.pseudorand` as well and just delete this function.  Can someone - @reaperhulk maybe - tell me why that would be a bad idea?

src/cryptography/hazmat/bindings/openssl/binding.py
glyph: Works for me.

src/cryptography/hazmat/bindings/openssl/binding.py
alex: Ugh, I'm reading the documentation for this API. It was clearly designed by somenoe who never actually thought very hard about it: https://www.openssl.org/docs/crypto/RAND_bytes.html .

I think maybe we should jsut change them both to be the same, and have `error=-1`?

glyph: A poorly designed API in OpenSSL? Shock, horror.

Happy to make them be the same; that seems to make the most sense.

alex:
Jenkins, retest this please

reaperhulk:
jenkins retest this please

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -21,10 +35,41 @@ class Binding(object):
...
     @classmethod
+    def _register_osrandom_engine(cls):
+        assert cls.lib.ERR_peek_error() == 0
+        looked_up_engine = cls.lib.ENGINE_by_id(cls._osrandom_engine_id)
+        if looked_up_engine != ffi.NULL:
+            raise RuntimeError("osrandom engine already registered")
+
+        cls.lib.ERR_clear_error()
+
+        engine = cls.lib.ENGINE_new()
```
alex: Throw in an `assert engine != cls.ffi.NULL` (outside of the `try` block)

glyph:
The last commit addresses @alex's last bit of feedback, so I think this is good to go now?

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -4,11 +4,25 @@
 
 from __future__ import absolute_import, division, print_function
 
+import os
 import threading
 
 from cryptography.hazmat.bindings._openssl import ffi, lib
 
 
+@ffi.callback("int (*)(unsigned char *, int)", error=-1)
```
alex: @reaperhulk do we need to call `ERR_put_error` on error here, or will stuff work ok without it?

alex:
Besides that comment this LGTM, @reaperhulk what do you think?

Now that this is more easily worked with (because python), we should file some follow up issues to write tests that stub out `os.urandom` and verify that the results are used correctly.

src/cryptography/hazmat/bindings/openssl/binding.py
reaperhulk: Things will work fine without it, but I'm curious why we're returning -1 here in case of error? 0 is what the old code returned.

alex: https://www.openssl.org/docs/crypto/RAND_bytes.html -1 seems more correct
than 0 (particularly for psueobytes)

On Sun, Jun 28, 2015 at 3:49 PM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/2073#discussion_r33426214>:
>
> >  import threading
> >
> >  from cryptography.hazmat.bindings._openssl import ffi, lib
> >
> >
> > +@ffi.callback("int (*)(unsigned char *, int)", error=-1)
>
> Things will work fine without it, but I'm curious why we're returning -1
> here in case of error? 0 is what the old code returned.
>
> —
> <https://github.com/pyca/cryptography/pull/2073/files#r33426214>.
>



-- 

reaperhulk: -1 is definitely the right response for pseudo (since that's a terrible response code system and we handled it that way in the past), but I'm not convinced it's the right one for RAND_bytes. It's *probably* not a big deal, but I do not know what OpenSSL does when you tell it the random engine doesn't support RAND_bytes as opposed to telling it that it's broken.

alex: Everysingle caller of `RAND_bytes` in `ssl/` uses `<= 0` to check the
status, so it seems to be immaterial.

On Sun, Jun 28, 2015 at 3:58 PM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/2073#discussion_r33426295>:
>
> >  import threading
> >
> >  from cryptography.hazmat.bindings._openssl import ffi, lib
> >
> >
> > +@ffi.callback("int (*)(unsigned char *, int)", error=-1)
>
> -1 is definitely the right response for pseudo (since that's a terrible
> response code system and we handled it that way in the past), but I'm not
> convinced it's the right one for RAND_bytes. It's *probably* not a big
> deal, but I do not know what OpenSSL does when you tell it the random
> engine doesn't support RAND_bytes as opposed to telling it that it's broken.
>
> —
> <https://github.com/pyca/cryptography/pull/2073/files#r33426295>.
>



-- 

reaperhulk:
Since we do have the ability to do some monkeypatching in tests I'd like to see at least one test that verifies the data from os.urandom is being copied into OpenSSL-land properly. It can really just call RAND_bytes directly.

Incidentally, I did a quick crappy test to see if there's any significant speed difference (as this method should result in at least one additional memory copy in addition to calling into Python) and a loop of 10000 fetches of 8192 bytes of randomness took about the same amount of time with both the old and new approach.

alex:
@reaperhulk 8192 is a pretty big read for `/dev/urandom`, if you've got the script handy, could you try with 16 byte reads?

reaperhulk:
100,000 16 byte reads takes ~0.6-0.7s with the C engine code, ~0.9-1.0s with this PR.

reaperhulk:
Here's an even more minimal script that attempts to solely measure the cost of the `RAND_bytes` call (my previous code allocated a new buffer and also pulled out the data after each call)

```python

buf = backend._ffi.new("char[16]")
for _ in range(100000):
    res = backend._lib.RAND_bytes(buf, 16)
    assert res == 1
```

On my machine running `time python testrand.py` (best run of 5)
### This PR
```
real	0m0.848s
user	0m0.567s
sys	0m0.276s
```
Variance between 0.848s and 0.860s

### Master
```
real	0m0.535s
user	0m0.326s
sys	0m0.205s
```
Variance between 0.535s and 0.554s


I don't necessarily think this is going to be a serious problem, just something to be aware of. It could conceivably negatively affect TLS handshake performance? PyOpenSSL won't be affected though because it doesn't activate this engine.

glyph:
I cranked up the iteration count by another 10x, switched to `xrange` instead of `range`, and measured with PyPy.

best of 3:

# This PR
```
real	0m2.193s
user	0m0.731s
sys	0m1.463s
```
Variance between 2.292 and 2.193

# Master

```
real	0m1.857s
user	0m0.407s
sys	0m1.447s
```
Variance between 1.857 and 1.945

It looks to me like we're pretty clearly just measuring function call / buffer copy overhead for your Python runtime here, and pypy does commensurately better.  How much random data does one TLS handshake require?  My suspicion is that this amount of additional overhead (on pypy, approximately 3µs/call - and it is per *call*, not per byte, as the overhead all but disappears with fewer calls generating the same number of bytes) will verge on undetectable.  Just ballpark-wise, http://www.semicomplete.com/blog/geekery/ssl-latency.html (from 2010) puts TLS handshake latency at 322000µs, so we'd have to be calling `RAND_bytes` 1000 times in the course of the handshake to cause a 1% degradation to performance.

alex:
Yeah, I think the performance implications of this are totally 100% fine :-)

glyph:
That last commit should address [@reaperhulk's feedback](https://github.com/pyca/cryptography/pull/2073#issuecomment-116356191). Anything else?

reaperhulk:
The patched test is failing on libressl.

glyph:
> The patched test is failing on libressl.

whyyyyyyyy

OK. Any tips on how to set up a libressl build environment that I can test with?

glyph:
For those following along at home, I did a `brew install libressl` and then this helpful shell function to switch OpenSSL's:
```bash
function kegger () {
    for kegtoadd in "$@"; do
        local kegpath="/usr/local/opt/$kegtoadd";
        export LDFLAGS="-L${kegpath}/lib $LDFLAGS";
        export CPPFLAGS="-I${kegpath}include $CPPFLAGS";
        export PATH="${kegpath}/bin:$PATH";
    done;
}
```

glyph:
[OpenSSL](https://github.com/openssl/openssl/blob/b39fc560612984e65ec30d7f37487303bf514fb3/crypto/rand/rand_lib.c#L154-L160):

```c
int RAND_bytes(unsigned char *buf, int num)
{
    const RAND_METHOD *meth = RAND_get_rand_method();
    if (meth && meth->bytes)
        return meth->bytes(buf, num);
    return (-1);
}
```

[LibreSSL](https://github.com/libressl/libressl/blob/ba04546060b835e279c0e8a26261e73a1964927c/src/crypto/rand/rand_lib.c#L80-L89):

```c
/*
 * Hurray. You've made it to the good parts.
 */
int
RAND_bytes(unsigned char *buf, int num)
{
	if (num > 0)
		arc4random_buf(buf, num);
	return 1;
}
```

See also https://github.com/libressl-portable/portable/issues/17

glyph:
Regarding the build failures:

- the travis failure is a spurious issue with installing dependencies on the docs builder.
- the codecov failure is a result of LibreSSL not being part of the test matrix; I added that code as a result of @reaperhulk's feedback, so I'm assuming LibreSSL *should* be in that matrix?

reaperhulk:
Libre is in our jenkins cluster and historically we haven't been able to correlate coverage from that. codecov may give us that ability now, but that's going to be a separate investigation. For now I'd say you should make a new function that triggers a skip if you pass a string with "LibreSSL" in it, call that from your test, and then add one more function that tests the skip (using `with pytest.raises(pytest.skip.Exception):`)

reaperhulk:
LGTM, awesome work @glyph. We should (probably as a separate PR) update our docs around the OS random engine to note explicitly that it uses the Python `os.urandom` implementation (and what that entails) along with a note that LibreSSL patches random itself.

alex:
Merge?

On Tue, Jun 30, 2015 at 8:18 PM, Paul Kehrer <notifications@github.com>
wrote:

> LGTM, awesome work @glyph <https://github.com/glyph>. We should (probably
> as a separate PR) update our docs around the OS random engine to note
> explicitly that it uses the Python os.urandom implementation (and what
> that entails) along with a note that LibreSSL patches random itself.
>
> —
> <https://github.com/pyca/cryptography/pull/2073#issuecomment-117377257>.
>



-- 

glyph:
YES PLEASE

glyph:
:tada: :balloon: :star: :star2: :star2: :star: :sparkles: :sparkling_heart: :+1: :ok: :fireworks:

PR #2077, opened 2015-06-27
add EC public/private key pair with c2pnb163v1 curve

reaperhulk:
This is an unusual and currently unsupported curve. We'll use it to improve branch coverage.


alex:
which branch will this help with?

reaperhulk:
I was about to say _load_ssh_ecdsa_public_key but damn it, ssh key. I actually have a better way to fix that branch. I'll leave this PR for the moment but let's not merge it unless we find a use for it.

alex:
This doesn't seem useful, so I'm going to close it.

PR #2079, opened 2015-06-27
Branch fixes

reaperhulk:
This removes the skip logic for tests when serialization is not available. We do not currently have a backend that doesn't support serialization and there's a good argument that we should require all backends to have basic serialization for non-hardware keys.

### DIFF
--- a/tests/hazmat/primitives/test_dsa.py
+++ b/tests/hazmat/primitives/test_dsa.py
@@ -30,21 +30,6 @@ from ...utils import (
 )
 
 
-def _skip_if_no_serialization(key, backend):
-    if not isinstance(
-        key,
-        (dsa.DSAPrivateKeyWithSerialization, dsa.DSAPublicKeyWithSerialization)
-    ):
-        pytest.skip(
-            "{0} does not support DSA key serialization".format(backend)
-        )
-
-
-def test_skip_if_no_serialization():
-    with pytest.raises(pytest.skip.Exception):
-        _skip_if_no_serialization("notakeywithserialization", "backend")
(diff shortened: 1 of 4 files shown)


.coveragerc
```diff
@@ -13,3 +13,4 @@ source =
 exclude_lines =
     @abc.abstractmethod
     @abc.abstractproperty
+    @property
```
alex: Why is this needed?

reaperhulk: https://codecov.io/github/pyca/cryptography/tests/test_interfaces.py#l-50 (I also see it locally during coverage. I vaguely recall "missed branch" being why we excluded the abc decorators as well)

alex: Ugh, those aren't about the `@property` lines, those are about those
methods never being invoked. Can you remove this for now and we can think
about alternative approaches?

On Sat, Jun 27, 2015 at 3:22 PM, Paul Kehrer <notifications@github.com>
wrote:

> In .coveragerc
> <https://github.com/pyca/cryptography/pull/2079#discussion_r33415262>:
>
> > @@ -13,3 +13,4 @@ source =
> >  exclude_lines =
> >      @abc.abstractmethod
> >      @abc.abstractproperty
> > +    @property
>
> https://codecov.io/github/pyca/cryptography/tests/test_interfaces.py#l-50
> (I also see it locally during coverage. I vaguely recall "missed branch"
> being why we excluded the abc decorators as well)
>
> —
> <https://github.com/pyca/cryptography/pull/2079/files#r33415262>.
>



-- 

PR #2070, opened 2015-06-27
Testing codecov/support#43

stevepeak:
Read more at https://github.com/codecov/support/issues/43 Thanks!

PR #2089, opened 2015-06-29
Switch to codecov

reaperhulk:
codecov is now fully working thanks to @stevepeak's hard work. Let's switch over!

Note: Since codecov has branch coverage we can now see that we are **not at 100%**. This means we can't set the status API to fail at 100% only right now. Fortunately, codecov supports a coverage on diff'd lines measurement so we can setting that to 100% while we work on covering the missed branches.

This supersedes #2070

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -120,7 +120,7 @@ script:
     - ./.travis/run.sh
 
 after_success:
-    - source ~/.venv/bin/activate && codecov --env TRAVIS_OS_NAME TOXENV OPENSSL
+    - source ~/.venv/bin/activate && bash <(curl -s https://raw.githubusercontent.com/codecov/codecov-bash/coveragepy/codecov) -e TRAVIS_OS_NAME,TOXENV,OPENSSL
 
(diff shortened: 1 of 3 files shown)


reaperhulk:
One slight hiccup, we can only require integral numbers for minimum coverage. So if we want the codecov checks to show up as passing in status API we have to set it to 99% until we get to 100%. @stevepeak is it possible to add decimal precision to that field?

alex:
@reaperhulk if we've got 100% required for new/changed lines, it ought to work out right?

reaperhulk:
@alex It should, yes.

reaperhulk:
Hmm, would be very nice if codecov didn't report failure during the first few submissions from travis. coveralls seems to know when travis is done -- is that something that could be added to codecov @stevepeak?

alex:
Personally I don't mind it.

stevepeak:
Wow! Thank you my friends for your attention and patience. I feel very honored having Codecov in this project! :+1:

PR #2090, opened 2015-06-29
Fix is now on main uploader

stevepeak:
Fix for branches is not formally available on Codecov. Thanks all for helping on this!
Cheers! :+1:


reaperhulk:
Jenkins, ok to test

PR #2092, opened 2015-07-02
name constraints - support leading periods

reaperhulk:
Name Constraints may have leading periods.

refs #1947

PR #2094, opened 2015-07-02
name constraints - support leading periods

reaperhulk:
replaces #2092


reaperhulk:
Clever github, I do have to rebase.

PR #2095, opened 2015-07-02
name constraints - support IP addresses with netmask

reaperhulk:
refs #1947

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -141,11 +141,30 @@ def _decode_general_name(backend, gn):
         oid = _obj2txt(backend, gn.d.registeredID)
         return x509.RegisteredID(x509.ObjectIdentifier(oid))
     elif gn.type == backend._lib.GEN_IPADD:
-        return x509.IPAddress(
-            ipaddress.ip_address(
-                _asn1_string_to_bytes(backend, gn.d.iPAddress)
-            )
-        )
+        data = backend._ffi.buffer(
+            gn.d.iPAddress.data, gn.d.iPAddress.length
+        )[:]
+        data_len = len(data)
+        if data_len == 8 or data_len == 32:
+            # This is an IPv4 or IPv6 Network and not a single IP. This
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -125,13 +125,30 @@ def _decode_general_name(backend, gn):
...
+            # ipaddress doesn't support packed bytes + netmask. Additionally,
+            # IPv6Network can only handle CIDR rather than the full 16 byte
+            # netmask. To handle this we convert the netmask to integer, then
+            # find the first 0 bit, which will be the prefix. If another 1
+            # bit is present after that the netmask is invalid.
+            base = ipaddress.ip_address(data[:data_len // 2])
+            netmask = utils.int_from_bytes(data[data_len // 2:], 'big')
+            bits = bin(netmask)[2:]
+            prefix = bits.find('0')
+            if bits[prefix:].find('1') != -1:
```
alex: Is this just `if "1" not in bits[prefix:]`? Also, `bits.find("0")` returns `-1` on non-existennce. Will that trigger a bug?

reaperhulk: Yeah it is. And definitely the former is a bug. I'll have to add one more vector for that.

alex:
tests are not passing

alex:
this needs a rebase

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -141,11 +141,34 @@ def _decode_general_name(backend, gn):
...
         return x509.RegisteredID(x509.ObjectIdentifier(oid))
     elif gn.type == backend._lib.GEN_IPADD:
-        return x509.IPAddress(
-            ipaddress.ip_address(
-                _asn1_string_to_bytes(backend, gn.d.iPAddress)
-            )
-        )
+        data = backend._ffi.buffer(
+            gn.d.iPAddress.data, gn.d.iPAddress.length
+        )[:]
```
alex: This should be `_asn1_string_to_bytes`

PR #2098, opened 2015-07-02
Adds certificate builder.

sigmavirus24:
This is a rebased version of #1980 and will be used to finish up the work in that PR.


alex:
There's a bunch of API changes we need to make this consistent with the CSRBuilder.

src/cryptography/x509.py
```diff
@@ -5,13 +5,14 @@
...
 import abc
+import datetime
 import ipaddress
 from enum import Enum
 
 import six
 
 from cryptography import utils
-from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives import dsa, ec, rsa, hashes
```
alex: `dsa`, `ec`, and `rsa` all come from `cryptography.hazmat.primitives.asymmetric`

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -83,6 +84,21 @@ def _encode_asn1_str(backend, data, length):
     s = backend._lib.ASN1_OCTET_STRING_new()
     s = backend._ffi.gc(s, backend._lib.ASN1_OCTET_STRING_free)
     backend._lib.ASN1_OCTET_STRING_set(s, data, length)
```
alex: Somehow `return s` got lost here

sigmavirus24: Sorry about that. The rebase must have eaten it.

docs/x509.rst
```diff
@@ -390,6 +390,89 @@ X.509 CRL (Certificate Revocation List) Object
...
         The extensions encoded in the CRL.
 
+X.509 Certificate Builder
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. class:: CertificateBuilder
+
+    .. method:: __init__()
+
+        Creates an empty certificate (version 1).
```
alex: We've removed the `__init__` docs, and replaced it with examples.

docs/x509.rst
```diff
@@ -390,6 +390,89 @@ X.509 CRL (Certificate Revocation List) Object
...
+
+    .. method:: set_version(version)
```
alex: All the `set_` prefixes have been dropped

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -83,6 +84,21 @@ def _encode_asn1_str(backend, data, length):
...
+
+
+def _make_asn1_int(backend, x):
```
alex: We've gone with an `encode` prefix

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -83,6 +84,21 @@ def _encode_asn1_str(backend, data, length):
...
+    i = backend._lib.ASN1_INTEGER_new()
+    # i = backend._ffi.gc(i, backend._lib.ASN1_INTEGER_free)
+    backend._lib.ASN1_INTEGER_set(i, x)
+    return i
+
+
+def _make_asn1_str(backend, x, n=None):
```
alex: This seems to dpulicate `_encode_asn1_str`

src/cryptography/x509.py
```diff
@@ -1489,3 +1491,156 @@ def sign(self, private_key, algorithm, backend):
...
+            self._public_key, self._serial_number, self._not_valid_before,
+            self._not_valid_after, self._extensions + [extension]
+        )
+
+    def sign(self, backend, private_key, algorithm):
+        """
+        Signs the certificate using the CA's private key.
+        """
+        if self._version is None:
+            self._version = Version.v1
```
alex: This shouldn't mutate the object; and if we're going to have a default I'm not sure `v1` is correct

docs/x509/reference.rst
```diff
@@ -388,6 +388,85 @@ X.509 CRL (Certificate Revocation List) Object
 
         The extensions encoded in the CRL.
 
+X.509 Certificate Builder
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. class:: CertificateBuilder
+
+    .. method:: set_version(version)
```
alex: This is no longer the name of the method.

src/cryptography/x509.py
```diff
@@ -1594,3 +1596,155 @@ def sign(self, private_key, algorithm, backend):
...
+        Adds an X.509 extension to the certificate.
+        """
+        if isinstance(extension, BasicConstraints):
+            extension = Extension(OID_BASIC_CONSTRAINTS, critical, extension)
+        elif isinstance(extension, SubjectAlternativeName):
+            extension = Extension(
+                OID_SUBJECT_ALTERNATIVE_NAME, critical, extension
+            )
+        else:
+            raise NotImplementedError('Unsupported X.509 extension.')
```
wbond: From all of the work I have been doing with X509 recently, I think the extension support should be improved here if there is the intent for this to be used.

According to RFC5280, if you are going to be creating CA certs (which is what basic constraints allows for), then conforming CAs MUST include the key usage extension.

For the sake of validating certificates, the authority key identifier and subject key identifier extensions should probably also be supported.

reaperhulk: @wbond we definitely intend to implement more extensions immediately following this PR. Some of that work (notably SAN) is already done but will be wired up later to prevent making this PR too large.

wbond: Sounds good, just figured I'd mention it.

src/cryptography/x509.py
```diff
@@ -1594,3 +1596,155 @@ def sign(self, private_key, algorithm, backend):
...
+            self._not_valid_after, self._extensions + [extension]
+        )
+
+    def sign(self, backend, private_key, algorithm):
+        """
+        Signs the certificate using the CA's private key.
+        """
+        builder = self
+        if self._version is None:
+            builder = self.version(Version.v3)
```
sigmavirus24: Our examples all use an explicit version. Do we want to support defaulting here? I did it because that's what the first pass on this did.

wbond: Hmm, you just got me thinking about versions. Should it even be possible to generate certificates that aren't v3? v3 has been a standard since 1996, and you can't use extensions with v2 or v1.

I am thinking v3 should be hard-coded.

https://tools.ietf.org/html/rfc5280#section-3.1

sigmavirus24: For what it's worth, the Version enum doesn't have v2 defined. It only has v1 and v3 defined. I'd love other people's thoughts on this.

reaperhulk: I'm okay with version going away for now. There are scenarios where I might want to generate v1 certificates, but it doesn't strike me as unreasonable to state that doing that is unsupported for now.

wbond: @reaperhulk Not questioning your reasoning, just looking to learn – in what situations is a v1 cert valuable?

alex: Writing tests for an X.509 library.

On Thu, Jul 23, 2015 at 1:45 PM, Will Bond <notifications@github.com> wrote:

> In src/cryptography/x509.py
> <https://github.com/pyca/cryptography/pull/2098#discussion_r35349775>:
>
> > +            if e.oid == extension.oid:
> > +                raise ValueError('This extension has already been set.')
> > +
> > +        return CertificateBuilder(
> > +            self._version, self._issuer_name, self._subject_name,
> > +            self._public_key, self._serial_number, self._not_valid_before,
> > +            self._not_valid_after, self._extensions + [extension]
> > +        )
> > +
> > +    def sign(self, backend, private_key, algorithm):
> > +        """
> > +        Signs the certificate using the CA's private key.
> > +        """
> > +        builder = self
> > +        if self._version is None:
> > +            builder = self.version(Version.v3)
>
> @reaperhulk <https://github.com/reaperhulk> Not questioning your
> reasoning, just looking to learn – in what situations is a v1 cert valuable?
>
> —
> <https://github.com/pyca/cryptography/pull/2098/files#r35349775>.
>



-- 

reaperhulk: What @alex said pretty much. There's not really any good reason beyond "feed it to other libraries as a test corpus to find security bugs".

wbond: On that note, if you could just go ahead and add support for policy constraints and then create some valid and invalid certificate paths utilizing policy constraints for the sake of path validation testing, that would be awesome! :smile:

I have run into the issue of trying to find good test vectors for x509 related to various extensions and constraints.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -988,6 +989,90 @@ def create_x509_csr(self, builder, private_key, algorithm):
...
+                raise NotImplementedError('Extension not yet supported.')
+
+            obj = _txt2obj(self, extension.oid.dotted_string)
+            extension = self._lib.X509_EXTENSION_create_by_OBJ(
+                self._ffi.NULL,
+                obj,
+                1 if extension.critical else 0,
+                _encode_asn1_str_gc(self, pp[0], r)
+            )
+            res = self._lib.X509_add_ext(x509_cert, extension, i)
```
reaperhulk: Given some other bugs we've had this should be checked to verify it doesn't return the position of the cert. Based on the signature it looks like it probably doesn't, but let's be sure :smile:

sigmavirus24: How might one check that?

reaperhulk: Sadly the only way is to start looking at function and macro definitions until you find what returns the integer.

reaperhulk: I looked at it. 1 is right. For reference:

```c
int X509_add_ext(X509 *x, X509_EXTENSION *ex, int loc)
    {
    return(X509v3_add_ext(&(x->cert_info->extensions),ex,loc) != NULL);
    }
```

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -988,6 +989,90 @@ def create_x509_csr(self, builder, private_key, algorithm):
 
         return _CertificateSigningRequest(self, x509_req)
 
+    def sign_x509_certificate(self, builder, private_key, algorithm):
+        # TODO: check type of private key parameter.
```
reaperhulk: We should go ahead and just handle this like we did in the CSR builder.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -988,6 +989,90 @@ def create_x509_csr(self, builder, private_key, algorithm):
...
+
+        # Set the subject's public key.
+        res = self._lib.X509_set_pubkey(
+            x509_cert, builder._public_key._evp_pkey
+        )
+        assert res == 1
+
+        # Set the certificate serial number.
+        serial_number = _encode_asn1_int(self, builder._serial_number)
+        self._lib.X509_set_serialNumber(x509_cert, serial_number)
```
reaperhulk: This should have its response checked ( == 1)

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -988,6 +989,90 @@ def create_x509_csr(self, builder, private_key, algorithm):
...
+            assert res == 1
+
+        # Set the issuer name.
+        res = self._lib.X509_set_issuer_name(
+            x509_cert, _encode_name(self, list(builder._issuer_name))
+        )
+        assert res == 1
+
+        # Sign the certificate with the issuer's private key.
+        res = self._lib.X509_sign(
```
reaperhulk: Could you write a test that attempts to sign a SHA512 certificate with a 512-bit (or smaller) RSA key? I'm guessing it's going to error and I'm curious where. I think we may need to more elegantly handle X509_sign response codes (and clear the OpenSSL error stack).

sigmavirus24: See 8690eff although it appeared to fail for me.

src/cryptography/hazmat/backends/openssl/backend.py
sigmavirus24: Done

src/cryptography/hazmat/backends/openssl/backend.py
sigmavirus24: Done

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: It absolutely should have failed, so I'm pleased that logic prevails. We'll need something like this:

```python
if res == 0:
    self._consume_errors()
    raise SomeError
```

We may need to check the error that comes back actually. Then we could give a bit more informative errors back (digest too big for key, etc). Of course, coverage becomes tricky then because we also need a path to catch errors we don't have cases for. Not sure what SomeError is either. Probably ValueError given our general tendencies?

(We need this logic in CSRBuilder too. I just tested and the same test triggers the bug)

reaperhulk: https://github.com/pyca/cryptography/pull/2175 contains the CSRBuilder fix. Should work fine here too.

docs/x509/reference.rst
```diff
@@ -388,6 +388,78 @@ X.509 CRL (Certificate Revocation List) Object
...
+        Adds an X.509 extension to the certificate.
+
+        :param extension: The :class:`~cryptography.x509.Extension` to add to
+            the certificate.
+
+    .. method:: sign(backend, private_key, algorithm)
+
+        Sign the certificate using the CA's private key.
+
+        :param algorithm: The
```
reaperhulk: This should have the other two params documented as well

src/cryptography/x509.py
```diff
@@ -1594,3 +1596,138 @@ def sign(self, private_key, algorithm, backend):
...
+                 public_key=None, serial_number=None, not_valid_before=None,
+                 not_valid_after=None, extensions=[]):
+        """
+        Creates an empty X.509 certificate (version 1).
+        """
+        self._version = Version.v3
+        self._issuer_name = issuer_name
+        self._subject_name = subject_name
+        self._public_key = public_key
+        self._serial_number = serial_number
```
reaperhulk: We should check that the serial number is not > 160-bit as RFC 5280 states that conforming CAs MUST NOT issue serials greater than 20 octets in length.

reaperhulk: I guess that means we should also check that it's not negative and that it's actually an integer :) (and we should document the zero or greater and 160-bits max)

src/cryptography/x509.py
```diff
@@ -1594,3 +1596,138 @@ def sign(self, private_key, algorithm, backend):
...
             raise ValueError("A CertificateSigningRequest must have a subject")
         return backend.create_x509_csr(self, private_key, algorithm)
+
+
+class CertificateBuilder(object):
+    def __init__(self, version=None, issuer_name=None, subject_name=None,
+                 public_key=None, serial_number=None, not_valid_before=None,
+                 not_valid_after=None, extensions=[]):
+        """
+        Creates an empty X.509 certificate (version 1).
```
reaperhulk: This comment is out of date, why don't we just remove it entirely?

src/cryptography/x509.py
reaperhulk: Oops I see we *are* enforcing integer type in the setter, but not the non-negative or 160-bit restriction.

src/cryptography/x509.py
```diff
@@ -1594,3 +1596,138 @@ def sign(self, private_key, algorithm, backend):
...
+            self._version, self._issuer_name, self._subject_name,
+            self._public_key, number, self._not_valid_before,
+            self._not_valid_after, self._extensions
+        )
+
+    def not_valid_before(self, time):
+        """
+        Sets the certificate activation time.
+        """
+        # TODO: require UTC datetime?
```
reaperhulk: Is this a legit TODO? (same goes for not_valid_after) We use naïve datetimes that are implicitly UTC (and documented as such) when calling not_valid_before or not_valid_after on a Certificate object so is there any reason not to do it here?

tests/hazmat/backends/test_openssl.py
```diff
@@ -478,6 +480,41 @@ def test_unsupported_ec_keys(self):
...
+    def test_requires_certificate_builder(self):
+        private_key = RSA_KEY_2048.private_key(backend)
+
+        with pytest.raises(TypeError):
+            backend.sign_x509_certificate(object(), private_key, DummyHash())
+
+    def test_checks_for_unsupported_extensions(self):
+        private_key = RSA_KEY_2048.private_key(backend)
+        builder = x509.CertificateBuilder().subject_name(x509.Name([
+            x509.NameAttribute(x509.OID_COUNTRY_NAME, u'US'),
```
reaperhulk: We shoudl slim down these tests where possible. There are a lot of cases where a bunch of NameAttributes (or lots of values in an extension) are set that are unnecessary.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -988,6 +989,102 @@ def create_x509_csr(self, builder, private_key, algorithm):
...
+
+        # Set the certificate serial number.
+        serial_number = _encode_asn1_int(self, builder._serial_number)
+        res = self._lib.X509_set_serialNumber(x509_cert, serial_number)
+        assert res == 1
+
+        # Set the "not before" time.
+        res = self._lib.ASN1_TIME_set(
+            self._lib.X509_get_notBefore(x509_cert),
+            calendar.timegm(builder._not_valid_before.timetuple())
```
reaperhulk: If you pass a datetime that is earlier than Jan 1 1970 you get a negative integer out of timegm, which is probably not good. We should probably check up in the setter to confirm not_valid_before doesn't predate the dawn of unix time.

src/cryptography/x509.py
sigmavirus24: I guess not. I just left it in there from before because I didn't know the right answer myself. `:D`

sigmavirus24:
I addressed some of the comments @reaperhulk and I'll take care of the rest tomorrow and add tests. If you can give advice about trimming down the tests, I'll work on that tomorrow.

docs/x509/reference.rst
```diff
@@ -388,6 +388,89 @@ X.509 CRL (Certificate Revocation List) Object
...
+X.509 Certificate Builder
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. class:: CertificateBuilder
+
+    .. method:: issuer_name(name)
+
+        Sets the issuer's distinguished name.
+
+        :param public_key: The :class:`~cryptography.x509.Name` that describes
```
reaperhulk: this should say param name

docs/x509/reference.rst
```diff
@@ -388,6 +388,89 @@ X.509 CRL (Certificate Revocation List) Object
...
+            the issuer (CA).
+
+    .. method:: subject_name(name)
+
+        Sets the subject's distinguished name.
+
+        :param public_key: The :class:`~cryptography.x509.Name` that describes
```
reaperhulk: param name

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -988,6 +989,102 @@ def create_x509_csr(self, builder, private_key, algorithm):
...
+        assert res == 1
+
+        # Set the subject's public key.
+        res = self._lib.X509_set_pubkey(
+            x509_cert, builder._public_key._evp_pkey
+        )
+        assert res == 1
+
+        # Set the certificate serial number.
+        serial_number = _encode_asn1_int(self, builder._serial_number)
```
reaperhulk: This will leak, as `X509_set_serialNumber` makes a copy. We'l need an _encode_asn1_int_gc function that calls _encode_asn1_int (take a look at _encode_asn1_str_gc).

docs/x509/reference.rst
```diff
@@ -388,6 +388,89 @@ X.509 CRL (Certificate Revocation List) Object
...
+
+        :param name: The :class:`~cryptography.x509.Name` that describes the
+            issuer (CA).
+
+    .. method:: subject_name(name)
+
+        Sets the subject's distinguished name.
+
+        :param name: The :class:`~cryptography.x509.Name` that describes the
+            subject (requester).
```
reaperhulk: The (requester) here sounds like it was copied from the CSRBuilder PR. We should remove it

tests/test_x509.py
```diff
@@ -22,7 +22,7 @@
 from cryptography.hazmat.primitives.asymmetric import dsa, ec, rsa
 
 from .hazmat.primitives.fixtures_dsa import DSA_KEY_2048
-from .hazmat.primitives.fixtures_rsa import RSA_KEY_2048
+from .hazmat.primitives.fixtures_rsa import RSA_KEY_512, RSA_KEY_2048
```
alex: Imports need to be ordered lexicographically, so 2048 comes before 512.

tests/test_x509.py
```diff
@@ -775,6 +775,350 @@ def test_hash(self, backend):
...
+        assert cert.not_valid_after == not_valid_after
+        basic_constraints = cert.extensions.get_extension_for_oid(
+            x509.OID_BASIC_CONSTRAINTS
+        )
+        assert basic_constraints.value.ca is False
+        assert basic_constraints.value.path_length is None
+
+    @pytest.mark.requires_backend_interface(interface=EllipticCurveBackend)
+    @pytest.mark.requires_backend_interface(interface=X509Backend)
+    def test_build_cert_with_dsa_private_key(self, backend):
```
alex: There are two methods here with teh same name.

docs/x509/reference.rst
```diff
@@ -388,6 +388,89 @@ X.509 CRL (Certificate Revocation List) Object
...
+
+        :param time: The `datetime.datetime` object (in UTC) that marks the
+            expiration time for the certificate.  The certificate may not be
+            trusted clients if it is used after this time.
+
+    .. method:: add_extension(extension)
+
+        Adds an X.509 extension to the certificate.
+
+        :param extension: The :class:`~cryptography.x509.Extension` to add to
```
reaperhulk: We don't actually add Extension here...but we also don't have an abc that all the extension classes are registered against. This is also an issue with the docs for CSRBuilder

tests/test_x509.py
```diff
@@ -775,6 +775,337 @@ def test_hash(self, backend):
...
+            x509.NameAttribute(x509.OID_STATE_OR_PROVINCE_NAME, u'Texas'),
+            x509.NameAttribute(x509.OID_LOCALITY_NAME, u'Austin'),
+            x509.NameAttribute(x509.OID_ORGANIZATION_NAME, u'PyCA'),
+            x509.NameAttribute(x509.OID_COMMON_NAME, u'cryptography.io'),
+        ])).public_key(
+            subject_private_key.public_key()
+        ).add_extension(
+            x509.BasicConstraints(ca=False, path_length=None), True,
+        ).add_extension(
+            x509.SubjectAlternativeName([x509.DNSName(u"cryptography.io")]),
```
reaperhulk: We should check for this extension in the cert or remove it from the test

tests/test_x509.py
```diff
@@ -775,6 +775,337 @@ def test_hash(self, backend):
...
+
+        builder = x509.CertificateBuilder().serial_number(
+            777
+        ).issuer_name(x509.Name([
+            x509.NameAttribute(x509.OID_COUNTRY_NAME, u'US'),
+        ])).subject_name(x509.Name([
+            x509.NameAttribute(x509.OID_COUNTRY_NAME, u'US'),
+        ])).public_key(
+            subject_private_key.public_key()
+        ).add_extension(
```
reaperhulk: These extensions aren't needed to test the error case here.

reaperhulk:
Other than my last few comments this looks good now. Awesome work so far @sigmavirus24.

I'm not entirely sure what to do about the type of `add_extension`. Maybe we need to define an abc called `ExtensionType` (or a better name) and register all the X509 extensions against it so we have an interface to describe? I'm okay with that discussion going outside this PR though since the issue already exists in CSRBuilder.

docs/x509/reference.rst
sigmavirus24: So I've updated these docs. Let me know if they're okay.

tests/test_x509.py
sigmavirus24: Done.

tests/test_x509.py
sigmavirus24: Removed.

sigmavirus24:
I can submit a different PR for the CSRBuilder docs if what I've done above is okay.

tests/test_x509.py
reaperhulk: Did you not push that commit? Looks like it's still here :)

sigmavirus24: Didn't add it to the index before committing the last one. Sorry. Should be fixed and running the tests now.

reaperhulk:
I think I'm happy now, but since this is so big I'd like @alex to give it a once over before merge.

docs/x509/reference.rst
```diff
@@ -388,6 +388,93 @@ X.509 CRL (Certificate Revocation List) Object
...
+            :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPublicKey`,
+            :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPublicKey` or
+            :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey`
+
+    .. method:: serial_number(serial_number)
+
+        Sets the certificate's serial number (an integer).  The CA's policy
+        determines how it attributes serial numbers to certificates.  The only
+        requirement is that this number uniquely identify the certificate given
+        the issuer.
```
alex: @reaperhulk can we make a normative recommendation here? (e.g. use a uuid or something?)

docs/x509/reference.rst
```diff
@@ -388,6 +388,93 @@ X.509 CRL (Certificate Revocation List) Object
...
+            identify this certificate (most notably during certificate
+            revocation checking).
+
+    .. method:: not_valid_before(time)
+
+        Sets the certificate's activation time.  This is the time from which
+        clients can start trusting the certificate.  It may be different from
+        the time at which the certificate was created.
+
+        :param time: The `datetime.datetime` object (in UTC) that marks the
```
alex: This should be `:class:` or use double backticks.

docs/x509/reference.rst
```diff
@@ -388,6 +388,93 @@ X.509 CRL (Certificate Revocation List) Object
...
+            activation time for the certificate.  The certificate may not be
+            trusted clients if it is used before this time.
+
+    .. method:: not_valid_after(time)
+
+        Sets the certificate's expiration time.  This is the time from which
+        clients should no longer trust the certificate.  The CA's policy will
+        determine how long the certificate should remain in use.
+
+        :param time: The `datetime.datetime` object (in UTC) that marks the
```
alex: Same here.

docs/x509/reference.rst
```diff
@@ -388,6 +388,93 @@ X.509 CRL (Certificate Revocation List) Object
...
+            expiration time for the certificate.  The certificate may not be
+            trusted clients if it is used after this time.
+
+    .. method:: add_extension(extension)
```
alex: Shouldn't this have a `critical` flag like CSR Builder does?

src/cryptography/hazmat/backends/openssl/backend.py
alex: Is tehre a reason not use `X509_set_notBefore` here? Seems more readable.

tests/hazmat/backends/test_openssl.py
```diff
@@ -478,6 +480,41 @@ def test_unsupported_ec_keys(self):
...
+            private_key.public_key()
+        ).serial_number(
+            777
+        ).not_valid_before(
+            datetime.datetime(1999, 1, 1)
+        ).not_valid_after(
+            datetime.datetime(2020, 1, 1)
+        )
+
+        builder._extensions.append(x509.Extension(
```
alex: eep. I think the approach we went for in CSR builder was to add support for an extension to the frontend, but not add it in the backend.

tests/hazmat/backends/test_openssl.py
alex: This is still the case.

alex:
Left a few comments. The docs could probably also use a few examples.

docs/x509/reference.rst
reaperhulk: UUID4 isn't a bad choice converted to int. Current CABForum baseline requirements say "CAs SHOULD generate non‐sequential Certificate serial numbers that exhibit at least 20 bits of entropy". You can also do things like microtime + urandom to get to 160 bits

alex: We should probably mention the "at least 20 minutes of entropy" line, and
maybe link to our secure random number generation page? (Which should
probably document how to get an itneger... I'll file a ticket on that)

On Sun, Aug 2, 2015 at 4:13 PM, Paul Kehrer <notifications@github.com>
wrote:

> In docs/x509/reference.rst
> <https://github.com/pyca/cryptography/pull/2098#discussion_r36047152>:
>
> > +
> > +    .. method:: public_key(public_key)
> > +
> > +        Sets the subject's public key.
> > +
> > +        :param public_key: The subject's public key. This can be one of
> > +            :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPublicKey`,
> > +            :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPublicKey` or
> > +            :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey`
> > +
> > +    .. method:: serial_number(serial_number)
> > +
> > +        Sets the certificate's serial number (an integer).  The CA's policy
> > +        determines how it attributes serial numbers to certificates.  The only
> > +        requirement is that this number uniquely identify the certificate given
> > +        the issuer.
>
> UUID4 isn't a bad choice converted to int. Current CABForum baseline
> requirements say "CAs SHOULD generate non‐sequential Certificate serial
> numbers that exhibit at least 20 bits of entropy". You can also do things
> like microtime + urandom to get to 160 bits
>
> —
> <https://github.com/pyca/cryptography/pull/2098/files#r36047152>.
>



-- 

tests/hazmat/backends/test_openssl.py
sigmavirus24: I'm confused by this comment. The CSRBuilder checks for extensions in [the backend][] as well as [the front end][]. In order to get proper test coverage, this seemed to be the easiest way to write the test that it would not allow a random extension.

[the backend]: https://github.com/pyca/cryptography/blob/93b5e3efab0337c51c84c27208d9034b607f09e2/src/cryptography/hazmat/backends/openssl/backend.py#L1030..L1039

[the front end]: https://github.com/pyca/cryptography/blob/93b5e3efab0337c51c84c27208d9034b607f09e2/src/cryptography/x509.py#L1572..L1585

sigmavirus24: I guess it could be rewritten to pass this as a member of a list to the constructor though. Is that what you want?

docs/x509/reference.rst
reaperhulk: Yeah noting 20 bytes of entropy is reasonable.

docs/x509/reference.rst
reaperhulk: Implementation does, so docs just need to note the critical bool

src/cryptography/hazmat/backends/openssl/backend.py
sigmavirus24: `¯\_(ツ)_/¯`. I didn't really write much of this at all. What would that look like?

I suppose we'd use `X509_set_notAfter` below too around L1055.

src/cryptography/x509.py
```diff
@@ -1594,3 +1598,146 @@ def sign(self, private_key, algorithm, backend):
...
+        """
+        Adds an X.509 extension to the certificate.
+        """
+        if isinstance(extension, BasicConstraints):
+            extension = Extension(OID_BASIC_CONSTRAINTS, critical, extension)
+        elif isinstance(extension, SubjectAlternativeName):
+            extension = Extension(
+                OID_SUBJECT_ALTERNATIVE_NAME, critical, extension
+            )
+        elif isinstance(extension. InhibitAnyPolicy):
```
alex: Comma, not a period :-)

docs/random-numbers.rst
```diff
@@ -1,3 +1,5 @@
+.. _secure_random_number_generation:
```
alex: No need for this,
```rest
:doc:`/random-numbers`
```

works

sigmavirus24: TIL

reaperhulk:
It looks like `sign_x509_certificate` is not on the `X509Backend` interface in `cryptography.hazmat.backends.interfaces`.

We also need a changelog entry (which will require a rebase or you'll see a conflict since we pushed some changes to that recently)

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: It wouldn't be a huge improvement. `X509_set_notBefore` takes an ASN1_TIME * so you'd still need to call `ASN1_TIME_set`.

PR #2100, opened 2015-07-02
Shift around the x509 code, all the extensions are going to have the same thing

alex:


alex:
(Don't fear the red, yes there's 1 less line covered, but there's also one less line!)

PR #2106, opened 2015-07-03
Added some nonsense around IA5STRING

alex:


alex:
I'm not sure this is strictly necessary, but this typedef has been this way since I was in the 2nd grade, so I'm not too worried.

PR #2108, opened 2015-07-04
Support OtherName for GeneralNames.

JoshData:
There are a few GeneralName types that aren't yet supported. I encountered an otherName in the wild (https://github.com/mail-in-a-box/mailinabox/issues/481). Rather than raising an exception and preventing the certificate from being loaded at all, it would be nice if the library is able to parse the certificate but leave the otherName as a placeholder.

This patch removes the `x509.UnsupportedGeneralNameType` exception and adds a new class `x509.UnsupportedGeneralName`.

What do you think?

I didn't update the tests. Wasn't sure quite how to do that.

### DIFF
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -695,6 +695,20 @@ General Name Classes
 
         :type: :class:`ObjectIdentifier`
 
+.. class:: OtherName
+
+    .. versionadded:: 1.0
+
+    This corresponds to an "otherName."  An otherName has a type identifier and a value represented in binary DER format.
+
+    .. attribute:: type_id
+
+        :type: :class:`ObjectIdentifier`
+
+    .. attribute:: value
+
+        :type: `bytes`
+
 X.509 Extensions
 ~~~~~~~~~~~~~~~~
 
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -699,7 +699,7 @@ General Name Classes
 
     .. versionadded:: 1.0
 
-    This corresponds to an "otherName."  An otherName has a type identifier and a value represented in binary DER format.
+    This corresponds to an ``otherName.``  An ``otherName`` has a type identifier and a value represented in binary DER format.
 
     .. attribute:: type_id
 
--- a/docs/x509.rst
+++ b/docs/x509.rst
@@ -981,6 +981,7 @@ X.509 Extensions
             :ref:`general name classes <general_name_classes>`.
 
         :returns: A list of values extracted from the matched general names.
+            The type of the returned values depends on the :class:`GeneralName`.
 
         .. doctest::
 
--- a/src/_cffi_src/openssl/asn1.py
+++ b/src/_cffi_src/openssl/asn1.py
@@ -155,6 +155,8 @@ int ASN1_UTCTIME_check(ASN1_UTCTIME *);
 
 /* Not a macro, const on openssl 1.0 */
 int ASN1_STRING_set_default_mask_asc(char *);
+
+int i2d_ASN1_TYPE(ASN1_TYPE *, unsigned char **);
 """
 
 CUSTOMIZATIONS = """
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -47,6 +47,17 @@ def _asn1_string_to_utf8(backend, asn1_string):
     return backend._ffi.buffer(buf[0], res)[:].decode('utf8')
 
 
+def _asn1_to_der(backend, asn1_type):
+    buf = backend._ffi.new("unsigned char **")
+    res = backend._lib.i2d_ASN1_TYPE(asn1_type, buf)
+    assert res >= 0
+    assert buf[0] != backend._ffi.NULL
+    buf = backend._ffi.gc(
+        buf, lambda buffer: backend._lib.OPENSSL_free(buffer[0])
+    )
+    return backend._ffi.buffer(buf[0], res)[:]
+
+
 def _decode_x509_name_entry(backend, x509_name_entry):
     obj = backend._lib.X509_NAME_ENTRY_get_object(x509_name_entry)
     assert obj != backend._ffi.NULL
@@ -157,8 +168,12 @@ def _decode_general_name(backend, gn):
             return x509.RFC822Name(
                 parts[0] + u"@" + idna.decode(parts[1])
             )
+    elif gn.type == backend._lib.GEN_OTHERNAME:
+        type_id = _obj2txt(backend, gn.d.otherName.type_id)
+        value = _asn1_to_der(backend, gn.d.otherName.value)
+        return x509.OtherName(x509.ObjectIdentifier(type_id), value)
     else:
-        # otherName, x400Address or ediPartyName
+        # x400Address or ediPartyName
         raise x509.UnsupportedGeneralNameType(
             "{0} is not a supported type".format(
                 x509._GENERAL_NAMES.get(gn.type, gn.type)
--- a/src/cryptography/x509.py
+++ b/src/cryptography/x509.py
@@ -1045,6 +1045,34 @@ class IPAddress(object):
         return not self == other
 
 
+@utils.register_interface(GeneralName)
+class OtherName(object):
+    def __init__(self, type_id, value):
+        if not isinstance(type_id, ObjectIdentifier):
+            raise TypeError("type_id must be an ObjectIdentifier")
+        if not isinstance(value, bytes):
+            raise TypeError("value must be a binary string")
+
+        self._type_id = type_id
+        self._value = value
+
+    type_id = utils.read_only_property("_type_id")
+    value = utils.read_only_property("_value")
+
+    def __repr__(self):
+        return "<OtherName(type_id={0}, value={1!r})>".format(
+            self.type_id, self.value)
+
+    def __eq__(self, other):
+        if not isinstance(other, OtherName):
+            return NotImplemented
+
+        return self.type_id == other.type_id and self.value == other.value
+
+    def __ne__(self, other):
+        return not self == other
+
+
 class GeneralNames(object):
     def __init__(self, general_names):
         if not all(isinstance(x, GeneralName) for x in general_names):
--- a/src/cryptography/x509.py
+++ b/src/cryptography/x509.py
@@ -1090,7 +1090,13 @@ class GeneralNames(object):
         return len(self._general_names)
 
     def get_values_for_type(self, type):
-        return [i.value for i in self if isinstance(i, type)]
+        # Return the value of each GeneralName, except for OtherName instances
+        # which we return directly because it has two important properties not
+        # just one value.
+        objs = (i for i in self if isinstance(i, type))
+        if type != OtherName:
+            objs = (i.value for i in objs)
+        return list(objs)
 
     def __repr__(self):
         return "<GeneralNames({0})>".format(self._general_names)
--- a/tests/test_x509_ext.py
+++ b/tests/test_x509_ext.py
@@ -1578,6 +1578,30 @@ class TestRSASubjectAlternativeNameExtension(object):
 
         assert 'Invalid rfc822name value' in str(exc.value)
 
+    def test_other_name(self, backend):
+        cert = _load_cert(
+            os.path.join(
+                "x509", "custom", "san_other_name.pem"
+            ),
+            x509.load_pem_x509_certificate,
+            backend
+        )
+
+        ext = cert.extensions.get_extension_for_oid(
+            x509.OID_SUBJECT_ALTERNATIVE_NAME
+        )
+        assert ext is not None
+        assert ext.critical is False
+
+        assert len(ext.value) == 1
+        assert list(ext.value)[0] == \
+            x509.OtherName(
+                x509.ObjectIdentifier("1.2.3.4"),
+                b'\x16\x0bHello World')
+
+        othernames = ext.value.get_values_for_type(x509.OtherName)
+        assert othernames == [b'\x16\x0bHello World']
+
 
 @pytest.mark.requires_backend_interface(interface=RSABackend)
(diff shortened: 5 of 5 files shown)


reaperhulk:
Thanks for the contribution @JoshData. We've been talking in IRC a bit about this and the consensus is that the best way to approach this case is to support otherName as a type that contains an OID type and a value of DER encoded bytes. That way people who know how to parse the otherName may choose to do so, but they'll remain opaque to cryptography. If you're interested in spearheading that PR that'd be great, otherwise we'll tackle it soon.

JoshData:
It may be nice to have this for the other unsupported types in the meanwhile anyway.

I'd be glad to take a stab at an OtherName type but I'm not sure I'll figure it out. Looks like I have to extend the struct definition. If you have time for a quick tip, that'd help me along.

alex:
Jenkins, ok to test.

alex:
@JoshData I've gone ahead and expanded the necessary bindings in #2111. Once that's merged in theory it's just a matter of adding the `OtherName` class, adding parsing support to that function, updating the test, and adding docs.

Feel free to ask away with any other questions!

JoshData:
Right so this is possibly over my head. It took me a good few hours just to figure out how to get your PR to take effect without `setup.py install`. Now I've got an OK working understanding of these openssl bindings things work.

I've got the type component of the otherName (an ASN1_OBJECT) as a Python string. But I don't see a way to DER-encode the value component of the otherName (a ASN1_TYPE). I see lots of i2d functions but not for ASN1_TYPEs. Tip?

JoshData:
Nevermind, I got it! `i2d_ASN1_TYPE` was very hard to find. It's a macro defined by other macros in openssl sources so it was particularly buried and Google was no help.

I've updated this PR in place (with `push --force`). I tested it with `vectors/cryptography_vectors/x509/custom/san_other_name.pem`, which now gives:

    GeneralNames([<OtherName(type_id=1.2.3.4, value=b'\x16\x0bHello World')>])>

and confirmed with `openssl asn1parse` that the DER encoding is legit:

        0:d=0  hl=2 l=  11 prim: IA5STRING         :Hello World

(although this is the first time I've ever run `asn1parse` so I have no idea what's going on there).

src/cryptography/x509.py
```diff
@@ -1044,6 +1044,31 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not self == other
 
+@utils.register_interface(GeneralName)
+class OtherName(object):
+    def __init__(self, type_id, value):
+        if not isinstance(type_id, six.text_type):
```
reaperhulk: This should actually be an `x509.ObjectIdentifier`.

src/_cffi_src/openssl/asn1.py
```diff
@@ -122,6 +122,9 @@
 ASN1_VALUE *ASN1_item_d2i(ASN1_VALUE **, const unsigned char **, long,
                           const ASN1_ITEM *);
 int ASN1_BIT_STRING_set_bit(ASN1_BIT_STRING *, int, int);
+
+/* ASN1 TYPE */
+int i2d_ASN1_TYPE(ASN1_TYPE *, unsigned char **);
```
reaperhulk: This should live in MACROS rather than FUNCTIONS

docs/x509.rst
```diff
@@ -695,6 +695,20 @@ General Name Classes
 
         :type: :class:`ObjectIdentifier`
 
+.. class:: OtherName
+
+    .. versionadded:: 0.96
```
reaperhulk: 1.0 please :)

docs/x509.rst
```diff
@@ -695,6 +695,20 @@ General Name Classes
...
+
+    This corresponds to an "otherName."  An otherName has a type identifier and a value represented in binary DER format.
+
+    .. attribute:: type_id
+
+        :type: :term:`text`
+
+    .. attribute:: value
+
+        :type: :term:`bytes`
```
reaperhulk: No need to wrap this in term (term requires its appearance in the glossary and causes the docs to fail to build properly).

reaperhulk:
Could you add a test to test_x509_ext that checks for otherName? We have `san_other_name.pem` as a test vector (currently used in `test_unsupported_other_name`) so you should be able to add a test to confirm that it is successfully parsing the data.

We'll also need to add a new test using something like `ediPartyName` to cover the UnsupportedGeneralNameType case, but that can be in a separate PR (and this one can be rebased once that merges).

Thanks for your work so far!

reaperhulk:
jenkins, ok to test.

JoshData:
Per your code comments-

> This should actually be an x509.ObjectIdentifier.

Got it. It now looks like:

    <OtherName(type_id=<ObjectIdentifier(oid=1.2.3.4, name=Unknown OID)>, value=b'\x16\x0bHello World')>

> This should live in MACROS rather than FUNCTIONS

Got it.

> 1.0 please :)
> No need to wrap this in term (term requires its appearance in the glossary and causes the docs to fail to build properly).

Got it (I think).

> Could you add a test to test_x509_ext that checks for otherName? 

Got it.

PR updated in place.

I wonder though if OtherName.value should return a tuple of the ObjectIdentifier and DER bytes, rather than just the DER bytes, so that get_values_for_type returns all of the OtherName's information?

src/cryptography/x509.py
```diff
@@ -1044,6 +1044,31 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not self == other
 
+@utils.register_interface(GeneralName)
+class OtherName(object):
+    def __init__(self, type_id, value):
+        if not isinstance(type_id, ObjectIdentifier):
+            raise TypeError("type_id must be a unicode string")
```
alex: The error message should say "ObjectIdentifier"

src/cryptography/x509.py
```diff
@@ -1044,6 +1044,31 @@ def __eq__(self, other):
...
+        if not isinstance(value, six.binary_type):
```
alex: You don't need to use `six.binary_type`, `bytes` is sufficient.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -157,6 +167,10 @@ def _decode_general_name(backend, gn):
             return x509.RFC822Name(
                 parts[0] + u"@" + idna.decode(parts[1])
             )
+    elif gn.type == backend._lib.GEN_OTHERNAME:
+        type_id = _obj2txt(backend, gn.d.otherName.type_id)
+        value = _asn1_to_der(backend, gn.d.otherName.value)
+        return x509.OtherName(x509.ObjectIdentifier(type_id), value)
     else:
         # otherName, x400Address or ediPartyName
```
alex: This comment should be udpated.

alex:
There's a handful of flake8 issues: https://travis-ci.org/pyca/cryptography/jobs/69610066

overall this patch is shaping up nciely, thanks!

JoshData:
OK. I made those changes, made my test a new test rather than replacing the existing `test_unsupported_other_name` to avoid a conflict with #2114, and rebased on master to get #2114 so the tests pass.

(If I had known this would take so many fixes I would have held off on the `push --force` until the end. Sorry about all of the clutter. Hopefully this is it...)

reaperhulk:
@JoshData In PRs we're happy to let the contributor do whatever they want. We can follow the PR so it's no problem. :)

alex:
Jenkins, add to whitelist.

alex:
add to whitelist

reaperhulk:
We're almost there! :smile:

Two things left that I see:
* We have several uncovered lines and missed branches in the `OtherName` class. If you take a look in `test_x509_ext` you can see classes for testing the various general names (e.g. the `TestRegisteredID` class). Looks like you'll need tests for invalid type_id, invalid value, equality, not equal, and repr.
* The docs job is failing spell check due to "otherName". I'd suggest handling this by putting double back ticks around it.

JoshData:
Cool cool. New commit (not squashed this time) added.

tests/test_x509_ext.py
```diff
@@ -1578,6 +1615,30 @@ def test_invalid_rfc822name(self, backend):
...
+        assert ext.critical is False
+
+        assert len(ext.value) == 1
+        assert list(ext.value)[0] == \
+            x509.OtherName(
+                x509.ObjectIdentifier("1.2.3.4"),
+                b'\x16\x0bHello World')
+
+        othernames = ext.value.get_values_for_type(x509.OtherName)
+        assert othernames == [b'\x16\x0bHello World']
```
alex: Hmm, not including the OID is a pretty huge issue here. What do you think about special casing `OtherName` to return the `OtherName` instance?

src/cryptography/x509.py
```diff
@@ -1045,6 +1045,34 @@ def __ne__(self, other):
...
+            raise TypeError("value must be a binary string")
+
+        self._type_id = type_id
+        self._value = value
+
+    type_id = utils.read_only_property("_type_id")
+    value = utils.read_only_property("_value")
+
+    def __repr__(self):
+        return "<OtherName(type_id={0}, value={1})>".format(
```
alex: This should be written `value={1!r}`.

tests/test_x509_ext.py
```diff
@@ -1147,6 +1147,43 @@ def test_ne(self):
...
+            x509.OtherName(b"notanobjectidentifier", b"derdata")
+
+        with pytest.raises(TypeError):
+            x509.OtherName(x509.ObjectIdentifier("1.2.3.4"), "notderdata")
+
+    def test_repr(self):
+        gn = x509.OtherName(x509.ObjectIdentifier("1.2.3.4"), b"derdata")
+        assert repr(gn) == (
+            "<OtherName(type_id=<ObjectIdentifier(oid=1.2.3.4, "
+            "name=Unknown OID)>, value=b'derdata')>"
```
alex: This will need a branch on `six.PY3` since there won't be a `b` prefix on python2 (there's a few other examples of this around)

tests/test_x509_ext.py
```diff
@@ -1147,6 +1147,43 @@ def test_ne(self):
         assert gn != object()
 
 
+class TestOtherName(object):
+    def test_invalid_args(self):
+        with pytest.raises(TypeError):
+            x509.OtherName(b"notanobjectidentifier", b"derdata")
+
+        with pytest.raises(TypeError):
+            x509.OtherName(x509.ObjectIdentifier("1.2.3.4"), "notderdata")
```
alex: This needs a `u` prefix on `"notderdata"` for python3.

tests/test_x509_ext.py
JoshData: Could also be done by having `value` return `self` in OtherName?

tests/test_x509_ext.py
JoshData: Is it worth modifying repr so it returns an identical string on Py2 and Py3?

tests/test_x509_ext.py
alex: Meaning `.type_id` an OID, `.<something creative>` be the der data, and
`.value` be `self`? I'm ok with that.

On Sun, Jul 5, 2015 at 5:56 PM, Joshua Tauberer <notifications@github.com>
wrote:

> In tests/test_x509_ext.py
> <https://github.com/pyca/cryptography/pull/2108#discussion_r33899392>:
>
> > +        )
> > +
> > +        ext = cert.extensions.get_extension_for_oid(
> > +            x509.OID_SUBJECT_ALTERNATIVE_NAME
> > +        )
> > +        assert ext is not None
> > +        assert ext.critical is False
> > +
> > +        assert len(ext.value) == 1
> > +        assert list(ext.value)[0] == \
> > +            x509.OtherName(
> > +                x509.ObjectIdentifier("1.2.3.4"),
> > +                b'\x16\x0bHello World')
> > +
> > +        othernames = ext.value.get_values_for_type(x509.OtherName)
> > +        assert othernames == [b'\x16\x0bHello World']
>
> Could also be done by having value return self in OtherName?
>
> —
> <https://github.com/pyca/cryptography/pull/2108/files#r33899392>.
>



-- 

tests/test_x509_ext.py
alex: Everywhere else we use the native bytes repr, and I think consistency with
that makes sense.

On Sun, Jul 5, 2015 at 5:57 PM, Joshua Tauberer <notifications@github.com>
wrote:

> In tests/test_x509_ext.py
> <https://github.com/pyca/cryptography/pull/2108#discussion_r33899414>:
>
> > @@ -1147,6 +1147,43 @@ def test_ne(self):
> >          assert gn != object()
> >
> >
> > +class TestOtherName(object):
> > +    def test_invalid_args(self):
> > +        with pytest.raises(TypeError):
> > +            x509.OtherName(b"notanobjectidentifier", b"derdata")
> > +
> > +        with pytest.raises(TypeError):
> > +            x509.OtherName(x509.ObjectIdentifier("1.2.3.4"), "notderdata")
> > +
> > +    def test_repr(self):
> > +        gn = x509.OtherName(x509.ObjectIdentifier("1.2.3.4"), b"derdata")
> > +        assert repr(gn) == (
> > +            "<OtherName(type_id=<ObjectIdentifier(oid=1.2.3.4, "
> > +            "name=Unknown OID)>, value=b'derdata')>"
>
> Is it worth modifying repr so it returns an identical string on Py2 and
> Py3?
>
> —
> <https://github.com/pyca/cryptography/pull/2108/files#r33899414>.
>



-- 

tests/test_x509_ext.py
JoshData: Err, I forgot that value was the name of the der property and not just an abstract method for getting the value. Nevermind. Special-casing sounds better.

JoshData:
Pretty sure I got it all.

JoshData:
Thanks for working through this with me, guys!

reaperhulk:
@JoshData thanks for your contribution. Feel free to add yourself to the AUTHORS.rst in a new PR :)

PR #2109, opened 2015-07-04
add some documentation around the requirement for xcode cli tools on mac

### LINKED ISSUE #1700: Document OS X installation
We implicitly assume our OS X users already have a compiler and the ffi headers (`xcode-select --install`). This is sometimes an unsafe assumption, especially as our project gets more popular. Let's document how to install on OS X.

reaperhulk:
fixes #1700

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -128,7 +128,9 @@ build.
 Building cryptography on OS X
 -----------------------------
 
-To install the Xcode command line tools on OS X open a terminal window and run:
+Building cryptography requires the presence of a C compiler and development
+headers. On OS X this is typically provided by Apple's Xcode development tools.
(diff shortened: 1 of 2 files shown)


docs/installation.rst
```diff
@@ -125,8 +125,21 @@ You'll also need to generate your own ``openssl.ld`` file. For example::
 You should replace the version string on the first line as appropriate for your
 build.
 
+Building cryptography on OS X
+-----------------------------
+
+To install the Xcode command line tools on OS X open a terminal window and run:
```
alex: This should probably explain _why_ you want to isntall xcode, perhaps lead with "Building cryptography requires having a compiler installed. [...]"

alex:
looks good.

PR #2110, opened 2015-07-04
remove docs for WithNumbers classes that are removed in 1.0

reaperhulk:
We deprecated these in 0.8 and removed them at the start of the 1.0 development cycle.


alex:
`CHANGELOG` needs to be updated

PR #2124, opened 2015-07-06
Fixed #2120 -- added __hash__ to x509.Cert

alex:

PR #2126, opened 2015-07-07
Fixed #2121 -- added __eq__ and __ne__ to CSRs

### LINKED ISSUE #2121: x509.CSR: implement equality
Since `x509.Certificate.__eq__` is implemented, I would expect a similar behavior for `x509.CertificateSigningRequest.__eq__`. Thanks!

```
(venv)jakub@iks:~/dev/letsencrypt/letsencrypt/acme/jose/testdata$ pip show cryptography
---
Name: cryptography
Version: 0.9.2
Location: /home/jakub/dev/letsencrypt/letsencrypt/venv/lib/python2.7/site-packages
Requires: idna, pyasn1, six, setuptools, enum34, ipaddress, cffi
(venv)jakub@iks:~/dev/letsencrypt/letsencrypt/acme/jose/testdata$ cat t2.py
from cryptography import x509

with open("cert.der") as cert_file:
    data = cert_file.read()
    cert1 = x509.load_der_x509_certificate(data, default_backend())
    cert2 = x509.load_der_x509_certificate(data, default_backend())

with open("csr.der") as req_file:
    data = req_file.read()
    req1 = x509.load_der_x509_csr(data, default_backend())
    req2 = x509.load_der_x509_csr(data, default_backend())

print repr(cert1)
print repr(req1)
print cert1 == cert2, req1 == req2
(venv)jakub@iks:~/dev/letsencrypt/letsencrypt/acme/jose/testdata$ python t2.py
<cryptography.hazmat.backends.openssl.x509._Certificate object at 0x7f44e221a550>
<cryptography.hazmat.backends.openssl.x509._CertificateSigningRequest object at 0x7f44e221a710>
True False
```

alex:
OpenSSL doesn't have X509_REQ_cmp, so I went with just comparing the DER bytes, is there a better way?

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -701,6 +701,14 @@ class _CertificateSigningRequest(object):
         self._backend = backend
         self._x509_req = x509_req
 
+    def __eq__(self, other):
+        if not isinstance(other, _CertificateSigningRequest):
+            return NotImplemented
+
+        self_bytes = self.public_bytes(serialization.Encoding.DER)
+        other_bytes = other.public_bytes(serialization.Encoding.DER)
+        return self_bytes == other_bytes
+
     def public_key(self):
         pkey = self._backend._lib.X509_REQ_get_pubkey(self._x509_req)
         assert pkey != self._backend._ffi.NULL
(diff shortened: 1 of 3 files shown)


reaperhulk:
The DER representation is canonical so that'll work fine. You could just compare fingerprints though (that's the hash of the DER bytes)

alex:
@reaperhulk a) CSR doesn't have a fingerprint method, b) what would the advantage of comparing a has be? seems like it would just be slower

reaperhulk:
Since there's no fingerprint method I'd say there's no advantage.

kuba:
what about `__hash__`, similarly to #2124?

alex:
When this lands I'll do a follow up PR, one at a time :-)

alex:
(I filed it now #2127)

Issue #2117, opened 2015-07-05
Add support for building CSRs with all general name types


alex:
- [x] `RFC822Name`
- [x] `DNSName`
- [x] `UniformResourceIdentifier`
- [x] `DirectoryName`
- [x] `RegisteredID`
- [x] `IPAddress`
- [x] `OtherName`

PR #2134, opened 2015-07-10
support DirectoryName encoding for general names

reaperhulk:
refs #2117

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -171,6 +171,15 @@ def _encode_subject_alt_name(backend, san):
             )
             assert obj != backend._ffi.NULL
             gn.d.registeredID = obj
+        elif isinstance(alt_name, x509.DirectoryName):
+            gn = backend._lib.GENERAL_NAME_new()
+            assert gn != backend._ffi.NULL
+            name = _encode_name(backend, alt_name.value)
+            # _encode_name registers the X509_NAME for gc so we'll duplicate
+            # a new one that is not gc'd for the struct
+            name = backend._lib.X509_NAME_dup(name)
+            gn.type = backend._lib.GEN_DIRNAME
+            gn.d.directoryName = name
         else:
             raise NotImplementedError(
                 "Only DNSName and RegisteredID supported right now"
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -171,6 +171,15 @@ def _encode_subject_alt_name(backend, san):
             )
             assert obj != backend._ffi.NULL
             gn.d.registeredID = obj
+        elif isinstance(alt_name, x509.DirectoryName):
+            gn = backend._lib.GENERAL_NAME_new()
+            assert gn != backend._ffi.NULL
+            name = _encode_name(backend, alt_name.value)
+            # _encode_name registers the X509_NAME for gc so we'll duplicate
+            # a new one that is not gc'd for the struct
+            name = backend._lib.X509_NAME_dup(name)
```
alex: Wouldn't just adding an `_encode_name_nogc` or something be easier :-)

reaperhulk: What about changing _encode_name to take a `gc` boolean arg? No default, so current call sites would become `_encode_name(backend, <data>, gc=True)`

Whatever pattern we choose will be duplicated for `_encode_asn1_str` and maybe `_txt2obj` as well.

alex: Not wild about a `gc` param; I'd prefer two functions (obviously one which
calls the other)

On Sat, Jul 11, 2015 at 12:29 AM, Paul Kehrer <notifications@github.com>
wrote:

> In src/cryptography/hazmat/backends/openssl/backend.py
> <https://github.com/pyca/cryptography/pull/2134#discussion_r34409541>:
>
> > @@ -171,6 +171,15 @@ def _encode_subject_alt_name(backend, san):
> >              )
> >              assert obj != backend._ffi.NULL
> >              gn.d.registeredID = obj
> > +        elif isinstance(alt_name, x509.DirectoryName):
> > +            gn = backend._lib.GENERAL_NAME_new()
> > +            assert gn != backend._ffi.NULL
> > +            name = _encode_name(backend, alt_name.value)
> > +            # _encode_name registers the X509_NAME for gc so we'll duplicate
> > +            # a new one that is not gc'd for the struct
> > +            name = backend._lib.X509_NAME_dup(name)
>
> What about changing _encode_name to take a gc boolean arg? No default, so
> current call sites would become _encode_name(backend, <data>, gc=True)
>
> Whatever pattern we choose will be duplicated for _encode_asn1_str and
> maybe _txt2obj as well.
>
> —
> <https://github.com/pyca/cryptography/pull/2134/files#r34409541>.
>



-- 

PR #2136, opened 2015-07-11
Encode othername

reaperhulk:
refs #2117 

depends on #2134 and #2135


alex:
ready for a rebase :-)

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -199,6 +199,28 @@ def _encode_subject_alt_name(backend, san):
...
+            other_name = backend._lib.OTHERNAME_new()
+            assert other_name != backend._ffi.NULL
+
+            type_id = backend._lib.OBJ_txt2obj(
+                alt_name.type_id.dotted_string.encode('ascii'), 1
+            )
+            assert type_id != backend._ffi.NULL
+            data = backend._ffi.new("unsigned char[]", alt_name.value)
+            data_ptr_ptr = backend._ffi.new("unsigned char **")
+            data_ptr_ptr[0] = data
```
alex: Why aren't we just leaving this null and letting OpenSSL fill it in here?

reaperhulk: Because despite being a ** we actually do have to provide the data. It's a call to a `d2i` function so you pass the bytes and it returns an object.

The function signature is weird because it should obviously just be a * but that's not how it was defined for some reason.

PR #2137, opened 2015-07-11
Encode rfc822name

reaperhulk:
refs #2117 

depends on #2134, #2135, #2136

PR #2141, opened 2015-07-12
expand RFC822Name to validate and (internally) IDNA encode

reaperhulk:
This will be used in the CSR builder. refs #2137

### DIFF
--- a/src/cryptography/x509.py
+++ b/src/cryptography/x509.py
@@ -6,7 +6,10 @@ from __future__ import absolute_import, division, print_function
 
 import abc
 import ipaddress
+from email.utils import parseaddr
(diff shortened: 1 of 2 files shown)


src/cryptography/x509.py
```diff
@@ -901,7 +904,24 @@ def __init__(self, value):
         if not isinstance(value, six.text_type):
             raise TypeError("value must be a unicode string")
 
+        name, address = parseaddr(value)
+        parts = address.split(u"@")
+        if name or len(parts) > 2 or not address:
```
alex: `len(parts) > 2` doesn't seem to actually be tested?

PR #2150, opened 2015-07-12
support URI general name encoding

reaperhulk:
depends on #2137 and #2149 

refs #2117


alex:
needs a rebase

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -12,6 +12,8 @@
 
 import six
 
+from six.moves import urllib_parse
```
alex: This is not used.

PR #2151, opened 2015-07-13
temporarily pin ipaddress to 1.0.7 to avoid a pypy bug

reaperhulk:
This will unblock our CI while we get this fixed upstream.


alex:
1.10 was just released fixing this; how do we want to handle the version requirement.

PR #2154, opened 2015-07-14
Fixed #2091 -- describe how to generate a CSR.

### LINKED ISSUE #2091: Document a recommended way to generate an x509 key pair.
A link or small sentence (perhaps in the FAQ) about the best way to go about generating a certificate would be nice.

alex:


glyph:
From reviewing https://jenkins.cryptography.io/job/cryptography-pull-request-docs/270/artifact/cryptography/docs/_build/html/x509/tutorial.html -

One thing about the `CertificateSigningRequestBuilder` is that it seems to put the wrong thing first.  `OID_COUNTRY_NAME`, `OID_STATE_OR_PROVINCE_NAME`, etc, are never verified and rarely presented to users.  Instead, I would make a CSR with just one `DNSName` first.

I also feel like there ought to be a clearly linked reference document.  How do I know what attributes are available to a `Name`?  Is order important or is this actually a `set`?  What other `OID`s can I use?  Which ones can I leave out?  Can I put things other than a `NameAttribute` into a name?  If a CA requires my technical administrator's email address to be in the CSR, where does that go?  Why `hashes.SHA256()`, what other hashes might be good, and why do I need to make a decision about that rather than getting a default?  (Actually, `serialization.BestAvailableEncryption` might be a good pattern to follow for hashes, too.)  Et cetera, et cetera, et cetera.

The structure of a CSR (as the structure of a certificate itself) is deeply complicated and weird, and I think that this tutorial needs to have a good exploration of building up from something minimal to something more complex, and it should do this in a way that explains how to look up more information about each step.  One of the big problems with "do SSL on your server" tutorials is that they copy-paste around huge numbers of unnecessary fields, and people think they are ritualistically necessary.  My favorite side-effect of this is that following only recommended instructions from trusted CAs, I have more than once operated a server with a signed certificate belonging to "Internet Widgets Pty. Ltd."

Personally, I like starting with a StartSSL-style CA, so you can show a CSR that basically has nothing in it but a key, and then build up more identifying information from there, one element at a time, so that people will end up with something that suits their needs, and not a template that has every SSL-secured site in the world claiming to be from "My Company" in San Francisco CA.

docs/x509/tutorial.rst
```diff
@@ -0,0 +1,79 @@
...
+    >>> # Write our key to disk for safe keeping
+    >>> with open("path/to/store/key.pem") as f:
+    ...     f.write(key.private_bytes(
+    ...         encoding=serialization.Encoding.PEM,
+    ...         format=serialization.PrivateFormat.TraditionalOpenSSL,
+    ...         encryption_algorithm=serialization.BestAvailableEncryption(b"passphrase"),
+    ...     ))
+
+If you've already generated a key you can load it with
+:func:`~cryptography.hazmat.primitives.serialization.load_pem_public_key`.
```
reaperhulk: This should be load_pem_private_key?

alex: fixed

On Thu, Jul 16, 2015 at 6:59 AM, Paul Kehrer <notifications@github.com>
wrote:

> In docs/x509/tutorial.rst
> <https://github.com/pyca/cryptography/pull/2154#discussion_r34776210>:
>
> > +    >>> # Generate our key
> > +    >>> key = rsa.generate_private_key(
> > +    ...     public_exponent=65537,
> > +    ...     key_size=2048,
> > +    ...     backend=default_backend()
> > +    ... )
> > +    >>> # Write our key to disk for safe keeping
> > +    >>> with open("path/to/store/key.pem") as f:
> > +    ...     f.write(key.private_bytes(
> > +    ...         encoding=serialization.Encoding.PEM,
> > +    ...         format=serialization.PrivateFormat.TraditionalOpenSSL,
> > +    ...         encryption_algorithm=serialization.BestAvailableEncryption(b"passphrase"),
> > +    ...     ))
> > +
> > +If you've already generated a key you can load it with
> > +:func:`~cryptography.hazmat.primitives.serialization.load_pem_public_key`.
>
> This should be load_pem_private_key?
>
> —
> <https://github.com/pyca/cryptography/pull/2154/files#r34776210>.
>



-- 

reaperhulk:
I'm going to merge this and open a new issue with the body of @glyph's comment for future improvement.

Issue #2061, opened 2015-06-23
[BLOCKER] Conditional names broken


reaperhulk:
When we switched to `set_source` in #1986 we broke conditional name binding. Instead of being removed as the lib is instantiated the NULL function pointers are available and, if called, cause a segfault.

reaperhulk:
The fix for this should also include a test that verifies that conditional names are removed. We can reliably test this by creating a conditional that will always be false.

reaperhulk:
I've been thinking about this for a bit and haven't come up with a silver bullet. Here's a potential solution though:

Pull the `Cryptography_HAS_*` declarations into a separate section and compile that. Use the resulting lib object to check for support and append the dependent types/functions/macros to the strings for the main compilation stage. Once that's done, compile the "real" lib objects and be done.

There are some challenges for this. For instance, we'll need to figure out how to delete the dylib generated by that first compilation (Donald?). Additionally, the `Cryptography_HAS_*` values are used by PyOpenSSL in various places (and potentially other libraries) so we will need to continue to have them in the main library.

I don't particularly like this solution, but I'm not seeing a better one at the moment. /cc @alex @dstufft

dstufft:
It's basically reinventing shit like the preprocessor and autotools, and since folks haven't figured out a better way than those solutions for that... I doubt we're probably going to either.

reaperhulk:
If we do it this way what's the best way to delete the temporary module we build?

dstufft:
I would just use the old API for it, since it should only exist in the build scripts and won't be mentioned in the setup.py I don't think you need to delete it, it'll get cleaned up with the rest, but I may be wrong on that.

alex:
Alternative -- wrap `lib` with something that doesn't allow access to things in `CONDITIONAL`, at runtime.

PR #2156, opened 2015-07-15
Proof of Concept: Conditional CDEF

dstufft:
This should not be merged! It is a proof of concept only.

This requires having CFFI installed from my fork on bitbucket (particularly [this commit](https://bitbucket.org/dstufft/cffi/commits/f7b3e65833d55c95fc1fe8d69c95f6c49fbf0d9f)).

This essentially works by modifying CFFI to give us a hook in the ``build_ext`` command when the FFI object exists but hasn't yet had it's C sources generated and we also have access to the ``build_ext`` command class so we can access things like libraries, include dirs, etc that were passed in via ``python setup.py build_ext ...``.

Essentially this inverts the old logic, instead of defining everything and then deleting some names from runtime, this only defines the unconditional names unconditionally and then defines a test and additional data to call ``ffi.cdef()`` with if that test passes. Then using the hook that I've modified CFFI to have we generate a small cffi module, compile it, check the value, and then conditionally add the items or not.

Some things I know this needs:

* It needs to pass all of the values in through ``set_source`` and not just libraries (like include dirs and such).
* We probably don't need to do the ``sys.path`` dance, just ensuring the tempdir is on there and then leaving it without cleaning up should be fine.
* Need to figure out what we want the data structure that defines conditional names should look like for real.
* Maybe all the ``#include`` statements should be added to each temporary module instead of needing to add them to each conditional test.
* Of course, actually move conditional names to this system.

I don't have the time right now to actually push this forward further, but hopefully this is enough that @reaperhulk or someone can pick up the mantle and take this forward.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,7 @@ import platform
 import subprocess
 import sys
 from distutils.command.build import build
+from distutils.command.build_ext import build_ext
 
 import pkg_resources
 
@@ -97,6 +98,30 @@ class PyTest(test):
         test_args = [os.path.join(base_dir, "tests")]
         errno = pytest.main(test_args)
         sys.exit(errno)
+
+
+class ConditionalCFFIBuildExt(build_ext):
+
+    def pre_run(self, ext, ffi):
+        from cffi import FFI
+        count = 0
+        for conditional in getattr(ffi, "_conditional_cdefs", []):
+            for attribute, (tc, rc) in conditional.items():
+                modname = "_cryptography_cffi_test_{0}".format(count)
+                tffi = FFI()
+                tffi.cdef("static const int {0};".format(attribute))
+                tffi.set_source(
+                    modname, tc,
+                    libraries=self.get_libraries(ext),
+                )
+                tmpdir = os.path.join(self.build_temp, "conditionals")
+                tffi.compile(tmpdir=tmpdir)
+                sys_path = sys.path[:]
+                sys.path.append(tmpdir)
+                mod = __import__(modname)
+                sys.path = sys_path
+                if getattr(mod.lib, attribute, None):
+                    ffi.cdef(rc)
 
 
 def keywords_with_side_effects(argv):
@@ -207,6 +232,7 @@ def keywords_with_side_effects(argv):
         return {
             "setup_requires": setup_requirements,
             "cmdclass": {
+                "build_ext": ConditionalCFFIBuildExt,
                 "test": PyTest,
             },
             "cffi_modules": cffi_modules
--- a/src/_cffi_src/openssl/aes.py
+++ b/src/_cffi_src/openssl/aes.py
@@ -23,13 +23,6 @@ int AES_set_decrypt_key(const unsigned char *, const int, AES_KEY *);
 """
 
 MACROS = """
-/* these can be moved back to FUNCTIONS once we drop support for 0.9.8h.
-   This should be when we drop RHEL/CentOS 5, which is on 0.9.8e. */
-int AES_wrap_key(AES_KEY *, const unsigned char *, unsigned char *,
-                 const unsigned char *, unsigned int);
-int AES_unwrap_key(AES_KEY *, const unsigned char *, unsigned char *,
-                   const unsigned char *, unsigned int);
-
 /* The ctr128_encrypt function is only useful in 0.9.8. You should use EVP for
    this in 1.0.0+. It is defined in macros because the function signature
    changed after 0.9.8 */
@@ -45,10 +38,6 @@ CUSTOMIZATIONS = """
 static const long Cryptography_HAS_AES_WRAP = 1;
 #else
 static const long Cryptography_HAS_AES_WRAP = 0;
-int (*AES_wrap_key)(AES_KEY *, const unsigned char *, unsigned char *,
-                    const unsigned char *, unsigned int) = NULL;
-int (*AES_unwrap_key)(AES_KEY *, const unsigned char *, unsigned char *,
-                      const unsigned char *, unsigned int) = NULL;
 #endif
 
 """
@@ -58,4 +47,25 @@ CONDITIONAL_NAMES = {
         "AES_wrap_key",
         "AES_unwrap_key",
     ],
+}
+
+
+CONDITIONAL_NAMES2 = {
+    "Cryptography_HAS_AES_WRAP": (
+        """
+/* OpenSSL 0.9.8h+ */
+#include <openssl/opensslv.h>
+#if OPENSSL_VERSION_NUMBER >= 0x0090808fL
+static const long Cryptography_HAS_AES_WRAP = 1;
+#else
+static const long Cryptography_HAS_AES_WRAP = 0;
+#endif
+        """,
+        """
+int AES_wrap_key(AES_KEY *, const unsigned char *, unsigned char *,
+                 const unsigned char *, unsigned int);
+int AES_unwrap_key(AES_KEY *, const unsigned char *, unsigned char *,
+                   const unsigned char *, unsigned int);
+        """,
+    ),
 }
--- a/src/_cffi_src/utils.py
+++ b/src/_cffi_src/utils.py
@@ -6,7 +6,17 @@ from __future__ import absolute_import, division, print_function
 
 import sys
 
-from cffi import FFI
+from cffi import FFI as _FFI
+
+
+class FFI(_FFI):
+
+    def __init__(self, *args, **kwargs):
+        super(FFI, self).__init__(*args, **kwargs)
+        self._conditional_cdefs = []
+
+    def conditional_cdefs(self, conditionals):
+        self._conditional_cdefs.extend(conditionals)
 
 
 def build_ffi_for_binding(module_name, module_prefix, modules, pre_include="",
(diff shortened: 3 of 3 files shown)


dstufft:
And yes, I did just reinvent some part of autotools.

public:
Why is this preferable to the current arrangement?

reaperhulk:
@public as of our switch to set_source in cffi 1.0 the current arrangement doesn't work. See #2061

public:
I see. I like the idea of doing something more autotoolsy. I'm not a massive fan of this particular API but it does seem to do the job.

I guess ideally I'd want something that could actually understand the #ifdefs and do all the glue for you.

dstufft:
@public The current arrangement doesn't work right now. When we switched to CFFI we lost the ability to do conditional names at all, so right now if you call a conditional name you don't get an attributeerror you get a segfault.

dstufft:
Also, it's possible to extend this out to act more like a preprocessor if we wanted. This was my "it's 1am and I wanted to make something functional so that hopefully someone else can pick up the torch from there" Proof of Concept :)

reaperhulk:
I'm writing these thoughts here as an infodump for what we're trying to accomplish:

cryptography compiles against many different versions of OpenSSL. Newer versions may have new functions/types that were not previously available. With cffi this creates a problem because the `cdef` method expects you to *a priori* know what types and functions are available. In the pre-cffi 1.0 world with `verify` we worked around this by creating a pattern where we would add something like this to the verify source:

```c
#ifdef OPENSSL_NO_EC
static const long Cryptography_HAS_EC = 0;

typedef void EC_KEY;
void (*EC_KEY_free)(EC_KEY *) = NULL;
EC_KEY *(*EC_KEY_new)(void) = NULL;
size_t (*EC_get_builtin_curves)(EC_builtin_curve *, size_t) = NULL;
/* and so on. lots more of these */
#else
static const long Cryptography_HAS_EC = 1;
#endif
```

Then after `verify` we'd have a `lib` object we could loop over and `delattr` each of the null function pointers. This was inelegant, but allowed us to use the C preprocessor to determine what version of OpenSSL we were compiling against and alter our support accordingly.

Unfortunately, the cffi 1.0+ `set_source` out-of-line API mode does not allow this pattern (we can't `delattr` items on the new object). This PR represents a potential avenue to resolve the problem. Ideally, a future release of cffi could support this sort of requirement directly, but what that looks like is not clear at this stage.

reaperhulk:
I'd like to see what @arigo thinks about these options and what could potentially be done in cffi to make this easier.

arigo:
I'm sure you can easily replicate the "delattr" hack by creating a new instance "lib" of a custom class and copying all attributes there, minus the ones that you used to "delattr".  But it remains a hack.  Maybe better would be to add official support for "#if" conditions in the cdef.  It should be reasonably possible and clean to generate a C file that uses the same "#if" conditions in order to only export what these "#if"s protect.

reaperhulk:
Yeah if an official way to do this via `#if` statements was created that'd be great. :) In the mean time the copy hack might be our best option (the approach in this PR already requires changes to cffi, but if we're going to do that I'd rather wait for `#if` in `cdef` :smile:).

Of course, we moved the conditional names to some files we no longer install, so we'll have to move them back into the `cryptography.*` namespace if we want to create a `ConditionalLib` object and copy the valid attrs over.

arigo:
There are also options like: if a C function is not defined, you can write it in the set_source() not as ``int(*foo)(int) = 0;`` but as ``int foo(int x) { errno=ENOSYS; return -1; }``, for example, so that it is still actually callable.

setup.py
```diff
@@ -99,6 +100,30 @@ def run_tests(self):
...
+                    modname, tc,
+                    libraries=self.get_libraries(ext),
+                )
+                tmpdir = os.path.join(self.build_temp, "conditionals")
+                tffi.compile(tmpdir=tmpdir)
+                sys_path = sys.path[:]
+                sys.path.append(tmpdir)
+                mod = __import__(modname)
+                sys.path = sys_path
+                if getattr(mod.lib, attribute, None):
```
alex: What's the default `None` for here?

alex:
The more I look at this PR, the more it looks like a pretty solid solution to me.

@dstufft are there any downsides I'm misisng?

reaperhulk:
@alex Primary downside is that the approach in this PR would require a cffi change without being as elegant as an `#if` in `cdef` change (although the latter is strictly hypothetical while the former is already present in dstufft's fork).

@arigo Unfortunately we have downstream consumers that use the presence/absence of an attribute as an indicator if the underlying OpenSSL supports a given function. Certainly erroring on the function is better than a null function pointer crash, but we do need to remove the function entirely to not break our API contract.

I'm kind of inclined to go with building a ConditionalLib object and just copying the attributes over for now? Unless @arigo thinks a conditional `cdef` is in the near future?

dstufft:
I personally think that this is a better solution than a ``ConditionalLib`` because it involves less work at runtime which will keep our runtime faster at the expensive of a longer build time. That is of course, assuming that @arigo would accept the small change made to CFFI for it. That change isn't really tied to conditional names, it's just providing a "before" hook in the compilation which could be used for many things. It's a nice general purpose thing that we could just happen to use for this purpose.

If at some point CFFI gets a conditional ``cdef`` then we can easily migrate to that, the runtime code won't have to change at all to cope with that fact, only the build code would.

dstufft:
Oh, and to answer @alex, the downside is basically a more complicated build which will require at least two passes (if not 1 pass per test depending exactly on implementation) which will result in a slower build.

setup.py
dstufft: I just have a habit of explicitly adding the ``None`` to ``getattr`` (even though it's not needed) to be explicit about the default (and for symmetry with ``{}.get(..., None)``),

reaperhulk:
@dstufft since this is a generally useful hook would you mind opening a PR against cffi so we can get it landed and released? :)

arigo:
@dstufft I can accept this change to cffi, but also, maybe it's overkill in this case: you could simply try to compile the test modules directly in your build_ffi() function.  This adds some time to run build_ffi(), as opposed to add it in the execution of the "build_ext" setuptools command, but as build_ffi() should not be called after installation anyway, it seems reasonable to me.

I'm trying to implement ``#if`` support but hitting a snag, so it's unclear at this point that it will be part of the soon-to-be-released cffi 1.2.0.

public:
Why did we decide against something like what was described here https://github.com/pyca/cryptography/pull/2156#issuecomment-122211074 ?

dstufft:
@arigo The reason we don't want to just compile the test modules directly in the build_ffi function is we don't have access to the options specified via distutils then. Like if someone does ``python setup.py build_ext -L /usr/local/lib/`` then the "real" module will get compiled with ``/usr/local/lib`` added to the library paths, but the test modules would not. The small change I made was primarily to provide a hook that would give us access to the distutils options prior to the CFFI having generated the C file and prior to it needing to know what names to expect.

@public It's a possibility, however it's pushing work onto the runtime that could live at build time, this would make the runtime slower (how much slower? I don't know I haven't tested it, but less work is always faster). On top of that it would require breaking down some of the seperation we currently have between build time and runtime, at the moment the information required to implement https://github.com/pyca/cryptography/pull/2156#issuecomment-122211074 is no longer installed as part of cryptography so it would require refactoring the cffi 1.0 support in order to make sure it got installed again. All in all I think the approach used by this PR will require less changes overall both to go from the master branch to something working and to get to an eventual place that cffi has ``#if`` support.

public:
@dstufft Does it require that? I'd have thought we could detect that the function pointer was NULL and hide it from the lib without needing the original binding files? It's not like we need to know the specific names to apply this check to is it? Seems pretty unlikely we'd ever want to actually expose a callable NULL pointer.

reaperhulk:
If we iterate over dir and compare everything to NULL that would work I guess? Definitely slower but maybe a simple workaround until cdef if.

Worth a basic benchmark I think

dstufft:
It would work I suppose, I still think doing work at build time is superior to doing work at runtime though.

reaperhulk:
Iterating and comparing won't work because cffi's lib object actually doesn't show that the function is a null pointer. You don't know until you call it and it crashes.

```
In [7]: backend._lib.X509_VERIFY_PARAM_set1_host
Out[7]: <function _openssl.CompiledLib.X509_VERIFY_PARAM_set1_host>
```

So that idea won't work. We do need the conditional info back in the installed files if we intend to do that as a temporary workaround.

Incidentally looping over every attr, comparing them against null, and copying them to a new object takes so little time that I can't tell even on 1 million iterations of timeit.

arigo:
You could do this:

```
ffi.cdef("""
    int foobar(int);
    #define cffi_enable_foobar ...
""")
ffi.set_source(.., """
    #ifdef SOME_CONDITION
      #define cffi_enable_foobar 1
    #else
      #define cffi_enable_foobar 0
      int (*foobar)(int) = 0;
""")
```

and then at runtime, you look for all names "lib.cffi_enable_X" and you copy or not the name "X" depending on the value of "cffi_enable_X".  You'd have one "cffi_enable_X" per function "X" that can appear conditionally.

arigo:
@dstufft added in cffi 1.2 the pre_run() hook.  Not documented for now; may be documented in future releases.  Only works with setuptools-based setup.py and API-level out-of-line ffi's.  See b989307a75ac in https://bitbucket.org/cffi/cffi .

reaperhulk:
This was resolved by #2180

PR #2163, opened 2015-07-22
Reduce number of isinstance calls in openssl/ciphers.py.

Ayrx:
This change reduces the number of `isinstance` calls to 1 at class initialization. The previous behaviour has a `isinstance` call on every `update()`. This change results in a noticeable speed up for code with many `update()` calls.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/ciphers.py
+++ b/src/cryptography/hazmat/backends/openssl/ciphers.py
@@ -24,6 +24,8 @@ class _CipherContext(object):
         self._operation = operation
         self._tag = None
 
+        self._is_gcm = isinstance(self._mode, modes.GCM)
+
         if isinstance(self._cipher, ciphers.BlockCipherAlgorithm):
             self._block_size = self._cipher.block_size
         else:
@@ -72,7 +74,7 @@ class _CipherContext(object):
             ctx, len(cipher.key)
         )
         assert res != 0
-        if isinstance(mode, modes.GCM):
+        if self._is_gcm:
             res = self._backend._lib.EVP_CIPHER_CTX_ctrl(
                 ctx, self._backend._lib.EVP_CTRL_GCM_SET_IVLEN,
                 len(iv_nonce), self._backend._ffi.NULL
@@ -107,7 +109,7 @@ class _CipherContext(object):
         # should be taken only when length is zero and mode is not GCM because
         # AES GCM can return improper tag values if you don't call update
         # with empty plaintext when authenticating AAD for ...reasons.
-        if len(data) == 0 and not isinstance(self._mode, modes.GCM):
+        if len(data) == 0 and not self._is_gcm:
             return b""
 
         buf = self._backend._ffi.new("unsigned char[]",
@@ -124,7 +126,7 @@ class _CipherContext(object):
         # even if you are only using authenticate_additional_data or the
         # GCM tag will be wrong. An (empty) call to update resolves this
         # and is harmless for all other versions of OpenSSL.
-        if isinstance(self._mode, modes.GCM):
+        if self._is_gcm:
             self.update(b"")
 
         buf = self._backend._ffi.new("unsigned char[]", self._block_size)
@@ -133,7 +135,7 @@ class _CipherContext(object):
         if res == 0:
             errors = self._backend._consume_errors()
 
-            if not errors and isinstance(self._mode, modes.GCM):
+            if not errors and self._is_gcm:
                 raise InvalidTag
 
             assert errors
@@ -154,7 +156,7 @@ class _CipherContext(object):
             else:
                 raise self._backend._unknown_error(errors[0])
 
-        if (isinstance(self._mode, modes.GCM) and
+        if (self._is_gcm and
            self._operation == self._ENCRYPT):
             block_byte_size = self._block_size // 8
             tag_buf = self._backend._ffi.new(


alex:
Can you show a benchmark and timing results with before/after?

Ayrx:
This is my test code.

```python
import timeit

from cryptography.hazmat.primitives.ciphers import algorithms, modes
from cryptography.hazmat.primitives.ciphers.base import Cipher

KEY = b'\xf8\xb3U)b\x82\xe5\x99\xe9\xc4A\xa1\x8eT\xb3\xac'
READ_CHUNK = 32768
IV = b"1234567890123456"

def crypto_bench():
    aes_cipher = Cipher(algorithms.AES(KEY), modes.CTR(IV), default_backend())
    aes = aes_cipher.encryptor()
    with open("/dev/zero", "rb") as f:
        for i in range(0, 100000):
            aes.update(f.read(READ_CHUNK))

if __name__ == "__main__":
    print(min(timeit.Timer(
        "crypto_bench()", setup="from test import crypto_bench"
    ).repeat(2, 100)))
```

Results on master:

    $ python test.py                                                                
    173.32433925295481

Results with this PR:

    $ python test.py
    172.19690058199922

The speed up isn't *large*, but it's pretty consistent. Worth it (imo) for a simple change that doesn't add any complexity.

public:
Seems fair. Might be interesting to see how doing all of these `if gcm` checks constantly effects performance too. If we should find a nice way to split the GCM stuff into a separate class that might make the code nicer too.

alex:
Personally, I don't think this change is worth it for <1% improvement.
Would be interested to see what @public's idea looks like though.

On Wed, Jul 22, 2015 at 11:26 AM, Alex Stapleton <notifications@github.com>
wrote:

> Seems fair. Might be interesting to see how doing all of these if gcm
> checks constantly effects performance too. If we should find a nice way to
> split the GCM stuff into a separate class that might make the code nicer
> too.
>
> —
> <https://github.com/pyca/cryptography/pull/2163#issuecomment-123757728>.
>



-- 

reaperhulk:
@Ayrx a profile might be helpful here to see if there are any useful hot spots.

src/cryptography/hazmat/backends/openssl/ciphers.py
```diff
@@ -107,7 +109,7 @@ def update(self, data):
         # should be taken only when length is zero and mode is not GCM because
         # AES GCM can return improper tag values if you don't call update
         # with empty plaintext when authenticating AAD for ...reasons.
-        if len(data) == 0 and not isinstance(self._mode, modes.GCM):
+        if len(data) == 0 and not self._is_gcm:
```
reaperhulk: This isn't actually called every `update` unless the data length is 0, so that's part of why you're not seeing much performance gain here.

reaperhulk:
I hereby hijack this PR conversation in the name of the (probably pointless) pursuit of performance! Here is a script:

```python
import cProfile

from cryptography.hazmat.primitives.ciphers import algorithms, modes
from cryptography.hazmat.primitives.ciphers.base import Cipher

KEY = b'\xf8\xb3U)b\x82\xe5\x99\xe9\xc4A\xa1\x8eT\xb3\xac'
CHUNK = "0" * 32768
IV = b"1234567890123456"

aes_cipher = Cipher(algorithms.AES(KEY), modes.CTR(IV), backend)


def crypto_bench():
    aes = aes_cipher.encryptor()
    for i in range(0, 100000):
        aes.update(CHUNK)

if __name__ == "__main__":
    cProfile.runctx("for _ in range(50): crypto_bench()", locals(), globals(), sort="time")
```

You'll notice this script hoists Cipher object instantiation and chunk reading entirely out of the loop since those are not what we care about. It does forget to finalize (oops, just saw that) but that won't affect the profiles materially. Here are some profiles on my laptop with progressively more aggressive optimizations. For brevity I have cut the bottom off the profiles where each function's time was zero (or very, very close).

## Master (2015-07-22)
```
         45002925 function calls in 75.762 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  5000000   40.292    0.000   40.292    0.000 {built-in method EVP_CipherUpdate}
  5000000   20.206    0.000   69.403    0.000 ciphers.py:103(update)
 10000000    6.750    0.000    6.750    0.000 {method 'new' of 'CompiledFFI' objects}
  5000000    3.315    0.000   72.718    0.000 base.py:134(update)
       50    3.007    0.060   75.762    1.515 bench.py:16(crypto_bench)
 15000150    1.087    0.000    1.087    0.000 {len}
  5000000    1.069    0.000    1.069    0.000 {method 'buffer' of 'CompiledFFI' objects}
       51    0.031    0.001    0.031    0.001 {range}
       50    0.001    0.000    0.002    0.000 ciphers.py:20(__init__)
      250    0.001    0.000    0.001    0.000 abc.py:128(__instancecheck__)
        1    0.001    0.001   75.762   75.762 <string>:1(<module>)
```

## Optimize len(data) calls in update
```
         35002931 function calls in 72.043 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  5000000   39.122    0.000   39.122    0.000 {built-in method EVP_CipherUpdate}
  5000000   18.393    0.000   65.841    0.000 ciphers.py:103(update)
 10000000    6.740    0.000    6.740    0.000 {method 'new' of 'CompiledFFI' objects}
  5000000    3.142    0.000   68.984    0.000 base.py:134(update)
       50    3.023    0.060   72.042    1.441 bench.py:16(crypto_bench)
  5000000    1.029    0.000    1.029    0.000 {method 'buffer' of 'CompiledFFI' objects}
  5000150    0.558    0.000    0.558    0.000 {len}
       51    0.030    0.001    0.030    0.001 {range}
       50    0.001    0.000    0.002    0.000 ciphers.py:20(__init__)
      250    0.001    0.000    0.001    0.000 abc.py:128(__instancecheck__)
        1    0.001    0.001   72.043   72.043 <string>:1(<module>)
```

Cut the number of `len` calls by 2/3rds and we gain ~0.5s. There's also some gain to be found in a reduciton of the total runtime of `update`. The total cumulative time differential is an artifact of not running these tests under super controlled conditions, but 0.5s out of 72 is still well under 1% performance improvement.

## Stop allocating all those outlen int * objects (+ previous optimization)

```
         30002985 function calls in 73.428 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  5000000   40.910    0.000   40.910    0.000 {built-in method EVP_CipherUpdate}
  5000000   17.966    0.000   66.862    0.000 ciphers.py:104(update)
  5000050    6.333    0.000    6.333    0.000 {method 'new' of 'CompiledFFI' objects}
  5000000    3.380    0.000   70.242    0.000 base.py:134(update)
       50    3.146    0.063   73.427    1.469 bench.py:16(crypto_bench)
  5000000    1.098    0.000    1.098    0.000 {method 'buffer' of 'CompiledFFI' objects}
  5000150    0.555    0.000    0.555    0.000 {len}
       51    0.034    0.001    0.034    0.001 {range}
       50    0.001    0.000    0.003    0.000 ciphers.py:20(__init__)
      250    0.001    0.000    0.001    0.000 abc.py:128(__instancecheck__)
        1    0.001    0.001   73.428   73.428 <string>:1(<module>)
```

We've killed another 5 million method calls and dropped `new` calls from 6.75s to 6.333s...a savings of a whole 6%, and again, much, much less than 1% on the total time consumed. So far so bad.

## Buffer reuse time (plus all previous optimizations why not)

```
         25003029 function calls in 58.293 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  5000000   37.895    0.000   37.895    0.000 {built-in method EVP_CipherUpdate}
  5000000   14.125    0.000   53.538    0.000 ciphers.py:107(update)
       50    2.392    0.048   58.293    1.166 bench.py:16(crypto_bench)
  5000000    2.327    0.000   55.865    0.000 base.py:134(update)
  5000000    1.118    0.000    1.118    0.000 {method 'buffer' of 'CompiledFFI' objects}
  5000150    0.400    0.000    0.400    0.000 {len}
       51    0.030    0.001    0.030    0.001 {range}
       50    0.001    0.000    0.002    0.000 ciphers.py:20(__init__)
        1    0.001    0.001   58.293   58.293 <string>:1(<module>)
```

In this case we see method new drop off the entire profile (it's down in the zeroes I deleted). So we save ~6.3s, which is a much more serious optimization. The challenge with this is that we now have to dynamically size and reuse a buffer for `update` calls. It looks something like this:

```
    def __init__(self, some, other, params):
        ...
        self._buf = None
        self._buflen = 0

    def update(self, data):
        # OpenSSL 0.9.8e has an assertion in its EVP code that causes it
        # to SIGABRT if you call update with an empty byte string. This can be
        # removed when we drop support for 0.9.8e (CentOS/RHEL 5). This branch
        # should be taken only when length is zero and mode is not GCM because
        # AES GCM can return improper tag values if you don't call update
        # with empty plaintext when authenticating AAD for ...reasons.
        data_len = len(data)
        if data_len == 0 and not isinstance(self._mode, modes.GCM):
            return b""

        # TODO: if one update call is very large it will keep a huge buffer
        # until the cipher object is garbage collected.
        if self._buflen < data_len:
            self._buf = self._backend._ffi.new(
                "unsigned char[]", data_len + self._block_size - 1
            )
            self._buflen = data_len

        res = self._backend._lib.EVP_CipherUpdate(
            self._ctx, self._buf, self._outlen, data, data_len
        )
        assert res != 0
        return self._backend._ffi.buffer(self._buf)[:self._outlen[0]]
```

So while this is quite a bit faster, it does have us reusing a buffer that has had previous data written into it (which, if we make a mistake, could create a class of bugs previously impossible for us) as well as making it possible to allocate arbitrarily large blocks of memory that aren't freed until the ciphercontext is garbage collected.

I ran the version with all optimizations 3 times and got an average runtime of 59.975s compared to 73.577s for no optimizations, which is ~18-19% faster in this one very specific benchmark.

alex:
This ones been sitting unchanged for a while. I think the conesnsus seems to be that just removing `isisntance` isn't worth it.

PR #2167, opened 2015-07-23
When encoding IP SANs, force packed values to be bytes()

hashbrowncipher:
In Python 2, the ipaddress library returns a bytearray, which can't be
serialized by _encode_asn1_str.  This fixes the issue by casting to bytes.

No change in Python 3.4:

>>> a = ipaddress.ip_address('10.1.1.1').packed
>>> bytes(a) is a
True

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -194,8 +194,9 @@ def _encode_subject_alt_name(backend, san):
         elif isinstance(alt_name, x509.IPAddress):
             gn = backend._lib.GENERAL_NAME_new()
             assert gn != backend._ffi.NULL
+            address = bytes(alt_name.value.packed)
             ipaddr = _encode_asn1_str(
-                backend, alt_name.value.packed, len(alt_name.value.packed)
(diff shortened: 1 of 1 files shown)


alex:
Jenkins, ok to test.

alex:
Hi @hashbrowncipher, we have some existing tests that pass with the current code, can you explain why you think this change is needed?

reaperhulk:
I suspect this is assuming the use of `py2-iapddress` rather than `ipaddress`.

hashbrowncipher:
Yep, I am using the py2-ipaddress module.  It looks like this was my mistake, as that module is far less popular than the ipaddress module.  Here is a test case that fails with the py2-ipaddress module:

```
from ipaddress import IPv4Address
from cryptography import x509
from cryptography.hazmat.backends.openssl.backend import _encode_subject_alt_name

_encode_subject_alt_name(backend, [x509.IPAddress(IPv4Address('10.1.1.1'))])
```

Given that there is already a PyPI module that does this the "right" way, I think it might be appropriate to close this pull request.

alex:
Yup, the `ipaddress` module from PyPI is what we officially depend on.

PR #2170, opened 2015-07-24
bindings for CRL encoding

reaperhulk:


reaperhulk:
Codecov looks like it didn't receive some of the coverage

alex:
@reaperhulk can you rerun some of the jobs so we get the coverage?

reaperhulk:
For some reason I can't seem to do that on mobile :(

Issue #2164, opened 2015-07-22
Issue compiling cryptography in master branch


DoctorMalboro:
I'm trying to compile Cryptography in Windows XP Professional SP3 and Python 2.7 with mingw, but I keep facing the same issue: http://pastebin.com/1Rbb6t1H

reaperhulk:
We have prebuilt wheels for use, but we only support compilation under windows using Visual Studio compilers.

DoctorMalboro:
@reaperhulk Thanks, but as I'm trying to fix the same issue as ticket #1524, I've tried installing the latest cryptography version available and had the same issue. Is this fixed in any of the prebuilt wheels?

reaperhulk:
@DoctorMalboro the next release may improve the situation for tools like cx_freeze, pyinstaller, etc but the current one (which includes the prebuilt wheels) does not. Sorry about that :(

To compile this yourself on Windows you'll need the VS2008 compiler + a version of OpenSSL that will link against it properly (which means you can't use the slproweb.com binary distribution any more). We actually build our own OpenSSLs (you can get the 32-bit build here https://jenkins.cryptography.io/job/openssl-win32-release/). You'll also need to verify you have the INCLUDE and LIB environment variables set properly of course.

DoctorMalboro:
@reaperhulk Thank you for the information, I'll give it a try!

reaperhulk:
I'm going to go ahead and close this issue since there's nothing for us to do.

PR #2172, opened 2015-07-24
Update URL for OpenSSL binaries

mhils:
The link for OpenSSL binaries should not point to the slproweb binaries anymore.

Maybe not the most beautiful fix, but after updating OpenSSL and then wasting a bit of time before I  discovered #2164, this may be worthwhile change for now.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -186,6 +186,6 @@ information, consult `Greg Wilson's blog post`_ on the subject.
 
 .. _`Homebrew`: http://brew.sh
 .. _`MacPorts`: http://www.macports.org
-.. _`pre-compiled binaries`: https://www.openssl.org/related/binaries.html
+.. _`pre-compiled binaries`: https://jenkins.cryptography.io/job/openssl-win32-release/
 .. _`bug in conda`: https://github.com/conda/conda-recipes/issues/110
 .. _`Greg Wilson's blog post`: http://software-carpentry.org/blog/2014/04/mr-biczo-was-right.html


reaperhulk:
Hmm, we should definitely fix this but I think I'd like some language that you can compile it yourself or use the binaries we build for our release infrastructure with links to both 32 and 64 bit builds.

reaperhulk:
Jenkins, ok to test

mhils:
One more thing that should be added: The binaries from https://jenkins.cryptography.io/job/openssl-win32-release/ do not work with static linking in their current form.

reaperhulk:
Oh, thank you for the reminder. We need to update those docs as well as the way we handle that in master has changed. I'll file an issue for the moment and will follow up with a PR later.

Would you mind updating this PR to add the links to both 32-bit and 64-bit so we can merge it? Thanks for your help so far!

mhils:
Thanks for the quick replies. I'll update the PR tomorrow. :+1: 

FWIW, I finally managed to got dynamic linking to work by applying the following patch:
```diff
diff --git a/src/_cffi_src/build_openssl.py b/src/_cffi_src/build_openssl.py
index dac3e4d..6af0aa8 100644
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -25,7 +25,8 @@ def _get_openssl_libraries(platform):

 def _get_openssl_windows_libraries(link_type):
     if link_type == "dynamic":
-        return ["libeay32", "ssleay32", "advapi32"]
+        return ["libeay32", "ssleay32", "advapi32",
+                "crypt32", "gdi32", "user32", "ws2_32"]
     elif link_type == "static" or link_type == "":
         return ["libeay32mt", "ssleay32mt", "advapi32",
                 "crypt32", "gdi32", "user32", "ws2_32"]
```
Is this just an issue on my end or shall I send out a PR for this as well?

reaperhulk:
Nope that's a bug, but it will be resolved with #2173 (basically there's no need for a link_type at all because the type is determined entirely by how the underlying library was built. The different names for dynamic and static libs is an artifact of how slproweb builds their OpenSSL distribution).

reaperhulk:
I updated the docs in #2174, so I'll close this one. If you think the proposed change is incomplete be sure to comment on the new PR!

PR #2174, opened 2015-07-24
remove windows link type, update docs

### LINKED ISSUE #2173: Update Windows linking code
We should remove the environment variable based linking and document that if you want to build it dynamic vs static then you should build the OpenSSL libs in your preferred fashion. The zip files we create in our jenkins job build them statically.

reaperhulk:
Supersedes #2172 and fixes #2173 

I'm on a terrible internet connection so if tests fail I'll fix tomorrow morning :smile:


mhils:
Looks good. :+1:
No additions from my end. :smiley:

PR #2175, opened 2015-07-25
handle RSA key too small and consume errors on CSR signature failure

reaperhulk:
This PR is not ready for merge as I'm not sure how to trigger the generic failure case (e.g. signature failure that isn't RSA key too small). Suggestions?

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -1057,12 +1057,12 @@ class Backend(object):
         )
         if res == 0:
             errors = self._consume_errors()
-            assert errors[0][1:] in (
+            assert errors[0][1:] == (
                 (
                     self._lib.ERR_LIB_RSA,
                     self._lib.RSA_F_RSA_SIGN,
                     self._lib.RSA_R_DIGEST_TOO_BIG_FOR_RSA_KEY
-                ),
+                )
             )
             raise ValueError("Digest too big for RSA key")
 
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -1058,11 +1058,9 @@ class Backend(object):
         if res == 0:
             errors = self._consume_errors()
             assert errors[0][1:] == (
-                (
-                    self._lib.ERR_LIB_RSA,
-                    self._lib.RSA_F_RSA_SIGN,
-                    self._lib.RSA_R_DIGEST_TOO_BIG_FOR_RSA_KEY
-                )
+                self._lib.ERR_LIB_RSA,
+                self._lib.RSA_F_RSA_SIGN,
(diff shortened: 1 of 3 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1055,7 +1055,19 @@ def create_x509_csr(self, builder, private_key, algorithm):
...
-        assert res > 0
+        if res == 0:
+            errors = self._consume_errors()
+            if errors[0][1:] in (
+                (
+                    self._lib.ERR_LIB_RSA,
+                    self._lib.RSA_F_RSA_SIGN,
+                    self._lib.RSA_R_DIGEST_TOO_BIG_FOR_RSA_KEY
+                ),
+            ):
```
alex: Just make this an assert?

reaperhulk: Derp, good idea.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1055,7 +1055,16 @@ def create_x509_csr(self, builder, private_key, algorithm):
         res = self._lib.X509_REQ_sign(
             x509_req, private_key._evp_pkey, evp_md
         )
-        assert res > 0
+        if res == 0:
+            errors = self._consume_errors()
+            assert errors[0][1:] in (
```
alex: Is there a reason to use `in` instead of `==`?

reaperhulk: Nope, left it in there from when I was looking for more than one error condition.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1055,7 +1055,16 @@ def create_x509_csr(self, builder, private_key, algorithm):
...
-        assert res > 0
+        if res == 0:
+            errors = self._consume_errors()
+            assert errors[0][1:] == (
+                (
+                    self._lib.ERR_LIB_RSA,
+                    self._lib.RSA_F_RSA_SIGN,
+                    self._lib.RSA_R_DIGEST_TOO_BIG_FOR_RSA_KEY
+                )
+            )
```
alex: You can drop the outer parens I guess.

tests/test_x509.py
```diff
@@ -1206,6 +1206,19 @@ def test_extended_key_usage(self, backend):
...
+        private_key = rsa.generate_private_key(65537, 512, backend)
+        builder = x509.CertificateSigningRequestBuilder()
+        builder = builder.subject_name(
+            x509.Name([x509.NameAttribute(x509.OID_COUNTRY_NAME, u'US')])
+        )
+
+        with pytest.raises(ValueError) as exc:
+            builder.sign(private_key, hashes.SHA512(), backend)
+
+        assert exc.value.message == "Digest too big for RSA key"
```
alex: No `.message` on py3k, use `str(exc.value)` I guess

PR #2179, opened 2015-07-30
[Discussion] Fernet: split `decrypt` into `verify` & `decrypt`

sotte:
(This is not a complete pull request, it's just meant for discussion.)

Fernet offers encryption and authentication. I argue that it should offer an interface that offers de-/encryption and verification. Currently only the de-/encryption part is supported by the interface. The verification happens internally.

I just made the smallest change possible and split the decrypt function into `verify` & `decrypt` which makes it possible to authenticate/verify the message without encrypting. Currently `verify` returns the data. It raises an exception if the message is not valid (see [here](https://cryptography.io/en/latest/development/submitting-patches/#api-considerations)) but that can be changed.

Usecase: you have a a lot of fernet encrypted data and just want to make sure that it is what you think it is.

```python
fernet = Fernet(key)
for encrypted_msg in big_encrypted_data_repo():
    try:
        fernet.verify(encrypted_msg)
    except InvalidToken:
        print('Compromised')
```

What do you think? Is this a useful feature?

### DIFF
--- a/src/cryptography/fernet.py
+++ b/src/cryptography/fernet.py
@@ -70,7 +70,7 @@ class Fernet(object):
         hmac = h.finalize()
         return base64.urlsafe_b64encode(basic_parts + hmac)
 
-    def decrypt(self, token, ttl=None):
+    def verify(self, token, ttl=None):
         if not isinstance(token, bytes):
             raise TypeError("token must be bytes.")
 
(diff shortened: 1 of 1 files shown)


glyph:
I don't understand the use-case.  What use is knowing the data is valid if you're not going to process it?  Given that you have to validate it again *when* you process it, what use is validating it ahead of time?

Rather than seeing what the code would look like, what I'd want to see before commenting on this intelligently would be what the doc patch would be.

My first reaction to this would be -0, though, because it seems the most likely consequence of this sort of a patch would be users of the API thinking that they needed to call `verify` first if they wanted verification, and thereby doing a lot of duplicate work.

alex:
I concur with @glyph, absent a compelling use case I don't think there's a reason to complicate teh API like this.

I'm going to close this issue, but if you have such a use case, we can reopen.

Thanks :sparkles:

PR #2180, opened 2015-08-01
create a ConditionalLibrary and remove unsupported items

### LINKED ISSUE #2061: [BLOCKER] Conditional names broken

reaperhulk:
This is an alternate fix to #2156 (fixes #2061)

The approach is to create a new lib object, copy all the names, then loop over the CONDITIONAL_NAMES and delete what we shouldn't have. This approach is not ideal, but should be strictly temporary. Some comments:

* Do we want to delete CONDITIONAL_NAMES from the _cffi_src? It doesn't do anything any more and the merged dictionary in this PR contains all that data.

* We could also try to be smarter (don't copy and then delete, but copy only if it should be copied) but that seemed like more work than porting our existing logic.

* Import performance does not seem materially affected by doing this (hooray) and this approach does not require a new cffi release.


src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -74,6 +79,17 @@ def _register_osrandom_engine(cls):
     def _ensure_ffi_initialized(cls):
         with cls._init_lock:
             if not cls._lib_loaded:
+                conditional_lib = _ConditionalLibrary()
+                for attr in dir(lib):
```
dstufft: All of this should probably live in the __init__ for  ``_ConditionalLibrary`` so that the API ends up being something like:

```python
class Binding:
    def _ensure_ffi_initialized(cls):
        ...
        cls.lib = _ConditionalLibrary(lib, CONDITIONAL_NAMES)
```

reaperhulk: :+1: good idea

dstufft:
* I think that the conditional names bit should exist in only one location. If it's not needed inside of ``_cffi_src`` then we should delete it from there and have it only live in ``src/cryptography/hazmat/bindings/openssl/_conditional.py``.

* Being smarter shouldn't be much more work, something like:

```python
class _ConditionalLibrary(object):

    def __init__(self, library, conditional_names):
        # Build up a list of names to exclude from copying.
        excluded_names = set()
        for condition, names in conditional_names.items():
            if not getattr(conditional_lib, condition):
                excluded_names |= set(names)
        
        # Copy over the names from library to _ConditionalLib, excluding any
        # conditional names that shouldn't exist.
        for attr in dir(library):
            # Skip names from the excluded_names list
            if attr in excluded_names:
                continue

            # Acutally do the copying
            setattr(self, attr, getattr(library, attr))
```

* I still prefer doing the work at build time, but this doesn't bother me that much.

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -74,6 +87,7 @@ def _register_osrandom_engine(cls):
     def _ensure_ffi_initialized(cls):
         with cls._init_lock:
             if not cls._lib_loaded:
+                cls.lib = _ConditionalLibrary()
```
dstufft: I know this will work as it's written, but I think it's cleaner to pass the lib and ``CONDITIONAL_NAMES`` dictionary into the ``_ConditionalLibrary``. Easier to test as well!

reaperhulk:
jenkins, retest this please.

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -23,11 +24,25 @@ def _osrandom_rand_status():
...
+    def __init__(self, lib, conditional_names):
+        excluded_names = set()
+        for condition, names in conditional_names.items():
+            if not getattr(lib, condition):
+                excluded_names |= set(names)
+
+        for attr in dir(lib):
+            if attr in excluded_names:
+                continue
+            else:
```
alex: This is simpler if you flip teh condition around `if attr not in excluded_names: setattr..`

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -23,11 +24,25 @@ def _osrandom_rand_status():
     return 1
 
 
+class _ConditionalLibrary(object):
```
alex: sooo, for reasons that I don't want to talk about, this will reduce performance on pypy.

Can you please write this as:

```py
import types

lib = types.ModuleType("name")
# add the attrs to lib here
```

alex:
@reaperhulk left a few comments
@dstufft are you cool with this approach, at least for now?

I didn't audit the CONDITIONAL_NAMES list to make sure it was complete and correct yet. Will wait until other things are done first.

dstufft:
Yea I'm OK with it for now.

alex:
Ok. Verified that `CONDITIONAL_NAMES` has the right stuff in it.

PR looks good to me, @dstufft can you please confirm?

dstufft:
LGTM

PR #2181, opened 2015-08-01
move urandom engine test

reaperhulk:
This test was in the bindings dir, which is incorrect. We do not set the urandom engine to default unless the openssl backend is loaded. The reason the test wasn't failing (even in the random order tests) is that the backends are loaded during pytest_generate_tests by a call to _available_backends. So no matter what order it occurred in the engine was already set to default. I discovered this when I tried to run the test_openssl.py bindings tests directly via pytest. Hooray global state.

### DIFF
--- a/tests/hazmat/backends/test_openssl.py
+++ b/tests/hazmat/backends/test_openssl.py
@@ -35,17 +35,17 @@ from ...utils import load_vectors_from_file, raises_unsupported_algorithm
 
 
 def skip_if_libre_ssl(openssl_version):
-    if b'LibreSSL' in openssl_version:
+    if u'LibreSSL' in openssl_version:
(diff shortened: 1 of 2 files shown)


tests/hazmat/backends/test_openssl.py
```diff
@@ -216,6 +230,19 @@ def test_bn_to_int(self):
         bn = backend._int_to_bn(0)
         assert backend._bn_to_int(bn) == 0
 
+    def test_actual_osrandom_bytes(self, monkeypatch):
+        skip_if_libre_ssl(backend.openssl_version_text())
```
alex: `version_text` returns a string, but `skip_if_libre_ssl` assumes bytes.

reaperhulk: Whoops. will fix when I get back to the apt

PR #2189, opened 2015-08-02
remove pep8/py3pep8 from travis because we can run those in jenkins

reaperhulk:
Just added a new jenkins PR builder with status API to give faster feedback on pep8 problems.


reaperhulk:
jenkins job doesn't want to behave and I'm too tired to investigate, so closing this for the moment :)

PR #2191, opened 2015-08-03
Make Homebrew install instructions more generic

tdsmith:
Homebrew supports installing to locations other than /usr/local.


reaperhulk:
jenkins, ok to test

Issue #2192, opened 2015-08-03
CertificateBuilder Extensions Support


reaperhulk:
We should support this minimum set of extensions for our 1.0 release:

- [x] SubjectKeyIdentifier
- [x] AuthorityKeyIdentifier
- [x] AuthorityInformationAccess
- [x] CRLDistributionPoints
- [x] KeyUsage
- [x] ExtendedKeyUsage
- [x] SubjectAlternativeName
- [x] BasicConstraints

Not blockers for tenth release but tracking:
- [x] CertificatePolicies (#2220)
- [x] OCSPNoCheck (PR coming after current ones land)
- [x] IssuerAlternativeName
- [x] InhibitAnyPolicy (PR coming after current ones land)
- [x] NameConstraints

reaperhulk:
SubjectKeyInformation and AuthorityKeyInformation have hashes of public key data that they want. Specifically they'd like to have the SHA1 hash of the BIT STRING value of "subjectPublicKey" (which is embedded in the subjectPublicKeyInfo structure). There's two primary ways we can do this.

* We can defer to OpenSSL (which is a pain but can probably be done with some calls to `X509_PUBKEY_set` and de-opaquing part of the `X509_PUBKEY` struct). This will require us to add a new attribute to the PublicKey interfaces (`key_identifier_hash` or something).
* We can do it ourselves in a classmethod on SubjectKeyInformation. This will require checking for each key type, converting to numbers, encoding the numbers into the appropriate sequence via pyasn1, then hashing the resulting bytes (which we can do with hashlib so we don't need to know what backend the key came from).

Both ways kind of suck, but I'm leaning towards the latter right now. Opinions?

alex:
Err, why not just `public_key.public_bytes(Encoding.DER,
Format.SubjectPublicKeyInfo)`?

On Tue, Aug 4, 2015 at 5:05 PM, Paul Kehrer <notifications@github.com>
wrote:

> SubjectKeyInformation and AuthorityKeyInformation have hashes of public
> key data that they want. Specifically they'd like to have the SHA1 hash of
> the BIT STRING value of "subjectPublicKey" (which is embedded in the
> subjectPublicKeyInfo structure). There's two primary ways we can do this.
>
>    - We can defer to OpenSSL (which is a pain but can probably be done
>    with some calls to X509_PUBKEY_set and de-opaquing part of the
>    X509_PUBKEY struct). This will require us to add a new attribute to
[...]

PR #2193, opened 2015-08-03
Support AuthorityInformationAccess in CertificateBuilder

reaperhulk:
This is part of the "minimum set" of extensions we should support for our first release with certificate generation. refs #2192


reaperhulk:
This should be considered blocked on #2196 + a rebase.

sigmavirus24:
Thank you for not asking me to do this because I would have been totally lost. :cake:

Issue #2162, opened 2015-07-21
Alternative ANS1 Encoder/Decoder


wbond:
I've been working on a bunch of cryptography stuff that is somewhat parallel to what the cryptography package is doing (a different philosophical approach), but in the process needed a better library than pyasn1.

With the recent work on generating CSRs and certs, I figured I'd share what I've been doing in the slight chance that it may be useful.

https://github.com/wbond/asn1crypto

It provides class-based declarative structure for parsing BER/DER and generating DER encodings, along with a PEM encoder and decoder. There are predefined classes for most of the ASN1 structures associated with PKI. The readme has an overview, with documentation about the core usage. Currently the details of the structures are referenced only via the source code since user-docs would be almost identical.

Currently I am in the process of building out a project that will utilize CMS structures, but along the way am building cert validation with OCSP, CRL, path building, etc. Because of this I am slowly adding high-level attributes and methods to the various classes to help with common tasks.

wbond:
Recently Python 2.6 support was added. Previously only 2.7+ were supported.

reaperhulk:
I've pulled an initial asn1crypto dep into #2435

alex:
Going to close this issue, as there's no actual work here.

PR #2204, opened 2015-08-05
SubjectKeyIdentifier classmethod

reaperhulk:
This class method is needed for creating the SKI extension that needs to be encoded when generating a certificate.

Unfortunately pyasn1 is not really geared towards giving us the data we need so the current solution is very inefficient. I'm not particularly worried about it from a performance perspective, but if there's an easy way to do this more efficiently (not by writing our own ASN.1 parser plz) I'm on board.

Longer term we could also look at the ASN1 lib mentioned in #2162.

refs #2192

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -1119,7 +1119,7 @@ X.509 Extensions
 
         The binary value of the identifier.
 
-    .. classmethod:: create_from_public_key(public_key)
+    .. classmethod:: from_public_key(public_key)
 
         .. versionadded:: 1.0
 
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -1119,6 +1119,21 @@ X.509 Extensions
 
         The binary value of the identifier.
 
+    .. classmethod:: create_from_public_key(public_key)
+
+        .. versionadded:: 1.0
+
+        Creates a new SubjectKeyIdentifier instance using the public key
+        provided to generate the appropriate digest. This should be the public
+        key that is in the certificate.
+
+        :param public_key: One of
+            :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPublicKey`
+            ,
+            :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPublicKey`
+            , or
+            :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey`.
+
 .. class:: SubjectAlternativeName
 
     .. versionadded:: 0.9
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -1125,7 +1125,9 @@ X.509 Extensions
 
         Creates a new SubjectKeyIdentifier instance using the public key
         provided to generate the appropriate digest. This should be the public
-        key that is in the certificate.
+        key that is in the certificate. The generated digest is the SHA1 hash
+        of the ``subjectPublicKey`` ASN.1 bit string. This is the first
+        recommendation in :rfc:`5280` section 4.2.1.2.
 
         :param public_key: One of
             :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPublicKey`
--- a/src/cryptography/utils.py
+++ b/src/cryptography/utils.py
@@ -47,16 +47,10 @@ else:
         return result
 
 
-if hasattr(int, "to_bytes"):
-    int_to_bytes = int.to_bytes
-else:
-    def int_to_bytes(integer, byteorder, signed=False):
-        assert byteorder == 'big'
-        assert not signed
-
-        hex_string = '%x' % integer
-        n = len(hex_string)
-        return binascii.unhexlify(hex_string.zfill(n + (n & 1)))
+def int_to_bytes(integer):
+    hex_string = '%x' % integer
+    n = len(hex_string)
+    return binascii.unhexlify(hex_string.zfill(n + (n & 1)))
 
 
 class InterfaceNotImplemented(Exception):
--- a/src/cryptography/utils.py
+++ b/src/cryptography/utils.py
@@ -5,6 +5,7 @@
 from __future__ import absolute_import, division, print_function
 
 import abc
+import binascii
 import inspect
 import struct
 import sys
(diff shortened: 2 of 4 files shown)


docs/x509/reference.rst
```diff
@@ -1119,6 +1119,21 @@ X.509 Extensions
 
         The binary value of the identifier.
 
+    .. classmethod:: create_from_public_key(public_key)
```
alex: Maybe name it `from_public_key`?

src/cryptography/x509.py
```diff
@@ -669,6 +681,29 @@ class SubjectKeyIdentifier(object):
...
+        # pyasn1 really doesn't want to give them to us. To get it we'll
+        # build an integer, hex it, then decode the hex.
+        bits = 0
+        for bit in spki.getComponentByName("subjectPublicKey"):
+            bits = bits << 1 | bit
+
+        # convert the integer to bytes
+        hex_string = '%x' % bits
+        n = len(hex_string)
+        data = binascii.unhexlify(hex_string.zfill(n + (n & 1)))
```
alex: Does it maybe make sense to have `public_key.public_bytes(DER, SubjectPublicKey)` (no `Info`)?

reaperhulk: We'd want to limit that to DER only and its utility as "the bytes you hash to generate a key identifier" is too limited to be of general benefit in my opinion.

docs/x509/reference.rst
reaperhulk: renamed.

src/cryptography/x509.py
```diff
@@ -669,6 +681,29 @@ class SubjectKeyIdentifier(object):
...
+        )
+        # the univ.BitString object is a tuple of bits. We need bytes and
+        # pyasn1 really doesn't want to give them to us. To get it we'll
+        # build an integer, hex it, then decode the hex.
+        bits = 0
+        for bit in spki.getComponentByName("subjectPublicKey"):
+            bits = bits << 1 | bit
+
+        # convert the integer to bytes
+        hex_string = '%x' % bits
```
alex: This should probably be `hex()`.

src/cryptography/x509.py
alex: Can you epxlain what algorithm is being implemented, I don't understand what's going on with the hex.

src/cryptography/x509.py
```diff
@@ -669,6 +681,29 @@ class SubjectKeyIdentifier(object):
...
         self._digest = digest
 
+    @classmethod
+    def from_public_key(cls, public_key):
+        # This is a very slow way to do this.
+        serialized = public_key.public_bytes(
+            serialization.Encoding.DER,
+            serialization.PublicFormat.SubjectPublicKeyInfo
+        )
+        spki, remaining = decoder.decode(
```
alex: `assert not remainign` or something?

src/cryptography/x509.py
reaperhulk: We can switch to hex() but then we'll need to [2:] to strip the 0x and .rstrip("L") to get rid of the L in Python 2. That seems uglier, but admittedly `%` is a weird beast in Python with its "not recommended but never going to be removed" status.

If you'd prefer it to swap to `hex(bits)[2:].rstrip("L")` I'll change it.

src/cryptography/x509.py
reaperhulk: We take the integer, convert to hex, zfill it to make sure it has an even number of characters, then unhexlify it to bytes. It's the reverse of `int.from_bytes(bytes, "big")`

```python
# python3
>>> import binascii
>>> value = 383842938479274659827346598273465298376452398476529038457
>>> hex_string = '%x' % value
>>> n = len(hex_string)
>>> data = binascii.unhexlify(hex_string.zfill(n + (n & 1)))
>>> data
b'\x0f\xa7\x81\xdc5\x1d\xd9\xa0%\x97\x964j\x01\xd8\xf1Z\xad\x93\xb1\xd4\x90\xf4y'
>>> int.from_bytes(data, "big")
383842938479274659827346598273465298376452398476529038457
>>> int.from_bytes(data, "big") == value
True
```

docs/x509/reference.rst
```diff
@@ -1119,6 +1119,21 @@ X.509 Extensions
 
         The binary value of the identifier.
 
+    .. classmethod:: from_public_key(public_key)
+
+        .. versionadded:: 1.0
+
+        Creates a new SubjectKeyIdentifier instance using the public key
+        provided to generate the appropriate digest. This should be the public
+        key that is in the certificate.
```
alex: From reading the RFC, it sounds like the SHA1 of the bytes is only one option, is that right?

src/cryptography/x509.py
alex: Can we factor this logic out into `utils.int_to_bytes` (and use the method on `int` where available)

docs/x509/reference.rst
reaperhulk: That is correct, although it's the only option I've ever seen used. I'll add some more prose to explain that it's the SHA1 of the subjectPublicKey BIT STRING and link the RFC section

alex: Thanks

On Sat, Aug 8, 2015 at 10:49 AM, Paul Kehrer <notifications@github.com>
wrote:

> In docs/x509/reference.rst
> <https://github.com/pyca/cryptography/pull/2204#discussion_r36578919>:
>
> > @@ -1119,6 +1119,21 @@ X.509 Extensions
> >
> >          The binary value of the identifier.
> >
> > +    .. classmethod:: from_public_key(public_key)
> > +
> > +        .. versionadded:: 1.0
> > +
> > +        Creates a new SubjectKeyIdentifier instance using the public key
> > +        provided to generate the appropriate digest. This should be the public
> > +        key that is in the certificate.
>
> That is correct, although it's the only option I've ever seen used. I'll
> add some more prose to explain that it's the SHA1 of the subjectPublicKey
> BIT STRING and link the RFC section
>
> —
> <https://github.com/pyca/cryptography/pull/2204/files#r36578919>.
>



-- 

src/cryptography/x509.py
```diff
@@ -669,6 +680,27 @@ class SubjectKeyIdentifier(object):
...
+            serialized, asn1Spec=_SubjectPublicKeyInfo()
+        )
+        assert not remaining
+        # the univ.BitString object is a tuple of bits. We need bytes and
+        # pyasn1 really doesn't want to give them to us. To get it we'll
+        # build an integer, hex it, then decode the hex.
```
alex: This comment is now a bit innacurate, the hex crap is behind an abstraction now.

src/cryptography/utils.py
```diff
@@ -46,6 +47,18 @@ def int_from_bytes(data, byteorder, signed=False):
         return result
 
 
+if hasattr(int, "to_bytes"):
+    int_to_bytes = int.to_bytes
+else:
+    def int_to_bytes(integer, byteorder, signed=False):
```
alex: Eh, fuck, `int.to_bytes` has an API where you need to include the desired length.

PR #2198, opened 2015-08-03
support CRLDistributionPoints in the CertificateBuilder

reaperhulk:
This PR should be considered blocked on #2193 and #2197 

refs #2192

PR #2206, opened 2015-08-05
refactor SAN encoding to separate out general names in openssl backend

reaperhulk:
refs #2198

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -241,6 +241,9 @@ def _encode_general_names(backend, names):
 
 def _encode_subject_alt_name(backend, san):
     general_names = _encode_general_names(backend, san)
+    general_names = backend._ffi.gc(
+        general_names, backend._lib.GENERAL_NAMES_free
+    )
     pp = backend._ffi.new("unsigned char **")
     r = backend._lib.i2d_GENERAL_NAMES(general_names, pp)
     assert r > 0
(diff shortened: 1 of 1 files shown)


alex:
The `GENERAL_NAMES_free` got totally lost in this PR.

reaperhulk:
That bug was in the PR this was split off from! Hooray splitting PRs to improve reviews. Fixing

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -228,18 +228,22 @@ def _encode_authority_information_access(backend, authority_info_access):
...
         assert res != 0
 
+    return general_names
+
+
+def _encode_subject_alt_name(backend, san):
+    general_names = _encode_general_names(backend, san)
+    general_names = backend._ffi.gc(
+        general_names, backend._lib.GENERAL_NAMES_free
+    )
```
alex: We usually try to keep the `ffi.gc()` near the allocation.

reaperhulk: I don't want to gc the allocation in this case because the refactored function will be used elsewhere (and needs to not be gc'd).

This does bring up an inconsistency in naming (our `_encode_<extension>` functions return a tuple of a char * and length and this new one returns a `sk_GENERAL_NAMES`). I'm happy to rename it (`_build_general_names`?) if that's helpful.

reaperhulk: That english was not terrific. What I mean is that the reason for refactoring this is that `_encode_general_names` will be used in the CRL encoding PR and I need to have the sk_GENERAL_NAMES it returns not be gc'd.

alex: Ok. I'm ok with this for now and will merge, but long term we should probably think about ways to clean about both the encoding functions and the GC bits.

PR #2207, opened 2015-08-05
Move validation code into CertificateBuilder constructor

bwhmather:
Realised earlier that values passed into the constructor of the CertificateBuilder (and other builders) are not validated.  This pull requests just shifts the validation code out of the chaining methods and into `__init__`.

If you call multiple methods in succession then validation code for earlier attributes will be called multiple times.  I think fixing this would be a premature optimisation though.

There is also a problem with `add_extension` taking an object which gets turned into an `Extension` and `__init__` taking a list of `Extensions`.  I think neither should accept `Extension`s but to achieve that, the original list of objects needs to be saved as well as the list of `Extension` objects.  I've split my idea of how to go about it into a second commit but would I am not currently very happy with it.  In particular I'm not sure what to do with the `critical` argument (for now have bundled it into a tuple, could we add it to the extension objects or possible have two arguments taking lists of critical and non-critical extensions?) and I'm not sure `_base_extensions` is the best name.  I would really appreciate some advice.

Will be happy to repeat the work on the other builder classes if everyone thinks it would be a good idea.

### DIFF
--- a/src/cryptography/x509/base.py
+++ b/src/cryptography/x509/base.py
@@ -311,116 +311,93 @@ class CertificateSigningRequestBuilder(object):
 
 
 class CertificateBuilder(object):
-    def __init__(self, issuer_name=None, subject_name=None,
-                 public_key=None, serial_number=None, not_valid_before=None,
-                 not_valid_after=None, extensions=[]):
+    def __init__(self, **kwargs):
         self._version = Version.v3
-        self._issuer_name = issuer_name
-        self._subject_name = subject_name
-        self._public_key = public_key
-        self._serial_number = serial_number
-        self._not_valid_before = not_valid_before
-        self._not_valid_after = not_valid_after
-        self._extensions = extensions
+        self._issuer_name = None
+        self._subject_name = None
+        self._public_key = None
+        self._serial_number = None
+        self._not_valid_before = None
+        self._not_valid_after = None
+        self._extensions = []
 
-    def issuer_name(self, name):
-        """
-        Sets the CA's distinguished name.
-        """
-        if not isinstance(name, Name):
-            raise TypeError('Expecting x509.Name object.')
-        if self._issuer_name is not None:
-            raise ValueError('The issuer name may only be set once.')
-        return CertificateBuilder(
-            name, self._subject_name, self._public_key,
-            self._serial_number, self._not_valid_before,
-            self._not_valid_after, self._extensions
-        )
+        self._apply_updates(**kwargs)
 
-    def subject_name(self, name):
-        """
-        Sets the requestor's distinguished name.
-        """
-        if not isinstance(name, Name):
-            raise TypeError('Expecting x509.Name object.')
-        if self._subject_name is not None:
-            raise ValueError('The subject name may only be set once.')
-        return CertificateBuilder(
-            self._issuer_name, name, self._public_key,
-            self._serial_number, self._not_valid_before,
-            self._not_valid_after, self._extensions
-        )
+    def _clone(self):
+        clone = self.__class__()
+        clone._issuer_name = self._issuer_name
+        clone._subject_name = self._subject_name
+        clone._public_key = self._public_key
+        clone._serial_number = self._serial_number
+        clone._not_valid_before = self._not_valid_before
+        clone._not_valid_after = self._not_valid_after
+        clone._extensions = list(self._extensions)
+        return clone
 
-    def public_key(self, key):
-        """
-        Sets the requestor's public key (as found in the signing request).
-        """
-        if not isinstance(key, (dsa.DSAPublicKey, rsa.RSAPublicKey,
-                                ec.EllipticCurvePublicKey)):
-            raise TypeError('Expecting one of DSAPublicKey, RSAPublicKey,'
-                            ' or EllipticCurvePublicKey.')
-        if self._public_key is not None:
-            raise ValueError('The public key may only be set once.')
-        return CertificateBuilder(
-            self._issuer_name, self._subject_name, key,
-            self._serial_number, self._not_valid_before,
-            self._not_valid_after, self._extensions
-        )
+    def _with_updates(self, **kwargs):
+        clone = self._clone()
+        clone._apply_updates(**kwargs)
+        return clone
 
-    def serial_number(self, number):
-        """
-        Sets the certificate serial number.
-        """
-        if not isinstance(number, six.integer_types):
-            raise TypeError('Serial number must be of integral type.')
-        if self._serial_number is not None:
-            raise ValueError('The serial number may only be set once.')
-        if number < 0:
-            raise ValueError('The serial number should be non-negative.')
-        if utils.bit_length(number) > 160:  # As defined in RFC 5280
-            raise ValueError('The serial number should not be more than 160 '
-                             'bits.')
-        return CertificateBuilder(
-            self._issuer_name, self._subject_name,
-            self._public_key, number, self._not_valid_before,
-            self._not_valid_after, self._extensions
-        )
+    def _apply_updates(self, issuer_name=None, subject_name=None,
+                       public_key=None, serial_number=None,
+                       not_valid_before=None, not_valid_after=None):
+        if issuer_name is not None:
+            if not isinstance(issuer_name, Name):
+                raise TypeError('Expecting x509.Name object for issuer name.')
+            if self._issuer_name is not None:
+                raise ValueError('The issuer name may only be set once.')
+            self._issuer_name = issuer_name
 
-    def not_valid_before(self, time):
-        """
-        Sets the certificate activation time.
-        """
-        if not isinstance(time, datetime.datetime):
-            raise TypeError('Expecting datetime object.')
-        if self._not_valid_before is not None:
-            raise ValueError('The not valid before may only be set once.')
-        if time <= _UNIX_EPOCH:
-            raise ValueError('The not valid before date must be after the unix'
-                             ' epoch (1970 January 1).')
-        return CertificateBuilder(
-            self._issuer_name, self._subject_name,
-            self._public_key, self._serial_number, time,
-            self._not_valid_after, self._extensions
-        )
+        if subject_name is not None:
+            if not isinstance(subject_name, Name):
+                raise TypeError('Expecting x509.Name object for subject name.')
+            if self._subject_name is not None:
+                raise ValueError('The subject name may only be set once.')
+            self._subject_name = subject_name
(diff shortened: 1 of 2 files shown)


bwhmather:
(rebased to fix python 2.6 test failures)

reaperhulk:
jenkins, ok to test

reaperhulk:
We don't document the `__init__` explicitly because we don't want people to use it, which is why we never hoisted the logic up into the initializer. Is there a use case you have for directly passing the values rather than just chaining?

bwhmather:
Ah ok.  I have a use case where arguments aren't sufficient so I figured that that was what chaining was for and that arguments were the preferred way.  It's an application where the client side of the certificate builder is filled in in one function (using arguments to the constructor) then the builder gets passed of to a different function which sets the issuer and signs the certificate.

If you want to have a look, it's part of a certificate authority I've been playing with as a side project (it's still in the frenzied hacking stage and not even close to being useful or safe).  Relevant parts are [pope_signing_service.signing](https://github.com/bwhmather/pope-signing-service/blob/develop/pope_signing_service/signing.py) and [pope_signing_service.certificates](https://github.com/bwhmather/pope-signing-service/blob/develop/pope_signing_service/certificates.py).

I think having the option to use arguments would be nice as it looks more compact and traditional code.  It also stops people like me running across the constructor but thinking it's safe.  As I mentioned before, the extension argument is something that needs to be thought about but it's a backwards compatible change so there is no hurry.

alex:
I disagree on the constructor, if Python let us, I'd say it should be
private. Since it doesn't, perhaps we should give each argument to init an
underscore prefix so it's clear they're private.

On Wed, Aug 5, 2015 at 4:25 PM, Ben Mather <notifications@github.com> wrote:

> Ah ok. I have a use case where arguments aren't sufficient so I figured
> that that was what chaining was for and that arguments were the preferred
> way. It's an application where the client side of the certificate builder
> is filled in in one function (using arguments to the constructor) then the
> builder gets passed of to a different function which sets the issuer and
> signs the certificate.
>
> If you want to have a look, it's part of a certificate authority I've been
> playing with as a side project (it's still in the frenzied hacking stage
> and not even close to being useful or safe). Relevant parts are
> pope_signing_service.signing
> <https://github.com/bwhmather/pope-signing-service/blob/develop/pope_signing_service/signing.py>
> and pope_signing_service.certificates
> <https://github.com/bwhmather/pope-signing-service/blob/develop/pope_signing_service/certificates.py>
> .
>
> I think having the option to use arguments would be nice as it looks more
> compact and traditional code. It also stops people like me running across
> the constructor but thinking it's safe. As I mentioned before, the
> extension argument is something that needs to be thought about but it's a
> backwards compatible change so there is no hurry.
>
> —
> <https://github.com/pyca/cryptography/pull/2207#issuecomment-128136835>.
>



-- 

reaperhulk:
@alex we could collapse the constructor to having no arguments and use a private class method to set all the attributes.

bwhmather:
Have pushed a new commit which fixes most of my complaints about my previous attempt just in case.  Have refactored validation into private mutating setters and added a `_clone` method.  The critical argument to `add_extension` has been replaced in `__init__` with two extension arguments: `extensions` and `optional_extensions`.  It's still down to how you want the API to look though.

@reaperhulk refactoring my `_clone` method to call `self.__class__()` instead of `self.__class__.__new__(self.__class__), and replacing `__init__` with the contents of `_init` should work for getting rid of the `__init__` arguments.  I'm happy to do it.

bwhmather:
I thought it would be worth rebasing this and checking it still works as I still think it's a good idea and at worst it gives you an excuse to close the pull request.

In summary, this patch makes it so that if someone tries to do this:

    cert = x509.CertificateBuilder(
        serial_number=uuid4().int,
        not_valid_before=now,
        not_valid_after=now + timedelta(days=730),
        subject_name=name,
        issuer_name=name,
        public_key=public_key,
    ).sign(...)

instead of this:

    cert = x509.CertificateBuilder()\
        .serial_number(uuid4().int)\
        .not_valid_before(now)\
        .not_valid_after(now + timedelta(days=730))\
        .subject_name(name)\
        .issuer_name(name)\
        .public_key(public_key)\
        .sign(...)

the validation code still runs and everything is safe.
The original method for assembling a certificate builder continues work and you can mix and match the two methods happily.

If you do decide to keep it will apply the same treatment to the other builders.

alex:
I agree that it's unfortunate that it's possibel to use it incorrectly now;
@reaperhulk what do you think abotu `_` prefixed names to make it clear
that it's not a public API?

On Tue, Sep 29, 2015 at 5:39 PM, Ben Mather <notifications@github.com>
wrote:

> I thought it would be worth rebasing this and checking it still works as I
> still think it's a good idea and at worst it gives you an excuse to close
> the pull request.
>
> In summary, this patch makes it so that if someone tries to do this:
>
> cert = x509.CertificateBuilder(
>     serial_number=uuid4().int,
>     not_valid_before=now,
>     not_valid_after=now + timedelta(days=730),
>     subject_name=name,
>     issuer_name=name,
>     public_key=public_key,
> ).sign(...)
>
> instead of this:
>
> cert = x509.CertificateBuilder()\
>     .serial_number(uuid4().int)\
>     .not_valid_before(now)\
>     .not_valid_after(now + timedelta(days=730))\
>     .subject_name(name)\
>     .issuer_name(name)\
>     .public_key(public_key)\
>     .sign(...)
>
> the validation code still runs and everything is safe.
> The original method for assembling a certificate builder continues work
> and you can mix and match the two methods happily.
>
> If you do decide to keep it will apply the same treatment to the other
> builders.
>
> —
> <https://github.com/pyca/cryptography/pull/2207#issuecomment-144199280>.
>



-- 

reaperhulk:
If we fail to properly validate here the consequence is that we can pass data to the backend that will cause a signature failure (and since OpenSSL, et al can fail in so many ways we might have an InternalError on an unhandled error code). This is bad, but (probably) not catastrophic.

I can definitely see users calling this without kwargs and not necessarily even realizing it's not the "right" way to do it (especially given the builders are the only chaining API in the entire library). I'm not a big fan of the significant increase in code for this, but I think we either need to raise an exception if you call this outside of the chaining API or else do the validation to protect users from themselves.

I'm not sure how to accomplish the former without something ugly like a secret guard `object()` though.

reaperhulk:
I'm going back through PRs we've let languish for too long (sorry!). I want to say we should just prefix the args to make it clear it's not public API, but it has just occurred to me that @mithrandi might be interested in benchmarking the performance difference between the "provide it all in the init" vs the chaining API approach. The latter generates a pile of new objects, but it's possible that the performance differential is irrelevant (and if you don't care feel free to say so @mithrandi :smile:).

mithrandi:
Okay, I wrote a benchmark: https://gist.github.com/mithrandi/85257ca7cd6305cd7b30ae0a6ca811d2

The differences between the two styles seem to be statistically insignificant on CPython 2.7; on PyPy, unchained style seems to be slightly faster, but not by enough to care (under 1%). I didn't test on Python 3 because I was lazy.

PyPy 5.0.1:
```
(cryptography-pypy) mithrandi@lorien ~/c/cryptography> python bench_builder.py 0
Chained style: (1.6999760695866177, 588.2435746540654)
(cryptography-pypy) mithrandi@lorien ~/c/cryptography> python bench_builder.py 0
Chained style: (1.695082187652588, 589.9418962008189)
(cryptography-pypy) mithrandi@lorien ~/c/cryptography> python bench_builder.py 0
Chained style: (1.698562502861023, 588.7331189259277)
(cryptography-pypy) mithrandi@lorien ~/c/cryptography> python bench_builder.py 1
Unhained style: (1.6938082660947529, 590.385594412998)
(cryptography-pypy) mithrandi@lorien ~/c/cryptography> python bench_builder.py 1
Unhained style: (1.6674516075535823, 599.717554302616)
(cryptography-pypy) mithrandi@lorien ~/c/cryptography> python bench_builder.py 1
Unhained style: (1.658584193179482, 602.9238697150577)
```

CPython 2.7.11:
```
(cryptography) mithrandi@lorien ~/c/cryptography> python bench_builder.py 0
Chained style: (1.5836537073528956, 631.4511785985821)
(cryptography) mithrandi@lorien ~/c/cryptography> python bench_builder.py 0
Chained style: (1.5371150970458984, 650.5693698031123)
(cryptography) mithrandi@lorien ~/c/cryptography> python bench_builder.py 0
Chained style: (1.5272971122495589, 654.7514507685397)
(cryptography) mithrandi@lorien ~/c/cryptography> python bench_builder.py 1
Unhained style: (1.5328672954014368, 652.3721936008255)
(cryptography) mithrandi@lorien ~/c/cryptography> python bench_builder.py 1
Unhained style: (1.5327594496987083, 652.4180948266659)
(cryptography) mithrandi@lorien ~/c/cryptography> python bench_builder.py 1
Unhained style: (1.515764676607572, 659.7330149150175)

**Comment by reaperhulk** [comment]
So performance is not a factor here at all. Thanks for writing that!

**Comment by bwhmather** [comment]
Thanks for following up.  Definitely my fault that it's languished.  Not helped by there not seeming to be any clear solution.

Maybe something like this would work to fix the problems with verbosity and with accidentally passing arguments to the constructor:

```python
class CertificateBuilder(object):
    def __init__(self):
        self._version = Version.v3
        self._issuer_name = None
        self._subject_name = None
        self._public_key = None
        self._serial_number = None
        self._not_valid_before = None
        self._not_valid_after = None
        self._extensions = []
       
    def _clone(self):
        clone = self.__class__()
        clone._issuer_name = self._issuer_name
        clone._subject_name = self._subject_name
        clone._public_key = self._public_key
        clone._serial_number = self._serial_number
        clone._not_valid_before = self._not_valid_before
        clone._not_valid_after = self._not_valid_after
        clone._extensions = list(self._extensions)
        return clone
            
    def issuer_name(self, issuer_name):
        """
        Sets the CA's distinguished name.
        """
        if not isinstance(issuer_name, Name):
            raise TypeError('Expecting x509.Name object for issuer name.')
        if self._issuer_name is not None:
            raise ValueError('The issuer name may only be set once.')

        clone = self._clone
        clone._issuer_name = issuer_name
        return clone
```
It's still not ideal in my opinion.  Setting private attributes from a different instance just seems wrong, even if `_clone` does ensure it is of the same class.

The other alternative would be to make the builder mutable but that would break backwards compatibility and consistency with the rest of the module, and being able to set the issuer name and extensions once is genuinely useful.

alex:
I don't think we're going to go this route.

PR #2210, opened 2015-08-05
AuthorityKeyIdentifier classmethod

reaperhulk:
Depends on #2209 and #2204

PR #2213, opened 2015-08-05
Support AuthorityKeyIdentifier in CertificateBuilder

reaperhulk:
This depends on #2212 and #2210 so no need to review until then. I just wanted to get it up so all the extensions we're going to support for 1.0 have outstanding PRs.

refs #2192


alex:
retest this please

PR #2217, opened 2015-08-06
Add OIDs to extension types and add ExtensionType interface

reaperhulk:
I'm putting up this PR to get some discussion going. Right now we have a set of extensions that have no common interface. Additionally, we don't directly associate the OID with them. Instead we create an extension object `Extension(oid, ext_class, critical)`. This is how the ASN.1 structure works, but I think there's significant value in attaching the OID as a property on the extension type class as well (BasicConstraints, SubjectAlternativeName, etc). At the moment our API requires you to know both the extension type class and the OID and the association between them is in the docs only. 

One improvement: `get_extension_by_oid` could now have a (far more friendly) `get_extension` that takes an extension type class. This does not actually require the oid attribute, but I wanted to mention it since I think we should add this in any case.

```python
ext = cert.get_extension(x509.BasicConstraints)
# instead of
ext = cert.get_extension_by_oid(x509.OID_BASIC_CONSTRAINTS)
```

We could also simplify some of our if/elif chains in CSRBuilder and CertificateBuilder:

```python
        if isinstance(extension, BasicConstraints):
            extension = Extension(OID_BASIC_CONSTRAINTS, critical, extension)
        elif isinstance(extension, ExtendedKeyUsage):
            extension = Extension(OID_EXTENDED_KEY_USAGE, critical, extension)
        elif isinstance(extension, SubjectAlternativeName):
            extension = Extension(
                OID_SUBJECT_ALTERNATIVE_NAME, critical, extension
            )
        elif isinstance(extension, KeyUsage):
            extension = Extension(OID_KEY_USAGE, critical, extension)
        elif isinstance(extension, InhibitAnyPolicy):
            extension = Extension(OID_INHIBIT_ANY_POLICY, critical, extension)
        else:
            raise NotImplementedError('Unsupported X.509 extension.')
```
becomes
```python
        ext = Extension(extension.oid, critical, extension)
        # NotImplementedError can be handled by the backend (which has that logic anyway)
```

This PR also proposes to have the extension type classes share a common interface. By doing so we get automatic testing for the presence of the `oid` attribute and we can use `isinstance` checks to give better errors out of code like the first example above. It will also help us by giving a common name for the extension type classes in the documentation.

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -922,8 +922,7 @@ X.509 Extensions
 
     The key usage extension defines the purpose of the key contained in the
     certificate.  The usage restriction might be employed when a key that could
-    be used for more than one operation is to be restricted. It corresponds to
-    :data:`OID_KEY_USAGE`.
+    be used for more than one operation is to be restricted.
 
     .. attribute:: oid
 
@@ -931,7 +930,7 @@ X.509 Extensions
 
         :type: :class:`ObjectIdentifier`
 
-        The OID associated with this extension type.
+        Returns :data:`OID_KEY_USAGE`.
 
     .. attribute:: digital_signature
 
@@ -1022,8 +1021,7 @@ X.509 Extensions
 
     Basic constraints is an X.509 extension type that defines whether a given
     certificate is allowed to sign additional certificates and what path
-    length restrictions may exist. It corresponds to
-    :data:`OID_BASIC_CONSTRAINTS`.
+    length restrictions may exist.
 
     .. attribute:: oid
 
@@ -1031,7 +1029,7 @@ X.509 Extensions
 
         :type: :class:`ObjectIdentifier`
 
-        The OID associated with this extension type.
+        Returns :data:`OID_BASIC_CONSTRAINTS`.
 
     .. attribute:: ca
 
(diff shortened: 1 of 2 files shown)


alex:
This looks like a pretty reasonable approach. Are you happy wtih the patch as is?

reaperhulk:
I am. Just wanted to explain my rationale in case there was disagreement.

alex:
Should the docs be updated to say "instead of the OID that corrosponds to this extension" to just say what value it is? Some of them have that info in the prose, which can be removed.

reaperhulk:
Agreed. Updated (and added an OID we actually were missing thanks to changing the docs in this way).

docs/x509/reference.rst
```diff
@@ -1051,6 +1081,14 @@ X.509 Extensions
     extension is only relevant when the certificate is an authorized OCSP
     responder.
 
+    .. attribute:: oid
+
+        .. versionadded:: 1.0
+
+        :type: :class:`ObjectIdentifier`
+
+        The OID associated with this extension type.
```
alex: This should be updated to match the others.

alex:
Besides that comment this LGTM

PR #2220, opened 2015-08-06
support encoding certificate policies in CertificateBuilder

reaperhulk:
I had this one mostly written already so I finished it. :smile:

Depends on #2231

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -152,23 +152,23 @@ def _encode_name_gc(backend, attributes):
 
 def _encode_certificate_policies(backend, certificate_policies):
     cp = backend._lib.sk_POLICYINFO_new_null()
-    assert cp != backend._ffi.NULL
+    backend.openssl_assert(cp != backend._ffi.NULL)
     cp = backend._ffi.gc(cp, backend._lib.sk_POLICYINFO_free)
     for policy_info in certificate_policies:
         pi = backend._lib.POLICYINFO_new()
-        assert pi != backend._ffi.NULL
+        backend.openssl_assert(pi != backend._ffi.NULL)
         res = backend._lib.sk_POLICYINFO_push(cp, pi)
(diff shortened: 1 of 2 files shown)


alex:
This is sufficiently complicated that I think the review is going to be a few passes, would you mind putting the bindings in their own PR so they're out of the way?

reaperhulk:
depends on #2227

alex:
Can be rebased

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -98,6 +98,20 @@ def _encode_asn1_str(backend, data, length):
...
+    """
+    Create an ASN1_UTF8STRING from a Python unicode string.
+    This object will be a ASN1_STRING with UTF8 type in OpenSSL and
+    can be decoded with ASN1_STRING_to_UTF8.
+    """
+    s = backend._lib.ASN1_UTF8STRING_new()
+    res = backend._lib.ASN1_STRING_set(
+        s, string.encode("utf8"), len(string.encode("utf8"))
+    )
+    assert res == 1
```
alex: `openssl_assert`

alex: (Applies to this entire PR really)

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -487,6 +576,7 @@ def _encode_crl_distribution_points(backend, crl_distribution_points):
     ExtensionOID.ISSUER_ALTERNATIVE_NAME: _encode_alt_name,
     ExtensionOID.EXTENDED_KEY_USAGE: _encode_extended_key_usage,
     ExtensionOID.AUTHORITY_KEY_IDENTIFIER: _encode_authority_key_identifier,
+    x509.OID_CERTIFICATE_POLICIES: _encode_certificate_policies,
```
alex: This should use `ExtensionOID`

PR #2231, opened 2015-08-09
Simplify Cert and CSR Builder

reaperhulk:

PR #2233, opened 2015-08-09
add support for OCSPNoCheck to the CertificateBuilder

reaperhulk:
depends on #2231

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -155,6 +155,15 @@ def _txt2obj_gc(backend, name):
     return obj
 
 
+def _encode_ocsp_nocheck(backend, ext):
+    """
+    The OCSP No Check extension is defined as a null ASN.1 value. We can just
+    return that value directly here in the pp, r tuple form the other
+    extension encoding functions use.
+    """
+    return [b"\x05\x00"], 2
+
+
 def _encode_key_usage(backend, key_usage):
     set_bit = backend._lib.ASN1_BIT_STRING_set_bit
     ku = backend._lib.ASN1_BIT_STRING_new()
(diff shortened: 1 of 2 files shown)


alex:
Can be rebased.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -155,6 +155,15 @@ def _txt2obj_gc(backend, name):
     return obj
 
 
+def _encode_ocsp_nocheck(backend, ext):
+    """
+    The OCSP No Check extension is defined as a null ASN.1 value. We can just
+    return that value directly here in the pp, r tuple form the other
+    extension encoding functions use.
```
alex: I'm not sure I understand this comment, since it is still returning a tuple.

Does OpenSSL not have a function that does this, I feel like it would contribute to the readability.

reaperhulk: We can build a null by doing some magic with `ASN1_TYPE` but since an ASN.1 null is always `\x05\x00` I'm not in favor of calling (several) C functions to get those two bytes.

The comment says "return that value directly here in the pp, r tuple form" so it isn't saying it won't return a tuple. Would it be more clear if it read "Rather than calling OpenSSL we can return a Python bytestring value in a list"?

alex: Yeah, that would be clearer.

On Sun, Aug 9, 2015 at 11:50 AM, Paul Kehrer <notifications@github.com>
wrote:

> In src/cryptography/hazmat/backends/openssl/backend.py
> <https://github.com/pyca/cryptography/pull/2233#discussion_r36589695>:
>
> > @@ -155,6 +155,15 @@ def _txt2obj_gc(backend, name):
> >      return obj
> >
> >
> > +def _encode_ocsp_nocheck(backend, ext):
> > +    """
> > +    The OCSP No Check extension is defined as a null ASN.1 value. We can just
> > +    return that value directly here in the pp, r tuple form the other
> > +    extension encoding functions use.
>
> We can build a null by doing some magic with ASN1_TYPE but since an ASN.1
> null is always \x05\x00 I'm not in favor of calling (several) C functions
> to get those two bytes.
>
> The comment says "return that value directly here in the pp, r tuple form"
> so it isn't saying it won't return a tuple. Would it be more clear if it
> read "Rather than calling OpenSSL we can return a Python bytestring value
> in a list"?
>
> —
> <https://github.com/pyca/cryptography/pull/2233/files#r36589695>.
>



-- 

PR #2237, opened 2015-08-09
Reduce the duplication in teh string versions of oids

alex:


alex:
FWIW, my plan is to do these one group at a time, and when they're all migrated, remove the dotted strings.

PR #2240, opened 2015-08-10
namespace the oids

reaperhulk:
Sorry about the size, but at least it's pretty straightforward.


alex:
@reaperhulk ugh, I hate to ask, but can you split this out one namespace at a time, reviewing them like this is difficult

PR #2244, opened 2015-08-10
split general names

reaperhulk:


alex:
@reaperhulk `_GENERAL_NAMES` is duplicated in this PR, it should be removed from `base.py`

PR #2250, opened 2015-08-10
resolve incorrect docs/naming around DSA (r, s) tuple encode/decode

### LINKED ISSUE #2246: Documentation of DSA signature references incorrect RFC
Documentation of DSA signatures (https://cryptography.io/en/latest/hazmat/primitives/asymmetric/dsa/#signing) references RFC 6979 which sounds strange. Same for the naming of de/encoding functions at https://cryptography.io/en/latest/hazmat/primitives/asymmetric/utils/#cryptography.hazmat.primitives.asymmetric.utils.decode_rfc6979_signature

But that RFC doesn't actually define the {r,s} encoding. The actual asn1 module can be found in RFC 3279 which defines both dsa-with-sha1 signature and Dss-Sig-Value which is the {r,s} sequence.

The references to RFC 6979 are actually unfortunate, because it defines deterministic DSA signatures, while cryptography.io exposes the randomised version using openssl's `DSA_sign`.

reaperhulk:
fixes #2246


alex:
Haven't substantivively reviewed yet, but can I propose this be an _extended_ deprecation period for this, not our normal 2 release one, since this is a function that's actually called by users, as opposed to the largely interface stuff.

reaperhulk:
I have no objection to an extended deprecation. Do we want to document what that is? How about triple the length? (6 releases) I wouldn't call documenting that a 1.0 blocker, but we should agree in principle at least.

Ayrx:
I feel like the CHANGELOG should state the reason why it's being renamed? Either in prose or with a link to the issue.

PR #2252, opened 2015-08-10
Update x509 docs to use namespace OIDs

reaperhulk:


docs/x509/tutorial.rst
```diff
@@ -57,14 +57,15 @@ a few details:
 .. code-block:: pycon
 
     >>> from cryptography import x509
+    >>> from cryptography.x509 import oid
```
alex: This is inconsistant with how we did this import everywhere else. Why not just import `NameOID` directly?

PR #2262, opened 2015-08-11
Ensure early exeption on non-bytes signature

viraptor:
Signature must be in bytes. If the check is skipped, verify() can
explode later in cffi call in _verify_pkey_ctx() for example.


reaperhulk:
This needs tests to check for the exception when passing non-bytes type and also the same logic should be added to _ECDSAVerificationContext.

alex:
@reaperhulk sorry I didn't look at this before -- our convention has mostly been to do these checks in `verifier()`, rather than the constructor. How would you feel about a PR to move those?

reaperhulk:
I thought about that a bit but originally decided it wasn't an issue since either way they'll have to be implemented in every backend. However, you're right that we're being inconsistent on where we check compared to other places in the code so I'm fine with changing it.

Issue #2255, opened 2015-08-11
add a get_extension_by_class to Extensions


reaperhulk:
We don't really need to require an OID and it would be friendlier to pass the class.

queenp:
Hi, I expressed an interest in this earlier on IRC, but don't quite get what the issue entails. Happy to do legwork reading up, but is there anywhere else I could read up on the API design/milestone stuff going on or to figure out what's required?

reaperhulk:
We have a method on the `Extensions` class called `get_extension_by_oid`. This issue is for adding a method called `get_extension_by_class` that takes an extension class (e.g. `BasicConstraints`) as an argument and returns the associated `Extension`.

queenp:
Gottit, thanks.

alex:
This is done!

PR #2267, opened 2015-08-12
added get_extension_for_class #2255

queenp:
Note: I've used "for", rather than get_extension_by_class, matching the original "get_extension_for_oid".

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -889,6 +889,21 @@ X.509 Extensions
             >>> cert.extensions.get_extension_for_oid(ExtensionOID.BASIC_CONSTRAINTS)
             <Extension(oid=<ObjectIdentifier(oid=2.5.29.19, name=basicConstraints)>, critical=True, value=<BasicConstraints(ca=True, path_length=None)>)>
 
+    .. method:: get_extension_for_class(extclass)
+
+        :param extclass: An extension class.
+
+        :returns: An instance of the extension class.
+
+        :raises cryptography.x509.ExtensionNotFound: If the certificate does
+            not have the extension requested.
+
+        .. doctest::
+
+            >>> from cryptography import x509
+            >>> cert.extensions.get_extension_for_class(x509.BasicConstraints)
+            <Extension(oid=<ObjectIdentifier(oid=2.5.29.19, name=basicConstraints)>, critical=True, value=<BasicConstraints(ca=True, path_length=None)>)>
+
 .. class:: Extension
 
     .. versionadded:: 0.9
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -891,6 +891,8 @@ X.509 Extensions
 
     .. method:: get_extension_for_class(extclass)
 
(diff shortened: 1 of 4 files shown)


reaperhulk:
Looks like you need to rebase this against current master to resolve some conflicts (we've been moving things around recently)

queenp:
Think it's sorted now.

src/cryptography/x509/extensions.py
```diff
@@ -80,6 +80,13 @@ def get_extension_for_oid(self, oid):
 
         raise ExtensionNotFound("No {0} extension was found".format(oid), oid)
 
+    def get_extension_for_class(self, extclass):
+        for ext in self:
+            if type(ext.value) == extclass:
```
reaperhulk: This should be done with isinstance rather than type.

reaperhulk:
You'll need to add a test that uses this method. We require 100% code coverage.

Thanks for your work so far!

queenp:
Sure thing, would you recommend that I mirror the full range of get_extension_for_oid checks?

reaperhulk:
In this case a single test that loads a cert, then uses the method to get the extension (and asserts on the result) should be sufficient.

reaperhulk:
Looks like to get full coverage you'll need to add a test that raises `ExtensionNotFound`.

reaperhulk:
jenkins, ok to test

queenp:
Sorry for the messy commits. First time contributing to a strictly quality controlled repo. Bit of a learning curve.

queenp:
I'm completely at a loss wrt what the branching condition is. My function and tests are almost identical to get_extension_for_oid. I can't find any condition that is left untested unless it's a quirk of isinstance()

docs/x509/reference.rst
```diff
@@ -889,6 +889,21 @@ X.509 Extensions
             >>> cert.extensions.get_extension_for_oid(ExtensionOID.BASIC_CONSTRAINTS)
             <Extension(oid=<ObjectIdentifier(oid=2.5.29.19, name=basicConstraints)>, critical=True, value=<BasicConstraints(ca=True, path_length=None)>)>
 
+    .. method:: get_extension_for_class(extclass)
+
```
reaperhulk: Could you add a `.. versionadded:: 1.1` here?

queenp:
Okay, I added that version tag to the doc page and literally didn't touch the test code or the function and code coverage panicked on the "isinstance" line again.

reaperhulk:
Weird... For the case where it doesn't find the extension try switching to using `x509/cryptography.io.pem` and have it look for `IssuerAlternativeName` (since BasicConstraints are actually present in the cryptography.io.pem cert)

queenp:
Sure, but that's not what's failing, so far as I can tell. It intermittently fails and passes the "if isinstance()" line (I've tried to eliminate any problems by making it as close to the branching logic of get_extension_for_oid() as possible).

reaperhulk:
Thanks @queenp! If you'd like to have your name in the AUTHORS file feel free to send a PR :)

PR #2268, opened 2015-08-12
added attribs to SSL_SESSION, SSL_SESSION_print. CIPHER_description

MiWCryptAnalytics:
For a project I am working on I need some session reflection directly from python. This change adds the required openssl SSL_SESSION struct attributes and functions within OpenSSL API

Additionally being able to query the TLS CipherSuite components used in a particular session with the CIPHER_description command lets us see what cipher mode we are operating in, Kx, etc.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -130,8 +130,17 @@ typedef ... SSL_METHOD;
 typedef ... SSL_CTX;
 
 typedef struct {
+    unsigned int key_arg_length;
+    unsigned char key_arg[...];
     int master_key_length;
     unsigned char master_key[...];
+    unsigned int session_id_length;
+    unsigned char session_id[...];
+    unsigned int sid_ctx_length;
+    unsigned char sid_ctx[...];
+    unsigned char *tlsext_tick;
+    size_t tlsext_ticklen;
+    long tlsext_tick_lifetime_hint;
     ...;
 } SSL_SESSION;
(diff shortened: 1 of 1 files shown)


alex:
Jenkins, ok to test.

MiWCryptAnalytics:
Hey thats some mighty fine tests you got there -- 
this breaks the build with libressl and centos5 (openssl 0.9.8)
I'll investigate now. Its possibly to do with function definition changes

It will be good come December this year when openssl 0.9.8 is out of support!

MiWCryptAnalytics:
centos 5 / rhel 5 uses a backported openssl 0.9.8e that does not contain TLS Extensions support. :-(

reaperhulk:
@MiWCryptAnalytics I think we can handle this but it's annoyingly complex. We'll need to add some C functions in CUSTOMIZATIONS that take an SSL_SESSION struct, dereference, and return the value you want (something like `Cryptography_get0_tlsext_hostname`, etc). We can then make those conditional bindings by adding a `Cryptography_HAS_TLSEXT_FIELDS` and stripping the functions on OpenSSLs that don't support it.

If you'd like, I can take over this PR to accomplish this as it requires relatively deep knowledge of the vagaries of our build process.

MiWCryptAnalytics:
@reaperhulk yes please :-) supporting openssl 0.9.8e is a nightmare, i wish you luck!

alex:
This was merged in #2282

PR #2276, opened 2015-08-14
Move signature type checks

viraptor:
Move the point of checking signatures, as suggested by alex in PR 2262.


reaperhulk:
LGTM, but since this was @alex's suggestion I'll let him confirm.

alex:
jenkins, ok to test

PR #2277, opened 2015-08-15
Update tutorial.rst as it was missing an import

shadyabhi:
The tutorial was missing an import.


reaperhulk:
jenkins, ok to test

PR #2279, opened 2015-08-16
Make py3pep8 tox target use a general python3.

Ayrx:
This saves us the effort from updating this file with every Python release. I don't think a specific Python 3 version matters when it comes to pep8 linting.


reaperhulk:
I have no objection to this...@alex?

PR #2282, opened 2015-08-19
additional SSL bindings

reaperhulk:
refs #2268 

This should steal the easy parts, but might still need a tweak or two.

One unfortunate thing about de-opaquing structs is that OpenSSL 1.1 will break this (as structs are all opaque and accessor functions exist). We'll cross that bridge when 1.1 comes out though...

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -242,19 +242,21 @@ void SSL_CTX_set_client_CA_list(SSL_CTX *, Cryptography_STACK_OF_X509_NAME *);
 
 /*  SSL_SESSION */
 void SSL_SESSION_free(SSL_SESSION *);
-int SSL_SESSION_print(BIO *, const SSL_SESSION *);
 
 /* Information about actually used cipher */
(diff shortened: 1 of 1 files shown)


reaperhulk:
More things need fixing, will look at it soon ;)

MiWCryptAnalytics:
@reaperhulk you rock! Having the 
<code>
int SSL_SESSION_print(BIO *fp, const SSL_SESSION *ses);
 char *SSL_CIPHER_description(const SSL_CIPHER *, char *buf, int size);
</code> functions is great! 

Presently I am accessing the SESSION struct attributes but being able to regex the SESSION_print output should offer the same visibility of the struct attribs
<code>
size_t tlsext_ticklen;
long tlsext_tick_lifetime_hint;
</code>
This is nice because the function exists in 0.9.8e (RHEL5) and the string simply wont contain the TLS ticket value. Later versions with TLSext will print these values.
If I recall the print function is only used in app s_client/s_server but its somewhat like an accessor function and one can use it until we have proper ones. They won't exist in 0.9.8e however ;-)

PR #2285, opened 2015-08-20
raise an error if trying to install on PyPy < 2.6

### LINKED ISSUE #2283: Installing cryptography with old pypy builds (and caches) bad wheels
If I install cryptography with pypy 2.5.1, the new cffi build mechanism isn't invoked and I get an installed package without any of the compiled modules. However, pip builds and caches `cryptography-1.0-py2-none-any.whl` which is platform-independent (because it has no compiled modules). This bad wheel then gets used for any further installs with any Python 2 interpreter, and happily installs a cryptography without any of the interesting bits.

To illustrate with an example:
```console
root@e1ffcbbe8f9d:/foo$ # Before cryptograpy is installed:
root@e1ffcbbe8f9d:/foo$ find ~/.cache/pip/wheels -name 'cryptography*'
root@e1ffcbbe8f9d:/foo$ /env-pypy-2.6.0/bin/pypy -c 'import cryptography.hazmat.bindings._openssl' 2>&1 | grep ImportError
ImportError: No module named cryptography

root@e1ffcbbe8f9d:/foo$ # Install cryptography using old pypy:
root@e1ffcbbe8f9d:/foo$ /env-pypy-2.5.1/bin/pip install cryptography -q
root@e1ffcbbe8f9d:/foo$ find ~/.cache/pip/wheels -name 'cryptography*'
/root/.cache/pip/wheels/6d/db/24/a5bbe660e3b632abb007ba6d8c91b5fa51a83bb53544c16709/cryptography-1.0-py2-none-any.whl

root@e1ffcbbe8f9d:/foo$ # After that, install cryptography using new pypy:
root@e1ffcbbe8f9d:/foo$ /env-pypy-2.6.0/bin/pip install cryptography -q
root@e1ffcbbe8f9d:/foo$ find ~/.cache/pip/wheels -name 'cryptography*'
/root/.cache/pip/wheels/6d/db/24/a5bbe660e3b632abb007ba6d8c91b5fa51a83bb53544c16709/cryptography-1.0-py2-none-any.whl

root@e1ffcbbe8f9d:/foo$ # The installed cryptography is now broken:
root@e1ffcbbe8f9d:/foo$ /env-pypy-2.6.0/bin/pypy -c 'import cryptography.hazmat.bindings._openssl' 2>&1 | grep ImportError
ImportError: No module named cryptography.hazmat.bindings._openssl
```

I think a reasonable solution is to explicitly fail to build or install on older versions of pypy. This would prevent the broken wheel from being being built, and would also give a useful error at install time instead of silently installing a nonfunctional cryptography.

reaperhulk:
fixes #2283 

Does `RuntimeError` make sense here?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,13 @@ if sys.version_info < (3, 4):
 if sys.version_info < (3, 3):
     requirements.append("ipaddress")
 
-if platform.python_implementation() != "PyPy":
+if platform.python_implementation() == "PyPy":
+    if sys.pypy_version_info < (2, 6):
+        raise RuntimeError(
+            "cryptography 1.0 is not compatible with PyPy < 2.6. Please "
+            "upgrade PyPy to use this library."
+        )
+else:
     requirements.append("cffi>=1.1.0")
     setup_requirements.append("cffi>=1.1.0")
 


alex:
`RuntimeError` seems fine

setup.py
```diff
@@ -45,7 +45,13 @@
 if sys.version_info < (3, 3):
     requirements.append("ipaddress")
 
-if platform.python_implementation() != "PyPy":
+if platform.python_implementation() == "PyPy":
+    if sys.pypy_version_info < (2, 6):
+        raise RuntimeError(
+            "cryptography 1.0 is not compatible with PyPy < 2.6. Please "
+            "upgrade PyPy or use an older version of cryptography."
```
alex: Drop the "older version of cryptography" bit. Dont' give people a wrong solution :-)

reaperhulk: Hehe, I was conflicted about that myself.

PR #2292, opened 2015-08-22
set the default stringmask to utf8

### LINKED ISSUE #2291: Use utf-8 strings in generated certs
Currently when adding unicode string to cert it end up as t61string or bmpstring.
Those are really ancient formats that are annoying to support in new code.

With some digging I would following workaround:

```
backend._lib.ASN1_STRING_set_default_mask_asc(b"utf8only")
```
As the comment says:

>  utf8only : only use UTF8Strings (RFC2459 recommendation for 2004)

Please make it default.  There are already bunch of root certs using this setting, so it's safe to enable for private certs too, at 2015...

reaperhulk:
This corrects a problem where older OpenSSL versions don't do this by default. fixes #2291

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -521,6 +521,12 @@ class Backend(object):
         self._ffi = self._binding.ffi
         self._lib = self._binding.lib
 
+        # Set the default string mask for encoding ASN1 strings to UTF8. This
+        # is the default for newer OpenSSLs for several years and is
+        # recommended in RFC 2459.
+        res = self._lib.ASN1_STRING_set_default_mask_asc(b"utf8only")
+        assert res == 1
+
         self._binding.init_static_locks()
 
         # adds all ciphers/digests for EVP


alex:
I don't love changing global state like this, but I don't have a better suggestion either.

alex:
Does not doing this actually break anything, or is it just unhygenic?

reaperhulk:
Once a UTF8 char gets encoded as T61String or BMPString software can't roundtrip it back to UTF8 properly unless it's already written to handle the broken encoding. OpenSSL itself will actually mangle it.

reaperhulk:
We actually changed this in pyopenssl a few months ago (https://github.com/pyca/pyopenssl/pull/234)

alex:
Ok. Thanks.

On Sat, Aug 22, 2015 at 5:35 PM, Paul Kehrer <notifications@github.com>
wrote:

> Once a UTF8 char gets encoded as T61String or BMPString software can't
> roundtrip it back to UTF8 properly unless it's already written to handle
> the broken encoding. OpenSSL itself will actually mangle it.
>
> —
> <https://github.com/pyca/cryptography/pull/2292#issuecomment-133756254>.
>



-- 

PR #2293, opened 2015-08-24
idempotent engine add

reaperhulk:
Threading issues keep cropping up. ENGINE_add ideally acquires a lock at the C layer via CRYPTO_w_lock (provided you have registered the locking callbacks) so let's try to use that...

I don't think this is a real fix, but it should be an improvement over the current situation and may resolve the issues for some people. We should keep thinking about how to truly fix this.

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/binding.py
+++ b/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -97,10 +97,6 @@ class Binding(object):
     @classmethod
     def _register_osrandom_engine(cls):
         _openssl_assert(cls.lib, cls.lib.ERR_peek_error() == 0)
-        looked_up_engine = cls.lib.ENGINE_by_id(cls._osrandom_engine_id)
-        if looked_up_engine != ffi.NULL:
-            raise RuntimeError("osrandom engine already registered")
-
         cls.lib.ERR_clear_error()
 
         engine = cls.lib.ENGINE_new()
@@ -113,7 +109,12 @@ class Binding(object):
             result = cls.lib.ENGINE_set_RAND(engine, cls._osrandom_method)
             _openssl_assert(cls.lib, result == 1)
             result = cls.lib.ENGINE_add(engine)
-            _openssl_assert(cls.lib, result == 1)
+            if result != 1:
+                errors = _consume_errors(cls.lib)
+                assert (
+                    errors[0].reason == cls.lib.ENGINE_R_CONFLICTING_ENGINE_ID
+                )
+
         finally:
             result = cls.lib.ENGINE_free(engine)
             _openssl_assert(cls.lib, result == 1)
@@ -171,3 +172,6 @@ class Binding(object):
                     mode, n, file, line
                 )
             )
+
+# init the static locks so we have a locking callback in C for engine init
+Binding.init_static_locks()
--- a/src/cryptography/hazmat/bindings/openssl/binding.py
+++ b/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -97,7 +97,6 @@ class Binding(object):
     @classmethod
     def _register_osrandom_engine(cls):
         _openssl_assert(cls.lib, cls.lib.ERR_peek_error() == 0)
-        cls.lib.ERR_clear_error()
 
         engine = cls.lib.ENGINE_new()
         _openssl_assert(cls.lib, engine != cls.ffi.NULL)
@@ -111,7 +110,8 @@ class Binding(object):
             result = cls.lib.ENGINE_add(engine)
             if result != 1:
                 errors = _consume_errors(cls.lib)
-                assert (
+                _openssl_assert(
+                    cls.lib,
                     errors[0].reason == cls.lib.ENGINE_R_CONFLICTING_ENGINE_ID
                 )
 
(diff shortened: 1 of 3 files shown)


src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -81,7 +77,12 @@ def _register_osrandom_engine(cls):
             result = cls.lib.ENGINE_set_RAND(engine, cls._osrandom_method)
             assert result == 1
             result = cls.lib.ENGINE_add(engine)
-            assert result == 1
+            if result != 1:
```
public: Does a non-1 return status always mean that the engine was already added?

reaperhulk: No, it can probably fail for other reasons. I can assert that the error stack contains a specific code to confirm that. That's mostly safe (sometimes OpenSSL changes error codes, but it seems unlikely that they'd change this one).

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -81,7 +77,20 @@ def _register_osrandom_engine(cls):
...
+                while True:
+                    code = cls.lib.ERR_get_error()
+                    if code == 0:
+                        break
+
+                    errors.append(code)
+
+                # the following error code corresponds to "conflicting engine
+                # id" in ENGINE_LIST_ADD
+                assert 638025831 in errors
```
alex: Is there not a constant for this in OpenSSL?

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -81,7 +77,20 @@ def _register_osrandom_engine(cls):
...
-            assert result == 1
+            if result != 1:
+                # Engine already added. Clear the error stack.
+                errors = []
+                while True:
+                    code = cls.lib.ERR_get_error()
+                    if code == 0:
+                        break
+
+                    errors.append(code)
```
alex: This code already exists in the backend, can we factor it into a standalone function that is used both here and there?

glyph:
Since this is where @public is suggesting the multi-interpreter fix will come from: If you're calling `init_static_locks` as part of binding initialization rather than at import time as specifically protected by the import lock, then the lock initialization *itself* is prone to race conditions and therefore differing ideas about what locks to use, and it reverts to previous case.

On the cffi ticket, arigo suggested to me that one could actually explicitly acquire the import lock (which is reentrant) with [`imp.acquire_lock()`](https://docs.python.org/2/library/imp.html#imp.acquire_lock).

Although this is possibly a problem because `imp` is [deprecated in 3](https://docs.python.org/3/library/imp.html#imp.acquire_lock), and 3.x apparently uses per-module locks and does *not* have a global import lock any more…

reaperhulk:
This PR has been updated to try the `imp.acquire_lock()` idea in #2299.

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -139,3 +149,17 @@ def _lock_cb(cls, mode, n, file, line):
...
+# (including across all subinterpreters if we're in a mod_wsgi environment)
+# initializing some C level locks, then initializing the binding and adding the
+# OS random engine. This should prevent engine addition race conditions,
+# although it's still unclear what happens when a subinterpreter which
+# has registered the locks is destroyed. Do the callback functions still work
+# or are all the objects associated with that subinterpreter destroyed?
+imp.acquire_lock()
+Binding.init_static_locks()
+Binding._ensure_ffi_initialized()
+imp.release_lock()
```
alex: This belongs in a different patch from the "make the other function idempotent"

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -97,10 +97,6 @@ def __init__(self):
     @classmethod
     def _register_osrandom_engine(cls):
         _openssl_assert(cls.lib, cls.lib.ERR_peek_error() == 0)
-        looked_up_engine = cls.lib.ENGINE_by_id(cls._osrandom_engine_id)
-        if looked_up_engine != ffi.NULL:
-            raise RuntimeError("osrandom engine already registered")
-
         cls.lib.ERR_clear_error()
```
alex: This line can be removed now

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -113,7 +109,12 @@ def _register_osrandom_engine(cls):
             result = cls.lib.ENGINE_set_RAND(engine, cls._osrandom_method)
             _openssl_assert(cls.lib, result == 1)
             result = cls.lib.ENGINE_add(engine)
-            _openssl_assert(cls.lib, result == 1)
+            if result != 1:
+                errors = _consume_errors(cls.lib)
+                assert (
+                    errors[0].reason == cls.lib.ENGINE_R_CONFLICTING_ENGINE_ID
+                )
```
alex: Should this be an `openssl_assert`?

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -171,3 +172,6 @@ def _lock_cb(cls, mode, n, file, line):
                     mode, n, file, line
                 )
             )
+
+# init the static locks so we have a locking callback in C for engine init
+Binding.init_static_locks()
```
alex: Should this go into a separate PR?

PR #2294, opened 2015-08-25
Adds name property to ObjectIdentifier

bjmc:
It would be convenient for my use case to have a `.name` property on the the ObjectIdentifier object. For example, if I wanted to get a dictionary of the details of a certificate subject, I could do:

`details = dict((attr.oid.name, attr.value) for attr in cert.subject)` or similar.

I tried this first using the `utils.read_only_property` function, but it triggered an infinite recursion error because `self` is used to look up the name. AFAIK, using the `@property` decorator will make a read-only attribute anyway, unless you specify a setter.


reaperhulk:
jenkins, ok to test

alex:
I'm not sure how I feel about this. Names are not really official, and are subject to change.

bjmc:
@alex Feel free to change the name of the property to something less canonical-sounding, like `display_name` or `human_readable` or something.

reaperhulk:
Certain names are extremely well-defined (e.g. name OIDs, extended key usage OIDs, and extension OIDs), so it'd be nice to expose these more readily, but names like `id-qt-cps` are just the name of the structure in an RFC and unless we're willing to state unequivocally that we'll Never Change Them™ I'm reluctant to have them be part of the public API.

bjmc:
How would you feel about adding this only to NameOID (or other selected, well-defined OIDs), and making it underscore-prefixed `._name` so it's clear that this isn't part of the normal, public API?

reaperhulk:
If we make it `_name` and don't document it I'm actually fine with it using the existing dict. Of course, being private that means we won't guarantee this method won't change/be renamed without warning in the future.

bjmc:
I changed this to be `_name`. It's fine by me if this is an undocumented, private method, and you can change it in the future if you need to.

alex:
Can you please merge/rebase on master? Then this LGTM.

bjmc:
Merged in master.

alex:
Jenkins, retest this please.

reaperhulk:
I'm :+1: on this.

alex:
Jenkins, retest this please.

reaperhulk:
jenkins, retest this please

bjmc:
Seem like Jenkins took labor day off?

reaperhulk:
Jenkins is having issues related to a plugin upgrade. I'll resolve those as soon as I can. :cry:

reaperhulk:
Let's try again. jenkins, retest this please.

PR #2295, opened 2015-08-25
Add set_cert_cb 

mhils:
This PR adds OpenSSL'S `set_cert_cb` to cryptography. I'm not an expert on cffi and closely followed what you did for ALPN, so please let me know if there are any issues. :smiley:

PR #2296, opened 2015-08-25
add conditional removal for SET_CERT_CB

reaperhulk:
Adds conditional I didn't notice was missing in #2295 

A followup PR could be done to add a test that imports the dictionary in `_conditional` and loops over it to verify that anything not present has been removed. That would have caught this mistake.


reaperhulk:
Err, no it wouldn't, since the HAS_* wouldn't be present in the dictionary the test would blithely pass.

Issue #2300, opened 2015-08-28
SSL_renegotiate is not present in OpenSSL bindings


kjav:
The function 'SSL_renegotiate' is present in OpenSSL but does not appear to be present in the bindings.

alex:
`int SSL_renegotiate(SSL *)` is the signature.

kjav:
I don't understand why the binding isn't present - I can't find the signature in the _conditional.py file

alex:
Things have to be explicitly bound, they don't automatically show up.

kjav:
Where are they bound (in which file)?

PR #2303, opened 2015-08-28
Added bindings for SSL_renegotiate_<pending/abbreviated>

kjav:
This is to facilitate their use in the pyopenssl library.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -206,6 +206,8 @@ int SSL_get_error(const SSL *, int);
 int SSL_do_handshake(SSL *);
 int SSL_shutdown(SSL *);
 int SSL_renegotiate(SSL *);
+int SSL_renegotiate_abbreviated(SSL *);
+int SSL_renegotiate_pending(SSL *);
 const char *SSL_get_cipher_list(const SSL *, int);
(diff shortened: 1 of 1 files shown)


kjav:
Related to issue https://github.com/pyca/cryptography/issues/2300

alex:
Jenkins, ok to test.

reaperhulk:
jenkins, ok to test

kjav:
@alex, I have a fork of the pyopenssl library which uses these bindings to add renegotiation functionality. When should I create the merge request? I'm worried the tests will fail as the cryptography version it uses might not be up to date with the changes in this merge.

reaperhulk:
Once this merges you can open a PR against pyopenssl. There are tests there that run against cryptography master. The tests against the latest release version will fail, but that's expected for now.

alex:
I don't think `SSL_renegotiate_abbreviated` exists in all versions of OpenSSL?

kjav:
Do you have an example?

reaperhulk:
Looks like it was added sometime in 1.0.1 so that'll have to be conditionally bound. @kjav look at the jenkins status check on this PR right now if you want to see where it's failing.

kjav:
I added this binding on a whim - it is not required for the pyopenssl renegotiation support (the reason for this merge). I have removed it.

kjav:
The checks have passed - can this be merged with master?

PR #2306, opened 2015-08-29
Support static linking on OS X

reaperhulk:
This will become our preferred method for building so that we can start shipping wheels for OS X.

We can't test static linking against Apple's 0.9.8 as they don't ship static libs, so we continue to link dynamically there (this includes all the jenkins builders, which build against system OpenSSL).


src/_cffi_src/build_openssl.py
```diff
@@ -11,15 +12,27 @@
 
 def _get_openssl_libraries(platform):
     # OpenSSL goes by a different library name on different operating systems.
-    if platform != "win32":
+    if platform == "darwin":
+        return _osx_libraries(
+            os.environ.get("CRYPTOGRAPHY_BUILD_STATIC", None)
```
alex: `None` is unnecsary here.

dstufft:
It's a little strange to have a environment variable called ``CRYPTOGRAPHY_BUILD_STATIC`` which doesn't actually cause it to be a static build, it just makes it so that we don't add any default libraries and you're expected to pass the libraries in from the outside. I think either this flag should cause the build script to actually build a static library, or we should name it something actually descriptive of what it does and not restrict it to just OSX.

reaperhulk:
Not restricting it to OS X is an interesting idea...

It's not possible to actually have it build statically on OS X with a single flag unless you do some magic with detecting the presence of homebrew, etc. I'm pretty against that level of magic so maybe a different name is appropriate. Do you have a suggestion?

reaperhulk:
Thinking about this a bit more I agree the name should change, but I **don't** think it should apply to anything other than OS X. On linux you need to pass the libraries plus a `-static` argument, so trying to make this more generic wouldn't make things simpler.

How does `CRYPTOGRAPHY_OSX_NO_LINK_FLAGS` sound?

dstufft:
I don't have a strong opinion on if it should work for other OSs or if it should just cram ``OSX`` into the name. The only reason I thought about other OSs was if, for some reason, there was something that had the same API as OpenSSL but required different libraries to link to, but that's probably not a very major use case.

I'm fine with ``CRYPTOGRAPHY_OSX_NO_LINK_FLAGS``.

reaperhulk:
jenkins, retest this please

PR #2307, opened 2015-08-29
Resolve an unusual test bug related to initializing the bindings

reaperhulk:
To make calls against the "SSL" parts of OpenSSL you need to call
SSL_library_init. There are multiple ways this can be called:

* If you're using the same OpenSSL in cryptography as you are in your
  Python then Python will call it for you.
* If you import the openssl backend.

These tests need SSL_library_init to be called. When run in our CI
SSL_library_init is called because during the parametrization step
the OpenSSL backend is imported (thus triggering it). However, you
can also run tests directly via py.test and without this change

py.test tests/hazmat/bindings/test_openssl.py

would crash if you had cryptography linked against a different OpenSSL
than your Python used.

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/binding.py
+++ b/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -92,6 +92,12 @@ class Binding(object):
             if not cls._lib_loaded:
                 cls.lib = build_conditional_library(lib, CONDITIONAL_NAMES)
                 cls._lib_loaded = True
+                # initialize the SSL library
+                cls.lib.SSL_library_init()
+                # adds all ciphers/digests for EVP
+                cls.lib.OpenSSL_add_all_algorithms()
+                # loads error strings for libcrypto and libssl functions
+                cls.lib.SSL_load_error_strings()
                 cls._register_osrandom_engine()
 
     @classmethod
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -529,13 +529,6 @@ class Backend(object):
 
         self._binding.init_static_locks()
 
-        # adds all ciphers/digests for EVP
-        self._lib.OpenSSL_add_all_algorithms()
-        # registers available SSL/TLS ciphers and digests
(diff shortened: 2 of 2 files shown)


alex:
Is there a good reason not to just have the bindings call `SSL_library_init`?

reaperhulk:
This revises the original approach and hoists library initialization (which we originally did in the backend) up into the binding. Since the bindings aren't really...useful...without calling these (and the calls themselves are idempotent) this should be a safe change. /me crosses fingers

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -91,14 +91,20 @@ def _ensure_ffi_initialized(cls):
         with cls._init_lock:
             if not cls._lib_loaded:
                 cls.lib = build_conditional_library(lib, CONDITIONAL_NAMES)
+                # init the OpenSSL locks
+                cls.init_static_locks()
```
alex: Moving this seems different than calling the OpenSSL init functions.

Also these should happen after setting `_lib_loaded` IMO

reaperhulk: I'll move init_static_locks back out. Although I think we should do that too it can be in a separate PR.

alex: Sounds good.

On Sat, Aug 29, 2015 at 9:05 PM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/2307#discussion_r38267563>:
>
> > @@ -91,14 +91,20 @@ def _ensure_ffi_initialized(cls):
> >          with cls._init_lock:
> >              if not cls._lib_loaded:
> >                  cls.lib = build_conditional_library(lib, CONDITIONAL_NAMES)
> > +                # init the OpenSSL locks
> > +                cls.init_static_locks()
>
> I'll move init_static_locks back out. Although I think we should do that
> too it can be in a separate PR.
>
> —
> <https://github.com/pyca/cryptography/pull/2307/files#r38267563>.
>



-- 

PR #2313, opened 2015-08-31
Random number docs

### LINKED ISSUE #2278: Random number generation docs are misleading
[The docs recommend against `random` module APIs and advocate `os.urandom()`](https://github.com/pyca/cryptography/blob/ae9ae087aeddb2157a6588a424ff5b14243f748c/docs/random-numbers.rst).

There is [`random.SystemRandom()`](https://docs.python.org/3/library/random.html#random.SystemRandom) that is built on top of `os.urandom()`:

```
>>> import random
>>> r = random.SystemRandom()
>>> r.getrandbits(160)
993486520646701389716655680137901746842047255567
>>> r.randint(2, 10)
4
>>> r.choice("abc")
'c'
>>> L = list(range(10))
>>> L
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
>>> r.shuffle(L)
>>> L
[3, 7, 1, 8, 2, 4, 0, 9, 6, 5]
```

I see that `getrandbits(160)` already has been reinvented as `int.from_bytes(os.urandom(20), byteorder="big")`.

Ayrx:
Fixes #2278.

I think he brings up a valid point because `random.SystemRandom().getrandbits()` *is* essentially `int.from_bytes(os.urandom(20), byteorder="big")`. No point reinventing the wheel when the standard module provides what we need.

### DIFF
--- a/docs/random-numbers.rst
+++ b/docs/random-numbers.rst
@@ -23,10 +23,11 @@ Windows.
 
 If you need your random number as an integer (for example, for
 :meth:`~cryptography.x509.CertificateBuilder.serial_number`), you can use
-``int.from_bytes`` to convert the result of ``os.urandom``:
+``getrandbits()`` from ``random.SystemRandom()``.
 
 .. code-block:: pycon
 
-    >>> serial = int.from_bytes(os.urandom(20), byteorder="big")
+    >>> r = random.SystemRandom()
+    >>> serial = r.getrandbits(160)
 
 .. _`always use your operating system's provided random number generator`: http://sockpuppet.org/blog/2014/02/25/safely-generate-random-numbers/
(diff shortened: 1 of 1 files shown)


skeuomorf:
I was under the impression that using `os.urandom()` directly was a conscious choice, given that the explicit nature of using it this way won't cause confusion for the "skimming" reader who's referred to tptacek's article. However, a case can be made for the cleanliness of using `SystemRandom()` as well.

Ayrx:
@skeuomorf I can buy that argument, but if so we should explicitly close this PR and the issue. :)

alex:
I believe @reaperhulk and I came to the same conclusion as @skeuomorf: The advice of "always use os.urandom" is simpler and less likely to lead to mistakes.

Ayrx:
@alex Cool. Should close this and #2278 then. :)

PR #2314, opened 2015-09-01
[WIP] can't redefine SSLv2_ functions on Solaris

etrauschke:
Trying to build cryptography on recent Solaris versions caused a problem with redefining SSLv2_ functions.

The cryptography code assumes, that if OPENSSL_NO_SSL2 is set, the following functions are not defined at all:
SSL_METHOD *SSLv2_method(void)
SSL_METHOD *SSLv2_client_method(void)
SSL_METHOD *SSLv2_server_method(void)

However, in Solaris the declarations have not been removed, just marked deprecated to not create linker issues with programs still looking for those.

This is a test to see if just putting in empty declarations instead of redefinitions will work for other platforms.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -431,9 +431,9 @@ static const long Cryptography_HAS_SECURE_RENEGOTIATION = 1;
 #endif
 #ifdef OPENSSL_NO_SSL2
 static const long Cryptography_HAS_SSL2 = 0;
-SSL_METHOD* (*SSLv2_method)(void) = NULL;
-SSL_METHOD* (*SSLv2_client_method)(void) = NULL;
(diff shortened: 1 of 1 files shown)


reaperhulk:
Unfortunately those methods are genuinely removed by the default distribution of OpenSSL when `OPENSSL_NO_SSL2` is set so the current approach errors since no symbol corresponding to the declaration is found in the object. Resolving this will probably require a macro check to see if it's being compiled on Solaris.

etrauschke:
Yes, that is what I was already expecting, I just wanted to confirm.
I think the best way to deal with this is to check for Solaris and the OpenSSL version we deprecated these methods. I'll check with our OpenSSL experts.

etrauschke:
Turns out our latest dev bits of OpenSSL will fix this issue, so I'm closing this. Sorry for the noise.

PR #2315, opened 2015-09-01
OpenSSL backend code for CRLs

etrauschke:
This is the backend code for the CRL objects. Looks like a bunch of additional automatic tests have been added so let me make sure they clear first.


etrauschke:
Looks like I have to change the test certs so that the 0.9.8 tests will not see a cert issuer extension marked as critical. Hopefully I can reuse some other cert for this, otherwise I might have to make a change to the vectors again.

etrauschke:
It looks like I'll have to change the test cert crl_all_reasons.pem so that the cert_issuer extension is not critical. Otherwise I would have to skip a bunch of tests in the 0.9.8 environments.

etrauschke:
So I'm not sure what the coverage thing is complaining about, I guess it's because the for loop never runs completely through (intentionally).
I guess if that is really important I can change the line to something like this:

```for r in crl.revoked_certificates.append(dummy_rev):```

**Comment by etrauschke** [comment]
Ended up completely changing the logic of the reason flag test. Coverage seems happy now.

**Comment by alex** [comment]
Jenkins, retest this please.

**Comment by reaperhulk** [comment]
jenkins, retest this please

**Comment by etrauschke** [comment]
Anybody could have a look at this?

File: docs/x509/reference.rst [hunk 1]
```diff
@@ -129,6 +144,52 @@ Loading Certificates
     >>> cert.serial
     2
 
+Loading Certificate Revocation Lists
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. function:: load_pem_x509_crl(data, backend)
+
+    .. versionadded:: 1.0
```
reaperhulk: This should be 1.1 now

docs/x509/reference.rst
```diff
@@ -129,6 +144,52 @@ Loading Certificates
...
+    :param backend: A backend supporting the
+        :class:`~cryptography.hazmat.backends.interfaces.X509Backend`
+        interface.
+
+    :returns: An instance of
+        :class:`~cryptography.x509.CertificateRevocationList`.
+
+.. function:: load_der_x509_crl(data, backend)
+
+    .. versionadded:: 1.0
```
reaperhulk: Same

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -30,32 +30,6 @@ def _obj2txt(backend, obj):
     return backend._ffi.buffer(buf, res)[:].decode()
 
 
-def _asn1_integer_to_int(backend, asn1_int):
```
reaperhulk: Would you mind moving this change to a separate PR that we can review/land first? That way we can shrink the size of this one and make it a bit easier to review.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -306,27 +280,12 @@ def public_key(self):
...
         asn1_time = self._backend._lib.X509_get_notBefore(self._x509)
-        return self._parse_asn1_time(asn1_time)
+        return self._backend._parse_asn1_time(asn1_time)
 
     @property
     def not_valid_after(self):
         asn1_time = self._backend._lib.X509_get_notAfter(self._x509)
-        return self._parse_asn1_time(asn1_time)
-
-    def _parse_asn1_time(self, asn1_time):
```
reaperhulk: This can be in the PR with the other moved functions as well.

reaperhulk:
It doesn't look like your merge captured all the commits (it's definitely missing the PR you opened earlier).

etrauschke:
Oh, please ignore this last push I merged my master and just did a push without specifying the branch so it updated this one unintentionally as well. I'm still testing my actual changes and will push once again once I('m done.

etrauschke:
Now the PR is ready for review. If this gets a bit cluttered I can also create a new PR.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,262 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+        if self.__last_update:
+            return self.__last_update
+
+        lu = self._backend._lib.X509_CRL_get_lastUpdate(self._x509_crl)
+        assert lu != self._backend._ffi.NULL
+        self.__last_update = self._backend._parse_asn1_time(lu)
+        return self.__last_update
+
+    @property
+    def revoked_certificates(self):
```
reaperhulk: I know this is in our current interface, but I'm wondering if it might be more consistent for us to just make CertificateRevocationList an iterable? That would look more like the way ``Extensions``, ``SubjectAlternativeName``, and several other x509 classes behave.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,262 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+        if self.__revoked:
+            return self.__revoked
+
+        revoked = self._backend._lib.X509_CRL_get_REVOKED(self._x509_crl)
+        assert revoked != self._backend._ffi.NULL
```
reaperhulk: I apologize for making you do this, but we've just switched (today) from asserts to making calls against `openssl_assert` when checking responses from the OpenSSL backend. Would you mind converting your asserts to this new approach?

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,262 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+        self.__serial_number = None
+        self.__revocation_date = None
+        self.__extensions = None
+
+    @property
+    def serial_number(self):
+        if self.__serial_number:
+            return self.__serial_number
+
+        bn = self._backend._lib.ASN1_INTEGER_to_BN(
```
reaperhulk: This can use `_asn1_integer_to_int` on the backend.

src/cryptography/hazmat/backends/openssl/x509.py
etrauschke: So you just want me to get rid of revoked_certificates() and replace it with an __iter__() and __len__() function? I can do that, I just tried to stay as close to the OpenSSL specification because I think this makes it easier for people to find out how to access the revoked certs.

src/cryptography/hazmat/backends/openssl/x509.py
etrauschke: I can do that.

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: Yeah, I think that's consistent with our current approach for x509 objects.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,262 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+        """
+        Returns the CertificateIssuer extension if it exists.
+        """
+        try:
+                return self.extensions.get_extension_for_oid(
+                    x509.OID_CERTIFICATE_ISSUER).value
+        except x509.ExtensionNotFound:
+                return None
+
+    def _build_crl_reason(self, ext):
```
reaperhulk: We've been calling these `decode` rather than `build`.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,262 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+        assert generalized_time != self._backend._ffi.NULL
+        generalized_time = self._backend._ffi.gc(
+            generalized_time, self._backend._lib.ASN1_GENERALIZEDTIME_free
+        )
+        time = self._backend._ffi.string(
+            self._backend._lib.ASN1_STRING_data(
+                self._backend._ffi.cast("ASN1_STRING *", generalized_time)
+            )
+        ).decode("ascii")
+        return datetime.datetime.strptime(time, "%Y%m%d%H%M%SZ")
```
reaperhulk: Do we have any existing extensions that are just bare values like this? The closest analogue to this extension is probably InhibitAnyPolicy, which is defined in the RFC as

```
   InhibitAnyPolicy ::= SkipCerts

   SkipCerts ::= INTEGER (0..MAX)
```

but we define an object with one property (`skip_certs`) for it. Do we want to do the same for invalidity date and reason code? It's arguably a pointless layer of indirection, so I'm interested in opinions. cc @alex

etrauschke:
I addressed the mentioned issues.
About CRLs as iterator: I added the iterator functionality but left the old interface in there too. Otherwise we'd have to change the interface in  src/cryptography/x509.py and I think it's easier for people to find what they're looking for since it's not intentional to just starting to iterate over the CRL object. However, I see it as a nice convenience function.

About the asserts: I changed all the ones in my code but saw that there are still some in x509.py and backend.py. Not sure if they need to get fixed as well.

About bare return values: I'd vote for leaving them like this. I feel like there is already a really thick layer of abstraction for getting the actual values of extensions or other x509 properties. I understand that this is necessary in a lot of cases but it's it's also a hassle to perform 10 rounds of unwrapping to get to the actual value. So if they are basic types, I would return them as such.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,267 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+                )
+
+            if oid == x509.OID_CRL_REASON:
+                value = self._decode_crl_reason(ext)
+            elif oid == x509.OID_INVALIDITY_DATE:
+                value = self._decode_invalidity_date(ext)
+            elif oid == x509.OID_CERTIFICATE_ISSUER and \
+                    self._backend._lib.OPENSSL_VERSION_NUMBER >= 0x10000000:
+                value = self._decode_cert_issuer(ext)
+            elif critical:
```
schlenk: The list is a bit short, would expect id-ce-deltaCRLIndicator and id-ce-certificateIssuer here too. The Authority Key Identifier is also often needed to find the right CA cert to validate the CRL.

etrauschke: These are CRL extensions, which are not supported in this PR, the ones you are looking at are CRL entry extensions, and this list is complete.

etrauschke:
<bump>

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,267 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+            if oid in seen_oids:
+                raise x509.DuplicateExtension(
+                    "Duplicate {0} extension found".format(oid), oid
+                )
+
+            if oid == x509.OID_CRL_REASON:
+                value = self._decode_crl_reason(ext)
+            elif oid == x509.OID_INVALIDITY_DATE:
+                value = self._decode_invalidity_date(ext)
+            elif oid == x509.OID_CERTIFICATE_ISSUER and \
```
reaperhulk: Could you switch this to using parentheses for multi-line rather than backslash?

Prior to 1.0 this extension was unsupported, correct? Does 0.9.8's CLI binary silently ignore it when you parse a CRL containing the extension?

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,267 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+
+        self.__extensions = x509.Extensions(extensions)
+        return self.__extensions
+
+    def get_reason(self):
+        """
+        Returns the CRLReason extension if it exists.
+        """
+        try:
+                return self.extensions.get_extension_for_oid(
```
reaperhulk: This indentation is off

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,267 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+                    x509.OID_CRL_REASON).value
+        except x509.ExtensionNotFound:
+                return None
+
+    def get_invalidity_date(self):
+        """
+        Returns the InvalidityDate extension if it exists.
+        """
+        try:
+                return self.extensions.get_extension_for_oid(
```
reaperhulk: Indentation

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,267 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+                    x509.OID_INVALIDITY_DATE).value
+        except x509.ExtensionNotFound:
+                return None
+
+    def get_certificate_issuer(self):
+        """
+        Returns the CertificateIssuer extension if it exists.
+        """
+        try:
+                return self.extensions.get_extension_for_oid(
```
reaperhulk: Indentation

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,267 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+    @property
+    def extensions(self):
+        if self.__extensions:
+            return self.__extensions
+
+        extensions = []
+        seen_oids = set()
+        extcount = self._backend._lib.X509_REVOKED_get_ext_count(
+            self._x509_revoked)
+        for i in range(0, extcount):
```
reaperhulk: You don't need the 0 argument in range here

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,267 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+
+    def _decode_crl_reason(self, ext):
+        enum = self._backend._lib.X509V3_EXT_d2i(ext)
+        self._backend.openssl_assert(enum != self._backend._ffi.NULL)
+        enum = self._backend._ffi.cast("ASN1_ENUMERATED *", enum)
+        enum = self._backend._ffi.gc(
+            enum, self._backend._lib.ASN1_ENUMERATED_free
+        )
+        code = self._backend._lib.ASN1_ENUMERATED_get(enum)
+        if code == 0:
```
reaperhulk: What about changing this to look like:

```python
try: 
    return {0: x509.ReasonFlags.unspecified, 1: x509.ReasonFlags.key_compromise, ...}[code]
except KeyError:
    raise ValueError
```

docs/x509/reference.rst
```diff
@@ -610,18 +704,35 @@ X.509 Revoked Certificate Object
 
         An integer representing the serial number of the revoked certificate.
 
+        .. doctest::
+
+            >>> crl.revoked_certificates[0].serial_number
```
reaperhulk: For this doctest (and others using `RevokedCertificate`) I'd prefer that we assigned `revoked_certificate = crl.revoked_certificates[0]` so that this would be `revoked_certificate.serial_number`.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,267 @@ def _decode_inhibit_any_policy(backend, asn1_int):
     return x509.InhibitAnyPolicy(skip_certs)
 
 
+@utils.register_interface(x509.RevokedCertificate)
+class _RevokedCertificate(object):
+    def __init__(self, backend, x509_revoked):
+        self._backend = backend
+        self._x509_revoked = x509_revoked
+
+        self.__serial_number = None
```
reaperhulk: Double underscore is inconsistent with our typical approach of single underscore -- would you mind changing it?

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,267 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+        return x509.GeneralNames(_decode_general_names(self._backend, gns))
+
+
+@utils.register_interface(x509.CertificateRevocationList)
+class _CertificateRevocationList(object):
+    def __init__(self, backend, x509_crl):
+        self._backend = backend
+        self._x509_crl = x509_crl
+
+        self.__revoked = None
```
reaperhulk: Same as previous double underscore comment.

src/cryptography/hazmat/backends/openssl/x509.py
etrauschke: Yes, this extensions is not supported in 0.9.8. I can see that definitions (like OIDs) have been added in later versions of 0.9.8 but full parsing support didn't come until later.
I haven't tried it with the CLI but I'd assume it would be ignored unless it's marked critical. What exactly are you concerned about it here? I can whip up a vbox with an 0.9.8 environment and test but I'm not sure what we are looking for. 
If we try to use it here with a 0.9.8 library we will get a NULL pointer when we're trying to parse the extension with X509V3_EXT_d2i().

src/cryptography/hazmat/backends/openssl/x509.py
etrauschke: I can change that, but I'd assume we want this to be formatted a bit differently, basically having one line per reason.

src/cryptography/hazmat/backends/openssl/x509.py
etrauschke: Sorry about that, most of the time I work on projects using 8 space indents. I just didn't pay attention and will fix all of them. I'm surprised this didn't get noticed by the pep8 runs.

reaperhulk:
Three items remaining:

* The docs for `CertificateRevocationList` should be updated to note that the object is iterable.
* I'd like @alex's opinion on retaining `revoked_certificates`. The CRL object itself is iterable (and our general convention has been to have iterables rather than a property), so I'm leaning towards saying we should remove it (and maybe add a `__repr__` to make it clear that the object is iterable).
* This PR caches some of the conversions it does. I have no objection to this approach, but it is different from our previous behavior in X509 (the openssl certificate implementation re-parses the properties every time they're accessed). Do we want to open an issue for doing this elsewhere @alex?

src/cryptography/hazmat/backends/openssl/x509.py
etrauschke: No problem. I just wanted to make sure they can not be accessed from outside of the class instead of just suggesting that they shouldn't be used.

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: Yeah, definitely -- just wanted to write it quickly ;)

etrauschke:
I added the caching for performance reasons and due to the design principle you guys use that a x509_* object is basically immutable. So I think doing the conversion every time somebody reads a property seems very expensive. This might not be an issue for a certificate which doesn't have a lot of properties but if you do that for a CRL which has a bunch of revoked certificates it might get expensive. It's also very little code overhead.

About the `revoked_certificates()` method I already stated my point of view that I think it's more obvious for users of the class but I don't feel very strongly about leaving or removing it. The nice thing of keeping it is that you can easily access a particular revoked cert by doing `crl.revoked_certificates()[4]`. But I guess i could also just add an `__getitem__()` method to make that work.

etrauschke:
I fixed most of the stuff. Still waiting for a final decision about the revoked_certificates() method removal.

etrauschke:
jenkins, retest this please

etrauschke:
retest this please

reaperhulk:
Jenkins only listens to a few people, but in this case the problem was travis having a failed job so I went ahead and restarted that manually.

etrauschke:
Cool, thanks. It actually did restart the jenkins tests, but not the travis ones.

etrauschke:
I also believe I have to sync this with the latest master at some point. I just didn't want to make the review any harder than it already is. Once we have the design discussions hashed out, I can merge my changes with the latest master.

reaperhulk:
You actually don't have any conflicts (and we've removed the merge requirement to keep up-to-date since it was so brutally difficult with the size of our CI) so no need to merge/rebase against master unless you want to.

etrauschke:
Oh, good to know. I was wondering why it didn't complain about the mismatched base build anymore and thought that you might have created a new branch, not updated the old one and then github thought there were no changes anymore. But this makes sense now.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -637,6 +639,264 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+        if self._extensions:
+            return self._extensions
+
+        extensions = []
+        seen_oids = set()
+        extcount = self._backend._lib.X509_REVOKED_get_ext_count(
+            self._x509_revoked)
+        for i in range(extcount):
+            ext = self._backend._lib.X509_REVOKED_get_ext(
+                self._x509_revoked, i)
```
alex: Can this code be written in terms of the extension parser class, it looks like yet?

alex:
I'd advocate for dropping the caching, and only offering `__iter__`, no property.

src/cryptography/hazmat/backends/openssl/x509.py
etrauschke: After having a quick look at this I think it could work. Let me play with it.

etrauschke:
Why wouldn't you want the caching? If necessary, I can put that into a different PR and also add it to the other x509 objects.

etrauschke:
I was able to convert my code to use the X509ExtensionParser. However, I had to add a way for ignoring extensions if they are not supported by the backend. Let me know if the way I came up with is acceptable.

I also saw that someone grouped the OIDs for Revoked certificates in a class named CRLExtensionsOID. This is technically not correct because these are Revoked Extensions and will create trouble once we support actual CRL extensions. I renamed the class to RevokedExtensionOID, however, I could see that someone might want to name them RevokedCertificateExtensionOID. I'm open to that.

I also removed the revoked_certificates property (made it protected and intended for class use only) and fixed the docs accordingly.

reaperhulk:
@etrauschke: Let's rename the `CRLExtensionOID` in another PR (and we'll probably need to alias it back to the old name with a deprecation cycle since it's part of our public API).

reaperhulk:
Let's also discuss the caching in a separate PR since it'd be awesome to land this and caching is orthogonal to the vast majority of the work here. (Sorry to make you do some extra work!)

reaperhulk:
It looks like the certificateIssuer revoked certificate extension is also something we can handle in 0.9.8. It's true that calling `X509V3_EXT_d2i` on the `ext` doesn't work, but in this case we can do the following:

`ext` is an `X509_EXTENSION *` which has an `ASN1_OCTET_STRING *value`.
An `ASN1_OCTET_STRING *` is a typedef for the struct `asn1_string_st` which has `int length` and `unsigned char *data`.
We can bind `GENERAL_NAMES *d2i_GENERAL_NAMES(GENERAL_NAMES **, const unsigned char **, long);` which should allow us to use the ASN1_OCTET_STRING data/length to obtain the `GENERAL_NAMES *` that we need to pass to the `_decode_general_names` function.

etrauschke:
@reaperhulk With the CRLExtensionOID I probably should have paid more attention when this change was made. This is unfortunate because it is basically wrong in the current form. They are not CRL extensions. Not sure if you would break anyones code yet since the backend support hasn't been in so the number of people using it is probably 0. However, I understand how complicated it is to change committed interfaces. Just one thing to keep in mind, if the backend code goes back, people might start using these extensions and then it gets even harder to change things around.

etrauschke:
The plan with the certificateIssuer sounds pretty nifty but how would we integrate this with the X509ExtensionParser which does run X509V3_EXT_d2i for us. We would have to shoehorn some extra code in there which seems ugly. Maybe we can look at this again if we find additional extensions which require newer OpenSSL versions to be properly supported.

etrauschke:
Before I forget it, I also changed the exception message for unsupported critical extensions so that it actually tells you that it is critical. Otherwise the message is confusing because the rule is that unknown exceptions are to be ignored unless they are marked critical.

reaperhulk:
Right now there are no extensions that we parse conditionally on underyling OpenSSL version. I'd very much prefer to keep it that way since it's extremely difficult to explain to users how the "same" version of cryptography might exhibit significantly different behavior. At the very least we need to raise a warning.

It is a shame that the naming is wrong (and that's on me!), but we can transition away from that name with our standard deprecation strategy without too much pain. We can put in a PR to change the name after this PR, update the docs to all point at the new name, but then keep the old name for compatibility for the next 2 releases. It would then be completely dropped in the 1.3 release.

etrauschke:
Ok, let me take a stab at the certIssuer custom parsing. I'll also remove the caching stuff for now and undo the naming change for CRLExtensionOID.

etrauschke:
Found a way to make this work, let me know what you think.
I had to add a new CRL pem file to vectors which includes an empty/invalid certIssuer entry extension. This is necessary for code coverage since I can't use the generic bailout for an invalid extension anymore. Let me know if you want me to push that in a separate PR. Also, I had to add a new binding for *d2i_GENERAL_NAMES which you might also want in a separate PR. So just let me know and I'll do that.

reaperhulk:
Yeah if you wouldn't mind dropping those in separate PRs we can get them landed immediately :)

tests/test_x509.py
```diff
@@ -52,6 +52,252 @@ def _load_cert(filename, loader, backend):
...
+    def test_invalid_cert_issuer_ext(self, backend):
+        crl = _load_cert(
+            os.path.join(
+                "x509", "custom", "crl_inval_cert_issuer_entry_ext.pem"
+            ),
+            x509.load_pem_x509_crl,
+            backend
+        )
+
+        with pytest.raises(ValueError):
```
reaperhulk: Can you walk me through the logic here? How was this extension encoded badly such that it should error? What makes the empty extension trigger the extension is invalid path?

etrauschke: What I did is I created a new custom extension which has the same OID as the certificate issuer extension but has a custom value. The way I did this is like that (in C using OpenSSL):
```
X509_REVOKED *r = X509_REVOKED_new();
int nid = OBJ_create("2.5.29.29", "Test", "Test Extension");
X509_REVOKED_add1_ext_i2d(r, nid, NULL, 0, 0);
```

Then when we get to `_decode_cert_issuer(backend, ext)` we get a NULL pointer as data_ptr_ptr (or something else which isn't a `GENERAL_NAMES` type)  which goes into `d2i_GENERAL_NAMES()` and returns a NULL pointer for which I explicitly test.

If that is not good enough, I could also create this CRL so that it sets an integer instead of a NULL value. Everything which is not `GENERAL_NAMES` should trigger this. However, I think at that point we are fairly far out in terms of error cases, I expect this code to never be triggered in the real world. I just needed something to make coverage happy.

reaperhulk: I'm okay with this -- just wanted to make sure I understood the approach. Thanks!

etrauschke: Now that I think about this, I wonder if we should just silently ignore it. I mean for the other extensions we don't check if the values actually make sense and we can pass a NULL value to `_decode_general_names()`, which just leads to an empty GeneralNames object.

reaperhulk: What about just creating a bad extension (use an OID that doesn't make sense for the context, like EKU or KU)?

etrauschke: If you are talking about using the EKU or KU OID this would just go down the code path for that particular OID, which is not the code I put in. If you just want to have a EKU or KU value for the certIssuer OID that would be the same as setting it to an integer or some other non-fitting ASN1 value. And if the result would be NULL in this case (which it most likely would be), are we just gonna forward this to `_decode_general_names()` or still use the bailout mechanism.
I think I put the bailout mechanism in to have an equivalent to the NULL checking of the result from `X509V3_EXT_d2i`.

reaperhulk: Yeah sorry, I was thinking about the fact that it wouldn't be in _REVOKED_EXTENSION_HANDLERS, but it'd still work on the `EXT_d2i` call. That said, you could just use an OID that corresponds to no extension and guarantee that `EXT_d2i` will return NULL that way.

etrauschke: I did not write that code around `EXT_d2i` so I'd assume this is already tested somewhere else.

I think the only question which still stand is: If we get a NULL value from `d2i_GENERAL_NAMES`, do we treat that as an error or do we just pass it along the stack.

etrauschke: Any opinion on the above? I guess in real live this will not make much of a difference, it would just be an inconsistency between the certIssuer entry extension and all other extensions. But implementing the extensions parsing on our own already sets this one apart from others.

reaperhulk: Sorry, I owe you a review of this updated code again. Will get to it tonight.

etrauschke: Cool, thanks. I just wanted to see if we need to change the behavior again, because before we have decided on what should happen you might just waste your time reviewing code which needs to be changed anyway.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -194,18 +199,22 @@ def parse(self, backend, x509_obj):
...
-                d2i = backend._lib.X509V3_EXT_d2i(ext)
-                if d2i == backend._ffi.NULL:
-                    backend._consume_errors()
-                    raise ValueError(
-                        "The {0} extension is invalid and can't be "
-                        "parsed".format(oid)
-                    )
-
-                value = handler(backend, d2i)
+                if self.unsupported_exts and oid in self.unsupported_exts:
```
reaperhulk: Let's put a comment here to the effect of "for extensions that openssl doesn't directly support we pass the extension object itself through to the handler"

tests/test_x509.py
reaperhulk: Okay, looking at this more I'd prefer it to ignore it like a normal NULL. We should already be triggering this code path with another test so we shouldn't need a special test case at all now that you're using the ExtensionParser class, right?

etrauschke: Yes, for everything which goes through the regular processing with `X509V3_EXT_d2i` we already handle the NULL case and I haven't change any of that code. It would be just about the certIssuer extension parsing.

reaperhulk: Let's leave this as is with just an additional comment next to that value error that states it's present because we're parsing the extension ourselves and can't rely on the check in the extensionparser.

reaperhulk:
I've restarted the CI job in #2417. Once that merges you can rebase this and add the two comments and I'll do a final review. Once again, thanks for sticking with this!

reaperhulk:
The vector has now merged. If you add those two comments I'll do a final paranoia-based review to make sure I haven't missed anything and we can get this merged!

docs/x509/reference.rst
```diff
@@ -129,6 +144,52 @@ Loading Certificates
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. function:: load_pem_x509_crl(data, backend)
+
+    .. versionadded:: 1.1
+
+    Deserialize a certificate revocation list (CRL) from PEM encoded data. PEM
+    requests are base64 decoded and have delimiters that look like
+    ``-----BEGIN X509 CRL-----``. This format is also known as
+    PKCS#10.
```
reaperhulk: PKCS10 defines CSRs -- is there anything about CRLs in there?

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -646,6 +658,205 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+        return self._backend._parse_asn1_time(
+            self._x509_revoked.revocationDate)
+
+    @property
+    def extensions(self):
+        return _REVOKED_CERTIFICATE_EXTENSION_PARSER.parse(
+            self._backend, self._x509_revoked
+        )
+
+    def get_reason(self):
```
reaperhulk: These helper methods aren't currently part of the interface. I'd prefer we remove this from the current PR and we can discuss adding these in a future PR.

etrauschke: You guys are no fun ;-)
Now every time I want to access any of these I have to do this:
`rev.extensions.get_extension_for_oid(x509.OID_CRL_REASON)` instead of `rev.get_reason()`.

I understand that this principle doesn't work so well for certificate extensions (because there are a lot more) but the amount for revoked certificates is pretty small. I have some ideas to create a universal solution which also works for cert extensions. I'll create a PR once this goes back.

reaperhulk:
I have a few small things I'd like to clean up but I'm satisfied that it's ready for merge now. Huge thanks @etrauschke. Be sure to send a PR adding yourself to `AUTHORS.rst` when you get a chance!

PR #2323, opened 2015-09-04
Add support for RSA_R_OAEP_DECODING_ERROR error flag.

EnTeQuAk:
This pull request adds support for the ``RSA_R_OAEP_DECODING_ERROR`` error flag referenced here: https://github.com/openssl/openssl/blob/master/include/openssl/rsa.h#L626

This error code is being raised on error condition (e.g wrong private key) for this code: https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#decryption

I actually didn't see any tests for ``RSA_R_PKCS_DECODING_ERROR`` flag and don't really know how to test such flags. Please let me know if you need any, I'd be happy for any guidance here :smiley:

### DIFF
--- a/tests/hazmat/backends/test_openssl.py
+++ b/tests/hazmat/backends/test_openssl.py
@@ -479,16 +479,15 @@ class TestOpenSSLRSA(object):
             )
         )
 
-        decrypted = RSA_KEY_512_ALT.private_key(backend).decrypt(
-            ciphertext,
-            padding.OAEP(
-                mgf=padding.MGF1(algorithm=hashes.SHA1()),
-                algorithm=hashes.SHA1(),
-                label=None
+        with pytest.raises(ValueError):
+            RSA_KEY_512_ALT.private_key(backend).decrypt(
+                ciphertext,
+                padding.OAEP(
+                    mgf=padding.MGF1(algorithm=hashes.SHA1()),
+                    algorithm=hashes.SHA1(),
+                    label=None
+                )
             )
-        )
-
-        assert decrypted == b'secure data'
 
 
 @pytest.mark.skipif(
--- a/tests/hazmat/backends/test_openssl.py
+++ b/tests/hazmat/backends/test_openssl.py
@@ -29,7 +29,7 @@ from cryptography.hazmat.primitives.ciphers.algorithms import AES
 from cryptography.hazmat.primitives.ciphers.modes import CBC, CTR, Mode
 
 from ..primitives.fixtures_dsa import DSA_KEY_2048
-from ..primitives.fixtures_rsa import RSA_KEY_2048, RSA_KEY_512
+from ..primitives.fixtures_rsa import RSA_KEY_2048, RSA_KEY_512, RSA_KEY_512_ALT
 from ..primitives.test_ec import _skip_curve_unsupported
 from ...utils import load_vectors_from_file, raises_unsupported_algorithm
 
@@ -479,7 +479,7 @@ class TestOpenSSLRSA(object):
             )
         )
 
-        decrypted = private_key.decrypt(
+        decrypted = RSA_KEY_512_ALT.private_key(backend).decrypt(
             ciphertext,
             padding.OAEP(
                 mgf=padding.MGF1(algorithm=hashes.SHA1()),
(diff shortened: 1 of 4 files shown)


reaperhulk:
@EnTeQuAk thanks for the PR! In this case I think a good test is whatever you used to produce the problem.

EnTeQuAk:
@reaperhulk added simple test that would fail on decryption without the support.

reaperhulk:
Your patch appears to be designed to have it properly raise a `ValueError` on a decryption failure rather than an `AssertionError`, correct? The test you put up passes without any issues on current trunk without your proposed patch.

EnTeQuAk:
@reaperhulk that's weird, yeah, if decryption fails (which it does not in the test) a ``ValueError`` is expected instead of an ``AssertionError``. Let me check why it's not failing without the patch, I used different rsa-keys here for testing but thought it'd be all the same.

I'll re-check

reaperhulk:
@EnTeQuAk your test doesn't expect a failure at all (it just encrypts/decrypts and asserts that decryption was successful), so it isn't going to exercise the failure case (which would require a `pytest.raises(ValueError)`). Did you copy the wrong test?

EnTeQuAk:
Ah no, it's not expecting a failure. The patch fixes a wrongly raised ``AssertionError``.

reaperhulk:
I am still a bit confused I think. The patch adds a new error code to the list of expected codes when `_handle_rsa_enc_dec_error` is called. This is *only* called when a decryption/encryption has failed though, so the test case you've provided can't exercise that code path.

EnTeQuAk:
Thanks, I've been blind. Updated the test :smile:

reaperhulk:
Thanks to a new GH feature you'll need to merge the latest master into this to be fully up to date before we can merge as well (sorry!)

tests/hazmat/backends/test_openssl.py
```diff
@@ -467,6 +467,28 @@ def test_unsupported_oaep_label_decrypt(self):
...
+            padding.OAEP(
+                mgf=padding.MGF1(algorithm=hashes.SHA1()),
+                algorithm=hashes.SHA1(),
+                label=None
+            )
+        )
+
+        with pytest.raises(ValueError):
+            RSA_KEY_512_ALT.private_key(backend).decrypt(
+                ciphertext,
```
alex: Can you please move the `RSA_KEY_512_ALT.private_key(backend)` call outside of the `with pytest.raises(ValueError)` block.

reaperhulk:
Also this new test passes for me on master without the patch as well, which suggests this issue is specific to a particular OpenSSL. What OS/distro/OpenSSL version are you seeing this on?

EnTeQuAk:
@reaperhulk I'm using OpenSSL 1.0.2d 9 Jul 2015 on ArchLinux

tests/hazmat/backends/test_openssl.py
EnTeQuAk: Done.

reaperhulk:
Sigh, stupid OpenSSL. I'm using 1.0.2d for my tests as well, but can't replicate. Figures.

reaperhulk:
jenkins, ok to test

EnTeQuAk:
Not cool. I'll try to recheck some time tomorrow (CET) what might be going on here. :disappointed:

tests/hazmat/backends/test_openssl.py
```diff
@@ -29,7 +29,7 @@
 from cryptography.hazmat.primitives.ciphers.modes import CBC, CTR, Mode
 
 from ..primitives.fixtures_dsa import DSA_KEY_2048
-from ..primitives.fixtures_rsa import RSA_KEY_2048, RSA_KEY_512
+from ..primitives.fixtures_rsa import RSA_KEY_2048, RSA_KEY_512, RSA_KEY_512_ALT
```
alex: This line is too long, you'll need to format it:

```py
from ..primitives.fixtures_rsa import (
    RSA_KEY_2048, RSA_KEY_512, RSA_KEY_512_ALT
)
```

reaperhulk:
@EnTeQuAk have you had a chance to revisit this or would you like us to take it over? It looks like there's just one small PEP8 problem.

EnTeQuAk:
@reaperhulk Hey, sorry, I did not have the time recently to figure it out. I'll fix the pep8 problem shortly, I'm just curious about the fact that the test works for you on master without the patch but it doesn't for me. I need to test this on various machines. In any case, I'm still on it.

PR #2332, opened 2015-09-10
upgrade os x travis jobs to use latest image

reaperhulk:


alex:
Tests are failing, and I'm now confused, why do we enable pyenv before it's installed?

Issue #2299, opened 2015-08-27
Regression: Cannot use cyrptography in embedded wsgi


Naddiseo:
We have discovered at $work that the 1.0 cryptography release does not work with apache2+mod_wsgi. The error log spits out "Truncated or oversized response headers received from daemon process", which seems to happen if the wsgi process dies. We've tracked the issue down to an upgrade with cryptography, and something to do with threading and/or the Python GIL (a work around is in the wsgi config to set [WSGIApplicationGroup](https://code.google.com/p/modwsgi/wiki/ConfigurationDirectives#WSGIApplicationGroup) to `%{GLOBAL}`). 

I did a git bisect from 0.9.3 and 1.0, and found that the issue was introduced in commit b3d37a5d485bcd295d1933d638180e9cd5d23478.

reaperhulk:
Wow, would not have expected it to come from that...

/cc @glyph

Naddiseo:
Me either, the commit itself doesn't look like it could cause something like this, I think that it may have been an earlier commit that introduced code paths that weren't called until the triggering commit.

reaperhulk:
Out of curiosity... Calling `backend.activate_builtin_random()` should disable this code path (you'll need to have a backend available in that scope of course. `from cryptography.hazmat.backends.openssl.backend import backend` will do it). Does this resolve your issue?

Naddiseo:
Yes, those two lines also resolve the issue.

glyph:
OK, so here's a hypothesis:

This is a problem because sub-interpreters have an entirely different Python namespace (different modules, etc) but the same C namespace (same GIL, same C-level global variables, shared libraries, etc).  I'm not sure *exactly* why initializing the engine a second time causes a problem, but it's C-level runtime initialization stuff, so  ¯\_(ツ)_/¯.  I'm sure that cryptography being surprised to discover an already-initialized OpenSSL causes all kinds of exciting issues.

@reaperhulk - you should like this one since it would also explain some of the other concurrency issues you're seeing at import time.

(The main hypothesis I guess I'm describing here is that subinterpreters are just a completely broken concept nobody should use, but maybe cryptography could be made to work around these issues.)

reaperhulk:
If we work on that assumption we can make engine addition idempotent at the C layer (see #2293) but since the random engine has callbacks in Python (that potentially point to functions not in the "current" sub-interpreter) it would still fail in the sub-interpreter...correct?

glyph:
[...]

PR #2336, opened 2015-09-13
acquire a global lock to init the OpenSSL singleton locks

reaperhulk:
...then init the library since we really only want to do that once anyway

This approach was suggested in #2299. Of note, in a subinterpreter environment a "random" subinterpreter will register the static locks and engine. This means it will contain the Python callbacks for them. If the subinterpreter is subsequently destroyed, do the locks and random engine continue to function? If not, this solution will not work.

This should be considered WIP until we have a solid answer to the above.

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/binding.py
+++ b/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -182,6 +182,8 @@ class Binding(object):
 # although it's still unclear what happens when a subinterpreter which
 # has registered the locks is destroyed. Do the callback functions still work
 # or are all the objects associated with that subinterpreter destroyed?
+# imp.acquire_lock() is global even in 3.4+ while import locks have moved to
+# per module granularity.
 try:
     imp.acquire_lock()
     Binding.init_static_locks()
--- a/src/cryptography/hazmat/bindings/openssl/binding.py
+++ b/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -5,6 +5,7 @@
 from __future__ import absolute_import, division, print_function
 
 import collections
+import imp
 import os
 import threading
 import types
@@ -171,3 +172,17 @@ class Binding(object):
                     mode, n, file, line
                 )
             )
+
+
+# https://github.com/pyca/cryptography/issues/2299 contains extensive
+# discussion about this workaround. In essence, we are acquiring a global lock
+# (including across all subinterpreters if we're in a mod_wsgi environment)
+# initializing some C level locks, then initializing the binding and adding the
+# OS random engine. This should prevent engine addition race conditions,
+# although it's still unclear what happens when a subinterpreter which
+# has registered the locks is destroyed. Do the callback functions still work
+# or are all the objects associated with that subinterpreter destroyed?
+imp.acquire_lock()
+Binding.init_static_locks()
+Binding._ensure_ffi_initialized()
+imp.release_lock()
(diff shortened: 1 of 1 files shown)


src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -139,3 +140,17 @@ def _lock_cb(cls, mode, n, file, line):
...
+# (including across all subinterpreters if we're in a mod_wsgi environment)
+# initializing some C level locks, then initializing the binding and adding the
+# OS random engine. This should prevent engine addition race conditions,
+# although it's still unclear what happens when a subinterpreter which
+# has registered the locks is destroyed. Do the callback functions still work
+# or are all the objects associated with that subinterpreter destroyed?
+imp.acquire_lock()
+Binding.init_static_locks()
+Binding._ensure_ffi_initialized()
+imp.release_lock()
```
alex: This should definitely be inside of a `finally` block. I still need to think of any of this makes sense though :-)

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -139,3 +140,19 @@ def _lock_cb(cls, mode, n, file, line):
...
+
+# https://github.com/pyca/cryptography/issues/2299 contains extensive
+# discussion about this workaround. In essence, we are acquiring a global lock
+# (including across all subinterpreters if we're in a mod_wsgi environment)
+# initializing some C level locks, then initializing the binding and adding the
+# OS random engine. This should prevent engine addition race conditions,
+# although it's still unclear what happens when a subinterpreter which
+# has registered the locks is destroyed. Do the callback functions still work
+# or are all the objects associated with that subinterpreter destroyed?
+try:
```
alex: When is the import lock not held here? If we replaced this with `assert imp.lock_held()` what happens?

reaperhulk: @glyph can probably more clearly articulate this, but my understanding is that Python 3.4+ uses per module locks rather than the global import lock so while for Py < 3.4 we could just call `init_static_locks` and `_ensure_ffi_initialized` at module scope and be safe, 3.4+ would require us to explicitly acquire the global import lock.

dstufft: I guess that 3.4+'s ``imp.acquire_lock()`` is still global and isn't per module as well?

alex: Ah, ok, so in 3.4/3.5, these refer to the real global lock.

The comment should mention this.

On Sun, Sep 13, 2015 at 4:02 PM, Donald Stufft <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/2336#discussion_r39351421>:
>
> > @@ -139,3 +140,19 @@ def _lock_cb(cls, mode, n, file, line):
> >                      mode, n, file, line
> >                  )
> >              )
> > +
> > +
> > +# https://github.com/pyca/cryptography/issues/2299 contains extensive
> > +# discussion about this workaround. In essence, we are acquiring a global lock
> > +# (including across all subinterpreters if we're in a mod_wsgi environment)
> > +# initializing some C level locks, then initializing the binding and adding the
> > +# OS random engine. This should prevent engine addition race conditions,
> > +# although it's still unclear what happens when a subinterpreter which
> > +# has registered the locks is destroyed. Do the callback functions still work
> > +# or are all the objects associated with that subinterpreter destroyed?
> > +try:
>
> I guess that 3.4+'s imp.acquire_lock() is still global and isn't per
> module as well?
>
> —
> <https://github.com/pyca/cryptography/pull/2336/files#r39351421>.
>



-- 

reaperhulk:
Since this doesn't work I'm closing it. We'll do what's in #2446 probably but a complete fix is still elusive.

PR #2337, opened 2015-09-14
Python 3.5 support

reaperhulk:
Also testing to see what our pypy version is pre-upgrade now.


alex:
pytest 2.7.3 is out, can you remove the pypy bits from here so we can get this merged?

alex:
Oh, you also need to update `install.sh` to install python 3.5

alex:
Also should be added to `tox.ini`

PR #2343, opened 2015-09-15
Implement ANSI X9.63 KDF

simo5:
Add ANSI X9.63 KDf function as it is needed for Let's encrypt.
This function is just a minor variation of ConcatKDFHash, but different enough that we need a separate name/implementation.

### DIFF
--- a/docs/hazmat/primitives/key-derivation-functions.rst
+++ b/docs/hazmat/primitives/key-derivation-functions.rst
@@ -506,6 +506,99 @@ Different KDFs are suitable for different tasks such as:
         ``key_material`` generates the same key as the ``expected_key``, and
         raises an exception if they do not match.
 
+.. currentmodule:: cryptography.hazmat.primitives.kdf.x963kdf
+
+.. class:: X963KDF(algorithm, length, otherinfo, backend)
+
+    .. versionadded:: 1.1
+
+    X963KDF (ANSI X9.63 Key Derivation Function) is defined by ANSI
+    in the `ANSI X9.63:2001`_ document, to be used to derive keys for use
+    after a Key Exchange negotiation operation.
+
+    SECG in `SEC 1 v2.0`_ recommends that
+    :class:`~cryptography.hazmat.primitives.kdf.concatkdf.ConcatKDFHash` be
+    used for new projects. This KDF should only be used for backwards
+    compatibility with pre-existing implementations.
+
+
+    .. warning::
+
+        X963KDF should not be used for password storage.
+
+    .. doctest::
+
+        >>> import os
+        >>> from cryptography.hazmat.primitives import hashes
+        >>> from cryptography.hazmat.primitives.kdf.x963kdf import X963KDF
+        >>> from cryptography.hazmat.backends import default_backend
+        >>> backend = default_backend()
+        >>> sharedinfo = b"ANSI X9.63 Example"
+        >>> xkdf = X963KDF(
+        ...     algorithm=hashes.SHA256(),
+        ...     length=256,
+        ...     sharedinfo=sharedinfo,
+        ...     backend=backend
+        ... )
+        >>> key = xkdf.derive(b"input key")
+        >>> xkdf = X963KDF(
+        ...     algorithm=hashes.SHA256(),
+        ...     length=256,
+        ...     sharedinfo=sharedinfo,
+        ...     backend=backend
+        ... )
+        >>> xkdf.verify(b"input key", key)
+
+    :param algorithm: An instance of a
+        :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm`
+        provider
+
+    :param int length: The desired length of the derived key in bytes.
+        Maximum is ``hashlen * (2^32 -1)``.
+
+    :param bytes sharedinfo: Application specific context information.
+        If ``None`` is explicitly passed an empty byte string will be used.
+
+    :param backend: A cryptography backend
+        :class:`~cryptography.hazmat.backends.interfaces.HashBackend`
+        provider.
+
+    :raises cryptography.exceptions.UnsupportedAlgorithm: This is raised
+        if the provided ``backend`` does not implement
+        :class:`~cryptography.hazmat.backends.interfaces.HashBackend`
+
+    :raises TypeError: This exception is raised if ``sharedinfo`` is not
+        ``bytes``.
+
+    .. method:: derive(key_material)
+
+        :param bytes key_material: The input key material.
+        :return bytes: The derived key.
+        :raises TypeError: This exception is raised if ``key_material`` is
+                            not ``bytes``.
+
+        Derives a new key from the input key material.
+
+    .. method:: verify(key_material, expected_key)
+
+        :param bytes key_material: The input key material. This is the same as
+                                   ``key_material`` in :meth:`derive`.
+        :param bytes expected_key: The expected result of deriving a new key,
+                                   this is the same as the return value of
+                                   :meth:`derive`.
+        :raises cryptography.exceptions.InvalidKey: This is raised when the
+                                                    derived key does not match
+                                                    the expected key.
+        :raises cryptography.exceptions.AlreadyFinalized: This is raised when
+                                                          :meth:`derive` or
+                                                          :meth:`verify` is
+                                                          called more than
+                                                          once.
+
+        This checks whether deriving a new key from the supplied
+        ``key_material`` generates the same key as the ``expected_key``, and
+        raises an exception if they do not match.
+
 
 Interface
 ~~~~~~~~~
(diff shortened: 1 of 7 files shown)


alex:
This can be rebased now

simo5:
Has already been rebased

simo5:
ok stupid git and its merge commits, rebased again

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -506,6 +506,93 @@ Different KDFs are suitable for different tasks such as:
...
 
+.. currentmodule:: cryptography.hazmat.primitives.kdf.x963kdf
+
+.. class:: X963KDF(algorithm, length, otherinfo, backend)
+
+    .. versionadded:: 1.0
+
+    X963KDF (ANSI X9.63 Key Derivation Function) is defined by ANSI
+    X9.63-2001 `ANSI X9.63:2001`_ document, to be used to
+    derive keys for use after a Key Exchange negotiation operation.
```
alex: Should we mention that this is derpcated by ANSI and should only be used by existing applications?

src/cryptography/hazmat/primitives/kdf/x963kdf.py
```diff
@@ -0,0 +1,70 @@
...
+        self._used = True
+
+        if not isinstance(key_material, bytes):
+            raise TypeError("key_material must be bytes.")
+
+        output = [b""]
+        outlen = 0
+        counter = 1
+
+        while (self._length > outlen):
```
alex: Parens can be dropped here.

alex:
This seems to be missing tests which use the vectors?

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -556,6 +643,7 @@ Interface
 
 .. _`NIST SP 800-132`: http://csrc.nist.gov/publications/nistpubs/800-132/nist-sp800-132.pdf
 .. _`NIST SP 800-56Ar2`: http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar2.pdf
+.. _`ANSI X9.63:2001`: http://webstore.ansi.org
```
reaperhulk: Let's make this link https

docs/hazmat/primitives/key-derivation-functions.rst
simo5: On Fri, 2015-09-18 at 17:23 -0700, Alex Gaynor wrote:
> Should we mention that this is derpcated by ANSI and should only be
> used by existing applications?

I do not have access to X9.63:2011 where is it said this is deprecated ?

alex: I could have sworn in looking into this we discovered that it was deprecated? /cc @reaperhulk

reaperhulk: http://www.secg.org/sec1-v2.pdf says "The NIST-800-56-Catenation-KDF should be used, except for backwards compatability with im- plementations already using one of the three other key derivation functions." (where one of the "three other" is this KDF)

simo5: It doesn't read to me as a deprecation notice, just a recommendation.
In fact let's encrypt is starting to use it now ...

reaperhulk: Since they've never provided context for their choice I'm still unconvinced that ConcatKDF wouldn't have been a better solution. I think it's reasonable to state that SECG's own documents suggest not using this except for compatibility purposes.

simo5: Can you propose text I would include ?

reaperhulk: Sure, how about "SEC 1 v2.0(link) recommends that ConcatKDF(link to our docs for ConcatKDF) be used for new projects. This KDF should only be used for backwards compatibility with pre-existing implementations."

simo5:
Should have addressed all concerns with this 2 commits, hopefully the CI will agree :)

tests/utils.py
```diff
@@ -760,3 +761,50 @@ def load_kasvs_ecdh_vectors(vector_data):
...
+            assert math.ceil(shared_secret_len / 8) * 2 == len(vector[3])
+        elif line.startswith("SharedInfo"):
+            if shared_info_len != 0:
+                vector[5] = line.split("=")[1].strip()
+                assert math.ceil(shared_info_len / 8) * 2 == len(vector[5])
+        elif line.startswith("key_data"):
+            vector[7] = line.split("=")[1].strip()
+            assert math.ceil(key_data_len / 8) * 2 == len(vector[7])
+            vectors.append(vector)
+            vector = ["", 0, 0, None, 0, None, 0, None]
```
reaperhulk: I find the usage of a list here very confusing when trying to read the tests. Could we switch it to a dict?

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -506,6 +506,99 @@ Different KDFs are suitable for different tasks such as:
         ``key_material`` generates the same key as the ``expected_key``, and
         raises an exception if they do not match.
 
+.. currentmodule:: cryptography.hazmat.primitives.kdf.x963kdf
+
+.. class:: X963KDF(algorithm, length, otherinfo, backend)
+
+    .. versionadded:: 1.0
```
reaperhulk: This needs to be 1.1

tests/hazmat/primitives/test_x963kdf.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+
+        okm = binascii.unhexlify(b"443024c3dae66b95e6f5670601558f71")
+
+        ckdf = X963KDF(hashes.SHA256(), 16, None, backend)
+
+        assert ckdf.derive(prk) == okm
+
+    def test_verify(self, backend):
+        prk = binascii.unhexlify(
```
reaperhulk: is prk, okm, etc the right names for these variables or is that an artifact from concatkdf?

tests/utils.py
simo5: I can change it to a dictionary.

tests/hazmat/primitives/test_x963kdf.py
simo5: It's the same definitions as used in ConcatKDF's CAVS files I guess.
I can change them to use z and key_data if you prefer

reaperhulk: It's not a big deal, but if you're updating for the other comments I'd appreciate it. Makes it a bit easier to follow what's happening when reviewing.

simo5:
I've updated tests and docs as requested

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -506,6 +506,99 @@ Different KDFs are suitable for different tasks such as:
...
+    .. versionadded:: 1.1
+
+    X963KDF (ANSI X9.63 Key Derivation Function) is defined by ANSI
+    in the `ANSI X9.63:2001`_ document, to be used to derive keys for use
+    after a Key Exchange negotiation operation.
+
+    SECG in `SEC 1 v2.0`_ recommends that
+    :class:`~cryptography.hazmat.primitives.kdf.concatkdf.ConcatKDFHash` be
+    used for new projects. This KDF should only be used for backwards
+    compatibility with pre-existing implementations.
```
alex: I'd `s/implementations/protocols/` here.

tests/utils.py
```diff
@@ -760,3 +761,51 @@ def load_kasvs_ecdh_vectors(vector_data):
...
+def load_x963_vectors(vector_data):
+    """
+    Loads data out of the X9.63 vector data
+    """
+
+    vectors = []
+
+    # Sets Metadata
+    hashname = None
+    vector = dict()
```
alex: Please use a dict literal here: `{}`

tests/utils.py
```diff
@@ -760,3 +761,51 @@ def load_kasvs_ecdh_vectors(vector_data):
...
+        elif line.startswith("SharedInfo"):
+            if shared_info_len != 0:
+                vector["SharedInfo"] = line.split("=")[1].strip()
+                silen = len(vector["SharedInfo"])
+                assert math.ceil(shared_info_len / 8) * 2 == silen
+        elif line.startswith("key_data"):
+            vector["key_data"] = line.split("=")[1].strip()
+            assert math.ceil(key_data_len / 8) * 2 == len(vector["key_data"])
+            vectors.append(vector)
+            vector = dict()
```
alex: `{}` here too

tests/utils.py
```diff
@@ -760,3 +761,51 @@ def load_kasvs_ecdh_vectors(vector_data):
...
+        elif line.startswith("[SharedInfo length"):
+            shared_info_len = int(line[1:-1].split("=")[1].strip())
+        elif line.startswith("[key data length"):
+            key_data_len = int(line[1:-1].split("=")[1].strip())
+        elif line.startswith("COUNT"):
+            count = int(line.split("=")[1].strip())
+            vector["hash"] = hashname
+            vector["count"] = count
+            vector["shared secret length"] = shared_secret_len
+            vector["SharedInfo length"] = shared_info_len
```
alex: Is there a reason not to use more "typical" names here: e.g. `"sharedinfo_length"`

reaperhulk:
Thanks @simo5, we're going to go ahead and merge this and @alex will put in a follow up PR to clean up the comments he had.

PR #2345, opened 2015-09-15
workaround for pytest 2.7.3's expectation of USERNAME on windows

reaperhulk:
We should remove this when a pytest that resolves https://github.com/pytest-dev/pytest/issues/1010 is released


alex:
@reaperhulk an alternate outcome might be a tox release that just adds `USERNAME ` to the default `passenv`

PR #2353, opened 2015-09-19
Adjusted the README to reflect that only PyPy 2.6 and above is supported

thedrow:
Since version 1.0 cryptography only installs on PyPy 2.6 and above. I changed the readme file to reflect that.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,7 +18,7 @@ Cryptography
 
 ``cryptography`` is a package which provides cryptographic recipes and
 primitives to Python developers.  Our goal is for it to be your "cryptographic
-standard library". It supports Python 2.6-2.7, Python 3.3+, and PyPy.
+standard library". It supports Python 2.6-2.7, Python 3.3+, and PyPy 2.6+. For support of PyPy versions below 2.6 use cryptography 0.9.x.
 
 ``cryptography`` includes both high level recipes, and low level interfaces to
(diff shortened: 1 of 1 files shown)


Ayrx:
I don't think we want to recommend older versions of cryptography.

thedrow:
What if one does need to support older PyPy versions because that's what currently deployed at his/hers production site?

reaperhulk:
Explicitly recommending older versions is something we'd prefer to avoid (as we are not supporting those versions), but I'd be okay with phrasing like "Support for PyPy < 2.6 was dropped after 0.9.3" (I realize this is just a semantic argument. Others may disagree entirely :smile:)

alex:
I agree with @reaperhulk, we made sure the CHANGELOG language also
encouraged people to move forward.

On Sat, Sep 19, 2015 at 11:09 AM, Paul Kehrer <notifications@github.com>
wrote:

> Explicitly recommending older versions is something we'd prefer to avoid
> (as we are not supporting those versions), but I'd be okay with phrasing
> like "Support for PyPy < 2.6 was dropped after 0.9.3" (I realize this is
> just a semantic argument. Others may disagree entirely [image: :smile:])
>
> —
> <https://github.com/pyca/cryptography/pull/2353#issuecomment-141678648>.
>



-- 

PR #2363, opened 2015-09-24
should have _asn1_* utility functions in a common place

etrauschke:
I think the utility functions to convert asn1 values should not be in x509.py but a more common place because they might be useful in other places too.
There are not that many right  now so backend.py might be good place. If there will be more in the future I wonder if it would be better to have a dedicated asn1_utils module for them.


reaperhulk:
I am fine with moving them here for now. I agree we should probably split them in the future though. This PR is splitting out the content from the X509 CRL PR correct?

etrauschke:
Yes, that is the intent. Once this is in, I will update the CRL pull request.

PR #2366, opened 2015-09-25
add some el capitan to our travis build matrix

reaperhulk:
I think we may want to revisit what our OS X build matrix looks like to try to trim it down if we can, but for the moment let's add system Python from El Capitan (10.11).


.travis.yml
```diff
@@ -134,6 +134,14 @@ matrix:
           os: osx
           osx_image: xcode7
           env: TOXENV=docs
+        - language: generic
```
alex: This is nit picky, but can you put these above `docs`

PR #2370, opened 2015-09-25
convert the rest of the openssl backend to using openssl_assert

reaperhulk:


src/cryptography/hazmat/backends/openssl/rsa.py
```diff
@@ -174,7 +174,7 @@ def __init__(self, backend, private_key, padding, algorithm):
 
             # Size of key in bytes - 2 is the maximum
             # PSS signature length (salt length is checked later)
-            assert self._pkey_size > 0
+            self._backend.openssl_assert(self._pkey_size > 0)
```
alex: This is an internal invariant

src/cryptography/hazmat/backends/openssl/rsa.py
```diff
@@ -363,7 +365,7 @@ def __init__(self, backend, public_key, signature, padding, algorithm):
...
```
alex: invariant

PR #2371, opened 2015-09-25
Backport first assert PR

reaperhulk:


alex:
@reaperhulk what's the deal with the coverage regression here?

reaperhulk:
Oops, this is going to be the same coverage 4.0 thing that got us in master. We'll need to backport those fixes too.

PR #2375, opened 2015-09-26
Convert asserts in bindings as well

alex:


src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -12,6 +13,42 @@
 from cryptography.hazmat.bindings.openssl._conditional import CONDITIONAL_NAMES
 
 
+_OpenSSLError = collections.namedtuple("_OpenSSLError",
+                                       ["code", "lib", "func", "reason"])
+
+
+class UnhandledOpenSSLError(Exception):
```
reaperhulk: Do we want to put this here and then import it from the backend? Two identical declarations is a bit odd. We're not documenting it but we also aren't calling this private.

alex: Err, it's not copied? I moved it.

reaperhulk: What in the hell. I swear this diff looks different now (but obviously I'm wrong). Ignore me.

PR #2360, opened 2015-09-21
SubjectKeyIdentifier equality now uses constant-time digest comparison.

obi1kenobi:
This hardens SubjectKeyIdentifier against potential timing attacks, despite the fact that there is no immediately obvious advantage an adversary could gain by executing such an attack.

PR #2384, opened 2015-09-29
Made AuthorityKeyIdentifier comparisons more resilient against timing.

obi1kenobi:
Similar to #2360, hardens the AuthorityKeyIdentifier equality check against timing attacks, despite there being no obvious exploit that leverages this vector.

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -156,8 +156,17 @@ class AuthorityKeyIdentifier(object):
         if not isinstance(other, AuthorityKeyIdentifier):
             return NotImplemented
 
+        if self.key_identifier is None:
+            key_ident_match = other.key_identifier is None
+        else:
+            key_ident_match = (
+                other.key_identifier is not None and
+                constant_time.bytes_eq(self.key_identifier,
+                                       other.key_identifier)
+            )
+
         return (
-            self.key_identifier == other.key_identifier and
+            key_ident_match and
             self.authority_cert_issuer == other.authority_cert_issuer and
             self.authority_cert_serial_number ==
             other.authority_cert_serial_number


obi1kenobi:
Keep in mind that the `authority_cert_issuer` equality check is almost certainly not constant time, and there is currently no facility in Python to compare lists against each other in constant time, and no function in your library that does the same. I would not mind writing a PR to do that, but I wanted to check with you first since it's a non-trivial amount of work.

Ayrx:
So, one concern I have with this PR (and #2360) is if there's a appreciable performance hit going from `==` to constant time comparisons?

If there is no useful information being leaked with a non-constant time compare (just like how comparing hashes in non-constant time doesn't leak any useful info), are we just doing security theatre by making everything constant time?

obi1kenobi:
I understand your concern, and I myself gave this some thought before deciding to submit the PR. I came up with three things.

* The common case of the comparison (I believe) is that all checks succeed, because this code is simply ensuring the validity of the x509 extension. This means that, constant-time or not, the code has to compare all fields, and that means going through the entire string and comparing it against the entire other string etc. This is also exactly how a constant-time comparison behaves when everything matches -- except it also behaves like this when something doesn't match. However, since that case is uncommon, any performance hit will be visible a negligible fraction of the time.

* All the targets that CI tests against, with the exception of Python 2.6, support `hmac.compare_digest` -- the built-in constant-time comparison written in C. This means that most users of this library will use an extremely efficient implementation. Keep in mind that this also skips all of Python's complicated method resolution for operators, because there is no need to decide whose method to call -- there's only one version of `hmac.compare_digest` and that's the one that is called.

* Lastly, just because I or anyone else here are unable to exploit this at the moment, does not mean that one cannot be found in general. Attacks only get *more* sophisticated, and never *less*, and even if there is an appreciable performance hit (which I believe there isn't), protecting the few users who may be subject to such an advanced attack is worth waiting a fraction of a second longer. After all, how many different certs (not counting retries) does a user to try to validate per second, realistically? If the answer is more than ~10 -- that user is probably already not using Python, but C, or at least a statically-typed language.

reaperhulk:
I'm torn here mostly because this is a huge increase in complexity for the comparison (unlike the simple change for SKI)

obi1kenobi:
I agree, there is a nontrivial increase in code complexity. However, all of that complexity has full coverage in tests, so the danger of bugs or unpredictable behavior is minimal.

I'm also a firm believer in "better safe than sorry" when it comes to security, especially given recent events like Heartbleed and Stagefright in mind.

alex:
I think we need to articulate a scenario where there's a possibility of
information leakage before preceeding any further.

On Fri, Oct 2, 2015 at 11:06 AM, Predrag Gruevski <notifications@github.com>
wrote:

> I agree, there is a nontrivial increase in code complexity. However, all
> of that complexity has full coverage in tests, so the danger of bugs or
> unpredictable behavior is minimal.
>
> I'm also a firm believer in "better safe than sorry" when it comes to
> security, especially given recent events like Heartbleed and Stagefright in
> mind.
>
> —
> <https://github.com/pyca/cryptography/pull/2384#issuecomment-145053978>.
>



-- 

Ayrx:
@obi1kenobi The situations aren't comparable. You are talking about *very* straightforward memory corruption type bugs vs a side channel attack that may not exist.

obi1kenobi:
@Ayrx I was referring to the impact of the vulnerabilities, not the exploitation details. Cryptography's stated goal is to be the "cryptographic standard library," and it is extremely popular, making a potential exploit devastating.

obi1kenobi:
Here is another idea on how we could proceed. Most of the complexity is introduced by attempting to prevent an adversary from finding out which part of the AKI (cert issuer, serial number or key identifier) was a mismatch. However, there are comparatively few cert issuers, and serial numbers are presumably *serial* so likely guessable by enumeration.

If we limited the scope to only verifying the key identifier in constant time, most of the complexity would go away -- it would likely be a 3-line diff rather than a 25-line one. I think this would give us effectively all of the benefits at very little added complexity cost.

alex:
Serial numbers aren't serial. They're required to be random data in all
modern certs.

On Fri, Oct 2, 2015 at 11:19 AM, Predrag Gruevski <notifications@github.com>
wrote:

> Here is another idea on how we could proceed. Most of the complexity is
> introduced by attempting to prevent an adversary from finding out which
> part of the AKI (cert issuer, serial number or key identifier) was a
> mismatch. However, there are comparatively few cert issuers, and serial
> numbers are presumably *serial* so likely guessable by enumeration.
>
> If we limited the scope to only verifying the key identifier in constant
> time, most of the complexity would go away -- it would likely be a 3-line
> diff rather than a 25-line one. I think this would give us effectively all
> of the benefits at very little added complexity cost.
>
> —
> <https://github.com/pyca/cryptography/pull/2384#issuecomment-145059234>.
>



-- 

obi1kenobi:
That makes sense, sorry -- I wasn't aware of that development. That said, am I correct in assuming that the serial number field is an integer?

I don't like the idea of brainstorming potential attack vectors before changing the code, because there is no possible way for us to know we haven't missed something. Instead, I think that I should be able to produce a more compact diff that achieves effectively the same thing with less complexity. Would that be acceptable?

src/cryptography/x509/extensions.py
```diff
@@ -156,8 +156,17 @@ def __eq__(self, other):
         if not isinstance(other, AuthorityKeyIdentifier):
             return NotImplemented
 
+        if self.key_identifier is None:
```
obi1kenobi: The main problem here is that `constant_time.bytes_eq()` raises exceptions if either argument is None. If you have any ideas on how to avoid the branches without introducing more complexity, I'd love to know.

obi1kenobi:
Any thoughts on the updated diff?

reaperhulk:
I'm a terrible person, but I think I need to see a threat model that this protects against before I'm willing to :+1: it. Sorry :disappointed: 

If other PyCA members disagree please speak up!

reaperhulk:
@alex what do you think here? Looking at it again, the current patch isn't a huge increase in complexity, so I'm less bothered than I was originally, so I guess I could say I'm neutral now.

reaperhulk:
After a bit more discussion we've decided that in the absence of even a hypothetical attack scenario (not a PoC, just an articulation of the risk) we're going to close this for now. We can reopen if someone is able to describe such a scenario.

glyph:
@obi1kenobi -

> I'm also a firm believer in "better safe than sorry" when it comes to security, especially given recent events like Heartbleed and Stagefright in mind.

I realize I'm coming to this after the discussion has already concluded, but since nobody called this out specifically, I wanted to add another point of agreement with the conclusion, because I think this is actually part of Cryptography's development philosophy which may not have been specifically stated:

Heartbleed, in particular, was the result of the increased complexity cost of additional security features.  DTLS is a useful thing, that you actually need to improve the security of datagram transports!  So if you want to avoid bugs like this, the most important thing is to mercilessly cut complexity at every point, and reject any “security feature” for which there is no coherent threat model.  Every additional line of code carries additional risk that it might be mistakenly implemented and that mistake exploited; the payoff of that line of code has to be justified by some clearly articulated threat.  After all, there are *tons* of input-dependent branches in Cryptography; it's not really feasible to make *everything* constant-time.

That's not to say that you have to demonstrate a *working exploit* in order to add some mitigation code, but you have to at least be able to explain how such a thing could possibly be constructed.  In this case I can't see how AuthorityKeyIdentifier comparisons could result in leaking any information about the underlying key material, and it doesn't sound like you can either.  (Standard Disclaimer: I Am Not A Cryptographer, I Don't Even Play One On The Internet)

PR #2387, opened 2015-10-01
add support for signature & tbs_certificate to Certificate

reaperhulk:
To actually verify that these signatures are correct we also need access to the `tbsCertificate` bytes. Then verification looks like:

```python
        # RSA specific
        verifier = signer_public_key.verifier(
            cert.signature, padding.PKCS1v15(), cert.signature_hash_algorithm
        )
        verifier.update(tbs)
        verifier.verify()
```

Without tbs you can't actually verify. If we're going down this path then we need to expose tbsCertificate as well (name suggestions?).

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -382,6 +382,26 @@ X.509 Certificate Object
             <Extension(oid=<ObjectIdentifier(oid=2.5.29.32, name=certificatePolicies)>, critical=False, value=<CertificatePolicies([<PolicyInformation(policy_identifier=<ObjectIdentifier(oid=2.16.840.1.101.3.2.1.48.1, name=Unknown OID)>, policy_qualifiers=None)>])>)>
             <Extension(oid=<ObjectIdentifier(oid=2.5.29.19, name=basicConstraints)>, critical=True, value=<BasicConstraints(ca=True, path_length=None)>)>
 
+    .. attribute:: signature
+
+        .. versionadded:: 1.2
+
+        :type: bytes
+
+        The bytes of the certificate's signature.
+
+    .. attribute:: tbs_certificate
+
+        .. versionadded:: 1.2
+
+        :type: bytes
+
+        The DER encoded bytes payload (as defined by :rfc:`5280`) that is hashed
+        and then signed by the private key of the certificate's issuer. This
+        data may be used to validate a signature, but use extreme caution as
+        certificate validation is a complex problem that involves much more
+        than just signature checks.
+
     .. method:: public_bytes(encoding)
 
         .. versionadded:: 1.0
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -390,7 +390,7 @@ X.509 Certificate Object
 
         The bytes of the certificate's signature.
 
-    .. attribute:: tbs_certificate
+    .. attribute:: tbs_certificate_bytes
 
         .. versionadded:: 1.2
 
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -327,8 +327,9 @@ class _Certificate(object):
         return self._backend._asn1_string_to_bytes(self._x509.signature)
 
     @property
-    def tbs_certificate(self):
+    def tbs_certificate_bytes(self):
         pp = self._backend._ffi.new("unsigned char **")
+        # the X509_CINF struct holds the tbsCertificate data
         res = self._backend._lib.i2d_X509_CINF(self._x509.cert_info, pp)
         self._backend.openssl_assert(res > 0)
         pp = self._backend._ffi.gc(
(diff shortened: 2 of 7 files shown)


danetrain:
We already have a ```public_bytes()``` method, so maybe the described method should be called ```tbs_bytes()```?

docs/x509/reference.rst
```diff
@@ -322,6 +322,14 @@ X.509 Certificate Object
             <Extension(oid=<ObjectIdentifier(oid=2.5.29.32, name=certificatePolicies)>, critical=False, value=<CertificatePolicies([<PolicyInformation(policy_identifier=<ObjectIdentifier(oid=2.16.840.1.101.3.2.1.48.1, name=Unknown OID)>, policy_qualifiers=None)>])>)>
             <Extension(oid=<ObjectIdentifier(oid=2.5.29.19, name=basicConstraints)>, critical=True, value=<BasicConstraints(ca=True, path_length=None)>)>
 
+    .. attribute:: signature
+
+        .. versionadded:: 1.1
+
+        :type: bytes
+
+        The bytes of the certificate's signature.
```
alex: Does this need some words to express that you probably don't want to use this?

etrauschke:
I'm happy you are planning to provide an interface to do the basic public_key verification but why does this have to be that complicated while having an interface to OpenSSL's X509_verify() does essentially the same thing in one call.

In general I could see the need to have several parameters you can pass to signer_public_key.verifier() but I fail to see when you would ever want to verify against a non-matching signature algorithm or padding scheme. Or is there some restriction in X509_verify() I'm missing here?

If I understand this correctly, the proposed solution has the same limitations X509_verify() has but is at least ten times harder to use (I think I would always have to copy&paste an example of this to get it right).

reaperhulk:
@alex and others have expressed a preference for Certificate to remain a "data only" object in the recent past. The intent of exposing `signature` and `tbs_certificate` would not be for people to build their own verification, but rather to support the verification classes being proposed in #2381.

I am sympathetic to the argument that `cert.verify(pub_key)` might be a nicer API, especially given that it allows utilization of backend specific methods like `X509_verify`. I'm also not a big fan of exposing `tbs_certificate`, as its only use is to hash the ASN.1 bytes and do sign/verify operations, which should be handled by the library anyway. That said, I'm not sure I feel strongly enough to argue about this just yet. I think we need a complete API proposal.

etrauschke:
If you want to keep the certificate data-only, you could still add some sort of utility function in the backend which takes a cert object and a public key and returns a verification result. This works as long as this function can get access to the underlying x509 object (which is the case with the current design).
It would still offer a straight-forward interface and wouldn't violate your greater design principles. 

However, it would be pretty hard to find for users since I think that functions in the backend are not supposed to be consumed by a random user. This could also be a good thing though because it prevents people from shooting themselves in the foot accidentally, believing that this is an universally, all-encompassing verification of a given certificate.

alex:
a) Merge conflict :-(
b) So in principle this seems reasonable, in practice it's way more "insidery" feeling than the rest of our API. Do you have suggestions on how to assauge that feeling?

reaperhulk:
Yeah documenting these in my head `tbs_certificate` is almost footgun-like. "tbs_certificate is the DER bytes payload that is hashed and then signed by the private key of the certificate's issuer. This data may be used to validate a signature, but use extreme caution as certificate validation is a complex problem that involves much more than just signature checks. Users are encouraged to use <thing that doesn't exist yet>."

reaperhulk:
Having thought about that a bit I'm actually more okay with exposing it with text similar to that. We can update it with the "Users are encouraged to use validation class" once we build the actual class.

docs/x509/reference.rst
```diff
@@ -382,6 +382,26 @@ X.509 Certificate Object
...
 
+    .. attribute:: signature
+
+        .. versionadded:: 1.2
+
+        :type: bytes
+
+        The bytes of the certificate's signature.
+
+    .. attribute:: tbs_certificate
```
alex: Is this the right way to expose this data? It feels a bit ad-hoc compared to `public_bytes` or something?

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -322,6 +322,20 @@ def signature_hash_algorithm(self):
...
         return _CERTIFICATE_EXTENSION_PARSER.parse(self._backend, self._x509)
 
+    @property
+    def signature(self):
+        return self._backend._asn1_string_to_bytes(self._x509.signature)
+
+    @property
+    def tbs_certificate(self):
+        pp = self._backend._ffi.new("unsigned char **")
+        res = self._backend._lib.i2d_X509_CINF(self._x509.cert_info, pp)
```
alex: Can you add a comment that says that X509_CINF holds all the tbs data

docs/x509/reference.rst
reaperhulk: I'm not sure what the best approach is here. tbsCertificate is a subset of the ASN1 that makes up the complete certificate object. I don't think I'd like it in `public_bytes` because it would be beyond confusing to require users to pass the serialization type (when 99.99% of the time you just want the whole thing) in addition to encoding.

rowland-smith:
Blessed are the code-mergers...

PR #2388, opened 2015-10-02
Doc update for test matrix + OS X install

reaperhulk:
Our information about what we test against was out of date. Here's the latest + a simplification of the OS X install docs now that we upload binary wheels for cffi (**no compiler required** :sparkles:)


alex:
doc tests failing on spelling

PR #2395, opened 2015-10-07
Document building static wheels on Linux.

### LINKED ISSUE #2338: Instructions for Static Wheels on Linux
It'd be great if we could deploy static wheels on Linux to allow python users to get the most up-to-date OpenSSL without affecting the rest of their distribution. However, as @reaperhulk and I found, this isn't totally trivial to do.

Obviously we can't generally distribute static wheels, Linux is painful like that, but we can definitely publish instructions for how to build and use them yourself. If we can work out how to do it, I'm happy to write those up.

Lukasa:
Resolves #2338.

This is a skeleton example of documenting building static wheels for Linux. I'm not wedded to any of this, the tone, the placement, or anything else, but I think it should be documented somewhere. Feedback encouraged!

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -58,6 +58,8 @@ to include the proper locations. For example:
     C:\> pip install cryptography
 
 
+.. _build-on-linux:
+
 Building cryptography on Linux
 ------------------------------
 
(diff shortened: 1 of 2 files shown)


docs/installation.rst
```diff
@@ -115,6 +117,58 @@ You'll also need to generate your own ``openssl.ld`` file. For example::
...
+    pip install -U wheel pip
+    curl -O https://openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz
+    tar xvf openssl-${OPENSSL_VERSION}.tar.gz
+    cd openssl-${OPENSSL_VERSION}
+    ./config no-shared no-ssl2 -fPIC --prefix=${CWD}/openssl
+    make && make install
+    cd ..
+    CFLAGS="-I${CWD}/openssl/include" LDFLAGS="-L${CWD}/openssl/lib" pip wheel cryptography
+
+.. _virtualenv: https://virtualenv.pypa.io/en/latest/
```
reaperhulk: Our convention is to put links at the very bottom of the doc

docs/installation.rst
```diff
@@ -115,6 +117,58 @@ You'll also need to generate your own ``openssl.ld`` file. For example::
...
+recent OpenSSL available to your Python programs.
+
+To do so, you should find yourself a machine that is as similar as possible to
+your target environment (e.g. your production environment): for example, spin
+up a new cloud server running your target Linux distribution. On this machine,
+install the Cryptography dependencies as mentioned in :ref:`build-on-linux`.
+Please also make sure you have `virtualenv`_ installed: this should be
+available from your system package manager.
+
+Then, run the following shell script. Feel free to adjust the OpenSSL version
```
reaperhulk: I'd prefer this to say something about going to openssl.org to find the latest version.

docs/installation.rst
```diff
@@ -115,6 +117,58 @@ You'll also need to generate your own ``openssl.ld`` file. For example::
...
+sufficiently-recent version of ``pip``. The Cryptography wheel in this
+directory contains a statically-linked OpenSSL binding, which ensures that you
+have access to the most-recent OpenSSL releases without corrupting your system
+dependencies.
+
+.. code-block:: console
+
+    set -e
+
+    OPENSSL_VERSION="1.0.2d"
```
reaperhulk: I don't want to have our docs suggesting an out of date OpenSSL if at all possible. Would it be better to put a comment above this that describes the format expected and then make the variable "VERSIONGOESHERE"? It would break "just paste this script" but prevent installing out of date OpenSSLs...

docs/installation.rst
Lukasa: Done.

docs/installation.rst
Lukasa: Done.

docs/installation.rst
Lukasa: Done.

reaperhulk:
:+1: looks good to me. I'll merge when tests pass. Thanks for the help @Lukasa

Lukasa:
Always a pleasure. =)

PR #2398, opened 2015-10-08
ECDH Exchange implementation

simo5:


reaperhulk:
This can be rebased now that the key exchange interface is merged (sure would be nice if github could be smart enough to remove it from the diff automatically...)

tests/utils.py
```diff
@@ -722,7 +722,7 @@ def load_kasvs_ecdh_vectors(vector_data):
         if line.startswith("["):
             tag = line.split()[0][1:]
         elif line.startswith("COUNT = "):
-            data["COUNT"] = int(line.split("=")[1], 16)
+            data["COUNT"] = int(line.split("=")[1])
```
alex: Can we write a unit test for the loader specifically?

simo5: You mean this: https://github.com/pyca/cryptography/blob/master/tests/test_utils.py#L3053 ?

docs/hazmat/primitives/asymmetric/ec.rst
```diff
@@ -122,6 +122,43 @@ Elliptic Curve Signature Algorithms
         :returns: A new instance of a :class:`EllipticCurvePublicKey`
             provider.
 
+Elliptic Curve Signature Algorithms
```
alex: This isn't a signature algorithm

tests/hazmat/primitives/test_ec.py
```diff
@@ -720,3 +726,88 @@ def test_signature_not_bytes(self, backend):
...
+                    vector['IUT']['y'],
+                    curve_type()
+                )
+            ).private_key(backend)
+
+            pubkey = peerkey.public_key()
+        except ValueError:
+            if vector['fail'] and vector['errno'] != 8:
+                return
+        except exceptions.InternalError:
```
alex: This is a bug in cryptography, and should not be worked around in the tests.

docs/hazmat/primitives/asymmetric/ec.rst
simo5: Oups, Cut&Paste error, thanks for spotting it

tests/hazmat/primitives/test_ec.py
simo5: Ok, so what should we do ?

simo5:
Ok I added a patch that deals with properly raising a value error if X or Y coordinates are invalid.
HTH,
Simo.

tests/hazmat/primitives/test_ec.py
```diff
@@ -720,3 +726,82 @@ def test_signature_not_bytes(self, backend):
...
+            key.exchange(DummyExchangeAlg())
+
+    def test_unsupported_key_derivation_function(self, backend):
+        key = ec.generate_private_key(ec.SECP384R1(), backend)
+        with pytest.raises(TypeError):
+            key.exchange(ec.ECDH(DummyExchangeAlg()))
+
+    def test_invalid_key_verification(self, backend):
+        with pytest.raises(exceptions.InvalidKey):
+            NullKDF().verify('a', 'b')
```
alex: These need to be bytes objects

simo5: right!

tests/hazmat/primitives/test_ec.py
```diff
@@ -720,3 +726,82 @@ def test_signature_not_bytes(self, backend):
...
+            key.exchange(DummyExchangeAlg())
+
+    def test_unsupported_key_derivation_function(self, backend):
+        key = ec.generate_private_key(ec.SECP384R1(), backend)
+        with pytest.raises(TypeError):
+            key.exchange(ec.ECDH(DummyExchangeAlg()))
+
+    def test_invalid_key_verification(self, backend):
+        with pytest.raises(exceptions.InvalidKey):
+            NullKDF().verify(bytes('a'), bytes('b'))
```
alex: These should just be `b'a'` and `b'b'`

tests/utils.py
alex: Doh! Indeed. Can this fix be pulled out into its own PR?

simo5: Spun off to #2409

tests/hazmat/primitives/test_ec.py
simo5: fixed

tests/utils.py
alex: #2409 had something different, not the loader changes.

reaperhulk:
For coverage:

* Add multibackend tests for `elliptic_curve_exchange_algorithm_supported ` that look like the tests for `elliptic_curve_signature_algorithm_supported` (in `test_multibackend.py`)
* To cover `elliptic_curve_exchange_algorithm_supported` in the openssl `backend.py`, use it in a function like `_skip_curve_unsupported` (found in `test_ec`).

tests/utils.py
simo5: Yeah it had the EC bug, do we really need a separate PR for a single number ?

simo5: Keep in mind this PR actually tests this change, so it seem, appropriate here.

docs/hazmat/primitives/asymmetric/ec.rst
```diff
@@ -122,6 +122,43 @@ Elliptic Curve Signature Algorithms
         :returns: A new instance of a :class:`EllipticCurvePublicKey`
             provider.
 
+Elliptic Curve Key Exchange algorithm
+-------------------------------------
+
+.. class:: ECDH(algorithm)
```
alex: I know we already landed the interface, but I wanted to ask: does it make more sense to pass the KDF to `exchange` or `agree` instead?

src/cryptography/hazmat/backends/openssl/ec.py
```diff
@@ -148,6 +151,32 @@ def verify(self):
...
+        self._kdf = algorithm
+
+    def agree(self, public_key):
+        pri_key = self._private_key._ec_key
+        pub_key = public_key._ec_key
+
+        group = self._backend._lib.EC_KEY_get0_group(pri_key)
+        z_len = (self._backend._lib.EC_GROUP_get_degree(group) + 7) // 8
+        assert z_len > 0
+        z_buf = self._backend._ffi.new("char[]", z_len)
```
alex: This should be `uint8_t`, not `char`

src/cryptography/hazmat/backends/openssl/ec.py
```diff
@@ -148,6 +151,32 @@ def verify(self):
...
+        peer_key = self._backend._lib.EC_KEY_get0_public_key(pub_key)
+
+        r = self._backend._lib.ECDH_compute_key(z_buf, z_len,
+                                                peer_key, pri_key,
+                                                self._backend._ffi.NULL)
+        assert r > 0
```
alex: These shoudl use `backend.openssl_assert`

tests/hazmat/primitives/test_ec.py
```diff
@@ -749,3 +765,99 @@ def test_signature_not_bytes(self, backend):
...
+    )
+    def test_key_exchange_with_vectors(self, backend, vector):
+        _skip_curve_unsupported(backend, ec._CURVE_TYPES[vector['curve']])
+        _skip_exchange_algorithm_unsupported(backend, ec.ECDH(NullKDF()),
+                                             ec._CURVE_TYPES[vector['curve']])
+
+        try:
+            z = self.key_exchange(backend, vector)
+        except ValueError:
+            assert vector['fail'] is True
```
alex: What method are you trying to capture an exception from?

tests/hazmat/primitives/test_ec.py
```diff
@@ -749,3 +765,99 @@ def test_signature_not_bytes(self, backend):
...
+
+        if vector['fail']:
+            if vector['errno'] in [7, 8]:
```
alex: This needs a comment explaining what it means.

docs/hazmat/primitives/asymmetric/ec.rst
ronf: I could imagine combining exchange() and agree(), such that it looked something like:

shared_key = private_key.exchange(ECDH(ECDHnoKDF()), peer_public_key)

or possibly:

shared_key = private_key.exchange(ECDH(), ECDHnoKDF(), peer_public_key)

I don't really much value in splitting the exchange() and agree() calls apart as it currently is.

simo5: I thing exchange/agree is an interface that makes sense
The first sets up the exchange properties, the second gives you the shared_key.
You may be setting up one exchange() and then call agree() many times over, each time with a different party.

tests/hazmat/primitives/test_ec.py
simo5: Any failure to create a key (due to invalid numbers) will cause a ValueError exception in key_exchange() that will bubble up.

alex: Are these errors coming from `agree()`, or from turning the numbers into
keys?

On Wed, Oct 14, 2015 at 11:19 PM, simo5 <notifications@github.com> wrote:

> In tests/hazmat/primitives/test_ec.py
> <https://github.com/pyca/cryptography/pull/2398#discussion_r42079921>:
>
> > +        load_vectors_from_file(
> > +            os.path.join(
> > +                "asymmetric", "ECDH",
> > +                "KASValidityTest_ECCStaticUnified_NOKC_ZZOnly_init.fax"),
> > +            load_kasvs_ecdh_vectors
> > +        )
> > +    )
> > +    def test_key_exchange_with_vectors(self, backend, vector):
> > +        _skip_curve_unsupported(backend, ec._CURVE_TYPES[vector['curve']])
> > +        _skip_exchange_algorithm_unsupported(backend, ec.ECDH(NullKDF()),
> > +                                             ec._CURVE_TYPES[vector['curve']])
> > +
> > +        try:
> > +            z = self.key_exchange(backend, vector)
> > +        except ValueError:
> > +            assert vector['fail'] is True
>
> Any failure to create a key (due to invalid numbers) will cause a
> ValueError exception in key_exchange() that will bubble up.
>
> —
> <https://github.com/pyca/cryptography/pull/2398/files#r42079921>.
>



-- 

docs/hazmat/primitives/asymmetric/ec.rst
ronf: In the most common case, you're using an ephemeral private key here, generated specifically for this exchange. It's very unlikely that you would ever want to use the key with more than one peer, and you really don't want to keep the key around once the shared key has been computed as that might allow an attacker to more easily figure out the shared key (since they only need that private key and the peer's public key that is often sent in the clear).

My ideal API here would actually build that in, generating the private key at random when the key exchange object was created, and providing APIs to return the public key to send to the peer and to generate the shared key from the peer's public key.

tests/hazmat/primitives/test_ec.py
simo5: generally turning numbers into keys

docs/hazmat/primitives/asymmetric/ec.rst
simo5: This is not ephemeral DH, ECDH may very well use actual long term private keys.

tests/hazmat/primitives/test_ec.py
alex: Maybe a silly question, but why do the vectors have cases that fail before
we ever get to doing ECDH with them?

On Wed, Oct 14, 2015 at 11:27 PM, simo5 <notifications@github.com> wrote:

> In tests/hazmat/primitives/test_ec.py
> <https://github.com/pyca/cryptography/pull/2398#discussion_r42080246>:
>
> > +        load_vectors_from_file(
> > +            os.path.join(
> > +                "asymmetric", "ECDH",
> > +                "KASValidityTest_ECCStaticUnified_NOKC_ZZOnly_init.fax"),
> > +            load_kasvs_ecdh_vectors
> > +        )
> > +    )
> > +    def test_key_exchange_with_vectors(self, backend, vector):
> > +        _skip_curve_unsupported(backend, ec._CURVE_TYPES[vector['curve']])
> > +        _skip_exchange_algorithm_unsupported(backend, ec.ECDH(NullKDF()),
> > +                                             ec._CURVE_TYPES[vector['curve']])
> > +
> > +        try:
> > +            z = self.key_exchange(backend, vector)
> > +        except ValueError:
> > +            assert vector['fail'] is True
>
> generally turning numbers into keys
>
> —
> <https://github.com/pyca/cryptography/pull/2398/files#r42080246>.
>



-- 

docs/hazmat/primitives/asymmetric/ec.rst
simo5: What we may want to change about the interface is that we pass in the KDF class, not the instance, and let agree() instantiate it internally every time it is called.
As is exchange+ agree can be called only once indeed because then the KDC will return an error as it already derived a key.

simo5: keyex = key.exchange(ec.ECDH(NullKDF))
keyex.agree(peer1)
keyex.agree(peer2)
...

ronf: I have no objection to the interface supporting both. I just see value in having a version of it that automatically creates a random key for you if you don't want to pass in an existing private key. However, it's only a handful of lines of code to build such a wrapper, so I can live with the API that's here.

simo5: NVM this is not possible as we'd need to passa  bunch of parameters for the KDF too ..
I guess then that shared_key = private_key.exchange(peer_public_key, ECDH(KDF(a, b, c))) is also a sensible interface.
I would put the peer key first.

tests/hazmat/primitives/test_ec.py
simo5: Ask NIST ? :)

docs/hazmat/primitives/asymmetric/ec.rst
ronf: Sounds good to me!

simo5: Another option would be something like:
keyex = ECDH()
shared_key = keyex.exchange(private_key, peer_public_key, KDF(..))

If private_key is None we generate one on the spot
the keyex.private_key will return you the private key should you need it ?

ronf: One problem with that is that you need to pass the curve if you don't pass in a private key. Here's the API I previously suggested which is along these lines:

```python
class ECDH:
    def __init__(self, curve):
        """Construct an ECDH private key"""

    def get_public(self):
        """Return the ECDH public key to send in the handshake"""

    def get_shared(self, public):
        """Return the shared key given the peer's public key"""
```

Instead of having a member which lets you access the private key, there's a method to return the public key corresponding to it, since that's what you need to give to the peer.

To allow an existing private key to be used, that could be passed as another argument to the constructor, though there would need to be a check that the curve & private key matched if we allow both to be passed in.

ronf: One other point here: The API can't require that you pass in the peer's public key to get back an object which lets you query your public key. If you do, there's no way for either side to send their public key since they're both waiting to receive the public key from their peer. You need two steps. The first step creates the private key and lets you get the corresponding public key to put in the handshake, and the second step takes the received public key and returns the shared key.

ronf: After thinking about it over night, I'd like to summarize what I think are a couple of possibilities for an alternate API:

1) Make the relative small change proposed above to combine exchange() and agree() on the private key. Typical usage would be:

```python
    shared_key = private_key.exchange(peer_public_key, ECDH(KDF(a, b, c)))
```

The KDF argument to ECDH() could default to ECDHnoKDF() so that this could be left off when not needed. The ephemeral ECDH case with no KDF would look something like:

```python
    private_key = ec.generate_private_key(curve, backend)
    public_key = private_key.public_key()   # Exchange this with peer
    shared_key = private_key.exchange(peer_public_key, ECDH())
```

2) Create separate ECDH and ECDHE (ephemeral ECDH) classes with a common API but different constructors:

```python
class ECDH:
    def __init__(self, private_key):
        """Begin an ECDH key exchange"""

    def get_public(self):
        """Return the public key to send to the peer"""

    def get_shared(self, peer_public_key, kdf):
        """Return the shared key given the peer's public key"""

class ECDHE(ECDH):
    def __init__(self, curve, backend):
        """Begin an ephemeral ECDH key exchange"""
```

The kdf argument to get_shared() could again default to ECDHnoKDF(). The ECDHE() variant would simply construct a random private key to pass to the parent's constructor. We may need to pass in a backend object to allow that.

In this case, ephemeral ECDHE with no KDF would look something like:

```python
    ecdh = ECDHE(curve, backend)
    public_key = ecdh.get_public()    # Exchange this with peer
    shared_key = ecdh.get_shared(peer_public_key)
```

The non-ephemeral one would look the same, except that a private_key would be passed to ECDH() instead of the curve & backend to ECDHE().

simo5: On 15/10/15 08:59, Ron Frederick wrote:
>> @@ -122,6 +122,43 @@ Elliptic Curve Signature Algorithms
>>           :returns: A new instance of a :class:`EllipticCurvePublicKey`
>>               provider.
>>
>> +Elliptic Curve Key Exchange algorithm
>> +-------------------------------------
>> +
>> +.. class:: ECDH(algorithm)
>
> After thinking about it over night, I'd like to summarize what I think are a couple of possibilities for an alternate API:
>
> 1) Make the relative small change proposed above to combine exchange() and agree() on the private key. Typical usage would be:
>
> ```python
>      shared_key = private_key.exchange(peer_public_key, ECDH(KDF(a, b, c)))
> ```
>
> The KDF argument to ECDH() could default to ECDHnoKDF() so that this could be left off when not needed. The ephemeral ECDH case with no KDF would look something like:

Note that there is no "ECDHnoKDF()" anymore, if you really want to avoid 
key derivation you'll have to create a NullKDF() function (look in the 
tests for an example). This is because we do not want to encourage users 
to avoid key derivation.

> ```python
>      private_key = ec.generate_private_key(curve, backend)
>      public_key = private_key.public_key()   # Exchange this with peer
>      shared_key = private_key.exchange(peer_public_key, ECDH())
> ```
>
> 2) Create separate ECDH and ECDHE (ephemeral ECDH) classes with a common API but different constructors:
>
> ```python
> class ECDH:
>      def __init__(self, private_key):
>          """Begin an ECDH key exchange"""
>
>      def get_public(self):
>          """Return the public key to send to the peer"""
>
>      def get_shared(self, peer_public_key, kdf):
>          """Return the shared key given the peer's public key"""
>
> class ECDHE(ECDH):
>      def __init__(self, curve, backend):
>          """Begin an ephemeral ECDH key exchange"""
> ```

I think having a separate ECDHE class to handle automatic generation of 
keys is a good idea.

> The kdf argument to get_shared() could again default to ECDHnoKDF(). The ECDHE() variant would simply construct a random private key to pass to the parent's constructor. We may need to pass in a backend object to allow that.
>
> In this case, ephemeral ECDHE with no KDF would look something like:
>
> ```python
>      ecdh = ECDHE(curve, backend)
>      public_key = ecdh.get_public()    # Exchange this with peer
>      shared_key = ecdh.get_shared(peer_public_key)
> ```
>
> The non-ephemeral one would look the same, except that a private_key would be passed to ECDH() instead of the curve & backend to ECDHE().

To avoid the need to add NullKDF() it would be sufficient to just 
provide a set_kdf() function, but that would encourage people to not use 
key derivation, I do not think that's a good idea.

ronf: In the case of SSH, a KDF is used, but not directly on the shared key returned by the key exchange operation. Part of the reason for this is that several keys are generated, and so this shared key is hashed with other pieces of information to generate a unique key stream for each of the keys. That means the key exchange needs to return the raw key value to use in multiple KDF calls.

Personally, I'd rather see the KDF removed entirely here. In the case where you only need a single KDF the caller can always apply it on the key value that's returned, and that will give them more flexibility about how it is applied, allowing for things like the SSH use case where multiple KDFs are needed.

simo5: As said, you can pass a NullKDF() to avoid key derivation. I think this is a better pattern, however ECDH() is also hazmat so if @alex and @reaperhulk agree we can allow to pass None and have it automatically ignored withing the code. They suggested to make it mandatory so I'd defer to them.

reaperhulk: So to be clear, `get_public` (in the API proposals @ronf provided this morning) is the spiritual equivalent of `exchange` and `get_shared` is `agree`?

Our current API has `exchange` added to the private key object, but I like the new one better.

```python
private_key = ec.generate_private_key(ec.SECP384R1(), default_backend())
ecdh = ECDH(private_key)
public_key = ecdh.get_public()
shared_key = ecdh.get_shared(peer_public_key, KDF())
```
instead of
```python
private_key = ec.generate_private_key(ec.SECP384R1(), default_backend())
keyex = private_key.exchange(ec.ECDH(KDF()))
sharedkey = keyex.agree(public_key)
```
with a future PR to add ECDHE like
```python
ecdhe = ECDHe(ec.SECP256R1(), default_backend())
public_key = ecdhe.get_public()
shared_key = ecdhe.get_shared(peer_public_key, KDF())
```

cryptography's APIs have strongly leaned toward explicit over implicit, so I'm in favor of requiring `get_shared` to have a KDF argument. As @simo5 said, you can pass a NullKDF for your purposes to obtain the key unmolested, but it seems prudent to strongly encourage the use of KDF by design (many consumers of this will inevitably end up using it outside the context of implementing an existing protocol like SSH).

ronf: Yes, I think that's a good summary.

Regarding the KDF, would it be acceptable to pass "None" explicitly as an argument (and not making it a default) instead of creating a NullKDF() function to pass in? I know I can do it by passing in "lambda x: x", but that's ugly.

Alternately, could NullKDF() be defined somewhere in "cryptography" (probably in hazmat.primitives.kdf, since it could be used in any place a KDF is expected)? That would be more readable than passing in None, and it would provide a standard name instead of each caller defining their own.

simo5: I do not think we should provide necessarily a get_public(), because you can already get the public_key out of the private key if you have pre-created it.
For ECDHE() we can simply have a private_key property which is the private_key that ECDHE() generated for you if you specify no private_key, then all you need to do to get the public_key() to send to the peer is:
ecdhe = ECDHE(ec.SECP256R1(), default_backend())
public_key = ecdhe.key.public_key()
shared_key = ecdhe.agree(peer_public_key, KDF())

I would like to keep the word agree() instead of get_shared as it convey the idea that a computation is being performed. Aslo whether the shared_key is really shared depends on the fact the correct KDF was used, so it may not be shared after all if you messed upo something (Ie you tried to "agree" on a shared key but failed :)

reaperhulk: I am neutral/slightly :+1: on both of these -- @alex, @ronf?

ronf: I'm slightly negative on having a "key" member variable and forcing the caller to know that it is a private key and they need to call a public_key() method on it to convert it to something they can exchange with their peer. It seems like the key exchange class should be providing the value to be exchanged directly. I'd be ok with naming the method "public_key()" instead of "get_public()", having the call be on the ecdh object and not a member of it, but my general leaning is to make method names be actions rather than nouns.

On the other point, I'm not a big fan of the agree() method name, but I agree with the point about making the name more clearly indicate a computation is occurring. Perhaps this could be something like compute_shared() or construct_shared(). Since the KDF is being passed as an argument there, passing the wrong KDF would be no different than having the wrong private or public key. It'll only be the same shared key if all three of those things are consistent on the two sides, but no matter what SOME key that could be shared will be returned.

simo5: A propertyt of ecdhe.public_key sounds ok to me.
I am not sure about a function, the public_key is not computed at the time you request it

On agree() I also considered briefly to call it compute() as in "compute this key exchange", but it doesn't read as well, compute_key() may do though I guess.

ronf: On private keys, public_key() is a method rather than a property, and I think that's because it actually creates a new public key object when it is called. I would think we'd want ECDH to be consistent with that if we kept the name the same. Even private_numbers() is a method.

As for compute_key(), I'm good with that if you prefer it to compute_shared() or compute_shared_key().

simo5: Please see #2415

simo5:
I created PR #2418 as an alternative implementation (heavily based on this PR) that uses the intrface defined in #2415

simo5:
replaced by #2420

PR #2402, opened 2015-10-09
minor fix to handle malformed certificates without hostname

ddcc:
Example certificate that triggers this behavior:
```
-----BEGIN CERTIFICATE-----
MIIDjjCCAzigAwIBAgIQFy56FtbECJlNvTojDFkPSDANBgkqhkiG9w0BAQUFADCB
hzEgMB4GCSqGSIb3DQEJARYRY2JydW5Ad2VzdGVsbC5jb20xCzAJBgNVBAYTAlVT
MQswCQYDVQQIEwJJTDEPMA0GA1UEBxMGQXVyb3JhMRAwDgYDVQQKEwdXZXN0ZWxs
MQwwCgYDVQQLEwNFbmcxGDAWBgNVBAMTD1dlc3RlbGwgSExBQiBDQTAeFw0wMTA5
MjcxNzMwNThaFw0xMTA5MjcxNzQwMTVaMIGHMSAwHgYJKoZIhvcNAQkBFhFjYnJ1
bkB3ZXN0ZWxsLmNvbTELMAkGA1UEBhMCVVMxCzAJBgNVBAgTAklMMQ8wDQYDVQQH
EwZBdXJvcmExEDAOBgNVBAoTB1dlc3RlbGwxDDAKBgNVBAsTA0VuZzEYMBYGA1UE
AxMPV2VzdGVsbCBITEFCIENBMFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBANVNxDxq
1F9n+UDmu7tjcIBvIMZB61sFUpCpcw/l2j3LFcPNEKAQTYyTO6HMpBaSZYwwYP94
fZVtnM8Xc4wJlzcCAwEAAaOCAXwwggF4MBMGCSsGAQQBgjcUAgQGHgQAQwBBMAsG
A1UdDwQEAwIBRjAPBgNVHRMBAf8EBTADAQH/MB0GA1UdDgQWBBT5vV1HXrfYekCQ
1+rV6FLSiaIm2TCCARAGA1UdHwSCAQcwggEDMIG/oIG8oIG5hoG2bGRhcDovLy9D
Tj1XZXN0ZWxsJTIwSExBQiUyMENBLENOPXcya2FkdixDTj1DRFAsQ049UHVibGlj
JTIwS2V5JTIwU2VydmljZXMsQ049U2VydmljZXMsQ049Q29uZmlndXJhdGlvbixE
Qz1obGFiLERDPWNvbT9jZXJ0aWZpY2F0ZVJldm9jYXRpb25MaXN0P2Jhc2U/b2Jq
ZWN0Y2xhc3M9Y1JMRGlzdHJpYnV0aW9uUG9pbnQwP6A9oDuGOWh0dHA6Ly93Mmth
ZHYuaGxhYi5jb20vQ2VydEVucm9sbC9XZXN0ZWxsJTIwSExBQiUyMENBLmNybDAQ
BgkrBgEEAYI3FQEEAwIBADANBgkqhkiG9w0BAQUFAANBAFpCg9Wp0XqsZiBKUsdl
2NmpDbXSfgX7btKT6NwdnzIIxKCg9O5uCEwCl21Roq51tqeN7MZxbjen7xZMESXF
PDA=
-----END CERTIFICATE-----
```


alex:
You're going to need to add a test case for this, otherwise the tests will fail because of missing coverage.

reaperhulk:
Thanks for the contribution. Could you add the example certificate into our test vectors (`vectors/cryptography_vectors/x509/custom` is the right dir), add an entry to the `test-vectors.rst` describing the content of the new vector, and then write a test that uses it to demonstrate the new code path?

Looking at the cert it appears the URI that's failing is actually: `ldap:///CN=Westell%20HLAB%20CA,CN=w2kadv,CN=CDP,CN=Public%20Key%20Services,CN=Services,CN=Configuration,DC=hlab,DC=com?certificateRevocationList?base?objectclass=cRLDistributionPoint`

With your hostname change it will return:

'ldap:/CN=Westell%20HLAB%20CA,CN=w2kadv,CN=CDP,CN=Public%20Key%20Services,CN=Services,CN=Configuration,DC=hlab,DC=com?certificateRevocationList?base?objectclass=cRLDistributionPoint'

which omits the `//` at the beginning. (I'm unsure if that's a bug or more correct?)

reaperhulk:
jenkins, ok to test

reaperhulk:
An empty dNSName would also trigger the same failure, so we should probably handle that case as well.

ddcc:
@reaperhulk: The library function `urlparse.urlunsplit()` conforms to [RFC 3986](https://tools.ietf.org/html/rfc3986.html#section-5.3), which seems to state that if the `authority`  component is defined, it is prepended with `//` and appended to the URI. In this case, since the `userinfo`, `host`, and `port` components of `authority` are empty, I believe it is excluded. The remaining `\` is appended from the `path` component, since it is absolute.

I have added two tests and updated the documentation.

reaperhulk:
The docs spell checker doesn't like `hostname` (line 201 on `test-vectors.rst`). You can either reword it or add hostname to `docs/spelling_wordlist.txt` to resolve the issue.

(Note to self, figure out why docs/pep8 jenkins jobs aren't linking in the status API)

reaperhulk:
Thanks for the contribution! Feel free to send a PR adding yourself to `AUTHORS.rst`.

PR #2403, opened 2015-10-10
add shebang for setup.py scripts

ddcc:
Minor patch to add a shebang to the `setup.py` scripts, this seems to be done by other packages such as `Twisted` and `setuptools`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 # This file is dual licensed under the terms of the Apache License, Version
 # 2.0, and the BSD License. See the LICENSE file in the root of this repository
 # for complete details.
--- a/vectors/setup.py
+++ b/vectors/setup.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 # This file is dual licensed under the terms of the Apache License, Version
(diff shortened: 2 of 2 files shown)


reaperhulk:
I'm assuming these exist for the weird edge case of someone marking `setup.py` as executable and then wanting to run `setup.py {install,test,build_ext,whatever}` without typing "python". /cc @dstufft to confirm that there's not some bizarre reason this is actually needed.

ddcc:
Yes, that's how I encountered it in the first place, although from an external non-UNIX filesystem that applies a +x mask to all files.

dstufft:
It's not really needed except for people who want to execute it directly.

reaperhulk:
I guess I don't see any real problem with adding this, but could you explain a bit more about the system you encountered that expected this behavior @ddcc? Might be useful to know about it for future reference

ddcc:
@reaperhulk It's a bit of a weird setup, but it was from inside a VMWare virtual machine accessing a shared folder on the host NTFS filesystem. Since the host doesn't support UNIX permissions, I believe the default behavior is to apply an executable mask to all files.

Aside from that, I forgot to prepend `python` to `setup.py` when attempting to run the tests.

alex:
Jenkins, ok to test.

PR #2406, opened 2015-10-11
update docs to point at newer, fancier openssl builder

reaperhulk:
This will also allow us to delete the two older openssl builder jobs in jenkins woo.


alex:
@stevepeak Hey Steve, is codecov having an issue? Status checks aren't updating here.

stevepeak:
Hey @alex sorry for the delay, there was a large work queue and some jobs were timing out. Thank you for your patience.

PR #2409, opened 2015-10-12
Catch Invalid X or Y points and raise a ValueError

simo5:
This fixes an assert trown when invalid X or Y points are used in EC curves.
Found while implementing ECDH and using CAVS test vectors

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -1776,9 +1776,13 @@ class Backend(object):
             self.openssl_assert(res == 1)
 
             res = self._lib.BN_cmp(bn_x, check_x)
-            self.openssl_assert(res == 0)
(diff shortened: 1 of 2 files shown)


reaperhulk:
To merge this we need test cases that trigger the new exceptions. Additionally, if any openssl errors may be added to the stack when these checks fail it should call the backend `_consume_errors()` method.

simo5:
Ok pushed a new version that adds specific tests and calls _consume_errors() before raising

reaperhulk:
One last thing, the top of the test method you added this to should have the line `         _skip_curve_unsupported(backend, ec.SECP521R1())` added (in addition to the existing line about `SECP256R1`)

simo5:
Ok, moved the tests to their own test function and added the skip-curve thing at the top of it.
This way if you have 256 but not 521 you skip only those tests.

PR #2415, opened 2015-10-15
Change the Key Exchange Context interface

simo5:
As discussed on PR #2398 propose a different Key Exchange Context interface.


reaperhulk:
So this interface would then be attached to the ECDH, ECDHE, DH, DHE classes rather than the keys, correct?

simo5:
Exactly

simo5:
Closing this one, replaced by #2419

PR #2418, opened 2015-10-15
ECDH exchange 2

simo5:
This is an altrernative ECDH Exchange implementation based on the alternative Key Exchange interface in PR #2415


simo5:
Killed by #2420 :-)

PR #2422, opened 2015-10-16
Add --no-cache-dir when building static wheels.

Lukasa:
Per @geradcoles suggestion in #2338 and certifi/python-certifi#26.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -166,7 +166,7 @@ dependencies.
     ./config no-shared no-ssl2 -fPIC --prefix=${CWD}/openssl
     make && make install
     cd ..
-    CFLAGS="-I${CWD}/openssl/include" LDFLAGS="-L${CWD}/openssl/lib" pip wheel cryptography
+    CFLAGS="-I${CWD}/openssl/include" LDFLAGS="-L${CWD}/openssl/lib" pip wheel --no-use-wheel cryptography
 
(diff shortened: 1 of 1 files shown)


alex:
cc: @dstufft

reaperhulk:
Should it be this flag or `--no-use-wheel`?

Lukasa:
That's a good question: @dstufft would know.

dstufft:
Probably you want ``--no-binary cryptography``, like:

```console
$ CFLAGS="-I${CWD}/openssl/include" LDFLAGS="-L${CWD}/openssl/lib" pip wheel --no-binary cryptogrpahy cryptography
```

I think that will do what you want... although it only works in the latest pip, the ``--no-use-wheel`` version is deprecated but does the same thing (though it applies to everything and not just cryptography).

dstufft:
Specifically, the ``--no-use-wheel`` version is more compat with older versions, but is deprecated in favor of ``--no-binary``.

Lukasa:
@dstufft Is there any good reason not to use --no-use-wheel, then? I'm worried about providing instructions that only work with bleeding-edge pip.

dstufft:
It'll go away at some point, but that's it.

Lukasa:
Ok, I've switched to `--no-use-wheel` instead.

Issue #2346, opened 2015-09-16
Elliptic curve serialization/deserialization from point value byte strings


ronf:
For ECDH key exchange and ECDSA signing/verification in AsyncSSH, I need the ability to serialize and deserialize EC point value encoded as byte strings from/to EC public keys.

The encoding of the point values into octet strings is described in http://www.secg.org/sec1-v2.pdf in section 2.3.3. This encoding is used by PKCS#1/PKCS#8 DER and PEM encoding of EC keys, as well as by SSH (with its own encoding of strings and integers replacing DER/PEM), so the existing DER/PEM serialization may already include this support. I'd just like to see it broken out so it can be used without constructing a full EC DER/PEM public key.

The serialization depends on the curve, as the number of bytes used to encode the x & y values of the point vary based on the curve's key size. That information is already available in EllipticCurvePublicNumbers, though, via curve.key_size. It just needs to be rounded up to the next even number of bytes. For now, I've coded this myself and it's just a few lines, but it would be nice to have this built into PyCA to avoid callers having to directly construct EllipticCurvePublicNumbers from raw x and y values. Here's an example of the encode and decode functions given the key length of the curve expressed in bytes:

```python
def decode_ec_point(keylen, data):
    """Decode an octet string into an elliptic curve point"""

    if data == b'\x00':
        return None, None
    elif data.startswith(b'\x04'):
        if len(data) == 2*keylen + 1:
            return (int.from_bytes(data[1:keylen+1], 'big'),
                    int.from_bytes(data[keylen+1:], 'big'))
        else:
            raise ValueError('Invalid elliptic curve point data length')
    else:
        raise ValueError('Unsupported elliptic curve point type')

def encode_ec_point(keylen, x, y):
    """Encode an elliptic curve point as an octet string"""

    if x is None:
        return b'\x00'
    else:
        return b'\x04' + x.to_bytes(keylen, 'big') + y.to_bytes(keylen, 'big')
```

The "None, None" return value here is how the "infinity" EC point value is serialized. There's also an encoding which does "point compression" to represent the y value as a single bit that's not supported here yet as I haven't needed it, but it is described in the link above. Since it involves some hairy math (prime modulo square root), we'd probably want to see if it is something supported by a backend if we wanted to add that.

wbond:
Also note that EC point compression is covered by https://www.google.com/patents/US6252960 until 2018. I ended up deciding not to support it in my library because of this issue.

alex:
[...]

PR #2427, opened 2015-10-17
ECDH take 4

alex:


reaperhulk:
Since we've been rapidly evolving this API and it's come partially back around I'd like @ronf to weigh in on whether he's happy with this.

ronf:
Is there a summary somewhere of the latest proposal? From the file diff, it looks like we're back to an exchange() method on the private key object rather than passing the key to the ECDH() object, but that exchange() covers both what exchange() and agree() did before. I'm ok with that, though I think passing the keys to ECDH() would have been cleaner. I also don't see a good way to do ephemeral DH or ECDH with this style of API, other than for the caller to be responsible for creating the ephemeral keys explicitly each time.

In this version, is the KDF removed entirely? I didn't see that showing up any more.

reaperhulk:
Correct ,the KDF is removed entirely.

ECDHE in this system would require a new ec key to be generated each time and then used to `exchange`. That does require a reasonable understanding of the mechanics of ECDHE though.

I am neutral on ECDH class vs this, but I'd be interested in knowing your reasoning for preferring a standalone class.

alex:
FWIW, I don't think ECDHE deserves special support, it's not a distinct
algorithm.

On Mon, Oct 19, 2015 at 2:57 PM, Paul Kehrer <notifications@github.com>
wrote:

> Correct ,the KDF is removed entirely.
>
> ECDHE in this system would require a new ec key to be generated each time
> and then used to exchange. That does require a reasonable understanding
> of the mechanics of ECDHE though.
>
> I am neutral on ECDH class vs this, but I'd be interested in knowing your
> reasoning for preferring a standalone class.
>
> —
> <https://github.com/pyca/cryptography/pull/2427#issuecomment-149312448>.
>



-- 

ronf:
I'm good with the KDF removal. That can easily be handled after the shared key data is returned.

As for ECDHE, the discussion proposed two options: One was a subclass of ECDH() with a different constructor but no other child methods, and another was to use a class method to achieve that. Both took an algorithm and backend instead of the private key itself. If we don't want whole separate class, the second option seemed like a good approach to me.

That said, the difference really is pretty minor. As long as both DH and ECDH support a top-level "generate_private_key()" method, I think this works, and is an improvement over the original exchange/agree proposal. I'm good with it.

reaperhulk:
Okay, I'm happy enough with this that I'll go ahead and merge. Thanks for your hard work on this @simo5 (and @alex) I'm going to file a follow up issue to write some docs for how to do ECDHE utilizing this API.

ronf:
I updated my local copy of AsyncSSH to use the new API, and it appears to be working. Here's what I ended up with to match the internal API I previously defined. The "_curves" dictionary maps byte-string curve names to PyCA objects (allowing me to swap this for another backend easily without changing the calling code):

```python
class ECDH:
    """A shim around PyCA for ECDH key exchange"""

    def __init__(self, curve_id):
        try:
            curve, _ = _curves[curve_id]
        except KeyError: # pragma: no cover, other curves not registered
            raise ValueError('Unknown EC curve %s' %
                             curve_id.decode()) from None

        self._priv_key = ec.generate_private_key(curve, backend)
        self._keylen = (curve.key_size + 7) // 8

    def get_public(self):
        """Return the public key to send in the handshake"""

        pub = self._priv_key.private_numbers().public_numbers
        return encode_ec_point(self._keylen, pub.x, pub.y)

    def get_shared(self, peer_public):
        """Return the shared key from the peer's public key"""

        x, y = decode_ec_point(self._keylen, peer_public)
        pub = ec.EllipticCurvePublicNumbers(x, y, self._priv_key.curve)
        peer_key = pub.public_key(backend)
        shared_key = self._priv_key.exchange(ec.ECDH(), peer_key)
        
        return int.from_bytes(shared_key, 'big')
```

Cryptography is missing functions to encode/decode EC point values as byte strings (see #2346) so I had to provide that myself. However, the rest was pretty straightforward.

PR #2430, opened 2015-10-17
Take 4: Add an Elliptic Curve Key Exchange Algorithm(ECDH)

simo5:
No classes this time, just a new method on ellipticCurvePrivateKey


reaperhulk:
#2427 supersedes this

PR #2435, opened 2015-10-20
encode countryName with PrintableString

### LINKED ISSUE #2407: cryptography.x509 encodes countryName as a Utf8String but it should be a PrintableString
[RFC 5280](https://www.ietf.org/rfc/rfc5280.txt) says that countryName should be as above, and unlike the other elements there (localeName etc.) cannot be a DirectoryString.

The consequences are that while it seems like `openssl s_server` is OK with certificates generated in this way, golang's (seemingly correct implementation which follows the RFC) cannot parse the certificate.

@reaperhulk says https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/backends/openssl/backend.py#L114 is what should be changed.

reaperhulk:
This commit adds a dependency on asn1crypto (thanks to wbond for pushing a release to pypi) for testing purposes to parse the certificate and confirm that countryName is encoded with PrintableString while other fields are UTF8String. This is a test only dep (but could eventually supplant pyasn1).

Fixes #2407

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -52,7 +52,7 @@ from cryptography.hazmat.primitives.ciphers.algorithms import (
 from cryptography.hazmat.primitives.ciphers.modes import (
     CBC, CFB, CFB8, CTR, ECB, GCM, OFB
 )
-from cryptography.x509.oid import ExtensionOID
+from cryptography.x509.oid import ExtensionOID, NameOID
 
 
 _MemoryBIO = collections.namedtuple("_MemoryBIO", ["bio", "char_ptr"])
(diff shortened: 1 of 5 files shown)


reaperhulk:
jenkins, retest this please

reaperhulk:
jenkins, retest this please

tests/test_x509.py
```diff
@@ -834,6 +838,43 @@ def test_build_cert(self, backend):
...
+            cert.public_bytes(serialization.Encoding.DER),
+            asn1Spec=rfc2459.Certificate()
+        )
+        tbs_cert = parsed.getComponentByName('tbsCertificate')
+        subject = tbs_cert.getComponentByName('subject')
+        issuer = tbs_cert.getComponentByName('issuer')
+        # \x13 is printable string. The first byte of the value of the
+        # node corresponds to the ASN.1 string type.
+        assert str(subject[0][0][0][1])[0] == "\x13"
+        assert str(issuer[0][0][0][1])[0] == "\x13"
```
alex: You can drop the `str`, `AttributeValue` is directly indexable.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -119,12 +119,14 @@ def _encode_name(backend, attributes):
     for attribute in attributes:
         value = attribute.value.encode('utf8')
         obj = _txt2obj_gc(backend, attribute.oid.dotted_string)
+        if attribute.oid == NameOID.COUNTRY_NAME:
+            # Per RFC5280 countryName should be encoded as PrintableString,
+            # not UTF8String
```
alex: Can we cite the specific section, I'm pretty sure in the future I will want to find where and would love to do better than "find"

reaperhulk: Sure. It's Appendix A.1: `X520countryName ::=     PrintableString (SIZE (2))`. Supposedly X.520 contains the normative description as well.

Issue #2433, opened 2015-10-20
Write some docs on how to do ECDHE


reaperhulk:
The new ECDH API now exists and we should add some docs explaining how to do ECDHE with it.

reaperhulk:
Done (for now)

PR #2437, opened 2015-10-20
add a little bit to the docs about ECDHE

reaperhulk:
This is a first draft to start some discussion.

refs #2433

### DIFF
--- a/docs/hazmat/primitives/asymmetric/ec.rst
+++ b/docs/hazmat/primitives/asymmetric/ec.rst
@@ -147,6 +147,10 @@ Elliptic Curve Key Exchange algorithm
         ... ).public_key()
         >>> shared_key = private_key.exchange(ec.ECDH(), peer_public_key)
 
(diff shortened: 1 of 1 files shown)


docs/hazmat/primitives/asymmetric/ec.rst
```diff
@@ -147,6 +147,8 @@ Elliptic Curve Key Exchange algorithm
         ... ).public_key()
         >>> shared_key = private_key.exchange(ec.ECDH(), peer_public_key)
 
+    When performing an ephemeral form of this exchange (commonly written ECDHE
+    or EECDH) you must generate a new private key for each ``exchange``.
```
alex: Can we use stronger prose here, like "Unless you are implementing a really bad protocol, you should be generating an ew private key for each exchange operation" (and link to the generate prive key docs), and maybe a link to a good reference on FS as well?

PR #2446, opened 2015-10-25
move lock initialization to during binding import

reaperhulk:
Previously we attempted to register our openssl locks only if the backend was initialized, but we should really just do it immediately. Consumers like PyOpenSSL already call init_static_locks after importing the binding and if a library wants to replace the locks with something else they can do so themselves.

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/binding.py
+++ b/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -172,3 +172,7 @@ class Binding(object):
                     mode, n, file, line
                 )
             )
+
+
+# OpenSSL is not thread safe until the locks are initialized.
+Binding.init_static_locks()
--- a/src/cryptography/hazmat/bindings/openssl/binding.py
+++ b/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -174,5 +174,7 @@ class Binding(object):
             )
 
 
-# OpenSSL is not thread safe until the locks are initialized.
+# OpenSSL is not thread safe until the locks are initialized. We initialize in
+# module scope to cause initialization whenever this module is imported (and
+# try to get some benefit from the import lock).
 Binding.init_static_locks()
--- a/src/cryptography/hazmat/bindings/openssl/binding.py
+++ b/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -174,7 +174,9 @@ class Binding(object):
             )
 
 
-# OpenSSL is not thread safe until the locks are initialized. We initialize in
-# module scope to cause initialization whenever this module is imported (and
-# try to get some benefit from the import lock).
+# OpenSSL is not thread safe until the locks are initialized. We call this
+# method in module scope so that it executes with the import lock. On
+# Pythons < 3.4 this import lock is a global lock, which can prevent a race
+# condition registering the OpenSSL locks. On Python 3.4+ the import lock
+# is per module so this approach will not work.
 Binding.init_static_locks()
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -172,3 +172,7 @@ def _lock_cb(cls, mode, n, file, line):
                     mode, n, file, line
                 )
             )
+
+
+# OpenSSL is not thread safe until the locks are initialized.
```
alex: Comment should explain why we're putting it at module scope?

reaperhulk: How about:

```
# OpenSSL is not thread safe until the locks are initialized. We initialize in
# module scope to cause initialization whenever this module is imported (and
# try to get some benefit from the import lock).
```

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -172,3 +172,9 @@ def _lock_cb(cls, mode, n, file, line):
                     mode, n, file, line
                 )
             )
+
+
+# OpenSSL is not thread safe until the locks are initialized. We initialize in
+# module scope to cause initialization whenever this module is imported (and
+# try to get some benefit from the import lock).
```
alex: Can you both shorten and expand here "We initialize at module scope so that we run with the import lock [explain why the import lock is relevant]"

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -172,3 +172,11 @@ def _lock_cb(cls, mode, n, file, line):
                     mode, n, file, line
                 )
             )
+
+
+# OpenSSL is not thread safe until the locks are initialized. We call this
+# method in module scope so that it executes with the import lock. On
+# Pythons < 3.4 this import lock is a global lock, which can prevent a race
+# condition registering the OpenSSL locks. On Python 3.4+ the import lock
+# is per module so this approach will not work.
```
alex: Can you remind me why we decided not to go with `imp.import_lock()`

reaperhulk: It didn't seem to work for people who were testing, but it won't hurt so maybe we should still do it?

alex: If that doesn't work, does any of this work?

On Mon, Oct 26, 2015 at 9:42 PM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/2446#discussion_r43076455>:
>
> > @@ -172,3 +172,11 @@ def _lock_cb(cls, mode, n, file, line):
> >                      mode, n, file, line
> >                  )
> >              )
> > +
> > +
> > +# OpenSSL is not thread safe until the locks are initialized. We call this
> > +# method in module scope so that it executes with the import lock. On
> > +# Pythons < 3.4 this import lock is a global lock, which can prevent a race
> > +# condition registering the OpenSSL locks. On Python 3.4+ the import lock
> > +# is per module so this approach will not work.
>
> It didn't seem to work for people who were testing, but it won't hurt so
> maybe we should still do it?
>
> —
> <https://github.com/pyca/cryptography/pull/2446/files#r43076455>.
>



-- 

reaperhulk: It definitely doesn't fix it, but it may help some users. I'm still completely at a loss for how to fix this without just doing it all at the C layer.

alex: Ok.

PR #2447, opened 2015-10-26
add support for encoding/decoding elliptic curve points

reaperhulk:
Heavily based on the work of @ronf in #2346.


docs/hazmat/primitives/asymmetric/utils.rst
```diff
@@ -28,3 +28,40 @@ Asymmetric Utilities
     :param int s: The raw signature value ``s``.
 
     :return bytes: The encoded signature.
+
+.. function:: encode_ec_point(key_length, x, y)
```
alex: Should this take `curve` instead of `key_length`?

reaperhulk: That's probably a better path. Fixes the ambiguity of bytes vs bits as well

ronf: Yeah - passing in curve here instead of directly passing key length is definitely preferable.

docs/hazmat/primitives/asymmetric/utils.rst
```diff
@@ -28,3 +28,50 @@ Asymmetric Utilities
     :param int s: The raw signature value ``s``.
 
     :return bytes: The encoded signature.
+
+.. function:: encode_ec_point(curve, x, y)
```
alex: Should this take a `PublicNumbers` instance?

reaperhulk: Hmm, that makes some sense to me but what do you think @ronf?

docs/hazmat/primitives/asymmetric/utils.rst
```diff
@@ -28,3 +28,50 @@ Asymmetric Utilities
...
+    `SEC 1 v2.0`_ section 2.3.3. This function only supports uncompressed
+    points.
+
+    :param curve: A
+        :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurve`
+        provider.
+
+    :param x: The x value of the point.
+
+    :type: int or None
```
alex: Not sure what you're trying to do here, but it's not rendering the way you want: https://jenkins.cryptography.io/job/cryptography-pr-docs/826/artifact/cryptography/docs/_build/html/hazmat/primitives/asymmetric/utils.html#cryptography.hazmat.primitives.asymmetric.utils.encode_ec_point

docs/hazmat/primitives/asymmetric/utils.rst
```diff
@@ -28,3 +28,50 @@ Asymmetric Utilities
...
+
+    .. versionadded:: 1.1
+
+    Decodes a byte string as described in `SEC 1 v2.0`_ section 2.3.3 to the
+    ``x`` and ``y`` integer values. This function only supports uncompressed
+    points.
+
+    :param curve: A
+        :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurve`
+        provider.
```
alex: I hate the language "provider", what about "instance"?

docs/hazmat/primitives/asymmetric/utils.rst
reaperhulk: x can be int or None, probably just need to indent?

src/cryptography/hazmat/primitives/asymmetric/utils.py
```diff
@@ -71,3 +72,36 @@ def encode_dss_signature(r, s):
     sig.setComponentByName('r', r)
     sig.setComponentByName('s', s)
     return encoder.encode(sig)
+
+
+def encode_ec_point(curve, x, y):
+    if not isinstance(curve, EllipticCurve):
+        raise TypeError("curve must be an EllipticCurve instance")
+
+    if x is None:
```
alex: wat. Explanatory comment please.

src/cryptography/hazmat/primitives/asymmetric/utils.py
```diff
@@ -71,3 +72,36 @@ def encode_dss_signature(r, s):
...
+        return b'\x00'
+    else:
+        # Get the ceiling of curve.key_size / 8
```
alex: Can you write "key_size is in bits. Conver to bytes and round up." instead?

src/cryptography/hazmat/primitives/asymmetric/utils.py
```diff
@@ -71,3 +72,36 @@ def encode_dss_signature(r, s):
...
+
+
+def decode_ec_point(curve, data):
+    if not isinstance(curve, EllipticCurve):
+        raise TypeError("curve must be an EllipticCurve instance")
+
+    if data == b'\x00':
+        return None, None
+    elif data.startswith(b'\x04'):
+        # Get the ceiling of curve.key_size / 8
```
alex: Same.

docs/hazmat/primitives/asymmetric/utils.rst
alex: I doubt indent works. I'd drop the `:type:` feature and describe it in prose.

alex:
Can we just drop support for null points? I'm not sure I've ever seen a use case for it. Unless I've missed something?

reaperhulk:
Probably a question for @ronf. I have no knowledge about null points.

alex:
(My usual reference, Go, doesn't handle null points)

On Mon, Oct 26, 2015 at 9:08 PM, Paul Kehrer <notifications@github.com>
wrote:

> Probably a question for @ronf <https://github.com/ronf>. I have no
> knowledge about null points.
>
> —
> <https://github.com/pyca/cryptography/pull/2447#issuecomment-151330787>.
>



-- 

reaperhulk:
If we don't handle them presumably we'd raise a `ValueError`?

alex:
Yeah, same as any other unsupported type

On Mon, Oct 26, 2015 at 9:21 PM, Paul Kehrer <notifications@github.com>
wrote:

> If we don't handle them presumably we'd raise a ValueError?
>
> —
> <https://github.com/pyca/cryptography/pull/2447#issuecomment-151334696>.
>



-- 

docs/hazmat/primitives/asymmetric/utils.rst
reaperhulk: (Turns out you can do this with `:type argname:`)

docs/hazmat/primitives/asymmetric/utils.rst
ronf: If we wanted to tie this to EllipticCurvePublicNumbers, I think it would make sense to have something like an encode_point() or to_point() method on that which returned the encoded bytes, and a decode_point() or from_point() class method that took encoded bytes and a curve and constructed an instance from them.

ronf:
I haven't specifically run across a need for null points for ECDH or ECDSA, and in fact a good generator point should not generate null points if you choose the private value properly. So, we could probably get away with not supporting it. However, it would be nice to have this eventually for completeness.

Does EllipticCurvePublicNumbers() have any way to hold null points yet? It doesn't seem to let you pass in None for x or y.

reaperhulk:
I actually like the idea of these being methods on `EllipticCurvePublicNumbers`. Null points aren't supported, but we can leave that unsupported until we run into a need for them.

src/cryptography/hazmat/primitives/asymmetric/ec.py
```diff
@@ -259,6 +259,34 @@ def __init__(self, x, y, curve):
...
+            b'\x04' + utils.int_to_bytes(self.x, byte_length) +
+            utils.int_to_bytes(self.y, byte_length)
+        )
+
+    @classmethod
+    def from_encoded_point(cls, curve, data):
+        if not isinstance(curve, EllipticCurve):
+            raise TypeError("curve must be an EllipticCurve instance")
+
+        if data == b'\x00':
```
alex: I think you should just drop this branch, only `data.startswith(b'\x04')` and the `else`.

src/cryptography/hazmat/primitives/asymmetric/ec.py
```diff
@@ -259,6 +259,34 @@ def __init__(self, x, y, curve):
...
+            byte_length = (curve.key_size + 7) // 8
+            if len(data) == 2 * byte_length + 1:
+                x = utils.int_from_bytes(data[1:byte_length + 1], 'big')
+                y = utils.int_from_bytes(data[byte_length + 1:], 'big')
+            else:
+                raise ValueError('Invalid elliptic curve point data length')
+        else:
+            raise ValueError('Unsupported elliptic curve point type')
+
+        return cls(x, y, curve)
```
alex: We only assign `x` and `y` once, I think you can just push this up to that line.

PR #2455, opened 2015-10-29
Error cleanly if the public and private keys to an ECDH key exchange …

alex:
…are on different curves

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/ec.py
+++ b/src/cryptography/hazmat/backends/openssl/ec.py
@@ -182,7 +182,7 @@ class _EllipticCurvePrivateKey(object):
                 _Reasons.UNSUPPORTED_EXCHANGE_ALGORITHM
             )
 
-        if type(peer_public_key.curve) is not type(self.curve):
+        if peer_public_key.curve.name != self.curve.name:
             raise ValueError(
                 "peer_public_key and self are not on the same curve"
             )
--- a/src/cryptography/hazmat/backends/openssl/ec.py
+++ b/src/cryptography/hazmat/backends/openssl/ec.py
@@ -182,6 +182,11 @@ class _EllipticCurvePrivateKey(object):
                 _Reasons.UNSUPPORTED_EXCHANGE_ALGORITHM
             )
 
+        if type(peer_public_key.curve) is not type(self.curve):
+            raise ValueError(
+                "peer_public_key and self are not on the same curve"
+            )
+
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/backends/openssl/ec.py
```diff
@@ -182,6 +182,11 @@ def exchange(self, algorithm, peer_public_key):
                 _Reasons.UNSUPPORTED_EXCHANGE_ALGORITHM
             )
 
+        if type(peer_public_key.curve) is not type(self.curve):
```
reaperhulk: This fails py3's pep8 with `E721 do not compare types, use 'isinstance()'`

alex: Yeah, I don't know what that means. What's teh right way to check if two curves rae equivilant.

ronf: I know this was already fixed another way, but for future reference you could do it as a type check by saying:

```python
    if not isinstance(peer_public_key.curve, type(self.curve)):
```

alex: Yeah, I wonder if we should just have an `__eq__` on curves.

On Thu, Oct 29, 2015 at 9:37 AM, Ron Frederick <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/2455#discussion_r43384597>:
>
> > @@ -182,6 +182,11 @@ def exchange(self, algorithm, peer_public_key):
> >                  _Reasons.UNSUPPORTED_EXCHANGE_ALGORITHM
> >              )
> >
> > +        if type(peer_public_key.curve) is not type(self.curve):
>
> I know this was already fixed another way, but for future reference you
> could do it as a type check by saying:
>
>     if not isinstance(peer_public_key.curve, type(self.curve)):
>
> —
> <https://github.com/pyca/cryptography/pull/2455/files#r43384597>.
>



-- 

ronf: Yeah, that would probably look cleaner. When writing an __eq__ for classes that have other members to compare, part of it is usually involves an isinstance() check of this sort before comparing the members. In this case, if these classes have no other state, the isinstance() by itself could  be enough.

That said, is there any reason these are classes instead of instances that could just be compared directly with "is"? Is it just for consistency with other parts of the API where arguments might be passed when constructing such an object?

One possibility here would be to make these singletons, so you still construct them from classes, but the constructor always returns the same instance.

PR #2456, opened 2015-10-29
Prefer travis_retry instead of bash shenangins.

alex:
This has a few advantages:

- reduces duplication which decreases changes of typos and makes sure we actually covera ll the line
- consistent number of retries
- better output on retries

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -132,7 +132,8 @@ matrix:
           env: TOXENV=docs
 
 install:
-    - ./.travis/install.sh
+    # source this so that travis_retry is available
+    - source .travis/install.sh
 
 script:
     - ./.travis/run.sh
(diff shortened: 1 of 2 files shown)


alex:
Eh, shit. Bash functions don't survive new invocations of the shell (which should have been obvious, but lol). We can either

- a) `source` the scripts instead of running them
- b) ignore this PR

@reaperhulk preference?

reaperhulk:
Sourcing the scripts to gain the helper functions seems reasonable to me. I guess if we want to do that we should put a comment explaining why we source rather than invoking directly though.

alex:
I don't understand why this makes 2.6 fail. @reaperhulk do you?

alex:
not worth it

PR #2458, opened 2015-10-29
Adding elliptic curve test fixtures

PeterHamilton:
This change adds pre-generated elliptic curve keys to be used in elliptic curve tests.


alex:
Jenkins, ok to test.

PeterHamilton:
What's the best way to resolve coverage for these fixtures? I don't just want to add arbitrary tests to cover testing data. I'm planning on using some of these fixtures in my subsequent certificate validation patch, but that doesn't address current coverage.

alex:
If you could replace one tests generate or load-from-vector using one of
these vlaues, that'd work.

On Thu, Oct 29, 2015 at 1:44 PM, Peter Hamilton <notifications@github.com>
wrote:

> What's the best way to resolve coverage for these fixtures? I don't just
> want to add arbitrary tests to cover testing data. I'm planning on using
> some of these fixtures in my subsequent certificate validation patch, but
> that doesn't address current coverage.
>
> —
> <https://github.com/pyca/cryptography/pull/2458#issuecomment-152262299>.
>



-- 

PeterHamilton:
Hmm, older versions of OpenSSL don't support elliptic curves. I'll find a better test to tweak.

PeterHamilton:
Not entirely sure why the build failed. It looks like there was a problem with the logging when setting up the Travis run. It didn't happen consistently across the different builds either.

alex:
@PeterHamilton looks like it's a bug in PyPy's nightly. I'm investigating, otherwise this patch lgtm.

reaperhulk:
How did you generate these? I'm assuming with ec.generate_key and then calling the `numbers` methods?

PeterHamilton:
@reaperhulk I actually used OpenSSL and then dumped the numbers via `serialization.load_pem_private_key` and `load_pem_public_key`. Kind of roundabout but it worked.

reaperhulk:
Okay, that's fine. Verified these load into keys so :+1:

PR #2459, opened 2015-10-30
pypy 4.0.0

alex:
only on linux, on OS X we're waiting for it to be in a pyenv release so we can get it in brew


alex:
jenkins, retest this please

PR #2467, opened 2015-11-04
these flags were actually added in 1.0.2beta2, not before that.

reaperhulk:
The check previously assumed all versions of 1.0.2 supported these flags, but in reality they were added in 1.0.2-beta2. The updated version check properly handles cases where users are using 1.0.2-beta1 :frowning: .

Fixes https://github.com/letsencrypt/letsencrypt/issues/1335


alex:
What is the `OPENSSL_VERSION_NUMBER` for a released 1.0.2?

reaperhulk:
Probably f but the only way to be sure is to check out the 1.0.2 tag

PR #2469, opened 2015-11-08
add progress bar to wheel downloads

reaperhulk:


tasks.py
```diff
@@ -66,15 +68,28 @@ def download_artifacts(session):
...
+            )
+            assert response.headers["content-length"]
+            print("Downloading {0}".format(artifact["fileName"]))
+            bar = ProgressBar(
+                expected_size=int(response.headers["content-length"]),
+                filled_char="="
             )
+            content = b''
+            for data in response.iter_content(chunk_size=8192):
+                content += data
```
alex: Don't add string in a loop like ths, it's O(n^2)

reaperhulk: Yeah I figured it was gonna be bad since it'll copy every time. I originally had this as an array and then it only gets concatenated at the end, but got lazy since I wanted the length. I'll refactor it quickly to be a separate length value.

alex: `io.BytesIO` is a good solution

On Sat, Nov 7, 2015 at 7:45 PM, Paul Kehrer <notifications@github.com>
wrote:

> In tasks.py
> <https://github.com/pyca/cryptography/pull/2469#discussion_r44219198>:
>
> >              )
> > +            content = b''
> > +            for data in response.iter_content(chunk_size=8192):
> > +                content += data
>
> Yeah I figured it was gonna be bad since it'll copy every time. I
> originally had this as an array and then it only gets concatenated at the
> end, but got lazy since I wanted the length. I'll refactor it quickly to be
> a separate length value.
>
> —
> <https://github.com/pyca/cryptography/pull/2469/files#r44219198>.
>



-- 

tasks.py
```diff
@@ -66,15 +69,28 @@ def download_artifacts(session):
...
+            bar = ProgressBar(
+                expected_size=int(response.headers["content-length"]),
+                filled_char="="
             )
+            content = io.BytesIO()
+            for data in response.iter_content(chunk_size=8192):
+                content.write(data)
+                bar.show(content.tell())
+                if bar.expected_size == content.tell():
+                    bar.done()
```
alex: Why not put the `bar.done()` outside the loop?

reaperhulk: No good reason at all. Probably no need to do the size check either assuming requests raises an error if iter_content fails before expected size. I'll just assert on it I guess.

PR #2470, opened 2015-11-08
[WIP] OpenSSL 1.1.0 support

reaperhulk:
1.1.0 alpha1 is coming soon (2015-12-10) with a public release in April 2016 so I'd like to start looking at how we're going to support it. The major change is the opaquing of nearly every struct. As a consequence we need to make our type declarations opaque as well and then use getter/setter functions as necessary.

Unfortunately, 1.1 adds many of these functions for the first time, so we need to declare them ourselves if OpenSSL < 1.1 (or is LibreSSL). To accomplish this I've copied the implementations directly from OpenSSL master. This has two drawbacks:

* If there is somehow a vulnerability in a getter/setter function we need to update our own version of it (and somehow track that this is necessary)
* The licensing for this is unclear. OpenSSL is moving to Apache 2 (https://www.openssl.org/blog/blog/2015/08/01/cla/) but for now it is still licensed under the "OpenSSL license", a BSD-like. Before landing anything we need clarity on what is needed to comply with licensing.

Until the entire conversion is complete OpenSSL 1.1 won't build (obviously). I will probably keep adding commits to this PR and then once it all works we can split it back apart for review. I'll also add a PR builder that uses openssl-master as we get close to successfully compiling.

This will break some downstream consumers (pyOpenSSL almost certainly), but I'm not sure how to avoid that. Does anyone have a different approach?

cc @alex @hynek @lukasa as you all might care about this.


src/_cffi_src/openssl/ssl.py
```diff
@@ -649,5 +634,69 @@
...
+/* Added in 1.0.1 but we need it in all versions now due to the great
+   opaquing. */
+#if OPENSSL_VERSION_NUMBER < 0x1000100fL
+/* from ssl/ssl_sess.c */
+int SSL_SESSION_set1_id_context(SSL_SESSION *s, const unsigned char *sid_ctx,
+                                unsigned int sid_ctx_len)
+{
+    if (sid_ctx_len > SSL_MAX_SID_CTX_LENGTH) {
+        SSLerr(SSL_F_SSL_SESSION_SET1_ID_CONTEXT,
+               SSL_R_SSL_SESSION_ID_CONTEXT_TOO_LONG);
```
alex: These constants don't exist on <1.1.0

hynek:
Looks like 1.1 will basically break everything so I wouldn’t hold my breath with its general adoption.

Maybe we’ll get a TLS layer in cryptographer faster? ;)

Lukasa:
If @hynek wants to avoid the work I can take the pyOpenSSL side of things. At least that way we can have a concrete idea of how painful it is in PyOpenSSL before we decide it's too much work. ;)

reaperhulk:
Shockingly this PR now builds properly against 1.1.0. The diff is even vaguely reasonable! We can discuss what to do with this once I'm back in the US. Boarding a flight now...

reaperhulk:
jenkins, retest this please

reaperhulk:
Closing in favor of #2826

PR #2475, opened 2015-11-12
these functions were added in 1.0.0, while CMS was added in 0.9.8h

reaperhulk:
We didn't catch this in our CI because all our 0.9.8 targets have CMS disabled or are older than 0.9.8h. Unlimited possible configurations...

### DIFF
--- a/src/_cffi_src/openssl/cms.py
+++ b/src/_cffi_src/openssl/cms.py
@@ -17,6 +17,7 @@ INCLUDES = """
 
 TYPES = """
 static const long Cryptography_HAS_CMS;
+static const long Cryptography_HAS_MORE_CMS;
 
 typedef ... CMS_ContentInfo;
 typedef ... CMS_SignerInfo;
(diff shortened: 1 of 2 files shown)


src/_cffi_src/openssl/cms.py
```diff
@@ -115,4 +116,13 @@
 CMS_SignerInfo *(*CMS_add1_signer)(CMS_ContentInfo *, X509 *, EVP_PKEY *,
                                    const EVP_MD *, unsigned int) = NULL;
 #endif
+#if !defined(OPENSSL_NO_CMS) && OPENSSL_VERSION_NUMBER < 0x1000000fL
```
alex: This could be nested inside the previous `ifdef`

alex: (Also the `const` could use a better name)

src/_cffi_src/openssl/cms.py
```diff
@@ -71,6 +72,15 @@
...
 #if !defined(OPENSSL_NO_CMS) && OPENSSL_VERSION_NUMBER >= 0x0090808fL
 static const long Cryptography_HAS_CMS = 1;
+#if OPENSSL_VERSION_NUMBER < 0x1000000fL
+static const long Cryptography_HAS_CMS_BIO_FUNCTIONS = 0;
+/* These functions were added in 1.0.0 */
+BIO *(*BIO_new_CMS)(BIO *, CMS_ContentInfo *) = NULL;
+int (*i2d_CMS_bio_stream)(BIO *, CMS_ContentInfo *, BIO *, int) = NULL;
+int (*PEM_write_bio_CMS_stream)(BIO *, CMS_ContentInfo *, BIO *, int) = NULL;
+#else
+static const long Cryptography_HAS_CMS_BIO_FUNCTIONS = 1;
```
alex: This needs to be defined in ths `HAS_CMS = 0` branch as well.

PR #2476, opened 2015-11-13
Add domain parameters as properties of elliptic curve objects

ronf:
This commit adds properties to each elliptic curve object which
hold the domain parameters associated with that curve. These
values are the finite field the curve is based on, two constants
which select a specific curve within that field, the x and y
coordinates of the base point of the curve, and the order of
the curve.

These values can be compared against externally provided domain
parameters to find a matching curve, such as when a PEM or DER
encoded EC key chooses to represent parameters explicitly rather
than referring to the curve by name.

### DIFF
--- a/docs/hazmat/primitives/asymmetric/ec.rst
+++ b/docs/hazmat/primitives/asymmetric/ec.rst
@@ -230,6 +230,13 @@ All named curves are providers of :class:`EllipticCurve`.
     SECG curve ``sect283k1``. Also called NIST K-283.
 
 
+.. class:: SECT239K1
+
+    .. versionadded:: 1.2
+
+    SECG curve ``sect239k1``.
+
+
 .. class:: SECT233K1
 
     .. versionadded:: 0.5
@@ -270,6 +277,13 @@ All named curves are providers of :class:`EllipticCurve`.
     .. versionadded:: 0.5
 
     SECG curve ``sect233r1``. Also called NIST B-233.
+
+
+.. class:: SECT163R1
+
+    .. versionadded:: 1.2
+
+    SECG curve ``sect163r1``.
 
 
 .. class:: SECT163R2
@@ -324,6 +338,48 @@ All named curves are providers of :class:`EllipticCurve`.
 Key Interfaces
 ~~~~~~~~~~~~~~
 
+.. class:: EllipticCurveField
+
+    .. versionadded:: 1.2
+
+    The field type associated with an elliptic curve.
+
+
+.. class:: EllipticCurvePrimeField
+
+    .. versionadded:: 1.2
+
+    A prime finite field where operations are done modulo a prime
+    number ``p``.
+
+    .. attribute:: p
+
+        :type: int
+
+        The prime modulus for this field.
+
+
+.. class:: EllipticCurveBinaryField
+
+    .. versionadded:: 1.2
+
+    A binary extension finite field using binary polynomials of order
+    ``m`` with operations done modulo an irreducible polynomial ``f``.
+
+    .. attribute:: m
+
+        :type: int
+
+        The order of the polynomial values in the field.
+
+    .. attribute:: f
+
+        :type: int list
+
+        The coefficients of the irreducible polynomial, excluding
+        ``m`` and ``1``.
+
+
 .. class:: EllipticCurve
 
     .. versionadded:: 0.5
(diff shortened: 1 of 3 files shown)


ronf:
Looks like I need additional unit tests for the eq/ne/hash/repr functions I added on the new EllipticCurveField classes. I'll try to get those in over the weekend and resubmit.

reaperhulk:
jenkins, add to whitelist

docs/hazmat/primitives/asymmetric/ec.rst
```diff
@@ -230,6 +230,13 @@ All named curves are providers of :class:`EllipticCurve`.
     SECG curve ``sect283k1``. Also called NIST K-283.
 
 
+.. class:: SECT239K1
+
+    .. versionadded:: 1.2
+
+    SECG curve ``sect239k1``.
```
reaperhulk: I assume you have some need for these curves? Do we have any source for test vectors for this and `sect163r1`?

ronf: I actually don't have a specific need for them and could remove them. However, they were listed in http://www.secg.org/sec2-v2.pdf along with all of the other curves already defined here and they were supported by OpenSSL, so I didn't really understand why they were excluded. I added them for completeness.

If you'd prefer to stick with the original list, I'm fine with dropping them.

reaperhulk: Yeah let's drop them for now. The reason we don't support them is simply that we don't have test vectors to confirm the underlying implementation. Without test vectors we don't want them to be listed as supported curves at this time.

ronf: Done.

ronf:
@reaperhulk, is there anything more you need from me on this? Anything you'd like me to change?

reaperhulk:
@ronf I still need to review it, sorry about the delay. I should be able to get to it this weekend.

reaperhulk:
@ronf I hate to ask you to do this, but would you mind splitting this PR into two parts? One with all the new constants/classes that hold them, and the other with the encode/decode point logic?

ronf:
@reaperhulk, The encode_point and decode_point functions are not actually new here. They were added in a previous check-in. I just moved the implementation from EllipticCurvePublicNumbers to EllipticCurve because I needed to use the encode function in the base_point() method, called by domain_parameters() to return all the new constants.

I chose to return the values as an encoded point rather than raw integers because that's how the parameters are encoded in PKCS#8 DER/PEM format EC public and private keys. The idea was to make it really easy to write a function that extracted data from these keys and matched on it, if you wanted to see if a certificate was using one of the well-known curves but was not encoded to refer to the curve by OID. I do this in AsyncSSH to support such certificates, since the SSH protocol requires that the curve name be provided.

I could split out the constants and support classes, but I would have to leave out the base_point() and domain_parameters() functions (or not use encoded points) if I didn't move the encode/decode calls.

reaperhulk:
Ah, I see. Yes, if you wouldn't mind splitting the constants and support classes I'd appreciate it. I'd like to get those constants landed as soon as we verify them (do you have a suggestion on how to do that by the way?).

ronf:
Ok - I've created a new pull request #2499 which has just the constants and support classes in it. I'll submit a new pull request with the remaining changes to add the base_point() and domain_parameters() functions once this has been merged in, since one change depends on the other.

Regarding verifying the constants, I did this myself before I did the original check-in by using the command line OpenSSL tool to create certificates with the "-param_enc explicit" option. This tells OpenSSL to encode the parameters explicitly in the certificate. I then used code I wrote for AsyncSSH to do ASN.1 parsing of the certificates to extract these parameters and pass them into my curve lookup function which compared them against the values returned by the domain_parameters() call in my original pull request and confirmed for each named curve that the values matched.

This will be a bit harder to do with only the constants in the new pull request, as the ASN.1 encoded domain parameters are encoded as point strings rather than integer values. However, if you decode the ASN.1 point strings into raw x & y values, it's possible to use this same basic approach.

ofek:
@ronf @reaperhulk @alex I strongly support exposing each curve's values for the mere reason we would not be able to support compressed public keys from [SEC 1 v2.0](http://www.secg.org/sec1-v2.pdf) section 2.3.3 otherwise. [https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicNumbers.from_encoded_point](url)

alex:
After a year plus of sitting on this, I'm going to close it. Protocols should already restrict themselves to less than "supporting every curve possible", which means you need logic for that -- further, with the move to curves with different structures like x25519, the API for this becomes further muddled. We'd rather people write the logic for their specific protocol than trying to express all of this.

ofek:
Haha coveralls.. brilliant

Issue #2477, opened 2015-11-13
ffi.callback() causes SELinux execmem violation


tiran:
The ffi.callback() wrapper uses libffi's closure API. Closures are build by generating and executing dynamic code at runtime. This is problematic because libffi creates memory mapped pages that are both writeable and executable at the same time. SELinux's memory protection prevents mmap() with PROT_WRITE | PROT_EXEC in order to mitigate buffer overflows and similar attack vectors.

The problem didn't show up earlier because cryptography hasn't used ffi.callback() much. In the past only the password callback made use of ffi.callback(). But recent versions of cryptography use ffi.callback() for CRYPTO thread locks and os.urandom engine. Now cryptography causes SELinux denials at import time. This causes trouble for applications like FreeIPA and OpenStack.
https://bugzilla.redhat.com/show_bug.cgi?id=1277224 https://bugzilla.redhat.com/show_bug.cgi?id=1249685

libffi has a new API that makes closures compatible with SELinux. The new API uses the same technologie as http://www.akkadia.org/drepper/selinux-mem.html with two file-backed MAP_SHARED memory mappings. Armin pointed out that the approach is problematic because MAP_SHARED mappings are also shared between parent process and all its forks. https://bitbucket.org/cffi/cffi/issues/231/writeable-memory-execution-execmem-with https://bitbucket.org/cffi/cffi/branch/ffi_closure_alloc

Armin and @glyph came up with a better solution https://bitbucket.org/cffi/cffi/issues/232/static-callbacks (WIP) The new solution is also faster. I will work on a PR when a version of cffi with the fix is available.

glyph:
Just for the record, while I understand the situation here, I think we should have an eye towards the fact that this just enables a "security" configuration that prohibits JITing.  One of the major advantages of using CFFI is that Cryptography JITs nicely with PyPy, and I think ecosystem-wise we should be pushing people to use PyPy as much as possible so that when they publish benchmarks of "python" they are looking at the fast one.

What does someone do in these environments if someone wants to run a node.js application, for example, where no pure-interpreter runtime is available?

PR #2463, opened 2015-11-01
Attempt to enable coverage on pypy

alex:
Fixes #785

PR #2481, opened 2015-11-18
Port callbacks to new static callback

tiran:
cffi 1.3.1 will introduce a new API to create static callbacks. Contrary to
the old callback API, static callbacks no longer depend on libffi's
dynamic code generation for closures. Static code has some benefits over
dynamic generation. For example the code is faster. Also it doesn't need
writeable and executable memory mappings, which makes it compatible with
SELinux's deny execmem policy.

https://bitbucket.org/cffi/cffi/issues/232/static-callbacks

Closes: #2477

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,8 +54,8 @@ if platform.python_implementation() == "PyPy":
             "upgrade PyPy to use this library."
         )
 else:
-    requirements.append("cffi>=1.1.0")
-    setup_requirements.append("cffi>=1.1.0")
+    requirements.append("cffi>=1.4.1")
+    setup_requirements.append("cffi>=1.4.1")
 
 # If you add a new dep here you probably need to add it in the tox.ini as well
 test_requirements = [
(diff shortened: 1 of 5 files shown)


reaperhulk:
The last time we upgraded our cffi requirement we encountered two things we'll want to keep in mind here:

* We need a PyPy release with the same code before we can use the new code path.
* If we move up our dep we kill our support for all older PyPys. Previously we considered this acceptable, but it was significantly more painful than expected (including totally breaking people using pypy on travis). Is there some way we can support both old-style and new-style callbacks simultaneously?

alex:
Please use an object, rather than a dict for userdata.

alex:
And this is now ready for a rebase!

reaperhulk:
Partial coverage is because we don't run coverage on pypy. We can solve this either by adding a job where we pin cffi to < 1.4 or merging #2463 despite its effect on our CI speed.

reaperhulk:
We've merged the pypy coverage fix, but there's a crash occurring on the pypy builder with 0.9.8 https://travis-ci.org/pyca/cryptography/jobs/100522841

PR #2484, opened 2015-11-18
backport openssl error msg

reaperhulk:
Include the full OpenSSL error in the exception message

This time with 100% more correct branch.


alex:
@reaperhulk uhh, ok, what's up with the coverage results?

reaperhulk:
codecov doesn't appear to have any results for the OS X builders. Looking at a random OS X travis builder it shows:  http://codecov.io/github/pyca/cryptography?ref=7731745acb27834205349d511857fde61854529e as the URL, which corresponds to build 7451...

Oh I know what happened here. I submitted this exact same PR but against master, then immediately closed it and submitted this one. Codecov appears to have gotten confused.

reaperhulk:
Or travis has. Let's see what happens when build 7452 is complete. That's the build that should be associated with this PR.

PR #2488, opened 2015-11-19
Change password callback to use userdata pointer

tiran:
Instead of a closure the pem_password_cb now uses the void *userdata
argument to exchange data with the callback function. It's a necessary
step to port all callbacks to new static callbacks.

See: #2477

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -40,6 +40,7 @@ from cryptography.hazmat.backends.openssl.x509 import (
     _Certificate, _CertificateRevocationList, _CertificateSigningRequest,
     _DISTPOINT_TYPE_FULLNAME, _DISTPOINT_TYPE_RELATIVENAME
 )
+from cryptography.hazmat.bindings._openssl import ffi as _ffi
 from cryptography.hazmat.bindings.openssl import binding
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import dsa, ec, rsa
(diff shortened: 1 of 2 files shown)


reaperhulk:
Is this ready for review @tiran?

tiran:
@reaperhulk yes, the patch is ready for review.

reaperhulk:
This LGTM as an intermediate step, although I'd like to see what the next step is here. Is that #2481 or do you have a different approach in mind now that you've done this?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -588,6 +588,35 @@ def _encode_crl_distribution_points(backend, crl_distribution_points):
...
+class _PasswordUserdata(object):
+    def __init__(self, backend, password):
+        self.backend = backend
+        self.password = password
+        self.called = 0
+        self.exception = None
+
+
+def _pem_password_cb(buf, size, writing, userdata_handle):
+    ud = backend._ffi.from_handle(userdata_handle)
```
alex: @reaperhulk interested in your thoughts here, IMO this shouldn't use the global `backend`, but instead it should directly use the global `ffi`?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -588,6 +588,35 @@ def _encode_crl_distribution_points(backend, crl_distribution_points):
...
+    ud.called += 1
+
+    if not ud.password:
+        ud.exception = TypeError(
+            "Password was not given but private key is encrypted."
+        )
+        return -1
+    elif len(ud.password) < size:
+        pw_buf = ud.backend._ffi.buffer(buf, size)
```
alex: same here.

tiran:
@reaperhulk  PR #2481 depends on this PR. The new approach for the password cb doesn't depend on new cffi features. It can be merged right now. On the other hand static callbacks require a different approach to store context specific information.

@alex from_handle() is the only place I can't access the proper backend. Any ffi instance will do. Which one do you want me to use?

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: Yeah we should just do `from cryptography.hazmat.bindings._openssl import ffi as _ffi` or something so that it's available in this scope.

reaperhulk:
@tiran Let's do `from cryptography.hazmat.bindings._openssl import ffi as _ffi` and then you can use that ffi.

tiran:
@reaperhulk I'll update the PR today.

alex:
@tiran if you can resolve the pep8 errors here we can do a final review

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -625,6 +626,35 @@ def _encode_general_subtree(backend, subtrees):
 }
 
 
+class _PasswordUserdata(object):
+    def __init__(self, backend, password):
+        self.backend = backend
```
alex: This is now unused.

tiran: I left it in in case it is useful in the future. It's gone now.

alex:
Looks good to me now!

PR #2489, opened 2015-11-19
add tbsCertList and signature interfaces to CRLs

etrauschke:
This is basically an extension of #2387 for CRLs.
Let me know if you want me to break this up but it seems fairly small.

### DIFF
--- a/src/_cffi_src/openssl/x509.py
+++ b/src/_cffi_src/openssl/x509.py
@@ -65,6 +65,7 @@ typedef struct {
 typedef struct {
     X509_CRL_INFO *crl;
     X509_ALGOR *sig_alg;
+    ASN1_BIT_STRING *signature;
     ...;
 } X509_CRL;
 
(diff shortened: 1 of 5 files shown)


src/_cffi_src/openssl/x509.py
```diff
@@ -65,6 +65,7 @@
 typedef struct {
     X509_CRL_INFO *crl;
     X509_ALGOR *sig_alg;
+    ASN1_BIT_STRING *signature;
```
reaperhulk: I don't think this should block, but we'll need to alter this for 1.1.0 using the new function `X509_CRL_get0_signature`. I'll deal with that in the 1.1.0 WIP PR though.

etrauschke: I didn't realize that. Thanks for taking care of it.

reaperhulk: Yeah, OpenSSL 1.1.0 is opaquing almost every struct. In most cases there are already accessors available, but for this (and the tbs payload) we'll have to define some functions since cffi won't know what the struct looks like.

PR #2495, opened 2015-11-28
Add support for 160 bit ARC4 keys

### LINKED ISSUE #2494: ARC4 should permit keys in range 40-2048 bits
ARC4 in this library is currently too restrictive on the key size. It should allow any key 5 to 256 bytes in length.

eakret:
This change adds support for 160 bit keys to the ARC4 cipher implementation. It also adds files for generating and verifying ARC4 test vectors for key lengths not included in RFC 6229, however only 160 bits is enabled at this time.

Closes #2494


alex:
Jenkins, ok to test.

alex:
Jenkins, retest this please.

reaperhulk:
This pretty much LGTM (great work!). One question: is there a specific reason you added the `msb` behavior in the vector generator?

eakret:
I added the `msb` bit so it would generate equivalent output to RFC 6229's test vectors if asked to use keys of lengths listed in RFC 6229. Key 1 in the RFC is an ascending byte counter starting at 1 but key 2 is a sha-256 hash from which they use the least significant bits.

reaperhulk:
Vectors confirmed, everything looks good. Thanks @eakret!

PR #2499, opened 2015-12-01
Add domain parameters as properties of elliptic curve objects (part 1)

ronf:
This commit is the first two submissions which add properties to each
    elliptic curve object holding the domain parameters associated with
    that curve. These values are the finite field the curve is based on,
    two constants which select a specific curve within that field, the
    x and y coordinates of the base point of the curve, and the order of
    the curve.
    
    This first commit adds the domain parameters as constants to each
    curve class, and a follow-up commit will add accessor functions to
    retrieve these values in a form easy to compare against externally
    provided domain parameters to find a matching curve, such as when a
    PEM or DER encoded EC key chooses to represent parameters explicitly
    rather than referring to the curve by name.


ronf:
I'm not sure what the docs failure here is -- it doesn't seem to be related to the changes I submitted.

reaperhulk:
It appears sphinx got updated to 1.3.2 and that is causing us some problems. Looks text encoding related.

reaperhulk:
#2500 should resolve this issue for now.

ronf:
jenkins, retest this please

src/cryptography/hazmat/primitives/asymmetric/ec.py
```diff
@@ -11,6 +11,55 @@
...
+
+
+class EllipticCurvePrimeField(EllipticCurveField):
+    def __init__(self, p):
+        self._p = p
+
+    p = utils.read_only_property("_p")
+
+    def __eq__(self, other):
+        return isinstance(other, self.__class__) and self.p == other.p
```
reaperhulk: Our typical pattern for `__eq__` is to do 

```python
        if not isinstance(other, thisclass):
            return NotImplemented
```

as that allows the other class to potentially implement comparison. Could you make that change for these `__eq__` methods (and add a quick `!= object()` in the tests to exercise that code path)?

reaperhulk:
Other than the one comment this LGTM.

src/cryptography/hazmat/primitives/asymmetric/ec.py
ronf: Done!

src/cryptography/hazmat/primitives/asymmetric/ec.py
```diff
@@ -11,6 +11,58 @@
 from cryptography import utils
 
 
+class EllipticCurveField(object):
+    """
+    Field type of an EllipticCurve
+    """
+
+
+class EllipticCurvePrimeField(EllipticCurveField):
```
alex: Why are we using inheritance here? In general we've avoided that for every other public API surface.

src/cryptography/hazmat/primitives/asymmetric/ec.py
```diff
@@ -109,99 +197,253 @@ def public_bytes(self, encoding, format):
 
 
 @utils.register_interface(EllipticCurve)
-class SECT571R1(object):
+class SECT571R1(EllipticCurve):
```
alex: same question here (and for all the other curves)

src/cryptography/hazmat/primitives/asymmetric/ec.py
```diff
@@ -109,99 +197,253 @@ def public_bytes(self, encoding, format):
...
+            "ffffffffffffffffffffffffe661ce18ff55987308059b18"
+            "6823851ec7dd9ca1161de93d5174d66e8382e9bb2fe84e47", 16)
 
 
 @utils.register_interface(EllipticCurve)
-class SECT409R1(object):
+class SECT409R1(EllipticCurve):
     name = "sect409r1"
     key_size = 409
+    field = EllipticCurveBinaryField(409, 87)
```
alex: these should use keyword arugments so it's clear what the different values are.

src/cryptography/hazmat/primitives/asymmetric/ec.py
reaperhulk: Looks like this should be a call to register_interface

ronf: The field() method on the EllipticCurve class returns an EllipticCurveField (either a prime field or a binary field). This parent class is intended to show that relationship. At the moment, there are no common methods that requires the child classes to inherit from it, but I think it's still good from a documentation standpoint to see the relationship here and have a type name I can include in the public docs for the return type of field().

For the curves, I changed them to inherit from EllipticCurve because in the full version of my check-in there are common methods now implemented on EllipticCurve for point encoding/decoding that all the curves share (moved out of EllipticCurvePublicNumbers because I need them for the base_point() method).

src/cryptography/hazmat/primitives/asymmetric/ec.py
ronf: I'm currently taking advantage of *args here to allow a variable number of polynomial exponents without requiring ugly extra parentheses and commas. A keyword version of this would end up looking like:

    field = EllipticCurveBinaryField(m=409, f=(87,))

I can do that, but I didn't think it really improved the readability.

ronf:
Ok - I've switched from inheritance to register_interface for now for both the EllipticCurveField subclasses and the EllipticCurve subclasses, and changed the places that construct EllipticCurveBinaryField objects to use keyword arguments.

I'm still unclear on how to avoid using inheritance when I add back in common functions that all the EllipticCurve subclasses need to share, though, such as the base_point() function and the whole reason for doing this, which was to add a domain_parameters() function to each of the curves that returns a hashable key of all the domain parameters for the class that could be used to look up a curve given its parameters.

ronf:
Any idea why the Jenkins build has started to fail here? It seems to be fine on all but one combination (CentOS 5, Python 2.6). I don't see any logs showing a specific error, though.

reaperhulk:
jenkins, retest this please

reaperhulk:
jenkins, retest this please

reaperhulk:
jenkins, retest this please

ronf:
Any other changes needed here?

reaperhulk:
@alex has talked to me a bit about this on IRC, so I'd like him to weigh in here.

alex:
Basically my concern is that there's a decent amount of new API surface, for very very few API consumers. Almost no protocols meaningfully support anything besides named curves, so most people will not need the domain parameters for anything. In addition storing these constants works perfectly fine outside of pyca/cryptography.

ronf:
While it's true that this lookup could be done externally, my reason for asking for this to be provided by the cryptography module is that I didn't like hardcoding constants in my application which were tied to algorithms I wasn't implementing there. It felt better to me to have these constants provided by the same package that provided the encryption code.

My reason for wanting to support something other than the named-curve case is that I have run across cases where someone has a key file which uses a well-known curve but the DER/PEM was written out with the explicit parameter values rather than a named curve OID. In those cases, the attempt to load the key would fail, even though the key is perfectly usable if only I had a way to determine which named curve those parameters corresponded to.

Ideally, there would be a way to query a crypto back-end like OpenSSL to get these constants rather than hard-coding them in cryptography's EC curve classes. That way, as OpenSSL added new curves, they would automatically be available without having to manually add these constants in cryptography. I know OpenSSL has these constants somewhere in it already as it can be used to construct DER/PEM keys which have the constants explicitly encoded in them, but I don't know if there's any way to query these constants using a public API today.

If the main concern here is the size of the API surface, that could be made much smaller. In terms of public API, all that's needed is a single function in the parent EllipticCurve class that returns a tuple of the domain parameters. The individual parameters wouldn't need to be directly accessible as properties, and even something like the base_point() function could be kept private, though I think there's still merit in moving the encode/decode point functions from EllipticCurvePublicNumbers to EllipticCurve. That allows conversions to be done without having to create an instance of an EC public key first.

reaperhulk:
I'm going to defer to @alex on this one. I don't feel strongly one way or another, but I'm adding this to the twelfth release milestone so we'll either accept or reject in the next week and a half :)

alex:
Switching over to the 13th release so we can take the time to make a good decision.

alex:
We've spent a lot of time discussing, and ultimately came down on the side of "no" -- I think this is too much API surface for too niche a usecase.

ronf:
For what it's worth, the API that I'm using in place of this being in PyCA is only two functions. One registers prime curve names and parameters and another does a lookup, returning the curve name given the parameters. With the PyCA version, we could have gotten this down to just a single top-level function if the parameters were constants in the curve classes. However, I can't really argue the point that this is a somewhat unusual requirement.

PR #2501, opened 2015-12-01
fix a warning in cffi

reaperhulk:
cffi doesn't want to guess the type, so we'll deopaque the enum and strip the values out of the lib if EC is unavailable. Fixes https://github.com/pyca/pyopenssl/issues/392

### DIFF
--- a/src/_cffi_src/openssl/ec.py
+++ b/src/_cffi_src/openssl/ec.py
@@ -30,9 +30,10 @@ typedef struct {
     const char *comment;
 } EC_builtin_curve;
 typedef enum {
-    POINT_CONVERSION_COMPRESSED = 2,
-    POINT_CONVERSION_UNCOMPRESSED = 4,
-    POINT_CONVERSION_HYBRID = 6
+    POINT_CONVERSION_COMPRESSED,
+    POINT_CONVERSION_UNCOMPRESSED,
+    POINT_CONVERSION_HYBRID,
+    ...
 } point_conversion_form_t;
 """
 
--- a/src/_cffi_src/openssl/ec.py
+++ b/src/_cffi_src/openssl/ec.py
@@ -213,7 +213,11 @@ typedef struct {
     int nid;
     const char *comment;
 } EC_builtin_curve;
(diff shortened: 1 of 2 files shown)


src/_cffi_src/openssl/ec.py
```diff
@@ -29,7 +29,11 @@
     int nid;
     const char *comment;
 } EC_builtin_curve;
-typedef enum { ... } point_conversion_form_t;
+typedef enum {
+    POINT_CONVERSION_COMPRESSED = 2,
+    POINT_CONVERSION_UNCOMPRESSED = 4,
+    POINT_CONVERSION_HYBRID = 6
```
alex: Do we need to declare the values of these and/or does cffi at least verify that we got them correct?

reaperhulk: Looks like we can drop the values and have cffi obtain them during the compile phase. Updating the PR.

reaperhulk:
This is r4r now.

src/_cffi_src/openssl/ec.py
```diff
@@ -208,7 +213,11 @@
     int nid;
     const char *comment;
 } EC_builtin_curve;
-typedef long point_conversion_form_t;
+typedef enum {
+    POINT_CONVERSION_COMPRESSED = 0,
+    POINT_CONVERSION_UNCOMPRESSED = 1,
+    POINT_CONVERSION_HYBRID = 2,
+} point_conversion_form_t;
```
alex: are the values necessary?

reaperhulk: Yep -- these aren't even the real values, but since this for the case where EC is not present we need to have the values present for cffi to extract and then we remove them from the final lib.

reaperhulk: Addendum: I guess if we don't put values the enum will be implicitly numbered by the compiler right? So maybe it would be okay to not put values in?

alex: Yes, enums have a default numbering

PR #2504, opened 2015-12-03
implement support for encoding name constraints

reaperhulk:
refs #2192

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,8 @@ Changelog
   :attr:`~cryptography.x509.Certificate.signature` and
   :attr:`~cryptography.x509.Certificate.tbs_certificate_bytes` attributes.
 * :class:`~cryptography.x509.NameConstraints` are now supported in the
-  :class:`~cryptography.x509.CertificateBuilder`.
+  :class:`~cryptography.x509.CertificateBuilder` and
+  :class:`~cryptography.x509.CertificateSigningRequestBuilder`.
 
 1.1.1 - 2015-11-19
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -58,6 +58,42 @@
 _MemoryBIO = collections.namedtuple("_MemoryBIO", ["bio", "char_ptr"])
 
 
+def _encode_name_constraints(backend, name_constraints):
```
alex: Can these be organized somewhere down bleow, with the rest of the extensions

reaperhulk: Sure. (This module is pretty unwieldy so we should probably move this stuff to another module, but that should be in another PR)

CHANGELOG.rst
```diff
@@ -9,6 +9,8 @@ Changelog
 * The :class:`~cryptography.x509.Certificate` class now has
   :attr:`~cryptography.x509.Certificate.signature` and
   :attr:`~cryptography.x509.Certificate.tbs_certificate_bytes` attributes.
+* :class:`~cryptography.x509.NameConstraints` are now supported in the
+  :class:`~cryptography.x509.CertificateBuilder`.
```
alex: CSR builder as well, I guess that's probably not practically relevant though?

reaperhulk: Some users might care (sometimes internal CAs allow issuance using extensions requested in the CSR), so I'll go ahead and mention CSRBuilder supports it too.

sigmavirus24:
I know nothing @mention-bot =P

reaperhulk:
Don't be so modest, it's touching code you had a hand in Ian! :grinning:

PR #2513, opened 2015-12-08
pin pytest to disallow 2.8.4 which breaks tests

alex:


alex:
Filed an issue about this with ptyest: https://github.com/pytest-dev/pytest/issues/1238

Will create an "unpin pytest" issue once this is resolved.

alex:
@stevepeak Is there a way to get visibility into why codecov hasn't reported a status here?

stevepeak:
@alex looks like it was a large queue of processing going on at this time. I've scaled up resources. Thank you :+1:

Issue #2503, opened 2015-12-02
SIGBUS on OS X


reaperhulk:
Replicating some info from https://github.com/pyca/pyopenssl/issues/391 into the correct issue tracker:

Cryptography binds functions that may or may not be available. Unfortunately, cffi provides us no method by which we can say "include this symbol if present, otherwise leave it out". Instead we need to call `cdef` to declare our types and functions, but we don't actually know if those functions exist until we invoke the compiler/linker. To handle this we developed a pattern whereby we use the C preprocessor to introspect our OpenSSL (via the headers) and, if symbols are missing, we declare them as null function pointers (or other values if they're not functions) and set a flag. During import of the Python module we then iterate over the set of conditional symbols and remove all attributes that aren't present from the `lib` object.

In this particular case we declared a flag `Cryptography_HAS_EC_NISTP_64_GCC_128` and then used it in [this conditional](https://github.com/pyca/cryptography/blob/master/src/_cffi_src/openssl/ec.py#L362-L371):

```cpp
#if defined(OPENSSL_NO_EC) || OPENSSL_VERSION_NUMBER < 0x1000100f || \
    defined(OPENSSL_NO_EC_NISTP_64_GCC_128)
static const long Cryptography_HAS_EC_NISTP_64_GCC_128 = 0;

const EC_METHOD *(*EC_GFp_nistp224_method)(void) = NULL;
const EC_METHOD *(*EC_GFp_nistp256_method)(void) = NULL;
const EC_METHOD *(*EC_GFp_nistp521_method)(void) = NULL;
#else
static const long Cryptography_HAS_EC_NISTP_64_GCC_128 = 1;
#endif
```

So far so good. The problem arises when building a universal library on OS X and then linking against the static version of it. To understand that we need a quick digression into what a universal library is on OS X...

Universal libraries are simply multiple CPU architectures crammed together in one file (typically using a tool called `lipo`). The linker simply looks at what arch it is linking for and uses those while ignoring the others. Universal libraries typically have i386 and x86_64 symbols in them these days, but just a few years ago you could find ppc, ppc64, i386, and x86_64 inside some libs.

[...]

PR #2515, opened 2015-12-09
remove the bindings for these x86_64 specific EC functions

reaperhulk:
We have no need to invoke them directly and their presence triggers a bug related to Fedora 23's hobbling of openssl EC functions (uugh)

This also fixes the SIGBUS issue in #2503, although that is more appropriately resolved via header fixes for universal libraries on OS X.


alex:
I am ok with this solution.

PR #2525, opened 2015-12-13
issue-2524

nbastin:
- Allows for any OID in access_method
- Validates OIDs at creation time
- Cleaned up all tests which used illegal OID values (other than those that should)
- Added a test for arbitrary access_method OIDs

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -239,7 +239,7 @@ class AuthorityInformationAccess(object):
 class AccessDescription(object):
     def __init__(self, access_method, access_location):
         if not isinstance(access_method, ObjectIdentifier):
-          raise TypeError("access_method must be an ObjectIdentifier")
+            raise TypeError("access_method must be an ObjectIdentifier")
 
         if not isinstance(access_location, GeneralName):
             raise TypeError("access_location must be a GeneralName")
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -238,11 +238,8 @@ class AuthorityInformationAccess(object):
 
 class AccessDescription(object):
     def __init__(self, access_method, access_location):
-        if not (access_method == AuthorityInformationAccessOID.OCSP or
-                access_method == AuthorityInformationAccessOID.CA_ISSUERS):
-            raise ValueError(
-                "access_method must be OID_OCSP or OID_CA_ISSUERS"
-            )
+        if not isinstance(access_method, ObjectIdentifier):
+          raise TypeError("access_method must be an ObjectIdentifier")
 
         if not isinstance(access_location, GeneralName):
             raise TypeError("access_location must be a GeneralName")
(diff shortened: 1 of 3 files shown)


alex:
Jenkins, ok to test.

reaperhulk:
Could you add a test verifying that we can successfully encode arbitrary OIDs using the CertificateBuilder? It should work fine with the current OpenSSL code, but a test will make sure we don't regress with other backends in the future.

Also, there are rules for OID values thanks to ASN.1 BER encoding, so we probably need to validate that. I believe the rules are that the second part cannot exceed 39 and the first must be {0, 1, 2}. If you attempt to encode values that violate this constraint in OpenSSL using `OBJ_txt2obj` (which is what we use under the covers to translate `ObjectIdentifier` when constructing certificate payloads) you'll get back a null pointer.

alex:
The ruels are
https://github.com/alex/rust-asn1/blob/master/src/utils.rs#L16-L18


On Sun, Dec 13, 2015 at 11:30 AM, Paul Kehrer <notifications@github.com>
wrote:

> Could you add a test verifying that we can successfully encode arbitrary
> OIDs using the CertificateBuilder? It should work fine with the current
> OpenSSL code, but a test will make sure we don't regress with other
> backends in the future.
>
> Also, there are rules for OID values thanks to ASN.1 BER encoding, so we
> probably need to validate that. I believe the rules are that the second
> part cannot exceed 39 and the first must be {0, 1, 2}. If you attempt to
> encode values that violate this constraint in OpenSSL using OBJ_txt2obj
> (which is what we use under the covers to translate ObjectIdentifier when
> constructing certificate payloads) you'll get back a null pointer.
>
> —
> <https://github.com/pyca/cryptography/pull/2525#issuecomment-164274214>.
>



-- 

nbastin:
@reaperhulk I'm not sure what test you're asking for re: CertificateBuilder, or at least how this would different from what TestObjectIdentifier already does (this is my first foray into the code, so perhaps there is something obvious here I simply don't understand).

Also, I wasn't trying to write a fully compliant ASN.1 parser, just adding some basic sanity check as previously you could make an OID of "foo" or "notanoid", or similar nonsense.  Perhaps really the library should pull in something like `pyasn1` rather than trying to reinvent the wheel across the board.

nbastin:
Test added as discussed on IRC.

reaperhulk:
Thanks for the test @nbastin. As @alex mentioned in IRC, would you mind opening a separate PR for the `__init__` change (plus adding the limited validation around encodable OIDs we mentioned earlier in this PR)? We can review/merge that and then rebase this change to be solely the expansion of support for arbitrary OIDs in AIA. Sorry about the extra work; we appreciate your efforts thus far!

nbastin:
Separate OID PR as 2530.

reaperhulk:
This can now be rebased/merged to resolve the conflicts. Also, @nbastin if you want to add yourself to `AUTHORS.rst` in a separate PR please feel free to do so!

PR #2528, opened 2015-12-14
require not_valid_after >= not_valid_before

### LINKED ISSUE #2496: CertificateBuilder should raise an error if not_valid_after is before not_valid_before

reaperhulk:
fixes #2496

### DIFF
--- a/src/cryptography/x509/base.py
+++ b/src/cryptography/x509/base.py
@@ -436,6 +436,11 @@ class CertificateBuilder(object):
         if time <= _UNIX_EPOCH:
             raise ValueError('The not valid before date must be after the unix'
                              ' epoch (1970 January 1).')
+        if self._not_valid_after is not None and time > self._not_valid_after:
+            raise ValueError(
+                'The not valid before date must be before the not valid after '
+                'date.'
+            )
         return CertificateBuilder(
             self._issuer_name, self._subject_name,
             self._public_key, self._serial_number, time,
@@ -453,6 +458,12 @@ class CertificateBuilder(object):
         if time <= _UNIX_EPOCH:
             raise ValueError('The not valid after date must be after the unix'
                              ' epoch (1970 January 1).')
+        if (self._not_valid_before is not None and
+                time < self._not_valid_before):
+            raise ValueError(
+                'The not valid after date must be after the not valid before '
+                'date.'
+            )
         return CertificateBuilder(
             self._issuer_name, self._subject_name,
             self._public_key, self._serial_number, self._not_valid_before,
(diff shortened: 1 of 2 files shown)


src/cryptography/x509/base.py
```diff
@@ -436,6 +436,11 @@ def not_valid_before(self, time):
         if time <= _UNIX_EPOCH:
             raise ValueError('The not valid before date must be after the unix'
                              ' epoch (1970 January 1).')
+        if self._not_valid_after is not None and time > self._not_valid_after:
```
alex: This allows them to be equal, do we care?

src/cryptography/x509/base.py
```diff
@@ -453,6 +458,12 @@ def not_valid_after(self, time):
         if time <= _UNIX_EPOCH:
             raise ValueError('The not valid after date must be after the unix'
                              ' epoch (1970 January 1).')
+        if (self._not_valid_before is not None and
+                time < self._not_valid_before):
```
alex: Same?

src/cryptography/x509/base.py
reaperhulk: Yeah, I decided equality is okay since a cert valid for 1s isn't out of spec (that's why the commit notes `>=`). The ValueError doesn't tell you that, but we could just say "the not valid before date must be <= the not valid after date" I suppose? I'm neutral on changing it, so if you want to let me know.

alex: It's fine, I just didn't know the spec off hand.

alex:
Jenkins, retest this please.

sigmavirus24:
A bit late, but :+1: .

etrauschke:
I know it's too late for this PR but in general: do we care that it is actually possible to create certificates which are not valid for testing purposes? I have seen this in several other areas, where cryptography is asserting something which OpenSSL actually allows even though it violates the standard.

The thing is that you can not assume that certificates with these issues are not used somewhere and so your software should be tested with invalid certs. Creating these is hard enough because most frameworks won't allow you to create them easily.

I understand that you need to find a balance so users don't shoot themselves in the foot but it would also be nice if there are still ways to create and parse certificates with invalid parameters.

reaperhulk:
@etrauschke I think there's a general desire for us to have a hazmat certificate builder that will allow us to do horrible things, so the likely long term here is that we build a hazmat builder and then have the recipe layer builder consume that with the additional restrictions on top.

etrauschke:
Yeah, that is what I was thinking. If you ever tried to create something even the openssl cli can't do, you usually start writing long C programs. Would be nice to be able to do this as part of cryptography.

PR #2529, opened 2015-12-14
Remove SSLv2 bindings.

### LINKED ISSUE #2527: Remove SSLv2 bindings
Nobody should use this, pyOpenSSL should already handle the case where it's unavailable since many OpenSSLs are compiled without it, let's just strip the whole thing.

Lukasa:
Resolves #2527. Not sure how well this will test.

BTW, I left in OP_NO_SSLv2 and OP_MSIE_SSLV2_RSA_PADDING. I did that because compiling out these bindings does not actually remove SSLv2 from OpenSSL, and so it still needs options that control its behaviour (or, you know, to disable it: I bloody hope no-one is setting OP_MSIE_SSLV2_RSA_PADDING).

I'll confirm that PyOpenSSL behaves appropriately under this patch.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -433,6 +433,10 @@ const long SSL_OP_LEGACY_SERVER_CONNECT = 0;
 static const long Cryptography_HAS_SECURE_RENEGOTIATION = 1;
 #endif
 
+/* Cryptography now compiles out all SSLv2 bindings. This exists to allow
+ * clients that use it to check for SSLv2 support to keep functioning as
+ * expected.
+ */
 static const long Cryptography_HAS_SSL2 = 0;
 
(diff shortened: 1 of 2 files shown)


src/_cffi_src/openssl/ssl.py
```diff
@@ -441,14 +432,8 @@
 #else
 static const long Cryptography_HAS_SECURE_RENEGOTIATION = 1;
 #endif
-#ifdef OPENSSL_NO_SSL2
+
 static const long Cryptography_HAS_SSL2 = 0;
```
reaperhulk: Could you put a comment that this exists for clients that may still check for SSLv2 support? We should probably figure out what a deprecation/removal policy is for things like this, but on the other hand the maintenance burden is essentially zero.

Lukasa: Done.

PR #2530, opened 2015-12-14
OID validation

nbastin:
Add OID validation based on ASN.1 rules, with tests and test modifications for those already using invalid OIDs.

### DIFF
--- a/src/cryptography/x509/oid.py
+++ b/src/cryptography/x509/oid.py
@@ -26,6 +26,11 @@ class ObjectIdentifier(object):
                     "Malformed OID: %s (non-integer nodes)" % (
                         self._dotted_string))
 
+        if len(nodes) < 2:
+            raise ValueError(
+                "Malformed OID: %s (insufficient number of nodes)" % (
+                    self._dotted_string)
+
         if intnodes[0] > 2:
             raise ValueError(
                 "Malformed OID: %s (first node outside valid range)" % (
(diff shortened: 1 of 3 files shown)


alex:
Jenkins, ok to test

reaperhulk:
Looks like `__init__` will raise an `IndexError` if you pass "1" as your dotted string. Do we care? Also, do we want to add checks for negative numbers? We're starting to get into more rigorous validation here so I'm not sure what we want to do. Opinion @alex?

alex:
I'd say yes to handling "1", no to negative numbers -- validation is mostly
a convenience for users, not a correctness thing. Open to being convinced
otherwise.

On Wed, Dec 16, 2015 at 11:12 PM, Paul Kehrer <notifications@github.com>
wrote:

> Looks like __init__ will raise an IndexError if you pass "1" as your
> dotted string. Do we care? Also, do we want to add checks for negative
> numbers? We're starting to get into more rigorous validation here so I'm
> not sure what we want to do. Opinion @alex <https://github.com/alex>?
>
> —
> <https://github.com/pyca/cryptography/pull/2530#issuecomment-165332521>.
>



-- 

alex:
Jenkins, retest this please.

ronf:
In the ASN.1 code I wrote for AsyncSSH, I treated the following as errors:

- Non-integer component numbers
- Negative component numbers
- OIDs with less than two components
- First component number greater than 2
- First component number less than 2 and second component number greater than 39

The last couple of checks are important because otherwise you can end up with an OID that decodes to a different value than what was provided. For instance, if you allow "0.80" or "1.40", that will get converted into "2.0" when decoded.

reaperhulk:
We've got the first 4 out of the 5 checks you list in this PR (the only thing missing is negative). I am okay with that for now. Thanks @nbastin!

PR #2537, opened 2015-12-21
[WIP] cffi init_once (1.4+) support in the openssl backend

reaperhulk:
cffi added `init_once` support in version 1.4. This can alleviate a race for us around the registration of the osrandom engine, but since we don't want to bump our PyPy minimum version beyond 2.6 we need to support both code paths. I am far from confident that the current PR represents the right approach, but it should serve as a starting point to figure out what we want to do.

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/binding.py
+++ b/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -9,6 +9,8 @@ import os
 import threading
 import types
 
+import cffi
+
 from cryptography.exceptions import InternalError
 from cryptography.hazmat.bindings._openssl import ffi, lib
 from cryptography.hazmat.bindings.openssl._conditional import CONDITIONAL_NAMES
@@ -16,6 +18,9 @@ from cryptography.hazmat.bindings.openssl._conditional import CONDITIONAL_NAMES
 
 _OpenSSLError = collections.namedtuple("_OpenSSLError",
                                        ["code", "lib", "func", "reason"])
+
+
+_init_once = True if cffi.__version__ >= '1.4.0' else False
 
 
 def _consume_errors(lib):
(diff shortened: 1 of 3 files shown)


src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -18,6 +20,9 @@
                                        ["code", "lib", "func", "reason"])
 
 
+_init_once = True if cffi.__version__ >= '1.4.0' else False
```
alex: Simplifies to `cffi.__version_info__ >= (1, 4, 0)`

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -174,9 +149,50 @@ def _lock_cb(cls, mode, n, file, line):
...
+
+# cffi 1.4+ has a new feature where we can guarantee a call only occurs once.
+# This is exactly what we want for initializing the OpenSSL library + adding
+# the osrandom engine since that is a shared global. However, we don't want
+# to reprise the cffi 1.0 transition where we broke PyPy < 2.6, so for now
+# we'll support both initialization methods. PyPy 4.1 should add init_once and
+# we can eventually require that as a minimum version.
+if _init_once:
+    ffi.init_once(_init_openssl, "init_openssl")
+    ffi.init_once(Binding.init_static_locks, "init_locks")
```
alex: Can we namespace these tags a bit more.

alex:
Can we do this without moving any of the code around? Ideally just altering the callsites of these functions?

src/cryptography/hazmat/bindings/openssl/binding.py
alex: e.g. `cryptography.`

alex:
Ooof, coverage is a mess here because PyPy is the one platform we don't have coverage on :-(

reaperhulk:
The coverage issue here is potentially resolved by merging #2463, although it will significantly increase our CI time. I don't really want to put this in the twelfth release though so we can wait on this one.

reaperhulk:
I'm unwilling to land a change like this at the moment. Mostly because until we have static callbacks sorted out I'm gunshy about making another major change. Kicking this out of the thirteenth release.

Issue #2288, opened 2015-08-21
Reading custom extensions


dainnilsson:
I need to be able to read the value of custom X509 extensions from a certificate. Currently, after loading the certificate using x509.load_pem_x509_certificate(), the extensions object is empty. As a reference, I want to be able to read out the `1.3.6.1.4.1.45724.2.1.1` extension from the following certificate:

```
-----BEGIN CERTIFICATE-----
MIICIjCCAQygAwIBAgIEIHHwozALBgkqhkiG9w0BAQswDzENMAsGA1UEAxMEdGVz
dDAeFw0xNTA4MTEwOTAwMzNaFw0xNjA4MTAwOTAwMzNaMCkxJzAlBgNVBAMTHll1
YmljbyBVMkYgRUUgU2VyaWFsIDU0NDMzODA4MzBZMBMGByqGSM49AgEGCCqGSM49
AwEHA0IABPdFG1pBjBBQVhLrD39Qg1vKjuR2kRdBZnwLI/zgzztQpf4ffpkrkB/3
E0TXj5zg8gN9sgMkX48geBe+tBEpvMmjOzA5MCIGCSsGAQQBgsQKAgQVMS4zLjYu
MS40LjEuNDE0ODIuMS4yMBMGCysGAQQBguUcAgEBBAQDAgQwMAsGCSqGSIb3DQEB
CwOCAQEAb3YpnmHHduNuWEXlLqlnww9034ZeZaojhPAYSLR8d5NPk9gc0hkjQKmI
aaBM7DsaHbcHMKpXoMGTQSC++NCZTcKvZ0Lt12mp5HRnM1NNBPol8Hte5fLmvW4t
Q9EzLl4gkz7LSlORxTuwTbae1eQqNdxdeB+0ilMFCEUc+3NGCNM0RWd+sP5+gzMX
BDQAI1Sc9XaPIg8t3du5JChAl1ifpu/uERZ2WQgtxeBDO6z1Xoa5qz4svf5oURjP
ZjxS0WUKht48Z2rIjk5lZzERSaY3RrX3UtrnZEIzCmInXOrcRPeAD4ZutpiwuHe6
2ABsjuMRnKbATbOUiLdknNyPYYQz2g==
-----END CERTIFICATE-----
```

reaperhulk:
[...]

PR #2543, opened 2015-12-22
support parsing CRL extensions in the OpenSSL backend

### LINKED ISSUE #2539: Support CRL extensions
Right now we support extensions on revoked certificates within the CRL, but ignore all top level CRL extensions. A list is available in section 5.2 of RFC 5280 (https://tools.ietf.org/html/rfc5280#section-5.2)

reaperhulk:
Unrecognized non-critical extensions will be ignored (the same as our current cert/csr parsing). Supporting arbitrary extensions more robustly is filed as #2288.

Fixes #2539


src/cryptography/x509/oid.py
```diff
@@ -234,6 +235,7 @@ class CertificatePoliciesOID(object):
     ExtensionOID.AUTHORITY_INFORMATION_ACCESS: "authorityInfoAccess",
     ExtensionOID.SUBJECT_INFORMATION_ACCESS: "subjectInfoAccess",
     ExtensionOID.OCSP_NO_CHECK: "OCSPNoCheck",
+    ExtensionOID.CRL_NUMBER: "CRLNumber",
```
alex: In oid-info the c is lower case?

alex:
One tiny comment. We did a good abstraction!

PR #2545, opened 2015-12-22
remove all mac 0.9.8 tests

### LINKED ISSUE #2544: Stop testing 0.9.8 system OpenSSL on OS X
We should remove this from our jenkins/travis configs. All our installation documentation references homebrew/macports openssl as a requirement, the vast majority of Mac users are getting a prebuilt binary wheel with statically linked OpenSSL, and of course 0.9.8 is EOL as of next week.

This will require an update to our docs to be more explicit (and to remove `OpenSSL 0.9.8-latest` from our list of tested configurations)

reaperhulk:
This will probably slow down the test infrastructure as we're running far more CommonCrypto tests than before. We should figure out what to do there before merging this.

Fixes #2544

### DIFF
--- a/.travis/run.sh
+++ b/.travis/run.sh
@@ -9,22 +9,14 @@ if [[ "$(uname -s)" == "Darwin" ]]; then
     PATH="$PYENV_ROOT/bin:$PATH"
     eval "$(pyenv init -)"
 
-    if [[ "${OPENSSL}" != "0.9.8" ]]; then
-        # set our flags to use homebrew openssl
-        export ARCHFLAGS="-arch x86_64"
-        # if the build is static we need different LDFLAGS
-        if [[ "${CRYPTOGRAPHY_OSX_NO_LINK_FLAGS}" == "1" ]]; then
-            export LDFLAGS="/usr/local/opt/openssl/lib/libssl.a /usr/local/opt/openssl/lib/libcrypto.a"
-        else
-            export LDFLAGS="-L/usr/local/opt/openssl/lib"
-        fi
-        export CFLAGS="-I/usr/local/opt/openssl/include"
-        # The Travis OS X jobs are run for two versions
-        # of OpenSSL, but we only need to run the
-        # CommonCrypto backend tests once. Exclude
-        # CommonCrypto when we test against brew OpenSSL
-        export TOX_FLAGS="--backend=openssl"
+    # set our flags to use homebrew openssl
+    # if the build is static we need different LDFLAGS
+    if [[ "${CRYPTOGRAPHY_OSX_NO_LINK_FLAGS}" == "1" ]]; then
+        export LDFLAGS="/usr/local/opt/openssl/lib/libssl.a /usr/local/opt/openssl/lib/libcrypto.a"
+    else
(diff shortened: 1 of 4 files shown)


.travis/run.sh
```diff
@@ -9,22 +9,14 @@ if [[ "$(uname -s)" == "Darwin" ]]; then
...
-        # CommonCrypto when we test against brew OpenSSL
-        export TOX_FLAGS="--backend=openssl"
+    # set our flags to use homebrew openssl
+    # if the build is static we need different LDFLAGS
+    if [[ "${CRYPTOGRAPHY_OSX_NO_LINK_FLAGS}" == "1" ]]; then
+        export LDFLAGS="/usr/local/opt/openssl/lib/libssl.a /usr/local/opt/openssl/lib/libcrypto.a"
+    else
+        export LDFLAGS="-L/usr/local/opt/openssl/lib"
     fi
+    export CFLAGS="-I/usr/local/opt/openssl/include"
```
Lukasa: I notice that `ARCHFLAGS` and `TOX_FLAGS` are gone here. Why?

Lukasa:
One note, otherwise LGTM.

.travis/run.sh
reaperhulk: ARCHFLAGS because we never needed it and it was cargo culted from the dark days of me not knowing what I was doing. TOX_FLAGS because we need to run the commoncrypto tests and previously we selectively switched them off unless we were testing system openssl. This was done to improve test runtime and not duplicate effort, but now we have no 0.9.8 builders so we need to run commoncrypto everywhere. I think it'd be sufficient to create a new COMMONCRYPTO variable and run the commoncrypto tests only when that's set, but I'm not sure everyone agrees with me :grinning:

Lukasa: That's fine. I suspected there were good reasons for these, I just didn't know what they were. In that case, this LGTM, but I can't merge so let's wait until someone with big green button privileges comes along.

alex:
Coverage; I don't fully understand the report.

reaperhulk:
Yeah that makes no sense...

alex:
This does appear to be a net time savings, FWIW.

alex:
@reaperhulk I can't think of a way to understand this, it appears to be a bug, should we ping Steve?

reaperhulk:
Yeah, I think so. @stevepeak, the combined diff we've got for this PR is exceptionally strange. Is there a bug we've uncovered?

stevepeak:
@reaperhulk not sure entirely all the data below lines up. Was there a rebase or something different in this pull request vs all the others?

- Base commit [1f32f2d](https://github.com/pyca/cryptography/commit/1f32f2d)
- Merge commit [c1eeb95](https://github.com/pyca/cryptography/commit/c1eeb9579252c355cdb795255c9305401bba7892) 
- Base Report: https://codecov.io/github/pyca/cryptography?ref=1f32f2d
- Head Report: https://codecov.io/github/pyca/cryptography?ref=c1eeb95

reaperhulk:
I don't recall rebasing this, but rebases do happen reasonably often in our PRs so it's possible it happened. We don't usually have a problem when we do that though. I'll rebase it now though to see if that resolves it.

PR #2552, opened 2015-12-23
CRLNumber needs to be a class for reasons

reaperhulk:


docs/x509/reference.rst
```diff
@@ -1695,6 +1695,27 @@ X.509 Extensions
 
         :type: int
 
+.. class:: CRLNumber(crl_number)
+
+    .. versionadded:: 1.2
+
+    The CRL number is a non-critical CRL extension that conveys a
```
alex: Err, what does non-critical mean in this context?

reaperhulk: That comes from the RFC, but it means it should not be marked as critical. We could just strip it though.

alex: I'd drop it, it kind of makes no sense.

docs/x509/reference.rst
```diff
@@ -1695,6 +1695,27 @@ X.509 Extensions
...
+    monotonically increasing sequence number for a given CRL scope and
+    CRL issuer. This extension allows users to easily determine when a
+    particular CRL supersedes another CRL. :rfc:`5280` requires that this
+    extension be present in conforming CRLs.
```
alex: So, it has to be included but should never be marked critical? Ugh, the IETF is teh worst.

reaperhulk: yuuup

alex:
besides those language tweeks this looks ok pending tests

src/cryptography/x509/extensions.py
```diff
@@ -109,6 +109,31 @@ def __repr__(self):
 
 
 @utils.register_interface(ExtensionType)
+class CRLNumber(object):
+    oid = ExtensionOID.CRL_NUMBER
+
+    def __init__(self, crl_number):
+        if not isinstance(crl_number, six.integer_types):
+            raise TypeError("crl_number must be an integer")
```
alex: This line is untested

reaperhulk:
Jenkins, retest this please

PR #2553, opened 2015-12-23
Build our own 0.9.8 on travis

reaperhulk:
This is an experiment to stop using the lucid hack and custom compile OpenSSL 0.9.8zh for coverage purposes on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ language: python
 cache:
     directories:
         - $HOME/.cache/pip
+        - $HOME/ossl-098
 
 matrix:
     include:
@@ -22,44 +23,14 @@ matrix:
           env: TOXENV=pypy
         - python: 2.6
           env: TOXENV=py26 OPENSSL=0.9.8
-          addons:
-              apt:
-                  sources:
-                      - lucid
-                  packages:
-                      - libssl-dev/lucid
         - python: 2.7
           env: TOXENV=py27 OPENSSL=0.9.8
-          addons:
-              apt:
-                  sources:
-                      - lucid
-                  packages:
-                      - libssl-dev/lucid
         - python: 3.3
           env: TOXENV=py33 OPENSSL=0.9.8
-          addons:
-              apt:
-                  sources:
-                      - lucid
-                  packages:
-                      - libssl-dev/lucid
         - python: 3.4
           env: TOXENV=py34 OPENSSL=0.9.8
-          addons:
-              apt:
-                  sources:
-                      - lucid
-                  packages:
-                      - libssl-dev/lucid
         - python: 3.5
           env: TOXENV=py35 OPENSSL=0.9.8
-          addons:
-              apt:
-                  sources:
-                      - lucid
-                  packages:
-                      - libssl-dev/lucid
         - python: pypy
           env: TOXENV=pypy OPENSSL=0.9.8
           addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: python
 cache:
     directories:
         - $HOME/.cache/pip
-        - $HOME/ossl-098
+        - $HOME/ossl-098l
 
 matrix:
     include:
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -66,7 +66,7 @@ else
       # This is 0.9.8l rather than zh because we have some branches for handling
       # < 0.9.8m that won't be exercised with a newer OpenSSL. (RHEL5 is 0.9.8e with
       # patches, but while that's in jenkins we don't get coverage data from it)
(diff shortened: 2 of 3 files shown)


alex:
This also removes several builds, can we make removing builds and switiching how we compile OpenSSL be two separate steps?

reaperhulk:
Coverage problems are due to being newer than 0.9.8m, when we have some branches based on that. I'll drop us to 0.9.8l I guess (yuck).

reaperhulk:
and 0.9.8l doesn't want to compile. going to let this one sit for a bit while I think about what to do...

.travis.yml
```diff
@@ -5,6 +5,7 @@ language: python
 cache:
     directories:
         - $HOME/.cache/pip
+        - $HOME/ossl-098
```
alex: Can we add the letter here, in the event we want to add more in the future I don't want untangling this to be a pain.

alex:
Besides that comment this looks ok to me, but I'd like another pair of eyes because this is a little wild.

Also, how long does compilation take? Keep in mind it will happen on the first push in any PR series.

reaperhulk:
It adds about 2-3 minutes to the runtime of the job within the build.

alex:
:+1:

.travis/install.sh
```diff
@@ -61,6 +61,28 @@ else
...
         pyenv global pypy-4.0.1
     fi
+    if [[ "${OPENSSL}" == "0.9.8" ]]; then
+      # download, compile, and install if it's not already present via travis cache
+      # This is 0.9.8l rather than zh because we have some branches for handling
+      # < 0.9.8m that won't be exercised with a newer OpenSSL. (RHEL5 is 0.9.8e with
+      # patches, but while that's in jenkins we don't get coverage data from it)
+      if [[ ! -f "$HOME/ossl-098l/bin/openssl" ]]; then
+        curl -O https://www.openssl.org/source/openssl-0.9.8l.tar.gz
+        tar zxvf openssl-0.9.8l.tar.gz
```
alex: Unless there's a compelling reason can we drop the `v`, it makes the logs hard to read.

.travis/install.sh
```diff
@@ -61,6 +61,28 @@ else
...
+        cd openssl-0.9.8l
+        echo "OPENSSL_0.9.8L_CUSTOM {
+            global:
+              *;
+        };" > openssl.ld
+        ./config no-asm no-ssl2 -Wl,--version-script=openssl.ld -Wl,-Bsymbolic-functions -fPIC shared --prefix=$HOME/ossl-098l
+        make depend
+        make install
```
alex: Should we be passing any `-j` here?

dstufft:
This change looks OK To me, though I agree with the dropping of ``v`` from the tar command.

.travis/install.sh
reaperhulk: we should be able to do a -j2 for a bit more speed

reaperhulk: I stand corrected, it looks like 0.9.8l does not properly have parallel build support enabled, so -j2 just results in a warning being logged that -j1 is used because the jobserver is unavailable.

alex: rofl, welp, ok

On Sat, Jan 2, 2016 at 2:47 PM, Paul Kehrer <notifications@github.com>
wrote:

> In .travis/install.sh
> <https://github.com/pyca/cryptography/pull/2553#discussion_r48686914>:
>
> > +    if [[ "${OPENSSL}" == "0.9.8" ]]; then
> > +      # download, compile, and install if it's not already present via travis cache
> > +      # This is 0.9.8l rather than zh because we have some branches for handling
> > +      # < 0.9.8m that won't be exercised with a newer OpenSSL. (RHEL5 is 0.9.8e with
> > +      # patches, but while that's in jenkins we don't get coverage data from it)
> > +      if [[ ! -f "$HOME/ossl-098l/bin/openssl" ]]; then
> > +        curl -O https://www.openssl.org/source/openssl-0.9.8l.tar.gz
> > +        tar zxvf openssl-0.9.8l.tar.gz
> > +        cd openssl-0.9.8l
> > +        echo "OPENSSL_0.9.8L_CUSTOM {
> > +            global:
> > +              *;
> > +        };" > openssl.ld
> > +        ./config no-asm no-ssl2 -Wl,--version-script=openssl.ld -Wl,-Bsymbolic-functions -fPIC shared --prefix=$HOME/ossl-098l
> > +        make depend
> > +        make install
>
> I stand corrected, it looks like 0.9.8l does not properly have parallel
> build support enabled, so -j2 just results in a warning being logged that
> -j1 is used because the jobserver is unavailable.
>
> —
> <https://github.com/pyca/cryptography/pull/2553/files#r48686914>.
>



-- 

PR #2558, opened 2015-12-24
fix a potential memory issue when retaining revoked certs from a CRL

### LINKED ISSUE #2557: CRL X509_REVOKED Memory Issue
In OpenSSl, `X509_CRL *` have a stack of `X509_REVOKED *`. In our parsing code we iterate over that using the `sk_X509_CRL_*` functions. However, `X509_REVOKED *` does not have a refcount, so it is freed when `X509_CRL_free` is called (which we register as a `gc` callback so it is called when the crl reference drops out of scope). Unfortunately, this is not safe. A user may choose to iterate the crl and save a `RevokedCertificate` instance, but this will point to garbage memory if the crl drops out of scope.

Ideally we could resolve this by incrementing the refcount, but since `X509_REVOKED *` doesn't have a refcount we're going to need to either retain the entire CRL struct for the lifetime of any sub-component (cue handwaving about how to do that) or else dupe every X509_REVOKED and pay the performance/memory penalty associated with that.

reaperhulk:
fixes #2557

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -747,8 +747,16 @@ def _decode_cert_issuer(backend, ext):
 
 @utils.register_interface(x509.RevokedCertificate)
 class _RevokedCertificate(object):
-    def __init__(self, backend, x509_revoked):
+    def __init__(self, backend, crl, x509_revoked):
         self._backend = backend
+        # The X509_REVOKED_value is a X509_REVOKED * that has
+        # no reference counting. This means when X509_CRL_free is
+        # called then the CRL and all X509_REVOKED * are freed. Since
+        # you can retain a reference to a single revoked certificate
+        # and let the CRL fall out of scope we need to retain a
+        # private reference to the CRL inside the RevokedCertificate
+        # object to prevent the gc from being called inappropriately.
+        self._crl = crl
         self._x509_revoked = x509_revoked
 
     @property
@@ -861,7 +869,10 @@ class _CertificateRevocationList(object):
             for i in range(num):
                 r = self._backend._lib.sk_X509_REVOKED_value(revoked, i)
                 self._backend.openssl_assert(r != self._backend._ffi.NULL)
-                revoked_list.append(_RevokedCertificate(self._backend, r))
+                revoked_certificate = _RevokedCertificate(
+                    self._backend, self, r
+                )
+                revoked_list.append(revoked_certificate)
 
         return revoked_list
 
--- a/tests/test_x509.py
+++ b/tests/test_x509.py
@@ -6,7 +6,6 @@ from __future__ import absolute_import, division, print_function
 
 import binascii
 import datetime
-import gc
 import ipaddress
 import os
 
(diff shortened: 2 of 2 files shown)


src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -861,7 +861,16 @@ def _revoked_certificates(self):
...
-                revoked_list.append(_RevokedCertificate(self._backend, r))
+                revoked_certificate = _RevokedCertificate(self._backend, r)
+                # The X509_REVOKED_value is a X509_REVOKED * that has
+                # no reference counting. This means when X509_CRL_free is
+                # called then the CRL and all X509_REVOKED * are freed. Since
+                # you can retain a reference to a single revoked certificate
+                # and let the CRL fall out of scope, we need to retain a
+                # private reference to the CRL inside the RevokedCertificate
+                # object to prevent the gc from being called inappropriately.
+                revoked_certificate._crl = self
```
alex: Please put this in the constructor, rather than hanging it off in random places.

reaperhulk: As I noted in the PR text, if we put it in the constructor then this becomes a part of the public interface despite being a backend specific implementation detail...

Except I just looked and __init__ isn't in the interface, so I can make that change without a problem. Maybe I should actually look at the interface before making a claim in the future... :stuck_out_tongue:

alex:
Is it possible to write a test for this, possibly one that segfaults on use-after-free without it?

etrauschke:
Thanks for catching this. The fix looks good to me.

tests/test_x509.py
```diff
@@ -173,6 +174,24 @@ def test_revoked_cert_retrieval(self, backend):
...
+        This test attempts to trigger the crash condition described in
+        https://github.com/pyca/cryptography/issues/2557
+        """
+        crl = _load_cert(
+            os.path.join("x509", "custom", "crl_all_reasons.pem"),
+            x509.load_pem_x509_crl,
+            backend
+        )
+        revoked = crl[11]
+        crl = "overwritten"
```
alex: IMO it's cleaner to just write `revoked = _load_cert(...)[11]` that way there's no confusing unused writes to locals.

tests/test_x509.py
```diff
@@ -173,6 +174,24 @@ def test_revoked_cert_retrieval(self, backend):
...
+        # force a gc collection to potentially X509_CRL_free if there are
+        # no references to the X509_CRL left.
+        gc.collect()
```
alex: Is this actually necessary? I feel like on CPython the refcounts work out so that it's unnecesary.

alex:
lgtm!

tests/test_x509.py
```diff
@@ -173,6 +173,20 @@ def test_revoked_cert_retrieval(self, backend):
...
+        """
+        revoked = _load_cert(
+            os.path.join("x509", "custom", "crl_all_reasons.pem"),
+            x509.load_pem_x509_crl,
+            backend
+        )[11]
+        assert revoked.revocation_date == datetime.datetime(2015, 1, 1, 0, 0)
+        assert revoked.serial_number == 11
+
     def test_extensions(self, backend):
```
etrauschke: Just curious, but does this test actually fail without the code changes?

alex: I tested it locally in a repl, it fails.

On Wed, Dec 23, 2015 at 9:22 PM, Erik Trauschke <notifications@github.com>
wrote:

> In tests/test_x509.py
> <https://github.com/pyca/cryptography/pull/2558#discussion_r48392416>:
>
> > @@ -173,6 +173,20 @@ def test_revoked_cert_retrieval(self, backend):
> >          # Check that len() works for CRLs.
> >          assert len(crl) == 12
> >
> > +    def test_revoked_cert_retrieval_retain_only_revoked(self, backend):
> > +        """
> > +        This test attempts to trigger the crash condition described in
> > +        https://github.com/pyca/cryptography/issues/2557
> > +        PyPy does gc at its own pace, so it will only be reliable on CPython.
> > +        """
> > +        revoked = _load_cert(
> > +            os.path.join("x509", "custom", "crl_all_reasons.pem"),
> > +            x509.load_pem_x509_crl,
> > +            backend
> > +        )[11]
> > +        assert revoked.revocation_date == datetime.datetime(2015, 1, 1, 0, 0)
> > +        assert revoked.serial_number == 11
> > +
> >      def test_extensions(self, backend):
>
> Just curious, but does this test actually fail without the code changes?
>
> —
> <https://github.com/pyca/cryptography/pull/2558/files#r48392416>.
>



-- 

PR #2561, opened 2015-12-24
Improve the performance of len(crl)

alex:


etrauschke:
While that probably helps, have you guys thought about bringing the object caches back to the x509 classes? I had them as part of the CRL backend code but was asked to take them out again.

Other than that, the code looks good to me even though I'm not sure if that buys you that much. Did you do some profiling on the code?

alex:
I haven't profiled, it takes it from O(n) -> O(1) though. It's a big win
particularly on large CRLs.

On Wed, Dec 23, 2015 at 11:37 PM, Erik Trauschke <notifications@github.com>
wrote:

> While that probably helps, have you guys thought about bringing the object
> caches back to the x509 classes? I had them as part of the CRL backend code
> but was asked to take them out again.
>
> Other than that, the code looks good to me even though I'm not sure if
> that buys you that much. Did you do some profiling on the code?
>
> —
> <https://github.com/pyca/cryptography/pull/2561#issuecomment-167040337>.
>



-- 

reaperhulk:
we may add some caching as it makes sense. We originally asked for them to be removed for consistency and to minimize the complexity of the PR :)

PR #2562, opened 2015-12-24
Add PEM_write_bio_DHparams binding

evilaliv3:
This pull request simply add PEM_write_bio_DHparams binding.

Adding this binding would be really valuable in order to make possible to generate DH parameters from the application and save them on a the filesystem using the PEM format.

### DIFF
--- a/src/_cffi_src/openssl/pem.py
+++ b/src/_cffi_src/openssl/pem.py
@@ -79,6 +79,7 @@ MACROS = """
 int PEM_write_bio_ECPrivateKey(BIO *, EC_KEY *, const EVP_CIPHER *,
                                unsigned char *, int, pem_password_cb *,
                                void *);
+int PEM_write_bio_DHparams(BIO *, DH *);
 """
 
 CUSTOMIZATIONS = """


alex:
Jenkins, ok to test.

evilaliv3:
what do you mean @alex? it's a not relevant error right?

alex:
@evilaliv3 that command was for Jenkins, to test it to run the tests. I haven't looked at why they're failing.

reaperhulk:
This is a consequence of the signature of this macro changing over time. In the most recent versions of OpenSSL it is declared with a const second argument, while older versions did not do this. To fix this move the declaration to the MACROS section (where we don't verify header declarations).

evilaliv3:
i see @reaperhulk, thanks.

going to do it and rebase the patch for a matter of clearness.

src/_cffi_src/openssl/pem.py
```diff
@@ -72,13 +72,13 @@
 int PEM_write_bio_RSAPublicKey(BIO *, const RSA *);
 
 EVP_PKEY *PEM_read_bio_PUBKEY(BIO *, EVP_PKEY **, pem_password_cb *, void *);
-int PEM_write_bio_PUBKEY(BIO *, EVP_PKEY *);
```
alex: Why was this removed?

alex:
Not sure what happened here, several extra commits appear to have been added in the rebase process.

src/_cffi_src/openssl/pem.py
evilaliv3: just error :) you are too fast!

i'm also rebasing against the current master

evilaliv3:
here we are @alex! the patch should be now correct; let's wait for the tests to finish!

PR #2559, opened 2015-12-24
CertificateRevocationListBuilder

reaperhulk:
This is the first of quite a few PRs.

Subsequent PRs will add revoked certificate support (yeah we can't add revoked certs in this PR), CRL extensions, CRL entry extensions (per revoked certificate extensions), and elliptic curve and DSA signing support.

PR #2565, opened 2015-12-24
add create_x509_crl interface

reaperhulk:
Since we enforce interfaces we need a dummy implementation in the OpenSSL backend to make this possible. Implementation remains in #2559.

### DIFF
--- a/docs/hazmat/backends/interfaces.rst
+++ b/docs/hazmat/backends/interfaces.rst
@@ -570,6 +570,26 @@ A specific ``backend`` may provide one or more of these interfaces.
         :returns: A new object with the
             :class:`~cryptography.x509.Certificate` interface.
 
+    .. method:: create_x509_crl(builder, private_key, algorithm)
+
+        .. versionadded:: 1.2
+
+        :param builder: An instance of
+            CertificateRevocationListBuilder.
+
+        :param private_key: The
+            :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKey`,
+            :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPrivateKey` or
+            :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePrivateKey`
+            that will be used to sign the CRL.
+
+        :param algorithm: The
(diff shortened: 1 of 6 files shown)


docs/hazmat/backends/interfaces.rst
```diff
@@ -570,6 +570,26 @@ A specific ``backend`` may provide one or more of these interfaces.
         :returns: A new object with the
             :class:`~cryptography.x509.Certificate` interface.
 
+    .. method:: create_x509_crl(builder, private_key, algorithm)
+
+        .. versionadded:: 1.2
+
+        :param builder: An instance of
+            CertificateRevocationListBuilder.
```
reaperhulk: This will become a link to the class in a future PR. Since the class doesn't exist yet it can't be linked :dancers:

alex: I assume this explodes if you linkify since the class doesn't exist?

docs/hazmat/backends/interfaces.rst
```diff
@@ -570,6 +570,26 @@ A specific ``backend`` may provide one or more of these interfaces.
...
+            :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPrivateKey` or
+            :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePrivateKey`
+            that will be used to sign the CRL.
+
+        :param algorithm: The
+            :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm`
+            that will be used to generate the CRL signature.
+
+        :returns: A new object with the
+            :class:`~cryptography.x509.CertificateRevocationList` interface.
```
alex: I think we're trying to get away from the "interface" nomenclature. Any reason not to "A new instance of CRL"?

reaperhulk: Nope, this is just an artifact of copy/paste on the other methods in this interface. We should probably put in a separate PR (or file an issue) to update the language on those. I'll update this one now.

alex:
I think you should just have it `raise NotImplementedError` and put a test in `test_opnessl.py`

Issue #2533, opened 2015-12-18
Add feature to create CRLs


joernheissler:
I'd like to be able to create CRLs.

`<reaperhulk> Wulf: that's a missing feature right now. A CertificateRevocationListBuilder needs to be implemented`

reaperhulk:
#2559 was the start of this

reaperhulk:
Since I need something to track outstanding work items:

- [ ] add_revoked_certificate support in CertificateRevocationListBuilder
- [ ] RevokedCertificateBuilder to build revoked certificates for the CRL builder (#2571)
- [ ] CRL Entry extension support in RevokedCertificateBuilder
- [ ] ECDSA signature support in CertificateRevocationListBuilder
- [ ] DSA signature support in CertificateRevocationListBuilder

PR #2568, opened 2015-12-25
add extension support to the CRLBuilder

reaperhulk:
refs #2559 and #2533

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -153,6 +153,17 @@ def _encode_name_gc(backend, attributes):
     return subject
 
 
+def _encode_crl_number(backend, crl_number):
+    asn1int = _encode_asn1_int_gc(backend, crl_number.crl_number)
+    pp = backend._ffi.new('unsigned char **')
+    r = backend._lib.i2d_ASN1_INTEGER(asn1int, pp)
+    backend.openssl_assert(r > 0)
+    pp = backend._ffi.gc(
+        pp, lambda pointer: backend._lib.OPENSSL_free(pointer[0])
+    )
+    return pp, r
+
+
 def _encode_certificate_policies(backend, certificate_policies):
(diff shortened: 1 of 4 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1510,7 +1530,27 @@ def create_x509_crl(self, builder, private_key, algorithm):
...
+            extension = self._lib.X509_EXTENSION_create_by_OBJ(
+                self._ffi.NULL,
+                obj,
+                1 if extension.critical else 0,
+                _encode_asn1_str_gc(self, pp[0], r)
+            )
+            self.openssl_assert(extension != self._ffi.NULL)
+            extension = self._ffi.gc(extension, self._lib.X509_EXTENSION_free)
+            res = self._lib.X509_CRL_add_ext(x509_crl, extension, i)
+            self.openssl_assert(res == 1)
```
alex: Ok, we already have this loop 2x, I hereby declare it to be time to write a function that does all the things the same way all the times.

alex: (That's me asking for a separate PR which refactors the existing two functions into a reusable thang)

reaperhulk: Hehe, I was going to do that as the next PR after this merged, but I can do it first :)

tests/test_x509_crlbuilder.py
```diff
@@ -144,6 +160,108 @@ def test_sign_empty_list(self, backend):
...
+        ).next_update(
+            next_update
+        ).add_extension(
+            extension, False
+        )
+
+        crl = builder.sign(private_key, hashes.SHA256(), backend)
+        assert len(crl) == 0
+        assert len(crl.extensions) == 1
+        ext = crl.extensions.get_extension_for_class(extension.__class__)
```
alex: Prefer `type(exetension)` here.

PR #2569, opened 2015-12-25
refactor x509 extension creation to make it a bit more reusable

reaperhulk:
Unfortunately X509 certs and CSRs add extensions differently, so we can't reuse quite as much as we'd like to...


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1524,6 +1501,30 @@ def create_x509_crl(self, builder, private_key, algorithm):
 
         return _CertificateRevocationList(self, x509_crl)
 
+    def _create_x509_extensions(self, extensions, handlers):
```
alex: Why not pass the ptr + function which adds to it, instead of returning a list.

reaperhulk: I'm not sure I follow. In the case of the CSR we need to create the extensions, push them onto a stack, then add that stack to the CSR. For the cert case we can just add the extensions directly to the X509 pointer. I could pass in the stack/x509 pointer and an add function, but the add functions have different arguments anyway.

alex: A sketch of what I'm describing:

```
handle_extensions(x509, extensions, X509_add_ext, handlers)

sk_extension = self._lib.sk_X509_EXTENSION_new_null()
handle_extensions(sk_extensions, extensions, lambda p, x, i: sk_X509_EXTENSION_push(p, x), handlers)
self._lib.X509_REQ_add_extensions(x509_req, sk_extension)

def handle_extensions(ptr, extensions, add_func, handler):
    for i, extension in enumerate(extensions):
        ...
        add_func(ptr, x509_extension, i)
```

PR #2573, opened 2015-12-25
remove deprecated methods

reaperhulk:
We forgot to do this when we started the twelfth release.

### DIFF
--- a/src/cryptography/hazmat/primitives/asymmetric/utils.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/utils.py
@@ -4,15 +4,11 @@
 
 from __future__ import absolute_import, division, print_function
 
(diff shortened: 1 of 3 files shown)


alex:
I think we were intentionally extending the deprecation period of these to _arbitrary future point_?

reaperhulk:
Hmm, I vaguely recall that now. I'll update the PR to just add a comment stating that we have no official timeline for removal of the DeprecatedIn10 functions.

reaperhulk:
Or I'll just put that in the other PR that adds some new deprecation since it's just a single comment.

alex:
Sounds good, you can also remove the usage of DeprecatedIn10 to just DeprecationWarning if you like?

PR #2574, opened 2015-12-25
rename CRLExtensionOID to CRLEntryExtensionOID

### LINKED ISSUE #2547: Rename CRLExtensionOID to CRLEntryExtensionOID
RFC 5280 labels the extension OIDs listed inside this enum as "CRL Entry" extensions. We should fix our nomenclature. I'm not sure what the deprecation cycle should be for this though. Do we want to follow the normal one or make it more extended?

reaperhulk:
Also document these OIDs.

fixes #2547

### DIFF
--- a/src/cryptography/x509/__init__.py
+++ b/src/cryptography/x509/__init__.py
@@ -29,9 +29,9 @@ from cryptography.x509.general_name import (
 )
 from cryptography.x509.name import Name, NameAttribute
 from cryptography.x509.oid import (
-    AuthorityInformationAccessOID, CRLExtensionOID, CertificatePoliciesOID,
-    ExtendedKeyUsageOID, ExtensionOID, NameOID, ObjectIdentifier,
-    SignatureAlgorithmOID, _SIG_OIDS_TO_HASH
+    AuthorityInformationAccessOID, CRLEntryExtensionOID,
+    CertificatePoliciesOID, ExtendedKeyUsageOID, ExtensionOID, NameOID,
+    ObjectIdentifier, SignatureAlgorithmOID, _SIG_OIDS_TO_HASH
 )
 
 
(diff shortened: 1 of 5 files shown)


src/cryptography/x509/__init__.py
```diff
@@ -29,9 +29,9 @@
 )
 from cryptography.x509.name import Name, NameAttribute
 from cryptography.x509.oid import (
-    AuthorityInformationAccessOID, CRLExtensionOID, CertificatePoliciesOID,
-    ExtendedKeyUsageOID, ExtensionOID, NameOID, ObjectIdentifier,
-    SignatureAlgorithmOID, _SIG_OIDS_TO_HASH
+    AuthorityInformationAccessOID, CRLEntryExtensionOID,
```
alex: Do we not need to keep exposing `CRLExtensionOID` here?

reaperhulk: We never documented it, but yeah we should. Fixing.

alex: Yeah, if we're going to go to the effort to deprecate it, we should make
sure it's undistrubed.

On Fri, Dec 25, 2015 at 7:11 PM, Paul Kehrer <notifications@github.com>
wrote:

> In src/cryptography/x509/__init__.py
> <https://github.com/pyca/cryptography/pull/2574#discussion_r48443894>:
>
> > @@ -29,9 +29,9 @@
> >  )
> >  from cryptography.x509.name import Name, NameAttribute
> >  from cryptography.x509.oid import (
> > -    AuthorityInformationAccessOID, CRLExtensionOID, CertificatePoliciesOID,
> > -    ExtendedKeyUsageOID, ExtensionOID, NameOID, ObjectIdentifier,
> > -    SignatureAlgorithmOID, _SIG_OIDS_TO_HASH
> > +    AuthorityInformationAccessOID, CRLEntryExtensionOID,
>
> We never documented it, but yeah we should. Fixing.
>
> —
> <https://github.com/pyca/cryptography/pull/2574/files#r48443894>.
>



-- 

PR #2576, opened 2015-12-26
make the Extensions class support indexing

reaperhulk:
No real reason not to support this, especially as we have a list internally.

I'm not clear if this is the right approach, so learn me if I'm doing something stupid :smile:

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -102,6 +102,9 @@ class Extensions(object):
     def __len__(self):
         return len(self._extensions)
 
+    def __getitem__(self, idx):
+        return self._extensions[idx]
+
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


src/cryptography/x509/extensions.py
```diff
@@ -102,6 +103,14 @@ def __iter__(self):
     def __len__(self):
         return len(self._extensions)
 
+    def __getitem__(self, idx):
+        if isinstance(idx, slice):
+            start, stop, step = idx.indices(len(self))
+            return self._extensions[start:stop:step]
+        else:
+            idx = operator.index(idx)
+            return self._extensions[idx]
```
alex: I think for this class just `return self._extensions[idx]` is all you need, you can drop all the other logic.

PR #2582, opened 2015-12-26
RevokedCertificateBuilder extensions

### LINKED ISSUE #2533: Add feature to create CRLs

reaperhulk:
refs and fixes #2533

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -53,7 +53,7 @@ from cryptography.hazmat.primitives.ciphers.algorithms import (
 from cryptography.hazmat.primitives.ciphers.modes import (
     CBC, CFB, CFB8, CTR, ECB, GCM, OFB
 )
-from cryptography.x509.oid import ExtensionOID, NameOID
+from cryptography.x509.oid import CRLEntryExtensionOID, ExtensionOID, NameOID
 
 
 _MemoryBIO = collections.namedtuple("_MemoryBIO", ["bio", "char_ptr"])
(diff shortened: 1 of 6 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -164,6 +164,68 @@ def _encode_crl_number(backend, crl_number):
...
+    x509.ReasonFlags.key_compromise: 1,
+    x509.ReasonFlags.ca_compromise: 2,
+    x509.ReasonFlags.affiliation_changed: 3,
+    x509.ReasonFlags.superseded: 4,
+    x509.ReasonFlags.cessation_of_operation: 5,
+    x509.ReasonFlags.certificate_hold: 6,
+    x509.ReasonFlags.remove_from_crl: 8,
+    x509.ReasonFlags.privilege_withdrawn: 9,
+    x509.ReasonFlags.aa_compromise: 10
+}
```
alex: I'm quite certain this dict already exists somewhere?

reaperhulk: There's the opposite (int to enum) in x509.py, and there's _CRLREASONFLAGS in backend.py, but the integer mappings are different (even though the names are the same) between how they're encoded in CRLDistributionPoints vs CRLReason inside a CRL entry.

alex: Can we at least put this text to to the int-to-enum one?

reaperhulk: For reference:

```

   CRLReason ::= ENUMERATED {
        unspecified             (0),
        keyCompromise           (1),
        cACompromise            (2),
        affiliationChanged      (3),
        superseded              (4),
        cessationOfOperation    (5),
        certificateHold         (6),
             -- value 7 is not used
        removeFromCRL           (8),
        privilegeWithdrawn      (9),
        aACompromise           (10) }
```

```
ReasonFlags ::= BIT STRING {
     unused                  (0),
     keyCompromise           (1),
     cACompromise            (2),
     affiliationChanged      (3),
     superseded              (4),
     cessationOfOperation    (5),
     certificateHold         (6),
     privilegeWithdrawn      (7),
     aACompromise            (8) }
```

We built a ReasonFlags enum that contains the union of those two, but then we map to the correct underlying bit value using these dicts.

alex: next*

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -694,13 +706,27 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+    x509.ReasonFlags.key_compromise: 1,
+    x509.ReasonFlags.ca_compromise: 2,
+    x509.ReasonFlags.affiliation_changed: 3,
+    x509.ReasonFlags.superseded: 4,
+    x509.ReasonFlags.cessation_of_operation: 5,
+    x509.ReasonFlags.certificate_hold: 6,
+    x509.ReasonFlags.remove_from_crl: 8,
+    x509.ReasonFlags.privilege_withdrawn: 9,
+    x509.ReasonFlags.aa_compromise: 10
+}
```
alex: Can we split just adding this dict into its own PR?

alex:
Ready for the next rebase :-)

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -164,6 +165,54 @@ def _encode_crl_number(backend, crl_number):
...
+    pp = backend._ffi.new('unsigned char **')
+    r = backend._lib.i2d_ASN1_GENERALIZEDTIME(time, pp)
+    backend.openssl_assert(r > 0)
+    pp = backend._ffi.gc(
+        pp, lambda pointer: backend._lib.OPENSSL_free(pointer[0])
+    )
+    return pp, r
+
+
+def _encode_certificate_issuer(backend, certificate_issuer):
```
alex: Is this different from `_encode_alt_name`?

reaperhulk: It is not. Removing.

PR #2589, opened 2015-12-27
add a comment and a dict we need in #2582

reaperhulk:


src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -679,7 +679,19 @@ def _decode_inhibit_any_policy(backend, asn1_int):
...
+#        keyCompromise           (1),
+#        cACompromise            (2),
+#        affiliationChanged      (3),
+#        superseded              (4),
+#        cessationOfOperation    (5),
+#        certificateHold         (6),
+#             -- value 7 is not used
+#        removeFromCRL           (8),
+#        privilegeWithdrawn      (9),
+#        aACompromise           (10) }
```
alex: FWIW, I don't feel strongly about having this comment, so if you don't want it I never intended to suggest adding it.

reaperhulk: We have a similar comment in another spot in this file. I think it's reasonably useful as (if nothing else) it lets you seek to where you need to be in RFC 5280 more quickly.

alex: :+1:

On Sat, Dec 26, 2015 at 8:13 PM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/2589#discussion_r48451014>:
>
> > @@ -679,7 +679,19 @@ def _decode_inhibit_any_policy(backend, asn1_int):
> >      return x509.InhibitAnyPolicy(skip_certs)
> >
> >
> > -_CRL_REASON_CODE_TO_ENUM = {
> > +#    CRLReason ::= ENUMERATED {
> > +#        unspecified             (0),
> > +#        keyCompromise           (1),
> > +#        cACompromise            (2),
> > +#        affiliationChanged      (3),
> > +#        superseded              (4),
> > +#        cessationOfOperation    (5),
> > +#        certificateHold         (6),
> > +#             -- value 7 is not used
> > +#        removeFromCRL           (8),
> > +#        privilegeWithdrawn      (9),
> > +#        aACompromise           (10) }
>
> We have a similar comment in another spot in this file. I think it's
> reasonably useful as (if nothing else) it lets you seek to where you need
> to be in RFC 5280 more quickly.
>
> —
> <https://github.com/pyca/cryptography/pull/2589/files#r48451014>.
>



-- 

Issue #2578, opened 2015-12-26
Implement __hash__ on all extensions we've implemented __eq__ on


reaperhulk:
We've fixed this haphazardly in the past, but we should probably consistently implement `__hash__` so extensions don't have odd behaviors like not being able to be inserted into hashable collections.

- [ ] CRLNumber
- [ ] AuthorityKeyIdentifier
- [ ] SubjectKeyIdentifier
- [ ] AuthorityInformationAccess
- [ ] AccessDescription
- [ ] BasicConstraints
- [ ] CRLDistributionPoints
- [ ] DistributionPoint
- [ ] CertificatePolicies
- [ ] PolicyInformation
- [ ] UserNotice
- [ ] NoticeReference
- [ ] ExtendedKeyUsage
- [ ] InhibitAnyPolicy
- [ ] KeyUsage
- [ ] NameConstraints
- [ ] Extension
- [ ] GeneralNames
- [ ] SubjectAlternativeName
- [ ] IssuerAlternativeName
- [ ] CRLReason
- [ ] InvalidityDate
- [ ] CertificateIssuer

reaperhulk:
Many of the remaining extensions use a list of items as their internal storage. Do we want to do something like hash(tuple(list))...? This could potentially be pretty expensive.

alex:
bleh. let's take some time and think for those, I don't think we have
external requests for these features yet, so we can take our time

On Sun, Dec 27, 2015 at 5:00 PM, Paul Kehrer <notifications@github.com>
wrote:

> Many of the remaining extensions use a list of items as their internal
> storage. Do we want to do something like hash(tuple(list))...? This could
> potentially be pretty expensive.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pyca/cryptography/issues/2578#issuecomment-167444021>.
>



-- 
"I disapprove of what you say, but I will defend to the death your right to
say it." -- Evelyn Beatrice Hall (summarizing Voltaire)
"The people's good is the highest law." -- Cicero
GPG Key fingerprint: 125F 5C67 DFE9 4084

Badg:
I'm going to preemptively apologize for an unintentionally long question, but: for those classes using lists for internal storage, is it a big deal if ```__hash__``` just reinstates ```object.__hash__```? 

[...]

PR #2594, opened 2015-12-27
implement hash on basicconstraints

reaperhulk:
If this approach looks good I'll probably do it on more than one extension for the next PR ;)

refs #2578

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -342,6 +342,9 @@ class BasicConstraints(object):
     def __ne__(self, other):
         return not self == other
 
+    def __hash__(self):
+        return hash((self.ca, self.path_length))
+
 
 @utils.register_interface(ExtensionType)
 class CRLDistributionPoints(object):
--- a/tests/test_x509_ext.py
+++ b/tests/test_x509_ext.py
@@ -837,6 +837,11 @@ class TestBasicConstraints(object):
             "<BasicConstraints(ca=True, path_length=None)>"
         )
 
+    def test_hash(self):
+        na = x509.BasicConstraints(ca=True, path_length=None)
+        na2 = x509.BasicConstraints(ca=True, path_length=None)
+        assert hash(na) == hash(na2)
+
     def test_eq(self):
         na = x509.BasicConstraints(ca=True, path_length=None)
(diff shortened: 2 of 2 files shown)


reaperhulk:
jenkins, retest this please

src/cryptography/x509/extensions.py
```diff
@@ -342,6 +342,9 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not self == other
 
+    def __hash__(self):
+        return hash(self.ca) ^ hash(self.path_length)
```
alex: So there's this (xoring each of the hashes), or `hash((self.ca, self.path_length))` (xoring a tuple of the items).

https://github.com/python/cpython/blob/master/Objects/tupleobject.c#L337-L359 is the algorithm for tuples, basically a xor + fancy stuff.

opinions?

reaperhulk: Seems like hashing the tuple and letting the underlying implementation xor is reasonable. I'll switch it.

alex: :+1:

On Sun, Dec 27, 2015 at 2:00 PM, Paul Kehrer <notifications@github.com>
wrote:

> In src/cryptography/x509/extensions.py
> <https://github.com/pyca/cryptography/pull/2594#discussion_r48457325>:
>
> > @@ -342,6 +342,9 @@ def __eq__(self, other):
> >      def __ne__(self, other):
> >          return not self == other
> >
> > +    def __hash__(self):
> > +        return hash(self.ca) ^ hash(self.path_length)
>
> Seems like hashing the tuple and letting the underlying implementation xor
> is reasonable. I'll switch it.
>
> —
> <https://github.com/pyca/cryptography/pull/2594/files#r48457325>.
>



-- 

tests/test_x509_ext.py
```diff
@@ -837,6 +837,11 @@ def test_repr(self):
             "<BasicConstraints(ca=True, path_length=None)>"
         )
 
+    def test_hash(self):
+        na = x509.BasicConstraints(ca=True, path_length=None)
+        na2 = x509.BasicConstraints(ca=True, path_length=None)
+        assert hash(na) == hash(na2)
```
alex: this should have a hash `!=` branch

PR #2599, opened 2015-12-28
Handle RSA_R_OAEP_DECODING_ERROR

reaperhulk:
Replaces #2323 

I've confirmed that this properly catches the error code on newer OpenSSL versions so I stole the branch, made it pass PEP8, and added a comment on the test to explain what it's for.

Big thanks to @EnTeQuAk for his work on this. Sorry it took so long to actually confirm it.

### DIFF
--- a/tests/hazmat/backends/test_openssl.py
+++ b/tests/hazmat/backends/test_openssl.py
@@ -30,9 +30,7 @@ from cryptography.hazmat.primitives.ciphers.algorithms import AES
 from cryptography.hazmat.primitives.ciphers.modes import CBC, CTR, Mode
 
 from ..primitives.fixtures_dsa import DSA_KEY_2048
-from ..primitives.fixtures_rsa import (
-    RSA_KEY_2048, RSA_KEY_512, RSA_KEY_512_ALT
-)
+from ..primitives.fixtures_rsa import RSA_KEY_2048, RSA_KEY_512
 from ..primitives.test_ec import _skip_curve_unsupported
 from ...utils import load_vectors_from_file, raises_unsupported_algorithm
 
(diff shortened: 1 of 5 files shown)


tests/hazmat/backends/test_openssl.py
```diff
@@ -456,6 +458,32 @@ def test_unsupported_oaep_label_decrypt(self):
                 )
             )
 
+    def test_invalid_oaep_decryption(self):
```
alex: Is there a reason not to put this test in test_rsa?

reaperhulk: Arguably this test exists to trigger a specific msg on the error stack in the openssl backend, but it definitely should raise the proper `ValueError` no matter the backend. We can move it (although I'd keep an altered version of the comment so we understand its value in the future) if you'd like.

alex: Yes, any future backend should also behave correctly here.

On Mon, Dec 28, 2015 at 12:08 AM, Paul Kehrer <notifications@github.com>
wrote:

> In tests/hazmat/backends/test_openssl.py
> <https://github.com/pyca/cryptography/pull/2599#discussion_r48463229>:
>
> > @@ -456,6 +458,32 @@ def test_unsupported_oaep_label_decrypt(self):
> >                  )
> >              )
> >
> > +    def test_invalid_oaep_decryption(self):
>
> Arguably this test exists to trigger a specific msg on the error stack in
> the openssl backend, but it definitely should raise the proper ValueError
> no matter the backend. We can move it (although I'd keep an altered version
> of the comment so we understand its value in the future) if you'd like.
>
> —
> <https://github.com/pyca/cryptography/pull/2599/files#r48463229>.
>



-- 

reaperhulk:
The ifdef check may be unneeded. It looks like RSA_R_OAEP_DECODING_ERROR is available on all versions of OpenSSL we care about.

PR #2602, opened 2015-12-30
Unrecognized extension support

### LINKED ISSUE #2288: Reading custom extensions

reaperhulk:
Unrecognized extensions are no longer ignored if they are non-critical -- instead we extract the DER encoded byte string and return it as the value of the Extension object.

This is being put up so I can get an opinion on whether this is a reasonable approach. I'll split out the test vector (and the `Extension` repr change) if we decide to go this route.

fixes #2288


alex:
I feel like there should be an "unkonwn extension" object, rather than just a byte string.

reaperhulk:
```
@utils.register_interface(ExtensionType)
class UnrecognizedExtension(object):
    def __init__(self, oid, value):
        if not isinstance(oid, ObjectIdentifier):
            raise TypeError("oid must be an ObjectIdentifier")
        self._oid = oid
        self._value = value

    oid = utils.read_only_property("_oid")
```

If we do it this way it might actually be possible to support encoding these extensions too...

alex:
That would be nice, a bit unfortunate that it requires duplicating the OID
on the read side.

On Wed, Dec 30, 2015 at 11:19 AM, Paul Kehrer <notifications@github.com>
wrote:

> @utils.register_interface(ExtensionType)
> class UnrecognizedExtension(object):
>     def __init__(self, oid, value):
>         if not isinstance(oid, ObjectIdentifier):
>             raise TypeError("oid must be an ObjectIdentifier")
>         self._oid = oid
>         self._value = value
>
>     oid = utils.read_only_property("_oid")
>     value = utils.read_only_property("_value")
>
> If we do it this way it might actually be possible to support encoding
> these extensions too...
>
> —
> <https://github.com/pyca/cryptography/pull/2602#issuecomment-168027220>.
>



-- 

PR #2604, opened 2015-12-30
add UnrecognizedExtension class

reaperhulk:
refs #2602 and #2288 

I'm calling it `UnrecognizedExtension` instead of `UnsupportedExtension` because the latter is an exception we're going to continue to raise for critical extensions we don't know how to parse. The language describing what this extension class is for could probably use some wordsmithing though.

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -1880,6 +1880,27 @@ X.509 Extensions
 
         :type: int
 
+.. class:: UnrecognizedExtension
+
+    .. versionadded:: 1.2
+
+    A generic extension class used to hold the raw value of **non-critical**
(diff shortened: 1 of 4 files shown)


docs/x509/reference.rst
```diff
@@ -1880,6 +1880,27 @@ X.509 Extensions
 
         :type: int
 
+.. class:: UnrecognizedExtension
+
+    .. versionadded:: 1.2
+
+    A generic extension class used to hold the raw value of **non-critical**
+    extensions that ``cryptography`` does not know how to parse. Extensions
+    marked critical will still raise
```
alex: No "still".

PR #2606, opened 2015-12-30
Fixed #2605 -- require setuptools 1.0

### LINKED ISSUE #2605: Set a minimum version on setuptools
Apparently it fails in hilarious ways with very very old setuptools (or even distribute). We should set a floor in `setup.py`.

@dstufft do you have opinions on what a reasonable floor would be?

alex:
Per @dstufft:

- 0.6.28 or something around there that supports wheels
- 0.7 is the version that introduced marker support in setuptools
- 1.0 is the version where setuptools started using and verifying tls

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ requirements = [
     "idna>=2.0",
     "pyasn1>=0.1.8",
     "six>=1.4.1",
-    "setuptools",
+    "setuptools>=1.0",
 ]
 setup_requirements = []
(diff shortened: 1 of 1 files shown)


dstufft:
I'm going to merge this. It says that coverage failed but if you go through and click through to the results, the results say it doesn't fail. It appears to be some sort of bug or missed update or something, but in any case it shouldn't affect this PR.

PR #2607, opened 2015-12-31
support unrecognized extensions in x509

reaperhulk:
One issue with this approach is that our helper method `get_extension_for_class` will return the first instance of UnrecognizedExtension and ignore all following instances. This is because we built that API on the assumption that the RFC 5280 prohibition on multiple extensions with the same OID had a 1:1 mapping with extension classes. We should decide what to do about that before merging.

replaces #2602 refs #2288

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -2435,7 +2435,8 @@ Exceptions
 
 .. class:: UnsupportedExtension
 
-    This is raised when a certificate contains an unsupported extension type.
+    This is raised when a certificate contains an unsupported extension type
+    that is marked critical.
 
     .. attribute:: oid
 
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -2436,7 +2436,7 @@ Exceptions
 .. class:: UnsupportedExtension
 
     This is raised when a certificate contains an unsupported extension type
-    that is marked critical.
+    that is marked ``critical``.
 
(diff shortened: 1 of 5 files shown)


alex:
`get_extension_for_class` should raise an exception on `UnrecognizedExtension`

docs/x509/reference.rst
```diff
@@ -2435,7 +2435,8 @@ Exceptions
 
 .. class:: UnsupportedExtension
 
-    This is raised when a certificate contains an unsupported extension type.
+    This is raised when a certificate contains an unsupported extension type
+    that is marked critical.
```
alex: Can you backtick `critical` here please?

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -213,6 +213,15 @@ def parse(self, backend, x509_obj):
                         "Critical extension {0} is not currently supported"
                         .format(oid), oid
                     )
+                else:
+                    # Dump the DER payload into an UnrecognizedExtension object
```
alex: This can't be implemented as a `handler` because `X509V3_EXT_d2i` returns `NULL` on rando oids I assume?

reaperhulk: Correct

alex:
Also this needs a CHANGELOG entry

PR #2613, opened 2015-12-31
deprecationwarning for OpenSSL < 1.0.1 as upstream has dropped support

### LINKED ISSUE #2510: Warn on OpenSSL 0.9.8?
Starting in 3.5 weeks OpenSSL 0.9.8 will officially be unsupported by the upstream team. It's unclear what this will mean for various downstreams (notable RHEL, CentOS, and OS X), but in practice it means there's likely to be a significantly decreased level of attention, research, and patching that goes into it.

I'd like to suggest that, starting with whatever release comes after January 1st, 2016, we emit a warning if users are linked against OpenSSL 0.9.8, suggesting they upgrade to a newer OpenSSL (or OS I guess?).

reaperhulk:
fixes #2510

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/binding.py
+++ b/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -8,6 +8,7 @@ import collections
 import os
 import threading
 import types
+import warnings
 
 from cryptography.exceptions import InternalError
 from cryptography.hazmat.bindings._openssl import ffi, lib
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -180,3 +181,12 @@ def _lock_cb(cls, mode, n, file, line):
 # condition registering the OpenSSL locks. On Python 3.4+ the import lock
 # is per module so this approach will not work.
 Binding.init_static_locks()
+
+if Binding.lib.SSLeay() < 0x10001000:
+    warnings.warn(
+        "OpenSSL versions less than 1.0.1 are no longer supported by the "
+        "OpenSSL project as of 2015-12-31. Your distribution/OS may provide "
+        "its own limited support, but we urge users to upgrade to a newer "
+        "OpenSSL as soon as possible.",
```
alex: I think we should go for a simpler message here, similar to what we do for Python 2.6, something that says, "Upstream has dropped support for this, it is no longer receiving security patches from them. Cryptography will drop support for this in an as-yet undetermined future release. Please upgrade" (but with cleaner language).

We should also add a CHANGELOG entry for increased visibility.

alex:
New test looks great, still needs a changelog entry.

alex:
Jenkins, retest this please.

PR #2614, opened 2016-01-01
add a small note about clearing the wheel cache

### LINKED ISSUE #2006: Compiling against different OpenSSL more complicated than in our docs
Since version 7, pip auto-wheels stuff caches it somewhere.  That usually *awesome*.

But it brings problems too.  So I’ve spent an hour yesterday trying desperately to recompile cryptography against my homebrew’s OpenSSL and I’ve failed.  In the end I `pip install`ed my git checkout.

I kept deleting stuff from my wheel house, tried to find caches but nothing.  It always pulled it out of nowhere and `-v` didn’t help either.  I’m gonna claim that if I’m too asinine to get this done, other people are likely too.

This morning I found it out with the help of dtruss (concretely with it complaining, that I don’t own `/Users/hynek/Library/Caches/pip/http` b/c dtruss runs as root.  So I killed all wheels there and got finally my juicy 1.0.2a.

We need to update our docs in this regard, this is madness.  Does pip have some `--force-recompile` option I’ve missed?  If not, that should be filed there too…

cc @dstufft

reaperhulk:
When compiling cryptography yourself if you have a problem and need to
recompile but the wheel built successfully you need to know to clear the
wheel cache. This is an edge case but worth documenting. Fixes #2006

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -58,6 +58,8 @@ to include the proper locations. For example:
     C:\> set INCLUDE=C:\OpenSSL-win64\include;%INCLUDE%
     C:\> pip install cryptography
 
+If you need to rebuild ``cryptography`` for any reason be sure to clear the
+local `wheel cache`_.
 
 .. _build-on-linux:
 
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -58,6 +58,8 @@ to include the proper locations. For example:
     C:\> set INCLUDE=C:\OpenSSL-win64\include;%INCLUDE%
     C:\> pip install cryptography
 
+If you need to rebuild ``cryptography`` for some reason be sure to clear the
```
alex: s/some/any/

PR #2617, opened 2016-01-01
don't add the NXCOMPAT and DYNAMICBASE flags if the compiler isn't msvc

### LINKED ISSUE #2339: Building on windows assumes MSVC
The result of issue #2021 causes `/NXCOMPAT /DYNAMICBASE` to be appended when building on windows, even if using mingw, resulting in a failed build.

reaperhulk:
fixes #2339 

I'm not a fan of this distutils code, but evidently it's the best we can do. And it'd be nice to not break mingw users.

### DIFF
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -7,7 +7,9 @@ from __future__ import absolute_import, division, print_function
 import os
 import sys
 
-from _cffi_src.utils import build_ffi_for_binding, extra_link_args
+from _cffi_src.utils import (
+    build_ffi_for_binding, compiler_type, extra_link_args
+)
 
 
 def _get_openssl_libraries(platform):
@@ -92,5 +94,5 @@ ffi = build_ffi_for_binding(
     pre_include=_OSX_PRE_INCLUDE,
     post_include=_OSX_POST_INCLUDE,
     libraries=_get_openssl_libraries(sys.platform),
-    extra_link_args=extra_link_args(sys.platform),
+    extra_link_args=extra_link_args(compiler_type()),
 )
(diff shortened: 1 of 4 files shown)


src/_cffi_src/utils.py
```diff
@@ -80,9 +82,18 @@ def build_ffi(module_name, cdef_source, verify_source, libraries=[],
 
 
 def extra_link_args(platform):
-    if platform != "win32":
+    if platform != "win32" or not _compiler_is_msvc():
```
Ayrx: Is there a scenario where the compiler is msvc but the platform isn't win32? Seems kinda redundant.

reaperhulk: Good point. I think I can refactor this significantly so `extra_link_args` takes the compiler_type rather than platform since that's what we actually care about here.

src/_cffi_src/build_openssl.py
```diff
@@ -92,5 +94,5 @@ def _osx_libraries(build_static):
     pre_include=_OSX_PRE_INCLUDE,
     post_include=_OSX_POST_INCLUDE,
     libraries=_get_openssl_libraries(sys.platform),
-    extra_link_args=extra_link_args(sys.platform),
+    extra_link_args=extra_link_args(compiler_type()),
```
alex: The `extra_link_args` call needs to be udpated in all the build files.

alex:
lgtm, will merge as soon as tests are green

alex:
@stevepeak Is codecov having an issue? The status doesn't appear to be being updated. (Happy New Year!)

stevepeak:
Hey @alex happy new years! We just had a classic large queue of jobs waiting. Scaling resources back up after the holidays :+1: 

PS we have a new major release in the mist, in the next month I would like to invite you to review it. Up for a look?

alex:
@stevepeak Thanks! And sure, we'd be happy to take a look!

PR #2620, opened 2016-01-01
@exarkun has requested that he be removed from mention-bot

reaperhulk:


exarkun:
:+1: , will merge when ci finishes tests and they're green.

PR #2627, opened 2016-01-02
Attempt to use coverage to quantify how many of the CONDITIONAL_NAMES…

alex:
… we really use


alex:
codecov's results were not as helpful as I'd hoped, I filed https://bitbucket.org/ned/coveragepy/issues/460/confusing-html-report-for-certain-partial with `coverage.py` to start with the issue, but at least some of the data appears to be there, so maybe I should file a bug with `codecov` as well?

PR #2625, opened 2016-01-02
full branch coverage by custom compiling 1.0.0

reaperhulk:
same approach as in #2553 for 0.9.8

fixes #2182

PR #2629, opened 2016-01-03
refactor travis custom openssl code to work with other versions

reaperhulk:
extracted from #2625 so it can be reviewed separately.

The big change here is modifying the SHLIB versioning and using rpath. Ensuring that the SHLIB is not the same as the version installed by the operating system (which is 1.0.0 for versions 1.0.0, 1.0.1, and 1.0.2 of OpenSSL) allows the dynamic linker to easily distinguish between the two versions. rpath then allows us to embed information on where the library we want to load is located so we don't have to interpose via LD_LIBRARY_PATH.

I tried using rpath alone, but it appears that even with LDFLAGS and CFLAGS if the SHLIB version is the same as what's available in /lib/x86_64-linux-gnu then it links that instead of what's in our custom dir.

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -69,7 +69,7 @@ else
         OPENSSL_DIR="ossl-098l"
     fi
     # download, compile, and install if it's not already present via travis cache
-    if [ ! -z "$OPENSSL_DIR" ]; then
+    if [ -n "$OPENSSL_DIR" ]; then
         if [[ ! -f "$HOME/$OPENSSL_DIR/bin/openssl" ]]; then
             curl -O https://www.openssl.org/source/openssl-$OPENSSL_VERSION_NUMBER.tar.gz
             tar zxf openssl-$OPENSSL_VERSION_NUMBER.tar.gz
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -62,26 +62,22 @@ else
         pyenv global pypy-4.0.1
     fi
     if [[ "${OPENSSL}" == "0.9.8" ]]; then
-      # download, compile, and install if it's not already present via travis cache
-      # This is 0.9.8l rather than zh because we have some branches for handling
-      # < 0.9.8m that won't be exercised with a newer OpenSSL. (RHEL5 is 0.9.8e with
-      # patches, but while that's in jenkins we don't get coverage data from it)
-      if [[ ! -f "$HOME/ossl-098l/bin/openssl" ]]; then
-        curl -O https://www.openssl.org/source/openssl-0.9.8l.tar.gz
-        tar zxf openssl-0.9.8l.tar.gz
-        cd openssl-0.9.8l
-        echo "OPENSSL_0.9.8L_CUSTOM {
-            global:
-              *;
-        };" > openssl.ld
-        ./config no-asm no-ssl2 -Wl,--version-script=openssl.ld -Wl,-Bsymbolic-functions -fPIC shared --prefix=$HOME/ossl-098l
-        make depend
-        make install
-      fi
-      export PATH="$HOME/ossl-098l/bin:$PATH"
-      export CFLAGS="-I$HOME/ossl-098l/include"
-      export LDFLAGS="-L$HOME/ossl-098l/lib"
-      export LD_LIBRARY_PATH="$HOME/ossl-098l/lib"
+        # We use 0.9.8l rather than zh because we have some branches for handling
+        # < 0.9.8m that won't be exercised with a newer OpenSSL. (RHEL5 is 0.9.8e with
+        # patches, but while that's in jenkins we don't get coverage data from it).
+        OPENSSL_VERSION_NUMBER="0.9.8l"
+        OPENSSL_DIR="ossl-098l"
+    fi
+    # download, compile, and install if it's not already present via travis cache
+    if [ ! -z "$OPENSSL_DIR" ]; then
+        if [[ ! -f "$HOME/$OPENSSL_DIR/bin/openssl" ]]; then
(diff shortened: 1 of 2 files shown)


.travis/install.sh
```diff
@@ -62,26 +62,22 @@ else
...
-      export LD_LIBRARY_PATH="$HOME/ossl-098l/lib"
+        # We use 0.9.8l rather than zh because we have some branches for handling
+        # < 0.9.8m that won't be exercised with a newer OpenSSL. (RHEL5 is 0.9.8e with
+        # patches, but while that's in jenkins we don't get coverage data from it).
+        OPENSSL_VERSION_NUMBER="0.9.8l"
+        OPENSSL_DIR="ossl-098l"
+    fi
+    # download, compile, and install if it's not already present via travis cache
+    if [ ! -z "$OPENSSL_DIR" ]; then
+        if [[ ! -f "$HOME/$OPENSSL_DIR/bin/openssl" ]]; then
```
alex: Is there a reason you added the `if [ ! -z "$OPENSSL_DIR" ]; then` check? instead of just the one for `bin/openssl`?

.travis/install.sh
```diff
@@ -62,26 +62,22 @@ else
...
+            curl -O https://www.openssl.org/source/openssl-$OPENSSL_VERSION_NUMBER.tar.gz
+            tar zxf openssl-$OPENSSL_VERSION_NUMBER.tar.gz
+            cd openssl-$OPENSSL_VERSION_NUMBER
+            ./config shared no-asm no-ssl2 -fPIC --prefix=$HOME/$OPENSSL_DIR
```
alex: Why'd we lose all the linker shenangins? (Specifically `--version-script` and `-Wl,-Bsymbol-functions`)

.travis/run.sh
```diff
@@ -24,10 +24,14 @@ else
...
     fi
     if [[ "${OPENSSL}" == "0.9.8" ]]; then
-      export PATH="$HOME/ossl-098l/bin:$PATH"
-      export CFLAGS="-I$HOME/ossl-098l/include"
-      export LDFLAGS="-L$HOME/ossl-098l/lib"
-      export LD_LIBRARY_PATH="$HOME/ossl-098l/lib"
+        OPENSSL_DIR="ossl-098l"
+    fi
+
+    if [ ! -z "$OPENSSL_DIR" ]; then
```
alex: Why was this branch added?

.travis/install.sh
reaperhulk: version-script wasn't actually doing anything useful unfortunately (verified with a whole lot of compiling over and over last night) and -Wl,-Bsymbolic-functions interacts poorly with -fPIC on OpenSSL 1.0.0t (for unclear reasons). We don't need it so I decided dropping it was the better part of valor.

alex: :+1:

On Sun, Jan 3, 2016 at 11:18 AM, Paul Kehrer <notifications@github.com>
wrote:

> In .travis/install.sh
> <https://github.com/pyca/cryptography/pull/2629#discussion_r48694475>:
>
> > -      export CFLAGS="-I$HOME/ossl-098l/include"
> > -      export LDFLAGS="-L$HOME/ossl-098l/lib"
> > -      export LD_LIBRARY_PATH="$HOME/ossl-098l/lib"
> > +        # We use 0.9.8l rather than zh because we have some branches for handling
> > +        # < 0.9.8m that won't be exercised with a newer OpenSSL. (RHEL5 is 0.9.8e with
> > +        # patches, but while that's in jenkins we don't get coverage data from it).
> > +        OPENSSL_VERSION_NUMBER="0.9.8l"
> > +        OPENSSL_DIR="ossl-098l"
> > +    fi
> > +    # download, compile, and install if it's not already present via travis cache
> > +    if [ ! -z "$OPENSSL_DIR" ]; then
> > +        if [[ ! -f "$HOME/$OPENSSL_DIR/bin/openssl" ]]; then
> > +            curl -O https://www.openssl.org/source/openssl-$OPENSSL_VERSION_NUMBER.tar.gz
> > +            tar zxf openssl-$OPENSSL_VERSION_NUMBER.tar.gz
> > +            cd openssl-$OPENSSL_VERSION_NUMBER
> > +            ./config shared no-asm no-ssl2 -fPIC --prefix=$HOME/$OPENSSL_DIR
>
> version-script wasn't actually doing anything useful unfortunately
> (verified with a whole lot of compiling over and over last night) and
> -Wl,-Bsymbolic-functions interacts poorly with -fPIC on OpenSSL 1.0.0t (for
> unclear reasons). We don't need it so I decided dropping it was the better
> part of valor.
>
> —
> <https://github.com/pyca/cryptography/pull/2629/files#r48694475>.
>



-- 

.travis/install.sh
reaperhulk: If we don't check to see if `OPENSSL_DIR` is set then it will check for `$HOME//bin/openssl` which will always return false. It will then try to download openssl, but `OPENSSL_VERSION_NUMBER` won't be set either so this will all fail.

The reason to separate them is so when we add 1.0.0 we can just add an `elif [[ "${OPENSSL}" == "1.0.0" ]]; then` to the top conditional and then share all the logic past that.

.travis/run.sh
reaperhulk: I didn't want to duplicate the variable exports if I didn't have to, so this conditional is so we can put an `elif [[ "${OPENSSL}" == "1.0.0" ]]; then` in the first if branch above and then share the logic.

reaperhulk:
jenkins, retest this please

alex:
One last nit should all the `! -z` be `-n`?

PR #2632, opened 2016-01-03
Quote a few things that shellcheck suggests

alex:


reaperhulk:
Whiny shellcheck, those are guaranteed to not have spaces in them! LGTM :+1:

PR #2635, opened 2016-01-03
Implement __hash__ on RFC822Name

alex:


reaperhulk:
I assume the intent is to implement __hash__ on all the generalname classes?

alex:
@reaperhulk yeah, it's a prereq to some of the other `__hash__` we want to implement.

PR #2637, opened 2016-01-03
Attempt to see where coverage is without 0.9.8

alex:
This will inform us on how much code we'd be able to delete, and give us a sense of how much burden we're actually assuming.


alex:
Closing this, will do a proper PR for dropping 0.9.8 when it's time.

PR #2638, opened 2016-01-04
Travis PyPy coverage

reaperhulk:
Evolution of #2463 to see if this is an acceptable solution.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -46,14 +46,6 @@ basepython = python2.7
 commands =
     sphinx-build -W -b linkcheck docs docs/_build/html
 
-# Temporarily disable coverage on pypy3 because of performance problems with
-# coverage.py on pypy3.
-[testenv:pypy3]
-commands =
-    pip list
-    python -c "from cryptography.hazmat.backends.openssl.backend import backend; print(backend.openssl_version_text())"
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -46,17 +46,9 @@ basepython = python2.7
...
-[testenv:pypy]
-commands =
-    pip list
-    python -c "from cryptography.hazmat.backends.openssl.backend import backend; print(backend.openssl_version_text())"
-    py.test --capture=no --strict {posargs}
-
-# Temporarily disable coverage on pypy3 because of performance problems with
-# coverage.py on pypy3.
-[testenv:pypy3]
+[testenv:pypy-nocoverage]
```
alex: This needs `basepython = pypy`, otherwise it's just running CPython I assume :-)

alex:
This looks like the right solution, emberassed I didn't htink of it.

tox.ini
reaperhulk: Fixed!

alex:
The bit in `upload_coverage` is broken, you can see if you expand the section in https://travis-ci.org/pyca/cryptography/jobs/100045029

alex:
oops.

PR #2639, opened 2016-01-05
Remove or rename readme

reaperhulk:
This PR will do one of the two. First let's try removing it.

Edit: Remove won't work. We need it on the Travis linux builders. Adding it back...

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -26,7 +26,7 @@ commands =
 deps =
     doc8
     pyenchant
-    readme>=0.5.1
+    readme_renderer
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -26,7 +26,7 @@ commands =
 deps =
     doc8
     pyenchant
-    readme>=0.5.1
+    readme_renderer>=0.5.1
```
alex: There's only an 0.7 release of this, so the condition doesn't make sense and can be dropped.

PR #2646, opened 2016-01-07
Static callbacks

reaperhulk:
Replaces #2481 (just to make travis throw away its weird cache on the previous PR)

### DIFF
--- a/src/_cffi_src/openssl/callbacks.py
+++ b/src/_cffi_src/openssl/callbacks.py
@@ -1,0 +1,111 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+from __future__ import absolute_import, division, print_function
+
+import cffi
+
+INCLUDES = """
+#include <openssl/ssl.h>
+#include <openssl/x509.h>
+#include <openssl/x509_vfy.h>
+"""
+
+TYPES = """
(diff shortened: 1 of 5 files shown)


reaperhulk:
This LGTM, but I'd like @alex to give it a review as well since this is a significant change.

src/_cffi_src/openssl/callbacks.py
```diff
@@ -0,0 +1,111 @@
...
+ * SSL_CTX_set_tlsext_servername_callback
+ */
+extern "Python" int Cryptography_tlsext_servername_cb(
+    const SSL *, int *, void *);
+
+/* x509_vfy.h
+ * int (*verify_cb)(int ok, X509_STORE_CTX *ctx)
+ */
+extern "Python" int Cryptography_verify_cb(int, X509_STORE_CTX *);
+"""
```
alex: There's a bunch of cb's included here that aren't used. These should not be provided for things we aren't using ourselves, because they _cannot_ be a public ok -- they can only ever have one consumer, and that has to be us.

tiran: Yes, you are totally correct. All static callbacks must not be used outside of cryptography and are reserved for internal use. I included the callbacks for future use in cryptography. I'm planning to work on a generic approach that can be used by e.g. PyOpenSSL or cryptography.tls.

I don't mind to remove or comment out unnecessary wrappers in the mean time. I can always cherry-pick them from a local patch.

PR #2647, opened 2016-01-07
opaque EVP_PKEY since EVP_PKEY_id exists

reaperhulk:
We should treat `EVP_PKEY` as fully opaque since we can get the type with `EVP_PKEY_id` anyway.


src/_cffi_src/openssl/evp.py
```diff
@@ -230,4 +229,8 @@
 EC_KEY *(*EVP_PKEY_get1_EC_KEY)(EVP_PKEY *) = NULL;
 int (*EVP_PKEY_set1_EC_KEY)(EVP_PKEY *, EC_KEY *) = NULL;
 #endif
+/* EVP_PKEY_id is not available on RHEL5 0.9.8e so we'll define our own */
+int Cryptography_EVP_PKEY_id(const EVP_PKEY *key) {
+    return key->type;
+}
```
alex: Is there a reason not to `#define Cryptography_EVP_PKEY_id EVP_PKEY_id` if it's available?

PR #2649, opened 2016-01-08
rename _create_mem_bio to _create_mem_bio_gc for consistency

reaperhulk:
For functions where we return a cdata object that is already registered for garbage collection we are (mostly) consistent with calling that `*_gc`. This makes us a bit more consistent.


alex:
:+1:

PR #2657, opened 2016-01-08
port the 1.2.1 changelog to master

reaperhulk:


Ivoz:
So @reaperhulk , is this bot going to notify me every time there's a change to `changelog.md`? Wondering what I can / should do about that.

reaperhulk:
@Ivoz it (unfortunately) will do that until there are contributors with more blame lines than you or all your blame lines disappear. We can add you to the exclusion list if you'd like though.

PR #2659, opened 2016-01-10
move openssl asn1 encode functions to a new module

reaperhulk:
While this diff looks big it's almost exclusively a cut/paste to a new location.


src/cryptography/hazmat/backends/openssl/encode_asn1.py
```diff
@@ -0,0 +1,603 @@
...
+
+from cryptography import x509
+from cryptography.hazmat.backends.openssl.x509 import (
+    _CRL_ENTRY_REASON_ENUM_TO_CODE, _DISTPOINT_TYPE_FULLNAME,
+    _DISTPOINT_TYPE_RELATIVENAME
+)
+from cryptography.x509.oid import CRLEntryExtensionOID, ExtensionOID, NameOID
+
+
+_MemoryBIO = collections.namedtuple("_MemoryBIO", ["bio", "char_ptr"])
```
alex: This is duplicated between here and `backend.py`

PR #2661, opened 2016-01-10
Write some tests for skip conditions in tests.

alex:
Without this these branches aren't excersised without 0.9.8, but conceptually they are needed.


reaperhulk:
:+1:

Issue #1660, opened 2015-02-13
Create a tool, that given a set of certificates, can automatically build a chain and validate a signature inside the context of that chain


fxdgear:
from a conversation with @reaperhulk he suggested I open an issue here for discussions.

reaperhulk:
@fxdgear's use case here is JWT, which allows you to have a certificate chain (https://tools.ietf.org/html/draft-ietf-jose-json-web-signature-41#section-4.1.6). To validate the token you need to build a chain from the signing certificate up to the root (which you presumably have obtained via some out of band method outside the JSON web token payload).

alex:
@reaperhulk this uses standard X.509 certificate verification right? So we need some sort of certificate store API, and then the ability to verify that a cert (or chain I guess?) is signed by it?

reaperhulk:
Yep. Of course, the devil is in the level of rigor we decide to employ. (OCSP/CRL checks all the way up the chain? If so, how do we do that without building an API that requires implicit outbound network egress)

alex:
I think it needs to be configurable (sadly) somehow, unclear what (if any)
the defaults should be.

On Fri, Feb 13, 2015 at 6:09 PM, Paul Kehrer <notifications@github.com>
wrote:

> Yep. Of course, the devil is in the level of rigor we decide to employ.
> (OCSP/CRL checks all the way up the chain? If so, how do we do that without
> building an API that requires implicit outbound network egress)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pyca/cryptography/issues/1660#issuecomment-74349545>.
>



-- 
"I disapprove of what you say, but I will defend to the death your right to
say it." -- Evelyn Beatrice Hall (summarizing Voltaire)
"The people's good is the highest law." -- Cicero
GPG Key fingerprint: 125F 5C67 DFE9 4084

Ayrx:
Does "tool" in this context mean a command-line program or just an API in cryptography?

alex:
API

On Fri, Feb 13, 2015 at 7:53 PM, Terry Chia <notifications@github.com>
wrote:

> Does "tool" in this context mean a command-line program or just an API in
> cryptography?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pyca/cryptography/issues/1660#issuecomment-74356073>.
>



-- 
"I disapprove of what you say, but I will defend to the death your right to
say it." -- Evelyn Beatrice Hall (summarizing Voltaire)
[...]

PR #2670, opened 2016-01-13
Add verify method on CertificateSigningRequest

joernheissler:


reaperhulk:
This issue touches some of the ideas in #1660 and #2460. In some ways it is simpler since a CSR will not have a chain to validate (and thus `verify` is not quite as dangerous as you're just validating that the signature matches the embedded public key), but some concerns still apply.

Past discussion has led to a decision that these objects will be "data only" and not have logic attached to them. So conceptually we might want to move this to be a top level function. We'd also probably want to bikeshed on the name a bit since `verify` still has unfortunate connotations.

That said, another option would be to implicitly do this validation on parse. Broken CSRs are generally a sign you should not consume them, so maybe it makes sense to raise an exception when parsing? (This would be an easier thing to say we should do if we had a hazmat x509 layer where you could load a CSR even if the signature was broken).

reaperhulk:
Oh and thank you for the contribution! We're always happy to see people contribute a PR for discussion!

alex:
I think I could be convinced that on a CSR, it's not a behavior really,
it's just a property `is_signature_valid`? What do you think?

On Wed, Jan 13, 2016 at 5:18 PM, Paul Kehrer <notifications@github.com>
wrote:

> This issue touches some of the ideas in #1660
> <https://github.com/pyca/cryptography/issues/1660> and #2460
> <https://github.com/pyca/cryptography/pull/2460>. In some ways it is
> simpler since a CSR will not have a chain to validate (and thus verify is
> not quite as dangerous as you're just validating that the signature matches
> the embedded public key), but some concerns still apply.
>
> Past discussion has led to a decision that these objects will be "data
> only" and not have logic attached to them. So conceptually we might want to
> move this to be a top level function. We'd also probably want to bikeshed
> on the name a bit since verify still has unfortunate connotations.
>
> That said, another option would be to implicitly do this validation on
> parse. Broken CSRs are generally a sign you should not consume them, so
> maybe it makes sense to raise an exception when parsing? (This would be an
> easier thing to say we should do if we had a hazmat x509 layer where you
> could load a CSR even if the signature was broken).
>
> —
> <https://github.com/pyca/cryptography/pull/2670#issuecomment-171454949>.
>



-- 

joernheissler:
I think it should be possible to load CSRs with wrong signatures, if the user really wants to. Else an Exception would be appropriate.
That means by verify (what's wrong with verify?) method isn't good enough, neither is a property.
How about an optional flag to the load functions that defaults to "safe mode" and a property to do verification in unsafe mode?

reaperhulk:
The issue with `verify` is that in general developers unfamiliar with cryptography (and specifically x509) conflate a call with that to the complete verification of the entire object. For example, when calling `verify` on our signature verification contexts you can trust that the entire payload has been verified and was signed by the key you expected.

In X509 this is trickier. For certificates verifying the certificate signature is merely part 1 of a long set of steps (build a chain to a trusted root, verify that each certificate itself has a valid signature signed by its parent, validate the key usage/extended key usage/extensions exist that allow the cert to do task x, fetch and check any CRL/OCSP data to confirm that each cert is not revoked, check that each cert is not expired and is in its validity window, etc), so `verify` becomes an arguably dangerous misnomer.

PKCS10 certificate signing requests are a bit simpler in that regard. They don't have revocation, key usages, certificate chains, etc, but the guarantees provided by verifying the signature are also commensurately less useful. In fact, the only thing verifying a signature on the CSR provides is assurance that whoever signed the requested subject DN, extensions, and public key possesses the private key associated with that public key. This assurance is not nothing, but it's also not necessarily what we'd want to call `verify`.

Sorry for the novella! To get this PR moving again, I'd support the `is_signature_valid` property that @alex proposed until such time as we have a `cryptography.hazmat.x509` layer and can then potentially switch to an API where we disallow broken signatures on load for the recipe layer and allow it inside hazmat.

alex:
Jenkins, ok to test.

alex:
@reaperhulk Does this look ok to you? Would be great to get this into 1.3

reaperhulk:
Yeah I'm okay with this approach for now (don't like the name but don't have a better suggestion :grinning:). Looks like memory is properly managed here too, so consider me :+1:

reaperhulk:
(We should add a changelog entry in a separate PR)

PR #2673, opened 2016-01-16
Add docstring to cryptography.hazmat

dveeden:
This shows the warning which is in the docs if someone does this:
```
>>>> import cryptography.hazmat
>>>> help(cryptography.hazmat)
```

### DIFF
--- a/src/cryptography/hazmat/__init__.py
+++ b/src/cryptography/hazmat/__init__.py
@@ -1,5 +1,11 @@
 # This file is dual licensed under the terms of the Apache License, Version
 # 2.0, and the BSD License. See the LICENSE file in the root of this repository
 # for complete details.
+"""
+Hazardous Materials
 
+This is a "Hazardous Materials" module. You should ONLY use it if you're
+100% absolutely sure that you know what you're doing because this module
+is full of land mines, dragons, and dinosaurs with laser guns.
+"""
 from __future__ import absolute_import, division, print_function


src/cryptography/hazmat/__init__.py
```diff
@@ -1,5 +1,10 @@
 # This file is dual licensed under the terms of the Apache License, Version
 # 2.0, and the BSD License. See the LICENSE file in the root of this repository
 # for complete details.
+"""Hazardous Materials
 
+You should ONLY use it if you're 100% absolutely sure that you know
+what you're doing because this module is full of land mines, dragons,
+and dinosaurs with laser guns.
```
alex: I think we want to use the exact originial text here, otherwise the `it` in the first line is an ambigious reference:

```
This is a "Hazardous Materials" module. You should **ONLY** use it if you're
100% absolutely sure that you know what you're doing because this module is
full of land mines, dragons, and dinosaurs with laser guns.
```

src/cryptography/hazmat/__init__.py
```diff
@@ -1,5 +1,10 @@
 # This file is dual licensed under the terms of the Apache License, Version
 # 2.0, and the BSD License. See the LICENSE file in the root of this repository
 # for complete details.
+"""Hazardous Materials
```
alex: Nit: newline immediately after the quotes please.

reaperhulk:
Jenkins, ok to test

alex:
@reaperhulk tests appear to be failing (on one platform) because tox is exploding on an env var with unicode (the author's name in this case). I've already filed an upstream bug with tox, in the meantime is there a way we can disable having this env var?

reaperhulk:
jenkins, retest this please

reaperhulk:
jenkins, retest this please

PR #2674, opened 2016-01-17
Attempt to reduce the duplication in declaring our test deps

alex:


alex:
Bleh, tox doesn't let you declare extras like this. Going to give up on fixing it and focus on just the `dev-requirements.txt`

setup.py
```diff
@@ -308,6 +308,9 @@ def run_tests(self):
 
     install_requires=requirements,
     tests_require=test_requirements,
+    extras_require={
```
reaperhulk: Does setuptools 1.0 handle extras properly? (I don't know that I would say it needs to since it's not unreasonable to ask developers to be vaguely current, just curious)

alex: I don't know. As you said, the only place we use this is development workflow, so I don't care :-)

PR #2677, opened 2016-01-18
add initial OCSP bindings

reaperhulk:
(Now we see if any of our platforms build with `no-ocsp` or some other weird OpenSSL build flag that makes this unavailable to us)


alex:
Should be able to remove the windows hijinks now

PR #2678, opened 2016-01-19
consolidate the windows specific header trickery we need to do

reaperhulk:
Import `e_os2.h` to get the proper Windows defines, then import `windows.h`. Hoist all of this up so it happens before anything else gets included so subsequent includes of things like cms, ssl, and ocsp will include without redefinition errors.

### DIFF
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -37,7 +37,11 @@ def _osx_libraries(build_static):
         return ["ssl", "crypto"]
 
 
-_OSX_PRE_INCLUDE = """
+_PRE_INCLUDE = """
+#include <openssl/e_os2.h>
+#if defined(OPENSSL_SYS_WINDOWS)
+#include <windows.h>
+#endif
 #ifdef __APPLE__
 #include <AvailabilityMacros.h>
 #define __ORIG_DEPRECATED_IN_MAC_OS_X_VERSION_10_7_AND_LATER \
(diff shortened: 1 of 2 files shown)


src/_cffi_src/build_openssl.py
```diff
@@ -60,6 +60,7 @@ def _osx_libraries(build_static):
     module_name="_openssl",
     module_prefix="_cffi_src.openssl.",
     modules=[
+        "windows-hack",
```
alex: Should this be a part of the `PRE_INCLUDE`?

src/_cffi_src/build_openssl.py
```diff
@@ -37,7 +37,11 @@ def _osx_libraries(build_static):
         return ["ssl", "crypto"]
 
 
-_OSX_PRE_INCLUDE = """
+_PRE_INCLUDE = """
+#include <openssl/e_os2.h>
```
alex: Does this need to be after the OS X crap to avoid a warning, or does it work out ok?

reaperhulk: It works out okay.

alex: :+1:

On Mon, Jan 18, 2016 at 10:11 PM, Paul Kehrer <notifications@github.com>
wrote:

> In src/_cffi_src/build_openssl.py
> <https://github.com/pyca/cryptography/pull/2678#discussion_r50068570>:
>
> > @@ -37,7 +37,11 @@ def _osx_libraries(build_static):
> >          return ["ssl", "crypto"]
> >
> >
> > -_OSX_PRE_INCLUDE = """
> > +_PRE_INCLUDE = """
> > +#include <openssl/e_os2.h>
>
> It works out okay.
>
> —
> <https://github.com/pyca/cryptography/pull/2678/files#r50068570>.
>



-- 

PR #2684, opened 2016-01-22
Formally deprecated support for OpenSSL 0.9.8

alex:
As proposed on the mailing list: https://mail.python.org/pipermail/cryptography-dev/2016-January/000551.html


Lukasa:
LGTM.

PR #2685, opened 2016-01-23
Require a newer setuptools.

alex:
This lets a long confusing comment.

@dstufft: PTAL :-)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ requirements = [
     "idna>=2.0",
     "pyasn1>=0.1.8",
     "six>=1.4.1",
-    "setuptools>=1.0",
+    "setuptools>=11.3",
 ]
 setup_requirements = []
 
--- a/src/cryptography/hazmat/backends/__init__.py
+++ b/src/cryptography/hazmat/backends/__init__.py
@@ -17,12 +17,7 @@ def _available_backends():
 
     if _available_backends_list is None:
         _available_backends_list = [
-            # setuptools 11.3 deprecated support for the require parameter to
-            # load(), and introduced the new resolve() method instead.
-            # This can be removed if/when we can assume setuptools>=11.3. At
-            # some point we may wish to add a warning, to push people along,
-            # but at present this would result in too many warnings.
-            ep.resolve() if hasattr(ep, "resolve") else ep.load(require=False)
+            ep.resolve()
             for ep in pkg_resources.iter_entry_points(
                 "cryptography.backends"
             )


reaperhulk:
setuptools 11.3 was released Jan 2015 and the maintenance burden for us supporting the older versions is essentially non-existent. People who package us for distributions (for example, CentOS 7 or Ubuntu 14.04) will just end up patching this code back in downstream if their setuptools isn't new enough anyway.

If there's a legitimate need for a newer version then I'm happy to up the dependency, but as is this doesn't seem compelling.

alex:
Are CentOS7 or Ubuntu 14.04 packaging newer versions of us?

dstufft:
On non Debuntu systems you'll end up fighting with the package manager about what the contents of ``/usr/lib/pythonX.Y/site-packages/setuptools/*`` is since both pip and their package manager will expect to be able to control it. Not in an of itself a big deal, but something to consider.

dstufft:
Otherwise though, this should work fine.

alex:
FWIW, the impetus for this was: we selected the 1.0 minimum somewhat at random, I decided to read the changelog and see if there was something better. The two I came up with were:

* 1.3: Fixes a TLS verification bug
* 11.3: Let's us delete some code and comments

So yeah, is this likely to break anything for people?

dstufft:
I can't think of anything it'd break that mentioning setuptools at all in ``install_requires`` wouldn't (due to the "upgrade everything" nature of ``pip install -U setuptools``). It just might mean that ``pip install cryptography`` (without the ``-U``) might end up doing an upgrade of setuptools as well, instead of that only happening if they use ``-U``.

alex:
And does the possibility of the packaging issue @reaperhulk mentioned concern you?

dstufft:
Not majorly. I think it'd only be a concern for backports style repositories of those distros since anyone pulling in a brand new version of cryptography at this point should be able to pull in a new version of setuptools too I think. Like, I don't think that Debian is going to add a new version of cryprography to Jesse except perhaps via jesse-backports. @mithrandi might have a different opinion.

mithrandi:
The odds of seeing a wholesale cryptography stable update (ie. in jessie, not jessie-backports) are pretty low unless there's a change in stable update policies.

And in jessie-backports, we can always backport a newer setuptools if it's necessary, so (speaking with my Debian Developer and python-cryptography maintainer hat on) I don't expect this to cause any problems.

reaperhulk:
Consider me outvoted I suppose!

alex:
Thanks @mithrandi !

PR #2703, opened 2016-01-31
Added two additional bindings that PyOpenSSL might want	

alex:


reaperhulk:
:+1: I also confirmed these signatures are present and unchanged back to 0.9.8b, although since we're dropping 0.9.8 support soon it doesn't really matter :)

PR #2707, opened 2016-02-02
some backends may error on non-block size aligned data in update

reaperhulk:
Some PKCS11 implementations require block size aligned data for the AES CBC mechanism (along with other block modes) in calls to update. This change allows the fernet tests to properly fail with InvalidToken when a ValueError is raised from update, not just finalize.


alex:
This is wrong and is incosisntent with our API. On such a backend, their Cipher implementnaation should buffer block sizes internally.

reaperhulk:
I think we may have to revisit this in the future depending on what guarantees we want to eventually be possible with backends (e.g. passing `char[]` CData objects so that memory can actually be zeroed), but for now I'll do buffering.

PR #2718, opened 2016-02-08
Actually allocate a buffer that is the correct size.

### LINKED ISSUE #2711: Intermittent segfault on Debian unstable / i386
I'm getting a segfault non-deterministically running the test suite on Debian unstable i386. I have not been able to reproduce the segfault on amd64, and it doesn't happen every time, but there seems to be a fairly high probability that I'll get a segfault somewhere during the test run.

An example of one such backtrace during a full test run is: https://gist.github.com/mithrandi/347185c625358b6ab9b9

I have reproduced the segfault (at least, I assume it's the same one) on Python 2.7, Python 3.4, and Python 3.5.

Running `tests/hazmat/backends/test_openssl.py::TestOpenSSLSerializationWithOpenSSL::test_pem_password_cb` by itself seems to segfault every time on 2.7, and with high probability on 3.x. Occasionally, the following test failure is observed instead of a segfault:
```
self = <tests.hazmat.backends.test_openssl.TestOpenSSLSerializationWithOpenSSL object at 0xedf1e9ac>

    def test_pem_password_cb(self):
        password = b'abcdefg'
        ffi_cb, userdata = backend._pem_password_cb(password)
        handle = backend._ffi.new_handle(userdata)
        buf = backend._ffi.new('char *')
        assert ffi_cb(buf, len(password) + 1, False, handle) == len(password)
        assert userdata.called == 1
>       assert backend._ffi.string(buf, len(password)) == password
E       assert b'abcdcfg' == b'abcdefg'
E         At index 4 diff: 99 != 101
E         Use -v to get the full diff

tests/hazmat/backends/test_openssl.py:619: AssertionError
```
An example of a backtrace from one of these is here: https://gist.github.com/mithrandi/5c1e09486154a092c5bb

As the backtraces differ, I'm not entirely sure there's only one bug here, but it seems likely that they're related at least.

mithrandi:
Fixes #2711

### DIFF
--- a/tests/hazmat/backends/test_openssl.py
+++ b/tests/hazmat/backends/test_openssl.py
@@ -611,10 +611,11 @@ class TestOpenSSLSerializationWithOpenSSL(object):
 
     def test_pem_password_cb(self):
         password = b'abcdefg'
+        buf_size = len(password) + 1
         ffi_cb, userdata = backend._pem_password_cb(password)
         handle = backend._ffi.new_handle(userdata)
-        buf = backend._ffi.new('char *')
-        assert ffi_cb(buf, len(password) + 1, False, handle) == len(password)
+        buf = backend._ffi.new('char[]', buf_size)
+        assert ffi_cb(buf, buf_size, False, handle) == len(password)
         assert userdata.called == 1
         assert backend._ffi.string(buf, len(password)) == password
 


alex:
Jenkins, test this please.

alex:
Jenkins, retest this please.

reaperhulk:
:+1: thanks for tracking this down @mithrandi!

PR #2729, opened 2016-02-24
Add bindings for cert validation on Windows.

Lukasa:
As part of the work I'm doing on certitude, I wanted to be able to use the Windows native crypto APIs to validate certificate chains. This diff exposes bindings used for doing just that.

This is obviously a bit tricky because it introduces a new bindings source, so I suspect I haven't done everything you need me to do on that front: please let me know what you need and I'll make the appropriate changes.

In the event that you want to test that everything works fine, you can use the test file available [here](https://gist.github.com/Lukasa/dc754d2193c53fdecde2).

### DIFF
--- a/src/_cffi_src/crypt32/crypt32.py
+++ b/src/_cffi_src/crypt32/crypt32.py
@@ -74,7 +74,7 @@ typedef struct _CERT_TRUST_STATUS {
     DWORD dwInfoStatus;
 } CERT_TRUST_STATUS, *PCERT_TRUST_STATUS;
 
-typedef const struct _CERT_CHAIN_CONTEXT* PCCERT_CHAIN_CONTEXT;
+typedef const struct _CERT_CHAIN_CONTEXT *PCCERT_CHAIN_CONTEXT;
 typedef struct _CERT_CHAIN_CONTEXT {
     DWORD cbSize;
     CERT_TRUST_STATUS TrustStatus;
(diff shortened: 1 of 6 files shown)


dstufft:
The OpenSSL and CommonCrypto bindings predate the ability to load backends from projects other than cryptography itself. I wonder if it would make more sense to make a new project, ``cryptography-crypt32`` for these bindings? The downside is that we'll have to solve some of the open questions around how we test "external" backends/bindings... however if we do it now we never have to deal with backwards compatibility concerns for the import statement for people using the crypt32 bindings directly (like we'll have to deal with for the OpenSSL and CommonCrypto bindings if we ever move them out).

I'm not sure if it's worth it to block this on that though, was just a thought that popped into my head.

Lukasa:
So the Jenkins build is failing but *not* on any of the Windows platforms and those failures seem to be more to do with Jenkins than with this change. Leaving that up to @reaperhulk.

src/_cffi_src/crypt32/crypt32.py
```diff
@@ -0,0 +1,144 @@
...
+    LONG lElementIndex;
+    void *pvExtraPolicyStatus;
+} CERT_CHAIN_POLICY_STATUS, *PCERT_CHAIN_POLICY_STATUS;
+
+typedef struct _CERT_TRUST_STATUS {
+    DWORD dwErrorStatus;
+    DWORD dwInfoStatus;
+} CERT_TRUST_STATUS, *PCERT_TRUST_STATUS;
+
+typedef const struct _CERT_CHAIN_CONTEXT* PCCERT_CHAIN_CONTEXT;
```
reaperhulk: The `*` here should be touching `PCCERT_CHAIN_CONTEXT`.

reaperhulk:
Until we have Windows coverage in our infra we can't really merge this. Since you've gone down the certitude route for now I'm going to go ahead and close this, but if you want to restart the discussion feel free to reopen (or yell at me at dinner)

PR #2731, opened 2016-02-24
Restructure repository to cleanly separate packages 

### LINKED ISSUE #1473: Better handle multiple packages

dstufft:
Currently we have the `cryptography` package at `.` and the `cryptography-vectors` package at `./vectors`. Instead, we'll make a sub directory under `.` for each individual package, named after the package to cleanly separate between them.

Fixes #1473 and is a stepping stone towards a solution to #1522


dstufft:
The coverage code was confused because coverage.py sees there is a directory named ``cryptography`` and assumes that it is what we're trying to cover instead of the package named ``cryptography``. There's no way around this currently besides naming the top level directory something other than ``cryptography``, so instead I added a new option to coverage.py in https://github.com/nedbat/coveragepy/pull/3, and temporarily switched this PR over to using that fork.

alex:
It's been over a year, there appears to be no interest here.

PR #2736, opened 2016-02-26
Added support for padding ANSI X.923

cedk:
I'm implementing an EBICS library and this protocol requires to pad using ANSI X.923.
So I think it will be good to have it in cryptography library.

### DIFF
--- a/docs/hazmat/primitives/padding.rst
+++ b/docs/hazmat/primitives/padding.rst
@@ -54,6 +54,47 @@ multiple of the block size.
             provider.
 
 
+.. class:: ANSIX923(block_size)
+
+    ANSI X.923 padding works by appending ``N-1`` bytes with the value of ``0``
+    and a last byte with the value of ``chr(N)``, where ``N`` is the number of
+    bytes required to make the final block of data the same size as the block
+    size. A simple example of padding is:
+
+    .. doctest::
+
+        >>> padder = padding.ANSIX923(128).padder()
+        >>> padded_data = padder.update(b"11111111111111112222222222")
+        >>> padded_data
+        '1111111111111111'
+        >>> padded_data += padder.finalize()
+        >>> padded_data
+        '11111111111111112222222222\x00\x00\x00\x00\x00\x06'
+        >>> unpadder = padding.ANSIX923(128).unpadder()
+        >>> data = unpadder.update(padded_data)
+        >>> data
+        '1111111111111111'
+        >>> data + unpadder.finalize()
+        '11111111111111112222222222'
+
+    :param block_size: The size of the block in bits that the data is being
+        padded to.
+    :raises ValueError: Raised if block size is not a multiple of 8 or is not
+        between 0 and 256.
+
+    .. method:: padder()
+
+        :returns: A padding
+            :class:`~cryptography.hazmat.primitives.padding.PaddingContext`
+            provider
+
+    .. method:: unpadder()
+
+        :returns: An unpadding
+            :class:`~cryptography.hazmat.primitives.padding.PaddingContext`
+            provider.
+
+
 .. class:: PaddingContext
 
     When calling ``padder()`` or ``unpadder()`` the result will conform to the
--- a/docs/hazmat/primitives/padding.rst
+++ b/docs/hazmat/primitives/padding.rst
@@ -56,10 +56,10 @@ multiple of the block size.
 
 .. class:: ANSIX923(block_size)
 
-    ANSI X.923 padding works by appending ``N-1`` bytes with the value of ``0``
-    and a last byte with the value of ``chr(N)``, where ``N`` is the number of
-    bytes required to make the final block of data the same size as the block
-    size. A simple example of padding is:
+    `ANSI X.923`_ padding works by appending ``N-1`` bytes with the value of
+    ``0`` and a last byte with the value of ``chr(N)``, where ``N`` is the
+    number of bytes required to make the final block of data the same size as
+    the block size. A simple example of padding is:
 
     .. doctest::
 
@@ -123,3 +123,5 @@ multiple of the block size.
         :raises TypeError: Raised if data is not bytes.
         :raises ValueError: When trying to remove padding from incorrectly
                             padded data.
+
+.. _`ANSI X.923`: https://en.wikipedia.org/wiki/Padding_%28cryptography%29#ANSI_X.923
--- a/docs/hazmat/primitives/padding.rst
+++ b/docs/hazmat/primitives/padding.rst
@@ -56,6 +56,8 @@ multiple of the block size.
 
 .. class:: ANSIX923(block_size)
 
+    .. versionadded:: 1.3
+
     `ANSI X.923`_ padding works by appending ``N-1`` bytes with the value of
     ``0`` and a last byte with the value of ``chr(N)``, where ``N`` is the
     number of bytes required to make the final block of data the same size as
--- a/src/cryptography/hazmat/primitives/padding.py
+++ b/src/cryptography/hazmat/primitives/padding.py
@@ -28,107 +28,76 @@ class PaddingContext(object):
         """
 
 
-class _BytePadding(object):
+def _byte_padding_check(block_size):
+    if not (0 <= block_size < 256):
+        raise ValueError("block_size must be in range(0, 256).")
+
+    if block_size % 8 != 0:
+        raise ValueError("block_size must be a multiple of 8.")
+
+
+def _byte_padding_update(buffer_, data, block_size):
+    if buffer_ is None:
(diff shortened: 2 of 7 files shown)


reaperhulk:
jenkins, ok to test

docs/hazmat/primitives/padding.rst
```diff
@@ -54,6 +54,47 @@ multiple of the block size.
             provider.
 
 
+.. class:: ANSIX923(block_size)
+
+    ANSI X.923 padding works by appending ``N-1`` bytes with the value of ``0``
```
reaperhulk: Is there a link to the ANSI X9.23 documentation that we can put here?

src/cryptography/hazmat/primitives/padding.py
```diff
@@ -122,3 +128,26 @@ def finalize(self):
         res = self._buffer[:-pad_size]
         self._buffer = None
         return res
+
+
+class ANSIX923(PKCS7):
```
alex: ANSIx923 should not be subclassing PKCS7, these classes should be unrelated.

reaperhulk: cryptography as a project has a general policy against subclassing. In this case I'd recommend creating a separate class and then factor out the block size checks from the `PKCS7` class so they can be called from both PKCS7 and the ANSIX923 classes.

src/cryptography/hazmat/primitives/padding.py
```diff
@@ -122,3 +128,26 @@ def finalize(self):
...
+
+    def _padding(self, size):
+        return six.int2byte(0) * (size - 1) + six.int2byte(size)
+
+
+@utils.register_interface(PaddingContext)
+class _ANSIX923UnpaddingContext(_PKCS7UnpaddingContext):
+
+    def _check_padding(self):
+        return True
```
reaperhulk: This doesn't seem right?

reaperhulk:
I've added some comments, but I'm sure @alex will have some as well. Thank you for your work so far -- it always makes me happy to see a PR with documentation and tests that show a clear understanding of the project's style!

src/cryptography/hazmat/primitives/padding.py
cedk: Why? There is no check in ANSI X.923 padding.

alex: Shouldn't we check that all the bytes are 0? Without that you are
vulnerable to attacks such as the one on SSLv3.

On Fri, Feb 26, 2016 at 5:14 PM, Cédric Krier <notifications@github.com>
wrote:

> In src/cryptography/hazmat/primitives/padding.py
> <https://github.com/pyca/cryptography/pull/2736#discussion_r54309644>:
>
> > +    def unpadder(self):
> > +        return _ANSIX923UnpaddingContext(self.block_size)
> > +
> > +
> > +@utils.register_interface(PaddingContext)
> > +class _ANSIX923PaddingContext(_PKCS7PaddingContext):
> > +
> > +    def _padding(self, size):
> > +        return six.int2byte(0) * (size - 1) + six.int2byte(size)
> > +
> > +
> > +@utils.register_interface(PaddingContext)
> > +class _ANSIX923UnpaddingContext(_PKCS7UnpaddingContext):
> > +
> > +    def _check_padding(self):
> > +        return True
>
> Why? There is no check in ANSI X.923 padding.
>
> —
> <https://github.com/pyca/cryptography/pull/2736/files#r54309644>.
>



-- 

docs/hazmat/primitives/padding.rst
cedk: I can put the link to wikipedia:
https://en.wikipedia.org/wiki/Padding_%28cryptography%29#ANSI_X.923

reaperhulk: :+1: since ANSI docs aren't freely available

src/cryptography/hazmat/primitives/padding.py
cedk: Or I could test that it is filled with zero?

cedk:
I think I have fixed all the remarks.
For the code coverage, I don't know how to make it clear because codecov complains about code that should never be reach 'NotImplemented'.

cedk:
I modified to use abstractmethod but it seems codecov did not update.
jenkins, ok to test

alex:
Codecov doesn't leave a new comment, it just updates the old one.

cedk:
But it reports missing coverage on code that is no more there.
Is it a way to force an update?

reaperhulk:
It might work to close/reopen the PR (I'll do that momentarily). If that doesn't work we can rebase/merge master.

reaperhulk:
Okay, Travis is running again. Fingers crossed (although I suspect we'll need to do the rebase/merge master trick to change the commit hash. That should un-confuse codecov)

cedk:
Did not change, so I pushed a rebase.

cedk:
ping

src/cryptography/hazmat/primitives/padding.py
```diff
@@ -122,3 +126,53 @@ def finalize(self):
         res = self._buffer[:-pad_size]
         self._buffer = None
         return res
+
+
+class PKCS7(_BytePadding):
```
reaperhulk: You're still using subclassing here (and for ANSIX923). Could you please remove the subclassing and instead make the size checks from the parent class's `__init__` be a function you call?

cedk: I don't get the point. I understood before that it was not clean to inherit from  PKCS7 to create ANSIX923. But here I moved into a private class the common code for any padding based on byte. For example, it could use to also implement ISO 10126 or ISO/IEC 7816-4.
If I don't use subclass, I will have to duplicate most of the code.

reaperhulk: This project has a general policy against subclassing for a variety of reasons. While it is simple here, we'd prefer the logic to be self-contained. Sorry about that.

cedk: But I can find existing exception to this rule like _AEADEncryptionContext
If I can not subclass in this case I will have to duplicate a good 200 lines of code (or design a poor manual subclassing behaviour), I think this is much bigger problem than your rule. I understand they are rules but I guess they should be considered as best practice and not fundamental at the point to not be broken when it is obviously better.

cedk:
I tried to rewrite it without sub-classing but I cannot without having such horrible code.

alex:
There's no reason that removing subclassing will require duplication of code. You can see the techniques employed for example here: https://github.com/pyca/cryptography/blob/master/src%2Fcryptography%2Fhazmat%2Fprimitives%2Fkdf%2Fconcatkdf.py to seperate code

cedk:
Exactly this example shows plenty of code duplication, I could rewrite such code with 30% less code.
Please could you provide why sub-classing is "forbidden"? I have seen no where in the documentation reference about such rule and I even found counter example https://github.com/pyca/cryptography/pull/2736#discussion_r55152142

alex:
It's probably not documented as well as it should be, and we're aware of
that exception. Nonetheless, this is a design decision we've made, many of
us were deeply influenced by https://www.youtube.com/watch?v=3MNVP9-hglc

On Sun, Mar 13, 2016 at 6:54 PM, Cédric Krier <notifications@github.com>
wrote:

> Exactly this example shows plenty of code duplication, I could rewrite
> such code with 30% less code.
> Please could you provide why sub-classing is "forbidden"? I have seen no
> where in the documentation reference about such rule and I even found
> counter example #2736 (comment)
> <https://github.com/pyca/cryptography/pull/2736#discussion_r55152142>
>
> —
> <https://github.com/pyca/cryptography/pull/2736#issuecomment-196076361>.
>



-- 

cedk:
OK, I rewrite it even if I think in this case sub-classing makes sense.
Let me know if this design is right according to the project policy.

docs/hazmat/primitives/padding.rst
```diff
@@ -54,6 +54,47 @@ multiple of the block size.
             provider.
 
 
+.. class:: ANSIX923(block_size)
```
reaperhulk: Could you add a `.. versionadded:: 1.3` to this section?

cedk: Done.

reaperhulk:
Would you mind adding a use after finalize, non-bytes, and invalid block size test? They can basically look identical to the PKCS7 tests, just want to make sure that if we split the shared functions in the future we still cover those paths. Other than that I think this LGTM. @alex would you mind giving it a final review as well (modulo those extra tests).

cedk:
Done.

reaperhulk:
Thanks for working with us on this!

PR #2739, opened 2016-02-27
Fixed #2732 -- added support for x.509 policy constraints extension

### LINKED ISSUE #2732: Add support for policyConstraints x.509 extension in parser
```
cryptography.x509.extensions.UnsupportedExtension: Critical extension <ObjectIdentifier(oid=2.5.29.36, name=policyConstraints)> is not currently supported
```

alex:
Two TODOs from my perspective:

- [x]  That test method needs a name
- [x] What's the way to get coverage on the uncovered path, find another vector?


reaperhulk:
Building a policyConstraints vector that contains an explicit but no inhibit is pretty easy so that's probably the best path, yes.

PR #2745, opened 2016-02-28
Added a changelog entry for policy constraints

alex:


sigmavirus24:
Are y'all still using mentionbot? If so, how is it working now. I keep getting subscribed without understanding why and each time I see:

> You’re receiving notifications because you were mentioned.

But I wasn't mentioned :/

alex:
mmm, now that you mention it, I don't think mentionbot has been leaving comments lately. I don't know how you're subscribed.

Issue #2750, opened 2016-03-01
Issue 1.2.3 using OpenSSL 1.0.2g


alex:

zx2c4:
````
/var/tmp/portage/dev-python/cryptography-1.2.2/work/cryptography-1.2.2-python2_7/temp.linux-x86_64-2.7/_openssl.c:726:6: error: conflicting types for ‘BIO_new_mem_buf’
 BIO *BIO_new_mem_buf(void *, int);
      ^
In file included from /usr/include/openssl/asn1.h:65:0,
                 from /var/tmp/portage/dev-python/cryptography-1.2.2/work/cryptography-1.2.2-python2_7/temp.linux-x86_64-2.7/_openssl.c:434:
/usr/include/openssl/bio.h:692:6: note: previous declaration of ‘BIO_new_mem_buf’ was here
 BIO *BIO_new_mem_buf(const void *buf, int len);
````

**Comment by zx2c4** [comment]
Fixed downstream here:
https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=fe3faddb9bf793b16a104bc5f0d304f14e8cf057

**Comment by mboersma** [comment]
This same bug also exists on Alpine Linux 3.3. If there is any way to fix it in pyca cryptography itself, that would be highly preferable for our needs.

**Comment by reaperhulk** [comment]
The 1.2.3 release (coming today) will fix this (*grumbles about OpenSSL changing a function signature in a security release*)

**Comment by mboersma** [comment]
@reaperhulk thanks! We'll keep our eyes out for the new release and test it.

**Comment by danielfaust** [comment]
Any ETA for this issue?

I have a minimal Alpine 3.3 Dockerfile.

```
FROM alpine:3.3

RUN apk --update add build-base libffi-dev openssl-dev python-dev py-pip
RUN pip install cryptography
```

alex:
Release is up now :-)

tjwebb:
I just ran into this using 1.5.2 (python 2.7). Am I missing something?

tjwebb:
Update: ok, I got past this issue by making sure the cryptography dependencies are also all updated to the latest versions.

anka-213:
@tjwebb How did you do this?

PR #2754, opened 2016-03-01
Build with openssl 1.0.2g

zx2c4:
It's a simple constness change.

Fixes: #2750
Signed-off-by: Jason A. Donenfeld <Jason@zx2c4.com>


reaperhulk:
Thanks for the patch -- we've actually got a fix already landed in our 1.2.x branch, which will become 1.2.3 momentarily. This particular approach will result in build failures for any version of OpenSSL that's not 1.0.2g due to the way we validate the bindings -- the best way to do it is to move this binding down into macros where we don't check function signatures. If you'd like to update this PR with that approach we'd be happy to merge your fix into master!

alex:
Closing in favor of #2755 -- thanks for contributing!

PR #2761, opened 2016-03-05
Fixed #2760 -- handle more types of bad RSA decrypts

### LINKED ISSUE #2760: Intermittent AssertionError on PrivateKey.decrypt
I think this is the same issue as seen in #2261

````
  File "/usr/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "test_encrypt.py", line 63, in do_encrypt_decrypt
    label=None
  File "/usr/local/lib/python2.7/unittest/case.py", line 475, in assertRaises
    callableObj(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/rsa.py", line 534, in decrypt
    return _enc_dec_rsa(self._backend, self, ciphertext, padding)
  File "/usr/local/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/rsa.py", line 76, in _enc_dec_rsa
    return _enc_dec_rsa_pkey_ctx(backend, key, data, padding_enum)
  File "/usr/local/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/rsa.py", line 105, in _enc_dec_rsa_pkey_ctx
    _handle_rsa_enc_dec_error(backend, key)
  File "/usr/local/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/rsa.py", line 146, in _handle_rsa_enc_dec_error
    assert errors[0].reason in decoding_errors

AssertionError: 
>>  assert [_OpenSSLError(code=67522692L, lib=4, func=101, reason=132)][0].reason in [106, 107, 121, 159]
    raise ValueError("Decryption failed.")
````

The above assertion error will cause anything from 3-10% of this nose test case to fail when trying to decrypt a payload with the wrong key (decrypt with correct key always works as expected):

`````
import json
from nose.tools import assert_equals, assert_not_equals, assert_raises

from cryptography.hazmat.primitives.asymmetric import padding, rsa
from cryptography.hazmat.primitives.hashes import SHA1


class TestEncrypt(object):
    def do_encrypt_decrypt(self):
        key = rsa.generate_private_key(
            public_exponent=65537,
            key_size=2048,
            backend=default_backend()
        )

        data = json.dumps(dict(secure='Message'))

        payload = key.public_key().encrypt(
            data,
            padding.OAEP(
                mgf=padding.MGF1(algorithm=SHA1()),
                algorithm=SHA1(),
                label=None
            )
        )
        assert_not_equals(data, payload)

        data_decrypted = key.decrypt(
            payload,
            padding.OAEP(
                mgf=padding.MGF1(algorithm=SHA1()),
                algorithm=SHA1(),
                label=None
            )
        )
        assert_equals(data, data_decrypted)

        another_key = rsa.generate_private_key(
            public_exponent=65537,
            key_size=2048,
            backend=default_backend()
        )
        assert_raises(
            ValueError,
            another_key.decrypt,
            payload,
            padding.OAEP(
                mgf=padding.MGF1(algorithm=SHA1()),
                algorithm=SHA1(),
                label=None
            )
        )

    def test_encrypt_decrypt_repeat(self):
        for _ in xrange(100):
            yield self.do_encrypt_decrypt
````

The reason is always 132 (`RSA_R_DATA_TOO_LARGE_FOR_MODULUS`)

cryptography 1.2.2
OpenSSL 1.0.1k 8 Jan 2015

alex:


alex:
@reaperhulk pretty sure the key belongs somewhere else, not sure about the ciphertext. Thoughts?

Also, thoughts on how we could use hypothesis to make these ops a part of our test suite?

reaperhulk:
Yeah the key should probably live in the RSA fixtures (if that specific key is even needed. Can we replicate with a diff key with same modulus size?

Not sure on hypothesis. File an issue so we can land that separately?

alex:
Good call.

On Sat, Mar 5, 2016 at 10:28 AM, Paul Kehrer <notifications@github.com>
wrote:

> Yeah the key should probably live in the RSA fixtures (if that specific
> key is even needed. Can we replicate with a diff key with same modulus size?
>
> Not sure on hypothesis. File an issue so we can land that separately?
>
> —
> <https://github.com/pyca/cryptography/pull/2761#issuecomment-192672525>.
>



-- 

alex:
@reaperhulk I was unable to make the test fail with any of our existing vectors; moved this to the fixtures file

PR #2762, opened 2016-03-05
Un-double the test doubles

alex:


tests/doubles.py
```diff
@@ -0,0 +1,41 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
```
reaperhulk: This is missing our standard from __future__ import

alex: good catch

reaperhulk:
jenkins, retest this please

PR #2770, opened 2016-03-07
HTTPS some links, in the odd event users of a cryptographic library w…

alex:
…ould care about authentication, integrity, or confidentiality


Lukasa:
LGTM.

Issue #1478, opened 2014-11-17
Use of key_size is a bit unfortunate


alex:
via @davidben

```
       DSAPrivateKey.key_size, DSAPublicKey.key_size, and
       EllipticCurve.key_size are pretty unfortunate, both in documentation
       and semantics. DSA and ECC keys have two size parameters, not one.

       For DSA, you've got the prime p and the subprime q. p's size gives
       the size of the public key. q's size gives the size of the private
       key and of one of the two numbers in a signature.

       Likewise, for ECC, there's the order and the degree. These turn out
       to have the same size in bits (but not the same value!) for NIST
       curves, but they are semantically very very different. The degree is
       the size of the underlying prime field. It sizes ONE of the TWO
       components of a point (public key, ECDH output) on the curve. (Docs
       just say "The bit length of the curve’s base point." which is
       ambiguous.) The order is the size of the group generated by the base
       point. It sizes a scalar, so a private key or one of the TWO scalars
       in an ECDSA signature.
```

PR #2772, opened 2016-03-07
potential fix for ambiguity in EC key_size

reaperhulk:
This PR is being pushed to discuss whether we want to pursue a fix like this for the ambiguity discussed in #1478. cc @davidben

If we like this approach then the next step would be to deprecate `key_size` on `EllipticCurve`.


simo5:
Do you forsee (or do there exists) curves where order and degree will be different ?

reaperhulk:
@simo5 non-NIST curves may have different order/degree values. That said, we may not support anything beyond this set of curves for ECDSA (ed25519 will be EdDSA and may or may not fit within the current API).

reaperhulk:
@alex Do you think we should move forward on this for 1.3?

PR #2773, opened 2016-03-07
NIST SP 800-108 Counter Mode KDF test vectors.

jared-mess:
NIST CAVP test vectors for Counter Mode KDF using HMAC-SHA1, HMAC-SHA224, HMAC-SHA256 and HMAC-SHA512 as the PRF.


alex:
Jenkins, ok to test.

reaperhulk:
Looks like these were pulled out of http://csrc.nist.gov/groups/STM/cavp/documents/KBKDF800-108/CounterMode.zip correct?

Normally we prefer to just land the file unchanged from the source so we can do a direct diff. I'm also curious why you excluded the SHA384 vectors and the CTRLOCATION=MIDDLE_FIXED and AFTER_FIXED values?

docs/development/test-vectors.rst
```diff
@@ -344,6 +344,7 @@ Key derivation functions
 * PBKDF2 (HMAC-SHA1) from :rfc:`6070`.
 * scrypt from the `draft RFC`_.
 * X9.63 KDF from `NIST CAVP`_.
+* SP 800-108 Counter Mode KDF (HMAC-SHA1, HMAC-SHA224, HMAC-SHA256, HMAC-SHA512) from `NIST CAVP`_.
```
reaperhulk: This line is too long and needs to be wrapped at 79 chars or less :)

reaperhulk:
There are a few comments on this PR that need to be addressed. @jared-mess have you had a chance to take a look? (If we get this landed in the next week this KDF can potentially be in the 1.3 release).

jared-mess:
@reaperhulk, The missing `SHA384` vectors was an accident (good catch). The reason why there are missing counter location vectors is because they aren't specified in the [SP 800-108 publication](http://csrc.nist.gov/publications/nistpubs/800-108/sp800-108.pdf). I followed the publication verbatim which is slightly inconsistent to the test vectors (the test vectors include the counter appearing in different locations, but the publication only specifies the counter before the fixed input.) 
```
1. n := ⎡L/h⎤.
2. If n > 2^r
    -1, then indicate an error and stop.
3. result(0):= ∅.
4. For i = 1 to n, do
    a. K(i) := PRF (KI, [i]2 || Label || 0x00 || Context || [L]2)
    b. result(i) := result(i-1) || K(i).
5. Return: KO := the leftmost L bits of result(n). 
``` 
More than happy to include those test vectors and update my code to enable the different counter locations. I also only included vectors where `RLEN=32_BITS` for simplicity. Would you like me to add in 8 and 16 bit too? (`struct.pack` doesn't seem to support 24 bits)

reaperhulk:
I've confirmed the updated vector file matches the file supplied by NIST in that zip file so I'm happy to merge this now. We can discuss the subset of vectors that actually get used in the tests in your other PR :) Thanks for working on this!

PR #2778, opened 2016-03-09
handle removal of M_ASN1_TIME_dup

reaperhulk:
OpenSSL 1.1.0 compat work

### DIFF
--- a/src/_cffi_src/openssl/asn1.py
+++ b/src/_cffi_src/openssl/asn1.py
@@ -155,4 +155,11 @@ ASN1_TYPE *d2i_ASN1_TYPE(ASN1_TYPE **, const unsigned char **, long);
 """
 
 CUSTOMIZATIONS = """
+/* This macro is removed in 1.1.0. We re-add it if required to support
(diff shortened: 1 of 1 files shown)


src/_cffi_src/openssl/asn1.py
```diff
@@ -155,4 +155,11 @@
 """
 
 CUSTOMIZATIONS = """
+/* This macro is removed in 1.1.0. We re-add it if required to support
+   pyOpenSSL versions older than whatever resolves
+   https://github.com/pyca/pyopenssl/issues/431 */
+#if !defined(M_ASN1_TIME_dup)
+#define M_ASN1_TIME_dup(a) (ASN1_TIME *)\
+    ASN1_STRING_dup((const ASN1_STRING *)a)
```
alex: Can you put all the `(ASN1_TIME *)` on the same line as this.

reaperhulk: I've fixed it now. This was a copy/paste from the OpenSSL source previously :)

PR #2786, opened 2016-03-09
DH_generate_parameters was deprecated in 0.9.8 and we don't use it

reaperhulk:
and neither does pyOpenSSL :smile_cat:


lvh:
This LGTM; no-one should probably ever have called this.

PR #2787, opened 2016-03-09
some ECDH functions we don't use that are no longer present in 1.1.0

reaperhulk:
pyOpenSSL does not use these.

### DIFF
--- a/src/_cffi_src/openssl/ecdh.py
+++ b/src/_cffi_src/openssl/ecdh.py
@@ -20,13 +20,6 @@ FUNCTIONS = """
 MACROS = """
 int ECDH_compute_key(void *, size_t, const EC_POINT *, EC_KEY *,
                      void *(*)(const void *, size_t, void *, size_t *));
-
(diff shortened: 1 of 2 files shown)


lvh:
I'm guessing pyOpenSSL is the only consumer we really care about? This obviously looks fine to me; my only concern is that something somewhere is calling this (but it seems improbable).

alex:
It's the only external bindings consumer that I'm aware of.

reaperhulk:
The only other people I can think of who might be using bindings separately would be @lukasa in his various pursuits or @glyph within Twisted. That said, I'm pretty sure that neither of them are using this :)

lvh:
Correct; it does not. this looks good to me; thanks @reaperhulk! Can I just merge this or would you prefer me to wait for @Lukasa?

alex:
Once tests pass this is fine to merge

Lukasa:
I don't use these. =)

alex:
Not sure what's going on with codecov, but it's clearly full coverage so I'm going to merge.

PR #2794, opened 2016-03-09
opaque HMAC_CTX, which requires some helper functions

reaperhulk:
The first somewhat complex 1.1.0 compatibility PR. In 1.1.0 `HMAC_CTX` is opaque. We don't reach inside it, but it now means we need allocate/free functions since we don't know its size any more. We add two functions that wrap the new ones on 1.1.0 and emulate them on 1.0.2 and lower. HMAC_CTX_new is reasonably straightforward, but free is a bit confusing. Free calls these functions in 1.1.0:

```c
static void hmac_ctx_cleanup(HMAC_CTX *ctx)
{
    EVP_MD_CTX_reset(ctx->i_ctx);
    EVP_MD_CTX_reset(ctx->o_ctx);
    EVP_MD_CTX_reset(ctx->md_ctx);
    ctx->md = NULL;
    ctx->key_length = 0;
    memset(ctx->key, 0, sizeof(HMAC_MAX_MD_CBLOCK));
}

void HMAC_CTX_free(HMAC_CTX *ctx)
{
    if (ctx != NULL) {
        hmac_ctx_cleanup(ctx);
        EVP_MD_CTX_free(ctx->i_ctx);
        EVP_MD_CTX_free(ctx->o_ctx);
        EVP_MD_CTX_free(ctx->md_ctx);
        OPENSSL_free(ctx);
    }
}



void EVP_MD_CTX_free(EVP_MD_CTX *ctx)
{
    EVP_MD_CTX_reset(ctx);
    OPENSSL_free(ctx);
}


int EVP_MD_CTX_reset(EVP_MD_CTX *ctx)
{
    if (ctx == NULL)
        return 1;

    /*
     * Don't assume ctx->md_data was cleaned in EVP_Digest_Final, because
     * sometimes only copies of the context are ever finalised.
     */
    if (ctx->digest && ctx->digest->cleanup
        && !EVP_MD_CTX_test_flags(ctx, EVP_MD_CTX_FLAG_CLEANED))
        ctx->digest->cleanup(ctx);
    if (ctx->digest && ctx->digest->ctx_size && ctx->md_data
        && !EVP_MD_CTX_test_flags(ctx, EVP_MD_CTX_FLAG_REUSE)) {
        OPENSSL_clear_free(ctx->md_data, ctx->digest->ctx_size);
    }
    EVP_PKEY_CTX_free(ctx->pctx);
#ifndef OPENSSL_NO_ENGINE
    ENGINE_finish(ctx->engine);
#endif
    memset(ctx, 0, sizeof(*ctx));

    return 1;
}
```

In 1.0.2 some of these functions do exist, they just have different names. For example, most of `EVP_MD_CTX_free` is inside `EVP_MD_CTX_cleanup` and and `HMAC_CTX_cleanup` is similar. So the code in this PR merges what it can together to emulate the behavior of 1.1.0 

For reference here are `EVP_MD_CTX_cleanup` and `HMAC_CTX_cleanup`:

```c
void HMAC_CTX_cleanup(HMAC_CTX *ctx)
    {
#ifdef OPENSSL_FIPS
    if (FIPS_mode() && !ctx->i_ctx.engine)
        {
        FIPS_hmac_ctx_cleanup(ctx);
        return;
        }
#endif
    EVP_MD_CTX_cleanup(&ctx->i_ctx);
    EVP_MD_CTX_cleanup(&ctx->o_ctx);
    EVP_MD_CTX_cleanup(&ctx->md_ctx);
    memset(ctx,0,sizeof *ctx);
    }

/* This call frees resources associated with the context */
int EVP_MD_CTX_cleanup(EVP_MD_CTX *ctx)
    {
#ifndef OPENSSL_FIPS
    /* Don't assume ctx->md_data was cleaned in EVP_Digest_Final,
     * because sometimes only copies of the context are ever finalised.
     */
    if (ctx->digest && ctx->digest->cleanup
        && !EVP_MD_CTX_test_flags(ctx,EVP_MD_CTX_FLAG_CLEANED))
        ctx->digest->cleanup(ctx);
    if (ctx->digest && ctx->digest->ctx_size && ctx->md_data
        && !EVP_MD_CTX_test_flags(ctx, EVP_MD_CTX_FLAG_REUSE))
        {
        OPENSSL_cleanse(ctx->md_data,ctx->digest->ctx_size);
        OPENSSL_free(ctx->md_data);
        }
#endif
    if (ctx->pctx)
        EVP_PKEY_CTX_free(ctx->pctx);
#ifndef OPENSSL_NO_ENGINE
    if(ctx->engine)
        /* The EVP_MD we used belongs to an ENGINE, release the
         * functional reference we held for this reason. */
        ENGINE_finish(ctx->engine);
#endif
#ifdef OPENSSL_FIPS
    FIPS_md_ctx_cleanup(ctx);
#endif
    memset(ctx,'\0',sizeof *ctx);

    return 1;
    }
```

### DIFF
--- a/src/_cffi_src/openssl/hmac.py
+++ b/src/_cffi_src/openssl/hmac.py
@@ -9,18 +9,17 @@ INCLUDES = """
 """
 
 TYPES = """
-typedef struct { ...; } HMAC_CTX;
+typedef ... HMAC_CTX;
 """
 
 FUNCTIONS = """
-void HMAC_CTX_init(HMAC_CTX *);
-void HMAC_CTX_cleanup(HMAC_CTX *);
-
 int Cryptography_HMAC_Init_ex(HMAC_CTX *, const void *, int, const EVP_MD *,
                               ENGINE *);
 int Cryptography_HMAC_Update(HMAC_CTX *, const unsigned char *, size_t);
 int Cryptography_HMAC_Final(HMAC_CTX *, unsigned char *, unsigned int *);
 int Cryptography_HMAC_CTX_copy(HMAC_CTX *, HMAC_CTX *);
+HMAC_CTX *Cryptography_HMAC_CTX_new(void);
+void Cryptography_HMAC_CTX_free(HMAC_CTX *ctx);
 """
 
 MACROS = """
@@ -78,6 +77,39 @@ int Cryptography_HMAC_CTX_copy(HMAC_CTX *dst_ctx, HMAC_CTX *src_ctx) {
 
     err:
         return 0;
+#endif
+}
+
+HMAC_CTX *Cryptography_HMAC_CTX_new(void) {
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L && !defined(LIBRESSL_VERSION_NUMBER)
+    return HMAC_CTX_new();
+#else
+    /* This uses OPENSSL_zalloc in 1.1.0, which is malloc + memset */
+    HMAC_CTX *ctx = (HMAC_CTX *)OPENSSL_malloc(sizeof(HMAC_CTX));
+    memset(ctx, 0, sizeof(HMAC_CTX));
+    /*if (ctx)
+        if (!HMAC_CTX_reset(ctx)) {
+            HMAC_CTX_free(ctx);
+            ctx = NULL;
+        }*/
+    return ctx;
+#endif
+}
+
+
+
+void Cryptography_HMAC_CTX_free(HMAC_CTX *ctx) {
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L && !defined(LIBRESSL_VERSION_NUMBER)
+    return HMAC_CTX_free(ctx);
+#else
+    if (ctx != NULL) {
+        ctx->md = NULL;
+        ctx->key_length = 0;
+        memset(ctx->key, 0, sizeof(HMAC_MAX_MD_CBLOCK));
+
+        HMAC_CTX_cleanup(ctx);
+        OPENSSL_free(ctx);
+    }
 #endif
 }
 """
--- a/src/_cffi_src/openssl/hmac.py
+++ b/src/_cffi_src/openssl/hmac.py
@@ -87,11 +87,6 @@ HMAC_CTX *Cryptography_HMAC_CTX_new(void) {
     /* This uses OPENSSL_zalloc in 1.1.0, which is malloc + memset */
     HMAC_CTX *ctx = (HMAC_CTX *)OPENSSL_malloc(sizeof(HMAC_CTX));
     memset(ctx, 0, sizeof(HMAC_CTX));
-    /*if (ctx)
-        if (!HMAC_CTX_reset(ctx)) {
-            HMAC_CTX_free(ctx);
-            ctx = NULL;
-        }*/
     return ctx;
 #endif
 }
--- a/src/_cffi_src/openssl/hmac.py
+++ b/src/_cffi_src/openssl/hmac.py
@@ -103,10 +103,6 @@ void Cryptography_HMAC_CTX_free(HMAC_CTX *ctx) {
     return HMAC_CTX_free(ctx);
 #else
     if (ctx != NULL) {
-        ctx->md = NULL;
-        ctx->key_length = 0;
-        memset(ctx->key, 0, sizeof(HMAC_MAX_MD_CBLOCK));
-
         HMAC_CTX_cleanup(ctx);
         OPENSSL_free(ctx);
     }
--- a/src/cryptography/hazmat/backends/openssl/hmac.py
+++ b/src/cryptography/hazmat/backends/openssl/hmac.py
@@ -20,10 +20,10 @@ class _HMACContext(object):
         self._backend = backend
 
         if ctx is None:
-            ctx = self._backend._ffi.new("HMAC_CTX *")
-            self._backend._lib.HMAC_CTX_init(ctx)
+            ctx = self._backend._lib.Cryptography_HMAC_CTX_new()
+            self._backend.openssl_assert(ctx != self._backend._ffi.NULL)
             ctx = self._backend._ffi.gc(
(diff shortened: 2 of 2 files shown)


src/_cffi_src/openssl/hmac.py
```diff
@@ -80,4 +79,37 @@
...
+    return HMAC_CTX_new();
+#else
+    /* This uses OPENSSL_zalloc in 1.1.0, which is malloc + memset */
+    HMAC_CTX *ctx = (HMAC_CTX *)OPENSSL_malloc(sizeof(HMAC_CTX));
+    memset(ctx, 0, sizeof(HMAC_CTX));
+    /*if (ctx)
+        if (!HMAC_CTX_reset(ctx)) {
+            HMAC_CTX_free(ctx);
+            ctx = NULL;
+        }*/
```
alex: Err, what's this commented out stuff?

src/_cffi_src/openssl/hmac.py
```diff
@@ -80,4 +79,37 @@
...
+
+
+void Cryptography_HMAC_CTX_free(HMAC_CTX *ctx) {
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L && !defined(LIBRESSL_VERSION_NUMBER)
+    return HMAC_CTX_free(ctx);
+#else
+    if (ctx != NULL) {
+        ctx->md = NULL;
+        ctx->key_length = 0;
+        memset(ctx->key, 0, sizeof(HMAC_MAX_MD_CBLOCK));
```
alex: `HMAC_CTX_cleanup` does all 3 of these

src/_cffi_src/openssl/hmac.py
reaperhulk: It looks like I commented it out because I'm dumb? That was the code as of 1.1.0-pre1. Current git master uses the below so I'll put that in:

```c
    if (ctx != NULL) {
        if (!HMAC_CTX_reset(ctx)) {
            HMAC_CTX_free(ctx);
            return NULL;
        }
    }
```

src/_cffi_src/openssl/hmac.py
reaperhulk: It does? I see the memset of the whole struct, which pretty much obviates the need to do any of these of course. Is that what you mean?

alex: Yeah, memsetting the memory of ctx to 0 is inclusing of setting md to NULL
and key_length to 0.

On Thu, Mar 10, 2016 at 8:35 AM, Paul Kehrer <notifications@github.com>
wrote:

> In src/_cffi_src/openssl/hmac.py
> <https://github.com/pyca/cryptography/pull/2794#discussion_r55678659>:
>
> > +            ctx = NULL;
> > +        }*/
> > +    return ctx;
> > +#endif
> > +}
> > +
> > +
> > +
> > +void Cryptography_HMAC_CTX_free(HMAC_CTX *ctx) {
> > +#if OPENSSL_VERSION_NUMBER >= 0x10100000L && !defined(LIBRESSL_VERSION_NUMBER)
> > +    return HMAC_CTX_free(ctx);
> > +#else
> > +    if (ctx != NULL) {
> > +        ctx->md = NULL;
> > +        ctx->key_length = 0;
> > +        memset(ctx->key, 0, sizeof(HMAC_MAX_MD_CBLOCK));
>
> It does? I see the memset of the whole struct, which pretty much obviates
> the need to do any of these of course. Is that what you mean?
>
> —
> <https://github.com/pyca/cryptography/pull/2794/files#r55678659>.
>



-- 

reaperhulk: :+1:

src/_cffi_src/openssl/hmac.py
reaperhulk: Oh, nevermind. I'm less dumb than I thought (but still dumb for leaving this here). So in 1.1.0 HMAC_CTX is opaque and most of the struct members are pointers:
```c
struct hmac_ctx_st {
    const EVP_MD *md;
    EVP_MD_CTX *md_ctx;
    EVP_MD_CTX *i_ctx;
    EVP_MD_CTX *o_ctx;
    unsigned int key_length;
    unsigned char key[HMAC_MAX_MD_CBLOCK];
};
```

In 1.0.2 and below they are directly in the HMAC_CTX struct:
```c
typedef struct hmac_ctx_st
    {
    const EVP_MD *md;
    EVP_MD_CTX md_ctx;
    EVP_MD_CTX i_ctx;
    EVP_MD_CTX o_ctx;
    unsigned int key_length;
    unsigned char key[HMAC_MAX_MD_CBLOCK];
    } HMAC_CTX;
```

So to initialize an HMAC_CTX in 1.1.0 you need to call that HMAC_CTX_reset function, but in 1.0.2 and below just memsetting the allocated memory to zero is sufficient before calling the init functions. So the right thing to do is just delete the commented out lines, as they serve no purpose in older OpenSSL.

src/_cffi_src/openssl/hmac.py
```diff
@@ -80,4 +79,28 @@
...
 #endif
 }
+
+HMAC_CTX *Cryptography_HMAC_CTX_new(void) {
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L && !defined(LIBRESSL_VERSION_NUMBER)
+    return HMAC_CTX_new();
+#else
+    /* This uses OPENSSL_zalloc in 1.1.0, which is malloc + memset */
+    HMAC_CTX *ctx = (HMAC_CTX *)OPENSSL_malloc(sizeof(HMAC_CTX));
+    memset(ctx, 0, sizeof(HMAC_CTX));
```
alex: Is there a reason not to just use `HMAC_CTX_init(ctx)` here?

reaperhulk: HMAC_CTX_init calls EVP_MD_CTX_init on `md_ctx`, `i_ctx` and `o_ctx`. It doesn't set initial values for `md`, `key_length`, or `key`.

PR #2798, opened 2016-03-10
opaque X509_EXTENSION

reaperhulk:
`X509_EXTENSION_get_object` and `X509_EXTENSION_get_data` in this PR *do not* increment refcount. The memory they return is tied to the lifetime of the parent object (which should already be registered for gc).

### DIFF
--- a/src/_cffi_src/openssl/x509.py
+++ b/src/_cffi_src/openssl/x509.py
@@ -36,7 +36,13 @@ typedef struct {
     ...;
 } X509_CINF;
 
-typedef ... X509_EXTENSION;
+/* TODO: opaque X509_EXTENSION. Cryptography no longer depends on it being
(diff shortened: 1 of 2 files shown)


reaperhulk:
Bleh, this will break pyopenssl because it reaches into the struct in one place. https://github.com/pyca/pyopenssl/blob/e6eac51b94b8172f5d398cb561e83c5327f5690c/src/OpenSSL/crypto.py#L1875

We shouldn't merge this for now (as I'd like to not break anything for the 1.3 release).

reaperhulk:
I've modified this PR to make the changes required so that cryptography itself no longer needs X509_EXTENSION to be non-opaque, but left the struct fully defined to avoid problems with pyOpenSSL. I'll do this for all the X509 structs so we can have most of the 1.1.0 support code landed before finally flipping the switch and opaquing them all.

PR #2802, opened 2016-03-11
move crypto_ex_data to macros, add i2d_re_X509_tbs & X509_get0_signature

reaperhulk:
And, of course, use them in the openssl bindings. These changes are a start towards opaquing all the X509 structs. The actual opaquing won't take place until the very end though to minimize pyOpenSSL breakage.

To ease review, here are the definitions for the added functions:

[X509_get0_signature](https://github.com/openssl/openssl/blob/902f3f50d051dfd6ebf009d352aaf581195caabf/crypto/asn1/x_x509.c#L217) (added in 1.0.2-beta1, which is 0x10002001L) 

[i2d_re_X509_tbs](https://github.com/openssl/openssl/blob/902f3f50d051dfd6ebf009d352aaf581195caabf/crypto/asn1/x_x509.c#L211) (not added until 1.0.2-beta3, which is 0x10002003L)

### DIFF
--- a/src/_cffi_src/openssl/x509.py
+++ b/src/_cffi_src/openssl/x509.py
@@ -152,12 +152,6 @@ X509_EXTENSION *X509_EXTENSION_dup(X509_EXTENSION *);
 X509_EXTENSION *X509_get_ext(X509 *, int);
 int X509_get_ext_by_NID(X509 *, int, int);
 
-/* CRYPTO_EX_DATA */
-int X509_get_ex_new_index(long, void *, CRYPTO_EX_new *, CRYPTO_EX_dup *,
-                          CRYPTO_EX_free *);
-int X509_set_ex_data(X509 *, int, void *);
-void *X509_get_ex_data(X509 *, int);
-
 int X509_EXTENSION_get_critical(X509_EXTENSION *);
 ASN1_OBJECT *X509_EXTENSION_get_object(X509_EXTENSION *);
 void X509_EXTENSION_free(X509_EXTENSION *);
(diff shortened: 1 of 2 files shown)


src/_cffi_src/openssl/x509.py
```diff
@@ -350,6 +354,33 @@
...
+}
+#endif
+/* Added in 1.0.2 but we need it in all versions now due to the great
+   opaquing. */
+#if OPENSSL_VERSION_NUMBER < 0x10002003L || defined(LIBRESSL_VERSION_NUMBER)
+/* from x509/x_x509.c */
+int i2d_re_X509_tbs(X509 *x, unsigned char **pp)
+{
+    /* In 1.0.1 and below cert_info is a pointer in the struct, so
+       we don't want to pass by reference. */
```
alex: I don't understand this comment.

src/_cffi_src/openssl/x509.py
```diff
@@ -350,6 +354,33 @@
...
+    /* ideally we also call x->cert_info->enc.modified = 1 as 1.0.2+ does, but
+       older OpenSSLs don't have the enc ASN1_ENCODING on the struct */
```
alex: a) This isn't a call, b) this comment ought to explain what the impact of this is

src/_cffi_src/openssl/x509.py
reaperhulk: That's because it makes little sense. I just tried to figure out what I meant. I *think* I was trying to say that in 1.1.0 the opaque X509 struct makes cert_info a direct member of the struct rather than a pointer. So in 1.1.0 the function definition looks like `return i2d_X509_CINF(&x->cert_info, pp);`. But this isn't a particularly useful comment so I should just remove it.

PR #2812, opened 2016-03-12
make multibackend error with an empty list.

### LINKED ISSUE #2717: Throw a specific exception when no backends are available
In e.g. #1958, an exception is thrown that in some cases appears to be due to the fact that no backends are available. It might be nice if in these cases a specific exception would be thrown.

reaperhulk:
fixes #2717

Open to better language for the exception (and should it be a ValueError?)

### DIFF
--- a/src/cryptography/hazmat/backends/multibackend.py
+++ b/src/cryptography/hazmat/backends/multibackend.py
@@ -31,7 +31,7 @@ class MultiBackend(object):
         if len(backends) == 0:
             raise ValueError(
                 "Multibackend cannot be initialized with no backends. If you "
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/backends/multibackend.py
```diff
@@ -28,6 +28,13 @@ class MultiBackend(object):
     name = "multibackend"
 
     def __init__(self, backends):
+        if len(backends) == 0:
+            raise ValueError(
+                "Multibackend cannot be initialized with no backends. If you "
+                "are seeing this error when trying to use default_backend "
```
alex: Can you put parens after `default_backend`?

PR #2813, opened 2016-03-12
Add AuthorityKeyIdentifier.from_issuer_subject_key_identifier

### LINKED ISSUE #2758: x509.AuthorityKeyIdentifier documented use is buggy
[Docs here](https://cryptography.io/en/latest/x509/reference/#cryptography.x509.AuthorityKeyIdentifier.from_issuer_public_key)

used in some of my own code like so:
```python
builder.add_extension(
    x509.AuthorityKeyIdentifier.from_issuer_public_key(ca_cert.public_key()), critical=False
    )
```
https://gist.github.com/jrabbit/e84b3c593e18bcf7d2dd#file-cert-py-L57
output cert fails to verify in `openssl x509 verify` & `certtool --verify`

the AKI it computes seems to be the SKI?

reaperhulk:
Fixes #2758

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,8 @@ Changelog
   to :class:`~cryptography.x509.CertificateSigningRequest`.
 * Fixed an intermittent ``AssertionError`` when performing an RSA decryption on
   an invalid ciphertext, ``ValueError`` is now correctly raised in all cases.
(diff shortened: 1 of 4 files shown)


alex:
Seems like there should be something to indicate which is prefere, this or AKI from public key?

Issue #2724, opened 2016-02-15
certificate.subject for certificate having numericstring values results in InternalError: Unknown OpenSSL error 


daa:
I have certificate with some subject's field have `NUMERICSTRING` value type - `1.2.643.100.1` and `1.2.643.3.131.1.1`, and when I try to get its `subject` property, cryptography raises `InternalError` exception:

```
Python 2.7.6 (default, Mar 22 2014, 22:59:56) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import cryptography
>>> cryptography.__version__
'1.2.2'
>>> from cryptography import x509
>>> from cryptography.hazmat.backends import default_backend
>>> with open('example.der', 'rb') as f:
...     cert = x509.load_der_x509_certificate(f.read(), default_backend())
... 
>>> cert.subject
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/x509.py", line 324, in subject
    return _decode_x509_name(self._backend, subject)
  File ".../lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/x509.py", line 52, in _decode_x509_name
    attributes.append(_decode_x509_name_entry(backend, entry))
  File ".../lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/x509.py", line 42, in _decode_x509_name_entry
    value = backend._asn1_string_to_utf8(data)
  File ".../lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 2255, in _asn1_string_to_utf8
    self.openssl_assert(res >= 0)
  File ".../lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 719, in openssl_assert
    return binding._openssl_assert(self._lib, ok)
  File ".../lib/python2.7/site-packages/cryptography/hazmat/bindings/openssl/binding.py", line 43, in _openssl_assert
    errors
cryptography.exceptions.InternalError: Unknown OpenSSL error. Please file an issue at https://github.com/pyca/cryptography/issues with information on how to reproduce this. ([])
```

It happens because OpenSSL's function `ASN1_STRING_to_UTF8` does not handle `NUMERICSTRING` type and returns `-1`, so `cryptography` fails. May be values with such types should be handled separately.
[example.zip](https://github.com/pyca/cryptography/files/130942/example.zip)

[...]

PR #2815, opened 2016-03-12
Error on unusual encodings

reaperhulk:
When encountering asn1 string types that can't be parsed via `ASN1_STRING_to_UTF8` let's error out with a TypeError for now.

Depends on #2814, refs #2724

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/decode_asn1.py
+++ b/src/cryptography/hazmat/backends/openssl/decode_asn1.py
@@ -710,7 +710,7 @@ def _asn1_string_to_utf8(backend, asn1_string):
     buf = backend._ffi.new("unsigned char **")
     res = backend._lib.ASN1_STRING_to_UTF8(buf, asn1_string)
     if res == -1:
-        raise TypeError(
+        raise ValueError(
             "Unsupported ASN1 string type. Type: {0}".format(asn1_string.type)
         )
 
--- a/src/cryptography/hazmat/backends/openssl/decode_asn1.py
+++ b/src/cryptography/hazmat/backends/openssl/decode_asn1.py
@@ -709,7 +709,11 @@ def _asn1_string_to_ascii(backend, asn1_string):
 def _asn1_string_to_utf8(backend, asn1_string):
     buf = backend._ffi.new("unsigned char **")
     res = backend._lib.ASN1_STRING_to_UTF8(buf, asn1_string)
-    backend.openssl_assert(res >= 0)
+    if res == -1:
+        raise TypeError(
+            "Unsupported ASN1 string type. Type: {0}".format(asn1_string.type)
+        )
+
     backend.openssl_assert(buf[0] != backend._ffi.NULL)
     buf = backend._ffi.gc(
         buf, lambda buffer: backend._lib.OPENSSL_free(buffer[0])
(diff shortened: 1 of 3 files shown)


alex:
Ready for a rebase

src/cryptography/hazmat/backends/openssl/decode_asn1.py
```diff
@@ -709,7 +709,11 @@ def _asn1_string_to_ascii(backend, asn1_string):
 def _asn1_string_to_utf8(backend, asn1_string):
     buf = backend._ffi.new("unsigned char **")
     res = backend._lib.ASN1_STRING_to_UTF8(buf, asn1_string)
-    backend.openssl_assert(res >= 0)
+    if res == -1:
+        raise TypeError(
```
alex: Should this be a `ValueError`? `TypeError` is usually on a "user called the function with the wrong type".

reaperhulk:
OpenSSL on CentOS5 can't load this certificate at all. Not sure what to do about that. I guess we could skip the test if we're on 0.9.8?

alex:
Do we know what about that OpenSSL makes it fail?

On Sat, Mar 12, 2016 at 8:31 AM, Paul Kehrer <notifications@github.com>
wrote:

> OpenSSL on CentOS5 can't load this certificate at all. Not sure what to do
> about that. I guess we could skip the test if we're on 0.9.8?
>
> —
> <https://github.com/pyca/cryptography/pull/2815#issuecomment-195742834>.
>



-- 

reaperhulk:
Looks like 'error:0D07808C:asn1 encoding routines:ASN1_ITEM_EX_D2I:mstring wrong tag'

alex:
The test is in `test_x509.py` so there's not really a good way to skip on older OpenSSL only :-/

reaperhulk:
An ugly hack but potentially workable:

```python
def skip_if_openssl_098(backend):
    if backend.name == "openssl" && backend._lib.OPENSSL_VERSION_NUMBER < 0x0100000fL:
        pytest.skip("This test requires OpenSSL 1.0.0 or newer")
```

I've got mixed feelings about putting backend specific skip behavior in though...

alex:
You could also put this test in `test_openssl.py`?

On Sat, Mar 12, 2016 at 11:34 AM, Paul Kehrer <notifications@github.com>
wrote:

> An ugly hack but potentially workable:
>
> def skip_if_openssl_098(backend):
>     if backend.name == "openssl" && backend._lib.OPENSSL_VERSION_NUMBER < 0x0100000fL:
>         pytest.skip("This test requires OpenSSL 1.0.0 or newer")
>
> I've got mixed feelings about putting backend specific skip behavior in
> though...
>
> —
> <https://github.com/pyca/cryptography/pull/2815#issuecomment-195769813>.
>



-- 

reaperhulk:
We could, but if/when we implement X509Backend on another backend we're likely to have the same "unusual ASN1String type" bug since it won't run a test using NUMERICSTRING. (Of course, maybe the backend has the same problem older OpenSSL has so then it wouldn't matter?)

I guess I'm okay with moving it for now and adding a versioned skip there.

alex:
:+1:

On Sat, Mar 12, 2016 at 11:38 AM, Paul Kehrer <notifications@github.com>
wrote:

> We could, but if/when we implement X509Backend on another backend we're
> likely to have the same "unusual ASN1String type" bug since it won't run a
> test using NUMERICSTRING. (Of course, maybe the backend has the same
> problem older OpenSSL has so then it wouldn't matter?)
>
> I guess I'm okay with moving it for now and adding a versioned skip there.
>
> —
> <https://github.com/pyca/cryptography/pull/2815#issuecomment-195770250>.
>



-- 

PR #2817, opened 2016-03-12
add two faq entries for common errors

### LINKED ISSUE #2816: Add documentation about mod_wsgi workaround
From #2774:

We should have a followup PR to add some language about mod_wsgi failures that links to a FAQ entry describing %{GLOBAL} as a workaround. (Or maybe we catch failures during engine initialization and raise a different error)

reaperhulk:
fixes #2816

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -23,5 +23,20 @@ OpenSSL. You're seeing this error because your copy of pip is too old to find
 our wheel files. Upgrade your copy of pip with ``pip install -U pip`` and then
 try install ``cryptography`` again.
 
+Starting ``cryptography`` using ``mod_wsgi`` produces an ``InternalError`` during a call in ``_register_osrandom_engine``
+-------------------------------------------------------------------------------------------------------------------------
+
+This happens because ``mod_wsgi`` uses sub-interpreters, which can cause a
+problem during initialization of the OpenSSL backend. To resolve this set the
+`WSGIApplicationGroup`_ to ``%{GLOBAL}`` in the ``mod_wsgi`` configuration.
+
+``cryptography`` raised an ``InternalError`` and I'm not sure what to do?
+-------------------------------------------------------------------------
+
+Frequently ``InternalError`` is raised when there are errors on the OpenSSL
+error stack that were placed there by other libraries that are also using
+OpenSSL. Try removing the other libraries and see if the problem persists.
+
 .. _`NaCl`: https://nacl.cr.yp.to/
 .. _`PyNaCl`: https://pynacl.readthedocs.org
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -23,5 +23,20 @@ OpenSSL. You're seeing this error because your copy of pip is too old to find
...
+This happens because ``mod_wsgi`` uses sub-interpreters, which can cause a
+problem during initialization of the OpenSSL backend. To resolve this set the
+`WSGIApplicationGroup`_ to ``%{GLOBAL}`` in the ``mod_wsgi`` configuration.
+
+``cryptography`` raised an ``InternalError`` and I'm not sure what to do?
+-------------------------------------------------------------------------
+
+Frequently ``InternalError`` is raised when there are errors on the OpenSSL
+error stack that were placed there by other libraries that are also using
+OpenSSL. Try removing the other libraries and see if the problem persists.
```
alex: Should we mention that the other alternative is a cryptography bug?

reaperhulk: Yeah probably. How about "If the error continues to occur please file an issue with pyca/cryptography providing whatever information we'll need to reproduce it."

alex: "If you have no other libraries using OpenSSL in your process, or they do
not appear to be at fault, it's possible that this is a bug in
``cryptography``, please file an issue with instructions on how to
reproduce it."

On Sat, Mar 12, 2016 at 3:30 PM, Paul Kehrer <notifications@github.com>
wrote:

> In docs/faq.rst
> <https://github.com/pyca/cryptography/pull/2817#discussion_r55927221>:
>
> > @@ -23,5 +23,20 @@ OpenSSL. You're seeing this error because your copy of pip is too old to find
> >  our wheel files. Upgrade your copy of pip with ``pip install -U pip`` and then
> >  try install ``cryptography`` again.
> >
> > +Starting ``cryptography`` using ``mod_wsgi`` produces an ``InternalError`` during a call in ``_register_osrandom_engine``
> > +-------------------------------------------------------------------------------------------------------------------------
> > +
> > +This happens because ``mod_wsgi`` uses sub-interpreters, which can cause a
> > +problem during initialization of the OpenSSL backend. To resolve this set the
> > +`WSGIApplicationGroup`_ to ``%{GLOBAL}`` in the ``mod_wsgi`` configuration.
> > +
> > +``cryptography`` raised an ``InternalError`` and I'm not sure what to do?
> > +-------------------------------------------------------------------------
> > +
> > +Frequently ``InternalError`` is raised when there are errors on the OpenSSL
> > +error stack that were placed there by other libraries that are also using
> > +OpenSSL. Try removing the other libraries and see if the problem persists.
>
> Yeah probably. How about "If the error continues to occur please file an
> issue with pyca/cryptography providing whatever information we'll need to
> reproduce it."
>
> —
> <https://github.com/pyca/cryptography/pull/2817/files#r55927221>.
>



-- 

PR #2824, opened 2016-03-14
remove some EC functions we don't use

reaperhulk:


alex:
The deleted functions need to be removed from `_conditional.py`

PR #2840, opened 2016-03-19
Fixed #2836 -- error out on OpenSSL 0.9.8 by default

### LINKED ISSUE #2836: Raise an error on openssl 0.9.8
With an env var (probably `CRYPTOGRAPHY_ALLOW_OPENSSL_098`) to allow disabling it for this one release.

alex:
- [x] Get tests passing on Travis and Jenkins by setting `CRYPTOGRAPHY_ALLOW_OPENSSL_098`
- [x] Figure out which exception type to use for the "you are trying to use openssl 0.9.8" error
- [x] Figure out what to do in `installation.rst`
- [x] Do we document `CRYPTOGRAPHY_ALLOW_OPENSSL_098`?
- [x] Write tests for `_verify_openssl_version` so that we have full coverage

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -46,7 +46,7 @@ Importing cryptography causes a ``RuntimeError`` about OpenSSL 0.9.8
 The OpenSSL project has dropped support for the 0.9.8 release series. Since it
 is no longer receiving security patches from upstream, ``cryptography`` is also
 dropping support for it. To fix this issue you should upgrade to a newer
-version of OpenSSL (1.0.1 or later), this may require you to upgrade to a newer
+version of OpenSSL (1.0.1 or later). This may require you to upgrade to a newer
 operating system.
 
 For the 1.4 release, you can set the ``CRYPTOGRAPHY_ALLOW_OPENSSL_098``
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -40,6 +40,19 @@ If you have no other libraries using OpenSSL in your process, or they do not
 appear to be at fault, it's possible that this is a bug in ``cryptography``.
 Please file an `issue`_ with instructions on how to reproduce it.
 
+Importing cryptography causes a ``RuntimeError`` about OpenSSL 0.9.8
+--------------------------------------------------------------------
+
+The OpenSSL project has dropped support for the 0.9.8 release series. Since it
+is no longer receiving security patches from upstream, ``cryptography`` is also
+dropping support for it. To fix this issue you should upgrade to a newer
+version of OpenSSL (1.0.1 or later), this may require you to upgrade to a newer
+operating system.
+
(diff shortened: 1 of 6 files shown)


reaperhulk:
Checklist looks good to me.

alex:
Ok, I'll start working on the easy ones. Would appreciate some input on the ones that are open questions.

reaperhulk:
I'm okay with documenting `CRYPTOGRAPHY_ALLOW_OPENSSL_098` on the FAQ page as a "this will work for one version only" kind of workaround.

What about `RuntimeError` or `SystemError` for the actual error? The latter is documented as "Raised when the interpreter finds an internal error, but the situation does not look so serious to cause it to abandon all hope", which doesn't quite fit what we're looking for here even though the name is pretty good.

alex:
`RuntimeError` sounds ok.

alex:
@reaperhulk do you have an opinion about what to do fro `installation.rst`?

reaperhulk:
Replace the 0.9.8 warning with something that says "As of cryptography 1.4 0.9.8 is no longer supported. See <faq link> for more details"? I'd probably leave the test matrix for now since we will still be testing against it for one more release.

docs/faq.rst
```diff
@@ -40,6 +40,19 @@ If you have no other libraries using OpenSSL in your process, or they do not
 appear to be at fault, it's possible that this is a bug in ``cryptography``.
 Please file an `issue`_ with instructions on how to reproduce it.
 
+Importing cryptography causes a ``RuntimeError`` about OpenSSL 0.9.8
+--------------------------------------------------------------------
+
+The OpenSSL project has dropped support for the 0.9.8 release series. Since it
+is no longer receiving security patches from upstream, ``cryptography`` is also
+dropping support for it. To fix this issue you should upgrade to a newer
+version of OpenSSL (1.0.1 or later), this may require you to upgrade to a newer
```
reaperhulk: Should be a period instead of a comma here.

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -217,6 +217,30 @@ def _lock_cb(cls, mode, n, file, line):
...
 
 
+def _verify_openssl_version(version):
+    if version < 0x10000000:
+        if os.environ.get("CRYPTOGRAPHY_ALLOW_OPENSSL_098"):
+            warnings.warn(
+                "OpenSSL version 0.9.8 is no longer supported by the OpenSSL "
+                "project, please upgrade. The next version of cryptography "
+                "will completely remove support for it.",
+                utils.DeprecatedIn12
```
reaperhulk: This should probably become `DeprecationWarning` because the other instance of `DeprecatedIn12` is going to be removed in 1.4 per our standard policy.

PR #2843, opened 2016-03-21
add a note on the install page about pip versioning on OS X

reaperhulk:
and update the pip version needed to get wheels.

Turns out many questions keep appearing on Stack Overflow and people link to our install docs so we should cross link this FAQ issue to make it easier to figure out why users are seeing this install error.


alex:
will merge whenever codecov finishes

PR #2845, opened 2016-03-21
Add __delattr__ to _ModuleWithDeprecations.

### LINKED ISSUE #2844: _ModuleWithDeprecations doesn't handle patching properly.
`_ModuleWithDeprecations` catches `__getattr__` and `__setattr__` to patch through to the underlying module, but does not intercept `__delattr__`. That means that if you're using something like `mock.patch`, the mock successfully lands in place, but cannot be removed: the mock was applied to the underlying module, but the delete comes from the proxy.

Should be easily fixed.

Lukasa:
Resolves #2844.

### DIFF
--- a/src/cryptography/utils.py
+++ b/src/cryptography/utils.py
@@ -120,6 +120,10 @@ class _ModuleWithDeprecations(object):
         setattr(self._module, attr, value)
 
     def __delattr__(self, attr):
+        obj = getattr(self._module, attr)
+        if isinstance(obj, _DeprecatedValue):
+            warnings.warn(obj.message, obj.warning_class, stacklevel=2)
+
         delattr(self._module, attr)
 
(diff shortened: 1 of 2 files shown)


Lukasa:
Test failure is spurious.

src/cryptography/utils.py
```diff
@@ -119,6 +119,9 @@ def __getattr__(self, attr):
     def __setattr__(self, attr, value):
         setattr(self._module, attr, value)
 
+    def __delattr__(self, attr):
+        delattr(self._module, attr)
```
alex: This should raise a warning if the attribute is deprecated, since deleting it won't work shortly.

alex:
Needs tests.

reaperhulk:
:+1: thanks @Lukasa

PR #2855, opened 2016-03-26
Remove redundant excludes from find_packages

### LINKED ISSUE #2854: Redundant exclude in setup.py's find_packages
I think the call can be reduced from 

```python
find_packages(
    where="src", exclude=["_cffi_src", "_cffi_src.*", "tests", "tests.*"]
)
```

to
```python
find_packages(where="src", exclude=["_cffi_src", "_cffi_src.*"])
```

because of the `where="src"`. I verified by printing the output from setup.py

RazerM:
Resolves #2854.

tests don't need excluded because they're not within the "src" directory.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -302,9 +302,7 @@ setup(
     ],
 
     package_dir={"": "src"},
-    packages=find_packages(
-        where="src", exclude=["_cffi_src", "_cffi_src.*", "tests", "tests.*"]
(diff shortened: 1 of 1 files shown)


reaperhulk:
jenkins, ok to test

reaperhulk:
looks good to me, thanks for the contribution!

PR #2857, opened 2016-03-27
Fixes #2856 -- add a fast path to _key_identifier_from_public_key for RSA keys

### LINKED ISSUE #2856: _key_identifier_from_public_key is pretty slow
Sure, okay, it even has a comment warning about this: https://github.com/pyca/cryptography/blob/master/src/cryptography/x509/extensions.py#L35

I just ran into this in the context of doing some testing; the bulk of the runtime in the test is issuing certificates in my fake CA, as part of that I'm using `SubjectKeyIdentifier.from_public_key` and `AuthorityKeyIdentifier.from_issuer_public_key`; I can cache the latter, but not the former.

alex:
And I'm a reasonably certain this is correct :-)

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -17,6 +17,7 @@ import six
 
 from cryptography import utils
 from cryptography.hazmat.primitives import constant_time, serialization
+from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey
 from cryptography.x509.general_name import GeneralName, IPAddress, OtherName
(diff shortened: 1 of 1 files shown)


alex:
Appears to be ~100x faster (o_O)

mithrandi:
I can confirm that this speeds things up a lot; only for RSA keys, of course.

reaperhulk:
RSA path looks good, haven't looked at the EC path yet.

reaperhulk:
EC path :+1: as well.

PR #2862, opened 2016-03-29
add ChaCha20Poly1305 test vectors from BoringSSL

reaperhulk:
We'll be able to support this when we merge OpenSSL 1.1.0 support, but we shouldn't merge until we decide how/if we're going to implement a one shot API to support this.

Vectors are from BoringSSL: https://boringssl.googlesource.com/boringssl/+/fef6fb592bb8277a555818c4152c9ebfcf9e5533/crypto/cipher/test/chacha20_poly1305_tests.txt


alex:
We want to start with separate vectors for ChaCha20 and Poly1305.

PR #2874, opened 2016-04-21
Make a paragraph slightly more clear

alex:


reaperhulk:
Super confused why codecov already posted a message to this PR...

reaperhulk:
The coverage on this is nonsensical as it appears to be off the wrong base. Merging despite that since it's clearly wrong.

PR #2875, opened 2016-04-22
Test with both old pypy and new pypy

alex:


.travis.yml
```diff
@@ -21,7 +21,9 @@ matrix:
         - python: 3.5
           env: TOXENV=py35
         - python: pypy
-          env: TOXENV=pypy
+          env: TOXENV=pypy PYPY_VERSION=4.0.1
```
reaperhulk: I believe we actually claim to support pypy 2.6, so if we want oldest/newest this should probably become 2.6?

docs/conf.py
```diff
@@ -176,4 +176,4 @@
 
 # Retry requests in the linkcheck builder so that we're resillient against
 # transient network errors.
-linkcheck_retries = 2
+linkcheck_retries = 3
```
reaperhulk: This looks like it belongs in another PR.

reaperhulk:
Aww damn. We need OldPyPy coverage for the non-static callback path. Maybe we need 4.0.1 for that still then?

alex:
Let's do all 3 I guess

PR #2877, opened 2016-04-24
Add OpenSSL bindings for blake2b and blake2s.

Ayrx:
This PR adds the bindings for Blake2b and Blake2s.


Ayrx:
@reaperhulk Think this can be merged first to lower the PR size of the next patch or do we want to wait for a complete implementation?

reaperhulk:
We shouldn't need this at all for unkeyed blake2, as it should be obtainable via `EVP_getdigestbyname`. Implementing blake2 at this point should mostly just be adding a `Blake2b` and `Blake2s` instance in hashes that have the right name for the hash backend to look up.

PR #2881, opened 2016-04-29
opaque BIO_METHOD and BIO. Move some functions that added const

reaperhulk:
pyOpenSSL treats these as opaque structs so we can actually do this safely.

### DIFF
--- a/src/_cffi_src/openssl/bio.py
+++ b/src/_cffi_src/openssl/bio.py
@@ -11,37 +11,8 @@ INCLUDES = """
 TYPES = """
 typedef struct bio_st BIO;
 typedef void bio_info_cb(BIO *, int, const char *, int, long, long);
-struct bio_method_st {
-    int type;
-    const char *name;
-    int (*bwrite)(BIO *, const char *, int);
-    int (*bread)(BIO *, char *, int);
-    int (*bputs)(BIO *, const char *);
-    int (*bgets)(BIO *, char *, int);
-    long (*ctrl)(BIO *, int, long, void *);
-    int (*create)(BIO *);
-    int (*destroy)(BIO *);
-    long (*callback_ctrl)(BIO *, int, bio_info_cb *);
-    ...;
-};
-typedef struct bio_method_st BIO_METHOD;
-struct bio_st {
-    BIO_METHOD *method;
-    long (*callback)(struct bio_st *, int, const char *, int, long, long);
-    char *cb_arg;
-    int init;
-    int shutdown;
-    int flags;
-    int retry_reason;
-    int num;
-    void *ptr;
-    struct bio_st *next_bio;
-    struct bio_st *prev_bio;
-    int references;
-    unsigned long num_read;
-    unsigned long num_write;
-    ...;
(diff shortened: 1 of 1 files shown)


src/_cffi_src/openssl/bio.py
```diff
@@ -11,37 +11,8 @@
...
-    int num;
-    void *ptr;
-    struct bio_st *next_bio;
-    struct bio_st *prev_bio;
-    int references;
-    unsigned long num_read;
-    unsigned long num_write;
-    ...;
-};
+typedef ... bio_st;
```
alex: Why does `bio_st` need to be exposed at all?

src/_cffi_src/openssl/bio.py
```diff
@@ -11,37 +11,8 @@
 TYPES = """
 typedef struct bio_st BIO;
```
alex: I think this should just be `typedef ... BIO`?

reaperhulk: `CRYPTO_mem_leaks(struct bio_st *);` is (obnoxiously) a function signature. We can do this if we want to move this function to MACROS and change its definition though.

alex: wtf? ok, that's fine I guess, maybe send a PR to upstream that seems like
an oversight?

On Fri, Apr 29, 2016 at 9:31 AM, Paul Kehrer <notifications@github.com>
wrote:

> In src/_cffi_src/openssl/bio.py
> <https://github.com/pyca/cryptography/pull/2881#discussion_r61577736>:
>
> > @@ -11,37 +11,8 @@
> >  TYPES = """
> >  typedef struct bio_st BIO;
>
> CRYPTO_mem_leaks(struct bio_st *); is (obnoxiously) a function signature.
> We can do this if we want to move this function to MACROS and change its
> definition though.
>
> —
> <https://github.com/pyca/cryptography/pull/2881/files/5c6fdf6dc6bdfa65259a69cd2f10f244e53c6fe8#r61577736>
>



-- 

alex:
@stevepeak Was there possibly a regression with codecov updating comments/status checks? The page on for this PR on codecov.io shows 100% coverage, but it's not reflected in the comment/status.

(The new UI looks _amazing_ btw! Congrats on shippnig!)

stevepeak:
Thank you @alex :) Sorry for the radio silence, had a number of issues to fix up.

I'll jump into see whats up here. Thank you!

alex:
@stevepeak Thanks! Look's like the comment is up to date now, but not the status check.

stevepeak:
I restarted the backend job. So there is some issue server side I need to check.

The status: Looks like its actually posted on the [merge commit](https://api.github.com/repos/pyca/cryptography/commits/b5d01d5888350a804ca47888e709d31637130e80/statuses). I'm fixing the detection of [merge commits here](https://github.com/codecov/codecov-python/commit/b26a4504492158358a04c053c1ad8af9c8fbfd8d). However, still need to handle this issue server side for posting the status on the correct commit.

PR #2882, opened 2016-04-29
CRYPTO_cleanup_all_ex_data is a macro and CRYPTO_add no longer exists

reaperhulk:
We don't use CRYPTO_add and neither does pyOpenSSL


reaperhulk:
Jenkins, retest this please

PR #2883, opened 2016-04-29
treat the DSA struct as opaque

reaperhulk:
The actual struct will be opaqued in a future PR once we make pyopenssl handle this properly.


reaperhulk:
I'm not sure why the CentOS 5 2.7 builder is crashing here. Will investigate.

alex:
CentOS builder remains broken here.

alex:
There's a lot going on in this PR, strongly recommend splitting it into multiple to simplify debugging

PR #2884, opened 2016-04-29
opaque the DH struct

reaperhulk:
We're not using it (yet) so I didn't add the getter/setter functions. We can add them if/when we implement classical DH.


lvh:
I guess we don't need to track that in a ticket since we'll do it anyway. Looks good to me once CI is happy.

lvh:
(Is the close/reopen to jiggle CI?)

reaperhulk:
Yep, codecov is having some challenges

PR #2896, opened 2016-05-04
add the scripts we use to build the wheels in jenkins to version control

reaperhulk:
This is a copy of exactly what the cryptography-wheel-builder job currently does. Once this is merged we can switch the job to just invoke these scripts.


alex:
@reaperhulk is this ready for review/merge?

.jenkins/mac-wheel.bat
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash -xe
```
alex: This should not have a `.bat` extension, should be `.sh`.

Also does this need `/bin/bash` or can `/bin/sh` work?

Issue #2901, opened 2016-05-05
Expose extension data for UnsupportedExtension


rudolphfroger:
Hi,
When a critical unrecognized extension is found in a certificate a UnsupportedExtension is raised. This exception holds the OID but not the extension data. I'd like to parse this unrecognized extension data myself but can't access it by catching the exception. Would it be acceptable to also encapsulate the extension data in the UnsupportedExtension exception?

Thanks, Rudolph

rudolphfroger:
BTW, I'll happily write a pull request if that will help

reaperhulk:
Attach it to the exception? That seems like a good approach. Our goal with raising the exception is to comply with the spirit of RFC 5280, which says:
```
A certificate-using system MUST reject the certificate if it encounters
a critical extension it does not recognize or a critical extension
that contains information that it cannot process.
```

Raising the exception but dropping the ASN.1 encoded data in it would allow users of this library to catch and process such extensions properly (or not catch them and by default the parser will fail appropriately).

So yes, please do submit a PR! If you haven't yet, take a quick glance at our [submitting patches](https://cryptography.io/en/latest/development/submitting-patches/) documentation to avoid some of the common pitfalls for new contributors.

rudolphfroger:
Thanks, hopefully my PR complies with your policies.

PR #2903, opened 2016-05-06
Expose extension data for UnsupportedExtension, issue #2901

rudolphfroger:


alex:
Jenkins, ok to test.

reaperhulk:
This will give you the value, but unfortunately since it raises an exception it will halt processing of all subsequent extensions (and won't give you previous ones either). This seems like it probably isn't what you actually want (e.g. in a certificate with 5 extensions, you'd only get the first unrecognized critical and would not have access to any of the others).

An alternate (but more involved) approach would be to come up with a public API for our extension parsing classes that allows users to register an OID and a callback. Something like

```python
def my_parser_func(data):
    // data is bytes, parse it however you need to
    a, b, c = parse_the_extension(data)
    return MyCustomExtensionClass(a, b, c)

x509.register_x509_extension(x509.ObjectIdentifier('1.2.3'), my_parser_func)

cert = x509.load_pem_x509_certificate(mypem, backend)
for ext in cert.extensions:
   print(ext)
```

With an approach like this you would see the instance of `MyCustomExtensionClass` be available in the `cert.extensions` iterable just like all the other extensions.

rudolphfroger:
Yes, that would be much better. I'll try to come with a patch.

alex:
Let's try to come up with an API that doesn't involve a global registry

rudolphfroger:
How about a method on Certificate instances to register extra extension handlers on the instance (to be used before iterating over 'extensions')? The same probably applies to CertificateRevocationList, CertificateSigningRequest and RevokedCertificate. Something like
```python
cert = x509.load_pem_x509_certificate(mypem, backend)
cert.register_extension_handler(x509.ObjectIdentifier('1.2.3'), my_parser_func)
for ext in cert.extensions:
   print(ext)
```

reaperhulk:
Having to register a handler every time you load a certificate seems like it would be annoying. In most cases users are presumably going to want repeated parsing to be capable of interpreting the same certificates. @alex do you have a suggestion? I was thinking maybe tying it to a backend, but that has complications with multibackend.

alex:
I _think_ we want a public API version of the OpenSSL backend's
ExtensionParser idea.

On Fri, May 6, 2016 at 7:47 PM, Paul Kehrer <notifications@github.com>
wrote:

> Having to register a handler every time you load a certificate seems like
> it would be annoying. In most cases users are presumably going to want
> repeated parsing to be capable of interpreting the same certificates.
> @alex <https://github.com/alex> do you have a suggestion? I was thinking
> maybe tying it to a backend, but that has complications with multibackend.
>
> —
> <https://github.com/pyca/cryptography/pull/2903#issuecomment-217588880>
>



-- 

rudolphfroger:
I'm not familiar with OpenSSL's ExtensionParser idea is but we could add an API which extends the `*_HANDLERS` dicts of `cryptography/hazmat/backends/openssl/decode_asn1.py`

```python
x509.register_extension_parser(decode_my_ext, backend)
```

alex:
We don't want the global state though, so you create a specific extension parser thing.

frasertweedale:
Big +1 for a way to process critical unsupported extensions (along with all
other extensions).

Consider the story where someone wants to use python-cryptography to write
a certificate inspector / pretty-printer.  Processing cannot barf on critical but
unknown extensions.

I'm undertaking a refactor of FreeIPA's certificate processing code and this
is a blocker to using python-cryptography.

rudolphfroger:
Alex mentioned an ExtensionParser class. I'm not sure what design he has in mind but I imagine it can be used to register extra extension parsers and which may be provided when parsing a certificate. This allows one to have multiple ExtensionParsers for different situations (i.e. no global extension parser registry).

Example:

```python
from cryptography.x509 import ExtensionParser

def my_parser_func(data):
    # data is bytes, parse it however you need to
    a, b, c = parse_the_extension(data)
    return MyCustomExtensionClass(a, b, c)

ext_parser = ExtensionParser()
ext_parser.register_extension_parser(x509.ObjectIdentifier('1.2.3'), my_parser_func)

cert = x509.load_pem_x509_certificate(mypem, extension_parser=ext_parser,
    backend=backend)

for ext in cert.extensions:
   print(ext)

def my_parser_func(data):
    # data is bytes, parse it however you need to
    x, y = parse_the_extension(data)
    return MyOtherCustomExtensionClass(x, y)

other_ext_parser = ExtensionParser()
other_ext_parser.register_extension_parser(x509.ObjectIdentifier('1.2.3'),
    my_other_parser_func)

other_cert = x509.load_pem_x509_certificate(my_other_pem, extension_parser=other_ext_parser,
    backend=backend)
```

simo5:
A parser is fine, but you still need a "generic" parser that is a catchall if you do not have a specialized one. In some cases you just need to inspect certificates even if you do not understand the critical extension.

rudolphfroger:
Adding a generic parser which just returns the ObjectIdentifier with the raw data would be trivial to add when an ExtensionParser interface would be available.

PR #2905, opened 2016-05-09
Replacing test_osrandom_engine_is_default.

grrrrrrrrr:
test_osrandom_engine_is_default depends on having a valid
sys.executable. This attribute is not always set (see
https://docs.python.org/2/library/sys.html#sys.executable ) so, in some
environments, this test fails. I moved the functionality of the test
into the setup and teardown methods so the correct behavior is still
tested.

### DIFF
--- a/tests/hazmat/backends/test_openssl.py
+++ b/tests/hazmat/backends/test_openssl.py
@@ -6,6 +6,9 @@ from __future__ import absolute_import, division, print_function
 
 import datetime
 import os
+import subprocess
+import sys
+import textwrap
 
 import pytest
 
@@ -226,6 +229,46 @@ class TestOpenSSLRandomEngine(object):
         current_default = backend._lib.ENGINE_get_default_RAND()
         name = backend._lib.ENGINE_get_name(current_default)
         assert name == osrandom_engine_name
+
+    @pytest.mark.skipif(sys.executable is None,
+                        reason="No Python interpreter available.")
+    def test_osrandom_engine_is_default(self, tmpdir):
+        engine_printer = textwrap.dedent(
+            """
+            import sys
+            from cryptography.hazmat.backends.openssl.backend import backend
+
+            e = backend._lib.ENGINE_get_default_RAND()
+            name = backend._lib.ENGINE_get_name(e)
+            sys.stdout.write(backend._ffi.string(name).decode('ascii'))
+            res = backend._lib.ENGINE_free(e)
+            assert res == 1
+            """
+        )
+        engine_name = tmpdir.join('engine_name')
+
+        # If we're running tests via ``python setup.py test`` in a clean
+        # environment then all of our dependencies are going to be installed
+        # into either the current directory or the .eggs directory. However the
+        # subprocess won't know to activate these dependencies, so we'll get it
+        # to do so by passing our entire sys.path into the subprocess via the
+        # PYTHONPATH environment variable.
+        env = os.environ.copy()
+        env["PYTHONPATH"] = os.pathsep.join(sys.path)
+
+        with engine_name.open('w') as out:
+            subprocess.check_call(
+                [sys.executable, "-c", engine_printer],
+                env=env,
+                stdout=out,
+                stderr=subprocess.PIPE,
+            )
+
+        osrandom_engine_name = backend._ffi.string(
+            backend._binding._osrandom_engine_name
+        )
+
+        assert engine_name.read().encode('ascii') == osrandom_engine_name
 
     def test_osrandom_sanity_check(self):
         # This test serves as a check against catastrophic failure.
--- a/tests/hazmat/backends/test_openssl.py
+++ b/tests/hazmat/backends/test_openssl.py
@@ -207,52 +207,29 @@ class TestOpenSSL(object):
         assert backend._ffi.buffer(buf)[0:length] == sample_data
 
 
+
+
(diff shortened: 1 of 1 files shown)


alex:
In what circumstances is `sys.executable` `None`?

alex:
Jenkins, ok to test.

glyph:
@alex - I am drawing on memory here, but I believe if a program embeds python and does not call `Py_SetProgramName` it won't have a `sys.executable`.  In a sense this is accurate; `sys.executable` presumes you have an executable which will start up Python, and in the context of an embedded interpreter you don't.

alex:
Uhhh, ok that makes sense. But in what circumstances are you embedding python _and_ running these tests. How does that even work?

glyph:
> in what circumstances are you embedding python and running these tests

you beat me to the punch, I was just typing that comment…

glyph:
@grrrrrrrrr - so you said, "in some environments" - in what environment did *you* see this happening?

grrrrrrrrr:
Hm we have an internal environment here where this is the case. It's pretty straightforward to work around this but I thought in the end, the test is still relying on a false assumption so I would rather fix it properly.

Do you think this is reason enough? It would also get rid of shelling out to another python process which would be a win imho.

alex:
Does bazel produce binaries combining python and the source files now? :D

I'm fine with this in principle, haven't reviewed the PR itself, though you probably want to take a look at the pep8 errors from CI.

glyph:
@grrrrrrrrr can you describe any properties of that environment besides "internal"?  would it be interesting, for example, to try to introduce a similar environment into cryptography's test matrix?

grrrrrrrrr:
Pep8 issues should be fixed.

Yeah, we are actually embedding Python into executables so there is no interpreter. I'm not sure if it would make sense to test for this, it's a pretty special environment.

reaperhulk:
@alex, @glyph would it be possible to take a look at this again?

glyph:
@reaperhulk My thought here is that this test is valuable and should not be removed.  The PR says "replace" but I see a diff that just deletes it.

If anything, this should be adding _more_ test coverage, since we're looking for an integration environment with no `sys.executable`.  I'd suggest adding a skipIf decorator to the test to only skip running it if `sys.executable` is `None`.

Clearly Cryptography cares about embedding integration environments though; apache being the big one.  So the work to set up additional CI for a stub embedded interpreter would not be wasted just for this.

reaperhulk:
I had not looked at the diff (shame on me). I agree that this should be a skip check and not a removal of the test though.

grrrrrrrrr:
This PR does not remove any functionality in the test but I
agree that it's hard to see.

What happens is that the test wants to verify that the default
random engine is osrandom. However, some of the tests here change
global state and change that default (it would be better if they
didn't imho!). The teardown method always restores the default
engine to be osrandom to revert those changes but this now means
that if the test does not run as the very first one of this
class, the backend might have been overwritten already. This is
the reason why it forks out to a new interpreter - to get a clean
state with the real default random engine.

The way I changed this now is that it stores the current default
engine in the setup method. This method runs before every test
and now guarantees that the default engine is correct before
every test.

This is actually stronger than the old test. It can now detect
the cases where

- something goes wrong in the restoration of the default BE. 

- some other, unrelated test set the engine to something
else. Then, tests were run using the wrong engine. The first test
here forks out and passes since it gets a good default. The
teardown method resets the engine to osrandom and using the wrong
engine remains undetected.


I tried to explain this in the comments but it's complicated. If
we want to make it more explicit, I could put the test back and
just repeat the assert from the teardown method. That would be a
bit easier to see, what do you think?

glyph:
The previous test ensures that a fresh Python process, loading nothing but Cryptography, initializes the osrandom engine.  This test ensures that, at some random point in the middle of the test suite, the osrandom engine is initialized.

This is not to mention the fact that the removed test is a specifically identified _test_, which can ostensibly fail, whereas the PR changes it to some setup code that can just encounter an error.  Repeating the same assert from the teardown would make this slightly less bad but wouldn't address the previous point.

Given what the test wants to test, I don't think there's any way of getting away from needing `sys.executable`.  There's no way to write the test this way where it wouldn't be confounded by setup code elsewhere in the test suite that exists only to initialize the osrandom engine.

tests/hazmat/backends/test_openssl.py
```diff
@@ -208,14 +208,30 @@ def notrandom(size):
...
+        e = backend._lib.ENGINE_get_default_RAND()
+        self.default_engine_name = backend._lib.ENGINE_get_name(e)
+
     def teardown_method(self, method):
         # we need to reset state to being default. backend is a shared global
         # for all these tests.
+
+        # First check that osrandom was actually the default before the test.
+        osrandom_engine_name = backend._binding._osrandom_engine_name
+        assert self.default_engine_name == osrandom_engine_name
```
alex: Why is this assert here, instead of in `setup`?

grrrrrrrrr: You are right, this does not make sense anymore. Moved to setup().

alex:
@reaperhulk pending coverage this LGTM, ack?

alex:
@stevepeak http://status.codecov.io/ doesn't show any issues, but still waiting on the status here; is there an issue ATM?

alex:
Doing a trick to try to resolve teh codecov thing.

PR #2919, opened 2016-05-25
Implement math ops on EllipticCurvePublicKey

npmccallum:
This permits cryptography to be used for protocols like JPAKE, SPAKE,
Hughes Diffie Hellman and Tang.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/ec.py
+++ b/src/cryptography/hazmat/backends/openssl/ec.py
@@ -40,6 +40,26 @@ def _truncate_digest_for_ecdsa(ec_key_cdata, digest, backend):
     return _truncate_digest(digest, order_bits)
 
 
+def _point_to_public_key(backend, group, point):
+    "Converts an EC_POINT to an EllipticCurvePublicKey."
+
+    backend.openssl_assert(group != backend._ffi.NULL)
+    backend.openssl_assert(point != backend._ffi.NULL)
+
+    ec_key = backend._lib.EC_KEY_new()
+    backend.openssl_assert(ec_key != backend._ffi.NULL)
+    ec_key = backend._ffi.gc(ec_key, backend._lib.EC_KEY_free)
+
+    res = backend._lib.EC_KEY_set_group(ec_key, group)
+    backend.openssl_assert(res == 1)
+
+    res = backend._lib.EC_KEY_set_public_key(ec_key, point)
+    backend.openssl_assert(res == 1)
+
+    evp_pkey = backend._ec_cdata_to_evp_pkey(ec_key)
+    return _EllipticCurvePublicKey(backend, ec_key, evp_pkey)
+
+
 def _ec_key_curve_sn(backend, ec_key):
     group = backend._lib.EC_KEY_get0_group(ec_key)
     backend.openssl_assert(group != backend._ffi.NULL)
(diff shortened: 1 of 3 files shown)


npmccallum:
Let me give some justification for this. There are a fairly large set of smaller protocols beyond the typical encryption/signing functions with elliptic curves. I gave some examples in the commit message. These algorithms are very often small variants of ECDH. That means that all the core operations are already implemented in python-cryptography. All that is needed is a small tweak to existing algorithms. These operations make this possible.

Further, now that python-cryptography is getting traction, there are libraries starting to be built on top of it. One such example is jwcrypto. An application that, for instance, needs to do SPAKE with JOSE will use jwcrypto to do the marshalling and get an EllipticCurvePublicKey. To complete the SPAKE algorithm all that the developer needs to do is one add and one multiply. But without my patch this is not possible.

Instead, the developer would have to marshall the EllipticCurvePublicKey again to SEC1 and then write his own interop with another library (like openssl). This is a lot of effort for two small operations when we already have all the context in python-cryptography, we just need to expose the operations. Most developers will just avoid the pyca stack rather than writing the interop.

One possible alternative implementation of this could be to create a secondary interface, such as EllipticCurvePublicKeyWithOperations. This would permit libraries that don't expose these operations to implement the base interface only. I'm open to this as well.

This code should be side-channel free. OpenSSL's API encapsulates the risky code of private key math; we simply call the API. The patch I have written is free of branches based on the private key. There is, however, a branch in the equality operation used for comparing public keys. We may be able to obviate this branch if it is a concern.

There are two things this code doesn't contain that I considered implementing (and may still). First, is exposure to the generator constants. The second is integer multipliers. These are particularly helpful when checking public keys against illicit values (for instance g^1).

src/cryptography/hazmat/backends/openssl/ec.py
```diff
@@ -302,3 +322,120 @@ def public_bytes(self, encoding, format):
...
+
+            res = self._backend._lib.EC_POINT_add(sg, sum, sp, op, bn_ctx)
+            self._backend.openssl_assert(res == 1)
+
+        return _point_to_public_key(self._backend, sg, sum)
+
+    def __sub__(self, public_key):
+        return self + -public_key
+
+    def __multiply(self, bignum):
```
sigmavirus24: More of a question for @reaperhulk, is using a double `_` to prefix the right stylistic thing?

reaperhulk: We typically use just a single underscore to denote a private method, but obviously `__` will make it genuinely inaccessible outside the class. I don't believe we do this anywhere else in cryptography but I'm okay with it. Others may disagree.

alex: I'd prefer to stick to our current style of single underscore.

On Wed, May 25, 2016 at 7:13 PM, Paul Kehrer <notifications@github.com>
wrote:

> <https://github.com/pyca/cryptography/pull/2919#discussion_r64669066>:
>
> > +        self._backend.openssl_assert(sum != self._backend._ffi.NULL)
> > +        sum = self._backend._ffi.gc(sum, self._backend._lib.EC_POINT_free)
> > +
> > +        with self._backend._tmp_bn_ctx() as bn_ctx:
> > +            res = self._backend._lib.EC_GROUP_cmp(sg, og, bn_ctx)
> > +            self._backend.openssl_assert(res == 0)
> > +
> > +            res = self._backend._lib.EC_POINT_add(sg, sum, sp, op, bn_ctx)
> > +            self._backend.openssl_assert(res == 1)
> > +
> > +        return _point_to_public_key(self._backend, sg, sum)
> > +
> > +    def __sub__(self, public_key):
> > +        return self + -public_key
> > +
> > +    def __multiply(self, bignum):
>
> We typically use just a single underscore to denote a private method, but
> obviously __ will make it genuinely inaccessible outside the class. I
> don't believe we do this anywhere else in cryptography but I'm okay with
> it. Others may disagree.
>
> —
> <https://github.com/pyca/cryptography/pull/2919/files/3120707ebd9de09329cb8ea28de9e02eb36fd98c#r64669066>
>



-- 

alex:
Jenkins, ok to test.

alex:
Tests are failing on CentOS here.

npmccallum:
Here is the latest version. All checks pass.

1. I migrated __multiply() => _multiply().
2. I made the addition and subtraction test more generic. Now, all supported curves are tested.
3. I removed the reference to EC_GROUP_cmp() which doesn't exist on old OpenSSLs.
4. I removed the branch in the equality operator.

npmccallum:
Any update?

reaperhulk:
@npmccallum I apologize for letting this float for many, many months. I've thought about this off and on for months and I believe we do need to support elliptic curve operations like this if we want to be the base for higher level cryptographic libraries in Python.

I am a bit wary of exposing it with the Python infix operators, but I can't really articulate why it bothers me. @lvh I know we've talked in the past about the general benefits of exposing elliptic curve math operations -- would you mind weighing one last time about the advantages (and potential disadvantages) of supporting this directly in cryptography? I'd like to make a final decision on whether we're going to accept this feature this week.

lvh:
1. I don't think that the output of operations that internally happen to be a group operation (like e.g. a DH exchange step) should be a group element object that understands addition and multiplication.
2. No opinions on general desirability of infix operators on group elements, but if you do, you should support `n * P` but not `P * n`. You should support point addition, and maybe subtraction (with the caveat that that isn't necessarily supported in every case).

I think you can do the "neat" API in a "as group element" wrapper. This would be useful, because if you want to implement e.g. SPAKE2, you don't care if the group is typically written multiplicatively or additively, and I don't care if it's an elliptic curve today and some other construct tomorrow.

tuxxy:
I'd like to push this PR up a bit.

It seems like this was being considered some time back. From the mailing list, @alex informed me that he was hesitant to merge this PR.

Cryptography is quite powerful. It is also one of the best libraries in Python for working with cryptographic primitives.

I would very much so love to make the case for this sort of arithmetic being implemented in this library. I don't necessarily need (or want, in particular) these operations to be performed from infix operators, but I do want them exposed in the library.

If Cryptography adds this in, it will become one of the best libraries for quickly prototyping and building cryptosystems with elliptic curves. Our company (@NuCypher) is in need of these operations and we would prefer to use a safer library such as Cryptography.io to use this.

So, I ask that the wonderful folks who maintain this please consider adding these in -- whether it's in the form of this PR or a different one.

reaperhulk:
As you can see from my comments over a year ago (ugh, that long?!) I am in favor of EC math being possible in the lib. However, since this is even more dangerous than usual I'd like to make code auditing for calls of this type very easy. For the purposes of discussion, what about a new module called `cryptography.hazmat.primitives.asymmetric.ec.dangerous_math` that contains functions for this? And does that make sense in a world where we'll be adding a `cryptography.hazmat.primitives.asymmetric.ed25519` once OpenSSL 1.1.1 comes out?

tuxxy:
We (@nucypher) need to implement the arithmetic for our proxy re-encryption library. I went ahead and forked and added in some BIGNUM arithmetic on the `EllipticCurvePrivateKey` object going along with using the Python infixes. [See this PR on our fork for the details of this.](https://github.com/nucypher/cryptography/pull/1)

To give you folks an idea of whats possible with arithmetic, I think you all will find interactive (non-interactive will be up shortly) proxy re-encryption interesting enough of an application. [See this code here for our Umbral spec.](https://github.com/tuxxy/nucypher-pre-python/tree/umbral-cryptography) For a quick summary, Umbral is what we call Split-key Threshold Proxy Re-encryption. It allows us to do, basically, Shamir's Secret Sharing on re-encryption keys for proxy re-encryption.

This is implemented completely with the additions in our above fork. 

As a frequent user of this library, it would be so amazing to be able to prototype and work on things with it rather than the alternatives.

Also, if we can get this arithmetic added in somehow, you will get the full support of me and @nucypher to help maintain it.

We're really pushing hard to have this added in so we can make some awesome stuff!

npmccallum:
All of the operations are logically valid for ed25519. But I do not know if OpenSSL supports all the operations.

tuxxy:
To keep anyone watching this updated:

We're not using this fork anymore to perform arithmetic. I've written a small module called [hazmat-math](https://github.com/tuxxy/hazmat-math/blob/master/hazmat_math/operations.py) that implements the major BN and EC_POINT operations on `_EllipticCurvePrivateKey` and `_EllipticCurvePublicKey` objects.

@reaperhulk Just curious, is this the sorta style of `dangerous_math` that you would expect to be in this theoretical addition?

reaperhulk:
Yes, that looks roughly like what I'd expect to see in that module.

earonesty:
FYI: we use the hazmat_math module all the time (threshold crypto, commitments, etc).  very useful.   maybe pyca can simply point to the hazmat_math repo officially in the docs

tuxxy:
I would recommend not doing this until I fix some of the issues at least. If someone at pyca is interested, let me know and I'll think about getting it merged into cryptography.io or fixing it up for public use.

reaperhulk:
3 years on clearly we're not really moving on this. Some other projects have sprung up using `cryptography` to do this sort of ultra-hazmat math, but for now we're probably not pulling it into `cryptography` proper.

PR #2930, opened 2016-05-30
these constants no longer exist in 1.1.0 and we weren't using them

reaperhulk:
Part of the 1.1.0 compatibility effort. These constants (and one macro) were removed post-pre5.


reaperhulk:
Sigh, another instance of status not being reported :(

alex:
Not sure why codecov is barfing, but this is fine.

PR #2932, opened 2016-05-31
Add extended Fernet classes to allow using 192- and 256-bit AES encryption

tknarr:
Refactored the Fernet code into FernetBase which implements the format and takes the AES key lengths from it's inputs or method arguments, and 3 derived classes (Fernet, ExtFernet192, ExtFernet256) which specify standard 128-bit or extended 192- and 256-bit keys respectively.

Fernet produces the same output as the original Fernet class. ExtFernet192 and ExtFernet256 change the version byte to 0x21 and 0x41 respectively to distinguish the data from standard Fernet and to indicate the key length needed.

The "invalid" set of tests for ExtFernet192/ExtFernet256 were omitted because they test code in the FernetBase class that doesn't differ based on the key length, the tests for Fernet will test that code.


tknarr:
Will open a new one after Travis checks succeed.

PR #2935, opened 2016-05-31
treat rsa struct as opaque

reaperhulk:


alex:
👍  will merge on passing tests

Issue #2943, opened 2016-06-02
serial/serial_number comedy


alex:
Certificate has serial, RevokedCertificate has serial_number. We should add an alias to un-comedy this

PR #2950, opened 2016-06-02
Add alias for Certificate serial as serial number

chellygel:
fixes: #2943

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -1573,7 +1573,7 @@ X.509 Extensions
             >>> from cryptography.hazmat.backends import default_backend
             >>> issuer_cert = x509.load_pem_x509_certificate(pem_data, default_backend())
             >>> x509.AuthorityKeyIdentifier.from_issuer_public_key(issuer_cert.public_key())
-            <AuthorityKeyIdentifier(key_identifier='X\x01\x84$\x1b\xbc+R\x94J=\xa5\x10r\x14Q\xf5\xaf:\xc9', authority_cert_issuer=None, authority__numbercert_serial_number=None)>
+            <AuthorityKeyIdentifier(key_identifier='X\x01\x84$\x1b\xbc+R\x94J=\xa5\x10r\x14Q\xf5\xaf:\xc9', authority_cert_issuer=None, authority_cert_serial_number=None)>
 
     .. classmethod:: from_issuer_subject_key_identifier(ski)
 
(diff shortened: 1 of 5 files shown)


docs/x509/reference.rst
```diff
@@ -1573,7 +1573,7 @@ X.509 Extensions
             >>> from cryptography.hazmat.backends import default_backend
             >>> issuer_cert = x509.load_pem_x509_certificate(pem_data, default_backend())
             >>> x509.AuthorityKeyIdentifier.from_issuer_public_key(issuer_cert.public_key())
-            <AuthorityKeyIdentifier(key_identifier='X\x01\x84$\x1b\xbc+R\x94J=\xa5\x10r\x14Q\xf5\xaf:\xc9', authority_cert_issuer=None, authority_cert_serial_number=None)>
+            <AuthorityKeyIdentifier(key_identifier='X\x01\x84$\x1b\xbc+R\x94J=\xa5\x10r\x14Q\xf5\xaf:\xc9', authority_cert_issuer=None, authority__numbercert_serial_number=None)>
```
alex: Not sure what happened here?

alex:
Jenkins, add to whitelist

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -58,6 +59,14 @@ def version(self):
 
     @property
     def serial(self):
+        warnings.warn(
+            "Certificate serial is deprecated, use serial_number instead.",
+            DeprecationWarning
```
alex: This should be a constant in the `cryptography.utils` module you'll see a few other examples there.

src/cryptography/utils.py
```diff
@@ -12,9 +12,11 @@
 import warnings
 
 
-# the functions deprecated in 1.0 are on an arbitrarily extended deprecation
-# cycle and should not be removed until we agree on when that cycle ends.
+# the functions deprecated in 1.0 and 1.04 are on an arbitrarily extended
```
alex: `1.04` is not correct here

tests/test_x509.py
```diff
@@ -486,19 +487,41 @@ def test_load_pem_cert(self, backend):
...
+        assert cert.serial_number == 2
+
+    def test_cert_serial_warning(self, backend):
+        cert = _load_cert(
+            os.path.join("x509", "PKITS_data", "certs", "GoodCACert.crt"),
+            x509.load_der_x509_certificate,
+            backend
+        )
+
+        warnings.simplefilter("always", utils.DeprecatedIn10)
```
alex: mmm, so I think you want to wrap this all in a `with warnings.catch_warning` to make sure the filter is restored: https://docs.python.org/2.7/library/warnings.html#warnings.catch_warnings

PR #2957, opened 2016-06-03
SSH serialization for public keys

### LINKED ISSUE #1744: OpenSSH public key serialization
We should support serialization of public keys to OpenSSH format.

alex:
fixes #1744

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -1745,7 +1745,23 @@ class Backend(object):
                     serialization._ssh_write_mpint(parameter_numbers.g) +
                     serialization._ssh_write_mpint(public_numbers.y)
                 )
-
+            else:
+                assert isinstance(key, ec.EllipticCurvePublicKey)
+                public_numbers = key.public_numbers()
+                curve_name = {
+                    ec.SECP256R1: b"nistp256",
+                    ec.SECP384R1: b"nistp384",
+                    ec.SECP521R1: b"nistp521",
+                }[type(public_numbers.curve)]
+                return b"ecdsa-sha2-" + curve_name + " " + base64.b64encode(
+                    serialization._ssh_write_string(
+                        b"ecdsa-sha2-" + curve_name
+                    ) +
+                    serialization._ssh_write_string(curve_name) +
+                    serialization._ssh_write_string(
+                        public_numbers.encode_point()
+                    )
+                )
         else:
             raise TypeError(
                 "format must be an item from the PublicFormat enum"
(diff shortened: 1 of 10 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1722,6 +1723,45 @@ def _public_key_bytes(self, encoding, format, evp_pkey, cdata):
...
+                    serialization._ssh_write_mpint(public_numbers.y)
+                )
+            else:
+                assert isinstance(key, ec.EllipticCurvePublicKey)
+                public_numbers = key.public_numbers()
+                curve_name = {
+                    ec.SECP256R1: b"nistp256",
+                    ec.SECP384R1: b"nistp384",
+                    ec.SECP521R1: b"nistp521",
+                }[type(public_numbers.curve)]
```
reaperhulk: Any curve outside of the 3 supported will raise a KeyError right now. This is probably not what we want.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1722,6 +1723,51 @@ def _public_key_bytes(self, encoding, format, evp_pkey, cdata):
                 write_bio = self._lib.i2d_RSAPublicKey_bio
 
             key = cdata
+        elif format is serialization.PublicFormat.OpenSSH:
+            if encoding is not serialization.Encoding.OpenSSH:
+                raise ValueError(
+                    "OpenSSH format must be used with OpenSSH encoding"
+                )
+            if isinstance(key, rsa.RSAPublicKey):
```
reaperhulk: The OpenSSH branch makes this function really long and a bit hard to read. What do you think about splitting this into a separate method? It will obviously only be called once but would be a method that takes a key and returns a bytestring.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1722,6 +1723,51 @@ def _public_key_bytes(self, encoding, format, evp_pkey, cdata):
...
+                        ec.SECP256R1: b"nistp256",
+                        ec.SECP384R1: b"nistp384",
+                        ec.SECP521R1: b"nistp521",
+                    }[type(public_numbers.curve)]
+                except KeyError:
+                    raise ValueError(
+                        "Only SECP256R1, SECP384R1, and SECP521R1 curves are "
+                        "supported by the SSH public key format"
+                    )
+                return b"ecdsa-sha2-" + curve_name + " " + base64.b64encode(
```
reaperhulk: " " needs to be bytes here too.

Issue #2982, opened 2016-06-18
Post-0.9.8 cleanup


reaperhulk:
We should clean up/refactor a few things once #2978 lands.

- [x] `d2i_AutoPrivateKey` might be usable in `load_der_private_key` now.
- [x] Investigate using `DSAparams_dup` to simplify some of our DSA code. We previously couldn't use it because it caused a `SIGILL` on CentOS 5 in 0.9.8e.
- [x] The RSA sign/verify code paths no longer have as many branches so we can likely simplify the logic.
- [x] Remove gunk from `.travis/`
- [x] `src/cryptography/hazmat/backends/openssl/ciphers.py`
- [x] `src/cryptography/hazmat/backends/openssl/backend.py`
- [x] `src/cryptography/hazmat/backends/openssl/dsa.py`
- [x] `src/cryptography/hazmat/backends/openssl/decode_asn1.py`
- [x] `0.9.8` sightings
    - [x] `src/_cffi_src/openssl/aes.py`
    - [x] `src/_cffi_src/openssl/bio.py`
    - [x] `src/_cffi_src/openssl/dsa.py`
    - [x] `src/_cffi_src/openssl/err.py`
    - [x] `src/_cffi_src/openssl/evp.py`
    - [x] `src/_cffi_src/openssl/nid.py`
    - [x] `src/_cffi_src/openssl/x509.py`
    - [x] `src/_cffi_src/openssl/x509_vfy.py`
    - [x] `src/_cffi_src/openssl/ssl.py`
- [x] `OPENSSL_VERSION_NUMBER` sightings
    - [x] `src/_cffi_src/openssl/cms.py`
    - [x] `src/_cffi_src/openssl/hmac.py`
    - [x] `src/_cffi_src/openssl/evp.py`
    - [x] `src/_cffi_src/openssl/rsa.py`
    - [x] `src/_cffi_src/openssl/ssl.py`
    - [x] `src/_cffi_src/openssl/x509.py`
    - [x] `src/_cffi_src/openssl/x509_vfy.py`
    - [x] `src/_cffi_src/openssl/x509v3.py`

PR #2987, opened 2016-06-18
0.9.8 error code conditionals no longer required

reaperhulk:
refs #2982 

tlsext error codes were added in 0.9.8f so we don't have to check for their presence any more.

Unlike the other PRs I removed the `Cryptography_HAS_*` constants in this one as I can't fathom a situation where an external consumer might have been checking for that...

### DIFF
--- a/src/_cffi_src/openssl/err.py
+++ b/src/_cffi_src/openssl/err.py
@@ -52,6 +52,10 @@ static const int ASN1_F_LONG_C2I;
 static const int ASN1_F_OID_MODULE_INIT;
 static const int ASN1_F_PARSE_TAGGING;
 static const int ASN1_F_PKCS5_PBE_SET;
+static const int ASN1_F_B64_READ_ASN1;
+static const int ASN1_F_B64_WRITE_ASN1;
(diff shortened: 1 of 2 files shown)


alex:
👍  lgtm

src/_cffi_src/openssl/err.py
```diff
@@ -193,6 +190,16 @@
...
+
+static const int ASN1_F_B64_READ_ASN1;
+static const int ASN1_F_B64_WRITE_ASN1;
+static const int ASN1_F_SMIME_READ_ASN1;
+static const int ASN1_F_SMIME_TEXT;
+static const int ASN1_R_NO_CONTENT_TYPE;
+static const int ASN1_R_NO_MULTIPART_BODY_FAILURE;
+static const int ASN1_R_NO_MULTIPART_BOUNDARY;
+static const int EVP_F_CAMELLIA_INIT_KEY;
+static const int EVP_R_CAMELLIA_KEY_SETUP_FAILED;
```
alex: Actually, I think there's existing groupings of consts that these could be added to.

PR #2991, opened 2016-06-18
Use `d` instead of `private_exponent` for consistency

djc:
True story: I used `e` instead of `d` because it seemed more closely related
to `private_exponent`. Should have looked it up, of course... but the docs could be better.


alex:
Jenkins, ok to test.

alex:
Mmm, this change would be backwards incompatible for anyone using keyword arguments.

djc:
How about this version, instead?

alex:

PR #2996, opened 2016-06-19
Simplify and remove some 0.9.8 code from the EVP bindings

alex:
The Cryptography_ symbol remains because pyOpenSSL uses. There's still other 0.9.8isms in this file.


alex:
Refs #2982

PR #2997, opened 2016-06-19
Remove 0.9.8isms from CMS code.

alex:
CMS is 'cryptographic message syntax', and not 'centers for medicaid and medicare', fyi


alex:
Refs #2982

PR #3004, opened 2016-06-20
Rewrite a comment.

alex:
d2i_AutoPrivateKey is less useful than we thought


alex:
Refs #2982

reaperhulk:
Well that's annoying >:(

alex:
Unless I missed something, yes sir.

PR #3010, opened 2016-06-21
Enforce that p > q to improve OpenSSL compatibility (fixes #2990)

### LINKED ISSUE #2990: rsa.rsa_recover_prime_factors() should return p > q
The documentation for `rsa_recover_prime_factors()` warns that it returns `p` and `q` such that `p < q`. However, things like OpenSSL and BoringSSL seem to require that `p > q`. Given this, would it be feasible to change the order around in cryptography so that it lines up with OpenSSL?

See also: http://crypto.stackexchange.com/questions/18084/in-rsa-why-does-p-have-to-be-bigger-than-q-where-n-p-times-q. @briansmith can provide more commentary if needed.

djc:


reaperhulk:
Looks like there are some test failures here

reaperhulk:
jenkins, ok to test

tests/hazmat/primitives/test_rsa.py
```diff
@@ -1981,10 +1981,10 @@ def test_recover_prime_factors(self, vector):
             private["private_exponent"]
         )
         # Unfortunately there is no convention on which prime should be p
-        # and which one q. The function we use always makes p < q, but the
+        # and which one q. The function we use always makes p > q, but the
         # NIST vectors are not so consistent. Accordingly, we verify we've
         # recovered the proper (p, q) by sorting them and asserting on that.
-        assert sorted([p, q]) == sorted([private["p"], private["q"]])
+        assert sorted([q, p]) == sorted([private["p"], private["q"]])
```
reaperhulk: This line should probably be `assert [p, q] == sorted([private["p"], private["q"]], reverse=True)` now. Reordering `[q, p]` inside `sorted` won't actually work since it'll sort by smallest first.

djc:
@reaperhulk Travis seems to like it, what do you think?

tests/hazmat/primitives/test_rsa.py
```diff
@@ -1985,10 +1985,11 @@ def test_recover_prime_factors(self, vector):
             private["private_exponent"]
         )
         # Unfortunately there is no convention on which prime should be p
-        # and which one q. The function we use always makes p < q, but the
+        # and which one q. The function we use always makes p > q, but the
         # NIST vectors are not so consistent. Accordingly, we verify we've
         # recovered the proper (p, q) by sorting them and asserting on that.
         assert sorted([p, q]) == sorted([private["p"], private["q"]])
+        assert p > q, (p, q)
```
reaperhulk: I don't understand what the `, (p, q)` is doing here?

src/cryptography/hazmat/primitives/asymmetric/rsa.py
```diff
@@ -246,7 +246,7 @@ def rsa_recover_prime_factors(n, e, d):
             if cand != 1 and cand != (n - 1) and pow(cand, 2, n) == 1:
                 # We have found a number such that (cand-1)(cand+1)=0 (mod n).
                 # Either of the terms divides n.
-                p = gcd(cand + 1, n)
+                q = gcd(cand + 1, n)
```
reaperhulk: Since we're sorting this now let's just leave these variable names untouched.

docs/hazmat/primitives/asymmetric/rsa.rst
```diff
@@ -508,8 +508,10 @@ this without having to do the math themselves.
 
     .. note::
 
-        When recovering prime factors this algorithm will always return ``p``
-        and ``q`` such that ``p < q``.
+        When recovering prime factors this algorithm will always returned ``p``
```
reaperhulk: `will always return` instead of `returned`.

docs/hazmat/primitives/asymmetric/rsa.rst
```diff
@@ -508,8 +508,10 @@ this without having to do the math themselves.
...
+        and ``q`` such that ``p > q``. Note: before 1.5, this function always
+        returned ``p`` and ``q`` such that ``p < q``. It was changed because
+        OpenSSL and related libraries generally require ``p > q``.
```
reaperhulk: OpenSSL doesn't require this though. How about `It was changed because libraries commonly require`?

tests/hazmat/primitives/test_rsa.py
djc: It's set as the message for the `AssertionError` if such an error is raised, so that you can see the values that led to the assertion trigger. I guess pytest may already rewrite assertion statements so that it does something friendly, however this is very useful outside pytest contexts.

Do you want me to remove it?

djc:
Addressed comments.

tests/hazmat/primitives/test_rsa.py
reaperhulk: Yeah, pytest rewrites asserts to give the values so let's remove that here.

djc:
Removed the assert ornament thingy.

reaperhulk:
@djc thanks for doing this! If you haven't added your name to `AUTHORS.rst` feel free to do so :)

djc:
Thank you for your patient reviews! Made another PR for adding myself in #3048.

Issue #2915, opened 2016-05-20
Let an application adjust minimum key length check in HOTP/TOTP


dholl:
I hope this is the correct place to report this issue, if not please redirect me.  :)

While trying to use TOTP in hazmat/primitives/twofactor/totp.py with secrets generated via google-authenticator, I encountered an incompatibility in key length check.  I found that TOTP objects each contain an HOTP object for the *OTP work, and within hotp.py, there is a check in HOTP's constructor that verifies the supplied key length:

       if len(key) < 16:
            raise ValueError("Key length has to be at least 128 bits.")

The problem here is that the key is supplied to this class _after_ decoding from base32, but the secrets generated by google-authenticator are 16 characters while in base32 format.  After base32-decoding one of these google-authenticator secrets, the resulting key is only 10 bytes (80 bits) long, and thus fails hotp.py's check.

From looking up the algorithm requirements in RFC 4226, I conclude that google-authenticator is generator insufficiently long keys.  Specifically, this text in section 4:

https://tools.ietf.org/html/rfc4226#section-4
   R6 - The algorithm MUST use a strong shared secret.  The length of the shared secret MUST be at least 128 bits.  This document RECOMMENDs a shared secret length of 160 bits.

So, the cryptography project is including a correct check of at least 128 bits, but the base-32 encoded keys from google-authenticator contain only 80 bits.

This issue has also been noted here:
https://security.stackexchange.com/questions/45053/why-does-google-cripple-the-2fa-google-authenticator-pam-module

Would the cryptography project consider adding an additional argument to the TOTP and HOTP constructors to permit a local code implementation to adjust this minimum length check?  This would be handy for application where compatibility with an existing google-authenticator deployment is considered "important enough" to downgrade the minimum key sizes...

Thank you,
David

reaperhulk:
Thanks for the detailed report, this is the right place.

Google Authenticator is, unfortunately, a very important source so I'm supportive of finding some way to do this. We could do:

```python
class HOTP(object):
    MINIMUM_KEY_LENGTH = 16

    def __init__(self, key, length, algorithm, backend):
        if not isinstance(backend, HMACBackend):
            raise UnsupportedAlgorithm(
                "Backend object does not implement HMACBackend.",
                _Reasons.BACKEND_MISSING_INTERFACE
            )

        if len(key) < self.MINIMUM_KEY_LENGTH:
            raise ValueError(
                "Key length has to be at least {} bytes.".format(
[...]

PR #3012, opened 2016-06-25
Add flag to toggle key length check for HOTP and TOTP.

Ayrx:
Potential fix for #2915. 

Out of all the API changes suggested, I feel like this one is the cleanest. It does go against project convention but it's the only API change that is backwards compatible. I'm not a fan of all the other suggestions that adds way too much complexity for my tastes.


Ayrx:
jenkins retest this please

tonyseek:
It looks well. 👍

reaperhulk:
Hmm, we discussed perhaps just making the key length check a function that could be overridden if needed in #2915. I'm not sure I'm a fan of that approach (as it'd be a private method and while it could be monkeypatched there's no guarantee that we wouldn't change the implementation in the future), but it seems worth making sure everybody agrees we don't want to do it that way.

Ayrx:
@reaperhulk Yeah, I saw that approach. I'm definitely not a fan of anything that encourages monkeypatching since imo that's just asking for trouble in the future. 

Regardless of which way we decide to go with eventually, it's good to have a concrete PR to discuss against.

Ayrx:
@alex, @reaperhulk: Just a poke on this. Any feedback so far? :)

reaperhulk:
I think I've made my peace with this approach. Needs a versionadded on the param though.

reaperhulk:
This is blocked on #3040

Ayrx:
jenkins retest this please

Ayrx:
All clear on the CI front now.

PR #3014, opened 2016-06-26
Fixed #3008 -- expose calculate max pss salt length	

### LINKED ISSUE #3008: Expose actual salt_length if padding.PSS.MAX_LENGTH is used
https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/  suggests to use salt_length=padding.PSS.MAX_LENGTH for signing with PSS padding.
To verify a signature, the salt_length must be known and cannot be determined from the signature.

Assuming in the verifier that the signer used MAX_LENGTH is not a proper solution as the signer is free to use different salt lengths. Some implementations may not even support the MAX_LENGTH feature, e.g. I couldn't find anything in BouncyCastle so far.

RFC3447 which specified the PSS algorithm also specifies an ASN.1 structure "RSASSA-PSS-params" for use in the "AlgorithmIdentifier" structure. It contains the parameters used for creating the signature, including the salt length.

I think we need a way to expose the result of _get_rsa_pss_salt_length to the user when creating a signature.
Alternatively, remove padding.PSS.MAX_LENGTH and add a function which computes this value so the user can pass the result back with the salt_length parameter.

alex:
Started working on the code, but then realized that I wasn't sure if it should go in `rsa` or `padding`. Opinion @reaperhulk ?

### DIFF
--- a/docs/hazmat/primitives/asymmetric/rsa.rst
+++ b/docs/hazmat/primitives/asymmetric/rsa.rst
@@ -330,6 +330,18 @@ Padding
     :class:`OAEP` should be preferred for encryption and :class:`PSS` should be
     preferred for signatures.
 
+
+.. function:: calculate_pss_max_salt_length(key, hash_algorithm)
+
+    :param key: An RSA public or private key.
+    :param hash_algorithm: A
+        :class:`cryptography.hazmat.primitives.hashes.HashAlgorithm`.
+    :returns int: The computed salt length.
+
+    Computes the length of the salt that :class:`PSS` will use if
+    :data:`PSS.MAX_LENGTH` is used.
+
+
 Mask generation functions
 -------------------------
 
--- a/docs/hazmat/primitives/asymmetric/rsa.rst
+++ b/docs/hazmat/primitives/asymmetric/rsa.rst
@@ -341,11 +341,10 @@ Mask generation functions
         Removed the deprecated ``salt_length`` parameter.
 
     MGF1 (Mask Generation Function 1) is used as the mask generation function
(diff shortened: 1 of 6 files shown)


alex:
(Still needs docs of course)

reaperhulk:
Since PSS is in padding I think this function probably belongs there too. Also, would it make sense to just have it take the integer key size rather than a key?

alex:
I kind of like it taking key, makes it easy for me to add a branch to error
if you pass some other key type.

On Sun, Jun 26, 2016 at 1:30 PM, Paul Kehrer <notifications@github.com>
wrote:

> Since PSS is in padding I think this function probably belongs there too.
> Also, would it make sense to just have it take the integer key size rather
> than a key?
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/pyca/cryptography/pull/3014#issuecomment-228612345>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAADBEFiIyIUiwjtHpXpyly6PjDZk5mMks5qPrcygaJpZM4I-leH>
> .
>



-- 

docs/hazmat/primitives/asymmetric/rsa.rst
```diff
@@ -330,6 +330,18 @@ Padding
     :class:`OAEP` should be preferred for encryption and :class:`PSS` should be
     preferred for signatures.
 
+
+.. function:: calculate_max_pss_salt_length(key, hash_algorithm)
```
reaperhulk: Needs a `.. versionadded:: 1.5`

reaperhulk: (and a changelog entry please)

PR #2826, opened 2016-03-14
OpenSSL 1.1.0 support

reaperhulk:
This PR makes the final changes required to compile against OpenSSL 1.1.0. It **must not** be merged until after the thirteenth release to avoid breaking pyOpenSSL.

PR #3017, opened 2016-06-26
Refs #2826 -- remove a no longer needed DSA binding

alex:


reaperhulk:
Looks loke switching away from this binding was done in pyopenssl 16.0 so :+1:

PR #3023, opened 2016-06-27
AES_ctr128_encrypt needs to be in macros as it's a conditional binding

reaperhulk:
Once 1.0.0 is dead we'll be able to just entirely remove this but for now it needs to be in macros.

### DIFF
--- a/src/_cffi_src/openssl/aes.py
+++ b/src/_cffi_src/openssl/aes.py
@@ -26,6 +26,9 @@ int AES_wrap_key(AES_KEY *, const unsigned char *, unsigned char *,
                  const unsigned char *, unsigned int);
 int AES_unwrap_key(AES_KEY *, const unsigned char *, unsigned char *,
                    const unsigned char *, unsigned int);
+"""
+
+MACROS = """
 /* The ctr128_encrypt function is only useful in 1.0.0. We can use EVP for
    this in 1.0.1+. */
 void AES_ctr128_encrypt(const unsigned char *, unsigned char *,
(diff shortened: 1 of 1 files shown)


src/_cffi_src/openssl/aes.py
```diff
@@ -26,22 +26,22 @@
...
                         size_t, const AES_KEY *, unsigned char[],
                         unsigned char[], unsigned int *);
 """
 
-MACROS = """
-"""
-
 CUSTOMIZATIONS = """
 static const long Cryptography_HAS_AES_WRAP = 1;
 #if OPENSSL_VERSION_NUMBER >= 0x10100000L && !defined(LIBRESSL_VERSION_NUMBER)
```
alex: Now that I look at this, can you explain the `OPENSSL_VERSION_NUMBER` check here?

alex:
I don't see this binding in my OpenSSL 1.1.0; can you explain what's going on here?

reaperhulk:
The compiler will error on this in 1.1.0 because the function isn't present and assigning a function pointer in this fashion doesn't satisfy the compiler's opinion of the function declaration.

alex:
Sorry, I don't follow. If the function doesn't exist in 1.1.0, what two things aren't matching?

reaperhulk:
```
void (*AES_ctr128_encrypt)(const unsigned char *, unsigned char *,
                            const size_t, const AES_KEY *,
                            size_t, const AES_KEY *,
                            unsigned char[], unsigned char[],
                            unsigned int *) = NULL;
```

does not satisfy the compiler as the implementation for the function declaration. This has always been true of our conditional bindings since you originally came up with the idea -- anything conditional has to sit in macros (for an example see cmac.py).

alex:
Hmm, so why wasn't this in macros before, it's always been conditional?

On Mon, Jun 27, 2016 at 10:22 PM, Paul Kehrer <notifications@github.com>
wrote:

> void (*AES_ctr128_encrypt)(const unsigned char *, unsigned char *,
>                             const size_t, const AES_KEY *,
>                             size_t, const AES_KEY *,
>                             unsigned char[], unsigned char[],
>                             unsigned int *) = NULL;
>
> does not satisfy the compiler as the implementation for the function
> declaration. This has always been true of our conditional bindings since
> you originally came up with the idea -- anything conditional has to sit in
> macros (for an example see cmac.py).
>
> —
> <https://github.com/pyca/cryptography/pull/3023#issuecomment-228930862>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAADBEJ0hu45-y8nJCMVSqgRAnrXDsbyks5qQIVKgaJpZM4I_mzk>
> .
>



-- 

reaperhulk:
It was in macros. It slipped into functions incorrectly as part of the 0.9.8 removal (my fault). This commit: https://github.com/pyca/cryptography/commit/40c6bf65

alex:
And nothing broke because we don't have any 1.1.0 in CI yet?

reaperhulk:
Correct. I'd add 1.1.0 but we'll fail on it until we can finish opaquing :(

PR #3037, opened 2016-07-02
Use a series of constants for OpenSSL version checks

alex:
N.B. I removed several qualifiers that were being used to express beta vs. release in OpenSSL version numbers. Reviewers please look closely!


src/_cffi_src/openssl/ssl.py
```diff
@@ -470,17 +470,19 @@
     return 1;
 }
 #endif
+
 /* Added in 1.0.2 but we need it in all versions now due to the great
    opaquing. */
-#if OPENSSL_VERSION_NUMBER < 0x10002001L || defined(LIBRESSL_VERSION_NUMBER)
```
reaperhulk: This isn't exactly the same comparison. We've had issues in the past with weird distros branching off beta versions of OpenSSL and then with insufficiently precise comparisons we get multiple definition bugs.

alex: Do you know of a specific platform impacted here? I'm reluctant to add more conditions for a theoretical case, given that all the CA passes.

reaperhulk: I don't remember specifically, but I know we had to change a definition much like this one to avoid a bug last year. You can probably dig it up if you hunt through the history, but that's a lot of work.

I'm extremely reluctant to merge a change like this though because I'm almost certain it will cause breakage as is.

alex: Ok; is it just this specific line or are there others as well?

On Tue, Jul 5, 2016 at 12:46 PM, Paul Kehrer <notifications@github.com>
wrote:

> In src/_cffi_src/openssl/ssl.py
> <https://github.com/pyca/cryptography/pull/3037#discussion_r69597499>:
>
> >  /* Added in 1.0.2 but we need it in all versions now due to the great
> >     opaquing. */
> > -#if OPENSSL_VERSION_NUMBER < 0x10002001L || defined(LIBRESSL_VERSION_NUMBER)
>
> I don't remember specifically, but I know we had to change a definition
> much like this one to avoid a bug last year. You can probably dig it up if
> you hunt through the history, but that's a lot of work.
>
> I'm extremely reluctant to merge a change like this though because I'm
> almost certain it will cause breakage as is.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/pyca/cryptography/pull/3037/files/d2bca036cbc79f1be53c92886048d4f267a7a13f#r69597499>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAADBC-5Usrb_FWA3b32Nei-vAYMZ-Hdks5qSoppgaJpZM4JDuXx>
> .
>



-- 

src/_cffi_src/openssl/x509.py
```diff
@@ -369,7 +369,7 @@
 #endif
 /* Added in 1.0.2 but we need it in all versions now due to the great
    opaquing. */
-#if OPENSSL_VERSION_NUMBER < 0x10002003L || defined(LIBRESSL_VERSION_NUMBER)
```
reaperhulk: We definitely want to keep this version (https://github.com/pyca/cryptography/commit/5d3965bbfad2364c45e3a5e4cb9cd11b87fd2d83)

src/_cffi_src/openssl/x509_vfy.py
```diff
@@ -196,7 +196,7 @@
 
 CUSTOMIZATIONS = """
 /* OpenSSL 1.0.2+ verification error codes */
-#if OPENSSL_VERSION_NUMBER >= 0x10002002L && !defined(LIBRESSL_VERSION_NUMBER)
```
reaperhulk: Probably need to handle this beta2 check separately as well (there's also one on line 215)

src/_cffi_src/openssl/cryptography.py
```diff
@@ -0,0 +1,56 @@
...
+#define CRYPTOGRAPHY_OPENSSL_102BETA2_OR_GREATER \
+    (OPENSSL_VERSION_NUMBER >= 0x10002002)
+#define CRYPTOGRAPHY_OPENSSL_110_OR_GREATER \
+    (OPENSSL_VERSION_NUMBER >= 0x10100000)
+
+#define CRYPTOGRAPHY_OPENSSL_LESS_THAN_101 \
+    (OPENSSL_VERSION_NUMBER < 0x10001000)
+#define CRYPTOGRAPHY_OPENSSL_LESS_THAN_102 \
+    (OPENSSL_VERSION_NUMBER < 0x10002000)
+#define CRYPTOGRAPHY_OPENSSL_LESS_THAN_102BETA1 \
```
reaperhulk: I think this can go away in favor of using LESS_THAN_102 (but the beta2 and beta3 checks should stay). Having said this, someone, somewhere will be using some crazy 1.0.2 forked before beta1 was even released...

alex: Done.

reaperhulk: Looks like you can remove this definition now since it's not used anywhere?

PR #3041, opened 2016-07-12
Fix DNSName wildcard encoding for NameConstraints

intgr:
Previously '.example.com' would get normalised to 'example.com', making it impossible to add wildcard NameConstraints.

### DIFF
--- a/tests/test_x509_ext.py
+++ b/tests/test_x509_ext.py
@@ -17,14 +17,31 @@ from cryptography import x509
 from cryptography.hazmat.backends.interfaces import (
     DSABackend, EllipticCurveBackend, RSABackend, X509Backend
 )
+from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.x509 import DNSName, NameConstraints, SubjectAlternativeName
 from cryptography.x509.oid import (
     AuthorityInformationAccessOID, ExtendedKeyUsageOID, ExtensionOID,
     NameOID, ObjectIdentifier
 )
 
+from .hazmat.primitives.fixtures_rsa import RSA_KEY_2048
 from .hazmat.primitives.test_ec import _skip_curve_unsupported
 from .test_x509 import _load_cert
+
+
+def _make_certbuilder(private_key):
+    name = x509.Name(
+        [x509.NameAttribute(NameOID.COMMON_NAME, u'example.org')])
+    return (
+        x509.CertificateBuilder()
+            .subject_name(name)
+            .issuer_name(name)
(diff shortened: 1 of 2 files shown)


reaperhulk:
Jenkins, ok to test.

tests/test_x509_ext.py
```diff
@@ -3526,3 +3529,34 @@ def test_invalid_certificate_policies_data(self, backend):
...
+
+
+class TestCertificateBuilderExtensions(object):
+    @pytest.mark.requires_backend_interface(interface=RSABackend)
+    @pytest.mark.requires_backend_interface(interface=X509Backend)
+    def test_nameconstraints(self, backend):
+        permitted = [u'.example.org', u'.\xf5\xe4\xf6\xfc.example.com',
+                     u'x2603.example.net']
+        private_key = RSA_KEY_2048.private_key(backend)
+        builder = _make_certbuilder(private_key)
```
reaperhulk: `_make_certbuilder` is likely a good idea to reduce some duplication across our tests, but since it's only in use in this one place I'd prefer that you inlined it here. I'd be happy to review a PR that uses this technique to simplify the tests in the future though!

tests/test_x509_ext.py
```diff
@@ -3526,3 +3529,34 @@ def test_invalid_certificate_policies_data(self, backend):
...
+        builder = builder.add_extension(
+            NameConstraints(permitted_subtrees=list(map(DNSName, permitted)),
+                            excluded_subtrees=[]), True)
+
+        cert = builder.sign(private_key, hashes.SHA1(), backend)
+        result = [x.value for x in cert.extensions[0].value.permitted_subtrees]
```
reaperhulk: Could you change this to use `get_extension_by_class(NameConstraints).value.permitted_subtrees`? While we do support indexing in this fashion I'd prefer it to be a bit more clear.

reaperhulk:
Sorry about the delay in reviewing this. Our decision to do unicode decoding continuously haunts me... This mostly looks good modulo the two comments I left!

intgr:
Updated the branch. I moved the new test into the `TestNameConstraintsExtension` class, which in hindsight seems more appropriate. And I added a new certbuilder test in `TestSubjectAlternativeName` because I also touched the code path dealing with SANs. So now there's a good reason to keep the `_make_certbuilder` function. :)

reaperhulk:
Thanks for your work on this! If you'd like to add your name to `AUTHORS.rst` feel free to send a PR!

intgr:
Oops, now I realize I had botched that commit message. Didn't see it before :(

Issue #3036, opened 2016-07-02
Purge our documentation of references to "providers"


alex:
These aren't a concept that's meaningful in Python, it's an allusion to `zope.interface`, which I claim most people aren't familiar with.

We should complete our simplification from "A {foo} provider" to "An instance of {foo}":

- [x] `docs/hazmat/backends/interfaces.rst`
- [x] `docs/hazmat/primitives/asymmetric/dh.rst`
- [x] `docs/hazmat/primitives/asymmetric/dsa.rst`
- [x] `docs/hazmat/primitives/asymmetric/ec.rst`
- [x] `docs/hazmat/primitives/asymmetric/rsa.rst`
- [x] `docs/hazmat/primitives/asymmetric/serialization.rst`
- [x] `docs/hazmat/primitives/cryptographic-hashes.rst`
- [x] `docs/hazmat/primitives/key-derivation-functions.rst`
- [x] `docs/hazmat/primitives/keywrap.rst`
- [x] `docs/hazmat/primitives/mac/cmac.rst`
- [x] `docs/hazmat/primitives/mac/hmac.rst`
- [x] `docs/hazmat/primitives/padding.rst`
- [x] `docs/hazmat/primitives/symmetric-encryption.rst`
- [x] `docs/hazmat/primitives/twofactor.rst`
- [x] `docs/x509/reference.rst`

PR #3052, opened 2016-07-21
Removed provider language from asymmetric primitives docs

gorisaka:
Refs #3036

### DIFF
--- a/docs/hazmat/primitives/asymmetric/dh.rst
+++ b/docs/hazmat/primitives/asymmetric/dh.rst
@@ -82,9 +82,8 @@ Key interfaces
         Generate a DH private key. This method can be used to generate many
         new private keys from a single set of parameters.
 
-        :return: A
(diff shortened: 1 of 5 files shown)


alex:
Jenkins, ok to test.

alex:
The examples which are just "A <foo>" are ok, it's only the "A <foo> provider" that need to be changed.

gorisaka:
Certainly but, in my opinion, being consistent with the phrasing would be better.

gorisaka:
But I can revise this PR.

alex:
I think minimizing the PR to just removing the ones that say "provider" would be best.

gorisaka:
Addressed feedback.

alex:
Ugh, looks like our CI is temporarily broken due to https://github.com/PyCQA/flake8-import-order/issues/79, so that's what's causing the failures.

alex:
Jenkins, retest this please.

PR #3053, opened 2016-07-22
Basic SecureTransport bindings.

Lukasa:
This is an early pass at SecureTransport bindings, exactly as written in [my scratch project](https://github.com/Lukasa/securetransportscratch). They will need revision to take into account cryptography's preferred style and supported features. We may also need to add additional bindings, depending on what we intend to use as the lowest supported SDK version.

This currently _also_ contains no private or unsupported APIs. Let's focus on the supported ones first, and then I'll come back and offer bindings for some of the private APIs as well, which the PyCA maintainers can consider separately.

### DIFF
--- a/src/_cffi_src/commoncrypto/secure_transport.py
+++ b/src/_cffi_src/commoncrypto/secure_transport.py
@@ -13,8 +13,21 @@ typedef ... *SSLContextRef;
 typedef const void *SSLConnectionRef;
 
 typedef enum {
+    kSSLServerSide,
+    kSSLClientSide
+} SSLProtocolSide;
+
+typedef enum {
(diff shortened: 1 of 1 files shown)


alex:
ready for a rebase

Lukasa:
Rebased.

alex:
@Lukasa tests failing here :-) If there's someway we can help work through these let me know

alex:
@Lukasa Is there a plan to work on this soon, or should we close this for now?

Lukasa:
I think I'm not planning to work on this further until I can sort out what the TLS ABC PEP would look like, so I'm happy for you to close this for a while if you'd like it cleaner. =)

PR #3055, opened 2016-07-22
Additional exclusion to keywords_with_side_effects

clayg:
At some point during build I do something like `python setup.py -k sdist` - but in my build cleanroom this is happening when mixing the source dist with the debian bits before cffi and stuff is installed, so it blows up.  It _seems_ like this code is trying to prevent exactly that and just missed an option, or maybe it's just supposed to work as soon as it sees `sdist`?  `sdist -k` or `-k sdist` didn't _seem_ to make any difference?  If this is the right way to go, I actually think `sdist` has a few other flags that should be added to this list...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -141,6 +141,7 @@ def keywords_with_side_effects(argv):
         '-q', '--quiet',
         '-v', '--verbose',
         '-V', '--version',
+        '-k', '--keep-temp',
         '--author',
         '--author-email',
         '--classifiers',


alex:
Jenkins ok to test.

alex:
@dstufft please advise :-)

clayg:
idk, I sorta get the feeling I'm doing something wrong ... in a different part of a different build some debian helper something does `python setup.py clean -a` and it blew up again...  I decided to just use a dh_override to make it just `python setup.py clean` w/o the `-a`

I tried to look at the code a little closer, and I sorta got the impression that `sdist` and `clean` wanted to pass through regardless of their options - but if that was the intention, I don't think it's quite coded up to work that way.

If no one is quite sure what that stuff in `setup.py` is there for anymore I'll probably have to try to do a better job of documenting how to reproduce the failure I'm seeing... and we'll just have to work back up to it.

alex:
Jenkins, retest this please.

reaperhulk:
cc @dstufft is there something we need to do here?

reaperhulk:
Closing due to lack of activity. We can reopen in the future if necessary.

PR #3062, opened 2016-07-27
Fixed openssl binding if no-cmac

jefferyto:
For instance, LEDE recently [removed CMAC](https://git.lede-project.org/?p=source.git;a=commit;h=41da31ac2ca2306e9cff070f650f6012af753517) from its OpenSSL package.


alex:
Jenkins, ok to test.

PR #3067, opened 2016-07-29
Disallow X509 certificate serial numbers bigger than 159 bits (#3064)

### LINKED ISSUE #3064: Non-RFC conformant X509 certificate serial number is generated
https://tools.ietf.org/html/rfc5280  4.1.2.2.  Serial Numbers

`Conforming CAs MUST NOT  use serialNumber values longer than 20 octets.`

Well, let's look to source code:

cryptography/hazmat/backends/openssl/backend.py :
```python
def _int_to_bn(self, num, bn=None):
    ...
    binary = num.to_bytes(int(num.bit_length() / 8.0 + 1), "big")
    ...
```

cryptography/x509/base.py:
```python
class CertificateBuilder:
    def serial_number(self, number):
        if utils.bit_length(number) > 160
```

Since serial number should be positive, for my example below it should add additional zero.

Sample program:
```python
import datetime
from cryptography import x509
from cryptography.hazmat.primitives.hashes import SHA256
from cryptography.x509 import NameOID
key = rsa.generate_private_key(public_exponent=65537, key_size=512, backend=default_backend())
serial = int.from_bytes(b'\xff' * 20, 'big')
subj = x509.Name([x509.NameAttribute(NameOID.COMMON_NAME, 'qwe')])
builder = x509.CertificateBuilder()
builder = builder.subject_name(subj)
builder = builder.issuer_name(subj)
builder = builder.not_valid_before(datetime.datetime.today())
builder = builder.not_valid_after(datetime.datetime.today())
builder = builder.serial_number(serial)
builder = builder.public_key(key.public_key())
cert = builder.sign(private_key=key, algorithm=SHA256(), backend=default_backend())
print(cert.public_bytes(encoding=serialization.Encoding.PEM).decode('ascii'))
```

`./qwe.py | openssl asn1parse -inform pem  -i` gives:

```
...
13:d=2  hl=2 l=  21 prim:   INTEGER           :FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
...
```

Note 21 byte (!)

socketpair:
FIxes #3064

### DIFF
--- a/tests/test_x509.py
+++ b/tests/test_x509.py
@@ -1689,12 +1689,55 @@ class TestCertificateBuilder(object):
 
     def test_serial_number_must_be_non_negative(self):
         with pytest.raises(ValueError):
-            x509.CertificateBuilder().serial_number(-10)
+            x509.CertificateBuilder().serial_number(-1)
+
+    def test_serial_number_must_be_positive(self):
+        with pytest.raises(ValueError):
+            x509.CertificateBuilder().serial_number(0)
+
+    @pytest.mark.requires_backend_interface(interface=RSABackend)
+    @pytest.mark.requires_backend_interface(interface=X509Backend)
+    def test_minimal_serial_number(self, backend):
+        subject_private_key = RSA_KEY_2048.private_key(backend)
+        builder = x509.CertificateBuilder().serial_number(
+            1
+        ).subject_name(x509.Name([
+            x509.NameAttribute(NameOID.COUNTRY_NAME, u'RU'),
+        ])).issuer_name(x509.Name([
+            x509.NameAttribute(NameOID.COUNTRY_NAME, u'RU'),
+        ])).public_key(
+            subject_private_key.public_key()
+        ).not_valid_before(
+            datetime.datetime(2002, 1, 1, 12, 1)
+        ).not_valid_after(
+            datetime.datetime(2030, 12, 31, 8, 30)
+        )
+        cert = builder.sign(subject_private_key, hashes.SHA256(), backend)
+        assert cert.serial_number == 1
+
+    @pytest.mark.requires_backend_interface(interface=RSABackend)
+    @pytest.mark.requires_backend_interface(interface=X509Backend)
+    def test_biggest_serial_number(self, backend):
+        subject_private_key = RSA_KEY_2048.private_key(backend)
+        builder = x509.CertificateBuilder().serial_number(
+            (1 << 159) - 1
+        ).subject_name(x509.Name([
+            x509.NameAttribute(NameOID.COUNTRY_NAME, u'RU'),
+        ])).issuer_name(x509.Name([
+            x509.NameAttribute(NameOID.COUNTRY_NAME, u'RU'),
+        ])).public_key(
+            subject_private_key.public_key()
+        ).not_valid_before(
+            datetime.datetime(2002, 1, 1, 12, 1)
+        ).not_valid_after(
+            datetime.datetime(2030, 12, 31, 8, 30)
+        )
+        cert = builder.sign(subject_private_key, hashes.SHA256(), backend)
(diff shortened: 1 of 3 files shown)


tests/test_x509.py
```diff
@@ -1691,10 +1691,45 @@ def test_serial_number_must_be_non_negative(self):
...
+            datetime.datetime(2002, 1, 1, 12, 1)
+        ).not_valid_after(
+            datetime.datetime(2030, 12, 31, 8, 30)
+        )
+        cert = builder.sign(subject_private_key, hashes.SHA256(), backend)
+        assert cert.serial_number == (1 << 159) - 1
+
     def test_serial_number_must_be_less_than_160_bits_long(self):
         with pytest.raises(ValueError):
-            # 2 raised to the 160th power is actually 161 bits
```
socketpair: There was test already (!) but it was broken (!)

```
In [1]: 2 ** 160
Out[1]: 1461501637330902918203684832716283019655932542976
In [2]: 1 << 159
Out[2]: 730750818665451459101842416358141509827966271488
```

alex:
Jenkins, ok to test

tests/test_x509.py
```diff
@@ -1691,10 +1691,45 @@ def test_serial_number_must_be_non_negative(self):
...
+            datetime.datetime(2002, 1, 1, 12, 1)
+        ).not_valid_after(
+            datetime.datetime(2030, 12, 31, 8, 30)
+        )
+        cert = builder.sign(subject_private_key, hashes.SHA256(), backend)
+        assert cert.serial_number == 0
+
+    @pytest.mark.requires_backend_interface(interface=RSABackend)
+    @pytest.mark.requires_backend_interface(interface=X509Backend)
+    def tets_biggest_serial_number(self, backend):
```
reaperhulk: typo here

tests/test_x509.py
```diff
@@ -1691,10 +1691,45 @@ def test_serial_number_must_be_non_negative(self):
         with pytest.raises(ValueError):
             x509.CertificateBuilder().serial_number(-10)
 
+    @pytest.mark.requires_backend_interface(interface=RSABackend)
+    @pytest.mark.requires_backend_interface(interface=X509Backend)
+    def tets_zero_serial_number(self, backend):
```
reaperhulk: typo here

tests/test_x509.py
socketpair: fixed

tests/test_x509.py
socketpair: fixed

socketpair:
Also, is it reasonable to add function, that sets random certificate serial number ? like
`.serial_number(int.from_bytes(os.urandom(20), 'big') >> 1)`

socketpair:
RFC 5280 4.1.2.2.  Serial Number:

```
 Note: Non-conforming CAs may issue certificates with serial numbers
   that are negative or zero.  Certificate users SHOULD be prepared to
   gracefully handle such certificates.
```

socketpair:
I'm not sure that revoked certificates should have that restrictions. Suppose we want to revoked certificate that was issued with 200-bit serial number (by mistake) ?

alex:
Jenkins, retest this please

reaperhulk:
I think we should go ahead and put the limitation in for now. We can relax it in the future if required, although I'd prefer to do that in a hypothetical hazmat version of this builder.

A random serial number generation function wouldn't be a bad idea, although that should be in a separate PR (and you would need to use the `cryptography.utils.int_from_bytes` function to make sure it works on Python 2)

socketpair:
Well, everything is fixed, will you merge that ?

reaperhulk:
Thanks for your contribution! If you'd like to send a PR to add yourself to `AUTHORS.rst` feel free!

PR #3078, opened 2016-08-01
Allow passing iterators where collections are expected

intgr:
This has tripped me up on several occasions. Iterators can only be enumerated once, breaking code like this in Python 3 for example:
```
san = SubjectAlternativeName(map(DNSName, lst))
```

This is also a slight behavior change if the caller modifies the list _after_ passing it to the constructor, because input lists are now copied. Which seems like a good thing.

Also:
* Name now checks that attributes elements are of type NameAttribute
* NoticeReference now allows notice_numbers to be any iterable

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -174,13 +174,15 @@ class AuthorityKeyIdentifier(object):
                 "must both be present or both None"
             )
 
-        if authority_cert_issuer is not None and not all(
-            isinstance(x, GeneralName) for x in authority_cert_issuer
-        ):
-            raise TypeError(
-                "authority_cert_issuer must be a list of GeneralName "
-                "objects"
-            )
+        if authority_cert_issuer is not None:
+            authority_cert_issuer = list(authority_cert_issuer)
+            if not all(
+                isinstance(x, GeneralName) for x in authority_cert_issuer
+            ):
+                raise TypeError(
+                    "authority_cert_issuer must be a list of GeneralName "
+                    "objects"
+                )
 
         if authority_cert_serial_number is not None and not isinstance(
             authority_cert_serial_number, six.integer_types
@@ -273,6 +275,7 @@ class AuthorityInformationAccess(object):
     oid = ExtensionOID.AUTHORITY_INFORMATION_ACCESS
 
     def __init__(self, descriptions):
+        descriptions = list(descriptions)
         if not all(isinstance(x, AccessDescription) for x in descriptions):
             raise TypeError(
                 "Every item in the descriptions list must be an "
@@ -386,6 +389,7 @@ class CRLDistributionPoints(object):
     oid = ExtensionOID.CRL_DISTRIBUTION_POINTS
 
     def __init__(self, distribution_points):
+        distribution_points = list(distribution_points)
         if not all(
             isinstance(x, DistributionPoint) for x in distribution_points
         ):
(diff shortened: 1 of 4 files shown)


intgr:
Updated: I found a few more places with the same problem. Writing tests for all 12 them would probably be excessive... Or would it?

alex:
Jenkins, ok to test

reaperhulk:
@intgr Since it's a problem that we don't want to regress on in the future I'd say having a test for every case (even if it's largely copy/paste) makes sense.

src/cryptography/x509/name.py
```diff
@@ -54,6 +54,10 @@ def __repr__(self):
 
 class Name(object):
     def __init__(self, attributes):
+        attributes = list(attributes)
+        if not all(isinstance(x, NameAttribute) for x in attributes):
+            raise TypeError("attributes must be a list of NameAttribute")
```
intgr: Behavior change: Name now checks that attributes elements are of type NameAttribute

src/cryptography/x509/extensions.py
```diff
@@ -676,9 +683,8 @@ def __ne__(self, other):
 class NoticeReference(object):
     def __init__(self, organization, notice_numbers):
         self._organization = organization
-        if not isinstance(notice_numbers, list) or not all(
-            isinstance(x, int) for x in notice_numbers
-        ):
+        notice_numbers = list(notice_numbers)
+        if not all(isinstance(x, int) for x in notice_numbers):
```
intgr: Behavior change: NoticeReference now allows notice_numbers to be any iterable, not just list

tests/test_x509_ext.py
```diff
@@ -1400,18 +1434,18 @@ def test_not_name(self):
 
     def test_repr(self):
         name = x509.Name([x509.NameAttribute(NameOID.COMMON_NAME, u'value1')])
-        gn = x509.DirectoryName(x509.Name([name]))
+        gn = x509.DirectoryName(name)
```
intgr: Fixed these tests, I assume these were unintentional, caught by the new check in `Name.__init__`

intgr:
Pull request updated. I also follow the same pattern for `Name(attributes=...)` arguments now.

Also:
* `Name` now checks that `attributes` elements are of type `NameAttribute`
* `NoticeReference` now allows `notice_numbers` to be any iterable

`DistributionPoint` requires its `reasons` argument to always be `frozenset`. I suppose this could made to accept any iterable as well, but I didn't change it because there are tests for that behavior. Is there a good reason why it requires a frozenset?

intgr:
Ping?

tests/test_x509_ext.py
reaperhulk: Great catch, but would you mind putting these fixes in as a separate PR and removing them from this one? Sorry to make you do some extra work :(

reaperhulk:
This looks good to me other than moving the test fixes to a separate PR...and having reviewed this now I'm actually okay with those fixes not being separated.

I'm a bit concerned about how to ensure that we're consistent with supporting iterators going forward, but we can talk more about that in the future.

Thanks for the contribution!

intgr:
Thanks!

> I'm a bit concerned about how to ensure that we're consistent with supporting iterators going forward, but we can talk more about that in the future.

Hmm, one approach would be using something like [attrs](https://attrs.readthedocs.io/en/stable/) for these classes, that way much of the boilerplate can be deleted and it's only necessary to declare something like `permitted_subtrees = attr.ib(convert=list, validator=elements_instance_of(GeneralName))`, which is more difficult to omit by accident.

But I'm not sure if you want to pull in an additional dependency for that. Or include a copy in the cryptography source tree.

PR #3081, opened 2016-08-05
cx_freeze support for default_backend

wolfthefallen:
Fix allows for cx_freeze, pyinstaller, ect to work with cryptography.hazmat.backends default_backend. When cryptography is export to binary for windows utilzing cx_freeze or other such tools, it fails to load any backends for default_backend, and provide the following error when called:

`"Multibackend cannot be initialized with no backends. If you are seeing this error when trying to use default_backend() please try uninstalling and reinstalling cryptography."'`

This fixs attempts to load known backends for crptography and provides them for MultiBackend when default_backend is called and no backends where provided.


alex:
Jenkins, ok to test.

alex:
This seems like a pretty ugly workaround for what is really a bug in those libraries: they attempt to infer what modules you're using, and they guess wrong :-) Far more ideal would be if they offered some sort of "must include modules" option.

wolfthefallen:
The library I currently see this issue in is Paramiko>=2.0.0 and utilizes default_backend. Which works just fine until the application is built utilizing cx_freeze, pyinstaller, or other such freeze program. After which it appears the entry point are not frozen or accessible by the exe program. 

I see that there has been several on going issues with freezing cryptography in the past, primarily with the use of default_backend. What is your recommendation to resolve this issue short of having all packages that utilizing cryptography stop utilizing default_backend?

saghul:
FWIW, same here.  How would everyone feel with a similar approach to this patch which tries to manually import certain backends `if hasattr(sys, 'frozen'): ...` ? AFAIK it's a convention all executable bundling framework thingies follow, to set `sys.frozen`.

reaperhulk:
The purist in me wants to say this isn't cryptography's bug and therefore the fix should reside outside of our library. Especially since the purpose of our entry points is to allow alternate backends to be plugged in. However, the alternate backend concept is still largely theoretical and we do still bundle the openssl backend. I think I'd be okay with a sys.frozen check that would pull in the openssl backend then? Would you be interested in submitting a PR along those lines @saghul?

wolfthefallen:
updated to check to see if sys is frozen, before manually attempting to import known packaged backends

src/cryptography/hazmat/backends/multibackend.py
```diff
@@ -12,7 +12,6 @@
     PEMSerializationBackend, RSABackend, ScryptBackend, X509Backend
 )
 
-
```
reaperhulk: This line needs to be re-added to pass pep8

reaperhulk:
I've grabbed the code here and opened a new PR (#3235) with a few tweaks. Thanks for your work on this @wolfthefallen. Once the new PR lands (assuming we decide to merge it) please feel free to send a PR adding yourself to `AUTHORS.rst`!

PR #3088, opened 2016-08-15
[twofactor] Update example code to use recommended 160 bits

brondsem:
I found the examples with `os.urandom(16)` generated URIs that Google Authenticator and Duo two-factor apps did not even recognize as supported.  This increases the key to the recommended 160 bits, and the URIs now work with both of those apps.


alex:
Jenkins, ok to test.

reaperhulk:
Looks good to me, thanks for the contribution!

PR #3093, opened 2016-08-16
ERR_load_RAND_strings changed function signature in 1.1.0

reaperhulk:
This is a hack to avoid breaking pyOpenSSL. Better suggestions welcome.


src/_cffi_src/openssl/rand.py
```diff
@@ -24,6 +23,12 @@
 """
 
 MACROS = """
+/* ERR_LOAD_RAND_STRINGS started returning an int in 1.1.0. Unfortunately we
```
alex: Name isn't all-caps; please make this consistent.

Issue #1209, opened 2014-07-02
Relicense as dual Apache / BSD 2 clause


alex:
This will allow folks with GPLv2 codebases to use us, and is a blocker for `PyCrypto` to be implemented on top of us. I've consulted with an IP lawyer to verify that this will allow more people to use Cryptography, while still giving us the protections of the Apache license.

Tasks:
- [x] Update `LICENSE` file appropriately
- [x] Update `setup.py` classifiers
- [x] Update `__about__.py` and `cryptography_vectors/__about__.py`
- [x] Figure out what to do with copyright headers in every file
- [x] Update contributing documentation
- [x] Receive consent of all folks who have contributed to cryptography in the past.
- [x] Changelog entry.

List of folks whose consent we need:
- [x] Alex Gaynor
- [x] Paul Kehrer
- [x] David Reid
- [x] Alex Stapleton
- [x] Donald Stufft
- [x] Ayrx
- [x] Laurens Van Houtven
- [x] Mohammed Attia
- [x] Jean-Paul Calderone
- [x] cyli
- [x] Hynek Schlawack
- [x] Julian Krause
- [x] Richard Wall
- [x] Ashwini Oruganti
- [x] PhiBo
- [x] Matthew Iversen
- [x] Christian Heimes
- [x] Chris Glass
- [x] Chirag Jog
- [x] Kubilay Kocak
- [x] Konstantinos Koukopoulos
- [x] timic
- [x] ashfall
- [x] Nic Young
- [x] Jeff Tang
- [x] Fedor Brunner
- [x] Daniele Sluijters
- [x] Cory Benfield
- [x] skeuomorf
- [x] Matt Iversen
- [x] Kyle Kelley
- [x] Kimmo Parviainen-Jalanko
- [x] manuels
- [x] kennwhite
- [x] Wouter Bolsterlee
- [x] Thomas Fenzl
- [x] Taewook Kang
- [x] Stephen Holsapple
- [x] Sascha Peilicke
- [x] Marcin Wielgoszewski
- [x] Jonathan Giannuzzi
- [x] Jay Parlar
- [x] D.S. Ljungmark
- [x] Arturo Filastò

PR #3097, opened 2016-08-23
correctly apply the Apache-2.0 License

graingert:


graingert:
According to the Apache 2.0 license you need to include the boilderplate, and you really need to include the copyright holders: https://github.com/pyca/cryptography/pull/3097/files#diff-0a77fdccb61e3d61ba13949e89b5bf80R1

sigmavirus24:
This is stemming from sigmavirus24/uritemplate#25. I'm fairly confident this can be closed.

graingert:
actually it looks like https://github.com/pyca/cryptography/issues/1209 "Figure out what to do with copyright headers in every file" was never done

sigmavirus24:
1. This change would have no effect on that.
2. In sigmavirus24/uritemplate#25 I linked you an example of how copyright is notified in *each* file in this project. Someone forgot to check a box on that issue. That is all.

graingert:
sure, but I think there's still an issue here. I'm not sure what Van Lindberg was consulted on, but I'm like 60% sure he would not contradict the explicit wording of the Apache-2.0 license.

reaperhulk:
When we consulted our lawyer it was explicitly to discuss relicensing to from Apache 2 to dual Apache 2 and BSD 3 clause. Within this discussion was a question about license headers, from which the current solution arose.

In this case it appears your concerns arise from a section after the end of terms and conditions and relates to one way (and notably not the *only* way) to make users of the software aware of the license it is provided under. That summary is just my opinion of this and I am definitively not a lawyer, but as we have consulted an IP lawyer about this very situation we're satisfied with our current approach.

PR #3101, opened 2016-08-24
remove two more constants that no longer exist and we don't use

reaperhulk:


alex:
Jenkins, retest this please.

PR #3102, opened 2016-08-24
two more functions that became const, one removed that we don't use

reaperhulk:


alex:
Jenkins, retest this please.

PR #3107, opened 2016-08-26
be a bit more robust about detecting locking callback declarations

reaperhulk:


alex:
Can you explain what's going on here/what this is fixing?

reaperhulk:
For 1.1.0 we originally tried to detect that the locking callbacks were gone by looking for `CRYPTO_LOCK` but in later betas OpenSSL added that back. So now it does an actual version check, but also checks for each type because the defines may not be present if a user activates `OPENSSL_NO_DEPRECATED` during compilation.

PR #3118, opened 2016-08-29
Mention that blake2 is not vulnerable to length-extension attacks

alex:


docs/hazmat/primitives/cryptographic-hashes.rst
```diff
@@ -120,7 +120,9 @@ SHA-2 family
 BLAKE2
 ~~~~~~
 
-`BLAKE2`_ is a cryptographic hash function specified in :rfc:`7693`.
+`BLAKE2`_ is a cryptographic hash function specified in :rfc:`7693`. BLAKE2 is
+not vulnerable to `length-extension attacks`_, which the SHA-family of hashes
```
reaperhulk: How about: BLAKE2 is not vulnerable to length-extension attacks, unlike the SHA-family of hashes.

PR #3123, opened 2016-08-30
fix an overindented line. not sure why our linters didn't catch this

reaperhulk:


reaperhulk:
jenkins, retest this please

PR #3130, opened 2016-09-02
Add bounds checking for Scrypt parameters.

### LINKED ISSUE #3129: scrypt bounds checking
```
[11:23:58]  <Alex_Gaynor>	reaperhulk: what happens if you pass a non-even n?
[11:24:10]  <Alex_Gaynor>	Or a negative value for any of the params?
```

Presumably it will fail with an assertion error on return from the call to `EVP_PBE_scrypt`, but we shouldn't allow those types of errors.

cc @Ayrx.

Ayrx:
Fixes #3129.

### DIFF
--- a/src/cryptography/hazmat/primitives/kdf/scrypt.py
+++ b/src/cryptography/hazmat/primitives/kdf/scrypt.py
@@ -25,6 +25,16 @@ class Scrypt(object):
         self._length = length
         if not isinstance(salt, bytes):
             raise TypeError("salt must be bytes.")
+
+        if n < 2 or (n & (n - 1)) != 0:
+            raise ValueError("n must be more than 1 and be a power of 2.")
+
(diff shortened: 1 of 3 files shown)


src/cryptography/hazmat/primitives/kdf/scrypt.py
```diff
@@ -25,6 +25,16 @@ def __init__(self, salt, length, n, r, p, backend):
         self._length = length
         if not isinstance(salt, bytes):
             raise TypeError("salt must be bytes.")
+
+        if n < 2 or (n & (n - 1)) != 0:
+            raise ValueError("n must be more than 1 and be a power of 2.")
```
alex: s/more/greater/

Ayrx:
This should be ready now.

PR #3139, opened 2016-09-04
Bind a pair of mem functions.

alex:


src/_cffi_src/openssl/crypto.py
```diff
@@ -57,6 +58,16 @@
 
 /* This was removed in 1.1.0 */
 void CRYPTO_lock(int, int, const char *, int);
+
+/* Signature changed significantly in 1.1.0, only expose there for sanity */
```
reaperhulk: This comment is confusing to me -- I think you mean we only expose it on 1.1.0+ and conditionally remove it on anything older?

alex:
Closing in favor of #3140, which is the only real consumer of this.

PR #3140, opened 2016-09-05
Memleak tests

alex:
They don't work,  because something calls `OpenSSL_malloc` in the subprocess before we initialize it. I assume this comes from the `Binding` init process

### DIFF
--- a/src/_cffi_src/openssl/crypto.py
+++ b/src/_cffi_src/openssl/crypto.py
@@ -33,6 +33,13 @@ static const int CRYPTO_LOCK_SSL;
 
 FUNCTIONS = """
 int CRYPTO_mem_ctrl(int);
+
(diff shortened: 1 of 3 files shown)


alex:
So, very close. For some reason it fails on py35 and py26 on windows... and the tracebacks make no sense to me.

@reaperhulk do you have opinions about how to dive in? really a pdb in there would go a long ways.

alex:
Ok, figured out py26, it was warnings on stderr. I still have absolutely no clue about py35

alex:
Boom, tests are green now!

reaperhulk:
Do we intend to run more substantial tests with the memory leak checks enabled in the future?

alex:
@reaperhulk I plan to add specific memleak tests for known historical leaks. I don't see a practical way to run our existing tests with leak checking.

PR #3152, opened 2016-09-15
Fix Fernet docs (changed key size to 256 bits)

sh4r3m4n:
Hi, looking at the source code of the Fernet cipher implementation i saw that it uses a 32 bytes (256 bits) key size. In the documentation it says its size it is of 128 bits so if i'm right it should be fixed.


reaperhulk:
Fernet uses a 32 byte key, but 16 bytes are AES key and 16 bytes are HMAC key so the 128-bit number was correct.

alex:
The docs are correct. Fernet has a 256 bit key, but 128 of those bits are for integrity. The AES key is only 128 bits.

sh4r3m4n:
Ok, i dint't realized it, sorry!

PR #3153, opened 2016-09-15
minor constant_time.bytes_eq refactor

Ofekmeister:
conditional return put inside one function


alex:
I don't think this significantly improves the readability of the code. Thanks for contributing though!

PR #3162, opened 2016-09-22
1.0.2i changed the way COMP_METHOD is exported if NO_COMP is set

reaperhulk:


src/_cffi_src/openssl/ssl.py
```diff
@@ -664,7 +664,8 @@
 #endif
 
 
-#if defined(OPENSSL_NO_COMP) || defined(LIBRESSL_VERSION_NUMBER)
+#if (defined(OPENSSL_NO_COMP) && CRYPTOGRAPHY_OPENSSL_LESS_THAN_102I) || \
```
alex: Please add a comment.

alex:
`CRYPTOGRAPHY_OPENSSL_LESS_THAN_102I` needs to be exposed in `cryptography.py`, not just declared

raycoll:
Any chance this will be backported to the older versions: 1.2.x, 1.3.x, 1.5.x?

reaperhulk:
cryptography has historically chosen to only support the latest version. Is there a reason you need to compile against a recent OpenSSL but can't run the latest cryptography?

PR #3170, opened 2016-09-26
test against 1.1.0a instead of 1.1.0

reaperhulk:


.travis/install.sh
```diff
@@ -65,7 +65,7 @@ else
         OPENSSL_DIR="ossl-100t"
     fi
     if [[ "${OPENSSL}" == "1.1.0" ]]; then
-        OPENSSL_VERSION_NUMBER="1.1.0"
+        OPENSSL_VERSION_NUMBER="1.1.0a"
         OPENSSL_DIR="ossl-110"
```
alex: So, master will do the wrong thing because of the cache, right?

reaperhulk: Ack, probably. Guess we'll see!

alex: It looks wrong on https://travis-ci.org/pyca/cryptography/jobs/162685354 as well

PR #3175, opened 2016-09-26
bump version and add changelog entry for 1.5.2

reaperhulk:


alex:
Confirmed that all the OS X and Windows builders are using 1.0.2j. Waiting for tests to pass now.

PR #3188, opened 2016-10-03
Update installation.rst

mattypiper:
Update openssl https URL, otherwise a 302 result screws up the curl/tar steps and confusion ensues.


alex:
Jenkins, ok to test.

alex:
tbh, this appears to work fine with the version of curl I have, but it is 301'ing, so why not.

PR #3192, opened 2016-10-08
[WIP] [Question] Add incremental API for `fernet` encryption

alexander255:
I know this is still lacking in several ways, but could somebody from the *corer* development team give feedback on whether this is something that is in line with the project goal?

### DIFF
--- a/src/cryptography/fernet.py
+++ b/src/cryptography/fernet.py
@@ -10,8 +10,6 @@ import os
 import struct
 import time
 
-import six
-
 from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes, padding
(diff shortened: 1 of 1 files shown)


alex:
This API involves exposing un-authenticated plaintext to users, and that is not consistent with the safety goals of the recipe layer, so we will not support this functionality.

alexander255:
You mean the fact that `_Decryptor` returns data sets without waiting for the completion of the HMAC validation?

alexander255:
Would it make sense to expose the `Decryptor` in `hazmat` instead?

reaperhulk:
If you're looking for a streaming API with authentication we already support that in the form of the AES GCM construct -- does this not meet your needs? Fernet is meant to be a one shot API specifically to prevent possible misuse of data before the payload is authenticated. This unfortunately makes it unsuitable for encryption of large payloads. We have some vague ideas for having a streaming/seekable encryption format like Fernet using framing, but for now the AES GCM API can give you what you want at the cost of no integrity guarantees until decryption is complete.

alexander255:
@reaperhulk You're right about that. I'm looking a way to efficiently stream large (100MB+) datasets using one-shot symmetrical encryption between clients over an potentially untrusted network of servers. Considering that the data indeed might be processed/disolayed in some way before the download completes, I must admit that trying to stream fernet was a bad idea…
OTOH simply chunking the data into ~10kB packets that are each encrypted separately (using the same key, but different IV) would add about 0.6% overhead without impairing the used crypto in any way, right?

reaperhulk:
Yep, framing in that fashion is fine for what you want, just not what we'd like to do as a generic solution. Of course, I'd be remiss if I didn't say you should really be using TLS if at all possible for network communication.

alexander255:
Ok, thanks for the feedback! One more thing I'm wondering about: Is it safe to just throw an unsigned length field in front of each fernet packet to declare the following packet's total length? Unless I'm missing something obvious (and this thread serves as public documentation that such a thing has happened before) the HMAC should "never" realistically be able to validate if the length was wrong, right?
And yes, I'm very painfully aware about not being to use TLS in this case. :confused:

PR #3193, opened 2016-10-10
encrypt our IRC channel to prevent forks from notifying.

reaperhulk:
Copying what pypa/pip already does.

Also remove the webhook for the buildtrends, which we don't use.


reaperhulk:
jenkins, retest this please

Issue #3195, opened 2016-10-14
Deadlock involving RSA PKeys


aglasgall:
Test case:

```
from OpenSSL.crypto import FILETYPE_PEM, load_privatekey, sign
import threading
import time
import sys
import traceback

def start_workers(N, key):

    for n in range(N):
        new_thread = threading.Thread(
            target=worker,
            name="daemon-worker-thread-{}".format(n),
            args=(n, key, ))
        new_thread.daemon = True
        print('start {}'.format(n))
        new_thread.start()


def worker(idx, key):
    try:
        while True:
            pkey = load_privatekey(FILETYPE_PEM, key)
            sign(pkey, 'some message that doesnt really matter', "sha256")
            print('worker {} did something!'.format(idx))
    except:
        traceback.print_exc()


if __name__ == '__main__':
    kfile = sys.argv[1]
    with open(kfile, 'rb') as fh:
        key = fh.read()

    start_workers(8, key)

    time.sleep(60)
    print('I am done')
```

Run with the path to a PEM-encoded RSA private key as an argument. It doesn't deadlock every time., but it definitely does at least a third to half of the time.

The cryptography locking callback is installed because python and cryptography are separately statically linked against openssl

aglasgall:
This may be caused by GC being triggered from the locking callback, resulting eventually in a finalizer for a PKey calling the locking callback.

aglasgall:
update: okay, I know that this case uses pyopenssl instead of just cryptography, but it the hang is definitely in cryptography's openssl lock callback, so...

reaperhulk:
Yep, this is cryptography's problem. What version of Python are you seeing this on? I've been unable to replicate so far on 2.7.10 on macOS, but that doesn't mean much.

aglasgall:
3.5.either 1 or 2 on Linux (built from source) against OpenSSL 1.0.2, uh, g, I think.

aglasgall:
Stack traces for the deadlock that we saw with 1.2.2 (possibly _not_ the same as the deadlock exhibited by this test case) follow:

py-bt:

```
Thread 18 (LWP 11012):
Traceback (most recent call first):
  <built-in method acquire of _thread.lock object at remote 0x7f2fb5935b98>
  File "[redacted]/lib/cryptography/hazmat/bindings/openssl/binding.py", line 189, in _lock_cb
    lock.acquire()
  <built-in method EVP_SignFinal of CompiledLib object at remote 0x7f2fb62cb848>
[...]

PR #3201, opened 2016-10-18
Remove Python OpenSSL locking callback and replace it with one in C

aglasgall:
The Python OpenSSL locking callback is unsafe; if GC is triggered
during the callback's invocation, it can result in the callback being
invoked reentrantly, which can lead to deadlocks. This patch replaces
it with one in C that gets built at compile time via cffi along with
the rest of the OpenSSL binding.

This is a straight lift of the callback and setup code from Python 3.5.2's _ssl module; the only changes were removing #ifs conditioned on various openssl versions because cryptography's minimum openssl version is higher than any of the conditioned blocks.

There are still several issues remaining after replacing it; in particular, the interpreter at least sometimes hangs on exit unless cryptography.hazmat.backends.openssl.backend.activate_builtin_random() is used (and though I've not seen it deadlock/hang if it is, that doesn't mean it won't), and I've also seen it dump core and abort because:

Fatal Python error: could not acquire lock for <_io.BufferedWriter name='<stdout>'> at interpreter shutdown, possibly due to daemon threads

So we still have work to do, but this is a start.

This is for #3195.

### DIFF
--- a/src/_cffi_src/openssl/callbacks.py
+++ b/src/_cffi_src/openssl/callbacks.py
@@ -5,6 +5,7 @@
 from __future__ import absolute_import, division, print_function
 
 import sys
+import sysconfig
 
 import cffi
 
@@ -12,6 +13,13 @@ INCLUDES = """
 #include <openssl/ssl.h>
 #include <openssl/x509.h>
 #include <openssl/x509_vfy.h>
+#include <openssl/crypto.h>
+
+"""
+
+if sysconfig.get_config_var("WITH_THREAD"):
+    INCLUDES += """
+#include <pythread.h>
 """
 
 TYPES = """
@@ -34,6 +42,7 @@ extern "Python" int Cryptography_pem_password_cb(char *, int, int, void *);
  */
 extern "Python" int Cryptography_rand_bytes(unsigned char *, int);
 extern "Python" int Cryptography_rand_status(void);
+
 """
 
 FUNCTIONS = """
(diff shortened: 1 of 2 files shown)


alex:
Jenkins, ok to test.

aglasgall:
Build failures are I think because I imported sysconfig to get at WITH_THREAD to mimic _ssl.c's behavior of only installing the callback if threading is enabled. There's probably a better way of doing that.

src/_cffi_src/openssl/callbacks.py
```diff
@@ -50,4 +59,81 @@
     # backwards compatibility for old cffi version on PyPy
     # and Python >=3.5 (https://github.com/pyca/cryptography/issues/2970)
     TYPES = "static const long Cryptography_STATIC_CALLBACKS;"
-    CUSTOMIZATIONS = "static const long Cryptography_STATIC_CALLBACKS = 0;"
+    CUSTOMIZATIONS = """static const long Cryptography_STATIC_CALLBACKS = 0;
+"""
+
+
+if sysconfig.get_config_var("WITH_THREAD"):
+    FUNCTIONS += """
```
tiran: I think you need an additional guard for OpenSSL < 1.1.0 and not LibreSSL. OpenSSL 1.1.0 has its own locking code.

reaperhulk:
@tiran Is it possible to reconcile the Python license with our Apache/BSD dual license? I want to be sure we're in the clear on code import like this, but I am utterly unfamiliar with the actual provisions of the license CPython is provided under.

src/_cffi_src/openssl/callbacks.py
```diff
@@ -5,13 +5,21 @@
...
 INCLUDES = """
 #include <openssl/ssl.h>
 #include <openssl/x509.h>
 #include <openssl/x509_vfy.h>
+#include <openssl/crypto.h>
+
+"""
+
+if sysconfig.get_config_var("WITH_THREAD"):
+    INCLUDES += """
```
reaperhulk: Style nit, let's put this all on one line.

src/_cffi_src/openssl/callbacks.py
```diff
@@ -34,6 +42,7 @@
  */
 extern "Python" int Cryptography_rand_bytes(unsigned char *, int);
 extern "Python" int Cryptography_rand_status(void);
+
```
reaperhulk: Style nit, no newline here.

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -178,43 +179,17 @@ def _ensure_ffi_initialized(cls):
...
+            if sysconfig.get_config_var("WITH_THREAD"):
+                # Use Python's implementation if available, importing _ssl
+                # triggers the setup for this.
+                __import__("ssl")
+                if cls.lib.CRYPTO_get_locking_callback() != cls.ffi.NULL:
+                    return
+                # If nothing else has setup a locking callback already, we set up
+                # our own
+                if not lib._setup_ssl_threads():
+                    raise InternalError("Threading is enabled and cryptography is"
```
reaperhulk: Instead of having this branch let's just do an `_openssl_assert`. Otherwise we'll have a coverage issue since we can't test this branch.

src/_cffi_src/openssl/callbacks.py
reaperhulk: I think 1.1.0 just noops the methods, but I agree we should guard.

tiran:
@reaperhulk I'm not a lawyer. You can ask VanL about the license.

reaperhulk:
To keep this PR updated: I have an open question to Van :)

alex:
Closing this in favor of #3226 which I'll use to drive this to completion. Thanks for diagnosing this and getting the ball rolling on a patch.

Issue #3199, opened 2016-10-18
Add support for multi-valued RDNs in Name


frasertweedale:
If a DN contains multi-valued RDNs, `Name` does expose all the attributes but does not
indicate which attributes go together in a set.  This could lead to bugs and even security
issues in applications using _cryptography_ due to lack of ability to distinguish between DNs,
e.g.

`cn=alice+uid=12345,dc=example,dc=com` versus `cn=alice,uid=12345,dc=example,dc=com`

I will be happy to implement support.  Let us first discuss API.

`Name` currently:
- is an iterable yielding each AVA in the DN, in X.500 "order" (order of AVAs from a single, multi-valued RDNs is unspecified)
- provides `get_attributes_for_oid` which returns attributes for the given OID in X.500 "order" (multi-valued RDNs do not affect order here, because RFC 5280 constrains each RDN to contain at most one value for any attribute type).
- is hashable

Proposal:
1. change `_attributes` to be a `list` of `set` of `NameAttribute`
2. add a new property `rdns` (name tentative) which yields the RDNs
3. preserve existing behaviour of `__len__` and `__iter__`, but document that iteration order of AVAs from a multi-valued RDN is unspecified
4. update `__eq__`, `__hash__`, etc to respect the full structure of a Name.
5. change `Name.__init__` to accept an iterable of `NameAttribute` (the existing API) **or** an iterable of `set` of `NameAttribute`.  In the former case, each individual `NameAttribute` shall be promoted into a singleton `set`.

PR #3202, opened 2016-10-19
Name: add support for multi-value RDNs

frasertweedale:
Implement initial design proposal for #3199.

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -1102,6 +1102,18 @@ X.509 CSR (Certificate Signing Request) Builder Object
     slash or comma delimited string (e.g. ``/CN=mydomain.com/O=My Org/C=US`` or
     ``CN=mydomain.com, O=My Org, C=US``).
 
+    Technically, a Name is a list of *sets* of attributes, called *Relative
+    Distinguished Names* or *RDNs*, although multi-valued RDNs are rarely
+    encountered.  The iteration order of values within a multi-valued RDN is
+    undefined.  If you need to handle multi-valued RDNs, the ``rdns`` property
+    gives access to an ordered list of :class:`RelativeDistinguishedName`
+    objects.
+
+    A Name can be initialized with an iterable of :class:`NameAttribute` (the
+    common case where each RDN has a single attribute) or an iterable of
+    :class:`RelativeDistinguishedName` objects (in the rare case of
+    multi-valued RDNs).
+
     .. doctest::
 
         >>> len(cert.subject)
@@ -1111,6 +1123,12 @@ X.509 CSR (Certificate Signing Request) Builder Object
         <NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.6, name=countryName)>, value=u'US')>
         <NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.10, name=organizationName)>, value=u'Test Certificates 2011')>
         <NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.3, name=commonName)>, value=u'Good CA')>
+
+    .. attribute:: rdns
+
+        .. versionadded:: 1.6
+
+        :type: list of :class:`RelativeDistinguishedName`
 
     .. method:: get_attributes_for_oid(oid)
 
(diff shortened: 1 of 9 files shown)


alex:
Jenkins, ok to test.

frasertweedale:
X509_NAME_ENTRY * became an opaque type in openssl 1.1.0;
there a function to get at the needed value but I just need to figure out
how to handle the different versions

frasertweedale:
phew, green at last

src/cryptography/hazmat/backends/openssl/decode_asn1.py
```diff
@@ -548,17 +556,17 @@ def _decode_crl_distribution_points(backend, cdps):
             else:
                 rns = cdp.distpoint.name.relativename
                 rnum = backend._lib.sk_X509_NAME_ENTRY_num(rns)
-                attributes = []
+                attributes = set()
```
reaperhulk: Could you clarify what's going on here? As I understand it this change will make every name entry be added to a single set, which means that all CRLDistributionPoint relativename names are multi-valued?

src/_cffi_src/openssl/x509name.py
```diff
@@ -76,4 +77,19 @@
 """
 
 CUSTOMIZATIONS = """
+#if CRYPTOGRAPHY_OPENSSL_110_OR_GREATER && !defined(LIBRESSL_VERSION_NUMBER)
+int Cryptography_X509_NAME_ENTRY_set(X509_NAME_ENTRY *ne) {
+    return X509_NAME_ENTRY_set(ne);
+}
+#else
+struct x509_name_entry_st {
```
frasertweedale: I think you're right.  Removed it.

src/cryptography/hazmat/backends/openssl/decode_asn1.py
reaperhulk: Sorry about the delay. I am okay with making a breaking change here if we implemented it incorrectly (which it appears we, or more specifically, I, did). I also agree that should be a separate PR. To make sure I understand this then, you have it all adding to a single set so that it behaves as much like an RDN as possible, correct?

docs/x509/reference.rst
```diff
@@ -1112,6 +1124,12 @@ X.509 CSR (Certificate Signing Request) Builder Object
         <NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.10, name=organizationName)>, value=u'Test Certificates 2011')>
         <NameAttribute(oid=<ObjectIdentifier(oid=2.5.4.3, name=commonName)>, value=u'Good CA')>
 
+    .. attribute:: rdns
```
reaperhulk: Now that `Name` is an iterable of `RelativeDistinguishedName` objects does it make sense to keep this attribute?

src/_cffi_src/openssl/x509name.py
```diff
@@ -17,7 +17,7 @@
 TYPES = """
 typedef ... Cryptography_STACK_OF_X509_NAME_ENTRY;
 typedef ... X509_NAME;
-typedef ... X509_NAME_ENTRY;
+typedef struct X509_name_entry_st X509_NAME_ENTRY;
```
reaperhulk: This can stay opaque.

docs/x509/reference.rst
frasertweedale: Iterating `RelativeDistinguishedName` still yields `NameAttribute` values, for compatibility (and it still adequately covers the 99%** use case where each RDN has exactly one attribute).

So ``rdns`` lets you get the sequence of actual RDNs, if you need them.

** guessing :)

src/_cffi_src/openssl/x509name.py
frasertweedale: typedef opacified :)

docs/x509/reference.rst
```diff
@@ -1112,6 +1124,12 @@ X.509 CSR (Certificate Signing Request) Builder Object
...
+
+        .. versionadded:: 1.6
+
+        :type: :class:`RelativeDistinguishedName`
```
reaperhulk: This returns an iterable of RDNs right?

frasertweedale: Yep; thanks. PR updated.

reaperhulk:
Thanks again for your work on this!

PR #3215, opened 2016-11-06
Fixes #3211 -- fixed hkdf's output with short length

### LINKED ISSUE #3211: HKDF key-length inconsistency
For too small key sizes, `HKDF.derive()` outputs an empty array instead of a small key:

Program:
```python
#!/usr/bin/env python3.5
import cryptography
from cryptography.hazmat.primitives.kdf.hkdf import HKDF

print("cryptography.io:{}".format(cryptography.__version__))

hkdf = HKDF(algorithm=hashes.SHA256(), length=4, salt=b"salt",
            info=b"some-test", backend=default_backend())

key = hkdf.derive(b"my secret passphrase")
print("Derived key: {}".format(key))
```

Output:
```
cryptography.io:1.5.2
Derived key: b''
```

Suggested fix:

I am not quite sure why the division by 8 in the snippet below was added. The cumulative size of the output array is always `self._algorithm.digest_size * len(output)` and thus we can stop after `self._algorithm.digest_size * len(output) >= self._length`. At first I thought this might be a clever trick taken from the paper, but I didn't find it there. I guess there was a mixup between bits and bytes at some point.

```python
# class HKDFExpand
def _expand(self, key_material):
        output = [b""]
        counter = 1

        while (self._algorithm.digest_size // 8) * len(output) < self._length:
            h = hmac.HMAC(key_material, self._algorithm, backend=self._backend)
            h.update(output[-1])
            h.update(self._info)
            h.update(six.int2byte(counter))
            output.append(h.finalize())
            counter += 1

        return b"".join(output)[:self._length]
```

alex:


reaperhulk:
This seems like a thing we should document as a security bug. Silently failing to deliver the requested number of bytes (even if that is a low number) is bad.

alex:
@reaperhulk I'll agree, shall we plan to backport and do an immediate release?

reaperhulk:
Yeah that's my inclination

alex:
@reaperhulk ok, I will _not_ put this in the changelog then, since it'll go in the 1.5.x changelog?

reaperhulk:
👍

PR #3229, opened 2016-11-10
New osrandom_engine in C

tiran:
Inspired by Python/random.c and the old implementation. Let's ask travis if it works on all platforms.

- [x] Test `/dev/urandom` path
- [x] Test ```getrandom()``` path
- [x] Test ```getentropy()``` path
- [x] Test `CryptGenRandom` path


glyph:
On the face of it, this is a terrible idea; ~400 lines of C to replace ~40 lines of Python; 2 open file descriptors for your whole process lifetime instead of 1.

However, knowing the history of this I am sure there's a very good reason for this change. Can you link it to an issue describing said reasons, please? I think it's super important to thoroughly document what is going on here, since it is incredibly confusing; without very clear explication I am concerned that a future maintainer will haplessly reintroduce the bug that this is fixing.

(Secretly, of course, I want to read that description and have a brilliant idea that will salvage the Python implementation and not have to drag around half a thousand lines of duplicative tricky wrapper C code; alas, if I haven't by now, I very much doubt I'll be able to.)

tiran:
Ha, mentionbot betrayed me to @glyph :)

@reaperhulk ask me to look into the matter. My patch is still in a very early phase. Basically I'm just abusing the build and CI testing to test my quick hack on other platforms. I'll take care of documentation later.

reaperhulk:
@glyph: The answer is that we run into endless problems with calling into Python from C on the margins. The most recent of which were apps like Kodi, which embed Python as a scripting option. A Kodi plugin adds cryptography, which registers a callback into Python...but there's only one instance of OpenSSL in the process so now things that aren't even aware of Python are attempting to call back into Python. You could argue that disabling the osrandom engine is the right solution in such a scenario but "right" and "things people will do" are not always overlapping sets.

We also continue to have subinterpreter problems which are probably soluble but I am **so very tired** of trying to fix it when moving it entirely to C will instantly resolve this entire class of issue.

Combine that with our interpreter shutdown issues in newer Pythons and the fact that we need to entirely abandon our Python-based locking callbacks (see: #3226 and linked issues) and there's a strong incentive to move this entirely to C because frankly users will have less problems.

The cost is that we are once again hoisting the complexity CPython formerly provided for us into our own domain and will be responsible for managing that. I don't believe we can continue to say "this isn't a big deal" though. Every time cryptography fails in a weird, undebuggable way when dangerous, unsupported libraries like PyCrypto do not we lose.

Sorry for the semi-rambling response, but I thought you deserved a reasonable dump of my thoughts on why this has become a desirable scenario. The tl;dr is "being reliably functional in all the environments Python is used in trumps the ideological purity of using `os.urandom`"

reaperhulk:
@tiran Thanks for the awesome work here so far. I look forward to reviewing it. 

As an aside: `CCRandomGenerateBytes` might be an avenue we could pursue in a future PR on the mac side (available in 10.10+). `SecRandomCopyBytes` is also available since 10.7 but we'd have to investigate the implementation of each of those to see if they're truly suitable.

tiran:
@reaperhulk The PR is ready for review. I added a ```CCRandomGenerateBytes``` implementation.

tiran:
All Net/Free/Open BSD seem to use the same syscall name ```SYS_getentropy```. I'm using ```#ifdef SYS_getentropy``` to select ```getentropy()``` instead of ```/dev/urandom```.
https://bz-attachments.freebsd.org/attachment.cgi?id=149067

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,396 @@
...
+
+static int osrandom_rand_status(void) {
+    return 1;
+}
+#endif /* CRYPTOGRAPHY_OSRANDOM_ENGINE_GETRANDOM */
+
+/****************************************************************************
+ * BSD getentropy
+ */
+#if CRYPTOGRAPHY_OSRANDOM_ENGINE == CRYPTOGRAPHY_OSRANDOM_ENGINE_GETRANDOM
```
reaperhulk: This should be `CRYPTOGRAPHY_OSRANDOM_ENGINE_GETENTROPY`

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,396 @@
...
+}
+
+static int osrandom_finish(ENGINE *e) {
+    return 1;
+}
+
+static int osrandom_rand_status(void) {
+    return 1;
+}
+#endif /* CRYPTOGRAPHY_OSRANDOM_ENGINE_GETRANDOM */
```
reaperhulk: This comment should be `CRYPTOGRAPHY_OSRANDOM_ENGINE_CC_RANDOM`

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,396 @@
...
```
reaperhulk: `CRYPTOGRAPHY_OSRANDOM_ENGINE_GETENTROPY`

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,396 @@
...
+                && st.st_ino == urandom_cache.st_ino) {
+            fd = urandom_cache.fd;
+            urandom_cache.fd = -1;
+            do {
+                n = close(fd);
+            } while (n < 0 && errno == EINTR);
+        }
+    }
+}
+#endif
```
reaperhulk: Should this have a `CRYPTOGRAPHY_OSRANDOM_NEEDS_DEV_URANDOM` comment?

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,396 @@
...
+
+/****************************************************************************
+ * Linux getrandom engine with fallback to dev_urandom
+ */
+
+#if CRYPTOGRAPHY_OSRANDOM_ENGINE == CRYPTOGRAPHY_OSRANDOM_ENGINE_GETENTROPY
```
reaperhulk: `CRYPTOGRAPHY_OSRANDOM_ENGINE_GETRANDOM`

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,396 @@
...
+    return 1;
+}
+
+static int osrandom_rand_status(void) {
+    if ((getrandom_works != 1) && (urandom_cache.fd < 0)) {
+        return 0;
+    }
+    return 1;
+}
+#endif /* CRYPTOGRAPHY_OSRANDOM_ENGINE_GETENTROPY */
```
reaperhulk: `CRYPTOGRAPHY_OSRANDOM_ENGINE_GETRANDOM`

src/_cffi_src/openssl/src/osrandom_engine.h
```diff
@@ -0,0 +1,71 @@
+#ifdef _WIN32
+  #include <Wincrypt.h>
+#else
+  #include <fcntl.h>
+  #include <unistd.h>
+   /* for defined(BSD) */
+  #include <sys/param.h>
+
+  #ifdef BSD
```
reaperhulk: Does FreeBSD define this? I think they typically use `__FreeBSD__`. What about macOS (which apparently has `SYS_getentropy` as of 10.12)?

src/_cffi_src/openssl/src/osrandom_engine.h
```diff
@@ -0,0 +1,71 @@
...
+ * TODO: check NetBSD and Darwin */
+#if defined(_WIN32)
+  /* Windows */
+  #define CRYPTOGRAPHY_OSRANDOM_ENGINE CRYPTOGRAPHY_OSRANDOM_ENGINE_CRYPTGENRANDOM
+#elif defined(__APPLE__) && defined(CRYPTOGRAPHY_HAVE_COMMON_RANDOM_H)
+  /* OSX 10.10+ */
+  #define CRYPTOGRAPHY_OSRANDOM_ENGINE CRYPTOGRAPHY_OSRANDOM_ENGINE_CC_RANDOM
+#elif defined(BSD) && defined(SYS_getentropy)
+  /* OpenBSD 5.6+ */
+  #define CRYPTOGRAPHY_OSRANDOM_ENGINE CRYPTOGRAPHY_OSRANDOM_ENGINE_GETRANDOM
```
reaperhulk: This should be `CRYPTOGRAPHY_OSRANDOM_ENGINE_GETENTROPY` with the naming fixes requested above

src/_cffi_src/openssl/src/osrandom_engine.h
```diff
@@ -0,0 +1,71 @@
...
+#elif defined(__linux__) && defined(SYS_getrandom)
+  /* Linux 3.4.17+ */
+  #define CRYPTOGRAPHY_OSRANDOM_ENGINE CRYPTOGRAPHY_OSRANDOM_ENGINE_GETENTROPY
```
reaperhulk: And this should be `CRYPTOGRAPHY_OSRANDOM_ENGINE_GETRANDOM`

src/_cffi_src/openssl/src/osrandom_engine.h
```diff
@@ -0,0 +1,71 @@
...
+  #define CRYPTOGRAPHY_OSRANDOM_NEEDS_DEV_URANDOM 1
+#endif
+
+/* Fall back to /dev/urandom */
+#ifndef CRYPTOGRAPHY_OSRANDOM_ENGINE
+  #define CRYPTOGRAPHY_OSRANDOM_ENGINE CRYPTOGRAPHY_OSRANDOM_ENGINE_DEV_URANDOM
+  #define CRYPTOGRAPHY_OSRANDOM_NEEDS_DEV_URANDOM 1
+#endif
+
+/* Open SSL 1.1.0+ has no ERR_R_RAND_LIB */
```
reaperhulk: What do you think about always passing 0 and not worrying about a macro here?

reaperhulk:
Coverage is failing because `_osrandom_rand_bytes` and `_osrandom_rand_status` need to be removed from `src/cryptography/hazmat/bindings/openssl/binding.py`

src/_cffi_src/openssl/src/osrandom_engine.c
tiran: Thx, I messed up some find and replace operations. The error was consistent over all files so it just worked.

src/_cffi_src/openssl/src/osrandom_engine.h
tiran: Yes, it's defined by Free, Open, Net and Dragon BSD in ```sys/param.h```

http://www.boost.org/doc/libs/1_60_0/libs/predef/doc/html/predef/reference/boost_os_operating_system_macros.html
http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system#BSD

tests/hazmat/backends/test_openssl.py
```diff
@@ -282,6 +269,14 @@ def test_activate_builtin_random_already_active(self):
         e = backend._lib.ENGINE_get_default_RAND()
         assert e == backend._ffi.NULL
 
+    def test_osrandom_engine_implementation(self):
+        name = backend.osrandom_engine_implementation()
+        assert name
+        if sys.platform.startswith('linux'):
+            assert name in ['getrandom', '/dev/urandom']
+        elif sys.platform == 'win32':
```
reaperhulk: Our CI can't cover this branch (we don't get coverage from our windows builders, although this is being executed) so we'll need to figure out a way to rewrite this. Maybe make the asserts a separate function that takes a string `_engine_name_assert(sys.platform)` and then

```python
def _engine_name_assert(platform):
    if platform.startswith('linux'):
        assert name in ['getrandom', '/dev/urandom']
   elif platform == 'win32':
        assert name == 'CryptGenRandom'
```

and then a separate test for `_engine_name_assert` that explicitly passes `win32`?

alex:
Failing docs tests are relevant; looks like `fallback` needs to be added to the wordlist :-)

tiran:
I added ``fallback`` to spelling wordlist. On my machine the doc tests are failing with a different error. My spell checker knows about fallback but does not recognize personalization:
```
hazmat/primitives/cryptographic-hashes.rst:129: (personalization)
```

reaperhulk:
https://codecov.io/gh/pyca/cryptography/compare/3a15b03e92c9fdeadff04ddd2ce505028b279b86...075de30232f1fd6f1e0d541844e01eab93a006b3#74657374732F68617A6D61742F6261636B656E64732F746573745F6F70656E73736C2E7079-280...281 is our coverage problem that's causing the status check failures.

We do have FreeBSD 10 (but not 11) in our jenkins cluster, but no OpenBSD and we don't get coverage from those builders right now. 😢

In an ideal world what would CI look like here?

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,461 @@
...
+        if (flags == -1) {
+            goto error;
+        } else if (fcntl(fd, F_SETFD, flags | FD_CLOEXEC) == -1) {
+            goto error;
+        }
+        /* Another thread initialized the fd */
+        if (urandom_cache.fd >= 0) {
+            do {
+                n = close(fd);
+            } while (n < 0 && errno == EINTR);
```
haypo: It would be kind to call PyErr_CheckSignals() to fail fast on CTRL+C: PEP 475.

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,461 @@
...
+             * close the fd. */
+            urandom_cache.fd = -1;
+        }
+    }
+    if (urandom_cache.fd < 0) {
+        fd = open("/dev/urandom", O_RDONLY);
+        if (fd < 0) {
+            goto error;
+        }
+        if (fstat(fd, &st)) {
```
haypo: You may call fstat() after setting CLOEXEC.

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,461 @@
...
+
+    fd = dev_urandom_fd();
+    if (fd < 0) {
+        return 0;
+    }
+
+    while (size > 0) {
+        do {
+            n = read(fd, buffer, (size_t)size);
+        } while (n < 0 && errno == EINTR);
```
haypo: Ditto: call PyErr_CheckSignals() on EINTR.

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,461 @@
...
+    int fd, n;
+
+    fd = dev_urandom_fd();
+    if (fd < 0) {
+        return 0;
+    }
+
+    while (size > 0) {
+        do {
+            n = read(fd, buffer, (size_t)size);
```
haypo: On UNIX, read() returns a ssize_t.

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,461 @@
...
+}
+
+static int dev_urandom_read(unsigned char *buffer, int size) {
+    int fd, n;
+
+    fd = dev_urandom_fd();
+    if (fd < 0) {
+        return 0;
+    }
+
```
haypo: You might check that size >= 0.

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,461 @@
...
+            return 0;
+        }
+        buffer += n;
+        size -= n;
+    }
+    return 1;
+}
+
+static void dev_urandom_close(void) {
+    if (urandom_cache.fd >= 0) {
```
haypo: Coding style: I suggest to return early using return if a test fail, to reduce the indentation level.

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,461 @@
...
+        struct stat st;
+
+        if (fstat(urandom_cache.fd, &st)
+                && st.st_dev == urandom_cache.st_dev
+                && st.st_ino == urandom_cache.st_ino) {
+            fd = urandom_cache.fd;
+            urandom_cache.fd = -1;
+            do {
+                n = close(fd);
+            } while (n < 0 && errno == EINTR);
```
haypo: If you decide to call PyErr_CheckSignals(), you may put this code into a subfunction.

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,461 @@
...
+        char dest[1];
+        n = syscall(SYS_getrandom, dest, sizeof(dest), GRND_NONBLOCK);
+        if (n < 0) {
+            /* Can fail with:
+             * - ENOSYS: Kernel does not support the syscall.
+             * - ENOPERM: seccomp prevents syscall.
+             * - EAGAIN: Kernel CRPNG has not been seeded yet.
+             * EINTR cannot occur for buflen < 256.
+             */
+            getrandom_works = 0;
```
haypo: I'm curious about EINTR: add assert(errno != EINTR); if you are confident :-)

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,461 @@
...
+        int fd = dev_urandom_fd();
+        if (fd < 0) {
+            return 0;
+        }
+    }
+    return 1;
+}
+
+static int osrandom_rand_bytes(unsigned char *buffer, int size) {
+    if (getrandom_works == 1) {
```
haypo: Coding style: I suggest to put dev_urandom_read() at the beginning with a return to avoid a level of indentation.

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,461 @@
...
+        if ((p == NULL) && (i == 0)) {
+            /* return required buffer len */
+            return len;
+        }
+        if ((p == NULL) || i < 0 || ((size_t)i <= len)) {
+            /* no buffer or buffer too small */
+            ENGINEerr(ENGINE_F_ENGINE_CTRL, ENGINE_R_INVALID_ARGUMENT);
+            return 0;
+        }
+        strncpy((char *)p, name, len);
```
haypo: memcpy() can be used here.

src/_cffi_src/openssl/src/osrandom_engine.h
```diff
@@ -0,0 +1,70 @@
...
+    /* Windows */
+    #define CRYPTOGRAPHY_OSRANDOM_ENGINE CRYPTOGRAPHY_OSRANDOM_ENGINE_CRYPTGENRANDOM
+  #elif defined(__APPLE__) && defined(CRYPTOGRAPHY_HAVE_COMMON_RANDOM_H)
+    /* OSX 10.10+ */
+    #define CRYPTOGRAPHY_OSRANDOM_ENGINE CRYPTOGRAPHY_OSRANDOM_ENGINE_CC_RANDOM
+  #elif defined(BSD) && defined(SYS_getentropy)
+    /* OpenBSD 5.6+ */
+    #define CRYPTOGRAPHY_OSRANDOM_ENGINE CRYPTOGRAPHY_OSRANDOM_ENGINE_GETENTROPY
+  #elif defined(__linux__) && defined(SYS_getrandom)
+    /* Linux 3.4.17+ */
```
haypo: 3.17+ no?

src/_cffi_src/openssl/src/osrandom_engine.h
```diff
@@ -0,0 +1,70 @@
...
+    #define CRYPTOGRAPHY_OSRANDOM_ENGINE CRYPTOGRAPHY_OSRANDOM_ENGINE_GETRANDOM
+  #else
```
haypo: You may also use getrandom() on Solaris 11.3+.

src/_cffi_src/openssl/src/osrandom_engine.c
tiran: The engine does not not use any Python API at all. I don't consider SIGINT a problem here. The engine is a low level part of OpenSSL's CPRNG. The CPRNG API should take care of of interrupts.

src/_cffi_src/openssl/src/osrandom_engine.c
tiran: fixed

tiran: fixed

src/_cffi_src/openssl/src/osrandom_engine.c
tiran: man getrandom promises that it works w/o EINTR for small values up to 256 bytes. It was one of the reasons I dislike Raymonds os.urandom(2500) call in random.py.

src/_cffi_src/openssl/src/osrandom_engine.h
tiran: 3.4.17 is right

src/_cffi_src/openssl/src/osrandom_engine.c
tiran: Why? It's a string.

src/_cffi_src/openssl/src/osrandom_engine.h
tiran: We don't have any test machines for Solaris. /dev/urandom is fine on all POSIX machines.

tests/hazmat/backends/test_openssl.py
```diff
@@ -282,6 +269,20 @@ def test_activate_builtin_random_already_active(self):
...
+        assert name in ['/dev/urandom', 'CCRandomGenerateBytes',
+                        'CryptGenRandom', 'getentropy', 'getrandom']
+        if sys.platform.startswith('linux'):
+            assert name in ['getrandom', '/dev/urandom']
+        if sys.platform == 'darwin':
+            assert name in ['CCRandomGenerateBytes', '/dev/urandom']
+        # TODO: enable when CI can handle coverage.py for these platforms
+        # if 'bsd' in sys.platform:
+        #     assert name in ['getentropy', '/dev/urandom']
+        # if sys.platform == 'win32':
```
reaperhulk: This branch can be uncommented now (I updated our jenkins configs so we now merge coverage across both jenkins and travis)

reaperhulk:
@tiran What do you think of doing this: https://github.com/pyca/cryptography/pull/3269/commits/d59d71b3373266200897d75b438935e0328826ad

This would simplify the code a bit and also let us cover all branches in our current CI. I was able to find/fix two syntax errors in the getentropy path with this change.

If it makes sense to you feel free to cherry-pick it into this PR (be sure you add `macOS` to the spelling wordlist too).

reaperhulk:
Jenkins, retest this please

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,430 @@
...
+ */
+#if CRYPTOGRAPHY_OSRANDOM_ENGINE == CRYPTOGRAPHY_OSRANDOM_ENGINE_GETENTROPY
+static const char *Cryptography_osrandom_engine_name = "osrandom_engine getentropy()";
+
+static int osrandom_init(ENGINE *e) {
+    return 1;
+}
+
+static int osrandom_rand_bytes(unsigned char *buffer, int size) {
+    Py_ssize_t len;
```
reaperhulk: Should probably be `ssize_t`?

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,430 @@
...
+    int res;
+    while (size > 0) {
+        /* OpenBSD restricts maximum buffer size to 256. */
```
reaperhulk: `macOS` does as well so this comment should be updated to mention both.

alex:
(Moved this to the 17th release, which we hope to get done in 2-3 weeks)

src/_cffi_src/openssl/src/osrandom_engine.c
tiran: getentropy() returns an int, http://man.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man2/getentropy.2

reaperhulk: Okay, so `int` instead of `Py_ssize_t` 😁

docs/hazmat/backends/openssl.rst
```diff
@@ -40,6 +40,10 @@ greater.
         Activates the OS random engine. This will effectively disable OpenSSL's
         default CSPRNG.
 
+    .. method:: osrandom_engine_implementation()
```
reaperhulk: .. versionadded:: 1.7 please

tiran: fixed

LICENSE
```diff
@@ -2,5 +2,6 @@ This software is made available under the terms of *either* of the licenses
 found in LICENSE.APACHE or LICENSE.BSD. Contributions to cryptography are made
 under the terms of *both* these licenses.
 
-The code used in the OpenSSL locking callback is derived from the same in
-Python itself, and is licensed under the terms of the PSF License Agreement.
+The code used in the OpenSSL locking callback and OS random engine is derived
+from the same in Python itself, and is licensed under the terms of the PSF
```
alex: Ugh, nit, not technically a bug in your PR (but if I go fix it separately it'll just cause a conflict). Can you `s/Python/CPython/`

docs/hazmat/backends/openssl.rst
```diff
@@ -81,6 +87,21 @@ details.
 Linux uses its own PRNG design. ``/dev/urandom`` is a non-blocking source
 seeded from the same pool as ``/dev/random``.
 
++------------------------------------------+------------------------------+
+| Windows                                  | ``CryptGenRandom()``         |
++------------------------------------------+------------------------------+
+| Linux >= 3.4.17 with working             | ``getrandom(GRND_NONBLOCK)`` |
+| ``SYS_getrandom`` syscall                |                              |
```
alex: You can drop the leading `SYS` here.

src/_cffi_src/openssl/src/osrandom_engine.h
```diff
@@ -0,0 +1,58 @@
...
+    #include <sys/random.h>
+  #endif
+
+  #ifdef __linux__
+    /* for SYS_getrandom */
+    #include <sys/syscall.h>
+    #ifndef GRND_NONBLOCK
+      #define GRND_NONBLOCK 0x0001
+    #endif /* GRND_NONBLOCK */
+  #endif /* _linux__ */
```
alex: Missing leading `_` here

src/_cffi_src/openssl/src/osrandom_engine.h
```diff
@@ -0,0 +1,58 @@
...
+  #endif
+
+  #ifdef __APPLE__
+    #include <sys/random.h>
+  #endif
+
+  #ifdef __linux__
+    /* for SYS_getrandom */
+    #include <sys/syscall.h>
+    #ifndef GRND_NONBLOCK
```
alex: In what circumstances it this not defined?

tests/hazmat/backends/test_openssl.py
```diff
@@ -282,6 +271,23 @@ def test_activate_builtin_random_already_active(self):
         e = backend._lib.ENGINE_get_default_RAND()
         assert e == backend._ffi.NULL
 
+    def test_osrandom_engine_implementation(self):
+        name = backend.osrandom_engine_implementation()
+        assert name in ['/dev/urandom', 'CCRandomGenerateBytes',
```
alex: `CCRandomGenerateBytes` is not used anywhere.

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,429 @@
...
+    if (CryptReleaseContext(hCryptProv, 0)) {
+        hCryptProv = 0;
+        return 1;
+    } else {
+        return 0;
+    }
+}
+
+static int osrandom_rand_status(void) {
+    if (hCryptProv == 0) {
```
alex: This simplified to `return hCryptoProv != 0;`

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,429 @@
...
+
+static int osrandom_init(ENGINE *e) {
+    return 1;
+}
+
+static int osrandom_rand_bytes(unsigned char *buffer, int size) {
+    int len, res;
+    while (size > 0) {
+        /* OpenBSD and macOS restrict maximum buffer size to 256. */
+        len = size > 256 ? 256: size;
```
alex: Space beore teh `:` please

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,429 @@
...
+
+static int osrandom_rand_status(void) {
+    if (hCryptProv == 0) {
+        return 0;
+    } else {
+        return 1;
+    }
+}
+
+static const char* osurandom_get_implementation(void) {
```
alex: `*` next to the function name, rather than the type please.

src/_cffi_src/openssl/src/osrandom_engine.c
alex: Is closign an `fd` in a loop safe? I seem to recall it being dangerous :-/

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,429 @@
...
+    /* We try to detect working getrandom until we succeed. */
+    if (getrandom_works != 1) {
+        long n;
+        char dest[1];
+        n = syscall(SYS_getrandom, dest, sizeof(dest), GRND_NONBLOCK);
+        if (n < 0) {
+            /* Can fail with:
+             * - ENOSYS: Kernel does not support the syscall.
+             * - ENOPERM: seccomp prevents syscall.
+             * - EAGAIN: Kernel CRPNG has not been seeded yet.
```
alex: In the even of `EAGAIN` I'm pretty sure we want to still use `getrandom`, not fall back.

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,429 @@
...
+    }
+    return 1;
+}
+
+static int osrandom_rand_bytes(unsigned char *buffer, int size) {
+    if (getrandom_works == 1) {
+        long n;
+        while (size > 0) {
+            do {
+                n = syscall(SYS_getrandom, buffer, size, GRND_NONBLOCK);
```
alex: What does this do on large (>256) `size`?

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -0,0 +1,429 @@
...
+    return dev_urandom_read(buffer, size);
+}
+
+static int osrandom_finish(ENGINE *e) {
+    dev_urandom_close();
+    return 1;
+}
+
+static int osrandom_rand_status(void) {
+    if (urandom_cache.fd < 0) {
```
alex: Simplifies to `return urandom_cache.fd > 0`

docs/hazmat/backends/openssl.rst
tiran: Unless you really insist, I'd rather keep ```SYS_getrandom``` because that is what the random engine actually uses. It's going to make it easier to distinguish between platforms with the syscall and platforms with a glibc wrapper later.

src/_cffi_src/openssl/src/osrandom_engine.h
tiran: The constant is defined in user land headers. Sometimes the running Kernel has the syscall but the userland hasn't caught up yet.

src/_cffi_src/openssl/src/osrandom_engine.c
tiran: Let's discuss this matter on IRC later. I don't want to repeat the /dev/urandom discussion from python-dev on github. :)

src/_cffi_src/openssl/src/osrandom_engine.c
tiran: http://man7.org/linux/man-pages/man2/getrandom.2.html

>    If the
>       /dev/urandom pool has been initialized, reads of up to 256 bytes will
>       always return as many bytes as requested and will not be interrupted
>       by signals.  No such guarantees apply for larger buffer sizes.  For
>       example, if the call is interrupted by a signal handler, it may
>       return a partially filled buffer, or fail with the error EINTR.

src/_cffi_src/openssl/src/osrandom_engine.c
tiran: ```>= 0``` or ``` != -1```. Theoretically the fd can by 0,

src/_cffi_src/openssl/src/osrandom_engine.c
alex: Ok, you know how to ping me.

alex:
The `EAGAIN` issue is the last remaining issue for me.

src/_cffi_src/openssl/src/osrandom_engine.c
socketpair: 1. Please give link to that conversation.
2. It would be nice to place explanation in documentation.

Issue #3197, opened 2016-10-15
Support for UniqueIdentifier (2.5.4.3) in Subject


jmcfcosta:
Hi,
How could I add an UniqueIdentifier field to a subject? Correct me if I'm wrong (I've only started using your lib recently) but it doesn't seem to be supported - trying to fetch the subject in a certificate having it raises a "ValueError: Unsupported ASN1 string type. Type: 3".
NameAttributes only accept text values but in this case a unique ID will be encoded as a bitstring.
Could support for this be added in the future or at least someone give me directions for a (temporary) workaround?

Thanks.

reaperhulk:
@jmcfcosta could you give us a sample certificate that contains a field like this?

castillar:
I'm having a similar problem with trying to parse a certificate containing such an identifier and getting the same "ValueError: Unsupported ASN1 string type. Type: 3" error when asking for either subject or issuer (it's a self-signed root, so those should be the same value). Since the certificate is part of several public trust stores, I've attached it here for reference, in PEM format.

PyASN1 decodes the subject as 
"Name(componentType=NamedTypes(NamedType('', RDNSequence()))).setComponents(RDNSequence().setComponents(RelativeDistinguishedName().setComponents(AttributeTypeAndValue(componentType=NamedTypes(NamedType('type', AttributeType('<no value>')), NamedType('value', AttributeValue()))).setComponents(AttributeType('2.5.4.10'), AttributeValue(hexValue='1305436973636f'))), RelativeDistinguishedName().setComponents(AttributeTypeAndValue(componentType=NamedTypes(NamedType('type', AttributeType('<no value>')), NamedType('value', AttributeValue()))).setComponents(AttributeType('2.5.4.3'), AttributeValue(hexValue='131041435432204543432053554449204341')))))"
so it's decodable, and all of the other attributes (public_bytes, not_valid_before, not_valid_after), etc. are fine. It's just Issuer and Subject that seem to give it heartburn.

[7e206939cc5fa883635f64c750ebf5fda9aee653.txt](https://github.com/pyca/cryptography/files/558833/7e206939cc5fa883635f64c750ebf5fda9aee653.txt)

reaperhulk:
[...]

PR #3233, opened 2016-11-11
add some new oids

reaperhulk:
Two more X.509 Name OIDs we need to care about (uniqueIdentifier has some additional requirements for parsing since it is a BITSTRING. see #3197)


src/cryptography/x509/oid.py
```diff
@@ -99,13 +99,15 @@ class NameOID(object):
     COUNTRY_NAME = ObjectIdentifier("2.5.4.6")
     LOCALITY_NAME = ObjectIdentifier("2.5.4.7")
     STATE_OR_PROVINCE_NAME = ObjectIdentifier("2.5.4.8")
+    STREET = ObjectIdentifier("2.5.4.9")
```
reaperhulk: Yep, should be that (and `streetAddress` in the name mapping). Updated

PR #3235, opened 2016-11-12
workaround for application bundling tools

reaperhulk:
This PR is based on #3081 and restructures it a little bit for test coverage.

### DIFF
--- a/src/cryptography/hazmat/backends/__init__.py
+++ b/src/cryptography/hazmat/backends/__init__.py
@@ -5,26 +5,11 @@
 from __future__ import absolute_import, division, print_function
 
 import sys
+
 import pkg_resources
 
 from cryptography.hazmat.backends.multibackend import MultiBackend
 
-
-def _build_frozen_backend_list():
-    # if found frozen then manually try to import known packaged backends
-    try:
-        from cryptography.hazmat.backends.commoncrypto.backend import backend as be_cc
-    except ImportError:
-        be_cc = None
-
-    try:
-        from cryptography.hazmat.backends.openssl.backend import backend as be_ossl
-    except ImportError:
-        be_ossl = None
-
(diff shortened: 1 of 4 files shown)


alex:
I didn't totally follow the original discussion, but I thought we only planned to do this if `sys.frozen`?

reaperhulk:
That was the original plan but while refactoring this it seemed like it was kinda pointless to check. If iter_entry_points fails for any reason we might as well fall back to the explicit enumeration. Plus checking `sys.frozen` makes a branch that's hard to cover :)

src/cryptography/hazmat/backends/__init__.py
```diff
@@ -16,7 +16,7 @@ def _available_backends():
     global _available_backends_list
 
     if _available_backends_list is None:
-        _available_backends_list = [
+        _entry_point_backends = [
```
alex: This shouldn't have an underscore, `_available_backends_list` does because it's the global name.

src/cryptography/hazmat/backends/__init__.py
```diff
@@ -29,8 +29,39 @@ def _available_backends():
...
+
+    try:
+        # CommonCrypto will only be present on macOS so failure is possible.
+        from cryptography.hazmat.backends.commoncrypto.backend import (
+            backend as be_cc
+        )
+    except ImportError:
+        be_cc = None
+
+    _found_backends = [be for be in (be_cc, be_ossl) if be is not None]
```
alex: I think this code can be restructured to be:

```py
from cryptography.hazmat.backends.openssl.backend import backend as be_ossl
backends = [be_ossl]
try:
    from cryptography.hazmat.backends.commoncrypto.backend import (
        backend as be_cc
    )
except ImportError:
    pass
else:
    backend.append(be_cc)

return backends
```

PR #3238, opened 2016-11-12
support prehashing in RSA sign

reaperhulk:
Support supplying previously hashed data to the `RSAPrivateKey.sign` method. Support for RSA verification and sign/verify for DSA/ECDSA will come in future PRs.

This addresses part of #1648

### DIFF
--- a/docs/hazmat/primitives/asymmetric/rsa.rst
+++ b/docs/hazmat/primitives/asymmetric/rsa.rst
@@ -167,6 +167,25 @@ There is a shortcut to sign sufficiently short messages directly:
     ...     hashes.SHA256()
     ... )
 
+
+If you have pre-hashed your data you can sign using that as well:
+
+.. versionadded:: 1.6
+
+.. doctest::
+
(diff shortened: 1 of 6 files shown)


alex:
It feels like there was a substantial refactoring here, any chance of splitting that out?

alex:
Ready for a rebase now.

alex:
Needs a test for what happens if `len(data) != algorith.digest_size`.

docs/hazmat/primitives/asymmetric/rsa.rst
```diff
@@ -167,6 +167,25 @@ There is a shortcut to sign sufficiently short messages directly:
     ...     hashes.SHA256()
     ... )
 
+
+If you have pre-hashed your data you can sign using that as well:
```
alex: I feel like this shouldn't be a part of the primary narrative, since it's relatively obscure. What do you think?

alex:
Needs a changelog entry

PR #3241, opened 2016-11-14
decrypt method returns bytes in example

2tunnels:
Just to make it clear.


reaperhulk:
Unfortunately due to the limitations of doctest this won't work in python 2.7. Are you aware of a way to fix that?

reaperhulk:
Closing for now since this is the way doctest works in py2.7. Feel free to reopen if there's some better way to do it though!

PR #3243, opened 2016-11-15
change derive_elliptic_curve_public_point to return EllipticCurvePubl…

reaperhulk:
…icKey and rename the interface method.

Per post-merge discussion on #3225.

### DIFF
--- a/docs/hazmat/backends/interfaces.rst
+++ b/docs/hazmat/backends/interfaces.rst
@@ -422,7 +422,7 @@ A specific ``backend`` may provide one or more of these interfaces.
         :returns: An instance of
             :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey`.
 
-    .. method:: derive_elliptic_curve_public_key(private_value, curve)
+    .. method:: derive_elliptic_curve_private_key(private_value, curve)
 
         :param private_value: A secret scalar value.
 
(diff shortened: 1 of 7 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1418,7 +1418,13 @@ def derive_elliptic_curve_public_point(self, private_value, curve):
...
             point_y = self._bn_to_int(bn_y)
 
-        return point_x, point_y
+        return self.load_elliptic_curve_public_numbers(
+            ec.EllipticCurvePublicNumbers(
+                point_x,
+                point_y,
+                curve
+            )
+        )
```
alex: Can the point ptr not be turned directly into an `EC_KEY`?

docs/hazmat/backends/interfaces.rst
```diff
@@ -422,14 +422,15 @@ A specific ``backend`` may provide one or more of these interfaces.
         :returns: An instance of
             :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey`.
 
-    .. method:: derive_elliptic_curve_public_point(private_value, curve)
+    .. method:: derive_elliptic_curve_public_key(private_value, curve)
```
alex: I realize the value you're computing is the public  key, but since you have the private value, is there a reason not to return a private key instance?

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: Good question. I'll look into it.

PR #3244, opened 2016-11-15
Add short_name and long_name to OIDs

tiran:
Fraser added mappings from OID to short names to FreeIPA, https://github.com/freeipa/freeipa/blob/master/ipapython/dn.py#L1423 . I'd prefer to have the information cryptography itself. The PR adds short_name and long_name attributes to ```ObjectIdentifier```.

@frasertweedale, I used ```_ssl.txt2obj()``` to translate ```OID``` to ```SN```. OpenSSL has different short names than FreeIPA, e.g. email and the jurisdiction OIDs.

### DIFF
--- a/src/cryptography/x509/oid.py
+++ b/src/cryptography/x509/oid.py
@@ -61,7 +61,15 @@ class ObjectIdentifier(object):
 
     @property
     def _name(self):
-        return _OID_NAMES.get(self, "Unknown OID")
+        return _OID_NAMES.get(self, ("Unknown OID", None))[0]
+
+    @property
+    def longname(self):
+        return _OID_NAMES.get(self, (None, None))[0]
+
+    @property
+    def shortname(self):
+        return _OID_NAMES.get(self, (None, None))[1]
 
     dotted_string = utils.read_only_property("_dotted_string")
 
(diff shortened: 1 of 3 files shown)


src/cryptography/x509/oid.py
```diff
@@ -61,7 +61,15 @@ def __hash__(self):
 
     @property
     def _name(self):
-        return _OID_NAMES.get(self, "Unknown OID")
+        return _OID_NAMES.get(self, ("Unknown OID", None))[0]
+
+    @property
+    def longname(self):
```
bjmc: When I suggested adding `name`, the maintainers wanted it to be `_name` so as not to commit to supporting these convenience properties in the long term. I suspect they will want the same here.

src/cryptography/x509/oid.py
```diff
@@ -61,7 +61,15 @@ def __hash__(self):
...
+        return _OID_NAMES.get(self, (None, None))[0]
+
+    @property
+    def shortname(self):
+        return _OID_NAMES.get(self, (None, None))[1]
```
bjmc: I think this reads a bit ugly, it's not immediately obvious to a casual reader what `[0]` and `[1]` are getting here. What about making all of these (shortname, longname) tuples be [`namedtuples`](https://docs.python.org/dev/library/collections.html#collections.namedtuple) instead? 

Namedtuples are lightweight, requiring no extra memory, and then you could have `.get(...).short` and `.get(...).long` or similar.

reaperhulk:
We've resisted having these be part of the official public API in the past because OID names aren't entirely standardized and we didn't want to leak OpenSSL-isms into cryptography.

frasertweedale:
:+1: to the addition of `USER_ID`, `POSTAL_ADDRESS` and `POSTAL_CODE` (which IMO should be a separate commit).

To the `shortname` proposal, I don't have strong feels.

reaperhulk:
As @frasertweedale mentioned, I'd like to see those 3 new OIDs added in a separate PR.

As for short_name/long_name, I'm negative on making this public API. The lack of standardization in these names frustrates me.

alex:
I'm also opposed to exposing this API, specific names for OIDs are basically all not-standardized. Going to close this since @reaperhulk and I are in agreement :-)

PR #3248, opened 2016-11-17
extend the cryptography cffi backend for pypy's brand new stdlib _ssl module

planrich:
Here are the changes I needed to apply to implement PyPy's new stdlib _ssl module. I hope these are useful and will be merged.

If some cffi methods are not added correctly (missing ifdef for version check, ...) just point it out, I'm happy to fix it.

PR #3249, opened 2016-11-19
add some BIO functions for pypy's ssl stdlib

reaperhulk:
refs #3248


reaperhulk:
For reference, https://github.com/openssl/openssl/commit/fb46be034816e5fe9f04fd39da960d34dbf2f52d is where `BIO_up_ref` was added

src/_cffi_src/openssl/bio.py
```diff
@@ -130,7 +133,15 @@
...
 long BIO_set_buffer_read_data(BIO *, void *, long);
 long BIO_set_nbio(BIO *, long);
+void BIO_set_retry_read(BIO *);
+void BIO_clear_retry_flags(BIO *);
 """
 
 CUSTOMIZATIONS = """
+#if CRYPTOGRAPHY_OPENSSL_LESS_THAN_110PRE4 || defined(LIBRESSL_VERSION_NUMBER)
+int BIO_up_ref(BIO *b) {
+    CRYPTO_add(&b->references, 1, CRYPTO_LOCK_BIO);
```
alex: This returns something, probably we should check the return value.

PR #3259, opened 2016-11-19
add a few more OIDs

reaperhulk:
pulled from #3244


tiran:
Thanks @reaperhulk . I haven't had time to split up #3244

PR #3270, opened 2016-11-21
PyPy's ssl module, the last missing macro and three macro functions

planrich:
Here are the last missing macro/functions needed. I'm not sure why `sk_GENERAL_NAME_freefunc` has the type with `struct GENERAL_NAME_st*` instead of `GENERAL_NAME*`. Without that change I could not pass `GENERAL_NAME_free` to an argument requiring a `sk_GENERAL_NAME_freefunc` type.


alex:
Jenkins, ok to test.

src/_cffi_src/openssl/ssl.py
```diff
@@ -688,6 +689,12 @@
 
 static const long Cryptography_HAS_SSL_CTX_CLEAR_OPTIONS = 1;
 
+#ifdef OPENSSL_NO_TLSEXT
```
reaperhulk: cryptography doesn't support OpenSSL compiled without TLS extensions so you can always assume this is not set. Do you still need it then?

src/_cffi_src/openssl/x509_vfy.py
```diff
@@ -200,6 +200,9 @@
                               size_t);
 int X509_VERIFY_PARAM_set1_ip_asc(X509_VERIFY_PARAM *, const char *);
 
+int sk_X509_OBJECT_num(Cryptography_STACK_OF_X509_OBJECT *);
+X509_OBJECT *sk_X509_OBJECT_value(Cryptography_STACK_OF_X509_OBJECT *, int);
+X509_VERIFY_PARAM * X509_STORE_get0_param(X509_STORE *);
```
reaperhulk: Style nit, this should be `X509_VERIFY_PARAM *X509_STORE_get0_param(X509_STORE *);` (space removed before function name)

src/_cffi_src/openssl/ssl.py
planrich: We use e.g. SSL_CTX_set_tlsext_servername_callback, I'm not sure if that is to be called without the macro OPENSSL_NO_TLSEXT defined.

planrich: Rethinking that, I'm fine with removing it... But this means PyPy cannot support any openssl version that would set the macro?

reaperhulk: Correct. By using cryptography as the basis of the module you're bound by the versions we've chosen to support, but OPENSSL_NO_TLSEXT is extremely uncommon and in practice we've never had a single bug report complaining about not supporting that configuration.

PR #3283, opened 2016-11-24
add support for encoding points to...

ofek:
the standard compressed format as described in [SEC 1 v2.0](http://www.secg.org/sec1-v2.pdf) section 2.3.3.


reaperhulk:
Your commit history is wildly divergent from master here. Would you mind fixing that? It looks like there's only one commit that actually matters here. If you're not familiar with git rebases I'd suggest just checking out a new copy of master and re-applying your changes.

alex:
Jenkins, ok to test

ofek:
@alex @reaperhulk r4r

PR #3286, opened 2016-11-25
Add binding to DTLSv1's timeout handling functions

Gawen:
2 patches to:

- ~~add the bindings of `DTLSv1_2_method()`, `DTLSv1_2_server_method()` and `DTLSv1_2_client_method()`.~~
- add the binding of `DTLSv1_get_timeout()` and `DTLSv1_handle_timeout()`.

Note that `DTLSv1_get_timeout()` is wrapped via `DTLSv1_get_timeout_wrapped()` to avoid `cffi` to manipulate a `struct timeval`. Pointers to the variables `tv_sec` and `tv_usec` are directly passed via the function's arguments.

Please let me know if you have any feedback which would help for this PR to be merged.

Thanks.

EDIT: as seen in https://github.com/pyca/cryptography/pull/3286#issuecomment-263015946 , this PR is now only about the DTLSv1.x timeout handling functions.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -730,8 +730,13 @@ long DTLSv1_get_timeout_wrapped(
     int r = DTLSv1_get_timeout(ssl, &tv);
 
     if (r == 1) {
-        if (ptv_sec) *ptv_sec = tv.tv_sec;
-        if (ptv_usec) *ptv_usec = tv.tv_usec;
+        if (ptv_sec) {
+            *ptv_sec = tv.tv_sec;
+        }
+
+        if (ptv_usec) {
+            *ptv_usec = tv.tv_usec;
+        }
     }
 
     return r;
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -725,7 +725,11 @@ static const long Cryptography_HAS_NPN_NEGOTIATED = 0;
 #endif
 
 /* Wrap DTLSv1_get_timeout to avoid cffi to handle a 'struct timeval'. */
-long DTLSv1_get_timeout_wrapped(SSL *ssl, time_t *ptv_sec, long int *ptv_usec) {
+long DTLSv1_get_timeout_wrapped(
+                                SSL *ssl,
+                                time_t *ptv_sec,
+                                long int *ptv_usec)
+{
     struct timeval tv = { 0 };
     int r = DTLSv1_get_timeout(ssl, &tv);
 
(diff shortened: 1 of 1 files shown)


alex:
Jenkins, ok to test.

alex:
Thanks for sending a pull request! 🎉 

What's the use case for adding these?

Gawen:
@alex the use case is to add DTLS support to pyopenssl (see https://github.com/Gawen/pyopenssl/tree/dev-dtlsv1). My patch needs these bindings.

Gawen:
```
ImportError: /home/travis/build/pyca/cryptography/.tox/py27/lib/python2.7/site-packages/cryptography/hazmat/bindings/_openssl.so: undefined symbol: DTLSv1_2_client_method
```

It appears the installed OpenSSL versions in the testing environments does not always support DTLSv1.2 ([see the failed cases here for example](https://jenkins.cryptography.io/job/cryptography-pr-builder/6063/)). This is an interesting situation, as the code could dynamically detect if the installed OpenSSL has it.

Any advise how to import `DTLSv1_2_{,client_,server_}method` *only* if it is present in the installed OpenSSL?

Gawen:
*I propose to handle the DTLSv1.2 support issue in a separate pull request* (see https://github.com/pyca/cryptography/pull/3286/commits/8dde92aad5db97fa176bf164783bdf9ba242edf4).

This pull request is now only about the DTLSv1.x's timeout handling functions.

src/_cffi_src/openssl/ssl.py
```diff
@@ -434,6 +434,10 @@
 long SSL_CTX_sess_misses(SSL_CTX *);
 long SSL_CTX_sess_timeouts(SSL_CTX *);
 long SSL_CTX_sess_cache_full(SSL_CTX *);
+
+/* DTLS support */
+long DTLSv1_get_timeout_wrapped(SSL *, time_t *, long int *);
```
reaperhulk: Style nit: Let's just declare the last arg as `long` (which is more consistent with our naming conventions).

src/_cffi_src/openssl/ssl.py
```diff
@@ -715,4 +719,23 @@
...
+long DTLSv1_get_timeout_wrapped(
+                                SSL *ssl,
+                                time_t *ptv_sec,
+                                long int *ptv_usec)
+{
+    struct timeval tv = { 0 };
+    int r = DTLSv1_get_timeout(ssl, &tv);
+
+    if (r == 1) {
+        if (ptv_sec) *ptv_sec = tv.tv_sec;
```
reaperhulk: Please wrap these in braces.

reaperhulk:
@Gawen: for binding functions that may not exist you'll want to either find the version of OpenSSL they were added and/or the `opensslconf.h` value that says they're not present (something like `OPENSSL_NO_DTLS` perhaps). You can then look in something like `ssl.py` to see how we do conditional bindings. I agree that should be a separate PR though.

Gawen:
@reaperhulk thanks for the feedback. This should be fixed.

I will work on a PR for DTLSv1.2 after this PR is merged.

reaperhulk:
https://github.com/reaperhulk/cryptography/commit/dc0dd43dc6be2dde0738530249b88cfb06cedc80 does this btw.

PR #3287, opened 2016-11-26
add support for encoding points to...

ofek:
the standard compressed format as described in [SEC 1 v2.0](http://www.secg.org/sec1-v2.pdf) section 2.3.3.

### DIFF
--- a/docs/hazmat/primitives/asymmetric/ec.rst
+++ b/docs/hazmat/primitives/asymmetric/ec.rst
@@ -158,13 +158,13 @@ Elliptic Curve Signature Algorithms
 
         :returns: A new instance of :class:`EllipticCurvePublicKey`.
 
-    .. method:: encode_point()
+    .. method:: encode_point(compress=False)
 
         .. versionadded:: 1.1
 
         Encodes an elliptic curve point to a byte string as described in
-        `SEC 1 v2.0`_ section 2.3.3. This method only supports uncompressed
-        points.
+        `SEC 1 v2.0`_ section 2.3.3. If `compress` is `True`, the point will
+        be encoded as a compressed public key.
 
         :return bytes: The encoded point.
 
--- a/docs/hazmat/primitives/asymmetric/ec.rst
+++ b/docs/hazmat/primitives/asymmetric/ec.rst
@@ -161,10 +161,13 @@ Elliptic Curve Signature Algorithms
     .. method:: encode_point(compress=False)
 
         .. versionadded:: 1.1
+        .. versionchanged:: 1.7
 
         Encodes an elliptic curve point to a byte string as described in
         `SEC 1 v2.0`_ section 2.3.3. If `compress` is `True`, the point will
         be encoded as a compressed public key.
+
+        :param bool compress: Flag for compression.
 
         :return bytes: The encoded point.
 
--- a/src/cryptography/hazmat/primitives/asymmetric/ec.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/ec.py
@@ -284,13 +284,18 @@ class EllipticCurvePublicNumbers(object):
     def public_key(self, backend):
         return backend.load_elliptic_curve_public_numbers(self)
 
-    def encode_point(self):
+    def encode_point(self, compress=False):
         # key_size is in bits. Convert to bytes and round up
         byte_length = (self.curve.key_size + 7) // 8
-        return (
-            b'\x04' + utils.int_to_bytes(self.x, byte_length) +
-            utils.int_to_bytes(self.y, byte_length)
-        )
+
+        if not compress:
+            return (
+                b'\x04' + utils.int_to_bytes(self.x, byte_length) +
+                utils.int_to_bytes(self.y, byte_length)
+            )
+
+        y_parity = b'\x03' if self.y & 1 else b'\x02'
+        return y_parity + utils.int_to_bytes(self.x, byte_length)
 
     @classmethod
     def from_encoded_point(cls, curve, data):
--- a/tests/hazmat/primitives/test_ec.py
+++ b/tests/hazmat/primitives/test_ec.py
@@ -192,10 +192,16 @@ def test_encode_point():
         16
     )
     pn = ec.EllipticCurvePublicNumbers(x, y, ec.SECP256R1())
-    data = pn.encode_point()
-    assert data == binascii.unhexlify(
+
+    uncompressed = pn.encode_point()
+    assert uncompressed == binascii.unhexlify(
         "04233ea3b0027127084cd2cd336a13aeef69c598d8af61369a36454a17c6c22ae"
         "c3ea2c10a84153862be4ec82940f0543f9ba866af9751a6ee79d38460b35f442e"
+    )
+
+    compressed = pn.encode_point(compress=True)
+    assert compressed == binascii.unhexlify(
+        "02233ea3b0027127084cd2cd336a13aeef69c598d8af61369a36454a17c6c22aec"
     )
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,8 @@ Changelog
   will need to upgrade.
 * Added support for Diffie-Hellman key exchange using
   :meth:`~cryptography.hazmat.primitives.asymmetric.dh.DHPrivateKeyWithSerialization.exchange`
+* Added support for encoding compressed public keys using
+  :meth:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicNumbers.encode_point`
 
 1.6 - 2016-11-22
 ~~~~~~~~~~~~~~~~


reaperhulk:
Jenkins, ok to test

ofek:
@alex @reaperhulk r4r

docs/hazmat/primitives/asymmetric/ec.rst
```diff
@@ -158,13 +158,13 @@ Elliptic Curve Signature Algorithms
 
         :returns: A new instance of :class:`EllipticCurvePublicKey`.
 
-    .. method:: encode_point()
+    .. method:: encode_point(compress=False)
 
         .. versionadded:: 1.1
```
reaperhulk: This needs a .. versionchanged:: 1.7

docs/hazmat/primitives/asymmetric/ec.rst
```diff
@@ -158,13 +158,13 @@ Elliptic Curve Signature Algorithms
...
 
         Encodes an elliptic curve point to a byte string as described in
-        `SEC 1 v2.0`_ section 2.3.3. This method only supports uncompressed
-        points.
+        `SEC 1 v2.0`_ section 2.3.3. If `compress` is `True`, the point will
+        be encoded as a compressed public key.
```
reaperhulk: This needs a doc line for the `compress` argument.

src/cryptography/hazmat/primitives/asymmetric/ec.py
```diff
@@ -284,13 +284,18 @@ def __init__(self, x, y, curve):
...
-            utils.int_to_bytes(self.y, byte_length)
-        )
+
+        if not compress:
+            return (
+                b'\x04' + utils.int_to_bytes(self.x, byte_length) +
+                utils.int_to_bytes(self.y, byte_length)
+            )
+
+        y_parity = b'\x03' if self.y & 1 else b'\x02'
```
reaperhulk: Is this generally true for all possible curves or limited to a subset?

tests/hazmat/primitives/test_ec.py
```diff
@@ -192,12 +192,18 @@ def test_encode_point():
...
+
+    uncompressed = pn.encode_point()
+    assert uncompressed == binascii.unhexlify(
         "04233ea3b0027127084cd2cd336a13aeef69c598d8af61369a36454a17c6c22ae"
         "c3ea2c10a84153862be4ec82940f0543f9ba866af9751a6ee79d38460b35f442e"
     )
 
+    compressed = pn.encode_point(compress=True)
+    assert compressed == binascii.unhexlify(
+        "02233ea3b0027127084cd2cd336a13aeef69c598d8af61369a36454a17c6c22aec"
```
reaperhulk: How do we know this is correct? Is there a source for test vectors for compressed points?

src/cryptography/hazmat/primitives/asymmetric/ec.py
ofek: This is standard and true of all curves.

tests/hazmat/primitives/test_ec.py
reaperhulk: Okay, looking at 2.3.3 I see that there are two procedures for determining the bit value that depend on the type of field. The code in this patch accounts for 2.2.1, which assumes that all supported curves are prime finite fields. cryptography also supports curves like `sect233r1` (aka NIST B-233), which is a finite field `F2m`. This code does not appear to handle that case?

lvh:
There are many ways to compress points. Are we okay with only supporting SEC compression for SEC curves? (That seems reasonable; I'm just checking.) You might run into trouble if you somehow end up with an anomalous curve with no obvious compression strategy; I know of at least 5 off-hand for F2m.

src/cryptography/hazmat/primitives/asymmetric/ec.py
lvh: Confirmed; although keep in mind that "standard" means "for SEC curves".

lvh:
Okay, so, here's how SECG point compression works for F2m curves.

You have a general curve equation y^2 + xy = x^3 + ax^2 + b. [1] A point with x=0 has order 2, so x!=0. Because x!=0, you can divide by it, so we can come up with an alternative equation with different parameters (x,z) instead of (x,y), where z=y/x. So, you get the equation:

z^2 + z = x +a + b/x^2 [2]

Note that the RHS can be factored:

z*(z+1) = x + a + b/x^2 [2]

Remember that a and b are public constants. The usual naive point compression has to do two things: communicate x, and then which one of two points (P or -P) it is. (Smarter point compression does roughly the same thing, but manages to cram that extra bit in to an x coordinate with cheese.) Either way; we need to reconstruct x, y  to find P = (x, y) or -P = (x, x+y).

The compressed point gives us x. Via [2], that gives us z*(z+1) and then z. Since z = y/x, y=zx; so we need only multiply by x to recover y.

So, easy! All we need to do is compute z, so that we can ship both z & 1 and x into our compressed point and perform the recovery.

This is the part where most people's eyes start glazing over because there are scary terms like "Frobenius" and "trace" and wrong-looking math where (a + b)^2 == a^2 + b^2. I could go into details of why that is, or you could just trust me ;)

The trace is defined as follows:

```
trace = lambda x: sum(x ** (2 ** i) for i in range(m))
```

I'm liberally using Python notation here; keep in mind that the math still happens in F2m. This turns out to be drastically simpler than it looks. Because of the (a + b)^2 = a^2 + b^2 magic, it turns out that trace(a^2) = trace(a)^2, and therefore its value is always 0 or 1 (!!!), or equivalently trace(a^2) = trace(a) [3].  OK. Still with me? Remember [2] (flipped around):

x + a + b/x^2 = z^2 + z 

so, consider the following equation, acquired by applying trace:

trace(x + a + b/x^2) = trace(z^2 + z)
= trace(z^2) + trace(z) [trace is linear]
= trace(z) + trace(z) [3] 
= 0               (F2m addition)

Well that was easy. We need one more bit of magic pixie dust: the half-trace. The half-trace is the same thing as the trace, but only for all the odd-numbered elements in the sum. It, similarly, has some awesome properties:

half-trace(q^2) + half-trace(q)
= trace(q) + q^(2^m) (cancel out except for the last one)
= trace(q) + q (F2m!)

So, half-trace(q^2) + half-trace(q) = trace(q) + q. If we set q = z^2+z as per our previous equation, and we already know the trace of that is 0, we get:

half-trace(q)^2 + half-trace(q) = q

But  q = z^2 + z = x + a + b/x^2  and we know that value (x is known, a, b constant), so we can compute the half-traces:

half-trace(x + a + b/x^2)^2 + half-trace( x + a + b/x^2) = z^2 + z

I think this simplifies away to z being either half-trace(q) or half-trace(q) + 1.

Most of this derivation is wrong for curves used for pairing. Don't do that. If you're feeling brave, HEHEC has this stuff.

lvh:
Oh, I forgot: you might be wondering why even bother shlepping around with the alternative representation, worrying about the endomorphism, et cetera. Well: finding the inverse for doing z=y/x is very very very expensive. Doing the trace computations, by leveraging the Frobenius properties, is much faster.

lvh:
(I don't know why the SECG algorithm doesn't mention the endomorphism implementation tricks; maybe it's mentioned elsewhere in the document. I'd assume so; it's just been a while since I last read it.)

ofek:
@lvh Thank you for the very detailed explanation! I understand this more now. However, I am still a bit perplexed. How exactly do you compute z? Is it simply:

`sum(x ** (2 ** i) for i in range(m) if i % 2)` ?

If so, that is inordinately slow. Perhaps that is just Python though?

Is there a 'trick' to make the computation less complex? For example, to derive y from x (and extra byte) on prime curves I learned in #bitcoin from Bitcoin core dev Gregory Maxwell one can leverage [Tonelli–Shanks algorithm](https://en.wikipedia.org/wiki/Tonelli%E2%80%93Shanks_algorithm). I've implemented that as follows:

```python
def x_to_y(x, y_bit, curve):
    a = curve.a
    b = curve.b
    field_size = curve.p

    y = pow(x**3 + a*x + b, (field_size + 1) // 4, field_size)

    if y & 1 != y_bit & 1:
        y = FIELD_SIZE - y

    return y
```

This executes in microseconds.

ofek:
@lvh Oh wait, is the exponentiation mod m? Like:

`sum(pow(x, 2 ** i, m) for i in range(m) if i % 2)` ?

lvh:
No, you shouldn't compute it that way. I was going from the expensive, inversion-based approach suggested in the SECG document to the way you'd actually implement it; as I mentioned after that implementation:

> I'm liberally using Python notation here; keep in mind that the math still happens in F2m

I'm inclined to suggest that this should be a backend method that might have a Python fallback implementation but generally calls OpenSSL `ec_GF2m_simple_point2oct`, `ec_GF2m_simple_oct2point` et cetera. You can see the OpenSSL implementation here: https://github.com/openssl/openssl/blob/6f0ac0e2f27d9240516edb9a23b7863e7ad02898/crypto/ec/ec2_oct.c#L127 (it starts getting good around 175; their `yxi` means y times x inverse; AKA my z).

ofek:
Taking a look now. Was the last thing I put correct though? It's mod m?

lvh:
@ofek I don't think so; the expression you posted looks like something for the half-trace; but it's still in GF2^m either way. However; the point is that you should never actually have to compute the trace; the curves are selected such that you never actually evaluate that expression. You can statically determine that the trace you care about is 0. If you care about computing the half-trace; that similarly simplifies down to a much simpler computation.

ofek:
@lvh Ok ty. What would you do for the Python fallback implementation?

lvh:
I'm not sure it should exist; that's up to @reaperhulk. If it does; as much of a transliteration of that OpenSSL code as possible. One difference is that the code I linked assumes yxi is already available, and that's not true for the current Python implementation of this type.

lvh:
Oops; my mistake it doesn't assume yxi is already available. Unfortunately, that means that you'll have to transliterate the field div instead, too. OpenSSL has the same high-level API as the SECG suggests by showing it as an inverse; it just relies on that implementation knowing that it can do it more efficiently on that curve.

ofek:
@reaperhulk @alex @lvh I apologize but I don't think I know enough about OpenSSL nor FFI to do what needs to be done here. Should I close this PR?

lvh:
I can't answer that one for you :) I don't really have enough free resources to implement this PR, but I can volunteer to review it. Presumably there should be a ticket; and if there is, this PR should be linked to it since it'd be unfortunate if the documentation on how to implement this feature disappeared into void.

ofek:
@reaperhulk @alex OTOH, if there is no one who will be able to take this on in the near future, I'm totally willing to finish the implementation of (de)compression now without calling into C if we merge https://github.com/pyca/cryptography/pull/2476 so we/users have access to curve values. Without access to each curve's properties, compression cannot be achieved. A python implementation may also be useful anyway for times when a backend other than OpenSSL is desired.

Note: As @lvh said, a Python implementation won't be slow. The issue is that curve attributes aren't exposed atm.

wbond:
@lvh You mentioned that you know of a bunch of different compression methods for F2m curves. Would you have time to point me in the direction of info about those?

lvh:
@wbond Sorry; I didn't notice that ping on the PR. Are you volunteering to implement them for this library?

Diego Aranha was nice enough to point out that RELIC has a pretty legible standard (i.e. SECG) F2m point compression implementation: https://github.com/relic-toolkit/relic/blob/master/src/eb/relic_eb_pck.c

Also, if you have access to HEHEC (or kindly ask to borrow my copy) you can check out section  13.3.7.

wbond:
@lvh No, I am not volunteering to work on this PR specifically. I maintain a Python ASN.1 library, and lurk around here the IRC channel to try and learn whatever I can. I do appreciate the info!

lvh:
@wbond OK, cool. An incomplete list, I'm sure, but just for binary curves, off the top of my head:

- SECG/IEEE P1363
- Seroussi's method
- King's method (https://www.iacr.org/archive/pkc2004/29470331/29470331.pdf)
- 2 methods by Lopez et al. ("Point Halving Revisited"; I found it on the IEEE site but nothing that's not paywalled, sorry!)

lvh:
There is also of course the "naive" method but that's pretty much just SECG without the prefix byte.

ofek:
@wbond "Field inversion and point halving revisited" [http://sci-hub.cc/10.1109/TC.2004.43](url)

lvh:
Woo, thanks @ofek :)

ofek:
@lvh No problem!

Hey I found another nice resource [https://www.security-audit.com/files/x9-62-09-20-98.pdf](url) section 4.2.2.  Is this the same method as SECG/IEEE P1363?  Does y<sub>p</sub>x<sub>p</sub><sup>-1</sup> mean inverse point?

lvh:
Yeah; that's just a different way of spelling Z=Y/X.

lvh:
(Some people prefer that notation because technically division between elements isn't defined; only multiplication by inverses is -- but it's a perfectly unambiguous notation, so I use it.)

ofek:
@lvh and z is actually `half-trace(x + a + b/x^2)^2 + half-trace( x + a + b/x^2)` ?

lvh:
Errr; get a second source to confirm because I'm not sure I want to walk all the way through that derivation again, but I'm pretty sure Z has two solutions (hence the extra bit): half-trace(q)  half-trace(q) + 1, where q=x + a + b/x^2. In practice, because the half-trace is linear, so you can compute it by adding together (in F2m, so this is just XOR!) some precomputed values (please don't implement this, let alone in Python) per 1-bit of the input.

PR #3290, opened 2016-11-29
Bind TLSEXT_STATUSTYPE_ocsp.

Lukasa:
This was missing from the earlier OCSP bindings, and is necessary for correctly-functioning client-side OCSP. Specifically, this is the only valid argument to `SSL_set_tlsext_status_type()`. PyOpenSSL can of course just hard-code the value, but given that we need to depend on the next version of Cryptography to get `OPENSSL_malloc` we may as well just do this properly.


reaperhulk:
Looks like it's available in every version of OpenSSL we test against but would you mind lookIng up when this constant was added?

Lukasa:
According to the changelog and git, 0.9.8h.

PR #3291, opened 2016-11-29
cffi bindings additions for pypy's _hashlib module

planrich:
There are some more bindings I used in the stdlib module. Unsure about OBJ_NAME, I found no OpenSSL functions that could access the fields, so I assume it is meant to be used directly?

### DIFF
--- a/src/_cffi_src/openssl/objects.py
+++ b/src/_cffi_src/openssl/objects.py
@@ -15,6 +15,8 @@ typedef struct {
     const char *name;
     const char *data;
 } OBJ_NAME;
+
+static const long OBJ_NAME_TYPE_MD_METH;
 """
 
 FUNCTIONS = """
@@ -36,8 +38,6 @@ void OBJ_NAME_do_all(int, void (*) (const OBJ_NAME *, void *), void *);
 MACROS = """
 /* OBJ_cleanup became a macro in 1.1.0 */
 void OBJ_cleanup(void);
-
-#define OBJ_NAME_TYPE_MD_METH ...
 """
 
 CUSTOMIZATIONS = """
--- a/src/_cffi_src/openssl/objects.py
+++ b/src/_cffi_src/openssl/objects.py
@@ -9,6 +9,12 @@ INCLUDES = """
 """
 
 TYPES = """
+typedef struct {
+    int type;
+    int alias;
+    const char *name;
+    const char *data;
+} OBJ_NAME;
 """
 
 FUNCTIONS = """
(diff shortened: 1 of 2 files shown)


alex:
Jenkins, ok to test.

src/_cffi_src/openssl/objects.py
```diff
@@ -24,11 +30,14 @@
...
 ASN1_OBJECT *OBJ_dup(const ASN1_OBJECT *);
 int OBJ_create(const char *, const char *, const char *);
+void OBJ_NAME_do_all(int, void (*) (const OBJ_NAME *, void *), void *);
 """
 
 MACROS = """
 /* OBJ_cleanup became a macro in 1.1.0 */
 void OBJ_cleanup(void);
+
+#define OBJ_NAME_TYPE_MD_METH ...
```
reaperhulk: We typically do this via "static const long" (assuming it is an integer) rather than letting cffi substitute on the define. If it's a char * then const char const *

reaperhulk: And it should live in types

src/_cffi_src/openssl/objects.py
```diff
@@ -9,6 +9,14 @@
 """
 
 TYPES = """
+typedef struct {
+    int type;
+    int alias;
+    const char *name;
+    const char *data;
+} OBJ_NAME;
```
reaperhulk: This struct isn't opaque in 1.1.0 so we can bind it this way -- how is it used in CPython right now though?

tiran: Yes, I didn't realize that we still use the struct directly.

tiran: The struct is still available in master: https://github.com/openssl/openssl/blob/master/include/openssl/objects.h#L935

planrich: the fields alias and name are used. CPython stdlib uses them to find out which hashing algorithms are available on the target platform. (it calls `OBJ_NAME_do_all(OBJ_NAME_TYPE_MD_METH, ...)`)

planrich: in the callback of `OBJ_NAME_do_all` it saves the names if it is not an alias.

reaperhulk: Okay, looking at the code there's `EVP_MD_do_all` in newer OpenSSL but that just calls `OBJ_NAME_do_all` anyway, so it appears this is indeed required. Okay 👍

PR #3292, opened 2016-11-29
Add minversion to tox.ini

tiran:
Cryptography uses new features from tox 2.4. Tox 2.3 happily ignores the
new config stanzes and doesn't install dependency. This can lead to strange
test failures.

With minversion=2.4, tox 2.3 fails to run properly:
$ tox
ERROR: tox version is 2.3.1, required is at least 2.4

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,4 +1,5 @@
 [tox]
+minversion = 2.4
 envlist = py26,py27,pypy,py33,py34,py35,docs,pep8,py3pep8
 
 [testenv]


tox.ini
```diff
@@ -1,4 +1,5 @@
 [tox]
+minversion=2.4
```
alex: Tiny nit, we use spaces around the `=`

derwolfe:
Interesting, I'd have expected https://github.com/pyca/cryptography/blob/master/dev-requirements.txt#L5 to have caught this. But, I think I misjudged when it was actually being invoked.

tiran:
@derwolfe I ran into issues on my local machine while I tried to run tests locally with my system's tox.

PR #3295, opened 2016-12-03
clean up int_from_bytes

ofek:
7x speed-up and code is more readable

### DIFF
--- a/src/cryptography/utils.py
+++ b/src/cryptography/utils.py
@@ -6,7 +6,6 @@ from __future__ import absolute_import, division, print_function
 
 import abc
 import binascii
-import codecs
 import inspect
 import sys
 import warnings
@@ -48,7 +47,7 @@ else:
         assert byteorder == 'big'
         assert not signed
 
-        return int(codecs.encode(data, 'hex'), 16)
+        return int(data.encode('hex'), 16)
 
 
 def int_to_bytes(integer, length=None):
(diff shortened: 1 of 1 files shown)


src/cryptography/utils.py
```diff
@@ -48,18 +49,7 @@ def int_from_bytes(data, byteorder, signed=False):
...
-        result = 0
-
-        while len(data) > 0:
-            digit, = struct.unpack('>I', data[:4])
-            result = (result << 32) + digit
-            # TODO: this is quadratic in the length of data
-            data = data[4:]
-
-        return result
+        return int(codecs.encode(data, 'hex'), 16)
```
alex: Can this be written `data.encode("hex")`, instead of using the `codecs` module?

alex:
DId this show up as a hotspot in a benchmark, or did you just happen across it?

alex:
Jenkins, ok to test

src/cryptography/utils.py
ofek: We can't because Python 3 is missing some built-in codecs [http://bugs.python.org/issue7475](url)

alex: This branch isn't used in python3 though, python3 always hits the `int.from_bytes` branch, right?

ofek: Yes, for 3.2+ (dunno about PyPy3). Would you like me to change it?

alex: Yeah, I think it's a bit cleaner.

PR #3297, opened 2016-12-03
DH serialization

palaviv:
This PR adds serialization for DH parameters and keys.
I have not finished writing tests yet but I would like to get some feedback before I continue.
This should solve Issue #2399.


reaperhulk:
Can you enumerate the list of ways OpenSSL can write public and private DH keys (as well as DH params) in a comment?

palaviv:
I hope I understand your question correctly @reaperhulk.
The key's can be written using SPKI or PKCS#8. You can see in the following link:
https://wiki.openssl.org/index.php/Diffie_Hellman#Using_the_Low_Level_APIs (under the code).

I am not sure about the parameters... The reading of the parameters has a comment "Transparently read in PKCS#3 or X9.42 DH parameters" but I couldn't find anything about the writing.

alex:
Jenkins, retest this please.

reaperhulk:
(Coverage issues here)

palaviv:
I added more tests that should now get this PR to 100% coverage. However I ran into a problem. In `load_der_public_key` we first try using `d2i_PUBKEY_bio` and if that fails we try `d2i_RSAPublicKey_bio` and for some reason `d2i_RSAPublicKey_bio` don't fail when it receives the DH params DER key. I am currently not sure how to solve this and would appreciate any advice.

palaviv:
Solved the problem I raised earlier by doing `DH_check` after calling `d2i_DHparams_bio`.

docs/hazmat/primitives/asymmetric/dh.rst
```diff
@@ -99,6 +99,25 @@ Group parameters
...
 
+    .. method:: public_bytes(encoding, format)
+
+        .. versionadded:: 1.7
+
+        Allows serialization of the key to bytes. Encoding (
+        :attr:`~cryptography.hazmat.primitives.serialization.Encoding.PEM` or
+        :attr:`~cryptography.hazmat.primitives.serialization.Encoding.DER`) and
+        format (
+        :attr:`~cryptography.hazmat.primitives.serialization.PublicFormat.SubjectPublicKeyInfo`)
```
reaperhulk: The serialization of DH parameters isn't actually a `subjectPublicKeyInfo` structure (which requires an algorithm identifier), but is instead just an ASN.1 sequence of two integers (`p` and `g`). Elsewhere in the code you refer to this as `PKCS3`.

Adding serialization for parameters is a new feature (we don't support it for `DSAParameters` at the moment) and since it is a "new" format I'd like to have this be added in a future PR. Could you remove this for now?

docs/hazmat/primitives/asymmetric/dh.rst
```diff
@@ -145,6 +164,35 @@ Key interfaces
 
         :return bytes: The agreed key. The bytes are ordered in 'big' endian.
 
+    .. method:: private_bytes(encoding, format, encryption_algorithm)
+
+        .. versionadded:: 1.7
```
reaperhulk: 1.8 (all 1.7s need to be 1.8 now)

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1049,7 +1059,18 @@ def load_pem_public_key(self, data):
                 evp_pkey = self._rsa_cdata_to_evp_pkey(rsa_cdata)
                 return _RSAPublicKey(self, rsa_cdata, evp_pkey)
             else:
-                self._handle_key_loading_error()
+                # PKCS3
```
reaperhulk: This is part of the code to remove and add to a future PR

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1092,8 +1113,23 @@ def load_der_public_key(self, data):
             return self._evp_pkey_to_public_key(evp_pkey)
         else:
             # It's not a (RSA/DSA/ECDSA) subjectPublicKeyInfo, but we still
-            # need to check to see if it is a pure PKCS1 RSA public key (not
-            # embedded in a subjectPublicKeyInfo)
+            # need to check to see if it PKCS3 dh parameters or
```
reaperhulk: This is part of the code to remove and add to a future PR

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1664,6 +1700,15 @@ def _public_key_bytes(self, encoding, format, key, evp_pkey, cdata):
                 write_bio = self._lib.i2d_RSAPublicKey_bio
 
             key = cdata
+        elif format is serialization.PublicFormat.PKCS3:
```
reaperhulk: This is part of the code to remove and add to a future PR

src/cryptography/hazmat/backends/openssl/dh.py
```diff
@@ -40,6 +41,20 @@ def parameter_numbers(self):
     def generate_private_key(self):
         return self._backend.generate_dh_private_key(self)
 
+    def public_bytes(self, encoding, format):
```
reaperhulk: This is part of the code to remove and add to a future PR

src/cryptography/hazmat/backends/openssl/dh.py
```diff
@@ -180,3 +206,17 @@ def public_numbers(self):
 
     def parameters(self):
         return _dh_cdata_to_parameters(self._dh_cdata, self._backend)
+
+    def public_bytes(self, encoding, format):
+        if format is serialization.PublicFormat.PKCS1:
```
reaperhulk: The error handling in `_public_key_bytes` won't handle the case where you pass openssh for both format and encoding (you'll get an assertion error because the openssh code path assumes the fallthrough is elliptic curve). Since we only support one publicformat here I think just doing `if format not serialization.PublicFormat.SubjectPublicKeyInfo: raise` is probably a reasonable path.

palaviv:
I removed the support in the traditionalopenssl serialization. There is no `PEM_write_bio_DHPrivateKey` so I am not sure if I should use `PEM_write_bio_DHparams`?

tests/hazmat/primitives/test_dh.py
```diff
@@ -285,3 +289,151 @@ def test_dh_vectors(self, backend, vector):
...
+        key_bytes = load_vectors_from_file(
+            key_path,
+            lambda pemfile: pemfile.read(), mode="rb"
+        )
+        vec = load_vectors_from_file(
+            os.path.join("asymmetric", "DH", "dhkey.txt"),
+            load_nist_vectors)[0]
+        key = loader_func(key_bytes, None, backend)
+        private_numbers = key.private_numbers()
+        assert private_numbers.x == int(vec["x"], 16)
```
reaperhulk: I don't understand quite what's going on in this test. You're loading a key, then testing the `private_numbers` conversion to see if the key you loaded matches the test vector. Then you're re-serializing and checking that the bytes == original_bytes. The latter appears to be the only part of the test we need here. Testing private_numbers is already covered in other tests is it not?

palaviv: This test covers both loading a key and serializing it back. I test that the values I read from `dhkey.txt` are indeed the one that I am expecting and not only that the result of the serialization is the same.

reaperhulk: It doesn't seem necessary to do that verification since we already have tests for the `numbers` methods that check that behavior.

palaviv: Which tests?

reaperhulk: It would appear the tests for private_numbers, etc don't actually test that the numbers they output are the ones we expected. So we *should* have a test for that, but I still think that should be a separate test than this :)

palaviv: I looked at what you do at test_dsa.py and tried to do the same. Hope this is what you were looking for.

reaperhulk:
Thanks again @palaviv!

reaperhulk:
(We'll need a separate PR to add a changelog entry)

PR #3312, opened 2016-12-12
Drop 1.0.0

alex:


tests/hazmat/bindings/test_openssl.py
```diff
@@ -79,11 +79,11 @@ def test_ssl_mode(self):
     def test_conditional_removal(self):
         b = Binding()
 
-        if b.lib.CRYPTOGRAPHY_OPENSSL_101_OR_GREATER:
-            assert b.lib.CMAC_Init
+        if b.lib.CRYPTOGRAPHY_OPENSSL_110_OR_GREATER:
+            assert b.lib.TLS_ST_OK
```
reaperhulk: Looks like this causes a failure on the libressl builder

reaperhulk:
Jenkins, retest this please

docs/faq.rst
```diff
@@ -40,27 +40,14 @@ If you have no other libraries using OpenSSL in your process, or they do not
...
-The OpenSSL project has dropped support for the 0.9.8 release series. Since it
-is no longer receiving security patches from upstream, ``cryptography`` is also
-dropping support for it. To fix this issue you should upgrade to a newer
-version of OpenSSL (1.0.1 or later). This may require you to upgrade to a newer
-operating system.
+Installing cryptography with OpenSSL 0.9.8 or 1.0.0 fails
+---------------------------------------------------------
+
+The OpenSSL project has dropped support for the 0.9.8 and 1.0.0 release
+serieses. Since they are no longer receiving security patches from upstream,
```
reaperhulk: series is both singular and plural, and serieses is definitely not a word 😄

src/cryptography/hazmat/backends/openssl/dsa.py
```diff
@@ -7,32 +7,14 @@
...
-    _calculate_digest_and_algorithm, _truncate_digest
+    _calculate_digest_and_algorithm
 )
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import (
     AsymmetricSignatureContext, AsymmetricVerificationContext, dsa
 )
 
 
-def _truncate_digest_for_dsa(dsa_cdata, digest, backend):
```
reaperhulk: The `_truncate_digest_for_ecdsa` doesn't have an equivalent comment, but could we check to see if that is no longer required? I just tested on 1.1.0 and it definitely does it, but hopefully ECDSA and DSA match up here.

tests/hazmat/backends/test_openssl.py
```diff
@@ -78,7 +75,7 @@ def test_supports_cipher(self):
         assert backend.cipher_supported(None, None) is False
 
     def test_aes_ctr_always_available(self):
```
reaperhulk: This test was only to verify our 0.9.8/1.0.0 trick of supporting CTR was present. It can be entirely removed now.

Issue #3119, opened 2016-08-29
Allow use of bytearray buffers in cipher context calls


reaperhulk:
As of cffi 1.7 `ffi.from_buffer` allows the use of bytearrays. This potentially allows us to support an API for encryption/decryption that significantly reduces memory copies and improves performance. @alex has suggested `update_into` (analogous to [socket.recv_into](https://docs.python.org/2/library/socket.html#socket.socket.recv_into). Some early performance tests indicate we'll see >25% performance improvement on some workloads (repeated large calls to `update_into`).

Here's roughly what an `update_into` could look like:

``` python
    def update_into(self, data, buf):
        if len(buf) < (len(data) + self._block_size - 1):
            raise ValueError("buffer must be bigger")

        buf = self._backend._ffi.from_buffer(buf)
        outlen = self._backend._ffi.new("int *")
        res = self._backend._lib.EVP_CipherUpdate(self._ctx, buf, outlen, data,
                                                  len(data))
        self._backend.openssl_assert(res != 0)
        return outlen[0]
```

Since this isn't present until cffi 1.7 we'd have to decide if we wanted to rev our required version or enable/disable this path based on cffi version. If we move our required version to 1.7 we'll have to drop support for every PyPy < 5.3.

(This is a spiritual continuation of some work I did in #2163)

alex:
Probably we could implement `update` in terms of `update_into` to reduce the amount of code we have to write.

I think `update` on `CipherContext` is the only var-sized output in our API, and therefore the only one justifying this attention.

reaperhulk:
We won't be able to implement `update` using `update_into` unless we're willing to require PyPy 5.3 and cffi 1.7 as minimum versions.

If we leave our minimum versions then we'll have to do some version checking on import (or else do it every time `update_into` is called, which seems somewhat counter to the performance goal of the API).

```python
from distutils.version import StrictVersion

import cffi

if StrictVersion(cffi.__version__) < StrictVersion("1.7.0"):
    # possibly mutate `update_into` to raise an error
```

reaperhulk:
That maybe should be `from pkg_resources import parse_version`. Dunno what the current preferred way is :)

PR #3316, opened 2016-12-13
fix a regression in int_from_bytes

reaperhulk:


alex:
test please

ofek:
Py2 doesn't have built-in `bytes`.

alex:
The name `bytes` exists on py2 just fine.

ofek:
Huh ty, TIL. [https://docs.python.org/3/whatsnew/2.6.html#pep-3112-byte-literals](url)

ofek:
Hey quick question wrt terminology, as I'm unfamiliar with this.

Generally, are internal utilities considered part of the public API? If not, is this actually classified as a regression? Looking at the offending line [https://github.com/mitya57/secretstorage/blob/master/secretstorage/dhcrypto.py#L36](url) it seems a bytearray is being passed to our int_from_bytes.

Cryptography doesn't use bytearrays at all atm (eventually maybe #3119)

alex:
Technically no, the public API is defined by what is documented. That said, we know we have a lot of downstream users, so we try to avoid breaking them (particularly unintentionally!)

PR #3318, opened 2016-12-13
changelog and version bump for 1.7.1

reaperhulk:


alex:
codecov is claiming this has less than 100% coverage, but a review of the data makes clear this is a bug. I am therefore overriding.

PR #3321, opened 2016-12-14
restore this constant, pyopenssl needs it

alex:


tiran:
I guess this means 1.7.2 will come out soon?

reaperhulk:
Nope this is only in master (it's due to the 1.0.0 removal)

PR #3322, opened 2016-12-14
Add DH serialization bindings

palaviv:
This PR add d2i_DHparams_bio, i2d_DHparams_bio bindings for use in PR #3297.


src/_cffi_src/openssl/x509.py
```diff
@@ -336,6 +336,9 @@
 int i2d_re_X509_CRL_tbs(X509_CRL *, unsigned char **);
 void X509_REQ_get0_signature(const X509_REQ *, const ASN1_BIT_STRING **,
                              const X509_ALGOR **);
+
+DH *d2i_DHparams_bio(BIO *, DH **);
+int i2d_DHparams_bio(BIO *, DH *);
```
alex: These should be in `dh.py`, not `x509.py`

PR #3328, opened 2016-12-19
add memory limit check for scrypt

### LINKED ISSUE #3323: scrypt failures on 32-bit arches
The Debian builds are failing on a whole bunch of arches (I think it's all of the 32-bit ones) in the scrypt  tests with `[_OpenSSLErrorWithText(code=101404844L, lib=6, func=181, reason=172, reason_text='error:060B50AC:digital envelope routines:EVP_PBE_scrypt:memory limit exceeded')]`.

You can see a full build log here: https://buildd.debian.org/status/fetch.php?pkg=python-cryptography&arch=i386&ver=1.7.1-1&stamp=1481779760

And the failing arches here: https://buildd.debian.org/status/package.php?p=python-cryptography

reaperhulk:
fixes #3323

### DIFF
--- a/tests/hazmat/primitives/test_scrypt.py
+++ b/tests/hazmat/primitives/test_scrypt.py
@@ -38,6 +38,8 @@ def test_memory_limit_skip():
     with pytest.raises(pytest.skip.Exception):
         _skip_if_memory_limited(1000, {"p": 16, "r": 64, "n": 1024})
 
+    _skip_if_memory_limited(2 ** 31, {"p": 16, "r": 64, "n": 1024})
+
 
 @pytest.mark.requires_backend_interface(interface=ScryptBackend)
 class TestScrypt(object):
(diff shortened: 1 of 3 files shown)


mithrandi:
Diff looks good to me; this is a comment rather than a review because I want to test this on Debian i386 first (should be able to do that later today).

tests/hazmat/primitives/test_scrypt.py
```diff
@@ -77,6 +97,7 @@ def test_password_not_bytes(self, backend):
 
     @pytest.mark.parametrize("params", vectors)
     def test_verify(self, backend, params):
+        _skip_if_memory_limited(backend._scrypt_mem_limit, params)
```
alex: is there a reason not to just write `sys.maxsize // 2` here instead of the backend dance?

reaperhulk:
jenkins, retest this please

tests/hazmat/primitives/test_scrypt.py
alex: I'm not wild about the tests consuming a non-public API that's not in the interface. Is there a reason not to just expose a `_MAX_MEMORY` constant in the `scrypt` module and have both use it?

reaperhulk: I'm fine with it as a non-public constant in the scrypt module. I just don't want to hard code it in the tests. I'll make the change.

PR #3329, opened 2016-12-19
add openssl_version_number & doc openssl_version_text

### LINKED ISSUE #3315: Document openssl_version_number
https://cryptography.io/en/latest/hazmat/backends/openssl/

reaperhulk:
fixes #3315

### DIFF
--- a/docs/hazmat/backends/openssl.rst
+++ b/docs/hazmat/backends/openssl.rst
@@ -45,9 +45,10 @@ greater.
 
         .. versionadded:: 1.8
 
-        :return int: The integer version of the OpenSSL library. This is
+        :return int: The integer version of the loaded OpenSSL library. This is
             defined in ``opensslv.h`` as ``OPENSSL_VERSION_NUMBER`` and is
-            typically shown in hexadecimal (e.g. ``0x1010003f``).
+            typically shown in hexadecimal (e.g. ``0x1010003f``). This is
+            not necessarily the same version as it was compiled against.
 
     .. method:: activate_osrandom_engine()
 
--- a/docs/hazmat/backends/openssl.rst
+++ b/docs/hazmat/backends/openssl.rst
@@ -38,7 +38,7 @@ greater.
 
     .. method:: openssl_version_text()
 
-        :return str: The friendly string name of the loaded OpenSSL library.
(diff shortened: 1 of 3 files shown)


alex:
Is there a reason not to break it down and return some sort of release object that offers this as structured data, instead of making people screw around with masks?

reaperhulk:
The devil's advocate in me says that any user who actually needs to test OpenSSL versions is probably going to be copying it directly out of an `opensslv.h` so the utility of a fancy object is somewhat limited.

reaperhulk:
This version check also won't help users who are on libressl since `OPENSSL_VERSION_NUMBER` is fixed to `0x20000000` there. We'd need to add `LIBRESSL_VERSION_NUMBER`, which...let's wait for a request.

docs/hazmat/backends/openssl.rst
```diff
@@ -36,6 +36,19 @@ greater.
...
+    .. method:: openssl_version_text()
+
+        :return str: The friendly string name of the loaded OpenSSL library.
+            This is not necessarily the same version as it was compiled against.
+
+    .. method:: openssl_version_number()
+
+        .. versionadded:: 1.8
+
+        :return int: The integer version of the OpenSSL library. This is
```
alex: Since you specified that `text` is the linked version, you probably ought to say this is the compiled version not the linked version.

tests/hazmat/backends/test_openssl.py
```diff
@@ -71,6 +71,9 @@ def test_openssl_version_text(self):
             backend.openssl_version_text().startswith("LibreSSL")
         )
 
+    def test_openssl_version_number(self):
+        backend.openssl_version_number() > 0
```
alex: assert

docs/hazmat/backends/openssl.rst
```diff
@@ -36,6 +36,20 @@ greater.
 
         The string name of this backend: ``"openssl"``
 
+    .. method:: openssl_version_text()
+
+        :return str: The friendly string name of the loaded OpenSSL library.
```
reaperhulk: Yeah we call `decode('ascii')` so you're right. Fixing.

Issue #3348, opened 2017-01-17
Enable Cryptography_STATIC_CALLBACKS for Python 3


tiran:
Now with locks and osrandom engine in C, is there any reason left to use dynamic callbacks for Python 3.5+? src/_cffi_src/openssl/callbacks.py still has the workaround for #2970. Starting with 1.7.0, ```bindings.ffi_callback()``` is only used for the password callback.

reaperhulk:
Yep we can revert that now.

PR #3350, opened 2017-01-17
Use static callbacks with Python 3.x again

tiran:
Static callbacks were disabled for Python 3.5+ to work around an issue
with subinterpreters, locking callbacks and osrandom engine. Locking
callback and osrandom engine were replaced with a C implementations in
version 1.6 and 1.7.

https://github.com/pyca/cryptography/issues/2970

Closes: #3348


reaperhulk:
import sys needs to be removed now

tiran:
Thanks for the fix up!

PR #3354, opened 2017-01-18
make cryptography fallback to /dev/urandom on mac on macOS < 10.12 even if SYS_getentropy is in the headers

### LINKED ISSUE #3332: build from source on OS X 10.11 with Xcode installed uses missing symbol
see https://github.com/Homebrew/homebrew-core/pull/8046

summary: if you have Xcode installed on MacOS 10.11, instead of "Command Line Tools", then the toolchain includes headers that are "compatible with both 10.12 and 10.11" by including `SYS_getentropy` in `<sys/syscall.h>` but having `_getentropy` as a "weak symbol" on 10.11.

This results in "Symbol not found: _getentropy":

```
File "/usr/local/Cellar/fabric/1.13.1/libexec/vendor/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/__init__.py", line 7, in <module>
File "/usr/local/Cellar/fabric/1.13.1/libexec/vendor/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 51, in <module>
    from cryptography.hazmat.bindings._openssl import ffi as _ffi
    ImportError: dlopen(/usr/local/Cellar/fabric/1.13.1/libexec/vendor/lib/python2.7/site-packages/cryptography/hazmat/bindings/_openssl.so, 2): Symbol not found: _getentropy
    Referenced from: /usr/local/Cellar/fabric/1.13.1/libexec/vendor/lib/python2.7/site-packages/cryptography/hazmat/bindings/_openssl.so
    Expected in: /usr/lib/libSystem.B.dylib in /usr/local/Cellar/fabric/1.13.1/libexec/vendor/lib/python2.7/site-packages/cryptography/hazmat/bindings/_openssl.so
```

The temporary fix for the above homebrew build was a tiny patch to `src/_cffi_src/openssl/src/osrandom_engine.h` to force CRYPTOGRAPHY_OSRANDOM_ENGINE_GETENTROPY to not be selected on MacOS 10.11. I guess you're supposed to have something like a configure script that attempts to both build and run a test executable to detect whether getentropy() is really available ... I dunno.

reaperhulk:
This will fix #3332 

However, the correct longterm approach is to do runtime detection of whether `SYS_getentropy` is available during `osrandom_init` like we do for the `getrandom` engine. Then we could revert this check entirely, as well as be able to generate wheels that would use the getentropy path when available on mac (right now all our wheels use /dev/urandom).

@tiran would you be interested in potentially adapting the getrandom code so that we can do the same thing in getentropy?

### DIFF
--- a/src/_cffi_src/openssl/src/osrandom_engine.h
+++ b/src/_cffi_src/openssl/src/osrandom_engine.h
@@ -13,6 +13,7 @@
 
   #ifdef __APPLE__
     #include <sys/random.h>
+    #include <AvailabilityMacros.h>
   #endif
 
   #ifdef __linux__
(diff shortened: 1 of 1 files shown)


tiran:
I'm rather busy this week and I'll be at conferneces and team meetings until Feb 3rd. I cannot promise to work on a fix before Feb 6th.

reaperhulk:
@tiran that's not a problem. This fix works fine and the better solution can be in 1.9 😄

ilovezfs:
Thanks! Any idea on time table for 1.7.3 or 1.8.0?

reaperhulk:
The plan is to release 1.8 around Feb 6-10 at the moment.

PR #3364, opened 2017-01-23
Remove DH generator size constraint

mtury:
I've been working on a `cryptography` based TLS module for Scapy (see PR [https://github.com/secdev/scapy/pull/476](https://github.com/secdev/scapy/pull/476)).

During test handshakes with `openssl`, I came across this limitation on the generator size of DH groups. Default DH group parameters with `openssl` are currently 2048-bit order with a 2048-bit generator. While we could debate on the impact of choosing a long generator over a value of 2 or 5, it does not seem appropriate to keep this constraint in the library.

### DIFF
--- a/src/cryptography/hazmat/primitives/asymmetric/dh.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/dh.py
@@ -87,9 +87,6 @@ class DHParameterNumbers(object):
         if q is not None and not isinstance(q, six.integer_types):
             raise TypeError("q must be integer or None")
 
-        if q is None and g not in (2, 5):
-            raise ValueError("DH generator must be 2 or 5")
-
         self._p = p
         self._g = g
         self._q = q
--- a/src/cryptography/hazmat/primitives/asymmetric/dh.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/dh.py
@@ -87,6 +87,9 @@ class DHParameterNumbers(object):
         if q is not None and not isinstance(q, six.integer_types):
             raise TypeError("q must be integer or None")
 
+        if g < 2:
+            raise ValueError("DH generator must be 2 or greater")
+
         self._p = p
         self._g = g
         self._q = q
--- a/tests/hazmat/primitives/test_dh.py
+++ b/tests/hazmat/primitives/test_dh.py
@@ -51,6 +51,11 @@ def test_dh_parameternumbers():
             None, None
         )
 
+    with pytest.raises(ValueError):
+        dh.DHParameterNumbers(
+            65537, 1
+        )
+
     params = dh.DHParameterNumbers(
         65537, 7, 1245
     )
--- a/tests/hazmat/primitives/test_dh.py
+++ b/tests/hazmat/primitives/test_dh.py
@@ -51,11 +51,6 @@ def test_dh_parameternumbers():
             None, None
         )
 
-    with pytest.raises(ValueError):
-        dh.DHParameterNumbers(
-            65537, 7
-        )
-
     params = dh.DHParameterNumbers(
         65537, 7, 1245
     )


reaperhulk:
To my knowledge the only reason we have this constraint is that we don't have tests for anything other than generator values of 2 or 5. Are you aware of anywhere we can get tests to cover larger generator sizes, or, failing that, a reason why we shouldn't care?

It seems clear it's desirable to remove it so I'd like to be confident that our test suite can handle the case properly in a future where we add a new backend that has a bug with larger generator sizes.

alex:
Jenkins, ok to test.

mtury:
NIST provides test vectors with 512-bit generators [here](http://csrc.nist.gov/groups/STM/cavp/key-establishment.html).

Unfortunately the keys cannot be directly imported because of a call to `openssl`'s `DH_check` through `cryptography/hazmat/backends/openssl/backend.py`. On the tests I ran, it returned the codes for `DH_NOT_SUITABLE_GENERATOR` (which is to be expected for any non-{2,3,5} generator) but also `DH_CHECK_P_NOT_SAFE_PRIME`.

However, when I bypassed this, I could check that the computed shared secret was the expected one.

What do you think? Should the backend check be made unreceptive to those two codes (I did not come across any safe prime on 5 different groups)? Or should we rather not test this?

palaviv:
@mtury you can pass DH_check by setting the appropriate `q` value. I assume that in all the cases that you have with such generator provide it.
I was planning to add support for prime-order subgroup size but didn't had the time yet.

mtury:
Indeed `q` is provided with the test vectors. But supporting the subgroup order with `cryptography` requires some work, and it is somewhat distinct from this PR.

Rather than weakening the backend check, I'd favor not testing long generators for now. The generator size distinction in `openssl` only intervenes in `dh_check.c` (and `dh_gen.c`). If anything should go wrong with long generators, we can expect the existing tests with small generators to fail anyway.

New vectors could be added as soon as subgroup orders are supported.

palaviv:
First I will start working on adding the subgroup order and hopefully will have a PR ready by next week. In any case I think that this is O.K to remove the `DH_check` and instead encourage people to use `dh_parameters_supported` instead.

markrwilliams:
@palaviv `DH_check` also fails when the generator is 2 and p mod 24 = 23.  This prevents the use of the [2048-bit MODP Group from RFC-3526](https://tools.ietf.org/html/rfc3526#page-3) and thus the implementation of the `diffie-hellman-group14-sha256` SSH key exchange algorithm.

While it's possible to clear the `DH_NOT_SUITABLE_GENERATOR` bit from `DH_check`'s error codes iff g = 2 and p mod 24 = 23, it's probably not worth it.  I'd like to get Twisted's SSH implementation using this DH implementation - would it be helpful to do a PR that removed the `DH_check` in favor of `dh_parameters_supported` ?

palaviv:
I think that will be the best solution @markrwilliams. I now have problems with DH_check in #3369 and that change will make it easier.

markrwilliams:
@palaviv Hm, looking at the code it appears that `dh_parameters_supported` is implemented in terms of `DH_check`.  Do you have a different approach for this?

reaperhulk:
(We talked about this a bit in IRC and @markrwilliams is going to look at pulling in a local copy of DH_check as `Cryptography_DH_check`)

palaviv:
I looked at the logs of the IRC talk (hope I did not miss anything). I don't think that a local copy of the new `DH_check` will help in this problem. @reaperhulk have you meant to #3369 that will be a great solution to the problem there (Would you like me to do it there @markrwilliams?).
The problem here will remain with the new code as without a `q` we only accept 2,5 as generators in `DH_check`.

> @palaviv Hm, looking at the code it appears that dh_parameters_supported is implemented in terms of DH_check. Do you have a different approach for this?

I think you misunderstood me @markrwilliams. My suggestion is to remove the `DH_check` in `load_dh_private_numbers` and in the documentation encourage the user to use `dh_parameters_supported`.

It seems like there are a lot of problems with `DH_check`. We could try `DH_check` first and then use our own function if that fails. I am not a mathematics expert but finding if a number is a primitive root modulo is very slow. We could make it faster by adding a list of known `p,g` pairs from the RFC.

markrwilliams:
> My suggestion is to remove the DH_check in load_dh_private_numbers and in the documentation encourage the user to use dh_parameters_supported.

Hm, so then users would call `dh_parameters_supported` on the assumption that they have the right `q` value for their `g` and `p`?  I like the flexibility of this but don't like that users could forget to call it.  

If we did take that approach, we'd still have to backport a `DH_check` that knows about `q` for OpenSSLs < 1.0.1, right?

> It seems like there are a lot of problems with DH_check. We could try DH_check first and then use our own function if that fails. I am not a mathematics expert but finding if a number is a primitive root modulo is very slow. We could make it faster by adding a list of known p,g pairs from the RFC.

I'm definitely not even a mathematics amateur :(  OpenSSH seems to just [check](https://github.com/openssh/openssh-portable/blob/master/dh.c#L259-L281) if the [public key passes very basic sanity checks](https://github.com/openssh/openssh-portable/blob/master/dh.c#L214-L257).  The most nuanced thing there is the bit checking.  Maybe that's good enough, and we can do this in our own replacement for `DH_check`?

@reaperhulk  what do you think?

palaviv:
> If we did take that approach, we'd still have to backport a DH_check that knows about q for OpenSSLs < 1.0.1, right?

Yes. I would suggest doing that in #3369 and not in the PR. Would you like me to do that or would you like to open a PR to my branch?

Doing what OpenSSH does sounds good as a fallback. I looked in [mbedtls](https://github.com/ARMmbed/mbedtls/blob/development/library/dhm.c#L100-L119) and they do the same.

> Hm, so then users would call dh_parameters_supported on the assumption that they have the right q value for their g and p? I like the flexibility of this but don't like that users could forget to call it.

How about adding a `check` parameter that will default to `True`?

markrwilliams:
> Would you like me to do that or would you like to open a PR to my branch?

I'll be happy to do a PR against your branch tonight :)

> How about adding a check parameter that will default to True?

I like this.  Since this is a `hazmat` primitive, a secure default seems sufficient to me.

reaperhulk:
Okay, so if I understand correctly, the order here is:

* Port the `DH_check` from 1.1.0 and use that if the detected openssl version < 1.1.0d (please make this a separate PR, not a part of 3369. It will ease review)
* Support DH subgroup order (#3369).
* Remove the DH size constraint while adding the 512-bit generator tests that @mtury found.

markrwilliams:
@reaperhulk Yes, with some additions:

* Before #3369 lands, running `DH_check` should be made controllable by an argument (probably to `DHParameters.generate_private_key` and `DHPublicNumbers.public_key`) with a default that ensure it runs.  That way if you don't have `q` handy but know that the prime is legal, you can still use it.  This seems OK because it's a safe default for a potentially dangerous option inside `hazmat`.  Thoughts?
* After #3369 lands, we should allow checking the validity of the public key as [mbedtls](https://github.com/ARMmbed/mbedtls/blob/development/library/dhm.c#L100-L119) and [OpenSSH](https://github.com/openssh/openssh-portable/blob/master/dh.c#L214-L257) do.   I imagine this would also be an argument to `generate_private_key`/`public_key`.

The two together make me think we should have an `enum` representing these possible checks:

```python
private_numbers = DHPrivateNumbers(...)
private_key = private_numbers.parameters().private_key(DHCheck.CHECK_PARAMETERS)
# -or-
private_key = private_numbers.parameters().private_key(DHCheck.CHECK_PUBLIC_KEY)
```

But if that happens, it would be happen after the PR to backport `DH_check`!

markrwilliams:
#3375 is ready to go, but is hard to test without the addition of q, which will come in #3369

reaperhulk:
We're close here thanks to a bunch of DH subgroup order work landing, but this will be in 1.9 rather than 1.8.

palaviv:
@mtury, @markrwilliams any of you planning on working on this or would you like me to do this?

mtury:
I was away for a few weeks, sorry. Unfortunately I won't have time to spend on the subject before the summer. Considering your involvement on the DH implementation, it would probably be better if you handled this.

palaviv:
I am starting to work on this but I am having a little hard time understanding the use cases. As much as I can understand @mtury would like to remove the generator size constraint and @markrwilliams would like to remove (or replace) the `DH_check` from `load_dh_private_numbers`.
@mtury is this needed after the subgroup support?
@markrwilliams would you mind posting a small use case so I will be clear on what you would like to do? Do you use previously calculated private and public key?

reaperhulk:
#3755 might also be relevant since removing DH_check would make it so you could pass alternate generators without complaint.

mtury:
@palaviv Unfortunately the subgroup support is virtually of no use to TLS, as the protocol does not provide a way to send subgroup orders. I discussed this briefly in #3414.

What prevents Scapy from using the library without any hack is basically a pre-DH_check test in DHParameterNumbers `__init__`. If DH_check ends up being removed, I believe this test should be removed too.

src/cryptography/hazmat/primitives/asymmetric/dh.py
```diff
@@ -87,9 +87,6 @@ def __init__(self, p, g, q=None):
         if q is not None and not isinstance(q, six.integer_types):
             raise TypeError("q must be integer or None")
 
-        if q is None and g not in (2, 5):
-            raise ValueError("DH generator must be 2 or 5")
```
mtury: Is this not redundant with DH_check?

reaperhulk: It is, in fact, worse than DH_check since DH_check allows 2, 3, and 5. Presumably scapy wants to convert Numbers into an instance of the object though, correct? So we need to potentially remove this check and also either remove or modify DH_check?

mtury: Actually once these two lines are removed it works fine. Indeed Scapy uses instances of DHParameterNumbers, but this never calls DH_check.

As far as I can tell, for now DH_check is called only in asymmetric/dh.py for `DHPrivateNumbers` through `load_dh_private_numbers` (or in test_dh.py through `dh_parameters_supported`). So apart from the library tests it is not called when dealing with `DHParameterNumbers` or `DHPublicNumbers`.

reaperhulk: Do you convert the DHParameterNumbers to a DHParameters via `parameters`? I guess we don't validate anything on that path so you can make a parameters and then call `generate_private_key`.

mtury: Indeed we do not use DHParameterNumbers unless for immediate conversion to DHParameters. We'd be fine with the workaround you suggest. Do you see anything else to add to the PR?

reaperhulk: I'm mostly trying to reconcile the existence of `DH_check` with the fact that we don't (and can't) validate here, but g=1 would still be a very bad thing to put here, so I think `g > 1` is probably what the modified check should be if we're going to merge this.

reaperhulk:
LGTM, let's see what CI has to say :)

(I hate FFDH)

mtury:
Now an error is raised when `g < 2`.
I did not keep the `q is None` check because we never want `g` to be less than 2 anyway.

tests/hazmat/primitives/test_dh.py
```diff
@@ -51,9 +51,9 @@ def test_dh_parameternumbers():
             None, None
         )
 
-    with pytest.raises(ValueError):
+    with pytest.raises(TypeError):
```
reaperhulk: This still needs to be `ValueError`

mtury: my bad

reaperhulk:
Thanks for sticking around for so long @mtury. Could I prevail on you to submit one more quick PR to update the docs to note the new restriction on DHParameterNumbers?

reaperhulk:
Nevermind! I went ahead and put it in as #3786

mtury:
Congrats for closing on the whole FFDH saga. :)

PR #3369, opened 2017-01-26
DH subgroup order (q)

palaviv:
This PR add support for DH subgroup order. Was discussed in #3364.
The current behavior of DH stay the same unless `q` is given.


reaperhulk:
Tests failing (probably due to DH_check, haven't looked)

palaviv:
You are correct @reaperhulk. It does fail on `DH_check` but only on openssl version 1.0.1. Support for checking with `q` was added in openssl/openssl@07981709. What do you usually do in this case?

markrwilliams:
Is it a terrible idea to backport the entire `DH_check` implementation?

markrwilliams:
#3375 makes a `DH_check` that knows about q available on all versions of OpenSSL (and LibreSSL)

reaperhulk:
This can be rebased now.

palaviv:
It seems like again something fails depending on the openssl version. I didn't find what is the problem yet...

reaperhulk:
It looks like `DHparams_dup` may not duplicate `q` on OpenSSL < 1.0.2 (including LibreSSL).

palaviv:
Should I do a `Cryptography_DHparams_dup` as we did for `DH_check`?

reaperhulk:
That would be more difficult here since DHparams_dup calls some other locally defined functions and we'd have to pull those in too. One potentially workaround would be to do a get0_pqg to extract the q and then set the q on DHparams ourselves.

src/cryptography/hazmat/backends/openssl/dh.py
```diff
@@ -8,14 +8,24 @@
...
     lib = backend._lib
     ffi = backend._ffi
 
     param_cdata = lib.DHparams_dup(dh_cdata)
     backend.openssl_assert(param_cdata != ffi.NULL)
     param_cdata = ffi.gc(param_cdata, lib.DH_free)
+    if lib.OPENSSL_VERSION_NUMBER < 0x10002000:
+        # In OpenSSL versions < 1.0.2 DHparams_dup don't copy q
+        q = ffi.new("BIGNUM **")
+        lib.DH_get0_pqg(dh_cdata, ffi.NULL, q, ffi.NULL)
```
reaperhulk: the returned bignum is not retained (`get0`) so you need to do a BN_dup here:

`q_dup = lib.BN_dup(q[0])`

`BN_dup` will return null if passed null so this is safe to do without an assertion. Then you can pass `q_dup` to `DH_get0_pqg`.

src/cryptography/hazmat/backends/openssl/dh.py
```diff
@@ -8,14 +8,24 @@
...
+        lib.DH_set0_pqg(param_cdata, ffi.NULL, q[0], ffi.NULL)
```
reaperhulk: The return value here needs to be checked

reaperhulk:
You'll also need to resolve the merge conflict of course (the test failures in jenkins right now are not relevant because it's pulling old commits since the latest commit introduces a conflict)

src/cryptography/hazmat/backends/openssl/dh.py
reaperhulk: I mean set0. Oops

palaviv:
> You'll also need to resolve the merge conflict of course (the test failures in jenkins right now are not relevant because it's pulling old commits since the latest commit introduces a conflict)

Is that a feature or a bug? Anyway I fixed the merge conflicts.

I need to add some tests for serialization with q so please don't merge until I do.

src/cryptography/hazmat/backends/openssl/dh.py
palaviv: fixed

src/cryptography/hazmat/backends/openssl/dh.py
```diff
@@ -9,14 +9,26 @@
...
 
-def _dh_cdata_to_parameters(dh_cdata, backend):
+def _dh_params_dup(dh_cdata, backend):
     lib = backend._lib
     ffi = backend._ffi
 
     param_cdata = lib.DHparams_dup(dh_cdata)
     backend.openssl_assert(param_cdata != ffi.NULL)
     param_cdata = ffi.gc(param_cdata, lib.DH_free)
+    if lib.OPENSSL_VERSION_NUMBER < 0x10002000 or lib.CRYPTOGRAPHY_IS_LIBRESSL:
```
reaperhulk: This feels like the sort of thing we're eventually going to leak memory on in LibreSSL if and when BNparams_dup copies q. (We would make a new q and set it, then lose the reference to the originally duped `q`). I think the only way to make sure we never leak here even if LibreSSL changes is to do a get0 and check to see if q is null.

reaperhulk: Alternately we can remove the branch, always do a `get0`, BN_free it, and then set the new duped one. That would be simpler, albeit a bit slower on OpenSSL >= 1.0.2

palaviv: I don't think that this is a problem as `set0` free the original before it set the new value.

reaperhulk: Ah, great. Then you're right!

reaperhulk:
Do you still need to add serialization tests for q or is this ready for review?

palaviv:
Working on that right now. Please wait with the review.

palaviv:
Added the tests for the serialization. Just need to rebase after #3388 will be merged.

reaperhulk:
Conditionally bind it here and then you can do something like:
```python
dh_types = [self._lib.EVP_PKEY_DH]
if self._lib.Cryptography_HAS_EVP_PKEY_DHX:
    dh_types.append(self._lib.EVP_PKEY_DHX)
...
if type in dh_types:
    dothething()
```

palaviv:
Sounds good. I change #3388.
Should I raise a custom error message or just fall to the `UnsupportedAlgorithm("Unsupported key type.")`?

reaperhulk:
@palaviv since OpenSSL has no idea what an `EVP_PKEY_DHX` is on versions where it doesn't exist then it can't occur, correct? So there shouldn't be any new error branch, just my example if becomes the elif on line 560.

tests/hazmat/primitives/test_dh.py
```diff
@@ -18,6 +19,11 @@
 from ...utils import load_nist_vectors, load_vectors_from_file
 
 
+skip_no_dhx = pytest.mark.skipif(
```
reaperhulk: What happens if you serialize with OpenSSL that doesn't have DHX as a key type?

palaviv: I don't have OpenSSL 1.0.1 on my machine so I will try to set up an environment with OpenSSL 1.0.1 and check that.

reaperhulk: A docker container with ubuntu 14.04 will work for that purpose. Hopefully that's a relatively easy path for you!

palaviv: Thanks for the tip @reaperhulk :smile:. I tested and the serialization is done without the `q` value so when we load the serialized value we receive everything but the `q`. When we try to load a serialized X9.42 we fail due to unknown error.

reaperhulk: So really if q is present and the user attempts to serialize or deserialize we should raise some sort of unsupported exception?

palaviv: done

palaviv:
The travis build have failed on something unrelated to this PR. Could someone restart it?

tests/hazmat/primitives/test_dh.py
```diff
@@ -528,3 +673,22 @@ def test_public_bytes_pkcs1_unsupported(self, backend):
...
+
+    @pytest.mark.skipif(
+        backend._lib.Cryptography_HAS_EVP_PKEY_DHX == 1,
+        reason="Requires OpenSSL without EVP_PKEY_DHX (1.0.2-)")
+    @pytest.mark.parametrize(
+        "vector",
+        load_vectors_from_file(
+            os.path.join("asymmetric", "DH", "RFC5114.txt"),
+            load_nist_vectors))
+    def test_dh_private_serizliation_with_q_unsupported(self, backend, vector):
```
mtury: Small typo here.
Impressive work otherwise!

tests/hazmat/primitives/test_dh.py
```diff
@@ -428,6 +529,26 @@ def test_private_bytes_unsupported_encryption_type(self, backend):
                 DummyKeySerializationEncryption()
             )
 
+    @pytest.mark.skipif(
```
reaperhulk: This is an OpenSSL backend specific test so it needs to live in `tests/hazmat/backends/test_openssl`

tests/hazmat/primitives/test_dh.py
```diff
@@ -528,3 +673,22 @@ def test_public_bytes_pkcs1_unsupported(self, backend):
             key.public_bytes(
                 serialization.Encoding.PEM, serialization.PublicFormat.PKCS1
             )
+
+    @pytest.mark.skipif(
```
reaperhulk: Same

tests/hazmat/primitives/test_dh.py
reaperhulk: (typo here)

tests/hazmat/primitives/test_dh.py
palaviv: done

reaperhulk:
You mentioned in a previous comment that if a user tries to load an X.942 key on 1.0.1 they'll get an InternalError. It doesn't look like we're catching that as an unsupported key type on 1.0.1 properly yet?

palaviv:
My comment was wrong. What we actually get is `ValueError("Could not deserialize key data.")` from `_handle_key_loading_error`. Would you like me to add a test for that?

reaperhulk:
Please do, thanks

palaviv:
Apparently I was wrong again :disappointed:. We have 4 possible cases that 3 of them result in `ValueError("Could not deserialize key data.")` but 1 of them raise `UnsupportedAlgorithm(
                "Unsupported public key algorithm.",
                _Reasons.UNSUPPORTED_PUBLIC_KEY_ALGORITHM
            )`.

The errors from each case are:
1. private PEM:
`[_OpenSSLError(code=101310620, lib=6, func=158, reason=156), _OpenSSLError(code=101118070, lib=6, func=111, reason=118), _OpenSSLError(code=151498765, lib=9, func=123, reason=13)]`
2. private DER:
`[_OpenSSLError(code=218529960, lib=13, func=104, reason=168), _OpenSSLError(code=218595386, lib=13, func=120, reason=58), _OpenSSLError(code=218640442, lib=13, func=131, reason=58)]`
3. public PEM:
`[_OpenSSLError(code=151441516, lib=9, func=109, reason=108)]`
4. public DER:
`[_OpenSSLError(code=218529960, lib=13, func=104, reason=168), _OpenSSLError(code=218546234, lib=13, func=108, reason=58), _OpenSSLError(code=218640442, lib=13, func=131, reason=58)]`

Should I add new exception cases to the `_handle_key_loading_error` method? I can add the following case:
```py
        elif errors[0][1:] in (
            (
                self._lib.ERR_LIB_ASN1,
                self._lib.ASN1_F_ASN1_CHECK_TLEN,
                self._lib.ASN1_R_WRONG_TAG
            ),
        ):
            raise UnsupportedAlgorithm(
                 "Unknown tag",
                 _Reasons.UNSUPPORTED_SERIALIZATION
            )
```
That covers both DER errors.
I can't find reason 58 or 13 in asn1.h any pointer on where I can find them?

P.S. Is that a mistake in `_handle_key_loading_error` at:
```py
        elif any(
            error[1:] == (
                self._lib.ERR_LIB_EVP,
                self._lib.EVP_F_EVP_PKCS82PKEY,
                self._lib.EVP_R_UNSUPPORTED_PRIVATE_KEY_ALGORITHM
            )
            for error in errors
        ):
            raise UnsupportedAlgorithm(
                "Unsupported public key algorithm.",
                _Reasons.UNSUPPORTED_PUBLIC_KEY_ALGORITHM
            )
```
shouldn't the exception be "Unsupported **private** key algorithm."

reaperhulk:
It should probably just say unsupported key algorithm since it can be public or private.

reaperhulk:
Please do add the new branches (with comments where appropriate to explain what triggers the code path).

To find the right error codes you'll need to track them through the lib they're being raised in. You can see which lib by looking at the codes in err.h (`ERR_LIB_EVP` is 6, etc)

tests/hazmat/primitives/test_dh.py
```diff
@@ -18,6 +19,11 @@
 from ...utils import load_nist_vectors, load_vectors_from_file
 
 
+skip_no_dhx = pytest.mark.skipif(
+    backend._lib.Cryptography_HAS_EVP_PKEY_DHX == 0,
```
reaperhulk: This is an internal implementation detail of OpenSSL so we can't have our skip checks work against this. Other backends won't have this available. If we need to do a check like this we want to do it via some mechanism like `dh_params_supported` or a new backend method to determine whether subgroup order is supported.

palaviv: Done

palaviv:
@reaperhulk is there anything more needed in the PR?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1795,6 +1836,9 @@ def dh_parameters_supported(self, p, g):
 
         return codes[0] == 0
 
+    def dhx_serialization_supported(self):
```
reaperhulk: DHX is an OpenSSL-ism. We need a more generic concept for what this is for other backends that may have similar limitations. How does the serialization differ?

palaviv: How about one of the following:
1. dh_seriazliation_with_subgroup_order_supported.
2. dh_RFC5114_serialization_supported.
3. dh_X9.42_serialization_supported.

reaperhulk: dh_x942_serialization_supported would be fine

palaviv: done

reaperhulk:
🎊 thanks palaviv! Would you mind putting in a PR for a changelog entry here? Also, what do we have left to do to make #3364 possible?

palaviv:
I am on vacation and wouldn't be able to help until Sunday. There is work needed to be done in #3364 and I am not sure if I will be able to work on that until the end of the month. Sorry.

PR #3375, opened 2017-01-31
Backport DH_check from OpenSSL 1.1.0.

markrwilliams:
OpenSSL 1.0.2's DH_check considers the q parameter, allowing it
validate more generators and primes; however, OpenSSL 1.1.0's DH_check
includes code to handle errors in BN functions, so it's preferred.

See https://github.com/openssl/openssl/commit/748e85308ef4f3e672975b3604ea2d76424fa404.

### DIFF
--- a/src/_cffi_src/openssl/dh.py
+++ b/src/_cffi_src/openssl/dh.py
@@ -16,7 +16,6 @@ FUNCTIONS = """
 DH *DH_new(void);
 void DH_free(DH *);
 int DH_size(const DH *);
-int DH_check(const DH *, int *);
 int DH_check_pub_key(const DH *, const BIGNUM *, int *);
 int DH_generate_key(DH *);
 int DH_compute_key(unsigned char *, const BIGNUM *, DH *);
@@ -40,6 +39,7 @@ MACROS = """
 int DH_generate_parameters_ex(DH *, int, int, BN_GENCB *);
 DH *d2i_DHparams_bio(BIO *, DH **);
 int i2d_DHparams_bio(BIO *, DH *);
+int Cryptography_DH_check(const DH *, int *);
 """
 
 CUSTOMIZATIONS = """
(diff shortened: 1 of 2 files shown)


markrwilliams:
`DH_check` isn't in LibreSSL [2.4.0](https://github.com/libressl-portable/openbsd/tree/libressl-v2.4.4/src/lib/libcrypto), the latest stable release, while [2.5.0](https://github.com/libressl-portable/openbsd/blob/libressl-v2.5.0/src/lib/libcrypto/dh/dh_check.c#L74-L118), the latest development release, has OpenSSL 1.0.2's version.  So no matter the LibreSSL version we use our backport.

src/_cffi_src/openssl/dh.py
```diff
@@ -114,4 +114,100 @@
...
+ * Check that p is a safe prime and
+ * if g is 2, 3 or 5, check that it is a suitable generator
+ * where
+ * for 2, p mod 24 == 11
+ * for 3, p mod 12 == 5
+ * for 5, p mod 10 == 3 or 7
+ * should hold.
+ */
+
+int Cryptography_DH_check(const DH *dh, int *ret)
```
reaperhulk: For review purposes, is this from 1.1.0d? Could you add a comment stating what version it came from?

src/_cffi_src/openssl/dh.py
```diff
@@ -114,4 +114,100 @@
...
+    ok = 1;
+ err:
+    if (ctx != NULL) {
+        BN_CTX_end(ctx);
+        BN_CTX_free(ctx);
+    }
+    return (ok);
+}
+#else
+int (*Cryptography_DH_check)(const DH *dh, int *ret) = DH_check;
```
reaperhulk: Our normal path for this sort of declaration is to do:

```
int Cryptography_DH_check(const DH *dh, int *ret) {
    return DH_check(dh, ret);
}
```

While this method will obviously work fine, could you make the change for consistency?

reaperhulk:
jenkins, add to whitelist

src/_cffi_src/openssl/dh.py
markrwilliams: [GitHub says 1.1.0pre6](https://github.com/openssl/openssl/commit/748e85308ef4f3e672975b3604ea2d76424fa404) - not sure which letter release that is.  How can I find out?

src/_cffi_src/openssl/dh.py
markrwilliams: This is a lot nicer because I can put the cffi declaration in `FUNCTIONS`!

src/_cffi_src/openssl/dh.py
reaperhulk: 1.1.0pre6 is itself a release number (it was the last pre-release prior to 1.1.0). So just add a comment saying it was updated to its current form in 1.1.0pre6

reaperhulk:
Tests are failing because the new error values are not defined in older versions of OpenSSL. (e.g. constants like `DH_CHECK_INVALID_Q_VALUE`).

src/_cffi_src/openssl/dh.py
```diff
@@ -117,6 +117,10 @@
 #endif
 
 #if CRYPTOGRAPHY_OPENSSL_LESS_THAN_110 || defined(LIBRESSL_VERSION_NUMBER)
+# define DH_CHECK_Q_NOT_PRIME            0x10
```
reaperhulk: These should each be ifndef for safety since we don't want to fail with duplicate definitions if/when LibreSSL adds these.

markrwilliams:
@reaperhulk anything else I can do to help this land?

reaperhulk:
I just need to get some time to review it. Will probably have to wait until Sunday, sorry!

reaperhulk:
Confirmed this matches the implementation in OpenSSL. `DH_UNABLE_TO_CHECK_GENERATOR` looks like it will be set in any case where generator is not 2 or 5. I assume we'll be making that an allowable code in the PR that lifts the generator restriction?

PR #3383, opened 2017-02-07
enforce password must be bytes when loading PEM/DER asymmetric keys

reaperhulk:
Previously we were using an ffi.buffer on the Python string, which was allowing text implicitly, but our documentation explicitly requires bytes.

This resolves an ambiguity that causes trouble in #3382


alex:
CHANGELOG please

PR #3388, opened 2017-02-11
Add EVP_PKEY_DHX

palaviv:
Need this for #3369.


reaperhulk:
EVP_PKEY_DHX doesn't appear to exist in OpenSSL < 1.0.2

palaviv:
It seems like the support was added in openssl/openssl@afb14cda. I don't see a easy way to backport this change. What do you think should be done @reaperhulk?

reaperhulk:
You still need to add this to https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/bindings/openssl/_conditional.py

PR #3395, opened 2017-02-13
ignore IntelliJ

ofek:


palaviv:
This is for a specific IDE. You should just put this in your global .gitignore file (http://stackoverflow.com/questions/7335420/global-git-ignore)

reaperhulk:
Agreed with @palaviv

Issue #3405, opened 2017-02-18
Compile C code with ASAN/UBSAN for tests


alex:

alex:
We've never managed to get this to work. It remains a good idea, but it's not going to happen.

PR #3406, opened 2017-02-18
Refs #3405 -- use ASAN/UBSAN when compiling on travis

alex:


reaperhulk:
I guess we'll see what versions of clang/gcc have this...

alex:
@reaperhulk That was my theory!

alex:
So, the precise builders do not have it.

reaperhulk:
What if we did this in Jenkins instead now that we have Jenkinsfile?

alex:
Doing this in Jenkins is a reasonable idea.

alex:
Closing this out since this patch is not working and I haven't been making time for it.

PR #3410, opened 2017-02-20
lazily import idna

### LINKED ISSUE #3409: Lazily load idna module
The idna package is a heavy-load package. ```idna.idnadata.scripts``` takes up a lot of time and memory to import. On my system it takes about 80ms to import the module. The import increases RSS of the process by almost 20 MB. That's a lot of time and memory for a command line script.

urllib3 uses lazy import for idna already, https://github.com/shazow/urllib3/blob/master/urllib3/contrib/pyopenssl.py#L167 . Would you consider to similar approach, too?

```
$ python -m timeit -s 'import sys' -- "sys.modules.pop('idna.idnadata', None); import idna.idnadata"
10 loops, best of 3: 79.7 msec per loop
```

```
$ python
>>> import psutil, os
>>> p = psutil.Process(os.getpid())
>>> pre = p.memory_info()
>>> import idna
>>> post = p.memory_info()
>>> pre
pmem(rss=15720448, vms=212979712, shared=7745536, text=4096, lib=0, data=8302592, dirty=0)
>>> post
pmem(rss=35950592, vms=242565120, shared=7880704, text=4096, lib=0, data=30937088, dirty=0)
>>> (post.rss - pre.rss) / 1024.**2
19.29296875
```

reaperhulk:
fixes #3409

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/decode_asn1.py
+++ b/src/cryptography/hazmat/backends/openssl/decode_asn1.py
@@ -9,8 +9,6 @@ import ipaddress
 
 from email.utils import parseaddr
 
-import idna
-
 import six
 
 from six.moves import urllib_parse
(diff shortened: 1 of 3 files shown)


intgr:
You could probably be even more lazy by detecting IDNA domain names from the `xn--` substring (in decoding) or whether it has non-ASCII characters (in encoding). If these aren't present, you could encode/decode in ASCII.

reaperhulk:
Obviated by #3411

Issue #3380, opened 2017-02-06
Provide GCM tag value during finalization step when decrypting


GaretJax:
Currently the GCM cipher mode requires the tag value to be passed in to the constructor when decrypting, but its value is not used until the finalization step. This is true for both the openssl and the commoncrypto backends.

In my use case, the data is structured as in `(iv, ciphertext, tag)`, which allows to treat the encrypted payload as a stream of data. The current API design prevents to decrypt the data by linearly reading from the input and requires it to be buffered in order to first get the tag from the end.

While I know that the decrypted output shall not be used before the finalization step, it is also true that allowing the tag to be passed in during the finalization step allows to avoid having to buffer the whole payload twice.

API wise, it would make sense to allow the tag to be passed to the `finalize` method. Currently, `InvalidTag` errors are raised only during the `finalize` call, so the main user-impacting changes would be the argument validation not occurring during initialization and decryptor construction, namely:
- https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/primitives/ciphers/modes.py#L166-L170
- https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/primitives/ciphers/modes.py#L175-L176
- https://github.com/pyca/cryptography/blob/a5865d9b630df7586785b683ed145b977ad7e0c4/src/cryptography/hazmat/primitives/ciphers/base.py#L109-L113

reaperhulk:
If we want to add this capability while maintaining API compatibility we'll want to keep validation in the GCM constructor probably because current users would expect to catch the exception there. We'd just also allow `None`. However, right now the AEADCipherContext interface defines finalize as taking no arguments, so we'd need to alter that to take a tag with default `None`. Additionally, in at least the OpenSSL backend the concrete implementation of `CipherContext` and `AEADCipherContext` are the same class, so we'd probably need to split those apart.

It doesn't seem unreasonable to support this use case, but whoever decides to implement it will have a bit more work than it appears at first glance.

phi-gamma:
Please have a look at #3421. Instead of adding an optional argument to
``finalize()`` I went with an explicit alternative ``finalize_with_tag()`` which
is only available for authenticated ciphers.

It’d be great if we could have this as it would match the options of other
crypto libraries.

Issue #3373, opened 2017-01-28
cryptography needs to be installed for self tests


0-wiz-0:
When testing cryptography, it complains:
````
cd .../cryptography && py.test-3.6
Traceback (most recent call last):
  File "/usr/pkg/lib/python3.6/site-packages/_pytest/config.py", line 325, in _getconftestmodules
    return self._path2confmods[path]
KeyError: local('.../cryptography/tests')

During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/pkg/lib/python3.6/site-packages/_pytest/config.py", line 356, in _importconftest
    return self._conftestpath2mod[conftestpath]
KeyError: local('.../cryptography/tests/conftest.py')

During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/pkg/lib/python3.6/site-packages/_pytest/config.py", line 362, in _importconftest
    mod = conftestpath.pyimport()
  File "/usr/pkg/lib/python3.6/site-packages/py/_path/local.py", line 650, in pyimport
    __import__(modname)
  File "/usr/pkg/lib/python3.6/site-packages/_pytest/assertion/rewrite.py", line 216, in load_module
    py.builtin.exec_(co, mod.__dict__)
  File ".../cryptography/tests/conftest.py", line 9, in <module>
    from cryptography.hazmat.backends import _available_backends
ModuleNotFoundError: No module named 'cryptography'
ERROR: could not load .../cryptography/tests/conftest.py
````
It would be great if one could test it before installing it.

reaperhulk:
This is unlikely to change any time soon. If someone has a proposal for how to make this possible without causing a regression in our testing strategies, etc then feel free to open an issue to discuss it. We're inside a `src` dir partially to prevent Python from picking up the non-installed version because we want to test against what actually gets installed. Since Python sdist/wheels are created from what you list in your setup.py and MANIFEST.in it is possible to have bugs like failing to include files in your installed package. Without testing directly against the installed version CI will miss bugs like that.

PR #3421, opened 2017-02-27
postpone GCM authentication tag requirement until finalization

phi-gamma:
Add a .finalize_with_tag() variant of the .finalize() function of
the GCM context. At the same time, do not enforce the requirement
of supplying the tag with the mode ctor. This facilitates
streamed decryption when the MAC is appended to the ciphertext
and cannot be efficiently retrieved ahead of decryption.

According to the GCM spec (section 7.2: “Algorithm for the
Authenticated Decryption Function”), the tag itself is not needed
until the ciphertext has been decrypted.

Addresses #3380

Signed-off-by: Philipp Gesang <philipp.gesang@intra2net.com>

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/ciphers.py
+++ b/src/cryptography/hazmat/backends/openssl/ciphers.py
@@ -78,6 +78,18 @@ class _CipherContext(object):
                 len(iv_nonce), self._backend._ffi.NULL
             )
             self._backend.openssl_assert(res != 0)
+            if operation == self._DECRYPT and \
+                    self._backend.openssl_version_number() < 0x10002000:
+                if mode.tag is None:
+                    raise NotImplementedError(
+                        "delayed passing of GCM tag requires OpenSSL >= 1.0.2."
+                        " To use this feature please update OpenSSL"
+                    )
+                res = self._backend._lib.EVP_CIPHER_CTX_ctrl(
+                    ctx, self._backend._lib.EVP_CTRL_GCM_SET_TAG,
+                    len(mode.tag), mode.tag
+                )
+                self._backend.openssl_assert(res != 0)
 
         # pass key/iv
         res = self._backend._lib.EVP_CipherInit_ex(
@@ -129,7 +141,8 @@ class _CipherContext(object):
             self.update(b"")
 
         if self._operation == self._DECRYPT and \
-                isinstance(self._mode, modes.ModeWithAuthenticationTag):
+                isinstance(self._mode, modes.ModeWithAuthenticationTag) and \
+                self._backend.openssl_version_number() >= 0x10002000:
             tag = self._mode.tag
             if tag is None:
                 raise ValueError(
--- a/src/cryptography/hazmat/backends/openssl/ciphers.py
+++ b/src/cryptography/hazmat/backends/openssl/ciphers.py
@@ -78,12 +78,6 @@ class _CipherContext(object):
                 len(iv_nonce), self._backend._ffi.NULL
             )
             self._backend.openssl_assert(res != 0)
-            if operation == self._DECRYPT:
-                res = self._backend._lib.EVP_CIPHER_CTX_ctrl(
-                    ctx, self._backend._lib.EVP_CTRL_GCM_SET_TAG,
-                    len(mode.tag), mode.tag
-                )
-                self._backend.openssl_assert(res != 0)
 
         # pass key/iv
         res = self._backend._lib.EVP_CipherInit_ex(
@@ -133,6 +127,19 @@ class _CipherContext(object):
         # and is harmless for all other versions of OpenSSL.
         if isinstance(self._mode, modes.GCM):
             self.update(b"")
+
+        if self._operation == self._DECRYPT and \
+                isinstance(self._mode, modes.ModeWithAuthenticationTag):
+            tag = self._mode.tag
+            if tag is None:
+                raise ValueError(
+                    "Authentication tag must be provided when decrypting."
+                )
+            res = self._backend._lib.EVP_CIPHER_CTX_ctrl(
+                self._ctx, self._backend._lib.EVP_CTRL_GCM_SET_TAG,
+                len(tag), tag
+            )
+            self._backend.openssl_assert(res != 0)
 
         buf = self._backend._ffi.new("unsigned char[]", self._block_size_bytes)
         outlen = self._backend._ffi.new("int *")
--- a/src/cryptography/hazmat/primitives/ciphers/modes.py
+++ b/src/cryptography/hazmat/primitives/ciphers/modes.py
@@ -161,21 +161,22 @@ class GCM(object):
         # len(initialization_vector) must in [1, 2 ** 64), but it's impossible
         # to actually construct a bytes object that large, so we don't check
         # for it
-        if min_tag_length < 4:
-            raise ValueError("min_tag_length must be >= 4")
-        if tag is not None and len(tag) < min_tag_length:
-            raise ValueError(
-                "Authentication tag must be {0} bytes or longer.".format(
-                    min_tag_length)
-            )
-
         if not isinstance(initialization_vector, bytes):
             raise TypeError("initialization_vector must be bytes")
-
-        if tag is not None and not isinstance(tag, bytes):
-            raise TypeError("tag must be bytes or None")
-
         self._initialization_vector = initialization_vector
+        self._set_tag(tag, min_tag_length)
+
+    def _set_tag(self, tag, min_tag_length=16):
+        if tag is not None:
+            if not isinstance(tag, bytes):
+                raise TypeError("tag must be bytes or None")
+            if min_tag_length < 4:
+                raise ValueError("min_tag_length must be >= 4")
+            if len(tag) < min_tag_length:
+                raise ValueError(
+                    "Authentication tag must be {0} bytes or longer.".format(
+                        min_tag_length)
+                )
         self._tag = tag
 
     tag = utils.read_only_property("_tag")
--- a/docs/hazmat/primitives/symmetric-encryption.rst
+++ b/docs/hazmat/primitives/symmetric-encryption.rst
@@ -292,7 +292,10 @@ Modes
     .. danger::
 
         When using this mode you **must** not use the decrypted data until
-        :meth:`~cryptography.hazmat.primitives.ciphers.CipherContext.finalize`
+        the appropriate finalization method
+        (:meth:`~cryptography.hazmat.primitives.ciphers.CipherContext.finalize`
+        or
+        :meth:`~cryptography.hazmat.primitives.ciphers.AEADDecryptionContext.finalize_with_tag`)
         has been called. GCM provides **no** guarantees of ciphertext integrity
         until decryption is complete.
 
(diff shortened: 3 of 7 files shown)


alex:
Jenkins, ok to test.

reaperhulk:
Looks like tests are failing here. In this case due to the number of failing GCM tests every build timed out in jenkins, which is very painful for the CI system. If you could please run them locally to minimize the probability of that happening again I'd appreciate it. Thanks for working on this!

phi-gamma:
> Looks like tests are failing here. In this case due to the
> number of failing GCM tests every build timed out in jenkins,
> which is very painful for the CI system. If you could please
> run them locally to minimize the probability of that happening
> again I'd appreciate it.

Sorry for the mess. Truth is that running the test suite is quite
tedious because of this issue:
https://github.com/pyca/cryptography/issues/3373 -- for my setup
this amounts to a workflow of dist tarball → custom RPM → install
as root (!) from shared location because no sudo/su → finally run
tests.

Running with “tox” isn’t an option either because of this issue:

    During handling of the above exception, another exception occurred:
      File "/src/python/cryptography/cryptography/.tox/py36/lib/python3.6/site-packages/_pytest/config.py", line 362, in _importconftest
        mod = conftestpath.pyimport()
      File "/src/python/cryptography/cryptography/.tox/py36/lib/python3.6/site-packages/py/_path/local.py", line 662, in pyimport
        __import__(modname)
      File "/src/python/cryptography/cryptography/.tox/py36/lib/python3.6/site-packages/_pytest/assertion/rewrite.py", line 216, in load_module
        py.builtin.exec_(co, mod.__dict__)
      File "/src/python/cryptography/cryptography/tests/conftest.py", line 10, in <module>
        from cryptography.hazmat.backends.openssl import backend as openssl_backend
      File "/src/python/cryptography/cryptography/.tox/py36/lib/python3.6/site-packages/cryptography/hazmat/backends/openssl/__init__.py", line 7, in <module>
      File "/src/python/cryptography/cryptography/.tox/py36/lib/python3.6/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 48, in <module>
        from cryptography.hazmat.bindings._openssl import ffi as _ffi
    ImportError: /src/python/cryptography/cryptography/.tox/py36/lib/python3.6/site-packages/cryptography/hazmat/bindings/_openssl.abi3.so: undefined symbol: OPENSSL_sk_num
    ERROR: could not load /src/python/cryptography/cryptography/tests/conftest.py

Anyways, I’ll update the PR after the tests suite completes.

phi-gamma:
Results with ``py.test-3`` for me:

    ========================== 80159 passed, 6306 skipped in 112.67 seconds ===========================

reaperhulk:
failing test is #3424 and nothing to do with this PR now

reaperhulk:
jenkins, retest this please

src/cryptography/hazmat/backends/openssl/ciphers.py
```diff
@@ -78,7 +78,8 @@ def __init__(self, backend, cipher, mode, operation):
                 len(iv_nonce), self._backend._ffi.NULL
             )
             self._backend.openssl_assert(res != 0)
-            if operation == self._DECRYPT:
+
+            if operation == self._DECRYPT and mode.tag is not None:
```
reaperhulk: This branch shouldn't be necessary any more since the tag is being set in `finalize`, correct?

src/cryptography/hazmat/primitives/ciphers/modes.py
```diff
@@ -161,24 +161,26 @@ def __init__(self, initialization_vector, tag=None, min_tag_length=16):
...
+                raise ValueError("min_tag_length must be >= 4")
+            if len(tag) < min_tag_length:
+                raise ValueError(
+                    "Authentication tag must be {0} bytes or longer.".format(
+                        min_tag_length)
+                )
         self._tag = tag
 
-    tag = utils.read_only_property("_tag")
+    tag = property(lambda self: getattr(self, "_tag"),
```
reaperhulk: Why make this change? It looks like this is to allow `self._ctx._mode.tag = tag` to work in `finalize_with_tag` but `self._ctx._mode._tag` will work without adding a public setter here.

phi-gamma:
> Tests are failing because this also needs to be implemented on the
> commoncrypto backend. Since you're presumably not developing on a mac it's
> okay to use CI to test some things 😄

I’m afraid I won’t be of much help there as I don’t have a Mac at
my disposal and I couldn’t find a free implementation of common
crypto to test against.

> I left a few comments, but this mostly looks okay to me. We
> also need to update the docs to add the AEADDecryptionContext in
> symmetric-encryption.rst with a .. versionadded:: 1.8

Thanks for the review, it’s much appreciated. The updated PR
contains a shot at documentation.

> This branch shouldn't be necessary any more since the tag is being set in
> finalize, correct?

Indeed, fixed.

> Why make this change? It looks like this is to allow self._ctx._mode.tag =
> tag to work in finalize_with_tag but self._ctx._mode._tag will work without
> adding a public setter here.

Reverted.

reaperhulk:
The commoncrypto fix looks like this:

```diff
-        if (self._operation == self._backend._lib.kCCDecrypt and
-                not constant_time.bytes_eq(
-                    self._tag[:len(self._mode.tag)], self._mode.tag
-                )):
-            raise InvalidTag
+        if self._operation == self._backend._lib.kCCDecrypt:
+            if self._mode.tag is None:
+                raise ValueError(
+                    "Authentication tag must be provided when decrypting."
+                )
+            if not constant_time.bytes_eq(
+                self._tag[:len(self._mode.tag)], self._mode.tag
+            ):
+                raise InvalidTag
```

However it looks like there's an issue with OpenSSL 1.0.1 that may be more serious. Tests using Ubuntu 12.04 are failing right now and since the travis job is timing out we can't see exactly what's going on. The way the docs job failed might be helpful, but I suspect tracking this completely down will require spinning up a docker container to replicate locally.

phi-gamma:
> The commoncrypto fix looks like this:

Applied, thanks. Let’s see what the CI has to say about it.

> However it looks like there's an issue with OpenSSL 1.0.1 that may be more
> serious.

Isn’t the 1.0.1 series obsolete since December?

> Tests using Ubuntu 12.04 are failing right now and since the travis
> job is timing out we can't see exactly what's going on. The way the docs job
> failed might be helpful, but I suspect tracking this completely down will
> require spinning up a docker container to replicate locally.

The oldest OpenSSL I have at my disposal is a 1.0.2k in an Arch VM. Tests seem
to run just fine on that one.

phi-gamma:
> The way the docs job failed might be helpful

Investigating the docs test I noticed that the line numbers given
there don’t match those on my systems:

    Exception raised:
          File "/opt/python/2.7.9/lib/python2.7/doctest.py", line 1315, in __run
            compileflags, 1) in test.globs
          File "<doctest default[0]>", line 57, in <module>
            tag
          File "<doctest default[0]>", line 40, in decrypt
            decryptor.authenticate_additional_data(associated_data)
          File "/home/travis/build/pyca/cryptography/.tox/docs/lib/python2.7/site-packages/cryptography/hazmat/primitives/ciphers/base.py", line 241, in authenticate_additional_data
            self._ctx.authenticate_additional_data(data)

→ Line 203 here.

          File "/home/travis/build/pyca/cryptography/.tox/docs/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/ciphers.py", line 190, in authenticate_additional_data
            self._backend.openssl_assert(res != 0)

→ Line 174 here.

          File "/home/travis/build/pyca/cryptography/.tox/docs/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 108, in openssl_assert
            return binding._openssl_assert(self._lib, ok)

→ Line 146 here.

          File "/home/travis/build/pyca/cryptography/.tox/docs/lib/python2.7/site-packages/cryptography/hazmat/bindings/openssl/binding.py", line 60, in _openssl_assert
            errors_with_text

→ This one is the same.

I’m not sure what to make of this discrepancy but if I had to
guess it’s running the tests against a stale tree.

reaperhulk:
Unfortunately while 1.0.1 is no longer supported by upstream it's quite heavily used by distributions (who may or may not be doing a good job of maintenance). We won't be able to drop support for it for probably 2 years still.

The line number discrepancy is probably because CI runs tests on merge commits. If you rebase against current master they'll likely match up.

This looks suspiciously like a bug in the OpenSSL 1.0.1 that ships with Ubuntu 12.04 since other 1.0.1 based distros are unaffected (CentOS 7, Debian 7, Debian 8), but we'll need to come up with a workaround/fix to be able to merge this.

phi-gamma:
> The line number discrepancy is probably because CI runs tests on merge commits.

Indeed, sorry for the noise.

> This looks suspiciously like a bug in the OpenSSL 1.0.1 that ships with Ubuntu 12.04 since other 1.0.1 based distros are unaffected (CentOS 7, Debian 7, Debian 8), but we'll need to come up with a workaround/fix to be able to merge this.

I grabbed a dusty 1.0.1u tarball and had a peek at ``EVP_EncryptInit(3)``
– GCM mode isn’t even documented, nor is CCM. That makes it hard to
say whether the API has changed.

reaperhulk:
I grabbed an ubuntu 12.04 container (`docker run -ti ubuntu:precise /bin/bash`), installed the necessary deps, and ran the tests against your PR...and it all passed. I'm curious if adding

```diff
+addons:
+    apt:
+        packages:
+          - openssl
+          - libssl-dev
+
 install:
+    - dpkg -l
     - ./.travis/install.sh
```
to `.travis.yml` will fix this. (The `dpkg -l` line is just for debug of course).

If it does fix this then the bug was resolved in an ubuntu-specific package release but is not present by default in Travis CI's precise container environment.

phi-gamma:
Interesting result. I’ve added the suggested changes as a separate commit.

reaperhulk:
Huh, well it upgraded it and still failed (Mac builds are failing due to the dpkg command, but I don't care about that at the moment). I'm not sure what's going on here 😞

phi-gamma:
How do you suggest we proceed? Maybe disallow late passing
of the tag for OpenSSL <= 1.0.1x? As long as it’s documented
it should be fine considering it’s most likely a consequence of
running and outdated library.

phi-gamma:
UTs went all green after guarding the new code with a backend
dependent version check. On OpenSSL 1.0.1 and below,
``NotImplementedError`` is raised if any of the functionality
added by the PR is accessed.

Note that since the problem appears to require not just any
legacy OpenSSL but is peculiar to some version as distributed
with Ubuntu, the version checks are likely to trigger on systems
that don’t actually exhibit the problem. However, since the 1.0.1
series went out of support five months ago, I consider that an
acceptable limitation.

The changes are contained in a separate commit that can be easily
reverted after dropping compatibility with affected OpenSSL
versions.

src/cryptography/hazmat/backends/openssl/ciphers.py
```diff
@@ -78,12 +78,19 @@ def __init__(self, backend, cipher, mode, operation):
...
             )
             self._backend.openssl_assert(res != 0)
-            if operation == self._DECRYPT:
-                res = self._backend._lib.EVP_CIPHER_CTX_ctrl(
-                    ctx, self._backend._lib.EVP_CTRL_GCM_SET_TAG,
-                    len(mode.tag), mode.tag
-                )
-                self._backend.openssl_assert(res != 0)
+            if operation == self._DECRYPT and \
+                    self._backend.openssl_version_number() < 0x10002000:
```
reaperhulk: I don't see a bug here, but LibreSSL reports its openssl_version_number as 0x20000000. It just so happens that LibreSSL doesn't have this bug, so this works, but it still makes me uncomfortable. I don't know that I have a better answer though. Thoughts @alex?

PR #3429, opened 2017-03-06
revert a change to our exceptions for asymmetric key loading

reaperhulk:
Previously when bad data was passed we'd raise ValuError. In the DH PR (#3369) we modified it so under some scenarios (that were too broad) it would raise UnsupportedAlgorithm. Since this was unnecessary this PR reverts it.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -1209,23 +1209,6 @@ class Backend(object):
                 _Reasons.UNSUPPORTED_CIPHER
             )
 
-        elif errors[0][1:] in (
-            (
-                self._lib.ERR_LIB_ASN1,
-                self._lib.ASN1_F_ASN1_CHECK_TLEN,
-                self._lib.ASN1_R_WRONG_TAG
-            ),
-            (
-                self._lib.ERR_LIB_PEM,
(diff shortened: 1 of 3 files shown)


alex:
Conceptually this looks ok, but still a handful of failing tests.

alex:
Still failing on some builders.

reaperhulk:
The current failure is because on OpenSSL < 1.0.2 trying to load a DER encoded DH X9.42 key raises a generic error, while the PEM path raises an error that we have previously defined to raise UnsupportedAlgorithm. Trying to make both PEM and DER return the same error here will be difficult, as we need to preserve the existing exception behaviors for key loading.

If we want to continue using the existing key loading and error handling paths then we're going to be stuck with raising ValueError for PEM/DER public keys and DER private, but UnsupportedAlgorithm for PEM private. Is this acceptable or should we create a new key loading & error handling path just for these DHX keys?

alex:
Jenkins, retest this please.

reaperhulk:
Jenkins, retest this please

palaviv:
Sorry for the late comment. I would like to suggest a different approach for the problem:
1. Change `UnsupportedAlgorithm` to inherit from `ValueError`
2. Create new exception that inherit from both  `UnsupportedAlgorithm` and `ValueError`.

PR #3437, opened 2017-03-09
it's called macOS and let's document openssl 1.1.0 installation on mac

reaperhulk:


reaperhulk:
If only I got paid for my shilling.

chrisglass:
Well, that was fast :)

alex:
Still waiting on coverage here.

Issue #3408, opened 2017-02-18
Parameters serialization 


palaviv:
I would like to add the option to serialize the `DHParameters` class. As metioned by @reaperhulk  in https://github.com/pyca/cryptography/pull/3297#discussion_r96549599 I agree that this should be a different method from the key's serialization.
I would like to suggest the following API:
1. Add `parameter_bytes` method to the `DHParameters`.
2. Add `load_pem_parameters` and `load_der_parameters` methods to serialization module and backends.

The above API should fit all parameters like `DSAParameters`.

palaviv:
@reaperhulk, @alex any thoughts? Can I start working on this?

reaperhulk:
This looks like the right approach. Just to be clear, what's the common use case for serialization here?

palaviv:
The use case if for ephemeral DH. Both parties only need the same parameters (p,g,q).

reaperhulk:
👍 what standardized serialization is there? I remember in the past PR there was a bare ASN.1 sequence -- is that a thing other libraries besides OpenSSL use, or is there a more formal structure specified in an RFC or PKCS somewhere?

palaviv:
I am not familiar with a lot of other libraries other than OpenSSL but In MbedTLS there is support for the ASN.1. I am not familiar with other structures but I will implement the ASN.1 and anything else that would be needed can be added in the future.

palaviv:
This has been solved by #3504.

PR #3441, opened 2017-03-10
Adding new exported functions

thierryba:
I'm using the backend directly to create for instance DHparams files. I needed to patch cryptography this way to have access to those functions. I hope to one day have a high level API for it. In the meantime, I would love if you could consider adding those functions. I'm seeing that other ones are there and are also useful.

Thanks for a great library!

### DIFF
--- a/src/_cffi_src/openssl/dsa.py
+++ b/src/_cffi_src/openssl/dsa.py
@@ -22,6 +22,7 @@ int DSA_sign(int, const unsigned char *, int, unsigned char *, unsigned int *,
              DSA *);
 int DSA_verify(int, const unsigned char *, int, const unsigned char *, int,
                DSA *);
+DH *DSA_dup_DH(const DSA *r);
 
 /* added in 1.1.0 to access the opaque struct */
 void DSA_get0_pqg(const DSA *, const BIGNUM **, const BIGNUM **,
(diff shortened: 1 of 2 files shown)


palaviv:
Hi @thierryba would something like what I wrote in #3408 be helpful for you as high level API?

thierryba:
@palaviv: yes that would help.

reaperhulk:
To merge this (and #3408 will probably need these anyway) we'll need to do conditional bindings so that compilation doesn't fail on OpenSSL versions without DHX. We have some [docs](https://cryptography.io/en/latest/development/c-bindings/) on how to do this and you can also look in the existing bindings to find examples.

thierryba:
I'm not super familiar with openssl and it macros but I can see this in pem.h (openssl 1.0.2, which is the minimum you guys support)

# ifndef OPENSSL_NO_DH
DECLARE_PEM_rw_const(DHparams, DH)
DECLARE_PEM_write_const(DHxparams, DH)
# endif

so it looks like Dhxparams is present if DHparams is. Or am I missing something?

reaperhulk:
We support 1.0.1 :)

alex:
Jenkins, ok to test

reaperhulk:
#3485 adds one of the two bindings here. If you still need the other one would you mind opening another PR? It'd be very helpful to explain the utility of that dup function as well.

PR #3447, opened 2017-03-13
Attempt to use coverage to quantify how many of the CONDITIONAL_NAMES we really use

alex:
(again)

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/_conditional.py
+++ b/src/cryptography/hazmat/bindings/openssl/_conditional.py
@@ -8,152 +8,179 @@ from __future__ import absolute_import, division, print_function
 # we can loop over them and delete them at runtime. It will be removed when
 # cffi supports #if in cdef
 
+Cryptography_HAS_CMS = lambda: [
+    "BIO_new_CMS",
+    "i2d_CMS_bio_stream",
+    "PEM_write_bio_CMS_stream",
+    "CMS_final",
+    "CMS_sign",
+    "CMS_verify",
+    "CMS_encrypt",
+    "CMS_decrypt",
+    "CMS_add1_signer",
+    "CMS_TEXT",
+    "CMS_NOCERTS",
+    "CMS_NO_CONTENT_VERIFY",
+    "CMS_NO_ATTR_VERIFY",
+    "CMS_NOSIGS",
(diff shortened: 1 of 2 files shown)


alex:
@reaperhulk ok, it looks like our remaining untested branches are:

- `Cryptography_HAS_CMAC`
- `Cryptography_HAS_EC`
- `Cryptography_HAS_ECDH`
- `Cryptography_HAS_ECDSA`
- `Cryptography_HAS_EC_CODES`
- `Cryptography_HAS_TLSv1_2`
- `Cryptography_HAS_RELEASE_BUFFERS`
- `Cryptography_HAS_SSL_OP_MSIE_SSLV2_RSA_PADDING`
- `Cryptography_HAS_NETBSD_D1_METH`
- `Cryptography_HAS_DTLS`

Should I just send a PR removing all these branches (probably two PRs), or are there any you want to argue in favor of keeping?

reaperhulk:
I've confirmed release buffers, CMAC, and TLS1.2 should all be unnecessary at this point. I'm confused why we have SSL_OP_MSIE_SSLV2_RSA_PADDING. Presumably pyOpenSSL, but pyOpenSSL doesn't check for presence at all.

Pretty sure `NETBSD_D1_METH` is a workaround for a NetBSD specific issue so we need to keep that.

Not sure about the others yet, although I'm inclined to believe we should be safe assuming EC support at this point...

alex:
Ok, now left remaining:

- `Cryptography_HAS_EC`
- `Cryptography_HAS_ECDH`
- `Cryptography_HAS_ECDSA`
- `Cryptography_HAS_EC_CODES`
- `Cryptography_HAS_NETBSD_D1_METH`
- `Cryptography_HAS_DTLS`

I guess I'll do one more PR to remove the EC stuff, and then we'll figure out what needs to happen with the DTLS stuff.

alex:
So now we're down to `Cryptography_HAS_NETBSD_D1_METH` and `Cryptography_HAS_DTLS`. It sounds like we need NetBSD in our CI to hit the former. Do we still need the vanilla DTLS branch?

alex:
Ok, we got the data and did the necessarry cleanups. Declaring victory!

PR #3451, opened 2017-03-14
Write a memory leak test for X.509 extensions

alex:
It's currently failing, and I can't figure out from where the memory is being allocated, because my `_backtrace()` hack has a single `??` frame.


alex:
If you prefer, I can cherry-pick everything but the x.509 test itself into its own PR.

alex:
Decided to split bits out into #3457, and found a better API for part of it in the process.

PR #3459, opened 2017-03-18
Update wheel scripts to match what Jenkins has

alex:


reaperhulk:
We never completed the switch to using these scripts in jenkins, so we should pull the latest from jenkins (for both windows and macOS, which both have been updated to do compile against 1.1.0 and generate py3.6 wheels), merge that, then switch the wheel builder to clone the repo and call the scripts.

alex:
Will this need to be backported to the release branch?

reaperhulk:
Not necessarily. If we have the wheel builder clone master to run the script we'll be fine on branches. The builder pulls the sdist from pypi so it doesn't use any local checkout.

alex:
👍

Issue #3461, opened 2017-03-19
Support parsing embedded SCT extension on X.509 certificates


alex:

3lixy:
Hi, I can see in cryptography.x509.oid that it has 
``
    PRECERT_SIGNED_CERTIFICATE_TIMESTAMPS = (
        ObjectIdentifier("1.3.6.1.4.1.11129.2.4.2")
    )
``
however i am getting UnrecognizedExtension for this extension.

Its value is:
``
[...]

PR #3464, opened 2017-03-20
Refs #3461 -- added the OID for the SCT x.509 extension

alex:
Just to get everybody's brain juice flowing:

- There's a distinct OID for embedding SCTs in an OCSP response
- This OID does not appear to have an associated name, so I didn't put it in the names dict.

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -2607,6 +2607,8 @@ instances. The following common OIDs are available as constants.
 
     .. attribute:: PRECERT_SIGNED_CERTIFICATE_TIMESTAMPS
 
+        .. versionadded:: 1.9
+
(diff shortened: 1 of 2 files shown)


docs/x509/reference.rst
```diff
@@ -2605,6 +2605,10 @@ instances. The following common OIDs are available as constants.
         the ``CRLNumber`` extension type. This extension only has meaning
         for certificate revocation lists.
 
+    .. attribute:: PRECERT_SIGNED_CERTIFICATE_TIMESTAMPS
```
reaperhulk: Oops, this should have a `.. versionadded:: 1.9`

alex: Done

PR #3469, opened 2017-03-21
Adds CryptographyException as base of exception hierarchy

bjmc:
Just a minor code hygiene item. I think it's nicer if library exceptions all inherit from a common base, so application code can catch the base class.


alex:
Sorry, but this will not happen.

People should not be generically handling exceptions out of a cryptography library, they should be handled narrowly and specifically. If you are not distinguishing between `AlreadyFinalized` and `InvalidSignature` your code is definitely doing something wrong.

PR #3473, opened 2017-03-22
remove key check

### LINKED ISSUE #3472: Very slow performance of ec.generate_private_key
On Windows with Python 3.6.0:
```bash
$ python -m timeit -s "from cryptography.hazmat.backends import default_backend;from cryptography.hazmat.primitives.asymmetric.ec import SECP256K1, derive_private_key;DEFAULT_BACKEND = default_backend();SECP256K1 = SECP256K1();import os" "derive_private_key(int.from_bytes(os.urandom(32), 'big'), SECP256K1, DEFAULT_BACKEND)"
1000 loops, best of 3: 772 usec per loop
$ python -m timeit -s "from cryptography.hazmat.backends import default_backend;from cryptography.hazmat.primitives.asymmetric.ec import SECP256K1, generate_private_key;DEFAULT_BACKEND = default_backend();SECP256K1 = SECP256K1()" "generate_private_key(SECP256K1, DEFAULT_BACKEND)"
1000 loops, best of 3: 1.91 msec per loop
```

ofek:
Fixes #3472.

@lvh Is this safe? [EC_KEY_generate_key](https://github.com/openssl/openssl/blob/master/crypto/ec/ec_key.c#L185) would never generate an invalid key right?

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -1271,9 +1271,6 @@ class Backend(object):
             res = self._lib.EC_KEY_generate_key(ec_cdata)
             self.openssl_assert(res == 1)
 
-            res = self._lib.EC_KEY_check_key(ec_cdata)
-            self.openssl_assert(res == 1)
-
             evp_pkey = self._ec_cdata_to_evp_pkey(ec_cdata)
 
             return _EllipticCurvePrivateKey(self, ec_cdata, evp_pkey)


alex:
Jenkins, ok to test

reaperhulk:
OpenSSL uses [ec_key_simple_generate_key](https://github.com/openssl/openssl/blob/master/crypto/ec/ec_key.c#L203) to generate typical EC keys. It generates a private key by creating a random number between 0 and group order via `BN_rand_range` and then doing an `EC_POINT_mul` to get the public key. That seems safe but EC (especially NIST EC) makes me second guess myself a lot so I wouldn't mind a quick confirmation that this is a safe way to generate such that we don't need the validation (here's what the validation code does: [ec_key_simple_check_key](https://github.com/openssl/openssl/blob/master/crypto/ec/ec_key.c#L275))

ofek:
There is strong signalling from a few people in `##crypto` that this is fine but @lvh could you please confirm?

lvh:
Hi! Sorry,  I didn't notice the original tag and assumed it was review bot pulling me in.

Yes, this is how you generate keys. Assuming the base point is in the main group, you're fine. For NIST curves this is always trivially true (because h=1), for non-NIST curves (notably Curve25519) this is true because the base point is carefully chosen to not be in any of the small subgroups.

I'm assuming the main thing you're worried about here is that the result would be off-curve? If so, that can't happen. This assert is basically only good for cosmic rays or whatever.

alex:
@lvh thanks for confirming.

PR #3478, opened 2017-03-23
Two additional bindings for CT

alex:


reaperhulk:
Are you going to need the ability to create and push onto these stacks when we get to SCT creation?

alex:
Don't know yet, taking it one day at a time :-)

PR #3479, opened 2017-03-23
Update symmetric-encryption.rst

gdmnbt:
Import default_backend so the example works out-of-the-box.

### DIFF
--- a/docs/hazmat/primitives/symmetric-encryption.rst
+++ b/docs/hazmat/primitives/symmetric-encryption.rst
@@ -340,6 +340,8 @@ Modes
     .. testcode::
 
         import os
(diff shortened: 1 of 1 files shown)


docs/hazmat/primitives/symmetric-encryption.rst
```diff
@@ -340,6 +340,8 @@ Modes
     .. testcode::
 
         import os
+        
+        from cryptography.hazmat.backends import default_backend
```
alex: Very tiny nit, can you remove the newline between the two `cryptography` imports?

alex:
Jenkins, ok to test

PR #3482, opened 2017-03-26
Attempt to simplify the libressl checing

alex:
The basic idea here is that the libressl version number is a giant lie, so don't take it into account when using our openssl version checking macros.


reaperhulk:
👍 love this.

PR #3484, opened 2017-04-01
Don't skip DH tests when dhx unsupported and no dhx is required

palaviv:
I noticed that in commit c10aa1bc85fcd4ea35aabafd200e9e3873b8b308 I changed the skipping and made a mistake. Now tests that don't need dhx will be executed even when not supported.


alex:
@reaperhulk can you take a look at this?

reaperhulk:
I'm not a fan of having the skip check care about whether or not it finds a specific substring in the path. It seems like either adding a new arg to the parameterization (is_dhx) and checking that would be a better path. Or simply splitting the tests apart.

PR #3485, opened 2017-04-01
Add PEM_write_bio_DHxparams, d2i_DHxparams_bio, i2d_DHxparams_bio

palaviv:
This is needed for the DH params serialization.


alex:
It looks like this probably needs to be a conditional binding.

palaviv:
I will reopen this when I fix this

palaviv:
I am not sure what is wrong here... @alex, @reaperhulk could you look?

reaperhulk:
Where are the function prototypes/macro definitions for d2i_DHxparams_bio and i2d? I can't find them even in 1.0.2.

palaviv:
I define them myself if CRYPTOGRAPHY_OPENSSL_LESS_THAN_110PRE6 is true in dh.py

src/_cffi_src/openssl/dh.py
```diff
@@ -227,4 +229,17 @@
     return DH_check(dh, ret);
 }
 #endif
+
+/* These functions were added in OpenSSL 1.1.0f commit d0c50e80a8 */
+#if CRYPTOGRAPHY_OPENSSL_LESS_THAN_110PRE6
```
reaperhulk: As your comment notes, the commit in question is not in a released version of 1.1.0 (1.1.0f is not yet released). This check says that anything newer than 1.1.0-pre6 has d2i_DHxparams_bio, which is incorrect.

Rather than trying to conditionally bind this based on the addition of macros in an unreleased version of OpenSSL I'd suggest just defining Cryptography_d2i_DHxparams_bio (and i2d) and just update the comment to note that we define our own to simplify support across all versions.

palaviv: I am currently failing on building at versions which are older then 1.1.0-pre6. When 1.1.0f will be released it will contain d0c50e80a8  thus this condition is valid.
I will try configuring `Cryptography_d2i_DHxparams_bio` and see if that helps...

palaviv:
@reaperhulk could you restart the travis build please.

palaviv:
I fail here because of `redeclared as different kind of symbol` but the declaration is the same. Any ideas on what I did wrong?

reaperhulk:
You need to move the declaration from functions to macros (this error is why we put the conditional bindings in macros).

palaviv:
Thanks @reaperhulk. This PR is very messy. Would you like me to rebase or open a new branch?

reaperhulk:
We squash merge so it doesn't really matter. You could rebase and squash your commits if you really want to though 😄

palaviv:
That's OK. r4r :smile:

alex:
prima facie this looks ok to me, but I'm going to let @reaperhulk review since I haven't been following this part of the code closely

src/_cffi_src/openssl/pem.py
```diff
@@ -82,7 +82,11 @@
                                unsigned char *, int, pem_password_cb *,
                                void *);
 int PEM_write_bio_DHparams(BIO *, DH *);
+int PEM_write_bio_DHxparams(BIO *, DH *);
 """
 
 CUSTOMIZATIONS = """
+#ifndef EVP_PKEY_DHX
```
palaviv: They are already conditional on `EVP_PKEY_DHX`

alex: @reaperhulk meant in `_conditional.py`

reaperhulk: EVP_PKEY_DHX gets defined if it doesn't exist in `evp.py`. This check is in `dh.py`, which gets loaded first (mostly because we decided to list the modules mostly alphabetically), but I'm not super comfortable with the bindings being order dependent in this fashion. Maybe we need `#if !defined(EVP_PKEY_DHX) && EVP_PKEY_DHX != -1` as a guard?

palaviv: I added a guard but not exactly the one you suggested. The `!` before the `defined(EVP_PKEY_DHX)` should not be there.

reaperhulk:
(Ugh, sorry about the huge delay here)

PR #3488, opened 2017-04-04
use setuptools.parse_version in _version_check

craigahobbs:
This eliminates the dependency on packaging.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,6 @@ VECTORS_DEPENDENCY = "cryptography_vectors=={0}".format(about['__version__'])
 requirements = [
     "idna>=2.1",
     "asn1crypto>=0.21.0",
-    "packaging",
     "six>=1.4.1",
     "setuptools>=11.3",
 ]
--- a/src/cryptography/utils.py
+++ b/src/cryptography/utils.py
@@ -10,7 +10,7 @@ import inspect
 import sys
 import warnings
 
-from packaging.version import parse
+from pkg_resources import parse_version
 
 
 # Several APIs were deprecated with no specific end-of-life date because of the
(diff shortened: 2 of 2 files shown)


alex:
Jenkins, ok to test.

alex:
@dstufft can you please tell us what to do here :-)

alex:
@craigahobbs what's the motivation here? Modern `setuptools` depends on `packaging` so it's not like this adds a new dependency.

dstufft:
I think that the only benefit to this is old setuptools and possibly future setuptools if we convince jaraco to re-bundle his deps. Whether that's worth it or not I dunno, but besides that the change looks fine on the surface.

alex:
Oh. I just realized, a downside of this is `pkg_resources` is super slow to import.

dstufft:
We already use ``pkg_resources`` to iterate the list of installed backends though right?

alex:
We moved that import into the function, so it's only imported when you call `default_backends`.

alex:
`_version_check` is called at module scope, so there's no equivilant solution.

dstufft:
Ah right.

craigahobbs:
Regarding motivation, it is solely to eliminate the dependency on packaging.  Not a huge deal.  We're using Python 3.6 and use the setuptools included with Python.  Interestingly, it seems to include packaging - it imports it like this "from pkg_resources.extern import packaging" - not that I'd recommend that.

reaperhulk:
Yeah we moved to depend directly on `packaging` to avoid the import time side effects of `pkg_resources`. So we can't switch back without causing a regression. Sorry, but thanks for the contribution anyway! It's definitely non-obvious why we're doing it this way.

refs #3347 and #3190

PR #3499, opened 2017-04-09
Add latest error to FAQ

kevinburke:
I matched on the exact error string I got when compiling, so hopefully
users will land on the FAQ first when searching for the error message
(instead of e.g. closed Github issues).

Includes the installation instructions you need to link against the
Homebrew OpenSSL while installing modules from pip, if your pip/Python
aren't installed by Homebrew.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -27,6 +27,18 @@ If you are using PyPy, we do not currently ship ``cryptography`` wheels for
 PyPy. You will need to install your own copy of OpenSSL -- we recommend using
 Homebrew.
 
+Compiling ``cryptography`` on macOS produces a ``fatal error: 'openssl/opensslv.h' file not found`` error
+---------------------------------------------------------------------------------------------------------
+
+You may get this message with the newest version of pip and a version of Python
+that can't install from wheels (like PyPy). Install OpenSSL using Homebrew
+(``brew install openssl``) and then link against Homebrew's OpenSSL when
+installing cryptography by running
+
+.. code-block::
+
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -27,6 +27,18 @@ If you are using PyPy, we do not currently ship ``cryptography`` wheels for
 PyPy. You will need to install your own copy of OpenSSL -- we recommend using
 Homebrew.
 
+Compiling ``cryptography`` on macOS produces a ``fatal error: 'openssl/opensslv.h' file not found`` error
+---------------------------------------------------------------------------------------------------------
+
+You may get this message with the newest version of pip and a version of Python
+that can't install from wheels (like PyPy). Install OpenSSL using Homebrew
```
alex: PyPy can install from wheels, it's just that we don't ship any wheels for it on PyPI.

docs/faq.rst
```diff
@@ -27,6 +27,18 @@ If you are using PyPy, we do not currently ship ``cryptography`` wheels for
...
+(``brew install openssl``) and then link against Homebrew's OpenSSL when
+installing cryptography by running
```
alex: I think it might be be better to just link to the write sectino of our installation docs.

alex:
Jenkins, ok to test.

alex:
This was solved elsewhere with the additino of the second paragraph for https://cryptography.io/en/latest/faq/#compiling-cryptography-on-macos-produces-a-fatal-error-openssl-aes-h-file-not-found-error

PR #3514, opened 2017-04-22
Refs #3508 -- try switching just the test requirements

alex:


alex:
Doesn't actually work; setuptools tries to parse everything, even if it's not going to be used.

PR #3516, opened 2017-04-23
jurisdictionCountryName also must be PrintableString

alex:
This needs a test, and for someone to find the citation.


alex:
Opinions on what to do until that change is in a released version of OpenSSL?

alex:
This needs a rebase now that OpenSSL released.

Issue #3522, opened 2017-04-28
Trailing whitespace in pem test vector


kaduk:
vectors/cryptography_vectors/asymmetric/Traditional_OpenSSL_Serialization/key1.pem has a trailing space character on the last line ("-----END RSA PRIVATE KEY----- ").  This causes a failure in the openssl test suite when I attempt to apply strict whitespace checking for private key loading.  (Arguably openssl should not break compatibility like that in a major release series, so I will back out that change, but did want to point out that there is trailing whitespace here, in case it is desired to remove it.)

alex:
That vector is originally from OpenSSL's tests, so if upstream would like to remove the trailing whitespace, I have no problem doing so. Do you want to send a PR?

alex:
Err, sorry that one comes from gnutls, but the space is something we added by accident:

https://gitlab.com/gnutls/gnutls/commit/f16ef39ef0303b02d7fa590a37820440c466ce8d#1c6ed56e181259ab27d6c38f04d56eb1c90c1816_0_54

reaperhulk:
👍 to removing the trailing whitespace

PR #3523, opened 2017-04-28
Remove trailing space in vector #3522

alex:
The original source of this vector does not have a space,  we added it by accident


reaperhulk:
fixes #3522

alex:
ugh, thanks

PR #3539, opened 2017-05-11
Fixes #3538 -- Make our OpenSSL EC verifier's implementation match the API

### LINKED ISSUE #3538: Remove `return True` in OpenSSL ecdsa verify implementation
https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/backends/openssl/ec.py#L89

This isn't part of our documented API

alex:


reaperhulk:
Jenkins, retest this please

PR #3543, opened 2017-05-17
Make the code compatible with the latest sphinx release

alex:


alex:
... apparently spell check no longer believes in contractions: https://jenkins.cryptography.io/job/cryptography-pr-docs/2804/console

alex:
I assume this is an unintended consequence of the new sphinx version. @dhellmann are you seeing this / do is there a workaround, or shall we wait for a fix?

(@reaperhulk, we may want to pin sphinx, as I suspect we're going to want to get some dev work done at PyCon!)

reaperhulk:
Yeah pinning seems like a good idea for the moment.

alex:
Filed https://bitbucket.org/dhellmann/sphinxcontrib-spelling/issues/13/with-sphinx-161-contractions-result-in

reaperhulk:
What is this doing now? I don't know anything about the sphinx 1.6.1 breakage other than it appears to be extremely widespread :)

alex:
Not sure I understand the question, this code works on both 1.6.1 and <1.6.1.

Basically I inlined the contents of `make_admonition` into out code and then constant propagated.

alex:
(This code being "this PR")

reaperhulk:
Okay, makes sense. I can look at that then. We're still blocked on the spellcheck issue though, correct?

alex:
Yes, but that blocks #3545, not this.

PR #3545, opened 2017-05-17
Unpin sphinx.

alex:
(This isn't ready for merge, but this is better than a tracking issue!)


alex:
https://bitbucket.org/dhellmann/sphinxcontrib-spelling/issues/13/with-sphinx-161-contractions-result-in is the upstream issue.

alex:
FWIW, I hypothesize that https://github.com/rfk/pyenchant/pull/94 would also resolve this.

alex:
With the pyenchant release, it's no longer mad about contractions, but possessives are still a problem. There's only two of them, so we could whitelist easily.

alex:
Hmm, actually Jenkins seems to have a pretty strange error message, while Travis CI has the possessive issues.

PR #3546, opened 2017-05-18
Added a binding that will be useful for AIA chasing

alex:


src/cryptography/hazmat/bindings/openssl/_conditional.py
```diff
@@ -166,4 +166,7 @@
         "sk_SCT_value",
         "SCT_LIST_free",
     ],
+    "Cryptography_HAS_X509_STORE_CTX_GET_ISSUER": [
+        "X509_STORE_set_get_issuer",
```
reaperhulk: get_get_issuer needs to be here too

PR #3549, opened 2017-05-19
2.6 is old news, for real deprecate it

alex:
Before this is final, I'd like signoff from our major downstream:

- [ ] @bitprophet
- [x] @glyph
- [x] @hynek 
- [x] @Lukasa 
- [ ] @abadger
- [ ] @SwartzCr
- [ ] Confirm this is a full list of downstreams


Lukasa:
I think that's fine by us: I have no problem using older cryptographies if we need to.

alex:
(If any of you need data on 2.6 usage for your libraries, please ask. I'm happy to go find the data for you.)

hynek:
Getting rid of 2.6 *and* being able to point at you if people yell at me is a clear win-win.

reaperhulk:
(I suspect certbot's needs will delay deprecation further)

glyph:
Thanks for checking with us :) but we dropped 2.6 in 2015: http://labs.twistedmatrix.com/2015/12/twisted-155-released.html

bitprophet:
(I remember Alex/Paul mentioning Let's Encrypt being kind of a blocker, in person yesterday, but assuming my input is still desired...) I'm still on the fence re: dropping it in the very short term, but I plan to drop 2.6 in the medium term for sure, and like Cory I suspect I can simply pin to current stable release if necessary. So :+1:.

alex:
Yes, to propagate all the state to this issue, it looks like a high enough percent of CertBot users are on 2.6 (CentOS/RHEL 6) that until there's some resolution to that this isn't an option. https://github.com/certbot/certbot/issues/2208 seems to be the most likely resolution. I'm going to leave this PR open as a reminder :-)

bmw:
Sorry for the delayed reply from Certbot. I just saw this.

Certbot/`letsencrypt` is planning to remove Python 2.6 support in our release next month. The resolution for this is certbot/certbot#2208 as Alex linked above. If for some reason we don't resolve this issue in time for the release, we can pin an older version of `cryptography` in our installer script `certbot-auto` which is the only installation method we support for systems using Python 2.6.

reaperhulk:
Thanks for the info @bmw! We have actually pushed back the 2.6 deprecation pending knowing more from our biggest downstreams (and ansible may prevent us from moving for the foreseeable future), but glad to hear certbot is pushing forward 😄

abadger:
Sorry for the delayed response as well.

I've been trying to convince @jimi-c otherwise but, for now at least, Ansible is going to support Python-2.6 on the controller-side (cryptography is a dep controller-side) until RHEL6/CentOS6 reach the End of Production phase, November 30, 2020 (  https://access.redhat.com/support/policy/updates/errata ).

Ansible's plans wrt cryptography are that the upcoming release, Ansible-2.4, is going to support both cryptography and pycrypto as backends for our controller-side code with a preference for cryptography if both are found.  Ansible has made use of cryptography, if found, for PBKDF2 for several releases already, and it also makes optional use of paramiko which has switched from pycrypto to cryptography as its backend in recent releases. Personally I'd like to get to the point where we can drop support for pycrypto at all but that's a change that's likely to be on the same glacial time-scale as dropping Python-2.6 support.

Currently, there are rpm packages for pycrypto but not cryptography in the EPEL6 respository (an addon repository for RHEL/CentOS6). @nirik would probably be the person who decides whether to package cryptography as a dep of ansible in EPEL6 in the future or continue to use pycrypto there.

Ansible's users who pip install on CentOS/RHEL6 will start to get cryptography by default starting with Ansible-2.4.  We've provided them with an environment variable that will let them install with pycrypto as a dependency instead but, as with all non-standard setup.py invocation, it will only be the power users who notice that.

I'll let jimi-c and nirik correct me if I'm wrong about anything or answer any more in-depth questions that are needed.  Thanks for checking with us!

alex:
For now we'll maintain 2.6 support, however fair warning, I think we're unlikely to keep it all the way through 2020

reaperhulk:
Going to close this since the 2.6 deprecation continues to be indefinite.

alex:
👍

PR #3576, opened 2017-05-23
test linkcheck

reaperhulk:
Once we see the actual job work we can drop the true from the if.


reaperhulk:
This was actually working fine. Alex was wrong.

PR #3577, opened 2017-05-23
Master should target a commit

reaperhulk:


Jenkinsfile
```diff
@@ -139,21 +139,29 @@ def checkout_git(label) {
...
-    if (label.contains("windows")) {
-        bat script
-    } else {
-        sh """#!/bin/sh
-        set -xe
-        ${script}
-        """
+        checkout([
+            $class: 'GitSCM',
+            branches: [[name: '*/master']],
```
alex: Will this work on non-master branches? (e.g. release branches)

reaperhulk: I don't think so. We might be able to do the interpolation trick we were trying previously?

Jenkinsfile
```diff
@@ -137,23 +137,34 @@ def checkout_git(label) {
...
+            submoduleCfg: [],
+            userRemoteConfigs: [[
+                'url': 'https://github.com/pyca/cryptography'
+            ]]
+        ])
+        if (label.contains("windows")) {
+            bat "git rev-parse HEAD"
+        } else {
+            sh "git rev-parse HEAD"
+        }
```
alex: This needs to be pulled out of this branch. Also you'll need to `cd cryptography` before you `rev-parse`, I assume?

PR #3578, opened 2017-05-23
Close stdout and stderr when spawning a process

alex:


alex:
@stevepeak hey, there appears to be some sort of merging bug, the two lines that this claims aren't covered make no sense, they are in the middle of function with lines both before and after them covered:
screen shot 2017-05-23 at 12 14 35 pm

Also want to point out that that logic codecov is using to show which blocks to expand seems a bit confused, like it's showing me random lines of code.

stevepeak:
@reaperhulk I reviewed a number of uploads and found Codecov was accurate. These two line are found in [this upload](https://codecov.s3.amazonaws.com/v4/raw/2017-05-23/12CC0B01F9906FB96C880A8A0ACFBA30/77f44dbef586c31c81bd7b4ea8f73dabf0d141ec/3e89cb0a-c931-4a1b-86bc-7f4c4f2e4697.txt) and other python 3.5 uploads. These lines were skipped in other uploads that were not 3.5 version.

Could be an issue with `coveragepy` collection? Can you run coverage locally to confirm?

PR #3596, opened 2017-05-24
add some explicit instructions on determining key type in a cert

### LINKED ISSUE #3541: Difficult to introspect on what kind of public key type a certificate uses
I use `cryptography` to walk through PEMs as I encounter them, and one operation I do is establish the type of public key in use.

Right now, the `public_key` method returns one of several classes based on the key type, and it looks like the key classes themselves don't give a method to ask what key type it is.

To illustrate what I mean, [here's my code right now](https://github.com/18F/domain-scan/blob/0bce58cbd5570c7996e71bf1827bcf56a6dbc8b4/scanners/sslyze.py#L232-L239):

```
        if leaf_key.__class__ == cryptography.hazmat.backends.openssl.rsa._RSAPublicKey:
            leaf_key_type = "RSA"
        elif leaf_key.__class__ == cryptography.hazmat.backends.openssl.dsa._DSAPublicKey:
            leaf_key_type = "DSA"
        elif leaf_key.__class__ == cryptography.hazmat.backends.openssl.ec._EllipticCurvePublicKey:
            leaf_key_type = "ECDSA"
        else:
            leaf_key_type == str(leaf_key.__class__)
```

This is also obviously hinging on a particular backend, if you look at the class strings -- these are the class types that actually came out when I introspected on them. The classes meant to be used as an abstraction by the backends aren't in use when I actually check the class of a key.

Anyway, I'm fully aware that this is a kludge -- and if there's a way I missed to do this that avoids this kludge, I'll happily switch.

reaperhulk:
fixes #3541

### DIFF
--- a/docs/x509/tutorial.rst
+++ b/docs/x509/tutorial.rst
@@ -152,3 +152,20 @@ Then we generate the certificate itself:
 
 And now we have a private key and certificate that can be used for local
 testing.
+
+Determining Certificate/CSR Key Type
+------------------------------------
+
+Certificates and certificate signing requests can be issued with multiple
+key types. You can determine what the key type is by using ``isinstance``
+checks:
+
+.. code-block:: pycon
+
+    >>> public_key = cert.public_key()
+    >>> if isinstance(public_key, rsa.RSAPublicKey):
+    ...     # Do something RSA specific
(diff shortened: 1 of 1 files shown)


alex:
Jenkins doesn't think CSR is a word.

PR #3601, opened 2017-05-25
only build master and our release branches

reaperhulk:


.travis.yml
```diff
@@ -104,3 +104,8 @@ notifications:
             - secure: "A93qvTOlwlMK5WoEvZQ5jQ8Z4Hd0JpeO53WYt8iIJ3s/L6AubkfiN7gwhThRtPnPx7DVMenoKRMlcRg76/ICvXEViVnGgXFjsypF0CzVcIay9pPdjpZjZHP735yLfX512RtxYEdEGwi5r25Z2CEFaydhhxNwfuMxGBtLUjusix4="
         use_notice: true
         skip_join: true
+
+branches:
```
alex: Please move this section to the top of the file, and add a comment explaining what that regex is trying to capture.

.travis.yml
```diff
@@ -8,6 +8,12 @@ cache:
         - $HOME/.cache/pip
         - $HOME/ossl-1/
 
+# Only build master and the version branches (e.g. 1.7.x)
+branches:
+    only:
+        - master
+        - /\d+\.\d+\.+x/
```
alex: Oops, from the docs it looks like `branches` also applies to `tags`, so please add a regex for them.

.travis.yml
```diff
@@ -8,6 +8,14 @@ cache:
         - $HOME/.cache/pip
         - $HOME/ossl-1/
 
+# Only build master, the version branches (e.g. 1.7.x), and 
+# version tags (which are apparently considered branches by travis)
+branches:
+    only:
+        - master
+        - /\d+\.\d+\.+x/
+        - /\d+\.\d+\.*\d*/
```
alex: The last `.` should be `\.?`, not `\.*`, we don't allow arbitrary numbers, we allow one or zero. Arguable it should be `(\.\d+)?`, since if you have a dot you need some numbers.

.travis.yml
```diff
@@ -8,6 +8,14 @@ cache:
         - $HOME/.cache/pip
         - $HOME/ossl-1/
 
+# Only build master, the version branches (e.g. 1.7.x), and
+# version tags (which are apparently considered branches by travis)
+branches:
+    only:
+        - master
+        - /\d+\.\d+\.+x/
```
alex: Oh gosh, I'm so sorry, one more note, no `+` on the final `\.`

PR #3603, opened 2017-05-25
fix compilation on 1.1.0f

reaperhulk:
refs https://github.com/openssl/openssl/pull/3360#issuecomment-304051546


src/_cffi_src/openssl/asn1.py
```diff
@@ -113,6 +111,10 @@
 long ASN1_ENUMERATED_get(ASN1_ENUMERATED *);
 int i2a_ASN1_INTEGER(BIO *, ASN1_INTEGER *);
 
+/* This became const ASN1_TIME in 1.1.0f */
+ASN1_GENERALIZEDTIME *ASN1_TIME_to_generalizedtime(ASN1_TIME *,
```
levitte: Shouldn't you add the `const` then?

reaperhulk: @levitte you'd think so but for various bad reasons that would cause other problems on older builds for us.

PR #3609, opened 2017-05-25
fix libressl error/refactor some error handling

### LINKED ISSUE #3608: Fix LibreSSL error msg
After fixing the compilation error we now have a bug in key loading IIRC that needs to be resolved on LibreSSL 2.5.x. (I do not like you LibreSSL)

reaperhulk:
Using CI to see the error and then actually fixing it...

fixes #3608

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -55,6 +55,11 @@ def configs = [
     ],
     [
         label: 'docker',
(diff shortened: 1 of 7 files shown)


alex:
Coveralls results appear to be nonsensical.

PR #3610, opened 2017-05-25
[DO NOT MERGE] Fixes #3548 - add a pyproject.toml

### LINKED ISSUE #3548: Add PEP 518 pyproject.toml

alex:
Don't merge until PIP X is out.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,6 +6,8 @@ include LICENSE.APACHE
 include LICENSE.BSD
 include README.rst
 
+include pyproject.toml
+
 recursive-include docs *
 recursive-include src/_cffi_src *.py *.c *.h
 prune docs/_build
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,6 @@
+[build-system]
+requires = [
+    "setuptools",
+    "wheel",
+    "cffi >= 1.4.1; platform_python_implementation != 'PyPy'",
+]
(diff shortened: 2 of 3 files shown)


alex:
pip x comes out in a week. I'd like to make a radical proposal:

1. We drop `setup_requires` entirely from our `setup.py`; with pip x you'll get cffi installed, if you're on an older `pip` or `setuptools` you can either upgrade to pip x or manually install `cffi`
2. We make `setup.py` error cleanly if you don't have `cffi` installed, telling you about (1)
3. We delete all the wild code for disabling `setup_requires` if the `setup.py` command doesn't need it.

Reasons:

1. Encourage people to upgrade to pip x
2. Working around this is easy, just `pip install cffi>1.7` if you can't upgrade `pip` for any reason
3. Let's us delete tons of obscure code
4. Helps kill off `easy_install` once and for all, thus protecting against scenarios like "using macOS<10.13 with system python and `pip install cryptography` successfully downloads `cryptography`, but then installing it fails to download `cffi` since that download is done by `setuptools`"

reaperhulk:
Top 10 versions (including both pip and setuptools) used to install cryptography 2.2.2 (latest version) since it was released:

image

I'm baffled why 1.5.4 is so enormous. It isn't on any other PyCA project.

dstufft:
https://packages.ubuntu.com/trusty/python-pip

alex:
This is blocked on:

- [ ] https://github.com/msgpack/msgpack-python/issues/299
- [ ] https://github.com/pypa/get-pip/issues/19

alex:
Turns out pip x only supports pyproject.toml with packages distributed as wheels, so we can't use it yet:

https://github.com/pypa/pip/issues/5229

alex:
Closing in favor of #4363

PR #3615, opened 2017-05-26
various backports for 1.8.2

reaperhulk:
sigh


reaperhulk:
This backport is actually jenkinsfile + 1.1.0f fix + sphinx pin because we need that combo to have CI pass (hopefully)

PR #3621, opened 2017-05-27
re-add the why of cryptography to the faq with some small updates

### LINKED ISSUE #3617: re-add "reasons for a new cryptography library" to faq
We removed it from the landing page, but a recent talk (https://www.youtube.com/watch?v=_kj9Sn0M-hs) used that language in a slide which suggests it is not entirely irrelevant. I think we should add it back to the FAQ (with some updated language to reflect our current views).

reaperhulk:
fixes #3617

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -14,6 +14,25 @@ to NaCl.
 
 If you prefer NaCl's design, we highly recommend `PyNaCl`_.
 
+Why use ``cryptography``?
+-------------------------
+
+If you've done cryptographic work in Python before you have likely encountered
+other libraries in Python such as *M2Crypto*, *PyCrypto*, or *PyOpenSSL*. In
+building ``cryptography`` we wanted to address a few issues we observed in the
+legacy libraries:
+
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -14,6 +14,25 @@ to NaCl.
...
+* Lack of PyPy and Python 3 support.
+* Lack of maintenance.
+* Use of poor implementations of algorithms (i.e. ones with known side-channel
+  attacks).
+* Lack of high level APIs.
+* Absence of algorithms such as
+  :class:`AES-GCM <cryptography.hazmat.primitives.ciphers.modes.GCM>` and
+  :class:`~cryptography.hazmat.primitives.kdf.hkdf.HKDF`.
+* Poor introspectability and thus poor testability.
+* Extremely error prone APIs and bad defaults.
```
alex: Can we reorder these to reflect the things we really think are most important these days? e.g. "Error prone APIs" should go at the top. Bad defaults should become "insecure defaults"

PR #3627, opened 2017-05-28
Sphinx 1.6.2 is also broken

alex:


setup.py
```diff
@@ -301,7 +301,7 @@ def run_tests(self):
             "doc8",
             "pyenchant",
             "readme_renderer >= 16.0",
-            "sphinx != 1.6.1",
+            "sphinx != 1.6.1, != 1.6.2",
```
reaperhulk: Is there progress in resolving our issue upstream? And, if not, should we just do `sphinx < 1.6.1` for now?

alex: Work is ongoing. I don't really want to pin `< 1.6.1` because then we're trapped with that, and we drag anything else that wants to run our docs in the same env backwards once this is fixed.

PR #3628, opened 2017-05-29
Simplify int to hex string conversion

alex:


reaperhulk:
nice that we can drop the or branch here.

PR #3630, opened 2017-05-29
add a jenkinsfile for updating homebrew openssl on jenkins builders

reaperhulk:


.jenkins/Jenkinsfile-Update-Homebrew-OpenSSL
```diff
@@ -0,0 +1,35 @@
...
+def build(label) {
+    node(label) {
+        try {
+            timeout(time: 30, unit: 'MINUTES') {
+                stage("Compile") {
+                    sh """
+                        set -xe
+
+                        /usr/local/bin/brew update
+                        /usr/local/bin/brew reinstall openssl@1.1 --build-bottle
```
alex: If you don't `--build-bottle` something hilarious happens right?

reaperhulk: Only on platforms where there's no pre-built bottle. At the moment sierra and yosemite both have bottles, but when sierra+1 comes out yosemite bottles will stop being built so we'll need `--build-bottle` then to prevent it from building OpenSSL with CPU-specific optimizations.

.jenkins/Jenkinsfile-Update-Homebrew-OpenSSL
```diff
@@ -0,0 +1,35 @@
...
+    def label = _label
+    builders[label] = {
+        build(label)
+    }
+}
+
+parallel builders
+
+/* We'd like to trigger a downstream job, but bugs. Commented out for now */
+/* Re-enable when the exception from https://ci.cryptography.io/job/cryptography-update-brew-openssl/6/console is fixed */
```
alex: Please file a bug for this once this is merged.

PR #3631, opened 2017-05-29
move MACContext to mac.py and eliminate interfaces.py

reaperhulk:


alex:
We should be nice and leave `interfaces.py` around to re-export `MACContext` with a deprecation warning.

reaperhulk:
Test I just added will fail until https://github.com/pytest-dev/pytest/issues/2441 is fixed.

PR #3632, opened 2017-05-29
reorganize our toc to look nicer and make more sense

reaperhulk:
New layout:

image


reaperhulk:
Note that this is not alphabetical but tries to roughly order by importance within the sections. I've got another PR that will reorder things alphabetically within hazmat primitives (and improve a few section titles), but that comes after the MAC PR merges.

PR #3638, opened 2017-05-29
update to new CI server for wheels + switch to using a job token

### LINKED ISSUE #3637: Update release automation for new wheel builder
Once #3636 is merged we need to update the release automation to trigger the new wheel builder and download the artifacts.

reaperhulk:
Fixes #3637 depends on #3636 

This switches back to using a job token instead of username/token because a job token makes more sense. Unfortunately you can't POST for that so we also switch to a GET.

### DIFF
--- a/release.py
+++ b/release.py
@@ -17,7 +17,10 @@ from clint.textui.progress import Bar as ProgressBar
 import requests
 
 
-JENKINS_URL = "https://jenkins.cryptography.io/job/cryptography-wheel-builder"
+JENKINS_URL = (
+    "https://ci.cryptography.io/job/cryptography-support-jobs/"
+    "job/wheel-builder/"
+)
 
 
 def run(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


release.py
```diff
@@ -17,7 +17,10 @@
 import requests
 
 
-JENKINS_URL = "https://jenkins.cryptography.io/job/cryptography-wheel-builder"
+JENKINS_URL = (
+    "https://ci.cryptography.io/job/cryptography-support-jobs/"
+    "job/wheel-builder/"
```
alex: Nit: remove trailing slash, so we don't produce duplicate slashes in our requests.

release.py
```diff
@@ -128,14 +131,11 @@ def release(version):
     )
     response.raise_for_status()
 
-    username = getpass.getpass("Input the GitHub/Jenkins username: ")
     token = getpass.getpass("Input the Jenkins token: ")
-    response = session.post(
+    response = session.get(
```
alex: Is this serious a get...

reaperhulk: Yep, doesn't respond to POST when using tokens for some stupid reason.

alex:
Computer

PR #3643, opened 2017-05-30
Remove a binding that hasn't been used for a while

alex:


reaperhulk:
Okay, so this will break users who upgrade cryptography but have a pyOpenSSL >=0.13 but < 16.0. I'm okay with that since `pip install -U pyopenssl` will fix it.

PR #3645, opened 2017-05-30
download artifacts from new jenkins

reaperhulk:
turns out new jenkins changed its JSON a bit.

Also broken from the 1.9 release:
git tag -s
both twine calls


alex:
Please file bugs for the other two issues.

PR #3504, opened 2017-04-17
Dh parameters serialization

palaviv:
This PR add support in DH parameters serialization as discussed in #3408.

PR #3648, opened 2017-05-30
Adding one more exported function for DSA

thierryba:
Hello,
I had made a PR long ago to add 2 functions that I'm using in a product.
Essentially, I want to have a way to generate dhparams file for apache. So I looked at the code inside openssl and came up with the following python code:

def generate_dh_file():
    filename = join(fwcontrol.get_certs_folder_path(), 'dhparams.pem')
    if os.path.exists(filename):
        print("File %s already present" % filename)
        return

    print("Generating DSA parameters, 2048 bit long prime...")
    from cryptography.hazmat.backends.openssl.backend import Backend
    b = Backend()

    dsa = b.generate_dsa_parameters(2048)
    dh = b._ffi.gc(b._lib.DSA_dup_DH(dsa._dsa_cdata), b._lib.DH_free)

    q_list = b._ffi.new("BIGNUM *[1]")
    p_list = b._ffi.new("BIGNUM *[1]")
    b._lib.DH_get0_pqg(dh, p_list, q_list, b._ffi.NULL)
    func = b._lib.PEM_write_bio_DHxparams if q_list[0] == 0 else b._lib.PEM_write_bio_DHparams
    out = b._create_mem_bio_gc()
    if  func(out, dh) == 0:
        raise Exception(b._lib.ERR_GET_REASON(b._lib.ERR_get_error()))
    open(filename, 'wb').write(b._read_mem_bio(out))
    print("written file to %s" % filename)

So basically I needed to add DSA_dup_DH.
Disclaimer: I'm not a security expert and I really just copied what openssl is doing.

### DIFF
--- a/src/_cffi_src/openssl/dsa.py
+++ b/src/_cffi_src/openssl/dsa.py
@@ -23,6 +23,9 @@ int DSA_sign(int, const unsigned char *, int, unsigned char *, unsigned int *,
 int DSA_verify(int, const unsigned char *, int, const unsigned char *, int,
                DSA *);
 
+/* added in 1.0.2 */
+DH *DSA_dup_DH(const DSA *r);
+
 /* added in 1.1.0 to access the opaque struct */
 void DSA_get0_pqg(const DSA *, const BIGNUM **, const BIGNUM **,
                   const BIGNUM **);
@@ -37,6 +40,11 @@ int DSA_generate_parameters_ex(DSA *, int, unsigned char *, int,
 """
 
 CUSTOMIZATIONS = """
+/* These functions were added in OpenSSL 1.0.2 */
+#if CRYPTOGRAPHY_OPENSSL_LESS_THAN_102
+DH *(*DSA_dup_DH)(const DSA *r) = null;
+#endif
+
 /* These functions were added in OpenSSL 1.1.0-pre5 (beta2) */
 #if CRYPTOGRAPHY_OPENSSL_LESS_THAN_110PRE5
 void DSA_get0_pqg(const DSA *d,


reaperhulk:
Does the API is #3504 cover your needs? We can already generate parameters but that PR will add serialization so this would hopefully capture your use case:

```python
parameters = dh.generate_parameters(
    generator=2, key_size=2048, backend=default_backend()
)
pem = parameters.parameter_bytes(serialization.Encoding.PEM, serialization.SomeFormat)
```

palaviv:
@thierryba why do you generate DSA params and not DH? Where in openssl they do this?

thierryba:
@palaviv They do it here https://github.com/openssl/openssl/blob/master/apps/dhparam.c

@reaperhulk That Api would suffice but using DSA is definitely faster and even if it is something we do rarely, we appreciate the speed of it.

reaperhulk:
I think I'm still not fully understanding this use case then. `DSA_dup_DH` doesn't seem like it should be faster unless you've already generated the DSA key for some other purpose besides DH kex?

thierryba:
@reaperhulk Like I said, I'm no expert. I can only say that
openssl dhparam -dsaparam 2048 is much faster than openssl dhparam 2048

form the dhparam docs (dsaparam docs):
If this option is used, DSA rather than DH parameters are read or created; they are converted to DH format. Otherwise, "strong" primes (such that (p-1)/2 is also prime) will be used for DH parameter generation. DH parameter generation with the -dsaparam option is much *faster*, and the recommended exponent length is shorter, which makes DH key exchange more efficient. Beware that with such DSA-style DH parameters, a fresh DH key should be created for each use to avoid small-subgroup attacks that may be possible otherwise.

reaperhulk:
Ah, I see. Yes, DSA param gen is much, much cheaper than DH param gen. With the new API you can accomplish this with:

```python
dsa_param = dsa.generate_parameters(2048, backend)
num = dsa_param.parameter_numbers()
dhparam = dh.DHParameterNumbers(p=num.p, q=num.q, g=num.g).parameters(backend)
# And the next call hasn't landed yet but should be in the next release
pem = dhparam.parameter_bytes(serialization.Encoding.PEM, serialization.SomeFormat)
```

thierryba:
@reaperhulk Cool. Looking forward to having that api ready.

PR #3652, opened 2017-05-31
FreeBSD 10 and higher use Fortuna, not Yarrow

ashemedai:
See https://www.freebsd.org/cgi/man.cgi?query=random&apropos=0&sektion=4&manpath=FreeBSD+12-current&arch=default&format=html#end for background info.

### DIFF
--- a/docs/hazmat/backends/openssl.rst
+++ b/docs/hazmat/backends/openssl.rst
@@ -92,8 +92,9 @@ When importing only the binding it is added to the engine list but
 OS random sources
 -----------------
 
-On macOS and FreeBSD ``/dev/urandom`` is an alias for ``/dev/random`` and
-utilizes the `Yarrow`_ algorithm.
+On macOS and FreeBSD ``/dev/urandom`` is an alias for ``/dev/random``. The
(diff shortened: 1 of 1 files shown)


docs/hazmat/backends/openssl.rst
```diff
@@ -93,7 +93,8 @@ OS random sources
 -----------------
 
 On macOS and FreeBSD ``/dev/urandom`` is an alias for ``/dev/random`` and
-utilizes the `Yarrow`_ algorithm.
+utilizes the `Yarrow`_ algorithm. FreeBSD version 10 and higher use the
+`Fortuna`_ algorithm.
```
alex: According to our docs, we only support FreeBSD 11 and up, so let's simplify this to saying macOS is Yarrow and FreeBSD is Fortuna

alex:
One small note, thanks for the PR!

alex:
@ashemedai you might need to push an empty commit or something; Travis seems confused.

ashemedai:
Jenkins error caused by: https://github.com/pypa/setuptools/issues/1042

alex:
Coverage results are manifestly impossible. Docs only so I'm merging.

PR #3658, opened 2017-06-02
make signature and verification contexts error better re: prehashed

### LINKED ISSUE #3656: Attempting to use a prehashed algorithm with ECDSA fails
```python
private_key = ec.generate_private_key(ec.SECP256R1(), default_backend())
signer = private_key.signer(ec.ECDSA(utils.Prehashed(hashes.SHA256())))
signer.update(preHashedBinary)
finalSignature = signer.finalize()
```

yields the error 

```
  File "utility.py", line 189, in signImage
    signer = private_key.signer(ec.ECDSA(utils.Prehashed(hashes.SHA256())))
  File "C:\Python27\lib\site-packages\cryptography\hazmat\backends\openssl\ec.py
", line 145, in signer
    self._backend, self, signature_algorithm.algorithm
  File "C:\Python27\lib\site-packages\cryptography\hazmat\backends\openssl\ec.py
", line 96, in __init__
    self._digest = hashes.Hash(algorithm, backend)
  File "C:\Python27\lib\site-packages\cryptography\hazmat\primitives\hashes.py",
 line 76, in __init__
    raise TypeError("Expected instance of hashes.HashAlgorithm.")
TypeError: Expected instance of hashes.HashAlgorithm.
```

From looking at the docs, signer should take an EllipticCurveSignatureAlgorithm. An EllipticCurveSignatureAlgorithm should be able to be prehashed. 

Additionally, this page: https://cryptography.io/en/latest/hazmat/primitives/asymmetric/utils/#cryptography.hazmat.primitives.asymmetric.utils.Prehashed

Doesn't have an example of how to use a Prehashed function with an ECDSA. If I'm attempting to use the Pre-hashed function incorrectly, I'd be more than willing to add some documentation on that function once I figure out what I'm doing wrong.

reaperhulk:
Fixes #3656

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/dsa.py
+++ b/src/cryptography/hazmat/backends/openssl/dsa.py
@@ -11,7 +11,8 @@ from cryptography.hazmat.backends.openssl.utils import (
 )
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import (
-    AsymmetricSignatureContext, AsymmetricVerificationContext, dsa
+    AsymmetricSignatureContext, AsymmetricVerificationContext, dsa,
+    utils as asym_utils
 )
 
 
@@ -41,6 +42,14 @@ def _dsa_sig_verify(backend, public_key, signature, data):
     if res != 1:
         backend._consume_errors()
         raise InvalidSignature
+
+
+def _check_not_prehashed(signature_algorithm):
+    if isinstance(signature_algorithm, asym_utils.Prehashed):
+        raise TypeError(
+            "Prehashed is only supported in the sign and verify methods. "
+            "It cannot be used with the signer or verifier context."
+        )
 
 
 @utils.register_interface(AsymmetricVerificationContext)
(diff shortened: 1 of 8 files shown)


src/cryptography/hazmat/backends/openssl/dsa.py
```diff
@@ -43,6 +44,14 @@ def _dsa_sig_verify(backend, public_key, signature, data):
         raise InvalidSignature
 
 
+def _check_not_prehashed(signature_algorithm):
+    if isinstance(signature_algorithm, asym_utils.Prehashed):
+        raise TypeError(
+            "Prehashed is only supported in the sign and verify methods. "
+            "It cannot be used with the signer or verifier context."
```
alex: I would simplify this to "It cannot be used with signer or verifier"

src/cryptography/hazmat/backends/openssl/ec.py
```diff
@@ -24,6 +25,14 @@ def _check_signature_algorithm(signature_algorithm):
             _Reasons.UNSUPPORTED_PUBLIC_KEY_ALGORITHM)
 
 
+def _check_not_prehashed(signature_algorithm):
```
alex: Ok, this function is identical, let's put this somewhere we can reuse it.

src/cryptography/hazmat/backends/openssl/rsa.py
```diff
@@ -25,6 +26,14 @@
 )
 
 
+def _check_not_prehashed(signature_algorithm):
+    if isinstance(signature_algorithm, asym_utils.Prehashed):
+        raise TypeError(
```
alex: Same

PR #3660, opened 2017-06-03
Fixes #3653 -- attempt to retry git clones up to 3 times

### LINKED ISSUE #3653: Retry failed git clones in jenkins

alex:


alex:
@stevepeak this is another case of codecov not showing the contents of the file at all.

stevepeak:
Thanks for poking :) I'm on it.

stevepeak:
*Update* found the issue. In [this upload](https://codecov.s3.amazonaws.com/v4/raw/2017-06-03/12CC0B01F9906FB96C880A8A0ACFBA30/fff7e19f06cc3cb9ae65df4b85d40aa5669af786/ae97dec3-e969-4b78-8cec-7efae806e346.txt) which is created by Jenkins-py34-windows64 *only* there is a file in the report called simply `utils.py` which is being resolved to `asymmetric/utils.py` therefore adding hundreds of extra lines to the report.

stevepeak:
We need to figure out a way to remove that file from the report. I'm not sure why it's there in the  first place.

reaperhulk:
That is super weird. We do have a lot of `utils.py` files, but there's nothing special about that build that should have more than any other report 😟

alex:
Right, all of our builds invoke coverage.py in the exact same way... interesting!

alex:
So, that `utils.py` is almost certainly `./tests/utils.py`, based on the number of lines.

alex:
And it looks like none of the files in `tests/` have that prefix on their `filename` attribute in the XML. This seems like a probable source of our troubles.

At a guess, codecov.io's resolution of bare-filename to exact path isn't stable, probably due to dictionary ordering or something. That explains why we only see this problem sometimes.

The fix is going to be to get the `filename` attribute to include the `test/` prefix I think.

stevepeak:
I'm happy to suggest ways to resolve this in the `codecov.yml` if you need help 👍

alex:
Would love to hear any suggestions!

I'm currently seeing if I can reproduce any part of this hypothesis chain locally :-)

alex:
I'm very confused, running `tox -e py34 -- tests/test_x509.py` produces `coverage` output with the full pathname.

alex:
Looking at https://ci.cryptography.io/blue/organizations/jenkins/pyca%2Fcryptography/detail/PR-3660/1/pipeline/42, the `coverage` output shows the full paths here, so somethign is getting screwed up in the `coverage.py` -> `codecov` translation I think?

alex:
Hah, `coverage xml -i` is where it happens!

alex:
https://bitbucket.org/ned/coveragepy/issues/578/incomplete-file-path-in-xml-report

stevepeak:
Ok. I'm about to be afk, but an idea I had is this:

```
fixes:
  - "::tests/"
  - "tests/src/cryptography/::src/cryptography/"
```

This will move **everything** to parent folder `tests/` then move `tests/src/cryptography/` back to  `src/cryptography/`. Hacky but should work while coveragepy is being resolved.

alex:
That's creative! I'll send a PR with it.

PR #3661, opened 2017-06-03
Attempted workaround for intermittent coverage issues

alex:
The root cause is https://bitbucket.org/ned/coveragepy/issues/578/incomplete-file-path-in-xml-report


alex:
Another fix might be to pin to `<4.4`.

PR #3663, opened 2017-06-03
deprecate signer/verifier on asymmetric keys

### LINKED ISSUE #3659: Deprecate Asymmetric Signature and Verification Contexts
I think we should deprecate these contexts (probably an extended deprecation cycle since it's likely that these APIs are used quite a bit) and steer people exclusively to using the one-shot sign/verify APIs. As part of this we can expand the documentation on Prehashed to show an example of how you can incrementally hash a large file and then pass that hash to the one-shot APIs. Hooray composition.

refs #3656 and #3658

reaperhulk:
fixes #3659


docs/hazmat/primitives/asymmetric/dsa.rst
```diff
@@ -106,26 +114,36 @@ You can get a public key object with
...
 ``verify()`` will raise an :class:`~cryptography.exceptions.InvalidSignature`
 exception if the signature isn't valid.
 
+If your data is too large to be passed in a single call, you can hash it
+separately and pass that value using
+:class:`~cryptography.hazmat.primitives.asymmetric.utils.Prehashed`.
+
+.. doctest::
+
+    >>> public_key = private_key.public_key()
```
alex: Drop this assigment, we already have a `public_key`

docs/hazmat/primitives/asymmetric/ec.rst
```diff
@@ -77,20 +68,52 @@ Elliptic Curve Signature Algorithms
...
     If the signature is not valid, an
     :class:`~cryptography.exceptions.InvalidSignature` exception will be raised.
 
+    If your data is too large to be passed in a single call, you can hash it
+    separately and pass that value using
+    :class:`~cryptography.hazmat.primitives.asymmetric.utils.Prehashed`.
+
+    .. doctest::
+
+        >>> public_key = private_key.public_key()
```
alex: same

docs/hazmat/primitives/asymmetric/rsa.rst
```diff
@@ -190,32 +195,39 @@ a public key to use in verification using
...
 :class:`~cryptography.exceptions.InvalidSignature` exception.
 
-There is a shortcut to verify sufficiently short messages directly:
+If your data is too large to be passed in a single call, you can hash it
+separately and pass that value using
+:class:`~cryptography.hazmat.primitives.asymmetric.utils.Prehashed`.
 
 .. doctest::
 
+    >>> public_key = private_key.public_key()
```
alex: once more, with feelign!

src/cryptography/utils.py
```diff
@@ -16,6 +16,10 @@
 # cycle ends.
 PersistentlyDeprecated = DeprecationWarning
 DeprecatedIn19 = DeprecationWarning
+# Asymmetric signature and verification contexts were deprecated in 2.0
+# However they are widely used and we will obey a longer than normal
+# deprecation cycle for them.
+PersistentlyDeprecatedIn20 = PendingDeprecationWarning
```
alex: Is there a reason not to use the same `PersistentlyDeprecated` class, or is your expectation that we'll migrate to them once this switches from `Pending` to regular?

reaperhulk:
I don't know how to write a test that catches this deprecation warning actually. It happens once and then is suppressed after that, but we also run a random test order job so we can't introduce an order dependent test. `warnings.simplefilter("always", PendingDeprecationWarning)` doesn't appear to work either. I assume I'm missing something -- ideas?

alex:
I'm very confused, I would have expected the code as written to work.

reaperhulk:
I can't reproduce this except in our tests (which I got down to two calls to `signer`). I have no idea what's going on.

alex:
Meaning if you copy the tests out of our overall suite, into it's own suite, you still can't reproduce?

reaperhulk:
This reproduces it in its own file:

```python
import pytest

from cryptography.hazmat.primitives.asymmetric import padding, rsa


def test_signer_deprecated():
    private_key = rsa.generate_private_key(65537, 512, backend)
    private_key.signer(
        padding.PKCS1v15(),
        hashes.SHA1()
    )


def test_signer_check_deprecated():
    private_key = rsa.generate_private_key(65537, 512, backend)
    with pytest.deprecated_call():
        private_key.signer(
            padding.PKCS1v15(),
            hashes.SHA1()
        )
```

reaperhulk:
Notably if you put the contextmanager around the first call then it all works.

alex:
Interesting! And what happens if you replace `.signer()` with just a function that does nothing but call `warnings.warn`?

reaperhulk:
It works then:

```python
import warnings

import pytest

from cryptography.hazmat.primitives.asymmetric import padding, rsa


def warn_me():
    warnings.warn("deprecated", PendingDeprecationWarning, stacklevel=2)


def test_signer_deprecated():
    private_key = rsa.generate_private_key(65537, 512, backend)
    private_key.signer(
        padding.PKCS1v15(),
        hashes.SHA1()
    )


def test_signer_check_deprecated():
    private_key = rsa.generate_private_key(65537, 512, backend)
    with pytest.deprecated_call():
        private_key.signer(
            padding.PKCS1v15(),
            hashes.SHA1()
        )


def test_warn_me():
    warn_me()


def test_warn_me_check_deprecated():
    with pytest.deprecated_call():
        warn_me()
```

So the output of that set of tests is: `.F..`

alex:
wat.

Does this mean that somewhere in the test suite there's a `signer` or `verifier` call that isn't marked as deprecated_call?

reaperhulk:
There's a ton of those actually (we call signer/verifier...a lot). But here's a minimal reproducer!

```python
import warnings

import pytest


def warn_me():
    warnings.warn("deprecated", PendingDeprecationWarning, stacklevel=2)


def i_call_warn_me():
    warn_me()


def test_warn_me():
    i_call_warn_me()


def test_warn_me_check_deprecated():
    with pytest.deprecated_call():
        i_call_warn_me()
```

alex:
So.... wtf. File a pytest bug I guess?

reaperhulk:
https://github.com/pytest-dev/pytest/issues/2469

src/cryptography/utils.py
reaperhulk: My main thinking was that it's useful to know at what point we persistently deprecated something. By grouping the persistent deprecations we have to look at the repository history or changelog to know how long it's been. Having written those words I think I'm going to just use `PersistentlyDeprecated` though, because looking at the changelog is not a burden.

PR #3665, opened 2017-06-03
Switched our backend to be a normal fixture in tests

alex:


alex:
I'm reasonably certain that `check_backend_support` can be folded into this as well, but one thing at a time.

PR #3678, opened 2017-06-05
when you :%s/\:/=/g sometimes there's a colon you shouldn't have touched

reaperhulk:
I had this fixed on another branch but didn't port it back to the vectors PR.


alex:
Ugh, not sure how I missed this in review.

PR #3686, opened 2017-06-06
X25519 Support

### LINKED ISSUE #3590: X25519 support
This is separate from #2968 so I can put it in it's own milestone.

Proposed API:

```py3
class X25519PublicKey(object):
    @classmethod
    def from_public_bytes(cls, bytes: bytes) -> X25519PublicKey:
        pass

    def public_bytes(self) -> bytes:
        pass
    

class X25519PrivateKey(object):
    @classmethod
    def generate(cls) -> X25519PrivateKey:
        pass

    def public_key(self) -> X25519PublicKey:
        pass

    @classmethod
    def _from_private_bytes(cls, bytes: bytes) -> X25519PrivateKey:
        pass

    def private_bytes(self) -> bytes:
        pass

    def exchange(self, public_key: X25519PublicKey) -> bytes:
        pass
```

reaperhulk:
depends on #3685 and #3684 and fixes #3590

### DIFF
--- a/docs/hazmat/primitives/asymmetric/x25519.rst
+++ b/docs/hazmat/primitives/asymmetric/x25519.rst
@@ -68,6 +68,14 @@ Key interfaces
 
         :returns: :class:`X25519PublicKey`
 
+        .. doctest::
+
+            >>> from cryptography.hazmat.primitives.asymmetric import x25519
+            >>> private_key = x25519.X25519PrivateKey.generate()
+            >>> public_key = private_key.public_key()
+            >>> public_bytes = public_key.public_bytes()
+            >>> loaded_public_key = x25519.X25519PublicKey.from_public_bytes(public_bytes)
+
     .. method:: public_bytes()
 
         :returns bytes: The raw bytes of the public key.
--- a/docs/hazmat/primitives/asymmetric/x25519.rst
+++ b/docs/hazmat/primitives/asymmetric/x25519.rst
@@ -42,6 +42,8 @@ Key interfaces
 
 .. class:: X25519PrivateKey
 
+    TODO: Document the other terminology used for X25519 private components.
+
     .. versionadded:: 2.0
 
     .. classmethod:: generate()
@@ -63,17 +65,19 @@ Key interfaces
 
 .. class:: X25519PublicKey
 
+    TODO: Document the other terminology used for X25519 public components.
+
     .. versionadded:: 2.0
 
     .. classmethod:: from_public_bytes(data)
 
-        :param bytes data: TODO WHAT IS THIS
+        :param bytes data: 32 byte public key.
 
         :returns: :class:`X25519PublicKey`
 
     .. method:: public_bytes()
 
-        :returns: TODO
+        :returns bytes: The raw bytes of the public key.
 
 
 .. _`Diffie-Hellman key exchange`: https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange
--- a/docs/hazmat/primitives/asymmetric/x25519.rst
+++ b/docs/hazmat/primitives/asymmetric/x25519.rst
@@ -10,9 +10,6 @@ X25519 is an elliptic curve `Diffie-Hellman key exchange`_ using `Curve25519`_.
 It allows two parties to jointly agree on a shared secret using an insecure
 channel.
 
-TODO: this is an elliptic curve (and a good one), should we mention it on
-the ec docs page?
-
 
 Exchange Algorithm
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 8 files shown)


docs/hazmat/primitives/asymmetric/x25519.rst
```diff
@@ -0,0 +1,84 @@
...
+
+        Generate an X25519 private key.
+
+        :returns: :class:`X25519PrivateKey`
+
+    .. method:: public_key()
+
+        :returns: :class:`X25519PublicKey`
+
+    .. method:: private_bytes()
```
alex: I thought we agreed that `private_bytes` wasn't needed?

reaperhulk: We did! And everywhere but the docs I removed it. Missed this one!

docs/hazmat/primitives/asymmetric/x25519.rst
```diff
@@ -0,0 +1,77 @@
...
+        :param X25519PublicKey peer_public_key: The public key for the
+            peer.
+
+        :returns bytes: A shared key.
+
+.. class:: X25519PublicKey
+
+    .. versionadded:: 2.0
+
+    .. classmethod:: from_public_bytes(data)
```
alex: Let's add an example of generating a private key, and then round tripping its public key.

src/cryptography/exceptions.py
```diff
@@ -19,6 +19,7 @@ class _Reasons(Enum):
     UNSUPPORTED_X509 = 8
     UNSUPPORTED_EXCHANGE_ALGORITHM = 9
     UNSUPPORTED_DIFFIE_HELLMAN = 10
+    UNSUPPORTED_X25519 = 11
```
alex: Why not just use one of the two existing relevant reasons?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1772,6 +1775,50 @@ def x509_name_bytes(self, name):
...
+        return _X25519PublicKey(self, evp_pkey)
+
+    def x25519_load_private_bytes(self, data):
+        # OpenSSL only has facilities for loading PKCS8 formatted private
+        # keys using the algorithm identifiers specified in
+        # https://tools.ietf.org/html/draft-ietf-curdle-pkix-03.
+        # This is the standard PKCS8 prefix for a 32 byte X25519 key.
+        # The form is:
+        # (SEQUENCE INTEGER ALGORITHMIDENTIFIER OCTETSTRING)
+        pkcs8_prefix = b'0.\x02\x01\x000\x05\x06\x03+en\x04"\x04 '
```
alex: This could probably use a longer comment, expanding on the ASN.1 parsing. Not everyone likes to parse it by hand (although I did)

tests/hazmat/primitives/test_x25519.py
```diff
@@ -0,0 +1,120 @@
...
+                ),
+                binascii.unhexlify(
+                    b"de9edb7d7b7dc1b4d35b61c2ece435373f8343c85b78674dadfc7e1"
+                    b"46f882b4f"
+                )
+            ]
+        ]
+    )
+    def test_public_bytes(self, private_bytes, public_bytes, backend):
+            private_key = X25519PrivateKey._from_private_bytes(private_bytes)
```
alex: overindent

tests/hazmat/primitives/test_x25519.py
```diff
@@ -0,0 +1,120 @@
...
+            [
+                binascii.unhexlify(
+                    b"77076d0a7318a57d3c16c17251b26645df4c2f87ebc0992ab177fba"
+                    b"51db92c2a"
+                ),
+                binascii.unhexlify(
+                    b"8520f0098930a754748b7ddcb43ef75a0dbf3a0d26381af4eba4a98"
+                    b"eaa9b4e6a"
+                )
+            ],
```
alex: This should be a tuple.

evilaliv3:
@reaperhulk: is there a reason why it was not designed the API for serializing private bytes?

Is there any internal that helps in case this is needed?

reaperhulk:
@evilaliv3 https://github.com/pyca/cryptography/issues/4386 exists to track that issue. If you wouldn't mind writing up your use case that will help us when deciding how to prioritize our dev efforts.

evilaliv3:
Actually many software implementations are using the key for more than just
key exchange (ecdh) but as well for encryption. This includes openssh and
openpgp.js for example as an alternative for NIST keys.

Is there a specific consideration you specifically implemented it only for
key exchange?

We are evaluating it in globaleaks where to implement a protocol similar to
WhatsApp/signal that also uses and store keys generated with this curve

evilaliv3:
Exactly, thanks!

In the meantime is there any way using internals of python-cryptography to
reach and export the key?

reaperhulk:
There's no way to export right now without using the C bindings themselves unfortunately. You'd need to use `i2d_PKCS8PrivateKey_bio` or `PEM_write_bio_PKCS8PrivateKey`.

PR #3689, opened 2017-06-06
compile all our mac builds against 1.1.0 in travis

reaperhulk:


.travis/install.sh
```diff
@@ -7,7 +7,7 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
     sw_vers
     brew update || brew update
 
-    brew outdated openssl || brew upgrade openssl
+    brew outdated openssl@1.1 || brew upgrade openssl@1.1
```
alex: `brew outdated` returns `0` if the argument isn't installed. I think you can just make this `brew install openssl@1.1`, it's not installed in the base image ATM.

.travis/install.sh
```diff
@@ -7,7 +7,7 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
     sw_vers
     brew update || brew update
 
-    brew outdated openssl || brew upgrade openssl
+    brew install openssl@1.1
```
alex: If we're using OpenSSL 1.1.0 AND 1.0.2, we need to upgrade and install.

reaperhulk: Yeah I'll re-add it once I see it all work (which it looks like it's going to). PS linkers are the worst.

PR #3699, opened 2017-06-11
implement CCM vector parser

reaperhulk:
Parsers are the worst. Unfortunately we can't use the existing ones because NIST has no consistency and CCM is different. In fact they're inconsistent even within the CCM vectors themselves!

Anyway, this writes a loader and tests the variants.


alex:
I hate everything. Who the hell came up with this gibberish format.

PR #3700, opened 2017-06-12
AES CCM support

reaperhulk:
Depends on #3698 and #3699. WIP until those merge.

### DIFF
--- a/docs/hazmat/primitives/aead.rst
+++ b/docs/hazmat/primitives/aead.rst
@@ -77,3 +77,91 @@ also support providing integrity for associated data which is not encrypted.
             doesn't validate this exception will be raised. This will occur
             when the ciphertext has been changed, but will also occur when the
             key, nonce, or associated data are wrong.
+
+.. class:: AESCCM(key)
+
+    .. versionadded:: 2.0
+
+    .. note:
+
+        AES-CCM is provided largely for compatibility with existing protocols.
+        Due to its construction it is not as computationally efficient as
+        other AEAD ciphers.
+
+    The AES-CCM construction is composed of the
+    :class:`~cryptography.hazmat.primitives.ciphers.algorithms.AES` block
+    cipher utilizing Counter with CBC-MAC (CCM) (specified in :rfc:`3610`).
+
+    :param bytes key: A 128, 192, or 256-bit key. This **must** be kept secret.
+
+    :raises cryptography.exceptions.UnsupportedAlgorithm: If the version of
+        OpenSSL does not support AES-CCM.
+
+    .. doctest::
+
+        >>> import os
+        >>> from cryptography.hazmat.primitives.ciphers.aead import AESCCM
+        >>> data = b"a secret message"
+        >>> aad = b"authenticated but unencrypted data"
+        >>> key = AESCCM.generate_key(bit_length=128)
+        >>> aesccm = AESCCM(key)
+        >>> nonce = os.urandom(13)
+        >>> ct = aesccm.encrypt(nonce, data, aad)
+        >>> aesccm.decrypt(nonce, ct, aad)
+        'a secret message'
+
+    .. classmethod:: generate_key(bit_length)
+
+        Securely generates a random AES-CCM key.
+
+        :param bit_length: The bit length of the key to generate. Must be
+            128, 192, or 256.
+
+        :returns bytes: The generated key.
+
+    .. method:: encrypt(nonce, data, associated_data, tag_length=16)
+
+        .. warning::
+
+            Reuse of a ``nonce`` with a given ``key`` compromises the security
+            of any message with that ``nonce`` and ``key`` pair.
+
+        Encrypts and authenticates the ``data`` provided as well as
+        authenticating the ``associated_data``.  The output of this can be
+        passed directly to the ``decrypt`` method.
+
+        :param bytes nonce: A value of between 7 and 13 bytes. The maximum
+            length is determined by the length of the ciphertext you are
+            encrypting and must satisfy the condition:
+            ``len(data) < 2 ** (8 * (15 - len(nonce)))``
+            **NEVER REUSE A NONCE** with a key.
+        :param bytes data: The data to encrypt.
+        :param bytes associated_data: Additional data that should be
+            authenticated with the key, but is not encrypted. Can be ``None``.
+        :param int tag_length: The length of the authentication tag. This
+            defaults to 16 bytes and it is **strongly** recommended that you
+            do not make it shorter unless absolutely necessary. Valid tag
+            lengths are 4, 6, 8, 12, 14, and 16.
+        :returns bytes: The ciphertext bytes with the tag appended.
+
+    .. method:: decrypt(nonce, data, associated_data, tag_length=16)
+
+        Decrypts the ``data`` and authenticates the ``associated_data``. If you
+        called encrypt with ``associated_data`` you must pass the same
+        ``associated_data`` in decrypt or the integrity check will fail.
+
+        :param bytes nonce: A value of between 7 and 13 bytes. This
+            is the same value used when you originally called encrypt.
+            **NEVER REUSE A NONCE** with a key.
+        :param bytes data: The data to decrypt (with tag appended).
+        :param bytes associated_data: Additional data to authenticate. Can be
+            ``None`` if none was passed during encryption.
+        :param int tag_length: The length of the authentication tag. This
+            defaults to 16 bytes. You only need to change this if your existing
+            ciphertext has a shorter tag. Valid tag lengths are 4, 6, 8, 12,
+            14, and 16.
+        :returns bytes: The original plaintext.
+        :raises cryptography.exceptions.InvalidTag: If the authentication tag
+            doesn't validate this exception will be raised. This will occur
+            when the ciphertext has been changed, but will also occur when the
+            key, nonce, or associated data are wrong.
--- a/src/cryptography/hazmat/backends/openssl/aead.py
+++ b/src/cryptography/hazmat/backends/openssl/aead.py
@@ -13,10 +13,13 @@ _DECRYPT = 0
 
 def _aead_cipher_name(cipher):
     from cryptography.hazmat.primitives.ciphers.aead import (
-        ChaCha20Poly1305
+        AESCCM, ChaCha20Poly1305
     )
-    assert isinstance(cipher, ChaCha20Poly1305)
-    return b"chacha20-poly1305"
+    if isinstance(cipher, ChaCha20Poly1305):
+        return b"chacha20-poly1305"
+    else:
+        assert isinstance(cipher, AESCCM)
+        return "aes-{0}-ccm".format(len(cipher._key) * 8).encode("ascii")
 
 
 def _aead_setup(backend, cipher_name, key, nonce, tag, tag_len, operation):
@@ -61,6 +64,18 @@ def _aead_setup(backend, cipher_name, key, nonce, tag, tag_len, operation):
     return ctx
 
 
+def _set_length(backend, ctx, data_len):
+    intptr = backend._ffi.new("int *")
+    res = backend._lib.EVP_CipherUpdate(
+        ctx,
+        backend._ffi.NULL,
+        intptr,
+        backend._ffi.NULL,
+        data_len
+    )
+    backend.openssl_assert(res != 0)
+
+
 def _process_aad(backend, ctx, associated_data):
     outlen = backend._ffi.new("int *")
     res = backend._lib.EVP_CipherUpdate(
@@ -78,10 +93,15 @@ def _process_data(backend, ctx, data):
 
 
 def _encrypt(backend, cipher, nonce, data, associated_data, tag_length):
+    from cryptography.hazmat.primitives.ciphers.aead import AESCCM
     cipher_name = _aead_cipher_name(cipher)
     ctx = _aead_setup(
         backend, cipher_name, cipher._key, nonce, None, tag_length, _ENCRYPT
     )
+    # CCM requires us to pass the length of the data before processing anything
+    # However calling this with any other AEAD results in an error
+    if isinstance(cipher, AESCCM):
+        _set_length(backend, ctx, len(data))
 
     _process_aad(backend, ctx, associated_data)
     processed_data = _process_data(backend, ctx, data)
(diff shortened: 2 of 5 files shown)


alex:
This can be rebased now.

docs/hazmat/primitives/aead.rst
```diff
@@ -77,3 +77,92 @@ also support providing integrity for associated data which is not encrypted.
...
+
+    .. method:: encrypt(nonce, data, associated_data, tag_length=16)
+
+        .. warning::
+
+            Reuse of a ``nonce`` with a given ``key`` compromises the security
+            of any message with that ``nonce`` and ``key`` pair.
+
+        Encrypts the ``data`` provided and authenticates the
+        ``associated_data``.  The output of this can be passed directly
```
alex: It also authenticates the ``data``, please rephrase.

docs/hazmat/primitives/aead.rst
```diff
@@ -77,3 +77,92 @@ also support providing integrity for associated data which is not encrypted.
...
+        to the ``decrypt`` method.
+
+        :param bytes nonce: A value of between 7 and 13 bytes. The maximum
+            length is determined by the length of the ciphertext you are
+            encrypting and must satisfy the condition:
+            ``len(data) < 2 ** (8 * (15 - len(nonce)))``
```
alex: Is this right? This indicates that the longer the nonce the shorter the ciphertext must be?

docs/hazmat/primitives/aead.rst
```diff
@@ -77,3 +77,92 @@ also support providing integrity for associated data which is not encrypted.
...
+            **NEVER REUSE A NONCE** with a key.
+        :param bytes data: The data to encrypt.
+        :param bytes associated_data: Additional data that should be
+            authenticated with the key, but does not need to be encrypted. Can
```
alex: s/does not need to be/is not/

docs/hazmat/primitives/aead.rst
```diff
@@ -77,3 +77,92 @@ also support providing integrity for associated data which is not encrypted.
...
+            be ``None``.
+        :param int tag_length: The length of the authentication tag. This
```
alex: Is there a reason to support truncated tags?

docs/hazmat/primitives/aead.rst
```diff
@@ -77,3 +77,92 @@ also support providing integrity for associated data which is not encrypted.
...
+        called encrypt with ``associated_data`` you must pass the same
+        ``associated_data`` in decrypt or the integrity check will fail.
+
+        :param bytes nonce: A value of between 7 and 13 bytes. This
+            is the same value used when you originally called encrypt.
+            **NEVER REUSE A NONCE** with a key.
+        :param bytes data: The data to decrypt (with tag appended).
+        :param bytes associated_data: Additional data to authenticate. Can be
+            ``None`` if none was passed during encryption.
+        :param int tag_length: The length of the authentication tag. This
```
alex: This param isn't in the signature, but I'm opposed to it :-)

src/cryptography/hazmat/backends/openssl/aead.py
```diff
@@ -11,8 +11,9 @@
 _DECRYPT = 0
 
 
-def _chacha20poly1305_setup(backend, key, nonce, tag, operation):
-    evp_cipher = backend._lib.EVP_get_cipherbyname(b"chacha20-poly1305")
+def _aead_setup(backend, cipher_name, key, nonce, tag, tag_len, operation):
```
alex: Please split these changes out into their own PR :-)

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1836,19 +1836,39 @@ def derive_scrypt(self, key_material, salt, length, n, r, p):
...
+    def aesccm_decrypt(self, key, nonce, data, associated_data, tag_length):
+        cipher_name = "aes-{0}-ccm".format(len(key) * 8).encode("ascii")
+        return aead._decrypt(
+            self, cipher_name, key, nonce, data, associated_data, tag_length
+        )
+
+    def aead_cipher_supported(self, cls):
+        # We do this ugly thing because circular import problems prevent
+        # importing aead so we could just do isinstance checks
+        if cls.__name__ == "ChaCha20Poly1305":
```
alex: Just importing the class in this function should work fine?

src/cryptography/hazmat/primitives/ciphers/aead.py
```diff
@@ -12,7 +12,7 @@
 
 class ChaCha20Poly1305(object):
     def __init__(self, key):
-        if not backend.chacha20poly1305_supported():
+        if not backend.aead_cipher_supported(self.__class__):
```
alex: prefer `type(self)`

src/cryptography/hazmat/primitives/ciphers/aead.py
```diff
@@ -52,3 +52,67 @@ def _check_params(self, nonce, data, associated_data):
...
+
+class AESCCM(object):
+    def __init__(self, key):
+        utils._check_bytes("key", key)
+        if len(key) not in (16, 24, 32):
+            raise ValueError("AESCCM key must be 128, 192, or 256 bits.")
+
+        self._key = key
+
+        if not backend.aead_cipher_supported(self.__class__):
```
alex: Same

src/cryptography/hazmat/primitives/ciphers/aead.py
```diff
@@ -52,3 +52,67 @@ def _check_params(self, nonce, data, associated_data):
...
+    def decrypt(self, nonce, data, associated_data, tag_length=16):
+        if associated_data is None:
+            associated_data = b""
+
+        self._check_params(nonce, data, associated_data, tag_length)
+        return backend.aesccm_decrypt(
+            self._key, nonce, data, associated_data, tag_length
+        )
+
+    def _validate_l(self, nonce, data_len):
```
alex: give this a clearer name please

src/cryptography/hazmat/primitives/ciphers/aead.py
```diff
@@ -52,3 +52,67 @@ def _check_params(self, nonce, data, associated_data):
...
+        if not isinstance(tag_length, int):
+            raise TypeError("tag_length must be an integer")
+
+        if tag_length not in (4, 6, 8, 12, 14, 16):
+            raise ValueError("Invalid tag_length")
+
+        utils._check_bytes("nonce", nonce)
+        utils._check_bytes("data", data)
+        utils._check_bytes("associated_data", associated_data)
+        if len(nonce) < 7 or len(nonce) > 13:
```
alex: `if not 7 <= len(nonce) <= 13:` is slightly more readable

docs/hazmat/primitives/aead.rst
reaperhulk: CCM tags change based on length specified (rather than GCM's trivial truncation) and some specs don't use full length tags (including some TLS 1.2 cipher suites which hopefully nobody is implementing as well as IPsec, which mandates 8 and 16 byte tag support and says 12 is optional).

docs/hazmat/primitives/aead.rst
reaperhulk: Can't avoid it when supporting multiple tag lengths (which, as previously noted, I think we need to).

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: Circular import because we import backend in cryptography.hazmat.primitives.ciphers.aead :(

docs/hazmat/primitives/aead.rst
reaperhulk: (Note to self: whatever we end up with on rephrasing we should also use it for chacha as this is the same text as the chacha docs use)

docs/hazmat/primitives/aead.rst
reaperhulk: Yes, the length of the ciphertext shortens the ciphertext because the length is encoded as part of the nonce. See https://tools.ietf.org/html/rfc3610#section-2

docs/hazmat/primitives/aead.rst
reaperhulk: (Same on porting new language to chacha)

src/cryptography/hazmat/primitives/ciphers/aead.py
reaperhulk: Sure, what's the advantage for future reference? Beyond the general "any time you don't directly reference a dunder that's a win"?

src/cryptography/hazmat/primitives/ciphers/aead.py
reaperhulk: In the RFC this is named `L` and the description is "Number of octets in length field". It represents the encoding of the length of the encrypted payload into bytes. `_validate_nonce_payload_length` maybe? I feel like `validate_l` is actually a good name given the fact that any debugging is going to necessitate a visit to the RFC.

alex:
Rebase required :-)

src/cryptography/hazmat/backends/openssl/aead.py
```diff
@@ -74,6 +86,9 @@ def _encrypt(backend, cipher_name, key, nonce, data, associated_data,
     ctx = _aead_setup(
         backend, cipher_name, key, nonce, None, tag_length, _ENCRYPT
     )
+    # CCM requires us to pass the length of the data before processing anything
+    if cipher_name[-3:] == b"ccm":
```
alex: I don't like checking on this string. Let's think of a better API.

Can we just call `_set_length` for all AEADs?

src/cryptography/hazmat/backends/openssl/backend.py
alex: Why does that prevent us from doing an import inside this function?

src/cryptography/hazmat/primitives/ciphers/aead.py
alex: How about `vaildate_lengths`, since it looks at `len(nonce)` and `len(data)`?

tests/hazmat/primitives/test_aead.py
```diff
@@ -31,7 +35,7 @@ def test_chacha20poly1305_unsupported_on_older_openssl(backend):
 
 
 @pytest.mark.supported(
-    only_if=lambda backend: backend.chacha20poly1305_supported(),
+    only_if=lambda backend: backend.aead_cipher_supported(ChaCha20Poly1305),
```
alex: If we're switching to an `aead_cipher_supported` API, can we split that out into a separate PR (lol)

src/cryptography/hazmat/backends/openssl/aead.py
reaperhulk: Just tried, nope :(

alex:
Ready for YAR (Yet Another Rebase)

src/cryptography/hazmat/backends/openssl/aead.py
```diff
@@ -84,6 +103,10 @@ def _encrypt(backend, cipher_cls, key, nonce, data, associated_data,
     ctx = _aead_setup(
         backend, cipher_name, key, nonce, None, tag_length, _ENCRYPT
     )
+    # CCM requires us to pass the length of the data before processing anything
```
alex: Move the import to the top of the function

src/cryptography/hazmat/backends/openssl/aead.py
```diff
@@ -84,6 +103,10 @@ def _encrypt(backend, cipher_cls, key, nonce, data, associated_data,
...
+    from cryptography.hazmat.primitives.ciphers.aead import AESCCM
+    if cipher_cls is AESCCM:
```
alex: what happened if we called this for other aeads?

src/cryptography/hazmat/backends/openssl/aead.py
```diff
@@ -111,12 +134,29 @@ def _decrypt(backend, cipher_cls, key, nonce, data, associated_data,
...
     _process_aad(backend, ctx, associated_data)
-    processed_data = _process_data(backend, ctx, data)
-    outlen = backend._ffi.new("int *")
-    res = backend._lib.EVP_CipherFinal_ex(ctx, backend._ffi.NULL, outlen)
-    if res == 0:
-        backend._consume_errors()
-        raise InvalidTag
+    # CCM has a different error path if the tag doesn't match. Errors are
+    # raised in Update and Final is irrelevant.
+    if cipher_name[-3:] == b"ccm":
```
alex: no string typing

src/cryptography/hazmat/backends/openssl/aead.py
```diff
@@ -13,11 +13,18 @@
 
 def _aead_cipher_name(cls, key_length):
```
alex: Ok, rather than pass the type here and doing all this weird `if key_lenght is None` stuff, let's pass an instance of the aead instead of the `cls`, then we can use the `key` attr.

src/cryptography/hazmat/backends/openssl/aead.py
```diff
@@ -111,12 +134,29 @@ def _decrypt(backend, cipher_cls, key, nonce, data, associated_data,
...
-    res = backend._lib.EVP_CipherFinal_ex(ctx, backend._ffi.NULL, outlen)
-    if res == 0:
-        backend._consume_errors()
-        raise InvalidTag
+    # CCM has a different error path if the tag doesn't match. Errors are
+    # raised in Update and Final is irrelevant.
+    if cipher_name[-3:] == b"ccm":
+        outlen = backend._ffi.new("int *")
+        buf = backend._ffi.new("unsigned char[]", len(data))
+        res = backend._lib.EVP_CipherUpdate(ctx, buf, outlen, data, len(data))
```
alex: Why can't this go through `_process_data`?

src/cryptography/hazmat/backends/openssl/aead.py
reaperhulk: internal error out of openssl

alex: that's obnoxious. can you put in the comment that calling this on other AEAD's is an internal error

src/cryptography/hazmat/backends/openssl/aead.py
reaperhulk: Because we need to check the error code and `_process_data` asserts on the error code and returns the buffer. Plus, calling EVP_CipherFinal returns zero with CCM (what a horribly overloaded API EVP is)

PR #3702, opened 2017-06-13
update pythons and pypy, a little prep for pypy3 in CI again

reaperhulk:


.travis/install.sh
```diff
@@ -26,20 +26,20 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
...
         py34)
-            pyenv install 3.4.5
-            pyenv global 3.4.5
+            pyenv install 3.4.6
+            pyenv global 3.4.6
             ;;
         py35)
-            pyenv install 3.5.2
-            pyenv global 3.5.2
+            pyenv install 3.5.3
```
markrwilliams: [Travis uses Python 3.5.2](https://travis-ci.org/pyca/cryptography/jobs/242682298#L117).  I guess the difference between microreleases isn't worth the additional build time of installing 3.5.3? Especially because there are Jenkins jobs to test against platform Pythons of all versions.

Issue #3509, opened 2017-04-20
ED25519 Support


atagar:
Hi there. Sorry if there's another issue tracking this. Presently [Stem](https://stem.torproject.org/) (part of [Tor](https://www.torproject.org/)) is using [PyNaCl](https://pypi.python.org/pypi/PyNaCl/) to [validate ed25519 certificates](https://gitweb.torproject.org/stem.git/commit/?id=727e14b9570ce3881228503ee04ace491de62009).

We use cryptography elsewhere and would love to use it for this too once there's support so we can drop the extra dependency. An [email from January](https://mail.python.org/pipermail/cryptography-dev/2017-January/000715.html) suggested this might be coming so just filing a tracking issue for us to watch so we know when we can switch.

If we should be following something else instead then feel free to resolve in favor of that.

Thanks! -Damian

alex:
This won't happen until OpenSSL has ed25519 support: https://github.com/openssl/openssl/issues/487

This can stay open, as none of our other tracking issues cover ed25519

atagar:
Damn that was quick. Thanks Alex!

Filed a tracking issue on our end to take advantage of this once it's available. Looking forward to it! :P

https://trac.torproject.org/projects/tor/ticket/22022

joelio:
+1 for this

himikof:
Ed25519 support was recently merged in OpenSSL master: openssl/openssl#3503.

reaperhulk:
Yep, we'll be implementing this when 1.1.1 is released.

atagar:
Wonderful, thanks!

reaperhulk:
I went ahead and implemented this as an experiment today: https://github.com/reaperhulk/cryptography/commit/fc663b7687f7696990b3074e5d721f95640887f6

That is a **very** WIP patch. It passes all tests, but there's no support for serialization of private keys, public key serialization is bare bytes only, there's no documentation, and no attempt has been made to share code across the x25519/ed25519 work (when it can easily be shared in several locations). Also, to use this you'll have to compile against OpenSSL 1.1.1-dev (which means cloning current git master, building it, and successfully linking against it instead of whatever other OpenSSL you may have resident on your system)

MaggieLplus:
Ed25519 TLS support was also merged to OpenSSL master: openssl/openssl#3585

reaperhulk:
[...]

PR #3707, opened 2017-06-18
ed25519 vector loader

reaperhulk:
Since we already have ed25519 vectors we can also land a vector loader. Refs #3509 

For reference, djb's ed25519 vectors are structured as a colon delimited array:
        0: secret key (32 bytes) + public key (32 bytes)
        1: public key (32 bytes)
        2: message (0+ bytes)
        3: signature (64 bytes) + message (0+ bytes)

For our future ed25519 support we want secret key, public key, message, and signature as separate items in a dictionary.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -822,11 +822,11 @@ def load_nist_kbkdf_vectors(vector_data):
 def load_ed25519_vectors(vector_data):
     data = []
     for line in vector_data:
-        x = line.split(':')
-        secret_key = x[0][0:64]
-        public_key = x[1]
-        message = x[2]
-        signature = x[3][0:128]
+        secret_key, public_key, message, signature, _ = line.split(':')
+        # In the vectors the first element is secret key + public key
+        secret_key = secret_key[0:64]
+        # In the vectors the signature section is signature + message
+        signature = signature[0:128]
         data.append({
             "secret_key": secret_key,
             "public_key": public_key,
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -817,3 +817,20 @@ def load_nist_kbkdf_vectors(vector_data):
             test_data[name.lower()] = value.encode("ascii")
 
     return vectors
+
+
+def load_ed25519_vectors(vector_data):
+    data = []
+    for line in vector_data:
+        x = line.split(':')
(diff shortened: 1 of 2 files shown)


tests/utils.py
```diff
@@ -817,3 +817,20 @@ def load_nist_kbkdf_vectors(vector_data):
             test_data[name.lower()] = value.encode("ascii")
 
     return vectors
+
+
+def load_ed25519_vectors(vector_data):
+    data = []
+    for line in vector_data:
+        x = line.split(':')
+        secret_key = x[0][0:64]
```
alex: Why are we slicing the key, are they the wrong length sometimes?

tests/utils.py
```diff
@@ -817,3 +817,20 @@ def load_nist_kbkdf_vectors(vector_data):
             test_data[name.lower()] = value.encode("ascii")
 
     return vectors
+
+
+def load_ed25519_vectors(vector_data):
+    data = []
+    for line in vector_data:
+        x = line.split(':')
```
alex: x is not a good name for this var.

tests/utils.py
```diff
@@ -817,3 +817,20 @@ def load_nist_kbkdf_vectors(vector_data):
...
+        secret_key = x[0][0:64]
+        public_key = x[1]
+        message = x[2]
+        signature = x[3][0:128]
```
alex: This could all be written better with some unpacking `secret_key, public_key, message, signature = line.split(":")`

tests/utils.py
reaperhulk: The info is in the initial PR comment :smile:

Here's what the indices are:

0: secret key (32 bytes) + public key (32 bytes)
1: public key (32 bytes)
2: message (0+ bytes)
3: signature (64 bytes) + message (0+ bytes)

So we're slicing because djb's vectors are a bit weird and concatenate info in ways we don't need/want.

tests/utils.py
reaperhulk: Do you want to see an unpack and then slice afterwards? (e.g. `signature = signature[0:128]`)?

reaperhulk: (I've done this)

PR #3709, opened 2017-06-19
add pypy3.5 to our test matrix

reaperhulk:
Let's see if this is too slow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ matrix:
           env: TOXENV=py36
         - env: TOXENV=pypy PYPY_VERSION=pypy2-5.3
         - env: TOXENV=pypy PYPY_VERSION=pypy2.7-5.8.0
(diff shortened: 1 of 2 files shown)


reaperhulk:
21 min vs 13 min. That's not really acceptable given our already excessive CI duration. Since we don't need the coverage off pypy3 let's see what happens if we re-add `pypy-nocoverage` (and since we no longer care about a Mac specific backend we can use it on the Mac side too)

reaperhulk:
With coverage turned off pypy3.5-5.8.0 drops to 5.5 minutes in our CI. I'm happy to have it run with that speed. Switching the mac pypy to no coverage drops it to a bit under 8 minutes from ~19-20 previously, which also seems worthwhile given our CI cycle constraints. We can always turn coverage back on if we need it on these builders in the future.

PR #3714, opened 2017-06-22
be clearer that 65537 is the right answer

reaperhulk:


docs/hazmat/primitives/asymmetric/rsa.rst
```diff
@@ -23,7 +23,8 @@ mathematical properties`_.
     provide more security, currently ``1024`` and below are considered
     breakable, and ``2048`` or ``4096`` are reasonable default key sizes for
     new keys. The ``public_exponent`` indicates what one mathematical property
-    of the key generation will be, ``65537`` should almost always be used.
+    of the key generation will be. Unless you have a special requirement for
```
alex: I've read this like 12 times, and I'm not really convinced it's stronger. halp

reaperhulk: I feel like it's stronger because devs should know if they have a special requirement (after all it's their requirement), while "should almost always" appears to cause them to second guess whether this is one of those situations? But maybe that just reduces to the same issue.

What about `Despite this being a configurable value, you should use 65537`.

alex: How about: "Unless you have a specific reason to do otherwise, you should always use 65537"

reaperhulk: 👍

PR #3721, opened 2017-06-24
doc the relationship between PrivateKey and PrivateKeyWithSerialization

reaperhulk:
Or at least do it better.

### DIFF
--- a/docs/hazmat/primitives/asymmetric/dh.rst
+++ b/docs/hazmat/primitives/asymmetric/dh.rst
@@ -143,6 +143,9 @@ Key interfaces
 
     .. versionadded:: 0.9
 
+    DH private key objects have methods from this interface as well as
+    :class:`DHPrivateKeyWithSerialization`.
+
     .. attribute:: key_size
 
         The bit length of the prime modulus.
@@ -173,8 +176,9 @@ Key interfaces
 
     .. versionadded:: 0.9
 
-    Inherits from
-    :class:`~cryptography.hazmat.primitives.asymmetric.dh.DHPrivateKey`.
+    This interface contains additional methods relating to serialization.
+    Any object with this interface also has all the methods from
+    :class:`DHPrivateKey`.
 
     .. method:: private_numbers()
 
--- a/docs/hazmat/primitives/asymmetric/dh.rst
+++ b/docs/hazmat/primitives/asymmetric/dh.rst
@@ -143,8 +143,8 @@ Key interfaces
 
     .. versionadded:: 0.9
 
-    DH private key objects have methods from this interface as well as
-    :class:`DHPrivateKeyWithSerialization`.
+    A DH private key that is not an :term:`opaque key` also implements
+    :class:`DHPrivateKeyWithSerialization` to provide serialization methods.
 
     .. attribute:: key_size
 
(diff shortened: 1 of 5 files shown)


docs/hazmat/primitives/asymmetric/ec.rst
```diff
@@ -441,7 +441,8 @@ Key Interfaces
     .. versionadded:: 0.5
 
     An elliptic curve private key for use with an algorithm such as `ECDSA`_ or
-    `EdDSA`_.
+    `EdDSA`_. Elliptic curve private key objects have methods from this
+    interface as well as :class:`EllipticCurvePrivateKeyWithSerialization`.
```
alex: Same.

docs/hazmat/primitives/asymmetric/rsa.rst
```diff
@@ -530,7 +530,8 @@ Key interfaces
 
     .. versionadded:: 0.2
 
-    An `RSA`_ private key.
+    An `RSA`_ private key. RSA private key objects have methods from this
+    interface as well as :class:`RSAPrivateKeyWithSerialization`.
```
alex: And one more.

reaperhulk:
I think there might have been some comments lost here Alex

docs/hazmat/primitives/asymmetric/dh.rst
```diff
@@ -173,8 +176,9 @@ Key interfaces
 
     .. versionadded:: 0.9
 
-    Inherits from
-    :class:`~cryptography.hazmat.primitives.asymmetric.dh.DHPrivateKey`.
+    This interface contains additional methods relating to serialization.
+    Any object with this interface also has all the methods from
+    :class:`DHPrivateKey`.
```
alex: This seems backwards?

docs/hazmat/primitives/asymmetric/dh.rst
```diff
@@ -143,6 +143,9 @@ Key interfaces
 
     .. versionadded:: 0.9
 
+    DH private key objects have methods from this interface as well as
+    :class:`DHPrivateKeyWithSerialization`.
```
alex: This seems backwards

reaperhulk: Do you mean the ordering? As in "DH private key objects have methods from DHPrivateKeyWithSerialization in addition to those on this interface" rather than the current phrasing?

alex: No, I mean it's very confusing/arguably wrong. Only DH private keys that implement the WithSerialization interface will have those methods; right now that's all the ones we ship, but I don't understand what this is trying to say.

reaperhulk: I can't come up with a clear way to say "Technically private keys may not implement serialization in the future but for now they all do and even when that's not true your code won't care unless you add support for opaque keys to your application".

How about `DH private key objects typically also implement DHPrivateKeyWithSerialization to provide serialization methods`?

The one problem with that phrasing is that we know words like "typically" or "usually" results in user confusion (see: 65537 on RSA generation).

alex: That phrasing sounds good; you can also try `Many DH private keys also implement [...]`?

reaperhulk: What about `DH private keys that are not opaque keys (link) also implement DHPrivateKeyWithSerialization to provide serialization methods`? Then opaque keys will link to a glossary section that would say:

An opaque key is a type of key that allows you to perform cryptographic operations such as encryption, decryption, signing, and verification, but does not allow access to the key itself. Typically an opaque key is loaded from a hardware security module (HSM) (link to wikipedia).

alex: Love it.

PR #3725, opened 2017-06-26
bind DTLS 1.2 methods

### LINKED ISSUE #3501: wishlist: DTLS 1.2 methods
Please consider adding the ``DTLSv1_2_*method`` symbols to the openssl bindings (``DTLSv1_*method`` are present at least in the latest release, but the newer ``DTLS_*method`` and the ``DTLSv1_2_*method`` are missing). It seems to me that adding those is sufficient to utilize DTLS 1.2 in the Python OpenSSL bindings.

DTLS version 1.2 is mandated for the transport layer security suite of the CoAP protocol that is increasingly being rolled out in IoT devices.

reaperhulk:
Fixes #3501 (Oh boy IoT).

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -26,7 +26,7 @@ static const long Cryptography_HAS_GET_SERVER_TMP_KEY;
 static const long Cryptography_HAS_SSL_CTX_SET_CLIENT_CERT_ENGINE;
 static const long Cryptography_HAS_SSL_CTX_CLEAR_OPTIONS;
 static const long Cryptography_HAS_DTLS;
-static const long Cryptography_HAS_DTLS12;
+static const long Cryptography_HAS_GENERIC_DTLS_METHOD;
 
 /* Internally invented symbol to tell us if SNI is supported */
 static const long Cryptography_HAS_TLSEXT_HOSTNAME;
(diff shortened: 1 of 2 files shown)


src/_cffi_src/openssl/ssl.py
```diff
@@ -590,6 +599,21 @@
 static const long TLS_ST_OK = 0;
 #endif
 
+#if defined(OPENSSL_NO_DTLS1_2_METHOD) || \
```
reaperhulk: Yes there really are this many ways to disable DTLS 1.2 support. Probably nothing uses them, but 🤷‍♂️

src/_cffi_src/openssl/ssl.py
```diff
@@ -331,6 +332,14 @@
 const SSL_METHOD *DTLSv1_server_method(void);
 const SSL_METHOD *DTLSv1_client_method(void);
 
+/* Added in 1.0.2 */
+const SSL_METHOD *DTLSv1_2_method(void);
+const SSL_METHOD *DTLSv1_2_server_method(void);
+const SSL_METHOD *DTLSv1_2_client_method(void);
```
alex: Can we skip the explicitly versioned ones and just do `DTLS_{server,client,}_method`? That's the prefered API + using `OP_NO` to disable unwanted protocols

reaperhulk: Seems reasonable to me.

PR #3729, opened 2017-06-26
remove python 3.3 from CI

reaperhulk:


alex:
This needs updates to: `README`, `CHANGELOG`, `installation`, `setup.py`, `cryptography/__init__.py` :-)

reaperhulk:
I was going to do that as a separate PR but since it actually causes a coverage issue to not remove the deprecation warning I'll do it all together :)

PR #3730, opened 2017-06-27
Run asan on jenkins mac builders

### LINKED ISSUE #3405: Compile C code with ASAN/UBSAN for tests

reaperhulk:
Fixes #3405, supersedes #3406

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -242,7 +242,7 @@ def build(toxenv, label, imageName, artifacts, artifactExcludes) {
                             codecov -e JOB_BASE_NAME,LABEL
                         """
                     } else if (label.contains("sierra") || label.contains("yosemite")) {
-                        if label.contains("sierra") {
+                        if (label.contains("sierra")) {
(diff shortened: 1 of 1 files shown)


reaperhulk:
Same issue here as the old PR actually. Despite using DYLD_INSERT_LIBRARIES asan won't load. No idea why.

reaperhulk:
DYLD_INSERT_LIBRARIES appears to be special in that it does not propagate through to subshells (assuming that's what tox uses?). So we can set it (via export or directly before we call tox) but it will not be present in the env tox has.

We *can* work around this via `setenv`, but then we can't use shell expansion to set the environment, which would make a tox -e asan target very fragile since the path is clang version dependent. Ideas?

Also, the 6 tests where we spawn subshells fail in this case still.

reaperhulk:
As an additional experiment I tried setting up asan via gcc on ubuntu rolling. It bombed out early in the tox run (before it even started collecting tests) claiming a bunch of memory leaks.

PR #3732, opened 2017-06-28
Fix for leaking memory in EllipticCurvePublicNumbers.public_key()

alex:


alex:
As far as I can tell, this leak has been there the entire time we've supported EC.

alex:
@reaperhulk once this is landed, do you have an opinion on backporting + cutting a release? The mozilla push team hits this and would be appreciative (but I could go either way)

reaperhulk:
Okay, confirmed this does what it should. I guess we assumed `EC_KEY_set_public_key_affine_coordinates` took ownership and didn't duplicate the BNs.

I'd prefer to just ship this with 2.0, but I think that's probably 2-3 weeks away. Is that timeline okay?

PR #3734, opened 2017-06-28
don't use long keys because they can be confusing wrt bits/bytes

### LINKED ISSUE #3733: KDF documentation mismatch: bytes vs bits
After doing some work with the KDF functions, I noticed an issue with the documentation (`docs/hazmat/primitives/key-derivation-functions.rst`). The `length` fields are often defined as `The desired length of the derived key in bytes.`. However, some of the examples clearly use bits as the `length` value (e.g., `KBKDFHMAC` uses 256, lines 637 vs 654).

I'm not too familiar with all of the different KDF functions and their parameters but the docs may need a cursory review to match the examples with the parameter descriptions.

reaperhulk:
fixes #3733


alex:
Can you update the docs for each of these params to be clear it's in bytes?

alex:
codecov is not having a good time, docs only so merging.

PR #3735, opened 2017-06-29
Run tests on buster

alex:


reaperhulk:
Is there any real difference between sid and buster for the next year or so?

alex:
No clue

reaperhulk:
cc @mithrandi 😁

mithrandi:
Since packages do not migrate from sid to buster instantly, there will be transient differences, but I don't think there is any need to test against both. Probably you only need to worry about sid until buster enters freeze at the end of the release cycle.

alex:
My personal attitude is that we may as well do it now, and not worry about it in the future. It doesn't add much time to CI.

reaperhulk:
I'm opposed mostly because we don't have the parallel capacity to do all our docker builds simultaneously at this point. I wouldn't care if we had a secondary docker builder, but we haven't decided to do that yet ;)

reaperhulk:
(We have more docker capacity now)

PR #3736, opened 2017-06-29
add manylinux1 wheel building to the wheel builder

reaperhulk:


.jenkins/Jenkinsfile-cryptography-wheel-builder
```diff
@@ -103,6 +121,29 @@ def build(version, label) {
...
+                sh """#!/usr/bin/env bash
+                    set -x -e
+                    # Because we are doing this as root in the container, but we write to a mounted dir that is outside the container
+                    # we need to make sure we set these files writable such that the jenkins user can delete them afterwards
+                    mkdir -p tmpwheelhouse
+                    mkdir -p wheelhouse
+                    chmod -R 777 tmpwheelhouse
+                    chmod -R 777 wheelhouse
+
+                    $linux32 /opt/python/$version/bin/pip install cffi six idna asn1crypto ipaddress enum34
```
reaperhulk: Is there a better way for us to get the deps for cryptography without enumerating them here? We do a pip install later with --no-index so we need all the deps.

.jenkins/Jenkinsfile-cryptography-wheel-builder
```diff
@@ -120,11 +161,25 @@ for (config in configs) {
...
-            node(label) {
-                stage(combinedName) {
-                    build(version, label)
+        if (label.contains("docker")) {
+            def imageName = config["imageName"]
+            def combinedName = "${imageName}-${version}"
+            builders[combinedName] = {
+                node(label) {
+                    stage(combinedName) {
+                        docker.image(imageName).inside("-u root") {
```
alex: Why do we need `-u root`?

reaperhulk: Because the docker image doesn't have the jenkins user. HOWEVER, that's only really required on the manylinux1 builds we do directly from the base container. Since we're generating our own here we should really update the dockerfile to add the user and then we don't need to do this nonsense.

alex: Ok, let's do that.

alex: This can be fixed now.

reaperhulk: (Turns out the manylinux1 images don't have virtualenv so if we switched away from root we'd need to add the new user to the container, install virtualenv, and then init and source a venv in the bash script. It was decided this was more complexity than just letting it operate as root)

alex:
Oh, please add a changelog!

reaperhulk:
There's more than just changelog to add. Also need to update the installation docs and probably add a section on how to avoid getting a wheel if you don't want one.

reaperhulk:
I've added a changelog entry. I'll put in a follow-up PR to add detail in installation.rst (working on it now).

PR #3741, opened 2017-06-30
Fixed a memory leak in EC private numbers

alex:


tests/hazmat/backends/test_openssl_memleak.py
```diff
@@ -190,21 +190,27 @@ def func(path):
             cert.extensions
         """), [path])
 
-    def test_ec_public_numbers_public_key(self):
+    def test_ec_private_numbers_private_key(self):
```
reaperhulk: Why are you altering this test and not adding another one?

alex: Because this test is a strict superset, it excersises both code paths.

reaperhulk: 👍

PR #3743, opened 2017-07-01
require a version of twine that uses pypi.org

alex:


dev-requirements.txt
```diff
@@ -3,6 +3,6 @@ clint
 coverage
 requests
 tox >= 2.4.1
-twine
+twine >=
```
reaperhulk: Shouldn't this have a version?

alex: ... lol yes, wtf

PR #3746, opened 2017-07-01
update linux install docs and release instructions

reaperhulk:


docs/installation.rst
```diff
@@ -74,31 +75,60 @@ local `wheel cache`_.
...
 .. code-block:: console
 
-    $ sudo yum install gcc libffi-devel python-devel openssl-devel
+    $ sudo apt-get install build-essential libssl-dev libffi-dev python3-dev
 
-You should now be able to build and install cryptography with the usual
+RHEL/CentOS
+~~~~~~~~~~~
+
+Replace ``python3-devel`` with ``python-devel`` if you're using Python 2.
```
alex: Are you sure `python3-devel` is a real thing?

reaperhulk: That's not in base repos, good point.

PR #3758, opened 2017-07-07
Remove DH check from load_dh_private_numbers

### LINKED ISSUE #3755: Diffie-Hellman inconsistency: Accepting broken values, rejecting good ones
I'm using cryptography 1.9 for python 3.4 on ubuntu 14.04 and python 3.5 on ubuntu 16.04.

The following is *not* a security issue.

I'm using Diffie-Hellman with extremely weak and broken values (for science). There are code paths where I get an error and there are code paths where my broken values are happily accepted. This is not a security issue since DH is not secure with broken or mitm-ed values (that is one reason why it is in the `hazmat` module). The other way round, there are also code paths which reject fine DH values.

The following works without complaint, loading my broken values.
```python
#!/usr/bin/env python3
from cryptography.hazmat.primitives.asymmetric import dh

#not a prime
notaprime = 2**512 - 1

pn = dh.DHParameterNumbers(notaprime, 2)
parameters = pn.parameters(default_backend())
bob_public = dh.DHPublicNumbers(pow(2, 42, notaprime), pn)
bob_public_key = bob_public.public_key(default_backend())
assert bob_public_key.key_size == 512
shared_key = parameters.generate_private_key().exchange(bob_public_key)
print(shared_key)
```

The shared key is completely broken (only one bit is set), but this is what we expect when using DH with a prime which is not a prime. PoC:
```python
kint = int.from_bytes(shared_key, 'big')
for i in range(512):
    if 2**i == kint:
        print('key bruteforced: {}'.format(i))
        assert shared_key == (2**i).to_bytes(64, 'big')
```

However, if I try the same thing slightly different, I get an error:
```python
#!/usr/bin/env python3
from cryptography.hazmat.primitives.asymmetric import dh

#not a prime
notaprime = 2**512 - 1
pn = dh.DHParameterNumbers(notaprime, 2)
parameters = pn.parameters(default_backend())
bob_public_key = peer_public_key = parameters.generate_private_key().public_key()
print(bob_public_key.key_size)
shared_key = parameters.generate_private_key().private_numbers()
print(shared_key.private_key(default_backend()))
```
```
  File "dh_reject.py", line 12, in <module>
    print(shared_key.private_key(default_backend()))
  File "/usr/local/lib/python3.4/dist-packages/cryptography/hazmat/primitives/asymmetric/dh.py", line 43, in private_key
    return backend.load_dh_private_numbers(self)
  File "/usr/local/lib/python3.4/dist-packages/cryptography/hazmat/backends/openssl/backend.py", line 1687, in load_dh_private_numbers
    raise ValueError("DH private numbers did not pass safety checks.")
ValueError: DH private numbers did not pass safety checks.
```

It is also expected behavior that those broken values are rejected.

My only concern is that it appears somewhat random to me whether my weak values are rejected or not.

The opposite also holds. There are good and strong DH values which get rejected.
```python
#!/usr/bin/env python3
import os
from binascii import unhexlify


# RFC3526 1536-bit DH values
p ="""FFFFFFFF FFFFFFFF C90FDAA2 2168C234 C4C6628B 80DC1CD1
      29024E08 8A67CC74 020BBEA6 3B139B22 514A0879 8E3404DD
      EF9519B3 CD3A431B 302B0A6D F25F1437 4FE1356D 6D51C245
      E485B576 625E7EC6 F44C42E9 A637ED6B 0BFF5CB6 F406B7ED
      EE386BFB 5A899FA5 AE9F2411 7C4B1FE6 49286651 ECE45B3D
      C2007CB8 A163BF05 98DA4836 1C55D39A 69163FA8 FD24CF5F
      83655D23 DCA3AD96 1C62F356 208552BB 9ED52907 7096966D
      670C354E 4ABC9804 F1746C08 CA237327 FFFFFFFF FFFFFFFF"""
p = int.from_bytes(unhexlify(p.replace('\n', '').replace(' ', '')), 'big')
g = 2

#run a DH exchange manually

## Alice:
xa = int.from_bytes(os.urandom(192), byteorder="big")
#ya is party a's public key; ya = g ^ xa mod p
ya = pow(g, xa, p)
print("Alice sends {},{},{}".format(ya, g, p))

## Bob:
#yb is party b's public key; yb = g ^ xb mod p
xb = int.from_bytes(os.urandom(192), byteorder="big")
yb = pow(g, xb, p)
# ZZ = (yb ^ xa)  mod p  = (ya ^ xb)  mod p
ZZ_b = pow(ya, xb, p)
print("Bob sends {}".format(yb))

## Alice:
ZZ_a = pow(yb, xa, p)

## Both have the same key
assert ZZ_a == ZZ_b


#the following works
from cryptography.hazmat.primitives.asymmetric import dh
pn = dh.DHParameterNumbers(p, 2)
parameters = pn.parameters(default_backend())
bob_public = dh.DHPublicNumbers(yb, pn)
bob_public_key = bob_public.public_key(default_backend())
assert bob_public_key.key_size == 1536
shared_key = parameters.generate_private_key().exchange(bob_public_key) #new key
print('worked', shared_key)

#the following fails
pn = dh.DHParameterNumbers(p, g)
parameters = pn.parameters(default_backend())
bob_public = dh.DHPublicNumbers(yb, pn)
bob_public_key = bob_public.public_key(default_backend())
assert bob_public_key.key_size == 1536
shared_key = dh.DHPrivateNumbers(xa, bob_public)
print(shared_key.private_key(default_backend())) #ValueError: DH private numbers did not pass safety checks.
```

debug: `dh.generate_parameters(generator=2, key_size=512, backend=default_backend()).parameter_numbers().p % 24` is always 11. For my prime p `p % 24`is  23

[Cryptography_DH_check](https://github.com/pyca/cryptography/blob/491fd7cc5aefff8064039d073331b41a2f606a63/src/_cffi_src/openssl/dh.py#L191) seems not to like 23 but wants 11.

I'm [not](https://crypto.stackexchange.com/questions/12961/diffie-hellman-parameter-check-when-g-2-must-p-mod-24-11) [the](https://groups.google.com/forum/#!msg/mailing.openssl.users/eEYp5ceyz8k/BlCOptLclfoJ) [first](https://github.com/nodejs/node-v0.x-archive/issues/2338) to notice this behavior of openssl. Yes, it is actually an issue of openssl. But somehow, I dislike the inconsistent behavior of pyhton-cryptography with regard to both, accepting obviously broken values as well as rejecting known good ones.

But probably this is not a bug at all.

Final disclaimer, in case this was not clear: This is **not** a security issue. Checking for bad DH values is a nice feature to prevent coding errors, but not a security feature. The following does **not** hold: "If we check that the DH values are good, then the DH exchange is secure." To get things secure, one needs to ensure the integrity and authenticity of a DH exchange. Though Cryptography_DH_check rejects my obviously broken value, it does not guarantee security (it is just a sanity checker, it would be approximately as hard to verify the *goodness* of DH values as just breaking DH).

reaperhulk:
DH_check turns out to block some bad values, but not all. It also blocks some **good** values. This causes problems. Removing this check will fix #3755 and #3364

Also, we don't attempt to check key consistency when loading from PEM or DER for DH anyway, so you could already load the rejected values via a previously serialized key.

`Cryptography_DH_check` is still called in `dh_parameters_supported` but I'm unconvinced we should call it there either. That can be debated later though.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -1772,13 +1772,6 @@ class Backend(object):
         res = self._lib.DH_set0_key(dh_cdata, pub_key, priv_key)
         self.openssl_assert(res == 1)
 
-        codes = self._ffi.new("int[]", 1)
-        res = self._lib.Cryptography_DH_check(dh_cdata, codes)
-        self.openssl_assert(res == 1)
-
-        if codes[0] != 0:
-            raise ValueError("DH private numbers did not pass safety checks.")
-
         evp_pkey = self._dh_cdata_to_evp_pkey(dh_cdata)
 
         return _DHPrivateKey(self, dh_cdata, evp_pkey)
--- a/tests/hazmat/primitives/test_dh.py
+++ b/tests/hazmat/primitives/test_dh.py
@@ -201,14 +201,6 @@ class TestDH(object):
         assert isinstance(deserialized_private,
                           dh.DHPrivateKeyWithSerialization)
 
-    def test_numbers_unsupported_parameters(self, backend):
-        params = dh.DHParameterNumbers(23, 2)
-        public = dh.DHPublicNumbers(1, params)
-        private = dh.DHPrivateNumbers(2, public)
-
-        with pytest.raises(ValueError):
-            private.private_key(backend)
-
     @pytest.mark.parametrize("with_q", [False, True])
     def test_generate_dh(self, backend, with_q):
         if with_q:


alex:
I haven't follow the DH development process (mostly because DH is 9000 times more complex than "just call `X25519()`"). I'm considered this removes all safety checks from this API, making it more footgunny than required.

If `DH_check` prevents certain perfectly good DH values, can we send a PR to OpenSSL to fix that?

reaperhulk:
#3755 has several links including some conversations dating back to 2002 on the OpenSSL mailing list.

The OpenSSL check also won't check any generator other than 2, 3, or 5. According to #3364, OpenSSL itself uses a 2048-bit generator for 2048-bit DH. This returns `DH_UNABLE_TO_CHECK_GENERATOR` so we could just ignore that error and not check generators outside 2, 3, 5.

To resolve the specific example in #3755 we could alter our own custom `Cryptography_DH_check` to allow p % 24 to be either 11 or 23. Some rationale for why this is a reasonable thing to do is here: https://crypto.stackexchange.com/questions/12961/diffie-hellman-parameter-check-when-g-2-must-p-mod-24-11

reaperhulk:
After thinking about this a bit longer I'd like to pursue a less extreme approach (at least at first).

To resolve #3364 I'll modify this PR to make the generator check in numbers be `> 1` (since the generator is used in the equation g<sup>k</sup>mod p a generator of less than 2 is unacceptable) and have us ignore `DH_NOT_SUITABLE_GENERATOR` for other values passed. This will also allow us to add the NIST vectors for testing (which use 512-bit and 1024-bit generators).

As a follow up to resolve #3755 I'll put in a PR to allow 23 as a p % 24 result. We'll CC lvh on that to get some confirmation that this is acceptable, but as far as I can tell it should be fine since it's just a choice between leaking a bit of the secret exponent (which OpenSSL chooses to do) or have your parameters only capable of generating half the possible values in Z<sup>*</sup><sub>p</sub>. Those two seem logically equivalent, but I'd like to hear a cryptographer confirm that :smile:

reaperhulk:
Closing in favor of #3768 (and a possible future PR)

PR #3760, opened 2017-07-07
explicitly pull the latest docker image in jenkins

reaperhulk:
Right now we don't need to do this since the same jenkins instance that builds the images and tags them also pushes them, so it is guaranteed to always have the latest. However, if we want to add another docker builder that is no longer true, so let's always pull.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -306,7 +306,7 @@ for (config in configs) {
             builders[combinedName] = {
                 node(label) {
                     stage(combinedName) {
-                        buildImage = docker.image(imageName)
+                        def buildImage = docker.image(imageName)
                         buildImage.pull()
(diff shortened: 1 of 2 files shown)


reaperhulk:
(This will fail until #3759 is merged)

alex:
@reaperhulk can you plesae rebase?

Jenkinsfile
```diff
@@ -306,7 +306,9 @@ for (config in configs) {
             builders[combinedName] = {
                 node(label) {
                     stage(combinedName) {
-                        docker.image(imageName).inside {
+                        buildImage = docker.image(imageName)
```
alex: Doh. Please put a `def` on this assignment. And then do a PR to all of the others fixing this :-(

PR #3763, opened 2017-07-08
Attempt to use coverage to quantify how many of the CONDITIONAL_NAMES we really use

alex:
The last time I sent this, we used it just as a temporary run. Now I want to propose we really use it :-)

### DIFF
--- a/docs/development/c-bindings.rst
+++ b/docs/development/c-bindings.rst
@@ -167,14 +167,21 @@ the necessarily type definitions are in place.
 Finally, add an entry to ``CONDITIONAL_NAMES`` with all of the things
 you want to conditionally export::
 
-    CONDITIONAL_NAMES = {
-        ...
-        "Cryptography_HAS_QUANTUM_TRANSMOGRIFICATION": [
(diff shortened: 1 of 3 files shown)


alex:
@reaperhulk `Cryptography_HAS_DTLS` and `Cryptography_HAS_EC2M` are apparently unused in CI. Objections to removing them?

reaperhulk:
According to the changelog `no-ec2m` was only added in 1.1.0 so it seems unlikely that anybody actually uses it (why is it even an option...). No DTLS might be somewhere, but I guess the only way we're going to learn about it is to drop support and see what happens. So, 👍 on removing both.

alex:
I'll do seperate PRs 👍

alex:
So, for whatever reason, codecov is not finding a coverage report on the CentOS7 builder, leading to this issue. wat.

@reaperhulk any clue why it fails?

reaperhulk:
Well that's bizarre. No idea what's going on. I guess we can pull down the image and run it ourselves to see what happens though.

PR #3766, opened 2017-07-08
Remove conditionals we never use.

alex:
Refs #3763


alex:
CentOS 7 appears to not have the EC2M stuff. I'm confused why it didn't show up on the original coverage report, but I'll put it back for now.

PR #3769, opened 2017-07-08
Fixed #3747 -- cache extensions on x.509 objects

### LINKED ISSUE #3747: Consider caching parsed extensions in x509.Certificates
Right now they get re-parsed on every access, this can be somewhat expensive, we should consider caching them on first access

alex:


reaperhulk:
I think this is worthy of a changelog entry since it has come up a few times previously (including during the work for CRLs). This also reminds me why I like Python. This decorator is elegant 🎉

PR #3770, opened 2017-07-09
remove cryptodev

reaperhulk:


src/cryptography/hazmat/bindings/openssl/_conditional.py
```diff
@@ -63,12 +63,6 @@ def cryptography_has_set_ecdh_auto():
     ]
 
 
-def cryptography_has_engine_cryptodev():
```
alex: dic tbelow

src/_cffi_src/openssl/engine.py
```diff
@@ -9,8 +9,6 @@
 """
 
 TYPES = """
-static const long Cryptography_HAS_ENGINE_CRYPTODEV;
```
alex: Should we leave this constant but set to `0`? Then there's no plausible backwards compat concern :-)

PR #3771, opened 2017-07-09
remove egd

reaperhulk:
EGD is bad and old and we removed it from pyOpenSSL in 16.0.

### DIFF
--- a/src/_cffi_src/openssl/rand.py
+++ b/src/_cffi_src/openssl/rand.py
@@ -9,7 +9,6 @@ INCLUDES = """
 """
 
 TYPES = """
-static const long Cryptography_HAS_EGD;
 """
 
 FUNCTIONS = """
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/bindings/openssl/_conditional.py
```diff
@@ -75,14 +75,6 @@ def cryptography_has_rsa_r_pkcs_decoding_error():
     ]
 
 
-def cryptography_has_egd():
```
alex: Remove from the dict below as well

src/_cffi_src/openssl/rand.py
```diff
@@ -9,7 +9,6 @@
 """
 
 TYPES = """
-static const long Cryptography_HAS_EGD;
```
alex: Should we leave this constant but set to `0`? Then there's no plausible backwards compat concern :-)

reaperhulk: Sure, why not. At some point it'd be good to remove it entirely but the two lines aren't going to be a maintenance burden in any way...

alex:
Landing despite the coverage because it appears to be a bug.

PR #3776, opened 2017-07-12
add a key derivation recipe

ofek:
Based on feedback from @dstufft's #3082, I put together a recipe for deriving keys from other keys or passwords. See https://github.com/pyca/cryptography/pull/3776/files#diff-148bf1cd66fb628e91d5ab3144654b29

Original participants: @alex @reaperhulk @tuxxy @lmctv

I'd definitely like feedback on what the default number of iterations for PBKDF2-HMAC should be.


reaperhulk:
Sorry, I was maybe not super clear about what I wanted to see for us to be able to move forward on this. I think we need to see a spec (ala fernet) that concretely describes everything about the recipe. What it's for, why each cryptographic primitive was chosen, explanation of parameters, and a versioning scheme that allows for migration over time.

Since the goal of this project is not to invent things (except where necessary) I'd also like to see a survey of the existing landscape to see if something like this already exists, and, if not, why not? Is there a good reason (e.g. the requirements end up being so specific that there's no use in a generic recipe) or is it more like the Fernet case where you choose some tradeoffs (effectively restricted message size) and gain significant safety in return.

tuxxy:
I like this idea a lot, but I agree with @reaperhulk.
I am absolutely interested in being a part of a project to get this specced out and defined. If this is something that people want (and I think it is), we can setup some communications channels and get to work on defining a spec for this recipe rather than just throwing some code together.

Again, I love where this is going, but I do agree that it needs some specification.

ofek:
Ok, I guess I could give that a shot.

@reaperhulk To be clear, to be considered for the recipe layer, a recipe should only handle cryptographically strong keys/salts if we provide the randomness for them?

reaperhulk:
I think that's probably going to be the position of this project, but you can debate it in the spec and see what people think. For now I'm going to close this PR since there's quite a bit of spec work to be done before I can review any code.

Issue #1199, opened 2014-06-30
Our GCM API exposes un-authenticated plaintext


alex:
That's like... bad. We should probably come up with an AEAD API. This also times into #1141, since arguably streaming makes no sense for authenticated stuff.

https://www.imperialviolet.org/2014/06/27/streamingencryption.html

money quote: """I will even claim that the existance of an API that can operate in a streaming fashion over large records (i.e. will encrypt and defer the authenticator and will decrypt and return unauthenticated plaintext) is a mistake."""

We'd need to design a new API, and deprecate the existing one.

simo5:
I'm curious,
why not simply use a chunked Fernet like construction where the MAC/digest of the previous chunk is used as input with the next chunk (to chain together all the parts) ? AEAD algorithms can handle this case very well.
Chunk Size can be defined in a (signed) header at the top of the file/stream, the header is he only special chunk and it is chained into the first chunk by its digest as well.
The key may be re-derived for each chunk or not.

reaperhulk:
I have a nebulous plan to extract the authenticated encryption construction in Tahoe-LAFS and adapt it into a recipe (CTR + a merkle tree to be able to rapidly auth arbitrary bytes and make it seekable), but I haven't had a chance to write that up.

Fernet isn't a great candidate for a chunked/streaming API since it isn't seekable due to the use of CBC.

simo5:
Do you want seekability in order to allow access to stored files and not just have a streaming format ?

reaperhulk:
I'm going to be comfortable closing this issue once we have an AESGCM AEAD class since we can link in the docs to have users use the one shot APIs by preference and only fall back to the nasty streaming API if absolutely necessary.

alex:
We now have an AES-GCM AEAD, so declaring this closed. Maybe some day we can think about deprecating the streaming API.

PR #3785, opened 2017-07-16
add AESGCM AEAD support

### LINKED ISSUE #3784: Implement One Shot AESGCM
Much like our ChaCha20Poly1305 and AESCCM classes we should do AESGCM. Then we can update our streaming GCM docs to point at the one shot as a better option if you aren't dealing with large volumes of data.

reaperhulk:
Fixes #3784 

I decided to name it `nonce` this time. This makes us inconsistent with the arg naming on the `GCM` mode itself, but I think we should call it a `nonce` since that's what it is. NIST's own GCM document calls it an initialization vector but then ALSO says it is synonymous with nonce.

### DIFF
--- a/docs/hazmat/primitives/aead.rst
+++ b/docs/hazmat/primitives/aead.rst
@@ -161,3 +161,74 @@ also support providing integrity for associated data which is not encrypted.
             doesn't validate this exception will be raised. This will occur
             when the ciphertext has been changed, but will also occur when the
             key, nonce, or associated data are wrong.
+
+.. class:: AESGCM(key)
+
+    .. versionadded:: 2.0
+
+    The AES-GCM construction is composed of the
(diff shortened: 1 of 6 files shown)


reaperhulk:
This is unfortunately going to add about 20-30s to each test run and some indeterminate amount of RAM consumed

docs/hazmat/primitives/aead.rst
```diff
@@ -161,3 +161,74 @@ also support providing integrity for associated data which is not encrypted.
             doesn't validate this exception will be raised. This will occur
             when the ciphertext has been changed, but will also occur when the
             key, nonce, or associated data are wrong.
+
+.. class:: AESGCM(key)
```
alex: Please document this _above_ CCM

reaperhulk:
This also fixes #1199

PR #3795, opened 2017-07-19
don't use pyenchant 1.6.9

reaperhulk:


alex:
For the record:
a) Why
b) Is there an upstream bug?

reaperhulk:
This is an experiment to see if it unbreaks our CI (macOS builder failure on TravisCI only with 1.6.9 right now). I have no idea if it's a wheel issue or a bug in 1.6.9 or completely unrelated to pyenchant at this point.

PR #3796, opened 2017-07-19
Switch to using the other style of environment markers

alex:


alex:
Note to self: branch-free setup.py is a prerequisite for shipping py3-stable-abi wheels.

alex:
@dstufft uhhh, any idea why this would result in a package not getting installed under `setup.py test`?

dstufft:
I know very little about ``setup.py test`` and how it resolved dependencies.

alex:
@jaraco -- @dstufft suggested you might have some idea why `setup.py test` is not installing things from `extra_requires` with install markers?

alex:
Chart of "Percent of cryptography installs with pip that are pip == 1.x by month":

alex:
@reaperhulk looking at the first 2 days of the month of august, we're down to 5.9% of installs being with pip 1.x

alex:
Now with a week of data, it's down to 5.5%:

```sql
SELECT
  STRFTIME_UTC_USEC(timestamp, "%Y-%m") AS yyyymm,
  ROUND(100 * SUM(CASE WHEN REGEXP_MATCH(details.installer.version, r"^1\.") THEN 1 ELSE 0 END) / COUNT(*), 1) AS percent_pip1,
  COUNT(*) as download_count
FROM
  TABLE_DATE_RANGE(
    [the-psf:pypi.downloads],
    DATE_ADD(CURRENT_TIMESTAMP(), -1, "year"),
    CURRENT_TIMESTAMP()
  )
WHERE
  details.installer.name == "pip" AND file.project == "cryptography"
GROUP BY
  yyyymm
ORDER BY
  yyyymm DESC
LIMIT 100
```

reaperhulk:
What does this do to users who install via setuptools rather than via pip?

alex:
setuptools version *mumble mumble* and above work fine.

alex:
(IIRC, it's version 0.7, incidentally the same version that started verifying shit in TLS, so I don't care about 0.6 _at all_)

alex:
https://github.com/pypa/setuptools/blob/master/CHANGES.rst#07

reaperhulk:
So the setuptools version is actually substantially older than the pip version. 4 years to pip 6's coming-up-on-3

alex:
I believe that's correct, yes.

alex:
@reaperhulk breaking setuptools 0.6 is arguably an important security measure :-)

reaperhulk:
Running that query gives 6.4% for August now. I think I'm convinced, especially since this won't affect setuptools unless you're on a ludicrously old version and all that happens in the worst case is some deps don't get installed.

PR #3797, opened 2017-07-19
hack hack hack

alex:


alex:
@reaperhulk so, uh, pinning pyenchant appears to fix it after all, but your PR came after the next release was out.

reaperhulk:
😠 well I merged your version of that pinning :)

PR #3800, opened 2017-07-20
We don't ever import cffi at runtime, I wonder if it's actually required...

alex:


alex:
Nope nope nope, definitely required.

PR #3805, opened 2017-07-21
simplify pyenchant requirement

alex:


alex:
I don't think any of them do. And also if they're shipping the very latest of cryptography, it shouldn't be a problem to do the same for pyenchant :-)

Issue #3812, opened 2017-07-26
 cryptography on OpenBSD 6.1


H6pOJTyp:
If you're filing a bug (as opposed to a feature request), please try the
following things:

* Upgrade to the latest version of ``setuptools`` and ``pip``
* Make sure you're on a supported version of OpenSSL
* Try with the latest version of ``cryptography``

If none of that works, please make sure to include the following information in
your bug report:

* Versions of Python, ``cryptography``, ``cffi``, ``pip``, and ``setuptools``
  you're using
* How you installed ``cryptography``
* Clear steps for reproducing your bug

H6pOJTyp:
cryptography on OpenBSD 6.1

The build fails with 
  cc1: error: -Werror=sign-conversion: No option -Wsign-conversion
After removing the option from src/_cffi_src/build_openssl.py, line 50,
it builds, but tests (with tahoe-lafs) show undefined symbols in hazmat/bindings/_openssl.so:
  __builtin_unreachable
  dev_urandom_read
The first is a non-standard gcc internal which appeared in gcc 4.5.
The second is strange, because OpenBSD has getentropy, so there should be no
calls to  dev_urandom_read at all.
I defined __builtin_unreachable as a NOOP, and dev_urandom_read as a function that immediately
aborts. Tahoe-lafs runs without problems after that, but this is no fix.

Versions:
cryptography 1d1653d3617ea826bb24a98bcea7eea587354809
Python 2.7.13
cffi 1.10.0
pip 9.0.1
setuptools 36.2.2
openbsd 6.1 stable amd64
gcc 4.2.1

reaperhulk:
Looks like we have a bug on the `CRYPTOGRAPHY_OSRANDOM_NEEDS_DEV_URANDOM` path with OpenBSD. `CRYPTOGRAPHY_OSRANDOM_ENGINE_GETENTROPY` is defined properly, but that path actually contains a urandom fallback to accommodate runtime detection of `SYS_getentropy` on macOS (https://github.com/pyca/cryptography/blob/master/src/_cffi_src/openssl/src/osrandom_engine.c#L211-L215). In the header we don't define `CRYPTOGRAPHY_OSRANDOM_NEEDS_DEV_URANDOM` for OpenBSD though (https://github.com/pyca/cryptography/blob/master/src/_cffi_src/openssl/src/osrandom_engine.h#L50-L55).

[...]

PR #3814, opened 2017-07-26
fix compilation under openbsd

reaperhulk:
refs #3812


alex:
We discussed on IRC that it'd be better to `#ifdef` out calling the dev-urandom fallback on OpenBSD

PR #3826, opened 2017-07-27
changelog and version bump for 2.0.2

reaperhulk:
This obviously should be merged after the backport for #3825

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+2.0.2 - 2017-07-27
+~~~~~~~~~~~~~~~~~~
+
+* Marked all symbols as hidden in the ``manylinux1`` wheel to avoid a
+  bug with symbol resolution in older Linux distributions.
+
 2.0.1 - 2017-07-26
 ~~~~~~~~~~~~~~~~~~
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,7 +5,7 @@ Changelog
 ~~~~~~~~~~~~~~~~~~
 
 * Marked all symbols as hidden in the ``manylinux1`` wheel to avoid a
-  bug with symbol resolution in older Linux distributions.
+  bug with symbol resolution in certain scenarios.
 
 2.0.1 - 2017-07-26
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+2.0.2 - 2017-07-27
```
alex: I guess it's the 27th somewhere...

CHANGELOG.rst
```diff
@@ -1,6 +1,12 @@
...
+~~~~~~~~~~~~~~~~~~
+
+* Marked all symbols as hidden in the ``manylinux1`` wheel to avoid a
+  bug with symbol resolution in older Linux distributions.
```
alex: The bug isn't with older Linux distributions, it just happens to materialize with a python shipped by an older distro.

I don't know what the right thing to write here is.

reaperhulk: "Symbol resolution in certain scenarios" 😛

alex: Sounds fine :D

srikalyan:
When is this going to be released? 2.0.1 is causing segfaults on ubuntu 14.04 :(

reaperhulk:
@srikalyan very shortly. We haven't been able to reproduce this on 14.04 with system Python, but we were able to reproduce on 12.04 and confirm this fixes it.

srikalyan:
I can confirm that this is broken on 14.04.5 (fresh system). I will let you know if the fix works on the 14.04.5

reaperhulk:
@srikalyan if you can replicate it right now would you mind trying https://f001.backblazeb2.com/file/dropshare-fenrir/cryptography-2.0.1-cp27-cp27mu-linux_x86_64.whl (that's a 2.0.1 wheel built with the fix). I'd like to confirm this works before releasing 2.0.2

srikalyan:
Sure give me like an hour?

reaperhulk:
I started the 2.0.2 release process already so it'll be up shortly. If 2.0.2 doesn't resolve it please file a new issue with a full reproducer so we can figure out the next level of this rabbit hole 😄

srikalyan:
I have already got 2.0.2 :) and it seems to fix the problem. Thank you everyone.

Issue #3357, opened 2017-01-19
Deprecate auto-idna in x509 layer


reaperhulk:
Trying to process IDNA within our certificate code was a mistake and has resulted in significant increases (rather than the desired decrease) in complexity for some of our downstream dependencies (see: https://github.com/kennethreitz/requests/blob/362da46e9a46da6e86e1907f03014384ab210151/requests/packages/urllib3/contrib/pyopenssl.py#L141). I'd like to see it removed, but let's talk a bit about whether we can actually achieve that goal.

Our current API requires that users do something like `x509.DNSName(u"mydömain.com")`. We also do this when parsing a certificate (we obtain the bytes, do some processing, `idna.decode` the parts we think we should, then pass the resulting string to the appropriate `GeneralName` class).

One potential path to removal would be to allow `DNSName` (and similar classes) to take bytes as well as unicode strings:

```python
class DNSName(object):
    def __init__(self, value):
        if isinstance(value, six.binary_type):
            self._value = decode_for_compat(value)
            self._bytes_value = value
        elif isinstance(value, six.text_type):
            self._value = value
            self._bytes_value = encode_to_bytes(value)
        else:
            raise TypeError("value must be a unicode string or bytes")

    value = utils.read_only_property("_value")
    bytes_value = utils.read_only_property("_bytes_value")
```

This approach preserves compatibility, but results in a confusing API that allows users on Python 2 to set values like "dömain" as bytes when they really wanted unicode. (Although we could potentially mitigate this by calling `.encode("ascii")` on `str` in Python 2.)

Once that's done we could deprecate unicode construction and at the same time start raising a `UserWarning` that `value` will be changing to match `bytes_value` in the future. Then once the deprecation time is complete we'd drop support for unicode and make `value` an alias to `bytes_value` and potentially deprecate `bytes_value` at that time. This is a major breaking change so I would be in favor of a year long deprecation cycle if we choose to pursue it.

alex:
- [ ] `DNSName`
- [ ] `RFC822Name`
- [ ] `UniformResourceIdentifier`

tiran:
I have been struggling with similar issues in Python's ssl module. The current implementation cannot verify and match IDN host names. It's also a bit of a mess, SNI callback and server_hostname are IDN U-labels, cert attributes are IDN A-labels. I have played with several approaches to fix the issue. So far only one approach is both simple enough to be memorable and not a potential source of security issues. It's also fully backwards compatible with ASCII-only host names.

[...]

PR #3830, opened 2017-07-29
Begin the deprecation of auto-idna for x509.DNSName

alex:
Refs #3357

### DIFF
--- a/tests/test_x509_ext.py
+++ b/tests/test_x509_ext.py
@@ -1476,6 +1476,11 @@ class TestDNSName(object):
         with pytest.warns(utils.DeprecatedIn21):
             name = x509.DNSName(u".\xf5\xe4\xf6\xfc.example.com")
         assert name.bytes_value == b".xn--4ca7aey.example.com"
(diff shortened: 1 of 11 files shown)


tests/test_x509_ext.py
```diff
@@ -2220,16 +2275,20 @@ def test_other_name(self, backend):
         assert othernames == [expected]
 
     def test_certbuilder(self, backend):
-        sans = [u'*.example.org', u'*.\xf5\xe4\xf6\xfc.example.com',
-                u'foobar.example.net']
+        sans = [b'*.example.org', b'*.\xf5\xe4\xf6\xfc.example.com',
```
reaperhulk: This should probably be `b'*.xn--4ca7aey.example.com'` or just be removed from the SAN list?  Previously this tested that it was encoding idna properly, but now they're all just bytes being passed along (and we're passing non-ascii bytes here, so we're sort of testing whether OpenSSL will tolerate such garbage)

alex: fixed

PR #3832, opened 2017-08-01
deprecate auto-idna on UniformResourceIdentifier

reaperhulk:
refs #3357

### DIFF
--- a/src/cryptography/x509/general_name.py
+++ b/src/cryptography/x509/general_name.py
@@ -180,16 +180,18 @@ class UniformResourceIdentifier(object):
                 value = self._idna_encode(value)
                 warnings.warn(
                     "UniformResourceIdentifier values should be passed as "
-                    "idna-encoded bytes, not strings. Support for passing "
-                    "unicode strings will be removed in a future version.",
+                    "bytes with the hostname idna encoded, not strings. "
+                    "Support for passing unicode strings will be removed in a "
+                    "future version.",
                     utils.DeprecatedIn21,
                     stacklevel=2,
                 )
             else:
                 warnings.warn(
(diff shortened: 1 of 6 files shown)


src/cryptography/x509/general_name.py
```diff
@@ -173,13 +173,34 @@ def __ne__(self, other):
...
-        if not isinstance(value, six.text_type):
-            raise TypeError("value must be a unicode string")
+        if isinstance(value, six.text_type):
+            try:
+                value = value.encode("ascii")
+            except UnicodeEncodeError:
+                value = self._idna_encode(value)
+                warnings.warn(
+                    "UniformResourceIdentifier values should be passed as "
+                    "idna-encoded bytes, not strings. Support for passing "
```
alex: `idna-encoded` is an over simplification right? the hostname portion should be idna-encoded, but the path segment should use the urlquote stuff, right?

src/cryptography/x509/general_name.py
```diff
@@ -173,13 +173,34 @@ def __ne__(self, other):
...
+                    "unicode strings will be removed in a future version.",
+                    utils.DeprecatedIn21,
+                    stacklevel=2,
+                )
+            else:
+                warnings.warn(
+                    "UniformResourceIdentifier values should be passed as "
+                    "idna-encoded bytes, not strings. Support for passing "
```
alex: same

src/cryptography/x509/general_name.py
reaperhulk: Yeah, maybe we just document it as "URIs should be passed as bytes with the hostname idna encoded"?

alex: Sounds right.

PR #3841, opened 2017-08-03
Fix weak linking of getentropy when compiling on older macOS

reaperhulk:
We use weak linking in macOS to determine if the getentropy symbol is
available. However, to do that we need to have a declaration that states
the function is __attribute((weak_import)) at compile time. On macOS
10.12 this is provided in sys/random.h, but on older macOS the
declaration doesn't exist at all, so we need to forward declare it
ourselves.

refs #3332

### DIFF
--- a/src/_cffi_src/openssl/src/osrandom_engine.h
+++ b/src/_cffi_src/openssl/src/osrandom_engine.h
@@ -13,6 +13,9 @@
 
   #ifdef __APPLE__
     #include <sys/random.h>
+    /* To support weak linking we need to declare this as a weak import even if
+     * it's not present in sys/random. */
+    extern int getentropy(void* buffer, size_t size) __attribute((weak_import));
   #endif
 
   #ifdef __linux__
--- a/src/_cffi_src/openssl/src/osrandom_engine.h
+++ b/src/_cffi_src/openssl/src/osrandom_engine.h
@@ -14,8 +14,8 @@
   #ifdef __APPLE__
     #include <sys/random.h>
     /* To support weak linking we need to declare this as a weak import even if
-     * it's not present in sys/random. */
-    extern int getentropy(void* buffer, size_t size) __attribute((weak_import));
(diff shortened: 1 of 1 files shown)


src/_cffi_src/openssl/src/osrandom_engine.h
```diff
@@ -13,6 +13,9 @@
 
   #ifdef __APPLE__
     #include <sys/random.h>
+    /* To support weak linking we need to declare this as a weak import even if
+     * it's not present in sys/random. */
+    extern int getentropy(void* buffer, size_t size) __attribute((weak_import));
```
alex: niiiiit: move the `*` in `void *buffer`

src/_cffi_src/openssl/src/osrandom_engine.h
```diff
@@ -13,6 +13,9 @@
 
   #ifdef __APPLE__
     #include <sys/random.h>
+    /* To support weak linking we need to declare this as a weak import even if
+     * it's not present in sys/random. */
```
alex: This comment isn't super clear. It should explicitly mention macOS < 10.12

alex:
@ilovezfs, will you want us to backport and do a release, or are you ok just cherry-picking this patch?

ilovezfs:
The fix works in my testing. Thanks!

>@ilovezfs, will you want us to backport and do a release, or are you ok just cherry-picking this patch?

A release, pretty please with sugar on top.

reaperhulk:
Okay, we'll backport this fix once it lands on master and do a 2.0.3.

ilovezfs:
@reaperhulk thank you :bowing_man:

JCount:
Thank you for the rapid fix! (Hopefully Apple will not further complicate things 😒)

reaperhulk:
This is now released in 2.0.3

JCount:
Thank you all! It is building on our CI now. 👍

PR #3849, opened 2017-08-06
Add is_signature_valid method on CertificateRevocationList

vpelletier:
Although this requires a parameter (unlike homonym on CSR), I believe such method is useful.


alex:
Thanks for submitting this!

It looks like there's a flake8 issue and a little bit of missing coverage (some of that is because of a typo in the test name: `text_` instead of `test_`). Can you take a look at those and then we'll get this a complete review!

vpelletier:
Thanks for the early review. Latest push should fix both issues, plus accidentaly disabled test hid a wrong filename (but otherwise, mechanism looks working).

reaperhulk:
Still one flake8 error (https://travis-ci.org/pyca/cryptography/jobs/261545711)

vpelletier:
Sorry for missing it. Also, latest push should fix the coverage decrease.

Side note about this PR: I initially thought about wrapping `X509_verify` in the same PR, but when I realised I would anyway need to pass the (immediate parent) CA certificate, I thought that rather than adding this it would seem better to directly wrap `X509_verify_cert`, which I expect would supersede `X509_verify`.

And side-tracking a bit further: the absence of `X509_verify_cert` wrapping is the last reason for me to use `pyOpenSSL` in my project (I need to validate it when receiving freshly issued/renewed cert before storing it, so without a SSL connection using it). Is there a reason for it not to be exposed via `cryptography` ?

PR #3861, opened 2017-08-12
add changelog and a warning for is_signature_valid on crl

reaperhulk:


CHANGELOG.rst
```diff
@@ -32,6 +32,8 @@ Changelog
   :attr:`~cryptography.x509.RFC822Name.value` attribute was deprecated, users
   should use :attr:`~cryptography.x509.RFC822Name.bytes_value` to access the
   raw value.
+* Added :attr:`~cryptography.x509.CertificateRevocationList.is_signature_valid`
```
alex: Wait, this shouldn't be attr, it should be meth.

reaperhulk: Ugh, we made it a property on CSR and a method on CRL then. Should we just change it to a property on CRL for consistency?

alex: How could you make it a property, don't you need to pass the issuers public key as an argument?

alex: CSR works because it's always self-signed.

reaperhulk: Indeed you do. Maybe I shouldn't suggest things right after waking up.

PR #3862, opened 2017-08-12
change x509 validation to allow parsing of some "bad" certs

### LINKED ISSUE #3857: Support non-standard X509 country names
I used the cryptography library to parse android APK certificates in androguard.
But there are some people, who create their certificates in a non standard way. Usually they do not put the two char country code, but instead use some string.
Here is an example of such an certificate: [ASN1. JavaScript Decoder with loaded Certificate](http://lapo.it/asn1js/#3082039506092A864886F70D010702A082038630820382020101310B300906052B0E03021A0500300B06092A864886F70D010701A08202D0308202CC30820289A003020102020466EDA548300B06072A8648CE380403050030363110300E06035504061307736961766173683110300E060355040A1307736961766173683110300E06035504031307736961766173683020170D3136303132353037323533305A180F32303534303532353037323533305A30363110300E06035504061307736961766173683110300E060355040A1307736961766173683110300E0603550403130773696176617368308201B83082012C06072A8648CE3804013082011F02818100FD7F53811D75122952DF4A9C2EECE4E7F611B7523CEF4400C31E3F80B6512669455D402251FB593D8D58FABFC5F5BA30F6CB9B556CD7813B801D346FF26660B76B9950A5A49F9FE8047B1022C24FBBA9D7FEB7C61BF83B57E7C6A8A6150F04FB83F6D3C51EC3023554135A169132F675F3AE2B61D72AEFF22203199DD14801C70215009760508F15230BCCB292B982A2EB840BF0581CF502818100F7E1A085D69B3DDECBBCAB5C36B857B97994AFBBFA3AEA82F9574C0B3D0782675159578EBAD4594FE67107108180B449167123E84C281613B7CF09328CC8A6E13C167A8B547C8D28E0A3AE1E2BB3A675916EA37F0BFA213562F1FB627A01243BCCA4F1BEA8519089A883DFE15AE59F06928B665E807B552564014C3BFECF492A0381850002818100E9B46A500AABCBB7B42E93477DB9F8AC7C819EADD392F4FA2DD7C10973C399D00D45796052BFB376567A518CFE1432A0F845E4BE75AA76462335BF7F84B6C634BB6C884B6AE2E7B6F1CC78C3BE6F98B65980832C40D9DA3D2B047CBA98B24C877B2ACE7908B0AF56FD62367DF355E011A0482C9B8751C373A8D131076CD01421A321301F301D0603551D0E04160414EEAC32E5F08781BB6F2F392814951B328868AE4C300B06072A8648CE3804030500033000302D0215009209EB12E966AAFAE281620E7DA9743774D7846302144BF36A9634967CA47639BBC10712571B036E13D531818E30818B020101303E30363110300E06035504061307736961766173683110300E060355040A1307736961766173683110300E0603550403130773696176617368020466EDA548300906052B0E03021A0500300B06072A8648CE3804010500042E302C02145DE23A86F0C2E45DCAD93EF8CAB675EC56A6FF4B021404C1A9F7F791631AE00D61FE8AE23F502C33FE84).
As you can see, someone used `siavash` as the country name.
From my point of view, such certificates are not common but also not very rare. I found out, that creating a new signing key in Android Studio does not reject the certificate if you enter something stupid in the country code field.

When you load the certificate (after removing the pkcs7 message around it), you will get an error message:
```
import binascii
from pyasn1.codec.der.decoder import decode
from pyasn1.codec.der.encoder import encode
from cryptography import x509

certhex  = b"3082039506092a864886f70d010702a082038630820382020101310b300906052b0e03021a0500300b06092a864886f70d010701a08202d0308202cc30820289a003020102020466eda548300b06072a8648ce380403050030363110300e06035504061307736961766173683110300e060355040a1307736961766173683110300e06035504031307736961766173683020170d3136303132353037323533305a180f32303534303532353037323533305a30363110300e06035504061307736961766173683110300e060355040a1307736961766173683110300e0603550403130773696176617368308201b83082012c06072a8648ce3804013082011f02818100fd7f53811d75122952df4a9c2eece4e7f611b7523cef4400c31e3f80b6512669455d402251fb593d8d58fabfc5f5ba30f6cb9b556cd7813b801d346ff26660b76b9950a5a49f9fe8047b1022c24fbba9d7feb7c61bf83b57e7c6a8a6150f04fb83f6d3c51ec3023554135a169132f675f3ae2b61d72aeff22203199dd14801c70215009760508f15230bccb292b982a2eb840bf0581cf502818100f7e1a085d69b3ddecbbcab5c36b857b97994afbbfa3aea82f9574c0b3d0782675159578ebad4594fe67107108180b449167123e84c281613b7cf09328cc8a6e13c167a8b547c8d28e0a3ae1e2bb3a675916ea37f0bfa213562f1fb627a01243bcca4f1bea8519089a883dfe15ae59f06928b665e807b552564014c3bfecf492a0381850002818100e9b46a500aabcbb7b42e93477db9f8ac7c819eadd392f4fa2dd7c10973c399d00d45796052bfb376567a518cfe1432a0f845e4be75aa76462335bf7f84b6c634bb6c884b6ae2e7b6f1cc78c3be6f98b65980832c40d9da3d2b047cba98b24c877b2ace7908b0af56fd62367df355e011a0482c9b8751c373a8d131076cd01421a321301f301d0603551d0e04160414eeac32e5f08781bb6f2f392814951b328868ae4c300b06072a8648ce3804030500033000302d0215009209eb12e966aafae281620e7da9743774d7846302144bf36a9634967ca47639bbc10712571b036e13d531818e30818b020101303e30363110300e06035504061307736961766173683110300e060355040a1307736961766173683110300e0603550403130773696176617368020466eda548300906052b0e03021a0500300b06072a8648ce3804010500042e302c02145de23a86f0c2e45dcad93ef8cab675ec56a6ff4b021404c1a9f7f791631ae00d61fe8ae23f502c33fe84"

pkcs7message = binascii.unhexlify(certhex)

message, _ = decode(pkcs7message)
cert = encode(message[1][3])
l = cert[1]
if not isinstance(l, int):
    l = ord(l)
cert = cert[2 + (l & 0x7F) if l & 0x80 > 1 else 2:]

certificate = x509.load_der_x509_certificate(cert, default_backend())

print(certificate.issuer)
print(certificate.fingerprint(hashes.SHA1()))
```

```
  File "Desktop\test.py", line 30, in <module>
    print(certificate.issuer)
  File "C:\Program Files\Anaconda3\lib\site-packages\cryptography\hazmat\backends\openssl\x509.py", line 101, in issuer
    return _decode_x509_name(self._backend, issuer)
  File "C:\Program Files\Anaconda3\lib\site-packages\cryptography\hazmat\backends\openssl\decode_asn1.py", line 65, in _decode_x509_name
    attribute = _decode_x509_name_entry(backend, entry)
  File "C:\Program Files\Anaconda3\lib\site-packages\cryptography\hazmat\backends\openssl\decode_asn1.py", line 56, in _decode_x509_name_entry
    return x509.NameAttribute(x509.ObjectIdentifier(oid), value)
  File "C:\Program Files\Anaconda3\lib\site-packages\cryptography\x509\name.py", line 27, in __init__
    "Country name must be a 2 character country code"
ValueError: Country name must be a 2 character country code
```

I know this is not really a bug, because the RFC says 2 char country code... But it would be nice to work with those "broken" certificates as well.

Versions (cryptography was installed via pip):
```
c:\>pip show cryptography
Name: cryptography
Version: 2.0.3
Summary: cryptography is a package which provides cryptographic recipes and primitives to Python developers.
Home-page: https://github.com/pyca/cryptography
Author: The cryptography developers
Author-email: cryptography-dev@python.org
License: BSD or Apache License, Version 2.0
Location: c:\program files\anaconda3\lib\site-packages
Requires: six, cffi, asn1crypto, idna

c:\>pip show cffi
Name: cffi
Version: 1.10.0
Summary: Foreign Function Interface for Python calling C code.
Home-page: http://cffi.readthedocs.org
Author: Armin Rigo, Maciej Fijalkowski
Author-email: python-cffi@googlegroups.com
License: MIT
Location: c:\program files\anaconda3\lib\site-packages
Requires: pycparser

c:\>pip show setuptools
Name: setuptools
Version: 27.2.0
Summary: Easily download, build, install, upgrade, and uninstall Python packages
Home-page: https://github.com/pypa/setuptools
Author: Python Packaging Authority
Author-email: distutils-sig@python.org
License: UNKNOWN
Location: c:\program files\anaconda3\lib\site-packages\setuptools-27.2.0-py3.6.egg
Requires:
c:\>python --version
Python 3.6.0 :: Anaconda 4.3.0 (64-bit)

c:\>pip --version
pip 9.0.1 from C:\Program Files\Anaconda3\lib\site-packages (python 3.6)
```

reaperhulk:
We now validate on these fields when adding the objects to the CertificateBuilder.

This PR does not attempt to migrate all validation logic. Each extension has logic that may be specific to Python or enforcement of an RFC restriction. Before migrating more logic I'd like to see examples of certificates that encode things in that specific invalid way.

Fixes #3857 and #3856

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -341,9 +341,6 @@ class BasicConstraints(object):
         if not isinstance(ca, bool):
             raise TypeError("ca must be a boolean value")
 
-        if path_length is not None and not ca:
-            raise ValueError("path_length must be None when ca is False")
-
         if (
             path_length is not None and
             (not isinstance(path_length, six.integer_types) or path_length < 0)
@@ -373,6 +370,10 @@ class BasicConstraints(object):
 
     def __hash__(self):
         return hash((self.ca, self.path_length))
+
+    def _validate(self):
+        if self.path_length is not None and not self.ca:
+            raise ValueError("path_length must be None when ca is False")
 
 
 @utils.register_interface(ExtensionType)
@@ -1140,6 +1141,13 @@ class Extension(object):
     oid = utils.read_only_property("_oid")
     critical = utils.read_only_property("_critical")
     value = utils.read_only_property("_value")
+
+    def _validate(self):
+        # Validate the extension if it has a _validate method. Not using
+        # a try/except here to avoid shadowing AttributeErrors that might
+        # occur when calling _validate
+        if getattr(self.value, "_validate", None) is not None:
+            self.value._validate()
 
     def __repr__(self):
         return ("<Extension(oid={0.oid}, critical={0.critical}, "
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,6 +29,7 @@ Changelog
 * Added support for :class:`~cryptography.hazmat.primitives.hashes.SHAKE128`
   and :class:`~cryptography.hazmat.primitives.hashes.SHAKE256` when using
   OpenSSL 1.1.1.
+* Allow parsing of more certificates that are not :rfc:`5280` compliant.
 * Added initial support for parsing PKCS12 files with
   :func:`~cryptography.hazmat.primitives.serialization.pkcs12.load_key_and_certificates`.
 * Added support for :class:`~cryptography.x509.IssuingDistributionPoint`.
--- a/src/cryptography/x509/base.py
+++ b/src/cryptography/x509/base.py
@@ -403,6 +403,8 @@ class CertificateSigningRequestBuilder(object):
             raise TypeError('Expecting x509.Name object.')
         if self._subject_name is not None:
             raise ValueError('The subject name may only be set once.')
+
+        name._validate()
         return CertificateSigningRequestBuilder(name, self._extensions)
 
     def add_extension(self, extension, critical):
@@ -414,6 +416,7 @@ class CertificateSigningRequestBuilder(object):
 
         extension = Extension(extension.oid, critical, extension)
         _reject_duplicate_extension(extension, self._extensions)
+        extension._validate()
 
         return CertificateSigningRequestBuilder(
             self._subject_name, self._extensions + [extension]
@@ -449,6 +452,8 @@ class CertificateBuilder(object):
             raise TypeError('Expecting x509.Name object.')
         if self._issuer_name is not None:
             raise ValueError('The issuer name may only be set once.')
+
+        name._validate()
         return CertificateBuilder(
             name, self._subject_name, self._public_key,
             self._serial_number, self._not_valid_before,
@@ -463,6 +468,8 @@ class CertificateBuilder(object):
             raise TypeError('Expecting x509.Name object.')
         if self._subject_name is not None:
             raise ValueError('The subject name may only be set once.')
+
+        name._validate()
         return CertificateBuilder(
             self._issuer_name, name, self._public_key,
             self._serial_number, self._not_valid_before,
@@ -563,6 +570,7 @@ class CertificateBuilder(object):
 
         extension = Extension(extension.oid, critical, extension)
         _reject_duplicate_extension(extension, self._extensions)
+        extension._validate()
 
         return CertificateBuilder(
             self._issuer_name, self._subject_name,
@@ -609,6 +617,8 @@ class CertificateRevocationListBuilder(object):
             raise TypeError('Expecting x509.Name object.')
         if self._issuer_name is not None:
             raise ValueError('The issuer name may only be set once.')
+
+        issuer_name._validate()
         return CertificateRevocationListBuilder(
             issuer_name, self._last_update, self._next_update,
             self._extensions, self._revoked_certificates
@@ -659,6 +669,8 @@ class CertificateRevocationListBuilder(object):
 
         extension = Extension(extension.oid, critical, extension)
         _reject_duplicate_extension(extension, self._extensions)
+        extension._validate()
+
         return CertificateRevocationListBuilder(
             self._issuer_name, self._last_update, self._next_update,
             self._extensions + [extension], self._revoked_certificates
@@ -733,6 +745,8 @@ class RevokedCertificateBuilder(object):
 
         extension = Extension(extension.oid, critical, extension)
         _reject_duplicate_extension(extension, self._extensions)
+        extension._validate()
+
         return RevokedCertificateBuilder(
             self._serial_number, self._revocation_date,
             self._extensions + [extension]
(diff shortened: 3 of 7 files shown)


tests/test_x509.py
```diff
@@ -3751,23 +3792,26 @@ def test_init_bad_value(self):
                 b'bytes'
             )
 
-    def test_init_bad_country_code_value(self):
-        with pytest.raises(ValueError):
-            x509.NameAttribute(
+    def test_bad_country_code_value(self):
```
reaperhulk: Should we keep these tests? This same logic is tested in `test_calls_validator` as well.

src/cryptography/x509/extensions.py
```diff
@@ -936,6 +937,13 @@ def __init__(self, oid, critical, value):
     critical = utils.read_only_property("_critical")
     value = utils.read_only_property("_value")
 
+    def _validate(self):
+        # Validate the extension if it has a _validate method
+        try:
+            self.value._validate()
+        except AttributeError:
```
alex: I don't like this, it silently consumes an `AttributeError` throw by `_validate()`

reaperhulk: Would you prefer every extension get a `_validate` and just have almost all of them be `pass`? Or do a `getattr`?

reaperhulk:
This is ready for review again. Still outstanding: whether we want to switch to adding `_validate()` to every extension class rather than the (new) getattr approach.

jcrowgey:
I have a dataset of about 20 certificates (and still growing) which break the validation steps in this library.  All were harvested in the wild.  If there's some way to publish these examples for you conveniently, I'd love to contribute.  Currently, I just monkey patch the __init__ methods on lots of the Extension and Name classes to remove the validation checks.

jcrowgey:
@2rs2ts what timing.  I just release (a few hours ago) the library I use which does the monkeypatching to allow loading of certs which don't always follow best practices.

https://github.com/jcrowgey/x5092json

2rs2ts:
@jcrowgey nice... I am not sure your library solves my problem, but it is indeed cool! 😄

reaperhulk:
This has been pushed 5 times now and is no less problematic now than it was 3 years ago. I think to get "bad" cert parsing we're going to need to have someone commit to the large amount of work necessary for a hazmat x509 parsing API. When that gets implemented we can have this.

In the meantime the monkeypatching model is probably the best path to bypass this type of validation if needed.

PR #3870, opened 2017-08-19
Remove branches for 1.0.2 betas.

alex:
There's no sense in which we actually support them


reaperhulk:
Are you also interested in doing this for 1.1.0? We have CRYPTOGRAPHY_OPENSSL_LESS_THAN_110PRE4, CRYPTOGRAPHY_OPENSSL_LESS_THAN_110PRE5, and CRYPTOGRAPHY_OPENSSL_LESS_THAN_110PRE6.

If memory serves we originally added these partially because that was when specific things were added, but also because we ran into some crazy situation where a distro was possibly using a beta version of 1.0.2? But it seems pretty reasonable to state that we do not support that behavior...

alex:
Yeah, I'll want to get the 1.1.0 pre ones as well, but wanted to do it separately.

My memory is that we did these as either paranoia, or because we really did add them during the beta period.

The only time I remember running into a distro which had packaged a beta something was from 1.0.1!

In any of these events, 1.0.2 is now several years old, if you haven't packaged a release...

PR #3871, opened 2017-08-20
Fix `load_rsa_public_numbers` documentation

dsanders11:
`load_rsa_public_numbers` takes an `RSAPublicNumbers` object as the name implies, not an `RSAPrivateNumbers`.


alex:
Thanks!

PR #3876, opened 2017-08-25
Remove legacy setup.py hacks

alex:
These dated back to when looking at setup.py the wrong way would compile all of our cffi modules. I don't believe this is the case any longer.


alex:
I don't understand what's going on with the py26 coverage.

reaperhulk:
cc @dstufft just to make sure we're not missing anything

alex:
Ok, after discussion we decided to hold off on this.

PR #3878, opened 2017-08-26
added binding support for rfc 5705

kelbyludwig:
I am working on a PR for pyopenssl that allows for exporting keying material and needed to expose this API.


kelbyludwig:
https://github.com/pyca/pyopenssl/pull/686 will fail without these updates to cryptography. if there is a smarter way of tackling this just let me know!

alex:
There's a few places where this deviates from our style guide: https://cryptography.io/en/latest/development/c-bindings/

Can you please fix those up?

(Also, for some reason we seem to be missing coverage, but that's not related to your patch. We'll need to fix that before this can be merged)

kelbyludwig:
@alex Thanks for pointing that guide out. I think my PR is aligned to those guidelines now.

src/_cffi_src/openssl/cryptography.py
```diff
@@ -25,6 +25,8 @@
 #include <windows.h>
 #endif
 
+#define CRYPTOGRAPHY_OPENSSL_101_OR_GREATER \
+    (OPENSSL_VERSION_NUMBER >= 0x1000100 && !CRYPTOGRAPHY_IS_LIBRESSL)
```
reaperhulk: We only support OpenSSL 1.0.1+ so this should always be true for OpenSSL. Additionally, it looks like `SSL_export_keying_material` is available in LibreSSL so I don't think you actually need to use a conditional binding at all here.

kelbyludwig: Makes sense. I assume that means I should remove the changes to `src/cryptography/hazmat/bindings/openssl/_conditional.py` as well?

reaperhulk: Yeah you should only need the function definition and none of the other boilerplate.

src/_cffi_src/openssl/ssl.py
```diff
@@ -520,6 +523,7 @@
 static const long Cryptography_HAS_SSL_OP_NO_TICKET = 1;
 static const long Cryptography_HAS_SSL_SET_SSL_CTX = 1;
 static const long Cryptography_HAS_NEXTPROTONEG = 1;
+static const long Cryptography_HAS_KEYING_EXPORT = 1;
```
reaperhulk: This (and the declaration) are also unnecessary now. Other than that this looks good!

kelbyludwig: you beat me to it :)

reaperhulk:
This LGTM now. Hopefully codecov will sort itself out once all the tests actually run 😄

reaperhulk:
Thanks for the contribution!

PR #3880, opened 2017-08-31
implement AES KW with padding (RFC 5649)

### LINKED ISSUE #3791: RFC 5649 support
RFC 3394 (AES Key Wrap) was added a while back. I'd like to request support for RFC 5649 (AES Key Wrap with Padding), since it builds off of RFC 3394. It looks like OpenSSL handled this back in 2015:

https://rt.openssl.org/Ticket/Display.html?id=3675&user=guest&pass=guest

Is this feasible for cryptography in the not-too-distant future?

Thanks,
Peter

reaperhulk:
fixes #3791

### DIFF
--- a/src/cryptography/hazmat/primitives/keywrap.py
+++ b/src/cryptography/hazmat/primitives/keywrap.py
@@ -76,14 +76,14 @@ def aes_key_wrap_with_padding(wrapping_key, key_to_wrap, backend):
     # pad the key to wrap if necessary
     pad = (8 - (len(key_to_wrap) % 8)) % 8
     key_to_wrap = key_to_wrap + b"\x00" * pad
-    r = [key_to_wrap[i:i + 8] for i in range(0, len(key_to_wrap), 8)]
-    if len(r) == 1:
+    if len(key_to_wrap) == 8:
         # RFC 5649 - 4.1 - exactly 8 octets after padding
         encryptor = Cipher(AES(wrapping_key), ECB(), backend).encryptor()
-        b = encryptor.update(aiv + r[0])
+        b = encryptor.update(aiv + key_to_wrap)
         assert encryptor.finalize() == b""
         return b
     else:
+        r = [key_to_wrap[i:i + 8] for i in range(0, len(key_to_wrap), 8)]
         return _wrap_core(wrapping_key, aiv, r, backend)
 
 
@@ -94,17 +94,18 @@ def aes_key_unwrap_with_padding(wrapping_key, wrapped_key, backend):
     if len(wrapping_key) not in [16, 24, 32]:
         raise ValueError("The wrapping key must be a valid AES key length")
 
-    r = [wrapped_key[i:i + 8] for i in range(0, len(wrapped_key), 8)]
-    encrypted_aiv = r.pop(0)
-    n = len(r)
-    if n == 1:
+    if len(wrapped_key) == 16:
         # RFC 5649 - 4.2 - exactly two 64-bit blocks
         decryptor = Cipher(AES(wrapping_key), ECB(), backend).decryptor()
-        b = decryptor.update(encrypted_aiv + r[0])
+        b = decryptor.update(wrapped_key)
         assert decryptor.finalize() == b""
         a = b[:8]
         data = b[8:]
+        n = 1
     else:
+        r = [wrapped_key[i:i + 8] for i in range(0, len(wrapped_key), 8)]
+        encrypted_aiv = r.pop(0)
+        n = len(r)
         a, r = _unwrap_core(wrapping_key, encrypted_aiv, r, backend)
         data = b"".join(r)
 
@@ -113,7 +114,7 @@ def aes_key_unwrap_with_padding(wrapping_key, wrapped_key, backend):
     #    MLI = LSB(32,A).
     # 3) Let b = (8*n)-MLI, and then check that the rightmost b octets of
     #    the output data are zero.
-    mli = struct.unpack(">I", a[4:])[0]
+    (mli,) = struct.unpack(">I", a[4:])
     b = (8 * n) - mli
     if (
         not bytes_eq(a[:4], b"\xa6\x59\x59\xa6") or not
--- a/src/cryptography/hazmat/primitives/keywrap.py
+++ b/src/cryptography/hazmat/primitives/keywrap.py
@@ -68,6 +68,63 @@ def _unwrap_core(wrapping_key, a, r, backend):
     return a, r
 
 
+def aes_key_wrap_with_padding(wrapping_key, key_to_wrap, backend):
+    if len(wrapping_key) not in [16, 24, 32]:
+        raise ValueError("The wrapping key must be a valid AES key length")
+
+    aiv = b"\xA6\x59\x59\xA6" + struct.pack(">i", len(key_to_wrap))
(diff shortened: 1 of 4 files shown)


src/cryptography/hazmat/primitives/keywrap.py
```diff
@@ -68,6 +68,63 @@ def _unwrap_core(wrapping_key, a, r, backend):
...
+        encryptor = Cipher(AES(wrapping_key), ECB(), backend).encryptor()
+        b = encryptor.update(aiv + r[0])
+        assert encryptor.finalize() == b""
+        return b
+    else:
+        return _wrap_core(wrapping_key, aiv, r, backend)
+
+
+def aes_key_unwrap_with_padding(wrapping_key, wrapped_key, backend):
+    if len(wrapped_key) < 16:
```
jlaine: Is this necessary since we assert the wrapping key is in [16, 24, 32] just below?

We don't do this length < 16 check for the wrapping function, so we should either do it for both wrap/unwrap or neither

tests/hazmat/primitives/test_keywrap.py
```diff
@@ -114,3 +114,68 @@ def test_unwrap_invalid_wrapped_key_length(self, backend):
...
+                )
+        else:
+            unwrapped_key = keywrap.aes_key_unwrap_with_padding(
+                wrapping_key, wrapped_key, backend
+            )
+            assert params["p"] == binascii.hexlify(unwrapped_key)
+
+    def test_unwrap_invalid_wrapped_key_length(self, backend):
+        # Keys to unwrap must be at least 16 bytes
+        with pytest.raises(ValueError):
```
jlaine: Should we check the exception's description to ensure we hit the right ValueError?

reaperhulk: Not a bad idea, especially since `pytest` turns out to have a nice `match` keyword for this

src/cryptography/hazmat/primitives/keywrap.py
```diff
@@ -68,6 +68,63 @@ def _unwrap_core(wrapping_key, a, r, backend):
...
 
+def aes_key_wrap_with_padding(wrapping_key, key_to_wrap, backend):
+    if len(wrapping_key) not in [16, 24, 32]:
+        raise ValueError("The wrapping key must be a valid AES key length")
+
+    aiv = b"\xA6\x59\x59\xA6" + struct.pack(">i", len(key_to_wrap))
+    r = [key_to_wrap[i:i + 8] for i in range(0, len(key_to_wrap), 8)]
+    pad = (8 - (len(r[-1]) % 8)) % 8
+    # Pad the last block if necessary
+    r[-1] = r[-1] + b"\x00" * pad
```
alex: This would be considerably easier to read if you did the padding on `key_to_wrap` and then converted to `r`.

src/cryptography/hazmat/primitives/keywrap.py
```diff
@@ -68,6 +68,63 @@ def _unwrap_core(wrapping_key, a, r, backend):
...
+    # 2) Check that 8*(n-1) < LSB(32,A) <= 8*n.  If so, let
+    #    MLI = LSB(32,A).
+    # 3) Let b = (8*n)-MLI, and then check that the rightmost b octets of
+    #    the output data are zero.
+    mli = struct.unpack(">I", a[4:])[0]
+    b = (8 * n) - mli
+    if (
+        not bytes_eq(a[:4], b"\xa6\x59\x59\xa6") or not
+        8 * (n - 1) < mli <= 8 * n or
+        data[-b:] != b"\x00" * b
```
alex: Should this be `bytes_eq`?

jlaine: Good point, not sure constant time comparison matters here. If it does, an (A or B or C) construction's not going to fly.

src/cryptography/hazmat/primitives/keywrap.py
```diff
@@ -68,6 +68,62 @@ def _unwrap_core(wrapping_key, a, r, backend):
...
+def aes_key_wrap_with_padding(wrapping_key, key_to_wrap, backend):
+    if len(wrapping_key) not in [16, 24, 32]:
+        raise ValueError("The wrapping key must be a valid AES key length")
+
+    aiv = b"\xA6\x59\x59\xA6" + struct.pack(">i", len(key_to_wrap))
+    # pad the key to wrap if necessary
+    pad = (8 - (len(key_to_wrap) % 8)) % 8
+    key_to_wrap = key_to_wrap + b"\x00" * pad
+    r = [key_to_wrap[i:i + 8] for i in range(0, len(key_to_wrap), 8)]
+    if len(r) == 1:
```
alex: Follow up, can we change this condition to be `if len(key_to_wrap) == 8`, and then move computing `r` into the `else` branch?

src/cryptography/hazmat/primitives/keywrap.py
```diff
@@ -68,6 +68,62 @@ def _unwrap_core(wrapping_key, a, r, backend):
...
+    if len(wrapped_key) < 16:
+        raise ValueError("Must be at least 16 bytes")
+
+    if len(wrapping_key) not in [16, 24, 32]:
+        raise ValueError("The wrapping key must be a valid AES key length")
+
+    r = [wrapped_key[i:i + 8] for i in range(0, len(wrapped_key), 8)]
+    encrypted_aiv = r.pop(0)
+    n = len(r)
+    if n == 1:
```
alex: Same thing here.

src/cryptography/hazmat/primitives/keywrap.py
```diff
@@ -68,6 +68,62 @@ def _unwrap_core(wrapping_key, a, r, backend):
...
+    else:
+        a, r = _unwrap_core(wrapping_key, encrypted_aiv, r, backend)
+        data = b"".join(r)
+
+    # 1) Check that MSB(32,A) = A65959A6.
+    # 2) Check that 8*(n-1) < LSB(32,A) <= 8*n.  If so, let
+    #    MLI = LSB(32,A).
+    # 3) Let b = (8*n)-MLI, and then check that the rightmost b octets of
+    #    the output data are zero.
+    mli = struct.unpack(">I", a[4:])[0]
```
alex: Use tuple unpacking, instead of indexing

src/cryptography/hazmat/primitives/keywrap.py
```diff
@@ -68,6 +68,63 @@ def _unwrap_core(wrapping_key, a, r, backend):
...
+    # 1) Check that MSB(32,A) = A65959A6.
+    # 2) Check that 8*(n-1) < LSB(32,A) <= 8*n.  If so, let
+    #    MLI = LSB(32,A).
+    # 3) Let b = (8*n)-MLI, and then check that the rightmost b octets of
+    #    the output data are zero.
+    (mli,) = struct.unpack(">I", a[4:])
+    b = (8 * n) - mli
+    if (
+        not bytes_eq(a[:4], b"\xa6\x59\x59\xa6") or not
+        8 * (n - 1) < mli <= 8 * n or not bytes_eq(data[-b:], b"\x00" * b)
```
alex: `8 * n` is just `len(data)` or something, right?

reaperhulk: It is. I think changing this makes it harder to understand since the comments (which are taken directly from the RFC) are no longer directly applicable, but I can do so if you want.

PR #3893, opened 2017-09-06
we pinned coverage for xml generation...let's try everywhere

reaperhulk:


alex:
@reaperhulk are you still interested in merging this now that coverage was otherwise resolved?

alex:
Looks like this isn't required.

PR #3899, opened 2017-09-08
[WIP] add support for the TLSFeature extension in x509

### LINKED ISSUE #3891: OCSP Must-Staple (TLS Feature extension)
We should support this. OpenSSL 1.1.0 explicitly supports this, but we could potentially just encode the extension ourselves (it's very simple) and then it would be supported regardless of version. Otherwise we'll have another scenario (like SCT) where the version of OpenSSL cryptography is compiled against influences what it's capable of parsing and generating. Opinions?

reaperhulk:
This extension is used for OCSP Must-Staple.

depends on #3894 (will rebase and add a new test demonstrating parsing in addition to the certificate signing/parsing we're doing now).

fixes #3891


docs/x509/reference.rst
```diff
@@ -1606,6 +1606,38 @@ X.509 Extensions
...
+
+        Returns :attr:`~cryptography.x509.oid.ExtensionOID.TLS_FEATURE`.
+
+.. class:: TLSFeatureType
+
+    .. versionadded:: 2.1
+
+    An enumeration of TLS Feature types.
+
+    .. attribute:: status_request
```
reaperhulk: RFC 7633 is a bit confusing, but it's possible that `status_request_v2` is also acceptable? Right now I believe LE will only issue with `status_request` though.

reaperhulk:
Rob Stradling noted that `status_request_v2` is technically permissible (and OpenSSL allows it: https://github.com/openssl/openssl/blob/master/crypto/x509v3/v3_tlsf.c#L51) so we'll have to be a bit fancier than the current WIP.

alex:
ready for a rebase.

src/cryptography/x509/__init__.py
```diff
@@ -59,6 +59,7 @@
 OID_SUBJECT_DIRECTORY_ATTRIBUTES = ExtensionOID.SUBJECT_DIRECTORY_ATTRIBUTES
 OID_SUBJECT_INFORMATION_ACCESS = ExtensionOID.SUBJECT_INFORMATION_ACCESS
 OID_SUBJECT_KEY_IDENTIFIER = ExtensionOID.SUBJECT_KEY_IDENTIFIER
+OID_TLS_FEATURE = ExtensionOID.TLS_FEATURE
```
alex: I don't think we do this for new oids

src/cryptography/x509/extensions.py
```diff
@@ -733,6 +733,46 @@ class OCSPNoCheck(object):
...
+
+        return self.features == other.features
+
+    def __ne__(self, other):
+        return not self == other
+
+    def __hash__(self):
+        return hash(tuple(self.features))
+
+    features = utils.read_only_property("_features")
```
alex: hmm, so conventionally we make the class iterable/subscriptable, instead of exposing the list as an attr

src/cryptography/x509/extensions.py
```diff
@@ -733,6 +733,46 @@ class OCSPNoCheck(object):
...
+
+
+class TLSFeatureType(Enum):
+    status_request = 5
```
alex: Comments describing the origins please

tests/x509/test_x509.py
```diff
@@ -2610,6 +2622,46 @@ def test_name_constraints(self, nc, backend):
 
     @pytest.mark.requires_backend_interface(interface=RSABackend)
     @pytest.mark.requires_backend_interface(interface=X509Backend)
+    @pytest.mark.parametrize(
+        "add_ext",
+        [
+            x509.TLSFeature([x509.TLSFeatureType.status_request]),
```
alex: empty list should be tested I guess

reaperhulk: This is done (but tested in test_x509_ext)

PR #3900, opened 2017-09-08
Add support for AES XTS

reaperhulk:
It's a widely used mode for disk encryption and it'd be nice if this library was usable for disk forensics why not.

We drop the non-byte aligned test vectors because according to NIST
http://csrc.nist.gov/groups/STM/cavp/documents/aes/XTSVS.pdf
"An implementation may support a data unit length that is not a
multiple of 8 bits." OpenSSL does not support this, so we can't
use those test vectors.

### DIFF
--- a/docs/hazmat/primitives/symmetric-encryption.rst
+++ b/docs/hazmat/primitives/symmetric-encryption.rst
@@ -485,7 +485,7 @@ Modes
         requires passing a 512-bit key.
 
     XTS (XEX-based tweaked-codebook mode with ciphertext stealing) is a mode
-    of operation for the AES block cipher that is used for _`disk encryption`.
+    of operation for the AES block cipher that is used for `disk encryption`_.
 
     **This mode does not require padding.**
 
--- a/docs/hazmat/primitives/symmetric-encryption.rst
+++ b/docs/hazmat/primitives/symmetric-encryption.rst
@@ -469,6 +469,29 @@ Modes
 
         a secret message!
 
+.. class:: XTS(tweak)
+
+    .. versionadded:: 2.1
+
+    .. warning::
+
+        XTS mode is meant for disk encryption and should not be used in other
+        contexts.
+
+    .. note::
+
+        AES XTS keys are double length. This means that to do AES 128
+        encryption in XTS mode you need a 256-bit key. Similarly, AES 256
+        requires passing a 512-bit key.
+
+    _`XTS` (XEX-based tweaked-codebook mode with ciphertext stealing) is a mode
+    of operation for the AES block cipher that is used for disk encryption.
+
+    **This mode does not require padding.**
+
+    :param bytes tweak: The tweak is a 16 byte value typically derived from
+        something like the disk sector number.
+
 
 Insecure modes
 --------------
@@ -744,6 +767,20 @@ Interfaces used by the symmetric cipher modes described in
         Exact requirements of the tag are described by the documentation of
         individual modes.
 
+
+.. class:: ModeWithTweak
+
+    .. versionadded:: 2.1
+
+    A cipher mode with a tweak.
+
+    .. attribute:: tweak
+
+        :type: bytes
+
+        Exact requirements of the tweak are described by the documentation of
+        individual modes.
+
 Exceptions
 ~~~~~~~~~~
 
@@ -766,3 +803,4 @@ Exceptions
 .. _`significant patterns in the output`: https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Electronic_Codebook_.28ECB.29
 .. _`International Data Encryption Algorithm`: https://en.wikipedia.org/wiki/International_Data_Encryption_Algorithm
 .. _`OpenPGP`: http://openpgp.org
+.. _`XTS`: https://en.wikipedia.org/wiki/Disk_encryption_theory#XTS
--- a/docs/hazmat/primitives/symmetric-encryption.rst
+++ b/docs/hazmat/primitives/symmetric-encryption.rst
@@ -476,12 +476,13 @@ Modes
     .. warning::
 
         XTS mode is meant for disk encryption and should not be used in other
-        contexts.
+        contexts. ``cryptography`` only supports XTS mode with
+        :class:`~cryptography.hazmat.primitives.ciphers.algorithms.AES`.
 
     .. note::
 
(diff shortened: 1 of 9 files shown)


alex:
needs a rebase (haven't reviewed)

src/cryptography/hazmat/primitives/ciphers/algorithms.py
```diff
@@ -24,7 +24,8 @@ def _verify_key_size(algorithm, key):
 class AES(object):
     name = "AES"
     block_size = 128
-    key_sizes = frozenset([128, 192, 256])
+    # 512 added to support AES-256-XTS, which uses 512-bit keys
+    key_sizes = frozenset([128, 192, 256, 512])
```
randombit: You'd also need 384 for AES-192-XTS, right?

reaperhulk: OpenSSL doesn't support AES-192-XTS (and NIST also doesn't publish test vectors), so we won't be supporting that key size.

I should probably add a bit to the XTS mode docs about only being 128/256 (with 256/512 keys)

docs/hazmat/primitives/symmetric-encryption.rst
```diff
@@ -469,6 +469,29 @@ Modes
...
+    .. versionadded:: 2.1
+
+    .. warning::
+
+        XTS mode is meant for disk encryption and should not be used in other
+        contexts.
+
+    .. note::
+
+        AES XTS keys are double length. This means that to do AES 128
```
alex: Nit: `AES-128`.

docs/hazmat/primitives/symmetric-encryption.rst
```diff
@@ -469,6 +469,29 @@ Modes
...
+        encryption in XTS mode you need a 256-bit key. Similarly, AES 256
+        requires passing a 512-bit key. AES 192 is not supported in XTS mode.
+
+    XTS (XEX-based tweaked-codebook mode with ciphertext stealing) is a mode
+    of operation for the AES block cipher that is used for `disk encryption`_.
+
+    **This mode does not require padding.**
+
+    :param bytes tweak: The tweak is a 16 byte value typically derived from
+        something like the disk sector number.
```
alex: More precise description please.

src/cryptography/hazmat/primitives/ciphers/algorithms.py
alex: I'm not wild about this. Once upon a time we thought about putting the key bytes in `XTS`. What factors led you to move away from this?

Do other crypto libraries do this the same way?

src/cryptography/hazmat/primitives/ciphers/modes.py
```diff
@@ -54,13 +63,25 @@ def tag(self):
         """
 
 
+def _check_aes_key_length(self, algorithm):
+    if algorithm.key_size > 256 and algorithm.name == "AES":
```
alex: Ugh, this is confusing and seems wrong. For one, it's equally true of any other block cipher right?

src/cryptography/hazmat/primitives/ciphers/algorithms.py
reaperhulk: Golang has a separate constructor for each mode, Crypto++ doesn't support XTS (or doesn't document it at least), and Botan has you get a cipher mode and then call `set_key` which needs to be the double length key.

I decided against having the "extra" key bytes be in the XTS mode constructor itself because it seems really bizarre to ask users to split the key in half and pass half to the AES object and half to the mode when it's all the key.

src/cryptography/hazmat/primitives/ciphers/modes.py
reaperhulk: Previously when `Cipher` called `validate_for_algorithm` against the mode we assumed that the key length had already been enforced by the constructor for the algorithm. Since now the length of the key is dependent on the  mode (but only for AES, which is the only algorithm where we support a mode with such behavior) to preserve previous behavior we need to do this check. (Side note: If we're okay with having it fail when calling encryptor/decryptor then we could remove this check here since it will fail with `UnsupportedAlgorithm` when it tries to instantiate an `EVP_CIPHER *`)

Other block ciphers are not affected because we didn't touch their allowed key sizes. It is true that a user attempting to do Camellia+XTS would see an apparent inconsistency with being unable to even provide a 512 bit key to the Camellia constructor, but I don't see a way to fix that without forcing consumers to (very confusingly IMO) split the key in half.

docs/hazmat/primitives/symmetric-encryption.rst
reaperhulk: Do you have language you want to see? I'm not sure how to make this more precise since the details appear to be almost entirely dependent on specific implementation. We could mention that reuse of a tweak is not catastrophic in the same way that nonce reuse is I guess?

I've added a little language around that but better wording welcome!

docs/hazmat/primitives/symmetric-encryption.rst
reaperhulk: Done

src/cryptography/hazmat/primitives/ciphers/modes.py
```diff
@@ -73,7 +94,31 @@ def __init__(self, initialization_vector):
         self._initialization_vector = initialization_vector
 
     initialization_vector = utils.read_only_property("_initialization_vector")
-    validate_for_algorithm = _check_iv_length
+    validate_for_algorithm = _check_iv_and_key_length
```
alex: Why is `CBC` the only thing that needs this?

alex: Err, nevermind, but doesn't ECB also need the AES length check?

tests/hazmat/primitives/test_aes.py
```diff
@@ -12,12 +12,44 @@
...
+        algorithms.AES(b"\x00" * 32), modes.XTS(b"\x00" * 16)
+    ),
+    skip_message="Does not support AES XTS",
+)
+@pytest.mark.requires_backend_interface(interface=CipherBackend)
+class TestAESModeXTS(object):
+    @pytest.mark.parametrize(
+        "vector",
+        # This list comprehension excludes any vector that does not have a
+        # data unit length that is divisible by 8
```
alex: Explain why.

PR #3904, opened 2017-09-12
docs for what bytes means for DNSName, URI, and RFC822Name

### LINKED ISSUE #3831: Better document what it means for a DNSName to be bytes

reaperhulk:
fixes #3831


reaperhulk:
I'd suggest we just ignore the clearly incorrect coverage here...

PR #3907, opened 2017-09-12
implement __hash__ on all GeneralName types

reaperhulk:
Needed to implement `__hash__` on AuthorityKeyIdentifier

I also noticed that some of our hash implementations are just hashing a single byte string, which means a trivial hash collision is possible if, for whatever reason, a user puts an instance of that object in a dict with the byte string itself. I've gone ahead and made the hash a tuple of `(type(self), self.value)` here -- should we do that everywhere or back it out on this PR?

### DIFF
--- a/src/cryptography/x509/general_name.py
+++ b/src/cryptography/x509/general_name.py
@@ -123,7 +123,7 @@ class RFC822Name(object):
         return not self == other
 
     def __hash__(self):
-        return hash(self.value)
+        return hash(self.bytes_value)
 
 
 def _idna_encode(value):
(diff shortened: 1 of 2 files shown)


reaperhulk:
Missing coverage is once again our 2.6 issue.

src/cryptography/x509/general_name.py
```diff
@@ -123,7 +123,7 @@ def __ne__(self, other):
         return not self == other
 
     def __hash__(self):
-        return hash(self.value)
+        return hash((type(self), self.bytes_value))
```
alex: Soooooo, in a very limited, technical sense, these are wrong.

Only rule of hashing, equal objects must have the same hash. 

If you make your own subclass of one of these, it won't.

Let's just do the simple, obviously correct thing, and use only `bytes_value` in the hash, it'll be correct always, and I'm not going to lose a lot of sleep over hash collisions from objects with the same values but different types. If you put an IP address in a DNSName, either you work for a major CA, or you're very confused about what a DNSName is.

PR #3908, opened 2017-09-12
coverage pin test part deux

reaperhulk:
We've closed this PR in the past because it probably doesn't do anything. And if it passes here it proves nothing. But if it fails then we know for absolute certain that this doesn't help and I can delete this branch forever!


reaperhulk:
Sigh, so no coverage problem, but that doesn't really prove anything.

Should we just merge this because it doesn't hurt? It also probably doesn't help, but ¯\_(ツ)_/¯

PR #3913, opened 2017-09-13
implement __hash__ on KeyUsage and ExtendedKeyUsage

reaperhulk:
refs #2578

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -726,6 +726,9 @@ class ExtendedKeyUsage(object):
     def __ne__(self, other):
         return not self == other
 
+    def __hash__(self):
+        return hash(tuple(self._usages))
+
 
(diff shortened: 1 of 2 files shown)


src/cryptography/x509/extensions.py
```diff
@@ -904,6 +907,15 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not self == other
 
+    def __hash__(self):
+        return hash((
+            self.digital_signature, self.content_commitment,
+            self.key_encipherment, self.data_encipherment,
+            self.key_agreement, self.key_cert_sign,
+            self.crl_sign, self.encipher_only,
+            self.decipher_only
```
alex: Why doesn't this need to use the private version of the attribute like `__eq__` does?

reaperhulk: It does need to. Test changed to catch that case and bug fixed.

PR #3919, opened 2017-09-14
add ChaCha20 support

### LINKED ISSUE #3264: ChaCha20 Support
Two things to answer here. Are we going to add ChaCha20 (separate from ChaCha20Poly1305)? It can be useful as a component in constructing other things so I think the answer is yes.

If yes, then how do we want to handle nonce/iv? The spec calls it a 96-bit nonce/iv with a 32-bit block counter. OpenSSL treats it as a single 128-bit value. Do we want to just treat it as an opaque byte string or separate out the two arguments and have the backends compose them as necessary? I'm leaning towards opaque byte string and documenting the way in which they're concatenated.

refs #2968

reaperhulk:
fixes #3264

### DIFF
--- a/docs/hazmat/primitives/symmetric-encryption.rst
+++ b/docs/hazmat/primitives/symmetric-encryption.rst
@@ -112,13 +112,16 @@ Algorithms
 
         In most cases users should use
         :class:`~cryptography.hazmat.primitives.ciphers.aead.ChaCha20Poly1305`
-        instead of this class.
+        instead of this class. `ChaCha20` alone does not provide integrity
+        so it must be combined with a MAC to be secure.
+        :class:`~cryptography.hazmat.primitives.ciphers.aead.ChaCha20Poly1305`
+        does this for you.
 
     ChaCha20 is a stream cipher used in several IETF protocols. It is
     standardized in :rfc:`7539`.
 
     :param bytes key: The secret key. This must be kept secret. ``256`` bits
-        in length.
+        (16 bytes) in length.
 
     :param bytes nonce: Should be unique, a :term:`nonce`. It is
         critical to never reuse a ``nonce`` with a given key.  Any reuse of a
@@ -142,7 +145,7 @@ Algorithms
         >>> from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
         >>> from cryptography.hazmat.backends import default_backend
         >>> nonce = os.urandom(16)
-        >>> algorithm = algorithms.ChaCha20(key,nonce)
+        >>> algorithm = algorithms.ChaCha20(key, nonce)
         >>> cipher = Cipher(algorithm, mode=None, backend=default_backend())
         >>> encryptor = cipher.encryptor()
         >>> ct = encryptor.update(b"a secret message")
--- a/docs/hazmat/primitives/symmetric-encryption.rst
+++ b/docs/hazmat/primitives/symmetric-encryption.rst
@@ -121,7 +121,7 @@ Algorithms
     standardized in :rfc:`7539`.
 
     :param bytes key: The secret key. This must be kept secret. ``256`` bits
-        (16 bytes) in length.
+        (32 bytes) in length.
 
     :param bytes nonce: Should be unique, a :term:`nonce`. It is
         critical to never reuse a ``nonce`` with a given key.  Any reuse of a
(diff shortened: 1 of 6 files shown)


src/cryptography/hazmat/backends/openssl/ciphers.py
```diff
@@ -59,6 +59,8 @@ def __init__(self, backend, cipher, mode, operation):
             iv_nonce = mode.initialization_vector
         elif isinstance(mode, modes.ModeWithNonce):
             iv_nonce = mode.nonce
+        elif isinstance(cipher, ciphers.algorithms.ChaCha20):
+            iv_nonce = cipher.nonce
```
alex: Why is this not just a ModeWithNonce?

docs/hazmat/primitives/symmetric-encryption.rst
```diff
@@ -104,6 +104,52 @@ Algorithms
...
 
+.. class:: ChaCha20(key)
+
+    .. versionadded:: 2.1
+
+    .. note::
+
+        In most cases users should use
+        :class:`~cryptography.hazmat.primitives.ciphers.aead.ChaCha20Poly1305`
+        instead of this class.
```
alex: Explain why

docs/hazmat/primitives/symmetric-encryption.rst
```diff
@@ -104,6 +104,52 @@ Algorithms
...
+
+    ChaCha20 is a stream cipher used in several IETF protocols. It is
+    standardized in :rfc:`7539`.
+
+    :param bytes key: The secret key. This must be kept secret. ``256`` bits
```
alex: byte length in parens please

docs/hazmat/primitives/symmetric-encryption.rst
```diff
@@ -104,6 +104,52 @@ Algorithms
...
+
+        .. note::
+
+            In :rfc:`7539` the nonce is defined as a 96-bit value that is later
+            concatenated with a block counter (encoded as a 32-bit
+            little-endian). If you have a separate nonce and block counter
+            you will need to concatenate it yourself before passing it. For
+            example if you have an initial block counter of 2 and a 96-bit
+            nonce the concatenated nonce would be
+            ``struct.pack("<i", 2) + nonce``.
```
alex: Is this concat backwards?

docs/hazmat/primitives/symmetric-encryption.rst
```diff
@@ -104,6 +104,52 @@ Algorithms
...
+
+    .. doctest::
+
+        >>> from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+        >>> from cryptography.hazmat.backends import default_backend
+        >>> nonce = os.urandom(16)
+        >>> algorithm = algorithms.ChaCha20(key,nonce)
```
alex: Space after commma

src/cryptography/hazmat/backends/openssl/ciphers.py
reaperhulk: It's not a mode, although I suppose we could just register the cipher as a ModeWithNonce. Opinion?

docs/hazmat/primitives/symmetric-encryption.rst
reaperhulk: Nope, you need to put the IBC on the front (see the `test_vectors` test)

docs/hazmat/primitives/symmetric-encryption.rst
reaperhulk: Done.

src/cryptography/hazmat/backends/openssl/ciphers.py
reaperhulk: I've gone ahead and made this change, but it still needs to have a separate `elif` because it's doing `cipher.nonce` instead of `mode.nonce`.

docs/hazmat/primitives/symmetric-encryption.rst
```diff
@@ -104,6 +104,55 @@ Algorithms
...
+        instead of this class. `ChaCha20` alone does not provide integrity
+        so it must be combined with a MAC to be secure.
+        :class:`~cryptography.hazmat.primitives.ciphers.aead.ChaCha20Poly1305`
+        does this for you.
+
+    ChaCha20 is a stream cipher used in several IETF protocols. It is
+    standardized in :rfc:`7539`.
+
+    :param bytes key: The secret key. This must be kept secret. ``256`` bits
+        (16 bytes) in length.
```
alex: Quite certain this isn't right.

reaperhulk: aspirational packing.

alex:
Codecov failed at merging, overriding at my discretion.

PR #3932, opened 2017-09-20
add Freshest CRL and Delta CRL Indicator test vectors

reaperhulk:


alex:
How do I verify that the _contents_ of these extensions is correct? I've confirmed they have the extension.

Issue #2546, opened 2015-12-22
CRL Extension Parsing


reaperhulk:
Here's the complete list of extensions from [RFC 5280](http://tools.ietf.org/html/rfc5280#section-5.2)
- [x] CRL Number
- [x] Authority Key Identifier
- [x] Issuer Alternative Name  (parsing support already present, just add to CRL extension parser dict)
- [x] Issuing Distribution Point (OpenSSL 1.0.0+ only)
- [x] Authority Information Access (parsing support already present, just add to CRL extension parser dict)
- [x] Freshest CRL (a.k.a. Delta CRL Distribution Point)
- [x] Delta CRL Indicator is also an extension

The last two extensions appear to be at least partially supported by OpenSSL but documentation is hard to come by...

There's also `Next CRL Publish`, a MS specific CRL extension.

reaperhulk:
Kicking this over to twenty four, cc @irinarenteria to see if she's still interested in completing this or if we should take over her branch and get it landed.

PR #3936, opened 2017-09-20
support delta crl indicator extension

reaperhulk:
This is an extension for CRLs

refs #2546


reaperhulk:
I'll rebase this to try to avoid the codecov issue (sigh) once another PR merges.

PR #3939, opened 2017-09-25
Expose FIPS funcs for OpenSSL.

ScottSturdivant:
While the merits of FIPS can certainly be debated, it would be nice if FIPS mode could be set if the backend library supports the option.  The following MR should enable that functionality.

If FIPS will never be accepted into the Cryptography library, is there a method that would enable this via some build-time parameters?

### DIFF
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -76,6 +76,7 @@ ffi = build_ffi_for_binding(
         "engine",
         "err",
         "evp",
+        "fips",
         "hmac",
         "nid",
         "objects",
--- a/src/_cffi_src/openssl/fips.py
+++ b/src/_cffi_src/openssl/fips.py
@@ -1,0 +1,30 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+from __future__ import absolute_import, division, print_function
+
+INCLUDES = """
+#include <openssl/crypto.h>
+"""
+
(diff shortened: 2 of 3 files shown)


reaperhulk:
I am okay with merging a PR that adds the FIPS functions to the bindings. There are reasons outside of developer control that result in needing such an unfortunate thing. I don't want to expose this control beyond the bindings, but this PR is (in principle) fine with me.

That said there's an error on the jenkins side you'll need to take a look at.

ScottSturdivant:
Thank you for considering this PR!  A few comments:

1. Is this the right behavior for LibreSSL?  It's not obvious to me how the creation of a function pointer to NULL will behave in Python - will `openssl.backend._lib.FIPS_mode_set(1)` return `None`, or would that function not exist at all?  Technically if you're calling `FIPS_mode_set()` with a non-zero value, you should check that the return value matches.  This isn't really the Python way though, and I'd feel that it would be safer if the FIPS functions simply were not present for LibreSSL.  If that'd be your preference too, how would that be accomplished?  Could they be removed from FUNCTIONS and instead be declared in CUSTOMIZATIONS, then remove the `_conditional.py` mods?
2. Would you prefer that all of my commits are squashed?

reaperhulk:
The conditional binding (with `Cryptography_HAS_FIPS`) means that the `_lib` in the binding object will not have those two functions when compiled against LibreSSL.

We squash commits on merge via GitHub so there's no need to squash them in the PR (unless you really want to).

ScottSturdivant:
Great, that would match my desire.  Let me know if there's anything else I can do.

PR #3942, opened 2017-09-26
Update the list of macOS versions we test on to match reality

alex:


reaperhulk:
Death to old versions!

PR #3944, opened 2017-09-30
Declare that 2.1 is the last version to support Python 2.6

alex:
FYI: @bitprophet, @glyph, @hynek, @Lukasa, @abadger, @bmw

Our current plan is 3-fold:

1) 2.1 will be the last release of cryptography to support Python 2.6;
2) If there is a need for a security release, we will backport it to 2.1 through the end of 2017
3) We'll help the certbot team land https://github.com/certbot/certbot/issues/2208 which should remove the majority of our downloads from 2.6

This is based on the dwindling rate of 2.6 users, and an increasing maintenance burden.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ Changelog
 
 .. note:: This version is not yet released and is under active development.
 
+* Python 2.6 support is deprecated, and will be removed in the next release of
+  ``cryptography``.
 * **BACKWARDS INCOMPATIBLE:** ``Whirlpool``, ``RIPEMD160``, and
   ``UnsupportedExtension`` have been removed in accordance with our
   :doc:`/api-stability` policy.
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -9,6 +9,8 @@ Changelog
 
 .. note:: This version is not yet released and is under active development.
 
+* Python 2.6 support is deprecated, and will be removed in the next release of
```
reaperhulk: Should we bold this?

alex: Any particular words you'd like bold, or want a `**DEPRECATION**` or `**BACKWARDS INCOMPATIBLE**` or something?

reaperhulk: **IMPENDING DOOM WARNING** 😀 

**FINAL DEPRECATION** might be okay. "Deprecation alone" seems insufficient? Or **SUPPORT CHANGE**

reaperhulk:
Less than 1% of our downloads in the last month are Python 2.6 for the latest version of cryptography according to BigQuery.

Issue #3943, opened 2017-09-27
Parsing a cert containing a raw IPv6 address as DNS SAN fails


jharbott:
The scenario is that I'm trying to access a server at e.g. `https://[2001:db8::17]/` running with a certificate that contains both `DNS:2001:db8::17` and `IP:2001:db8::17` via urllib3, which in turn uses the `cryptography` module in order to parse the certificate.

With released version 2.0.3, there is an error already in `cert.extensions.get_extension_for_class(x509.SubjectAlternativeName)`, if I install the current git master, the error moves to a later stage when `ext.get_values_for_type(x509.DNSName)` is called.

After asking in `#cryptography-dev` I used the workarounf of dropping the `DNS:...` part from the certificate, but in that case the certificate is no longer accepted by the standard `httplib` module, see https://bugs.python.org/issue23239, so we really need a fix to accept that part here.

jharbott:
[cert.pem.txt](https://github.com/pyca/cryptography/files/1336544/cert.pem.txt)

jharbott:
Looks like this simple patch would work for me, avoiding calling a deprecated function or so it claims. Sadly next thing to happen is `urllib3` blowing up in another idna check, but that's a different issue.

```
diff --git a/src/cryptography/x509/extensions.py b/src/cryptography/x509/extensions.py
index eb4b927..06a99ef 100644
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -1180,7 +1180,7 @@ class GeneralNames(object):
         # just one value.
         objs = (i for i in self if isinstance(i, type))
         if type != OtherName:
-            objs = (i.value for i in objs)
+            objs = (i.bytes_value for i in objs)
         return list(objs)
 
     def __repr__(self):
```

jharbott:
Patch v2:

```
diff --git a/src/cryptography/x509/extensions.py b/src/cryptography/x509/extensions.py
index eb4b927..e4bedce 100644
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -21,7 +21,7 @@ from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey
 from cryptography.x509.certificate_transparency import (
     SignedCertificateTimestamp
 )
-from cryptography.x509.general_name import GeneralName, IPAddress, OtherName
+from cryptography.x509.general_name import GeneralName, IPAddress, OtherName, DNSName
[...]

PR #3945, opened 2017-10-01
[WIP] potential solution for get_values_for_type

reaperhulk:
Potential resolution to #3943 

This is incomplete (no docs, no tests) but I wanted to put it up so we can have a discussion about this approach vs a new method `get_values_for_type_ng`


reaperhulk:
(We're not going to do it this way)

PR #3946, opened 2017-10-04
disable smart quotes in sphinx to workaround a spellchecker issue

reaperhulk:
This makes it so we don't have to pin sphinx while we wait for sphinxcontrib-spelling to handle smart quotes.

Refs https://github.com/sphinx-contrib/spelling/issues/1

Obviates #3545

### DIFF
--- a/docs/docutils.conf
+++ b/docs/docutils.conf
@@ -1,0 +1,2 @@
+[parsers]
+smart_quotes: no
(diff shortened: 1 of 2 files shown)


reaperhulk:
This should be reverted once https://github.com/sphinx-contrib/spelling/pull/2 and https://github.com/sphinx-doc/sphinx/pull/4110 land and are in released versions of sphinx and sphinxcontrib-spelling.

PR #3950, opened 2017-10-05
Fix: format param doc string

gesslerpd:


alex:
Thank you for your contribution, however the docs are correct as is; a `Certificate` object's `public_bytes` method does not take a `format` argument: https://github.com/pyca/cryptography/blob/master/src/cryptography/x509/base.py#L171-L175

gesslerpd:
Oops, my bad, was looking in the wrong part of the docs, sorry

PR #3951, opened 2017-10-09
Backwards incompatible change to DNSName

reaperhulk:
During this release cycle we decided to officially deprecate passing U-labels to our GeneralName constructors. At first we tried changing this in a purely backwards compatible way, but get_values_for_type made that untenable. This PR modifies DNSName to take three different types. U-label strings (which raises a deprecation warning), A-label strings (the new preferred type), and bytes (which are assumed to be decodable to unicode strings). The latter, while supported, is primarily intended for use by our parser and allows us to return the actual encoded data in a certificate even if it has not been properly encoded to A-label before the certificate is created. (Of course, if the certificate contains invalid utf8 sequences this will still fail, but let's handle one catastrophic failure at a time).

Note that this change is only backwards incompatible in the case where the user is passing a string that contains an internationalized domain name U-label. In that scenario the user will pass a string containing unicode characters and receive back a string encoded to A-label when calling `value`.

refs #3943 

Once URI and RFC822Name are complete we'll also need to update the changelog.

### DIFF
--- a/src/cryptography/x509/general_name.py
+++ b/src/cryptography/x509/general_name.py
@@ -131,8 +131,8 @@ def _idna_encode(value):
     for prefix in ['*.', '.']:
         if value.startswith(prefix):
             value = value[len(prefix):]
-            return prefix.encode('ascii') + idna.encode(value)
-    return idna.encode(value)
+            return prefix + idna.encode(value).decode("ascii")
+    return idna.encode(value).decode("ascii")
 
 
 @utils.register_interface(GeneralName)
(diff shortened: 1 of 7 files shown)


src/cryptography/x509/general_name.py
```diff
@@ -131,82 +131,49 @@ def _idna_encode(value):
...
+                    "idna. Support for passing unicode strings (aka U-label) "
+                    " will be removed in a future version.",
                     utils.DeprecatedIn21,
                     stacklevel=2,
                 )
-        elif not isinstance(value, bytes):
-            raise TypeError("value must be bytes")
-
-        self._bytes_value = value
+        elif isinstance(value, six.binary_type):
```
alex: Why are we accepting `bytes` here?

tests/x509/test_x509_ext.py
```diff
@@ -1662,33 +1662,36 @@ def test_key_cert_sign_crl_sign(self, backend):
 
 class TestDNSName(object):
     def test_init(self):
-        with pytest.warns(utils.DeprecatedIn21):
-            name = x509.DNSName(u"*.\xf5\xe4\xf6\xfc.example.com")
-        assert name.bytes_value == b"*.xn--4ca7aey.example.com"
+        name = x509.DNSName(u".xn--4ca7aey.example.com")
```
alex: Why did you drop the `*` fromt his test?

src/cryptography/x509/general_name.py
reaperhulk: If we don't accept bytes then if a cert has actual unicode characters in it we will (incorrectly) encode it to an A-label. With a bytes path we can preserve the exact data from the cert (provided it is a valid UTF-8 sequence).

tests/x509/test_x509_ext.py
reaperhulk: That'd be a mistake.

src/cryptography/x509/general_name.py
alex: Hmm, is it better to accept bytes, or to have the internal creation path do `n = DNSName.__new__(); n._value = value`, hacky, but doesn't expand our public API.

reaperhulk: I'm fine with bypassing the constructor. I wasn't planning on documenting the bytes path anyway (callers should pass strings).

Only nuance then is to make sure we have a test that allows a parsed unicode generalname to properly be encoded. We may already have that test but I'll look.

alex: I can probably find you a garbage cert from a public CA if you need it.

PR #3953, opened 2017-10-11
backwards incompatible change to RFC822Name

reaperhulk:
During this release cycle we decided to officially deprecate passing U-labels to our GeneralName constructors. At first we tried changing this in a purely backwards compatible way but get_values_for_type made that untenable. This PR modifies RFC822Name to accept two types: U-label strings (which raises a deprecation warning) and A-label strings (the new preferred type). There is also a constructor for RFC822Name that bypasses validation so we can parse garbage out of certificates
(and round trip it if necessary).

refs #3951

### DIFF
--- a/tests/x509/test_x509.py
+++ b/tests/x509/test_x509.py
@@ -795,10 +795,6 @@ class TestRSACertificate(object):
             u'xn--biztosts-fza2j.hu', u'*.xn--biztosts-fza2j.hu'
         ]
 
-    def test_non_ascii_rfc822name(self, backend):
-        # TODO
-        pass
-
     def test_all_subject_name_types(self, backend):
         cert = _load_cert(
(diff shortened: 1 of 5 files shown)


tests/x509/test_x509.py
```diff
@@ -795,6 +795,10 @@ def test_non_ascii_dns_name(self, backend):
             u'xn--biztosts-fza2j.hu', u'*.xn--biztosts-fza2j.hu'
         ]
 
+    def test_non_ascii_rfc822name(self, backend):
+        # TODO
+        pass
```
alex: ...

reaperhulk: Oh yeah 😄  oops, that test turns out to not be needed but I forgot to remove it.

PR #3955, opened 2017-10-11
Update docs and changelog for URI, RFC822Name, and DNSName

reaperhulk:
As of 2.1 we want users to pass A-label strings, but we still accept U-label strings and auto-encode on their behalf (with a deprecation warning). However, we do not return U-label in the value attribute so we are making a backwards incompatible change for users that utilize internationalized domain names.

Document all of that!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,27 +14,20 @@ Changelog
 * **BACKWARDS INCOMPATIBLE:** ``Whirlpool``, ``RIPEMD160``, and
   ``UnsupportedExtension`` have been removed in accordance with our
   :doc:`/api-stability` policy.
+* **BACKWARDS INCOMPATIBLE:** :attr:`~cryptography.x509.DNSName.value`,
+  :attr:`~cryptography.x509.RFC822Name.value`, and
+  :attr:`~cryptography.x509.UniformResourceIdentifier.value` will now return
+  an :term:`A-label` string if you provided a :term:`U-label`
+  internationalized domain to the constructor. See below for additional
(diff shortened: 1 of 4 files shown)


CHANGELOG.rst
```diff
@@ -14,27 +14,20 @@ Changelog
 * **BACKWARDS INCOMPATIBLE:** ``Whirlpool``, ``RIPEMD160``, and
   ``UnsupportedExtension`` have been removed in accordance with our
   :doc:`/api-stability` policy.
+* **BACKWARDS INCOMPATIBLE:** :attr:`~cryptography.x509.DNSName.value`,
+  :attr:`~cryptography.x509.RFC822Name.value`, and
+  :attr:`~cryptography.x509.UniformResourceIdentifier.value` will now return
+  an :term:`A-label` string if you provided a :term:`U-label`
+  internationalized domain to the constructor. See below for additional
+  deprecations related to this change.
```
alex: More importantly, they return a-labels for certs with a-labels, instead of returning u-labels

docs/glossary.rst
```diff
@@ -85,5 +85,10 @@ Glossary
         prefix ``xn--``. To create an A-label from a unicode domain string use
         a library like `idna`_.
 
+    U-label
+        The presentational unicode form of an internationalized domain
+        name. U-labels use unicode characters outside the ASCII range and
+        are encoded into A-labels in certificates.
```
alex: How about "are encoded as A-labels when stored in certificates"?

alex:
Merge conflict.

PR #3962, opened 2017-10-11
expunge python 2.6

reaperhulk:


src/cryptography/utils.py
```diff
@@ -101,12 +101,8 @@ def verify_interface(iface, klass):
...
 
 
-if sys.version_info >= (2, 7):
-    def bit_length(x):
-        return x.bit_length()
-else:
-    def bit_length(x):
-        return len(bin(x)) - (2 + (x <= 0))
+def bit_length(x):
+    return x.bit_length()
```
alex: We should just inline this into all the callers now; in this PR or a seperate one?

.travis.yml
```diff
@@ -19,10 +19,6 @@ branches:
 matrix:
     include:
         # these are just to make travis's UI a bit prettier
-        - python: 2.6
-          env: TOXENV=py26
-        - python: 2.7
```
reaperhulk: whoops what

src/cryptography/utils.py
reaperhulk: This is effectively public API (I've seen other projects using this). If we want to get rid of it I think we need a deprecation cycle.

alex: We should probably at least stop using it ourselves; I'm ok with keeping it around.

PR #3969, opened 2017-10-12
auto-cancel jobs in jenkins when a new commit is pushed to a PR/branch

reaperhulk:
does not apply to master


reaperhulk:
Apparently can't do this without whitelisting some method calls I don't want to whitelist, so closing for now.

PR #3970, opened 2017-10-12
random hack to maybe support older pips

alex:


lepture:
This environment marker doesn't working for PyPy, even with the latest setuptools.

https://travis-ci.org/idan/oauthlib/jobs/289548171

alex:
You need to run `pip install -U setuptools` before `pip install coveralls`.

Because they are both in the same installation command, `cryptography` is being installed before `setuptools` is upgraded (`coveralls` depends on `urllib3[secure]` which depends on `cryptography`).

lepture:
@alex thanks

PR #3973, opened 2017-10-12
random hack to maybe support older pips (#3970)

alex:


reaperhulk:
TIL “Pull requests that have a failing status can’t be merged on a phone.”. But you can just load the desktop site soooo....

PR #3989, opened 2017-10-19
Run ubuntu rolling tests on py3

alex:
ubuntu artuful moves from 3.5 to 3.6


alex:
Dunno what's going on here, closing and going to retry.

PR #3990, opened 2017-10-19
begin deprecation of backend (x509 module first)

reaperhulk:
With this PR the following functions/methods no longer require passing a backend:

load_pem_x509_crl
load_der_x509_crl
load_pem_x509_certificate
load_der_x509_certificate
load_pem_x509_csr
load_der_x509_csr
CertificateBuilder.sign
CSRBuilder.sign
CRLBuilder.sign

This does not raise a deprecation warning at this time, but eventually we'll want to raise deprecation warning if backend is passed.

Tests here are a bit stupid, suggestions welcome.


alex:
Needs docs

docs/x509/reference.rst
```diff
@@ -128,38 +128,40 @@ Loading Certificates
         >>> cert.serial_number
         2
 
-.. function:: load_der_x509_certificate(data, backend)
-
-    .. versionadded:: 0.7
+.. function:: load_der_x509_certificate(data, backend=Nonej)
```
jlaine: Nonej => None

intgr:
I'm not surprised that this is happening, but just curious, is there any discussion you can point to about the deprecation of backends?

alex:
Unfortunately these discussions were overwhelmingly in person, and also a bit on IRC, so there's not a clear place to point you.

PR #3992, opened 2017-10-19
Correctly record the minimum enum34 version

alex:


alex:
Ok, turns out the actual problem is that someone `pip install enum`. Closing this.

PR #4000, opened 2017-10-26
clearly error out on older setuptools

alex:


setup.py
```diff
@@ -14,11 +14,21 @@
...
 import pkg_resources
 
+import setuptools
 from setuptools import find_packages, setup
 from setuptools.command.install import install
 from setuptools.command.test import test
 
 
+if (
+    pkg_resources.SetuptoolsVersion(setuptools.__version__) <
```
dstufft: You probably want ``pkg_resources.parse_version``, it'll use ``SetuptoolsVersion`` on ones that have it, or it'll use the old mechanism on older ones.

PR #4014, opened 2017-11-08
Use a different warning class so users get warnings

alex:
I expect this will break a few tests, will require a cleanup commit

### DIFF
--- a/src/cryptography/utils.py
+++ b/src/cryptography/utils.py
@@ -12,7 +12,7 @@ import warnings
 
 
 # We use a UserWarning subclass, instead of DeprecationWarning, because CPython
-# decided deprecation warnings should be invisbly be default.
+# decided deprecation warnings should be invisble be default.
 class CryptographyDeprecationWarning(UserWarning):
     pass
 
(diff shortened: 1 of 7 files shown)


njsmith:
`FutureWarning` is kinda similar to `DeprecationWarning` but is shown by default – it might be more appropriate than the vague `UserWarning`.

It's probably a good idea to make empty warning subclasses so that people can easily filter on them if they want:
```python
class CryptographyDeprecationWarning(FutureWarning):
    pass
```

A policy you might want to consider is using a `DeprecationWarning` in X.Y+1, and then switching to something visible in X.Y+2, so diligent projects can catch it early and fix it before you start yelling at people. Or you could go straight to yelling at people, that works too.

src/cryptography/utils.py
```diff
@@ -11,11 +11,16 @@
 import warnings
 
 
+# We use a UserWarning subclass, instead of DeprecationWarning, because CPython
+# decided deprecation warnings should be invisbly be default.
```
reaperhulk: typo

PR #4016, opened 2017-11-09
nit in LICENSE, itself doesn't make sense here

alex:


LICENSE
```diff
@@ -3,5 +3,5 @@ found in LICENSE.APACHE or LICENSE.BSD. Contributions to cryptography are made
 under the terms of *both* these licenses.
 
 The code used in the OpenSSL locking callback and OS random engine is derived
-from the same in CPython itself, and is licensed under the terms of the PSF
-License Agreement.
+from the same in CPython, and is licensed under the terms of the PSF License
```
reaperhulk: How about we just say "derived from CPython". "the same in CPython" still doesn't really make much sense.

PR #4021, opened 2017-11-15
added brainpool ec-curves

joerichter-stash:
Adding brainpool curve names as specified in RFC5639. They are supported in pyopenssl and usable with this patch.

### DIFF
--- a/src/cryptography/hazmat/primitives/asymmetric/ec.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/ec.py
@@ -228,6 +228,54 @@ class SECP192R1(object):
     key_size = 192
 
 
+@utils.register_interface(EllipticCurve)
+class BRAINPOOLP256R1(object):
+    name = "brainpoolP256r1"
+    key_size = 256
+
+
+@utils.register_interface(EllipticCurve)
+class BRAINPOOLP256T1(object):
+    name = "brainpoolP256t1"
+    key_size = 256
+
+
+@utils.register_interface(EllipticCurve)
+class BRAINPOOLP320R1(object):
+    name = "brainpoolP320r1"
+    key_size = 320
+
+
+@utils.register_interface(EllipticCurve)
+class BRAINPOOLP320T1(object):
+    name = "brainpoolP320t1"
+    key_size = 320
+
+
+@utils.register_interface(EllipticCurve)
+class BRAINPOOLP384R1(object):
+    name = "brainpoolP384r1"
+    key_size = 384
+
+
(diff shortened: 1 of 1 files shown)


reaperhulk:
Do you have a specific use case for the brainpool curves?

joerichter-stash:
Yes, i'm working on a project for the german smart-meter infrastructure where those curves are mandatory for certificates by specification. Using pyopenssl OpenSSL.crypto.get_elliptic_curves() the returned list states that the brainpool curves are supported, but further down in the cryptography module they where not supported. So for my implementation it was the straight forward way to add them.

reaperhulk:
So the spec mandates x509 certificates signed using ECDSA over brainpool curves? Would it be possible to provide a sample cert and a link to the spec? Thanks for the help so far, it's very useful to have context when we're deciding whether to expose additional curve parameters like this!

joerichter-stash:
https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/TechnischeRichtlinien/TR03109/TR-03109-3_Kryptographische_Vorgaben.pdf?__blob=publicationFile&v=1

Why would you drop ECC below 256bit? According to https://de.wikipedia.org/wiki/Elliptic_Curve_Cryptography#Effizienz_und_Sicherheit a 160bit ECDH key has the equivalent security as a 1024bit RSA/DH key. And additionally the educated user of this library would not longer have the option which cipher he wants to use.

alex:
1024-bit RSA/DH is totally unacceptable at this point in time, so that's a poor argument.

One of the goals of this project is to make it harder for our users to make cryptographic mistakes and to push the ecosystem forward.

joerichter-stash:
I'm using keys >=256bit ECC keys. I just wanted to add all curves supported by pyopenssl. If this is an requirement, i can drop <256bit curves from my pull request.

schlenk:
Probably a better link to the crypto parts of the spec (still only german), which is referenced by the short paper linked by @joerichter-stash 
https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/TechnischeRichtlinien/TR03116/BSI-TR-03116-3.pdf?__blob=publicationFile&v=6

No wonder BSI wants the Brainpool curves, as they were involved in specifying them...

joerichter-stash:
The original specification of the curves can be found here: https://tools.ietf.org/html/rfc5639

alex:
Adding a few specific ones, rather than a ton of them has a strong 👍 from me.

reaperhulk:
closing in favor of #4129

PR #4035, opened 2017-12-01
Fix ASN1 string type encoding for several Name OIDs

### LINKED ISSUE #4034: Problematic type for emailAddress in CSR
Hello.
We generate CSR with cryptography, and after update from 2.0.3 to 2.1.3 we catch exception from openssl:
```
[actics@e7470 trash]$ openssl ca -days 365 -in request.csr -out certificate.cer -config openssl.cnf -batch
Using configuration from openssl.cnf
Check that the request matches the signature
Signature ok
The Subject's Distinguished Name is as follows
commonName            :ASN.1 12:'example.com'

emailAddress type needs to be of type IA5STRING
```
I found commit with problem changes: 72c92f5ed1
In my case, I was able to fix it with add private argument to x509.NameAttribute, but i don't like to use private objects :) 
```x509.NameAttribute(x509.NameOID.EMAIL_ADDRESS, emailAddress, _type=_ASN1Type.IA5String)```
Am I doing something wrong or is it a bug?
Thank you

### PR DESCRIPTION
**Pull request by reaperhulk**
When we changed over to the new type encoding system we didn't verify that the new code exactly matched the ASN1 string types that OpenSSL was previously choosing. This caused serialNumber, dnQualifier, emailAddress, and domainComponent to change from their proper encodings to UTF8String as of version 2.1.

Now we check to see if there's a sentinel value (indicating no custom type has been passed) and then check if the OID has a different default than UTF8. If it does, we set it.

This PR also adds tests for the ASN1 string type of ever supported NameOID.

Fixes #4034

### DIFF
--- a/src/cryptography/x509/name.py
+++ b/src/cryptography/x509/name.py
@@ -68,10 +68,7 @@ class NameAttribute(object):
         # to look up whether the OID has a non-UTF8 type. If it does, set it
         # to that. Otherwise, UTF8!
         if _type == _SENTINEL:
-            try:
-                _type = _NAMEOID_DEFAULT_TYPE[oid]
-            except KeyError:
-                _type = _ASN1Type.UTF8String
+            _type = _NAMEOID_DEFAULT_TYPE.get(oid, _ASN1Type.UTF8String)
 
         if not isinstance(_type, _ASN1Type):
             raise TypeError("_type must be from the _ASN1Type enum")
--- a/src/cryptography/x509/name.py
+++ b/src/cryptography/x509/name.py
@@ -27,6 +27,14 @@ class _ASN1Type(Enum):
 
 _ASN1_TYPE_TO_ENUM = dict((i.value, i) for i in _ASN1Type)
 _SENTINEL = object()
+_NAMEOID_DEFAULT_TYPE = {
+    NameOID.COUNTRY_NAME: _ASN1Type.PrintableString,
+    NameOID.JURISDICTION_COUNTRY_NAME: _ASN1Type.PrintableString,
+    NameOID.SERIAL_NUMBER: _ASN1Type.PrintableString,
+    NameOID.DN_QUALIFIER: _ASN1Type.PrintableString,
+    NameOID.EMAIL_ADDRESS: _ASN1Type.IA5String,
+    NameOID.DOMAIN_COMPONENT: _ASN1Type.IA5String,
+}
 
 
 class NameAttribute(object):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/cryptography/x509/name.py [hunk 1]
```diff
@@ -50,17 +58,20 @@ def __init__(self, oid, value, _type=_SENTINEL):
...
+        # multiple RFCs including 2459, 3280, and 5280. In general UTF8String
+        # is preferred (2459), but 3280 and 5280 specify several OIDs with
+        # alternate types. This means when we see the sentinel value we need
+        # to look up whether the OID has a non-UTF8 type. If it does, set it
+        # to that. Otherwise, UTF8!
         if _type == _SENTINEL:
-            _type = _ASN1Type.UTF8String
+            try:
+                _type = _NAMEOID_DEFAULT_TYPE[oid]
+            except KeyError:
```
alex: Just use `.get`

tests/x509/test_x509.py
```diff
@@ -3668,6 +3762,47 @@ def test_unsupported_subject_public_key_info(self, backend):
...
+            NameOID.JURISDICTION_STATE_OR_PROVINCE_NAME,
+            _ASN1Type.UTF8String
+        ),
+        (NameOID.BUSINESS_CATEGORY, _ASN1Type.UTF8String),
+        (NameOID.POSTAL_ADDRESS, _ASN1Type.UTF8String),
+        (NameOID.POSTAL_CODE, _ASN1Type.UTF8String),
+    ]
+
+    def test_default_types(self):
+        for pair in TestNameAttribute.EXPECTED_TYPES:
```
alex: use unpacking

tests/x509/test_x509.py
```diff
@@ -1650,6 +1650,59 @@ def test_encode_nonstandard_aia(self, backend):
...
+        ).serial_number(
+            777
+        ).not_valid_before(
+            datetime.datetime(1999, 1, 1)
+        ).not_valid_after(
+            datetime.datetime(2020, 1, 1)
+        ).sign(private_key, hashes.SHA256(), backend)
+
+        for dn in (cert.subject, cert.issuer):
+            for pair in TestNameAttribute.EXPECTED_TYPES:
```
alex: and here

tests/x509/test_x509.py
```diff
@@ -2747,6 +2800,47 @@ def test_build_ca_request_with_unicode(self, backend):
...
+                ),
+                x509.NameAttribute(
+                    NameOID.JURISDICTION_STATE_OR_PROVINCE_NAME, u"value"
+                ),
+                x509.NameAttribute(NameOID.BUSINESS_CATEGORY, u"value"),
+                x509.NameAttribute(NameOID.POSTAL_ADDRESS, u"value"),
+                x509.NameAttribute(NameOID.POSTAL_CODE, u"value"),
+            ])
+        ).sign(private_key, hashes.SHA256(), backend)
+        for pair in TestNameAttribute.EXPECTED_TYPES:
```
alex: and here

PR #4037, opened 2017-12-01
The HKDF limit is actually 255 * digest_length_in_bytes

### LINKED ISSUE #4032: Bug in HKDF?
I think the computation of [`max_length`](https://github.com/pyca/cryptography/blob/66460d8f62b3f27a009bb61be6ce7675c8451b6e/src/cryptography/hazmat/primitives/kdf/hkdf.py#L70) in `src/cryptography/hazmat/primitives/kdf/hkdf.py` is wrong.

[RFC5869](https://tools.ietf.org/html/rfc5869) states on page 3 that the input `L` of the HKDF-Expand function describes the "length of output keying material in octets (<= 255*HashLen)".
An octet consists of 8 bit. 

Currently, `max_length` is computed as:

```
max_length = 255 * (algorithm.digest_size // 8)
```

The problem is, that `algorithm.digest_size` returns the size of the digest in bytes. (There are 8 bits per byte). Therefore, the division by 8 is wrong, and thus, `max_length` is unnecessarily small.

(same applies for the computation of `salt` as well ([line 33](https://github.com/pyca/cryptography/blob/66460d8f62b3f27a009bb61be6ce7675c8451b6e/src/cryptography/hazmat/primitives/kdf/hkdf.py#L33)), in the case where `salt is None`)

reaperhulk:
Previously we had a bug where we divided digest_size by 8...but HashAlgorithm.digest_size is already in bytes.

RFC 5869:
```
      L        length of output keying material in octets
               (<= 255*HashLen)
```

fixes #4032

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Changelog
 .. note:: This version is not yet released and is under active development.
 
 * **BACKWARDS INCOMPATIBLE:** Support for Python 2.6 has been dropped.
+* Resolved a bug in ``HKDF`` that incorrectly constrained output size.
 * Added token rotation support to :doc:`Fernet </fernet>` with
   :meth:`~cryptography.fernet.MultiFernet.rotate`.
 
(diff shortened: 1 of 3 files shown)


alex:
I'd a like a test that it returns the ocrrect result for values between these lengths.

alex:
I am not :-) Generate one against a foreign implementation I guess.

PR #4068, opened 2018-01-04
DH interfaces existed in 0.9 but we didn't implement until 1.7

### LINKED ISSUE #4067: `dh.generate_parameters` wasn't supported before 1.7
As far as I can tell, the docs claiming support for `dh.generate_parameters` (and other things) starting from version 0.9 are wrong. The git blame suggests that these were first included in tag 1.7: https://github.com/pyca/cryptography/blame/master/docs/hazmat/primitives/asymmetric/dh.rst

I think they were first added by https://github.com/pyca/cryptography/commit/495f21a9c61befd4a61054dd1b6080164a7400e9#diff-6be0e8a855120dce88594543c96c1a05 which also touches the changelog of 1.7.

reaperhulk:
This is misleading to our users since they understandably assume that "added in 0.9" means it works.

Fixes #4067

### DIFF
--- a/docs/hazmat/primitives/asymmetric/dh.rst
+++ b/docs/hazmat/primitives/asymmetric/dh.rst
@@ -68,7 +68,7 @@ Group parameters
 
 .. function:: generate_parameters(generator, key_size, backend)
 
-    .. versionadded:: 0.9
+    .. versionadded:: 1.7
 
     Generate a new DH parameter group for use with ``backend``.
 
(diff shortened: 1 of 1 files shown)


thomwiggers:
> I kind of wonder if we should be deleting versionadded markers that are sufficiently old, at some point they're just noise and if you need to know you're better off reading the git history.

I unfortunately given the state of this package in e.g. Debian Oldstable, it's still very valuable to know if something was added pre or post even a very old version like 0.6. The `versionadded` for versions prior to that _probably_ should be ok to remove, but the 0.9 ones are still very useful imo.

PR #4070, opened 2018-01-05
just a quick confirmation that it really is an x25519 evp key

reaperhulk:


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1914,6 +1914,7 @@ def x25519_load_private_bytes(self, data):
         evp_pkey = backend._lib.d2i_PrivateKey_bio(bio.bio, self._ffi.NULL)
         self.openssl_assert(evp_pkey != self._ffi.NULL)
         evp_pkey = self._ffi.gc(evp_pkey, self._lib.EVP_PKEY_free)
+        assert self._lib.EVP_PKEY_id(evp_pkey) == self._lib.EVP_PKEY_X25519
```
alex: `openssl_assert`?

reaperhulk: Yeah I can switch it. I went with assert since this is sort-of-invariant (any key loaded this way is prefixed with the PKCS8 X25519 structure), but it's plausible that a parsing bug could somehow ignore that with a different inner value so okay.

Issue #4065, opened 2017-12-27
Why not use `EC_POINT_clear_free` and `BN_clear_free`?


tuxxy:
Hey, folks!

Just a question -- why aren't `EC_POINT_clear_free` and `BN_clear_free` used? Instead, `EC_POINT_free` and `BN_free` are used.

It is my understanding that the formerly named functions clear the memory, then free it. This is safer compared to just freeing the memory.

Are there issues associated with using this? I already know about the nuances with how memory is handled on the system (ie: swap files, etc).

Much less, why aren't these simply exposed in the `backend`? `EC_POINT_clear_free` is exposed, but `BN_clear_free` is not.

alex:
This would be a good addition.

reaperhulk:
For anyone who wants to tackle this, a good way would be ~3 PRs.  Exposing `BN_clear_free`, using `BN_clear_free` in appropriate places, and using `EC_POINT_clear_free` in appropriate places.

tuxxy:
With this said, maybe it would also be a good idea to expose and use `BN_secure_new` where possible, too?

@alex @reaperhulk Any thoughts on this?

tuxxy:
@reaperhulk I'm bringing my discussion of `EC_POINT_clear_free` over here.

It seems that `EC_POINT_free` is only used in one place, as far as I can tell, and it's on a public value.

Using similar logic to the work on implementing `BN_clear_free`, should we just skip using `EC_POINT_clear_free`, then? See L1394 on `hazmat.backends.openssl.backend.py`

reaperhulk:
Yeah let's skip it then.

tuxxy:
Alright. I think #4072 will close out this issue, then. 🐧

PR #4072, opened 2018-01-05
Use `BN_clear_free` in places where `BN_free` is being used

tuxxy:
### What this does:
1. Implements `BN_clear_free` in places where `BN_free` was being used on private values. (See #4065)
2. Keeps the use of `BN_free` on public values.

### Why?
1. `BN_clear_free` is a security improvement. By replacing `BN_free` with this, whenever the garbage collector is called it will: 1) clear the memory with zeros, then 2) free the memory.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -353,7 +353,7 @@ class Backend(object):
         rsa_cdata = self._ffi.gc(rsa_cdata, self._lib.RSA_free)
 
         bn = self._int_to_bn(public_exponent)
-        bn = self._ffi.gc(bn, self._lib.BN_free)
+        bn = self._ffi.gc(bn, self._lib.BN_clear_free)
 
         res = self._lib.RSA_generate_key_ex(
             rsa_cdata, key_size, bn, self._ffi.NULL
@@ -1359,7 +1359,7 @@ class Backend(object):
             ec_cdata, public.x, public.y)
 
         private_value = self._ffi.gc(
-            self._int_to_bn(numbers.private_value), self._lib.BN_free
+            self._int_to_bn(numbers.private_value), self._lib.BN_clear_free
         )
         res = self._lib.EC_KEY_set_private_key(ec_cdata, private_value)
         self.openssl_assert(res == 1)
@@ -1394,7 +1394,7 @@ class Backend(object):
         point = self._ffi.gc(point, self._lib.EC_POINT_free)
 
         value = self._int_to_bn(private_value)
-        value = self._ffi.gc(value, self._lib.BN_free)
+        value = self._ffi.gc(value, self._lib.BN_clear_free)
 
         with self._tmp_bn_ctx() as bn_ctx:
             res = self._lib.EC_POINT_mul(group, point, value, self._ffi.NULL,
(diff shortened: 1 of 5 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -353,7 +353,7 @@ def generate_rsa_private_key(self, public_exponent, key_size):
         rsa_cdata = self._ffi.gc(rsa_cdata, self._lib.RSA_free)
 
         bn = self._int_to_bn(public_exponent)
-        bn = self._ffi.gc(bn, self._lib.BN_free)
+        bn = self._ffi.gc(bn, self._lib.BN_clear_free)
```
alex: Not needed here, this is a public value

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1499,8 +1499,8 @@ def _ec_key_set_public_key_affine_coordinates(self, ctx, x, y):
                 "Invalid EC key. Both x and y must be non-negative."
             )
 
-        x = self._ffi.gc(self._int_to_bn(x), self._lib.BN_free)
-        y = self._ffi.gc(self._int_to_bn(y), self._lib.BN_free)
+        x = self._ffi.gc(self._int_to_bn(x), self._lib.BN_clear_free)
```
alex: These are also public

src/cryptography/hazmat/backends/openssl/decode_asn1.py
```diff
@@ -695,7 +695,7 @@ def _asn1_to_der(backend, asn1_type):
 def _asn1_integer_to_int(backend, asn1_int):
     bn = backend._lib.ASN1_INTEGER_to_BN(asn1_int, backend._ffi.NULL)
     backend.openssl_assert(bn != backend._ffi.NULL)
-    bn = backend._ffi.gc(bn, backend._lib.BN_free)
+    bn = backend._ffi.gc(bn, backend._lib.BN_clear_free)
```
alex: I'm pretty sure these are always public, since they're from an X.509 cert -- does that sound right @reaperhulk ?

src/cryptography/hazmat/backends/openssl/encode_asn1.py
```diff
@@ -29,7 +29,7 @@ def _encode_asn1_int(backend, x):
     # machine's native integer limits (note: `int_to_bn` doesn't automatically
     # GC).
     i = backend._int_to_bn(x)
-    i = backend._ffi.gc(i, backend._lib.BN_free)
+    i = backend._ffi.gc(i, backend._lib.BN_clear_free)
```
alex: Same.

src/cryptography/hazmat/backends/openssl/decode_asn1.py
reaperhulk: Every point where we utilize this is within X509 and is public, yes. I guess that doesn't preclude us using it for non-public things in the future, although I can't currently think of any scenarios where we'd want to do that.

tuxxy: Great, I'll take care of this then.

tuxxy:
Ah, it looks like I mistakenly threw in my branch for adding `EC_POINT_clear_free` here.

I can remove this and push it to another PR, or we can keep it in. @alex @reaperhulk Your call..

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1391,10 +1391,10 @@ def derive_elliptic_curve_private_key(self, private_value, curve):
 
         point = self._lib.EC_POINT_new(group)
         self.openssl_assert(point != self._ffi.NULL)
-        point = self._ffi.gc(point, self._lib.EC_POINT_free)
+        point = self._ffi.gc(point, self._lib.EC_POINT_clear_free)
```
tuxxy: See here.

tuxxy: Mind you that this is also a public value. Other than that, not sure if I see a use for `EC_POINT_clear_free` if I'm going off of the logic from the implementation of `BN_clear_free`.

alex:
If you rebase this branch on master it should clean it up.

PR #4073, opened 2018-01-06
Drop conda workaround from installation docs

alex:
I don't think it's relevant anymore


reaperhulk:
I don't believe you codecov.

PR #4074, opened 2018-01-06
add 1200 byte HKDF test vector and a generator/verifier for it

reaperhulk:
refs #4037 

(Note that to run the Python script you'll have to edit the current master HDKF implementation to remove its length limit)


docs/development/custom-vectors/hkdf/verify_hkdf.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+	}
+	return verifier(l, unhexlify(ikm), unhexlify(okm))
+}
+
+func main() {
+	if validateVectors("vectors/cryptography_vectors/KDF/hkdf-generated.txt") {
+		fmt.Println("HKDF OK.")
+	} else {
+		fmt.Println("HKDF failed.")
```
alex: This really should exit with a non-0 status code.

PR #4081, opened 2018-01-12
Remove some unused headers.

davidben:
cms and conf appear to be unused in both cryptography.io and PyOpenSSL.

@reaperhulk suggested that patches to unwind the mess of unused bindings in `_cffi_src` would be welcome. To start with, here are a couple of files that I believe can be dropped wholesale.

As before, I haven't actually gotten things to build yet, but presumably the CI will do strictly more than me anyway.


src/_cffi_src/openssl/conf.py
```diff
@@ -1,21 +0,0 @@
-# This file is dual licensed under the terms of the Apache License, Version
```
alex: These were explicitly added because a user asked for them; I don't think we can just remove them.

reaperhulk:
closing in favor of #4131

PR #4099, opened 2018-02-05
Add bindings for SRTP support

jlaine:
I spotted a number of new bindings being added to support DTLS, which is great!

One useful application is negotiating SRTP keys using DTLS, for instance for use with WebRTC:

https://tools.ietf.org/html/rfc5764

To do so, you need bindings for SSL_set_tlsext_use_srtp and/or SSL_CTX_set_tlsext_use_srtp, which this pull request provides.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -405,6 +405,9 @@ long SSL_set_tlsext_status_type(SSL *, long);
 long SSL_CTX_set_tlsext_status_cb(SSL_CTX *, int(*)(SSL *, void *));
 long SSL_CTX_set_tlsext_status_arg(SSL_CTX *, void *);
 
+int SSL_CTX_set_tlsext_use_srtp(SSL_CTX *, const char *);
+int SSL_set_tlsext_use_srtp(SSL *, const char *);
+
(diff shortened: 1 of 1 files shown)


reaperhulk:
(These are technically in `srtp.h` but `ssl.h` includes it so I'm okay with them being in this file).

jlaine:
Thanks a lot!

Once this gets released and DTLSv1_method trickles into pyopenssl's SSL.Context, I'd be happy to follow up with a patch for pyopenssl to add SSL.Context.set_tlsext_use_srtp(self, profiles).

jlaine:
FYI you can find an example of asyncio-based code which performs DTLS handshake and SRTP keying here:

https://gist.github.com/jlaine/4c39326b251a43423b477d14ed67b21b

Thanks to the bindings you added I'll be able to use cryptography's bindings, and eventually pyopenssl hopefully.

reaperhulk:
Thanks for contributing! (Feel free to send a PR to add your name to `AUTHORS.rst`)

You can actually submit a PR to pyOpenSSL now if you'd like. The CI job will fail, but we have several jobs in the CI that test against the master branch of cryptography so if those pass then once we do a cryptography release all you'll need to do is rebase (and bump the minimum version of cryptography required).

PR #4112, opened 2018-02-21
openssl 1.1.1 testing

reaperhulk:


reaperhulk:
Once this lands I can submit the ed25519 PR, although we probably shouldn't land it before 1.1.1 final.

alex:
```
19:58:12 <reaperhulk> (I don't know if we actually want to merge it or not. It passes all tests so we could, but it's also unreleased software so ¯\_(ツ)_/¯)
20:00:21 <Alex_Gaynor> reaperhulk: looks good, dunno if we want to land it either
20:00:34 <Alex_Gaynor> reaperhulk: maybe let's keep it fresh and land when we get closer to their final release?
20:00:40 <reaperhulk> seems fair
```

alex:
@reaperhulk Can you please bump this to pre2?

Issue #4109, opened 2018-02-16
Expose `BN_rand` operations?


cygnusv:
Right now it is difficult to implement advanced cryptographic algorithms on top of cryptography.io. 
One of the problems is sampling random big numbers. This could be easily solved by exposing some of the `BN_rand` operations from OpenSSL:

https://www.openssl.org/docs/man1.0.2/crypto/BN_rand.html

I think that it is only necessary to expose `BN_rand` and `BN_rand_range`. Perhaps, this also could be used inside cryptography.io (not sure about this).

reaperhulk:
As discussed in IRC there's some questions about whether we want to expose arbitrary bignum math (or ec point math) to enable primitives to be built from our base. We have been reluctant to do this since the probability of misuse feels pretty high (and safely using these super-primitives can be difficult since the Python layer can potentially introduce a variety of side channels depending on how it's utilized).

If you'd like to discuss whether it makes sense to build an API for this sort of thing in cryptography we can go down that path, but I have no objection to adding `BN_rand` as a binding in the short term. It exists in every OpenSSL we support and is also available in boring (something we'd like to potentially target in the future).

cygnusv:
That sounds reasonable. I feel that the scope of that discussion (i.e. allowing arbitrary operations for bignums and EC points) is much bigger than this current issue, so probably you'd want to handle that separately. For sure @tuxxy and myself will be very interested in weighing in.

For our short-term needs, `BN_rand` operations are enough. 

Thanks for your time!

PR #4118, opened 2018-02-23
Expose BN_rand operations

tuxxy:
### What this does:
1. Exposes `BN_rand` and `BN_rand_range` in the OpenSSL backend.
2. Solves #4109


tests/hazmat/backends/test_openssl.py
```diff
@@ -169,6 +169,32 @@ def test_bn_to_int(self):
         bn = backend._int_to_bn(0)
         assert backend._bn_to_int(bn) == 0
 
+    def test_bn_rand(self):
```
alex: We don't need tests for bindings like these. This code can be removed.

tuxxy: 👍

Issue #3740, opened 2017-06-30
Add certbot to downstream tests in CI


alex:

alex:
@SwartzCr if we want to run the `certbot` tests in our CI -- but with a local cryptography (from some random git rev) instead of the pypi package -- how should we go about it? The only important bits are the one that use cryptography, so I think we're primarily interested in `acme` and `cerbot`, not the various DNS or server packages (but if it's easier to run everything I won't say no :-))

PR #4120, opened 2018-02-24
Add certbot to downstream tests (#3740)

jlaine:


reaperhulk:
This looks right to me but you're correct that jenkins isn't building it. I wonder if this is some security restriction by jenkins since you're not a maintainer for the project? I'll cherry pick this commit and submit as a separate PR as a quick experiment.

PR #4122, opened 2018-02-25
Add certbot to downstream tests (#3740)

reaperhulk:
This is a direct clone of #4120 to see if jenkins likes it better when I submit.


reaperhulk:
Jenkins is too clever and matched the commit hash. /me adds an empty commit.

reaperhulk:
It would appear that's precisely why. Once this passes (to confirm certbot downstream works as expected) I'll close this PR and merge the original.

Issue #4005, opened 2017-10-31
DH docs should do a better job telling people to use a KDF


alex:
https://cryptography.io/en/latest/hazmat/primitives/asymmetric/x25519/#exchange-algorithm (and the same in the ECDH and FFDH docs)

We should a) include a KDF in the example, b) use more words to describe wtf we're saying about kdfs and better motivate the _why_

jlaine:
This is a little arcane for a "good first issue"!

From my reading so far, this is what I've come up with : 

    For most applications the shared_key should be passed to a key derivation function in order to produce a key with the expected number of random bits, and to destroy mathematical structure in the shared_key which might be used by an attacker.

PR #4124, opened 2018-02-28
Document motivation for a KDF after key-exchange (#4005)

jlaine:


jlaine:
I don't understand the codecov failure, I haven't touched any code..

reaperhulk:
I still don't have a better suggestion for this language, but in the meantime the examples should be updated to use a KDF.

docs/hazmat/primitives/asymmetric/dh.rst
```diff
@@ -19,7 +19,9 @@ Exchange Algorithm
 ~~~~~~~~~~~~~~~~~~
 
 For most applications the ``shared_key`` should be passed to a key
-derivation function.
+derivation function in order to produce a key with the expected number of
+random bits, and to destroy mathematical structure in the shared key which
```
reaperhulk: After soliciting several people for ideas on phrasing here how about:

"...should be passed to a key derivation function. This allows mixing of additional information into the key, derivation of multiple keys, and destroys any structure that may be present."

Not really much better I guess ;)

jlaine:
For future reference, the output of exectack looks like:

```
- execstack.check/cryptography/hazmat/bindings/_constant_time.abi3.so
- execstack.check/cryptography/hazmat/bindings/_openssl.abi3.so
- execstack.check/cryptography/hazmat/bindings/_padding.abi3.so
```

If I enable exectack on a .so it looks like:

```
- execstack.check/cryptography/hazmat/bindings/_constant_time.abi3.so
X execstack.check/cryptography/hazmat/bindings/_openssl.abi3.so
- execstack.check/cryptography/hazmat/bindings/_padding.abi3.so
```

=> the test is "if any line starts with X", we have a problem

PR #4127, opened 2018-03-05
add X509_CRL_dup function

ddcc:
required by https://github.com/pyca/pyopenssl/pull/367


src/_cffi_src/openssl/x509.py
```diff
@@ -144,6 +144,7 @@
 int X509_REVOKED_set_revocationDate(X509_REVOKED *, ASN1_TIME *);
 
 X509_CRL *X509_CRL_new(void);
+X509_CRL *X509_CRL_dup(X509_CRL *crl);
```
reaperhulk: Our style for bindings is to not include variable names here. Otherwise this LGTM.

ddcc: Done

PR #4128, opened 2018-03-05
don't allow GeneralNames to be an empty list

### LINKED ISSUE #4102: SubjectAlternativeName() should not allow to pass empty list

reaperhulk:
It doesn't make logical sense (at least in our higher level APIs) to allow empty GeneralNames.

fixes #4102


alex:
flake8 issue needs resolving

Issue #3994, opened 2017-10-23
Add a test in our wheel builder that all .so's are noexecstack


alex:
Follow up to #3993

jlaine:
How can I test a change to the Jenkins config?

It looks like we need something like:

    unzip tmpwheelhouse/*.whl -d execstack.check
    (execstack execstack.check/cryptography/hazmat/bindings/*.so | grep '^X') && exit 1

reaperhulk:
You can't directly test the change, what you'd need to do is grab the docker image and run the script yourself to see if it works as expected.

```docker run --rm -ti pyca/cryptography-manylinux1:x86_64 /bin/bash``` will get you in.

PR #4130, opened 2018-03-05
Check all .so's are noexecstack (#3994)

jlaine:


.jenkins/Jenkinsfile-cryptography-wheel-builder
```diff
@@ -134,6 +134,8 @@ def build(version, label, imageName) {
                         CFLAGS="-I/opt/pyca/cryptography/openssl/include -Wl,--exclude-libs,ALL" \
                         $linux32 /opt/python/$version/bin/pip wheel cryptography==$BUILD_VERSION -w tmpwheelhouse/ --no-binary cryptography --no-deps
                     $linux32 auditwheel repair tmpwheelhouse/cryptography*.whl -w wheelhouse/
+                    unzip wheelhouse/*.whl -d execstack.check
+                    (execstack execstack.check/cryptography/hazmat/bindings/*.so | grep '^X') && exit 1
```
reaperhulk: I don't believe execstack is present in the cryptography-manylinux1 image. Additionally, the vagaries of docker in jenkins (which is itself in a docker container) means you'll need to `chmod -R 777 execstack.check` or else we'll have permissions errors on cleanup.

To add execstack you'll want to do a `RUN yum -y install prelink && yum -y clean all` in the existing dockerfile (https://github.com/pyca/infra/blob/master/cryptography-manylinux1/Dockerfile)

jlaine: Can I count on "unzip" being present?

alex: Yes, unzip is in the image

reaperhulk: Okay execstack is now in the container, but the chmod will still need to be done.

jlaine:
For future reference, the output of exectack looks like:

```
- execstack.check/cryptography/hazmat/bindings/_constant_time.abi3.so
- execstack.check/cryptography/hazmat/bindings/_openssl.abi3.so
- execstack.check/cryptography/hazmat/bindings/_padding.abi3.so
```

If I enable exectack on a .so it looks like:

```
- execstack.check/cryptography/hazmat/bindings/_constant_time.abi3.so
X execstack.check/cryptography/hazmat/bindings/_openssl.abi3.so
- execstack.check/cryptography/hazmat/bindings/_padding.abi3.so
```

=> the test is "if any line starts with X", we have a problem

PR #4133, opened 2018-03-05
install only the minimum required to build our docs for rtd

### LINKED ISSUE #4132: Figure out how to fix docs build on rtd
It appears to be running into https://github.com/sphinx-doc/sphinx/issues/3976

reaperhulk:
We don't need to install sphinxcontrib-spelling or doc8 in rtd so let's stop trying.

Requires a modification to the rtd config after merge (to point at rtd-requirements.txt)

Fixes #4132


setup.py
```diff
@@ -296,12 +296,14 @@ def run_tests(self):
         ":platform_python_implementation != 'PyPy'": ["cffi >= 1.7"],
 
         "test": test_requirements,
-        "docstest": [
-            "doc8",
-            "pyenchant >= 1.6.11",
+        "docs": [
             "readme_renderer >= 16.0",
```
alex: Isn't this for docstest?

PR #4129, opened 2018-03-05
Brainpool curves

reaperhulk:
Add the brainpool curves required for German smart meters and no more than that.

supersedes #4021

PR #4141, opened 2018-03-14
add brainpool p256r1, p384r1, and p512r1 test vectors from OpenSSL

reaperhulk:
refs #4129 

This is a reformatting of data from https://github.com/openssl/openssl/blob/346149c164605cf0e1b930d3f42a1f28f3babcae/test/recipes/30-test_evp_data/evppkey.txt#L4981


alex:
Does adding vectors from OpenSSL which we will then use as tests against OpenSSL make any sense?

reaperhulk:
Running the test vectors ourselves does confirm that our code properly calls the underlying functions. I guess it might guard against regression in OpenSSL if they update something and improperly change the vectors, but that's not really a very realistic case. I mostly want these to confirm that we haven't mis-wired anything when doing ECDH with unusual curves.

alex:
Ok. Are there no 3rd party vectors we can use?

reaperhulk:
https://tools.ietf.org/html/rfc7027#appendix-A may be a good alternative. Could probably reformat those into the NIST ECDH vector format pretty easily.

Issue #4091, opened 2018-01-25
feature request: API to check TTL on secure message


janwijbrand:
For secure messages with a short TTL it can be useful to know if the TTL is about to expire, so a renewed message can be issued. Would it technically be possible to have a API call on the `cryptography.fernet.Fernet()` instance to determine how many seconds are left?

reaperhulk:
The TTL is not part of the encrypted payload so parsing it is easy. However, unless you verify the HMAC you don't know if the TTL has been tampered with. That may be safe in some use cases (https://github.com/dolph/keyless-fernet allows this), but is not generically secure. I'm opposed to adding anything that parses TTL without checking HMAC, but would be happy to take a patch to add an API that, given a fernet key, tells you how many seconds are left until expiry (or raises a token error if invalid).

janwijbrand:
Thanks for the reply.

I did assume the API would indeed require the key in order to have a validated check on the TTL. I have no idea yet whether I can come up with a patch. Any pointers where to start is appreciated.
.

reaperhulk:
You might take a look at how we calculate whether ttl has expired on decrypt (https://github.com/pyca/cryptography/blob/master/src/cryptography/fernet.py#L97-L103) and see if it makes sense to refactor that into a public method that raises or returns.

PR #4144, opened 2018-03-16
Add API for retrieving the seconds-to-expiry for the token, given a TTL.

janwijbrand:
As discussed in issue #4091 perhaps there could be an API to check the number of seconds before the token expires (against a given TTL). This PR is still work in progress. I'd like to know if the implementation is anywhere near acceptable. If it is, I'll add TTL support to MultiFernet and add documentation. Thanks in advance for your feedback.

### DIFF
--- a/docs/fernet.rst
+++ b/docs/fernet.rst
@@ -91,7 +91,7 @@ has support for implementing key rotation via :class:`MultiFernet`.
         :raises cryptography.fernet.InvalidToken: If the ``token``'s signature
                                                   is invalid, this exception
                                                   is raised.
-        :raises TypeError: This exception is raised if ``data`` is not
+        :raises TypeError: This exception is raised if ``token`` is not
                            ``bytes``.
 
 
--- a/docs/fernet.rst
+++ b/docs/fernet.rst
@@ -80,6 +80,20 @@ has support for implementing key rotation via :class:`MultiFernet`.
         :raises TypeError: This exception is raised if ``token`` is not
                            ``bytes``.
 
+    .. method:: age(token)
+
+        Computes the age of the token. The caller can then decide if the token
+        is about to expire and for example reissue a new token.
+
+        :param bytes token: The Fernet token. This is the result of calling
+                            :meth:`encrypt`.
+        :returns int: The token's age in seconds.
+        :raises cryptography.fernet.InvalidToken: If the ``token``'s signature
+                                                  is way invalid, this
+                                                  exception is raised.
+        :raises TypeError: This exception is raised if ``data`` is not
+                           ``bytes``.
+
 
 .. class:: MultiFernet(fernets)
 
--- a/docs/fernet.rst
+++ b/docs/fernet.rst
@@ -89,8 +89,8 @@ has support for implementing key rotation via :class:`MultiFernet`.
                             :meth:`encrypt`.
         :returns int: The token's age in seconds.
         :raises cryptography.fernet.InvalidToken: If the ``token``'s signature
-                                                  is way invalid, this
-                                                  exception is raised.
+                                                  is invalid, this exception
+                                                  is raised.
         :raises TypeError: This exception is raised if ``data`` is not
                            ``bytes``.
 
--- a/src/cryptography/fernet.py
+++ b/src/cryptography/fernet.py
@@ -74,6 +74,13 @@ class Fernet(object):
         timestamp, data = Fernet._get_unverified_token_data(token)
         return self._decrypt_data(data, timestamp, ttl)
 
+    def ttl(self, token, ttl):
+        current_time = int(time.time())
+        timestamp, data = Fernet._get_unverified_token_data(token)
+        # Verify the token was not tampered with, but do not fail on expiry.
+        self._decrypt_data(data, timestamp, None)
+        return ttl - (current_time - timestamp)
+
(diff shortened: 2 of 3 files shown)


src/cryptography/fernet.py
```diff
@@ -74,6 +74,13 @@ def decrypt(self, token, ttl=None):
         timestamp, data = Fernet._get_unverified_token_data(token)
         return self._decrypt_data(data, timestamp, ttl)
 
+    def ttl(self, token, ttl):
+        current_time = int(time.time())
+        timestamp, data = Fernet._get_unverified_token_data(token)
+        # Verify the token was not tampered with, but do not fail on expiry.
+        self._decrypt_data(data, timestamp, None)
```
reaperhulk: This will, in addition to verifying the HMAC, decrypt the data. While you're throwing that away we should probably just refactor `_decrypt_data` so that these lines (https://github.com/pyca/cryptography/blob/master/src/cryptography/fernet.py#L105-L110) are in a separate method that we can call.

tests/test_fernet.py
```diff
@@ -122,6 +123,51 @@ def test_bad_key(self, backend):
...
+        ts = "1985-10-26T01:20:01"
+        current_time = calendar.timegm(iso8601.parse_date(ts).utctimetuple())
+        monkeypatch.setattr(time, "time", lambda: current_time)
+
+        issued_ts = struct.pack('>Q', calendar.timegm(
+            datetime.datetime(1985, 10, 26, 1, 19, 1).utctimetuple()))
+
+        f = Fernet(base64.urlsafe_b64encode(b"\x00" * 32), backend=backend)
+        with pytest.raises(InvalidToken):
+            token = base64.urlsafe_b64encode(b"\x80" + issued_ts)
```
reaperhulk: Instead of trying to construct the token manually please just call `encrypt` on Fernet to get a real token.

tests/test_fernet.py
```diff
@@ -122,6 +123,51 @@ def test_bad_key(self, backend):
...
+        current_time = calendar.timegm(iso8601.parse_date(ts).utctimetuple())
+        monkeypatch.setattr(time, "time", lambda: current_time)
+        assert f.ttl(token, ttl=60) == 1
+
+        # Check the TTL two minutes later. We're a minute over due in this
+        # case.
+        ts = "1985-10-26T01:21:00"
+        current_time = calendar.timegm(iso8601.parse_date(ts).utctimetuple())
+        monkeypatch.setattr(time, "time", lambda: current_time)
+        assert f.ttl(token, ttl=60) == 1
```
reaperhulk: This does not appear to test what the comment says (it is a duplicate of the previous).

janwijbrand:
Thanks for the feedback @reaperhulk. Concerning "why pass a TTL?": you are right, the caller can decide if a new token needs to be handed out based on the age of the current token. So'd rename `def ttl()` to `def age()` then.

I'll address the other comments too.

janwijbrand:
By "just" computing the age, I think the "Also, just to complicate things a bit we actually allow a bit of slop in calculating expiry of a token"-comment is no longer relevant.

tests/test_fernet.py
janwijbrand: Ah, that was a mistake. Fixed it.

tests/test_fernet.py
janwijbrand: Fixed.

src/cryptography/fernet.py
janwijbrand: I factored it out to `def _verify_signature()`.

janwijbrand:
I think I processed the comment and feedback. Please let me know if there's anything missing.

tests/test_fernet.py
```diff
@@ -122,6 +122,44 @@ def test_bad_key(self, backend):
         with pytest.raises(ValueError):
             Fernet(base64.urlsafe_b64encode(b"abc"), backend=backend)
 
+    def test_timestamp_ttl_overdue(self, monkeypatch, backend):
```
reaperhulk: This test doesn't appear to be related to this feature. We should already have this test elsewhere in `test_fernet`.

janwijbrand: removed duplicate test

docs/fernet.rst
```diff
@@ -80,6 +80,20 @@ has support for implementing key rotation via :class:`MultiFernet`.
...
+        Computes the age of the token. The caller can then decide if the token
+        is about to expire and for example reissue a new token.
+
+        :param bytes token: The Fernet token. This is the result of calling
+                            :meth:`encrypt`.
+        :returns int: The token's age in seconds.
+        :raises cryptography.fernet.InvalidToken: If the ``token``'s signature
+                                                  is invalid, this exception
+                                                  is raised.
+        :raises TypeError: This exception is raised if ``data`` is not
```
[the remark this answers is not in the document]
janwijbrand: fixed.

janwijbrand:
Anything else I could do to refine the PR for it to be merged?

janwijbrand:
Please let me know if I need to amend or additional work for the PR to be mergeable.

reaperhulk:
This mostly looks good. After some discussion we have some additional API changes we'd like to see, but I'll go ahead and cherry-pick this and make the changes.

PR #4145, opened 2018-03-18
DSA parameters should support 224 for q length

outcoldman:
I am trying to make Golang to talk to Python and using this library.
In golang keys are generated with `L2048N224` (https://golang.org/pkg/crypto/dsa/#ParameterSizes) and after making this small fix I do not see any issue - I can load my parameters and work with my keys.

### DIFF
--- a/src/cryptography/hazmat/primitives/asymmetric/dsa.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/dsa.py
@@ -130,8 +130,8 @@ def generate_private_key(key_size, backend):
 def _check_dsa_parameters(parameters):
     if parameters.p.bit_length() not in [1024, 2048, 3072]:
         raise ValueError("p must be exactly 1024, 2048, or 3072 bits long")
-    if parameters.q.bit_length() not in [160, 256]:
-        raise ValueError("q must be exactly 160 or 256 bits long")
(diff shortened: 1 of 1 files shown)


outcoldman:
And if somebody is looking for the same problem, workaround is


```
import collections

# create your own class with p, q, g
DSAParameterNumbers = collections.namedtuple('DSAParameterNumbers', ('p', 'q', 'g'))
parameters = default_backend().load_dsa_parameter_numbers(parameters_numbers)
```

reaperhulk:
I believe this is an artifact of the past when OpenSSL didn't support this. We actually have test vectors available for q=224, but they're being filtered out of the vector set as we load them right now. I'll put up a PR that fixes that + cherry picks this commit and we'll see if CI is happy 😄

reaperhulk:
Okay, merged as #4147 thanks for the contribution! You can send a PR adding yourself to `AUTHORS.rst` if you'd like!

outcoldman:
@reaperhulk thank you for such a quick merge! (also created https://github.com/pyca/cryptography/pull/4148)

Curious, when can I see this in release on pypi?

reaperhulk:
We're planning to do a 2.2 early next week.

PR #4146, opened 2018-03-18
remove arbitrary rc4 key size limitations

jbremer:
What's the reasoning behind a limited set of possible key sizes for `ARC4` (https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/primitives/ciphers/algorithms.py#L106) when the OpenSSL backend can take any key size up to 256 bytes (https://github.com/openssl/openssl/blob/master/crypto/rc4/rc4_skey.c#L45)?
Note that `PyCrypto` also doesn't have this limitation.
No additional unit tests are included (sorry), but the existing ones should still work.

### DIFF
--- a/src/cryptography/hazmat/primitives/ciphers/algorithms.py
+++ b/src/cryptography/hazmat/primitives/ciphers/algorithms.py
@@ -103,10 +103,9 @@ class CAST5(object):
 @utils.register_interface(CipherAlgorithm)
 class ARC4(object):
     name = "RC4"
-    key_sizes = frozenset([40, 56, 64, 80, 128, 160, 192, 256])
 
     def __init__(self, key):
-        self.key = _verify_key_size(self, key)
+        self.key = key
 
     @property
     def key_size(self):


alex:
I'm loathe to expand this given how awful an alogrithm RC4 is. What's the use case here?

jbremer:
Virtually any user of RC4, including the examples displayed at the RC4 wikipedia page, uses arbitrarily-sized keys. This includes proprietary protocols etc.

alex:
Do you have a specific use case, or is this just a general change?

jbremer:
As an example, the Panda Banker uses (among other sizes) a 256-byte key for decrypting specific data (target URLs of the C2), see also https://www.arbornetworks.com/blog/asert/let-pandas-zeus-zeus-zeus-zeus/.

alex:
Is this something you're running into, or just an example?

jbremer:
It's something where I'm currently forced to use `PyCrypto` instead of `cryptography`, yes ;)

reaperhulk:
Okay we've talked about this in IRC quite a bit and our consensus is that we're not going to expand the set. For weird edge cases like this you can easily monkeypatch `_verify_key_size` if you want to bypass the restriction. While that is not public API and we don't support it, it's very unlikely to be broken by us either.

PR #4152, opened 2018-03-19
don't make a wheel for vectors

reaperhulk:
It makes me upload 26MB more data and almost nobody will install the vectors anyway.

### DIFF
--- a/release.py
+++ b/release.py
@@ -100,7 +100,7 @@ def release(version):
     run("git", "push", "--tags")
 
     run("python", "setup.py", "sdist")
-    run("python", "setup.py", "sdist", "bdist_wheel", cwd="vectors/")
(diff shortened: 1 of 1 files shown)


alex:
I'm soft opposed to this; the motivation is kind of questionable  ;-) and wheeling everything is cool!

reaperhulk:
Heh, the motivation is based entirely on bandwidth available. https://github.com/pypa/warehouse/issues/994#issuecomment-371987295 would make this unnecessary since we could delegate release responsibility more securely.

alex:
I'm going to wontfix, on the basis of the fact that there's a better long term solution coming down the pike, and we don't do releases that frequently -- next time just ask me to run the script :-)

PR #4155, opened 2018-03-19
Add urllib3 to downstream tests

jlaine:


jlaine:
Don't trust the "ok" on the tests, since Jenkins did not pick up the modified Jenkinsfile

jlaine:
Can we merge this if urllib3 is now unbroken?

reaperhulk:
I need to hijack it and submit it as a PR to verify that it all works before we can merge.

jlaine:
Sounds good. If you're wondering about the [socks] bit I blindly copied it over from pyopenssl's downstream test.

PR #4180, opened 2018-03-30
Raise ve on bad gt

jcrowgey:
Instead of `openssl_assert` that the return from `ASN1_TIME_to_generalizedtime` is not null, we can check if it is null (indicating a failure to convert the time) and then raise a ValueError.


src/cryptography/hazmat/backends/openssl/decode_asn1.py
```diff
@@ -734,7 +734,15 @@ def _parse_asn1_time(backend, asn1_time):
     generalized_time = backend._lib.ASN1_TIME_to_generalizedtime(
         asn1_time, backend._ffi.NULL
     )
-    backend.openssl_assert(generalized_time != backend._ffi.NULL)
+    if generalized_time == backend._ffi.NULL:
+        raise ValueError(
+            "Couldn't parse ASN.1 time as generalizedtime {}".format(
```
alex: Switch `{}` to `{!r}` please, and drop the `.decode()` call.

tests/x509/test_x509.py
```diff
@@ -3752,6 +3752,18 @@ def test_unsupported_subject_public_key_info(self, backend):
...
+    def test_bad_time_in_validity(self, backend):
+        cert = _load_cert(
+            os.path.join(
+                "x509", "badasn1time.pem"
+            ),
+            x509.load_pem_x509_certificate,
+            backend,
+        )
+
+        with pytest.raises(ValueError):
```
alex: Please add an assert that the message on the exception contains the expected value

reaperhulk: This can be done with `pytest.raises(ValueError, match='string')` btw

jcrowgey: I guess that value and contains are slightly ambiguous here.  To clarify, would you prefer something like this:

```
assert exc.value.message == "Couldn't parse ASN.1 time as generalizedtime '19020701025736Z'"
```

Or something like this:

```
assert exc.value.message.find('19020701025736Z') > -1
```

reaperhulk: The former please. Yuo can do `with pytest.raises(ValueError, match='Couldn't parse ASN.1 time as generalizedtime'):` and it will do a match on that.

alex:
Please match on the date string, not the rest of the message :-)

On Mar 30, 2018 6:07 PM, "Paul Kehrer" <notifications@github.com> wrote:

*@reaperhulk* commented on this pull request.
------------------------------

In tests/x509/test_x509.py
<https://github.com/pyca/cryptography/pull/4180#discussion_r178392932>:

> @@ -3752,6 +3752,18 @@ def test_unsupported_subject_public_key_info(self, backend):
             cert.public_key()

+    def test_bad_time_in_validity(self, backend):
+        cert = _load_cert(
+            os.path.join(
+                "x509", "badasn1time.pem"
+            ),
+            backend,
+        )
+

The former please. Yuo can do with pytest.raises(ValueError,
match='Couldn't parse ASN.1 time as generalizedtime'): and it will do a
match on that.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/pyca/cryptography/pull/4180#discussion_r178392932>, or mute
the thread

tests/x509/test_x509.py
jcrowgey: Oh, that's very nice.  I will push one more commit.

PR #4184, opened 2018-04-04
OpenSSL 1.0.2o has switched to winsock2

thierryba:
So here we need to make sure we don't simply include windows but only the parts that we want.

### DIFF
--- a/src/_cffi_src/openssl/callbacks.py
+++ b/src/_cffi_src/openssl/callbacks.py
@@ -11,7 +11,10 @@ INCLUDES = """
 #include <openssl/crypto.h>
 
 #ifdef _WIN32
+#define WIN32_LEAN_AND_MEAN
 #include <Windows.h>
+#include <Wincrypt.h>
+#include <Winsock2.h>
 #else
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


alex:
Can you explain what problem this is solving? I just realized we don't have CI for Windows+1.0.2, we only test 1.1.0 with Windows in CI.

Do we fail to build on 1.0.2 on Windows?

Thanks!

thierryba:
Yes it fails with toms of errors about symbols being redeclared with different linkage type. That seems to be happening from including both winsock 1 and 2.
So here I'm define WIN32_LEAN_AND_MEAN to avoid including the whole windows world in 1 include statement ;). And then the only thing that was still needed was to add wincrypt.h

thierryba:
Of course from the logs it looks like for some other setup this is failing...

reaperhulk:
Looks like `    build\temp.win32-2.7\Release\_openssl.c(2285) : error C2079: 'tv' uses undefined struct 'timeval'`

I believe that comes from `Winsock2.h` so that might be required as well?

thierryba:
let me add it... but it might then fail because openssl prior to 1.0.2o is including winsock.h...

reaperhulk:
Well this passes our CI -- @thierryba are you able to confirm if this works on OpenSSL < 1.0.2o?

reaperhulk:
Thank you for the contribution!

PR #4194, opened 2018-04-14
Update URLs for new pypi!

alex:


alex:
We're blocked ATM until either https://github.com/msgpack/msgpack-python/issues/299 is resolved of we have infra blacklist pip x on wheezy

PR #4200, opened 2018-04-18
Expose OpenSSL constant time bignum arithmetic

### LINKED ISSUE #4199: Some OpenSSL bignum arithmetic operations are not in constant time
### What's wrong:
1. Some arithmetic operations exposed in the [OpenSSL bignum binding](https://github.com/pyca/cryptography/blob/master/src/_cffi_src/openssl/bignum.py) are not in constant time.
2. These functions are specifically:
    1. `BN_div`
    2. `BN_mod_inverse`
    3. `BN_mod_exp`

### How to resolve:
1. The OpenSSL solution to this is to call `BN_set_flags` on the secret BIGNUM and set the `BN_FLG_CONSTTIME` flag. [The OpenSSL docs reference this here.](https://github.com/openssl/openssl/blob/master/include/openssl/bn.h#L61)
2. Expose the needed operations:
    1. `BN_MONT_CTX` operations.
    2. `BN_mod_exp_mont`, and `BN_mod_exp_mont_consttime` 
    3. `BN_set_flags`, `BN_get_flags`, and the `BN_FLG_CONSTTIME` flag.

### Notes:
1. Cryptography.io isn't affected by these timing issues. I have clarified with @alex and @reaperhulk.
2. If you are a downstream user utilizing these operations on private/secret values, then _you are affected_.

tuxxy:
### What this does:
1. Exposes the following:
    1. `BN_set_flags` and `BN_get_flags`
    2. `BN_MONT_CTX_new`, `BN_MONT_CTX_set`, and `BN_MONT_CTX_free`
    3. `BN_mod_exp_mont`, `BN_mod_exp_mont_consttime`
    4. `BN_FLG_CONSTTIME` flag for setting BNs to use constant time operations in `BN_div`, `BN_mod_inverse`, and `BN_mod_exp_mont`.
2. Resolves #4199

### Notes:
1. When `BN_FLG_CONSTTIME` is set on an OpenSSL BN and `BN_mod_inverse`, `BN_div`, or `BN_mod_exp_mont` is called OpenSSL will, in turn, call `BN_mod_inverse_no_branch`, `BN_div_no_branch`, and `BN_mod_exp_mont_consttime` respectively.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -40,3 +40,4 @@ PGP key fingerprints are enclosed in parentheses.
 * Chris Wolfe <chriswwolfe@gmail.com>
 * Jeremy Lainé <jeremy.laine@m4x.org>
 * Denis Gladkikh <denis@gladkikh.email>
+* John Pacific <me@johnpacific.com> (2CF6 0381 B5EF 29B7 D48C 2020 7BB9 71A0 E891 44D9)
--- a/src/_cffi_src/openssl/bignum.py
+++ b/src/_cffi_src/openssl/bignum.py
@@ -10,11 +10,17 @@ INCLUDES = """
 
 TYPES = """
 typedef ... BN_CTX;
+typedef ... BN_MONT_CTX;
 typedef ... BIGNUM;
 typedef int... BN_ULONG;
 """
 
 FUNCTIONS = """
+#define BN_FLG_CONSTTIME ...
+
+void BN_set_flags(BIGNUM *, int);
+int BN_get_flags(const BIGNUM *, int);
+
 BIGNUM *BN_new(void);
 void BN_free(BIGNUM *);
 void BN_clear_free(BIGNUM *);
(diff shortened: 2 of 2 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1461,6 +1461,17 @@ def _tmp_bn_ctx(self):
         finally:
             self._lib.BN_CTX_end(bn_ctx)
 
+    @contextmanager
+    def _tmp_bn_mont_ctx(self, modulus):
```
reaperhulk: Since we don't use this in cryptography this will cause a failure in CI since it is uncovered code. This probably belongs in a third party library unless/until we have a need for BN with montgomery multiplication for a primitive in cryptography.

tuxxy:
I mean, it's ultimately up to you guys, but I don't see a point in not
including it, to be honest.

The library already exposes arithmetic that it doesn't use. Why not expose
this to allow easier/cleaner operations?
I'll certainly remove it if needed, but it seems like such an obvious
accesory to the rest of what's provided.

The bindings and interfaces here are so nice for prototyping and
experimenting (seriously, it's easy af to do ECC prototypes with this lib).
I would hate to leave an interface like this up to eventual tech debt
someday.

alex:
Yes, I agree with @reaperhulk, there's no need to add the private python method that we don't use.

tuxxy:
Alright, I'll pull it out and push it up.

tuxxy:
Just rebased and cleaned up the commit history. Any updates on this?

PR #4202, opened 2018-04-20
add custom extensions functions for openssl >=1.0.2

DenisLila:
Add these methods https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_add_client_custom_ext.html


alex:
Thanks for sending a PR! What's the use case for adding these?

DenisLila:
We'd like to start using [Token Binding](https://datatracker.ietf.org/doc/draft-ietf-tokbind-protocol) for our auth tokens in some of our python tools, and I believe we need at least client_custom_ext to [negotiate it](https://datatracker.ietf.org/doc/draft-ietf-tokbind-negotiation).

alex:
👍

PR #4205, opened 2018-04-23
Add Session functions, necessary to implement new features in Python 3.6

amauryfa:
This change was already applied in the PyPy repository, and test_ssl passes.


alex:
Looks like there's a compilation error on some OpenSSL versions.

reaperhulk:
It appears to fail on any version of OpenSSL < 1.1.0 because the fallback getters attempt to access struct elements that don't exist. In < 1.1.0 they were named `tlsext_tick` and `tlsext_tick_lifetime_hint` so the functions need to be modified for that.

amauryfa:
Updated. The tests pass now.

PR #4208, opened 2018-04-24
Add more verbose logging to try to diagnose the cause of the pip inst…

alex:
…all failures in CI


alex:
This is wrong, the install is failing in tox...

PR #4210, opened 2018-04-29
Fix build with LibreSSL 2.7

Sp1l:
For readability of the block I duplicated the version stuff in `src/_cffi_src/openssl/cryptography.py`.
The rest are small changes for things that LibreSSL has not added from the 1.1.0 branch.

Tested with
  - OpenSSL 1.1.0h
  - LibreSSL 2.7.2
on 2.1.4 version

Signed-off-by: Bernard Spil <brnrd@FreeBSD.org>

PR #4211, opened 2018-04-30
Test libre

reaperhulk:
This tests #4210 with libre 2.7.2 in jenkins


Sp1l:
All packages are OK on FreeBSD. Cross-tested

 - Python 2.7 & 3.6
 - FreeBSD 10.4 & 11.1
 - amd64 & i386
 - OpenSSL 1.0.1u & 1.0.2o & 1.1.0h & 1.1.1-pre5 & LibreSSL 2.7.2

Build logs (poudriere) on https://keg.brnrd.eu/ (enumerating all 32 logs here was a bit much)

reaperhulk:
The CI is failing on LibreSSL 2.7.2 here. It looks like the defines probably aren't quite right. There are several symbols not present in 2.7.2 (You can see them in the error output from jenkins)

cactus:
Does something need to be added [here](https://github.com/pyca/cryptography/blob/c81df6b255e4e767419bccbeff033497fd1324e9/src/_cffi_src/openssl/ssl.py#L691) so libressl also takes the `Cryptography_HAS_GENERIC_DTLS_METHOD = 0` path?

Sp1l:
I've been digging and found some more issues where LibreSSL is missing methods. Expect an update to the 4210 issue shortly.

Sp1l:
Just pushed an update to #4210 that fixes all 'implicit declaration' warnings during build and allows make test to run.

Issue #4136, opened 2018-03-06
[Feature request] X448


maqp:
OpenSSL [now](https://github.com/openssl/openssl/issues/1398) supports X448 key exchange for TLS. Please add bindings for it.

reaperhulk:
@maqp just to make sure I understand the request, do you want to see X448 key exchange support like we added X25519?

maqp:
@reaperhulk: Yes, a hazmat module is all I'm requesting.

reaperhulk:
Okay, this will be possible once 1.1.1 ships, although we'll need to decide what the serialization model looks like. We may also want to revisit PKCS8 support for X25519. Previously OpenSSL was using non-standard OIDs (because the standards post-dated OpenSSL's implementation) and I have no idea if that's been fixed/normalized in some fashion.

PR #4215, opened 2018-05-02
Add X448 support

derwolfe:
Adds support for #4136

A few comments and questions:

1. I don't think I'm handling the openssl version identifier correctly. I'm currently allowing any version as long as it is greater than 1.1.1 dev (I'm using pre6). 
2. I don't know asn1/pkcs8 well. To get the prefix for the key, I've just generated a key then looked for the prefix. I do think that the comment regarding asn1 would be helpful to others, but I don't know how to write it in a helpful way.
3. The exchange code is nearly identical to x25519's. Would you rather I DRY this up or just leave them effectively copied but independent?

Thanks!


derwolfe:
I’m linking against 111pre6 locally and I think I’ll need to update the travis configuration to use it too, otherwise x448 tests will fail or be skipped.

derwolfe:
> I don't think I'm handling the openssl version identifier correctly. I'm currently allowing any version as long as it is greater than 1.1.1 dev (I'm using pre6).

I've looked in the openssl includes and the version number that I think I should be targeting is `10101006L`.

docs/hazmat/primitives/asymmetric/x448.rst
```diff
@@ -0,0 +1,104 @@
...
+    >>> # In a real handshake the peer_public_key will be received from the
+    >>> # other party. For this example we'll generate another private key and
+    >>> # get a public key from that. Note that in a DH handshake both peers
+    >>> # must agree on a common set of parameters.
+    >>> peer_public_key = X448PrivateKey.generate().public_key()
+    >>> shared_key = private_key.exchange(peer_public_key)
+    >>> # Perform key derivation.
+    >>> derived_key = HKDF(
+    ...     algorithm=hashes.SHA256(),
+    ...     length=32,
```
derwolfe: I believe that these and the others should be 56 and not 32 (double checking now).

maqp:
@derwolfe Not to rush you or anything, but any thoughts on the progress of this PR?

reaperhulk:
@derwolfe has semi-recently started a new job where he has less free time for OSS. He's given his blessing for someone else to take over this PR. I'll be doing it after ed25519 and x25519 serialization if nobody else gets to it first.

PR #4219, opened 2018-05-07
Remove unused BN bindings.

davidben:
These appear to be unused in both cryptography.io and PyOpenSSL.

### DIFF
--- a/src/_cffi_src/openssl/bignum.py
+++ b/src/_cffi_src/openssl/bignum.py
@@ -10,14 +10,21 @@ INCLUDES = """
 
 TYPES = """
 typedef ... BN_CTX;
+typedef ... BN_MONT_CTX;
 typedef ... BIGNUM;
 typedef int... BN_ULONG;
 """
 
 FUNCTIONS = """
+#define BN_FLG_CONSTTIME ...
+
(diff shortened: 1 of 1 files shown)


davidben:
Windows failure looks unrelated:
```
Processing c:\jenkins\workspace\4219-y7qc74epgy2ragn@4\cryptography\vectors

Collecting coverage==4.3.4

  Could not find a version that satisfies the requirement coverage==4.3.4 (from versions: )

No matching distribution found for coverage==4.3.4
```

davidben:
It was pointed out that https://github.com/nucypher/pyUmbral relies on a bunch of these symbols, so I'll update this to restore the symbols it uses.

Probably worth adding them to CI, and perhaps arranging for some more supported interface for them. In particular, they use `BN_FLG_CONSTTIME` which may be changed in OpenSSL later and is generally error-prone:
https://github.com/openssl/openssl/issues/6078

tuxxy:
@davidben Just saw this PR, thanks for thinking about us. If those get removed, we'd be up a creek without a paddle. Much appreciated! :)

With regards to getting pyUmbral in CI, if that's something the project is interested in, I think we'd be okay in assisting in that.

PR #4225, opened 2018-05-11
Clean up unused EC bindings.

davidben:
Hopefully this one is less complicated than BIO.

A lot of these are really OpenSSL internals, like the EC_METHOD
business, support for custom curves which are a bad idea, and weird
non-standard serializations like taking the usual point serialization
and treating it as a single BIGNUM.

I also didn't remove things when they're arguably part of a set. E.g.
EC_POINT_add is used, but EC_POINT_dbl isn't. However, they both set at
the same abstraction level (basic point operations), so it's strange to
have one without the other.

I also kept EC_POINT_is_on_curve because, although it is not used,
OpenSSL prior to 1.1.0 doesn't perform this important check in
EC_POINT_set_affine_coordinates_GFp (though it does in some of the
functions which ultimately call it, like
EC_KEY_set_public_key_affine_coordinates, what cryptography.io actually
uses), so one should not expose the latter without the former.

### DIFF
--- a/src/_cffi_src/openssl/ec.py
+++ b/src/_cffi_src/openssl/ec.py
@@ -27,34 +27,16 @@ typedef struct {
 typedef enum {
     POINT_CONVERSION_COMPRESSED,
     POINT_CONVERSION_UNCOMPRESSED,
-    POINT_CONVERSION_HYBRID,
     ...
 } point_conversion_form_t;
 """
 
 FUNCTIONS = """
-EC_GROUP *EC_GROUP_new(const EC_METHOD *);
 void EC_GROUP_free(EC_GROUP *);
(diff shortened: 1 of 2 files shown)


src/_cffi_src/openssl/ec.py
```diff
@@ -182,25 +114,8 @@
 #if defined(OPENSSL_NO_EC2M)
 static const long Cryptography_HAS_EC2M = 0;
 
-const EC_METHOD *(*EC_GF2m_simple_method)() = NULL;
-
-int (*EC_POINT_set_affine_coordinates_GF2m)(const EC_GROUP *, EC_POINT *,
```
alex: This was removed here, but still exists in another place in this file, leading to an error loading the .so

PR #4227, opened 2018-05-11
Fix some callback type signatures.

davidben:
Hopefully this PR is even less complicated. :-)

SSL_CTX_set_psk_server_callback:
https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_psk_server_callback.html
https://github.com/openssl/openssl/blob/OpenSSL_1_0_2/ssl/ssl.h#L1355
https://github.com/openssl/openssl/blob/OpenSSL_1_1_0/include/openssl/ssl.h#L734

SSL_CTX_set_tlsext_servername_callback:
https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_tlsext_servername_callback.html
https://github.com/openssl/openssl/blob/OpenSSL_1_0_2/ssl/s3_lib.c#L3964
https://github.com/openssl/openssl/blob/OpenSSL_1_1_0/ssl/s3_lib.c#L3499

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -738,7 +738,7 @@ void (*SSL_CTX_set_psk_server_callback)(SSL_CTX *,
                                             SSL *,
                                             const char *,
                                             unsigned char *,
-                                            int
+                                            unsigned int
                                         )) = NULL;
 void (*SSL_CTX_set_psk_client_callback)(SSL_CTX *,
                                         unsigned int (*)(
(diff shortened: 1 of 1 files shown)


src/_cffi_src/openssl/ssl.py
```diff
@@ -252,7 +252,7 @@
                                          SSL *,
                                          const char *,
                                          unsigned char *,
-                                         int
+                                         unsigned int
```
alex: https://www.openssl.org/docs/man1.1.0/ssl/SSL_CTX_set_psk_server_callback.html shows this as just `int`?

davidben: The code uses `unsigned int`, so any deviation here is just a documentation error. git says it was fixed here:
https://github.com/openssl/openssl/commit/93a048a1d8f1f6307aea0c1e7854a40b2daf236f#diff-9e87706727f317514b6408167292621f

Mismatching function pointer types, even slightly like this, is undefined behavior in C (though realistically okay). This was only building because cryptography.io doesn't treat those type errors as fatal in its build. They're just flagging warnings.

PR #4234, opened 2018-05-13
Match functions to the respective OpenSSL and LibreSSL versions (including LibreSSL 2.7!)

vishwin:
In the footsteps of #4210 and #4211, fix at least some of the remaining fallout. During this process, introduce some additional LibreSSL defines. Not only does LibreSSL 2.7 not implement every new function from OpenSSL 1.1.0, but certain functions from both OpenSSL 1.1.0 and 1.0.2 actually appeared in LibreSSL 2.1.4 and 2.5.1 instead of 2.7.0. In the future we may need some sort of separation between the OpenSSL and LibreSSL defines.

I tested this on FreeBSD 12-CURRENT under LibreSSL 2.7.3, OpenSSL 1.0.2o (from base system and ports) and OpenSSL 1.1.0h. Ran the Fernet example from the README and test imported the OpenSSL python library, both with no problems.


reaperhulk:
Looks like some pep8 issues here, but I'm concerned about the direction these macros are taking at this point. The complexity is very high and we're doing things like returning true for `CRYPTOGRAPHY_OPENSSL_110_OR_GREATER` when it's LibreSSL >= 2.7.0, but then adding additional checks in some places to say "oh but if it's *actually* libre then ignore the previous.

I don't have a good suggestion on how to handle this but I'm not comfortable merging as is.

On the bright side we'll be moving the libre builder into travis shortly so we can test 2.7.x more easily to confirm that it works in CI. Previously we tested in Jenkins but the Jenkins builders won't trust changes coming from non-committers.

vishwin:
I'll probably just separate the logic in the macros given the creeping complexity we're all seeing, as in only having the `CRYPTOGRAPHY_OPENSSL` return true when actually OpenSSL and vice versa for LibreSSL. Places where any `CRYPTOGRAPHY_OPENSSL` macros are in use will then need to add the corresponding `CRYPTOGRAPHY_LIBRESSL` macro. I think this will allow for a cleaner distinction between features found between the two SSL implementations, as said previously, different OpenSSL features appeared in LibreSSL earlier than 2.7, some don't exist in libre at all, etc. This can even account for the possibility of LibreSSL having something that OpenSSL doesn't.

I am continuing to discuss with sp1l on how to best tackle this.

alex:
Subsumed by #4270

PR #4238, opened 2018-05-14
Future proofing use of the six python version constants

ericwb:
After reading [1], noticed that cryptography uses a lot of if six.PY3
blocks. The issue with this is that whenever Python 4 is released,
this code in the else block will be executed even though it was
only intended for Python 2.

[1] http://astrofrog.github.io/blog/2016/01/12/stop-writing-python-4-incompatible-code/

Signed-off-by: Eric Brown <browne@vmware.com>


alex:
@ericwb would you mind changing all of these to use `if not six.PY2` instead? This will help us keep the diff smaller, making it easier to review.

alex:
Great, thanks. Once that's done I'll review and merge!

PR #4241, opened 2018-05-14
Validate the public/private halves of EC keys on import.

davidben:
OpenSSL's API is a little finicky. If one sets the public key before the
private key, it does not validate that they match. If set in the other
order, it does validate this.

In particular, KASValidityTest_ECCStaticUnified_NOKC_ZZOnly_init.fax
describes error code 7 as:

  Result = F (7 - IUT's Static private key d changed-prikey validity)

Reordering the two operations makes those tests to fail on key import,
which is what CAVP appears to have intended.

### DIFF
--- a/tests/hazmat/primitives/test_ec.py
+++ b/tests/hazmat/primitives/test_ec.py
@@ -1068,8 +1068,8 @@ class TestECDH(object):
                 ec._CURVE_TYPES[vector['curve']]()
             )
         )
-        # Errno 5-7 indicates a bad public or private key, this doesn't test the
-        # ECDH code at all
+        # Errno 5-7 indicates a bad public or private key, this doesn't test
+        # the ECDH code at all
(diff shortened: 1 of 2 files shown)


tests/hazmat/primitives/test_ec.py
```diff
@@ -1068,9 +1068,9 @@ def test_key_exchange_with_vectors(self, backend, vector):
                 ec._CURVE_TYPES[vector['curve']]()
             )
         )
-        # Errno 5 and 6 indicates a bad public key, this doesn't test the ECDH
-        # code at all
-        if vector['fail'] and vector['errno'] in [5, 6]:
+        # Errno 5-7 indicates a bad public or private key, this doesn't test the
```
alex: This line is longer than our 80 (79?) char max. Can you please wrap it so the linter is happy :-)

davidben: Done. (Seems to be 79.)

PR #4252, opened 2018-05-17
Make AuthorityKeyIdentifier docs reflect reality

thomdixon:
The `AuthorityKeyIdentifier.authority_cert_issuer` docs state that it returns a `Name` instance, but it [actually returns a list of `GeneralName` instances or `None`](https://github.com/pyca/cryptography/blob/master/src/cryptography/x509/extensions.py#L157).

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -1713,7 +1713,7 @@ X.509 Extensions
 
     .. attribute:: authority_cert_issuer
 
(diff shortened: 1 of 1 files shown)


thomdixon:
Broken build has nothing to do with this change (`coverage==4.3.4` distribution cannot be found on Windows x64 + Python 3.6).

thomdixon:
Thanks for the quick review!

PR #4253, opened 2018-05-18
[WIP] A strange segfault - is it just us?

jMyles:
This is a single commit with a test that shows some odd behavior that we've been experiencing on pyUmbral.  Specifically, we get a segfault when we ask whether `EC_POINT_is_on_curve` when that point isn't in fact on the curve (under a specific set of conditions).

Obviously, even if this is true, and not simply some sort of artifact of our config, I'm not looking for a merge on this actual commit; I'll refactor into a proper regression test or whatever the other contributors prefer.

### DIFF
--- a/tests/hazmat/backends/test_openssl.py
+++ b/tests/hazmat/backends/test_openssl.py
@@ -620,3 +620,57 @@ class TestOpenSSLDHSerialization(object):
         )
         with pytest.raises(ValueError):
             loader_func(key_bytes, backend)
+
+    def test_segfault_from_wrong_curve(self):
+        """
+        In pyUmbral, we're getting a strange segfault.
+
+        We have a test in which when we show that using a curve to generate a point,
+        and then a different curve for EC_GROUP_new_by_curve_name properly borks.
+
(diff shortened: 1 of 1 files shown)


jMyles:
@alex, sure man - that was expected.  I'm just wondering if there's any scenery or anything that might be causing this segfault.  Best case scenario, it isn't reproduced in the build and then we know that there's something specific to our environment.

It does however segfault when I install with a fresh cryptography.io on my system (although I'm not using any kind of isolation, so that doens't rule out a local effect).

Anyway, I know that there's not much you're going to be able / willing to do since it's a strictly private interface.  My motivation is primarily curiosity at this point.

Nice seeing you this week!  I wish we got to hang out a bit more; I did a very short PyCon this year.

jMyles:
Looks like it's just us.

jMyles:
So, to close the loop on this, it was indeed a bug in OpenSSL (albeit not present in any of the versions used in the cryptography travis build config):

https://github.com/openssl/openssl/issues/6302

PR #4259, opened 2018-05-22
Fixed build errors on HP-UX.

dumol:
Thanks for `cryptography`! We package it with https://github.com/chevah/python-package on a number of OS'es. 

In latest revisions we are using `cryptography` 2.2.2 on HP-UX with these minor changes, required for building with the proprietary compiler on HP-UX 11i v3.

### DIFF
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -46,7 +46,7 @@ def _extra_compile_args(platform):
     When we drop support for CRYPTOGRAPHY_OPENSSL_LESS_THAN_110 we can
     revisit this.
     """
-    if platform not in [ "win32", "hp-ux11" ]:
+    if platform not in ["win32", "hp-ux11"]:
         return ["-Wconversion", "-Wno-error=sign-conversion"]
     else:
         return []
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -46,7 +46,7 @@ def _extra_compile_args(platform):
     When we drop support for CRYPTOGRAPHY_OPENSSL_LESS_THAN_110 we can
     revisit this.
     """
-    if not platform in [ "win32", "hp-ux11" ]:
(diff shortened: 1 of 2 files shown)


src/_cffi_src/build_openssl.py
```diff
@@ -46,7 +46,7 @@ def _extra_compile_args(platform):
     When we drop support for CRYPTOGRAPHY_OPENSSL_LESS_THAN_110 we can
     revisit this.
     """
-    if platform != "win32":
+    if not platform in [ "win32", "hp-ux11" ]:
```
reaperhulk: To make pep8 happy this line should read `if platform not in ["win32", "hup-ux11"]:` (reordered not in and removed spaces in the list).

reaperhulk:
While we're not going to expand our test infra to explicitly test against hpux this diff is small and quite reasonable. Once the pep8 issue is resolved I'm happy to merge it. Thanks for contributing!

src/_cffi_src/openssl/crypto.py
```diff
@@ -143,6 +143,7 @@
 }
 
 void Cryptography_free_wrapper(void *ptr, const char *path, int line) {
-    return free(ptr);
+    free(ptr);
+    return;
```
alex: You can just drop the `return` line, `void` functions don't need to return.

dumol: Thanks, did just that.

src/_cffi_src/build_openssl.py
```diff
@@ -46,7 +46,7 @@ def _extra_compile_args(platform):
     When we drop support for CRYPTOGRAPHY_OPENSSL_LESS_THAN_110 we can
     revisit this.
     """
-    if platform != "win32":
+    if platform not in [ "win32", "hp-ux11" ]:
```
reaperhulk: This will still fail pep8 because it doesn't want a space after `[` or before `]`.

dumol: Ooppss…

PR #4261, opened 2018-05-24
deprecate pythons without hmac.compare_digest

### LINKED ISSUE #4258: Drop Wheezy Builder?
Wheezy is EOL as of May 31, 2018 (https://wiki.debian.org/LTS). Should we drop it out of our test matrix going forward? We still have OpenSSL 1.0.1 coverage through building on Ubuntu 14.04 (supported through April 2019).

reaperhulk:
fixes #4258

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,9 @@ Changelog
 
 * Added :meth:`~cryptography.fernet.Fernet.extract_timestamp` to get the
   authenticated timestamp of a :doc:`Fernet </fernet>` token.
+* Support for Python 2.7.x without ``hmac.compare_digest`` has been deprecated.
+  We will require Python 2.7.7 or higher (or 2.7.6 on Ubuntu) in the next
+  ``cryptography`` release.
 
 .. _v2-2-2:
 
--- a/src/cryptography/hazmat/primitives/constant_time.py
+++ b/src/cryptography/hazmat/primitives/constant_time.py
@@ -5,7 +5,9 @@
 from __future__ import absolute_import, division, print_function
 
 import hmac
+import warnings
 
+from cryptography import utils
 from cryptography.hazmat.bindings._constant_time import lib
 
 
@@ -17,6 +19,12 @@ if hasattr(hmac, "compare_digest"):
         return hmac.compare_digest(a, b)
 
 else:
+    warnings.warn(
+        "Support for your Python version is deprecated. The next version of "
+        "cryptography will remove support. Please upgrade to a 2.7.x "
+        "release that supports hmac.compare_digest as soon as possible.",
+        utils.DeprecatedIn23,
+    )
     def bytes_eq(a, b):
         if not isinstance(a, bytes) or not isinstance(b, bytes):
             raise TypeError("a and b must be bytes.")
--- a/src/cryptography/hazmat/primitives/constant_time.py
+++ b/src/cryptography/hazmat/primitives/constant_time.py
@@ -25,6 +25,7 @@ else:
         "release that supports hmac.compare_digest as soon as possible.",
         utils.DeprecatedIn23,
     )
+
     def bytes_eq(a, b):
         if not isinstance(a, bytes) or not isinstance(b, bytes):
             raise TypeError("a and b must be bytes.")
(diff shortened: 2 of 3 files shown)


alex:
flake8

bmw:
It doesn't look like the version of Python 2 in Ubuntu Trusty has `hmac.compare_digest`. Do you plan to keep (security) support for the version of Python there until the OS is EOL'd next year?

alex:
That's pretty upsetting... yeah, I guess we'll probably wait. I assume trusty is a significant portion of your users?

bmw:
Yeah. It's about 10% of our users so waiting until April of next year would help us a lot.

On a related note, we're currently rethinking how we distribute Certbot. If you're interested, I'm happy to write up the approaches we're considering to see if we're on the same page for how long we'd need/want to support different versions of Python, OpenSSL, etc.

alex:
Yup, for 10% of your users we'll wait.

And yeah, since you're probably our most significant downstream, and we're your finickiest upstream I think it'd make sense for @reaperhulk and I to at least be aware of what you're thinking.

reaperhulk:
Yeah please don't hesitate to reach out @bmw. In this case I thought (completely wrongly and I'm not sure how I came to this conclusion) that Ubuntu had backported `compare_digest` into their Trusty 2.7.5.

Given the volume of Trusty users we have even outside of certbot we'll need to maintain support for the foreseeable future.

bmw:
We may also be your downstream that tries to run on the most ancient OSes.

Thanks for keeping Trusty support. It avoids us having to write code specifically for the platform to transition them to Python 3.

As for our new packaging plan, we haven't decided exactly what we're doing yet, but I can tell you how what we're thinking would mean for what we'd ideally like to see from `cryptography`. If you want to get into the weeds, I followed the suggestion from someone at PyCon and shared it in detail with [a post](https://mail.python.org/mm3/archives/list/distutils-sig@python.org/thread/QMTQCDVFSBABR55EFCMMHNSORXYG5NPE/#QMTQCDVFSBABR55EFCMMHNSORXYG5NPE) to the distutils-sig mailing list.

This may change, but the relevant highlights of what we're thinking right now are:

* Only support Python 3. We probably will drop Python 2 support in the next year.
* We will continue to use version of Python 3 provided by the OS package manager except on RHEL based distros. For us, this mainly means we'd want to keep support for the version of Python 3 in each version of Debian and Ubuntu until the distro release is EOL'd.
* We will continue to pull our Python dependencies from PyPI so any new Python dependencies (as long as they don't have non-Python dependencies) should be no problem at all.
* Thanks to your Linux wheels, OpenSSL support is largely a nonissue, however, about 3% of our users are on ARM. This is mainly Raspbian, so keeping support for the version of OpenSSL in Debian's releases would be nice.

Related to that last note though, Certbot currently shells out to `openssl` as a workaround for https://github.com/pyca/pyopenssl/issues/306. Any interest in getting support for that feature added to `cryptography` or `pyopenssl`? This would allow us to remove our dependency on `openssl`.

reaperhulk:
@bmw I'd be interested in adding OCSP support to cryptography, yes. Does the API need to validate the signatures on the responses or just parse the data into a form you can use to validate it yourself?

bmw:
Awesome. Thanks @reaperhulk.

I think we'd prefer the former, but the latter is all we really need.

bmw:
One aspect of the OCSP feature we would like to see is the ability to get the raw OCSP response. We need this for https://github.com/certbot/certbot/issues/956 which one of my coworkers has almost finished implementing.

reaperhulk:
@bmw Does this sound like roughly what you'd expect?

Certbot constructs an OCSP request with cryptography's APIs, serializes it, and then makes an http request with whatever library is desirable (be it async or otherwise). The response is a raw DER encoded OCSPResponse, which can be passed to cryptography for parsing as well as written to disk (after being verified that it is valid) to provide as a stapled response.

orbli:
passing by, thumbing up for OCSP support through cryptography api

bmw:
I just noticed that OCSP support was added in 2.4. Thank you so much for adding this.

reaperhulk:
@bmw you're welcome, sorry about forgetting to tell the certbot team that we had released it!

bmw:
Not a problem! Being one of our more important dependencies, I should probably subscribe to updates about the work you all are doing anyway.

PR #4280, opened 2018-06-13
add downstream tests for awslabs/aws-dynamodb-encryption-python

mattsb42-aws:
Add downstream tests for https://github.com/awslabs/aws-dynamodb-encryption-python

### DIFF
--- a/.travis/run.sh
+++ b/.travis/run.sh
@@ -54,6 +54,13 @@ else
             pip install -e .
             pytest -m local -l
             ;;
+        dynamodb-encryption-sdk)
+            git clone --depth=1 https://github.com/awslabs/aws-dynamodb-encryption-python
+            cd aws-dynamodb-encryption-python
+            pip install -r test/requirements.txt
+            pip install -e .
+            pytest -m "local and not slow and not veryslow and not nope" -l
+            ;;
         certbot)
             git clone --depth=1 https://github.com/certbot/certbot
             cd certbot
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,6 +63,8 @@ matrix:
         - python: 2.7
           env: DOWNSTREAM=aws-encryption-sdk
         - python: 2.7
+          env: DOWNSTREAM=dynamodb-encryption-sdk
+        - python: 2.7
           env: DOWNSTREAM=certbot OPENSSL=1.1.0h
         - python: 2.7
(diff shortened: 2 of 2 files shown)


.travis/run.sh
```diff
@@ -54,6 +54,13 @@ else
             pip install -e .
             pytest -m local -l
             ;;
+        dynamodb-encryption-sdk)
+            git clone --depth=1 https://github.com/awslabs/aws-dynamodb-encryption-python
+            cd aws-dynamodb-encryption-python
+            pip install -r test/requirements.txt
+            pip install -e .
+            pytest -m "local and not slow and not veryslow and not nope" -l
```
alex: I lol'd

alex:
The tests seem to be failing 👀

mattsb42-aws:
Yeah, was just digging through the logs (side note, might be worth revisiting the whitelisted branches? makes it harder to run the CI prior to a PR). Looks like it's choking on RSA-OAEP-SHA256, which requires OpenSSL >= 1.0.2.

alex:
Oh gosh, it makes me so happy that our new downstream test harness makes it easy to swap OpenSSLs

.travis/run.sh
mattsb42-aws: Haha, yup. Of the many levels of hypothesis-powered combinatorically exploding test scenarios, I figure we want the smallest set here.

reaperhulk:
@mattsb42-aws for future reference we're totally okay with iterating a bit in a PR. A few build failures is fine. 😄 We have more parallel builders than your fork too so the jobs will run faster anyway.

mattsb42-aws:
@reaperhulk Fair enough. I was under the impression that Travis ran fork CI in the context of the fork owner, so the builds for a fork wouldn't affect the main repo CI.

alex:
@mattsb42-aws I cannot begin to tell you how much that wouldn't bother us :-) FWIW, we're not going to remove the branch restriction because otherwise every time I send a PR via the github edit-in-the-web UI, the tests get run twice and it makes @reaperhulk sad.

mattsb42-aws:
@alex Haha, fair enough. I guess I could always just make an ephemeral branch on my fork that adds itself to the whitelist and gets constantly rebased on top of any branch I'm actually working on. Would be nice if the whitelist controls could be set per fork as well as per branch.

PR #4288, opened 2018-06-19
Normalize signatures to low S representation

webmaster128:
This is a private PR used to store a diff. No plans to get that merged anywhere


webmaster128:
Wrong target repository. Sorry everyone!

PR #4289, opened 2018-06-19
Add clearer message when key type is not bytes

### LINKED ISSUE #4157: Raise clearer error message when passing a unicode value as a symmetric key
Python: 2.7.3
cryptography==2.1.4
cffi==1.11.5
pip 9.0.1

```
import os

iv = os.urandom(16)
cipher = Cipher(algorithms.AES(key), modes.CFB8(iv),
                            backend=default_backend())
decryptor = cipher.decryptor()
raw_text = decryptor.update(encoded_text) + decryptor.finalize()
```

If `key` is unicode it will raise error.
```

/env/local/lib/python2.7/site-packages/cryptography/hazmat/primitives/ciphers/base.pyc in decryptor(self)
    125     def decryptor(self):
    126         ctx = self._backend.create_symmetric_decryption_ctx(
--> 127             self.algorithm, self.mode
    128         )
    129         return self._wrap_ctx(ctx, encrypt=False)

/env/local/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/backend.pyc in create_symmetric_decryption_ctx(self, cipher, mode)
    270 
    271     def create_symmetric_decryption_ctx(self, cipher, mode):
--> 272         return _CipherContext(self, cipher, mode, _CipherContext._DECRYPT)
    273 
    274     def pbkdf2_hmac_supported(self, algorithm):

/env/local/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/ciphers.pyc in __init__(self, backend, cipher, mode, operation)
    108             cipher.key,
    109             iv_nonce,
--> 110             operation
    111         )
    112         self._backend.openssl_assert(res != 0)
``` 

I have this fix temporary
```key = key.encode('utf-8') if isinstance(key, unicode) else key```
But should we do thi in `backends/openssl/ciphers.py` ?

### PR DESCRIPTION
**Pull request by VMois**
fixes #4157 

Should I add more tests, or add check key type in other modules (RSA, DSA etc.)?

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/ciphers.py
+++ b/src/cryptography/hazmat/backends/openssl/ciphers.py
@@ -55,6 +55,8 @@ class _CipherContext(object):
                 _Reasons.UNSUPPORTED_CIPHER
             )
 
+        utils._check_bytes("{0} key".format(cipher.name), cipher.key)
+
         if isinstance(mode, modes.ModeWithInitializationVector):
             iv_nonce = mode.initialization_vector
         elif isinstance(mode, modes.ModeWithTweak):
--- a/src/cryptography/hazmat/backends/openssl/ciphers.py
+++ b/src/cryptography/hazmat/backends/openssl/ciphers.py
@@ -55,8 +55,6 @@ class _CipherContext(object):
                 _Reasons.UNSUPPORTED_CIPHER
             )
 
-        utils._check_bytes("{0} key".format(cipher.name), cipher.key)
-
         if isinstance(mode, modes.ModeWithInitializationVector):
             iv_nonce = mode.initialization_vector
         elif isinstance(mode, modes.ModeWithTweak):
--- a/src/cryptography/hazmat/primitives/ciphers/algorithms.py
+++ b/src/cryptography/hazmat/primitives/ciphers/algorithms.py
@@ -12,6 +12,9 @@ from cryptography.hazmat.primitives.ciphers.modes import ModeWithNonce
 
 
 def _verify_key_size(algorithm, key):
+    # Verify that the key is instance of bytes
+    utils._check_bytes("{0} key".format(algorithm.name), key)
+
     # Verify that the key size matches the expected key size
     if len(key) * 8 not in algorithm.key_sizes:
         raise ValueError("Invalid key size ({0}) for {1}.".format(
--- a/src/cryptography/hazmat/primitives/ciphers/algorithms.py
+++ b/src/cryptography/hazmat/primitives/ciphers/algorithms.py
@@ -13,7 +13,7 @@ from cryptography.hazmat.primitives.ciphers.modes import ModeWithNonce
 
 def _verify_key_size(algorithm, key):
     # Verify that the key is instance of bytes
-    utils._check_bytes("Key", key)
+    utils._check_bytes("key", key)
 
     # Verify that the key size matches the expected key size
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: src/cryptography/hazmat/backends/openssl/ciphers.py [hunk 1]
```diff
@@ -55,6 +55,8 @@ def __init__(self, backend, cipher, mode, operation):
                 _Reasons.UNSUPPORTED_CIPHER
             )
 
+        utils._check_bytes("{0} key".format(cipher.name), cipher.key)
```
alex: I think it'd make more sense for this code to live in `algorithms.py`, specifically in `_verify_key_size`, which is already used by all the algorithm classes, and which will provide an earlier error message

VMois: You're right. I will change it. What about tests?

alex: They should be changed to verify the `AES` constructor errors on non-bytes arguments

VMois:
When I'm using key of string type it works without an error (code example from related issue). Should it be allowed?

src/cryptography/hazmat/primitives/ciphers/algorithms.py
```diff
@@ -12,6 +12,9 @@
 
 
 def _verify_key_size(algorithm, key):
+    # Verify that the key is instance of bytes
+    utils._check_bytes("{0} key".format(algorithm.name), key)
```
alex: Let's drop the fancier error message and just use `key`, it'll be a small performance win, and it's consistent with all the other places we raise this error, and it should be obvious what key it is from context.

alex:
No. Strings are not a valid key type, it's a bug that they'd ever be allowed. Keys must be bytes.

src/cryptography/hazmat/primitives/ciphers/algorithms.py
VMois: Okay. Just "key" string.

VMois:
I will change `_check_bytes` function, so when string will be passed it will also throw an error. Okey?

alex:
What change do you think `_check_bytes` needs? It looks complete to me.

VMois:
> No. Strings are not a valid key type, it's a bug that they'd ever be allowed. Keys must be bytes.

So, we accept string as key but it is not a good practice?

A lot of tests of python3 failed in travis. Could you look, please.

alex:
Looks like this patch caught a bunch of legitimate bugs in our code.

VMois:
I would say I'm lucky :) 

Some keys, for example, from 3DES tests are of string type but should be in bytes. But you have already understood this.

alex:
It should be possible to merge/rebase this on master and get the tests passing.

VMois:
One job is failed https://travis-ci.org/pyca/cryptography/jobs/394446823 due to "No output received in 10m". Could you run travis test suit one more time.

src/cryptography/hazmat/primitives/ciphers/algorithms.py
```diff
@@ -12,6 +12,9 @@
 
 
 def _verify_key_size(algorithm, key):
+    # Verify that the key is instance of bytes
+    utils._check_bytes("Key", key)
```
alex: nit: lowercase `key` please

VMois: Done.

PR #4292, opened 2018-06-20
Exposed OpenSSL prime methods

### LINKED ISSUE #4287: [Feature request] expose BIGNUM prime methods
For our research (@Tribler) we use OpenSSL's `BN_generate_prime_ex` and `BN_is_prime_ex` through custom bindings. Is this something you would want to support in your bindings (hazmat backend)?

If you're interested, I can turn https://github.com/qstokkink/cryptography/commit/31d2bc21e60c6842be2927aa576834b3cf2bf6a4 into a polished PR.

qstokkink:
Resolves #4287 

This PR adds support for the following OpenSSL calls:

 - Function `BN_generate_prime_ex`, for generating (safe) primes
 - Function `BN_is_prime_ex`, for checking primality
 - Macro `BN_prime_checks_for_size`, giving the recommended amount of rounds for primality checking

### DIFF
--- a/src/_cffi_src/openssl/bignum.py
+++ b/src/_cffi_src/openssl/bignum.py
@@ -72,6 +72,12 @@ BIGNUM *BN_mod_inverse(BIGNUM *, const BIGNUM *, const BIGNUM *, BN_CTX *);
 int BN_num_bytes(const BIGNUM *);
 
 int BN_mod(BIGNUM *, const BIGNUM *, const BIGNUM *, BN_CTX *);
+
+/* The following 3 prime methods are exposed for Tribler. */
+int BN_generate_prime_ex(BIGNUM *, int, int, const BIGNUM *,
+                         const BIGNUM *, BN_GENCB *);
+int BN_is_prime_ex(const BIGNUM *, int, BN_CTX *, BN_GENCB *);
+const int BN_prime_checks_for_size(int);
 """
 
 CUSTOMIZATIONS = """


tests/hazmat/backends/test_openssl.py
```diff
@@ -169,6 +169,52 @@ def test_bn_to_int(self):
         bn = backend._int_to_bn(0)
         assert backend._bn_to_int(bn) == 0
 
+    def test_bn_generate_prime(self):
```
reaperhulk: We don't add tests for our bindings only so you'll probably want to pull these tests into your own codebase.

qstokkink:
@reaperhulk resolved.

src/_cffi_src/openssl/bignum.py
```diff
@@ -72,6 +72,12 @@
 int BN_num_bytes(const BIGNUM *);
 
 int BN_mod(BIGNUM *, const BIGNUM *, const BIGNUM *, BN_CTX *);
+
+/* The following 3 prime methods are exposed for Tribler. */
+int BN_generate_prime_ex(BIGNUM *, int, int, const BIGNUM *,
+                         const BIGNUM *, BN_GENCB *);
+int BN_is_prime_ex(const BIGNUM *, int, BN_CTX *, BN_GENCB *);
+static const int BN_prime_checks_for_size(int);
```
reaperhulk: Any reason you chose to declare this as static?

qstokkink: That's what the documentation (https://cryptography.io/en/latest/development/c-bindings/) gave as the format for `#define`'s. I'll change it to non-static then.

reaperhulk: Ah, yeah. The static keyword doesn't hurt here so it's fine, I was just curious why. Probably should update our docs to note that static makes sense for constants and not for macros though.

PR #4294, opened 2018-06-21
Add OID for RSASSA-PSS X.509 signature algorithm (RFC 4056)

intgr:
In 2005, IETF devised a more secure padding scheme to replace PKCS #1 v1.5. To make sure that nobody can properly support or use it, they mandated lots of complicated parameters in the certificate, unlike any other X.509 signature scheme.

https://tools.ietf.org/html/rfc4056

`_SIG_OIDS_TO_HASH` and `Certificate.signature_hash_algorithm` cannot be supported as-is, because the hash algorithm is defined in the signature algorithm parameters, not by the OID itself.

I wasn't quite sure what to use as the OID "name", OpenSSL calls it `Signature Algorithm: rsassaPss` which is ugly.

RFC 4056 defines `id-RSASSA-PSS OBJECT IDENTIFIER ::= { pkcs-1 10 }` so I opted for `RSASSA-PSS` as the name. This is in line with ECDSA/DSA, which also use the identifier name from RFC.


reaperhulk:
Naming is always tricky, but the RFC is a good choice. Could you add it to the docs (`x509/reference.rst`) with a `.. versionadded:: 2.3`?

intgr:
Added to documentation.

alex:
That issue was resolved; it was #4297.

PR #4304, opened 2018-07-01
Use RFC 4514 Distinguished Name formatting in Name str() and repr()

intgr:
The current output for str()/repr() is very verbose and ugly. Thankfully RFC 4514 defines a standard string representaton.

For now, I have no plans to implement RFC 4514 for input, but it would be a very welcome change as well.

Alternatively instead of str()/repr(), the classes could have a `distinguished_name()` method, but I think this is more helpful. Once I have an agreement that this is a good way to go, I will update docs as well.


reaperhulk:
If we're going to make this change then I definitely would want a method call that returns this string that `repr` uses. We don't consider `repr` to be part of the public contract so just changing that doesn't guarantee we won't alter it in the future. I don't have a strong opinion on the repr alteration but I'm happy to merge a method that returns this string serialization. Are there any additional test vectors we can use to verify conformance with the presumably annoying edge cases around escaping?

Preserving order on RDNs is definitely desirable, but should be submitted as a separate PR please.

Thanks for your contribution so far!

intgr:
> If we're going to make this change then I definitely would want a method call that returns this string that repr uses

Yeah, I currently implemented it in the `__str__` and then made `__repr__` use that. Is that what you had in mind or do you think there should be a 3rd method?

intgr:
> Preserving order on RDNs is definitely desirable, but should be submitted as a separate PR please.

Done: #4306. I deliberately kept it as a separate commit because I already expected that.

reaperhulk:
This can be rebased now. Let's go down the route of `distinguished_name` (or similar) name rather than `__str__` since the cryptography API doesn't use `str()` anywhere else.

reaperhulk:
@intgr still interested in working on this?

intgr:
@reaperhulk Hrm, I am actually, I just need to get my procrastination under control... :)

intgr:
Added some documentation and rebased.

intgr:
On a second thought, I don't like the `distinguished_name` method name much. It's long and also the `Name` object itself could just as well be called a "distinguished name".

What I could come up with were `as_string`, `as_text`, hrm. Although I think the most Pythonic way would be `__str__`. I changed it to `as_string` for now.

reaperhulk:
`rfc_4514_string`?

intgr:
Sorry, `rfc_4514_string` seems really ugly to me. Do you think it's an improvement over `as_string`? :)

reaperhulk:
`as_string` doesn't describe what kind of string it is and is inconsistent with our current API (which strongly favors an explicit and easily auditable approach). I agree that it's not a particularly pretty name, but it is descriptive. This is a form of serialization so you *could* argue it should be an enum passed to a `public_bytes` method, but I think that's probably a step too far. @alex do you have an opinion?

intgr:
OK since we don't have a 3rd opinion, I rebased this and renamed the method to `rfc4514_string` (without the `_` after `rfc` -- this way is more consistent with other usages in cryptography code).

Looks like I managed to appease the coverage checks too this time around :)

PR #4305, opened 2018-07-02
Add testing support of Python 3.7

ericwb:
So pyca/cryptography's Python classifier already claims support
of Python 3.7, but there is no testing matrix for it. This patch
adds the recently released Python 3.7 to the matrix of testing.
It requires sudo:true and xenial to pass.

Fixes Issue #4301

Signed-off-by: Eric Brown <browne@vmware.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,10 @@ matrix:
           env: TOXENV=py35
         - python: 3.6
           env: TOXENV=py36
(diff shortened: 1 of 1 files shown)


alex:
@reaperhulk do you have opinions on doing this vs. waiting for "native" 3.7 support from travis?

PR #4306, opened 2018-07-04
Make RelativeDistinguishedName preserve attribtue order

intgr:
Duplicate attributes now raise an error instead of silently discarding
duplicates.

From #4304:
> Preserving order on RDNs is definitely desirable, but should be submitted as a separate PR please.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,9 @@ Changelog
   ``cryptography`` release.
 * Fixed multiple issues preventing ``cryptography`` from compiling against
   LibreSSL 2.7.x.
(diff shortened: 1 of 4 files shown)


reaperhulk:
The docs job is failing here.

As a side note: comparisons of DNs are supposed to ignore RDN ordering, but I'm okay with providing strong ordering just for string stability.

intgr:
> The docs job is failing here.

Heh, I noticed today, just a little before you sent that message. Hopefully pushed a correct version now.

> comparisons of DNs are supposed to ignore RDN ordering

Yep, I've implemented it as an order-preserving set. Comparisons are still based on `frozenset`, but iteration uses a `list`.

PR #4307, opened 2018-07-04
since the generator order is 570 bits this should be 570

### LINKED ISSUE #4271: incorrect key_size of sect571r1
Hello!

https://github.com/pyca/cryptography/blob/17c8f126c7c7d5ce886112a6e924277a7b203f25/src/cryptography/hazmat/primitives/asymmetric/ec.py#L138

The value there should be 570. From [the standard](http://www.secg.org/sec2-v2.pdf) the order of the published generator is 

```py
>>> order = 0x03FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFE661CE18FF55987308059B186823851EC7DD9CA1161DE93D5174D66E8382E9BB2FE84E47
>>> print(len(bin(order))-2)
570
```

reaperhulk:
but key_size is nonsense and we'll deprecate it next

fixes #4271

### DIFF
--- a/src/cryptography/hazmat/primitives/asymmetric/ec.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/ec.py
@@ -135,7 +135,7 @@ EllipticCurvePublicKeyWithSerialization = EllipticCurvePublicKey
 @utils.register_interface(EllipticCurve)
 class SECT571R1(object):
     name = "sect571r1"
-    key_size = 571
+    key_size = 570
 
 
 @utils.register_interface(EllipticCurve)


alex:
Is it possible to write a test for this? Possibly using `from_encoded_point`, which I believe looks at `key_size`?

reaperhulk:
Is there any point given that we're going to deprecate this right after and never touch any of it again until we remove it entirely?

reaperhulk:
I did confirm that the bit size of the order of the generator is indeed 570 which means the "bit length of the secret scalar" (as we documented `key_size` to be) will be <= 570 in all cases.

Issue #3331, opened 2016-12-19
Integrate tests from wycheproof


alex:
https://github.com/google/wycheproof:

- [x] `aes_cbc_pkcs5_test.json`
- [x] `aes_cmac_test.json`
- [x] `aes_gcm_test.json`
- [x] `chacha20_poly1305_test.json`
- [x] `dsa_test.json`
- [x] `ecdh*_test.json`
- [x] `ecdsa*_test.json`
- [x] `kw_test.json`
- [x] `kwp_test.json`
- [x] `rsa_signature*_test.json`
- [x] `x25519_test.json`

reaperhulk:
They've published test vectors we can actually use now: https://github.com/google/wycheproof/tree/master/testvectors

alex:
For anyone interested in working on this, I think we should structure this so CI clones that repo and runs them -- do not copy them all in.

mithrandi:
Debian packages are built without network access, so there would need to be a way to disable the downloading. @alex's suggestion seems reasonable since I could ship the vectors in the source package.

reaperhulk:
When we add the KWP vectors we should test to see if it would catch the 0 mod 8 bug we had.

PR #4310, opened 2018-07-05
Refs #3331 -- added initial wycheproof integration, starting with x25519, rsa, and keywrap

alex:


alex:
@reaperhulk so what do you make of: https://travis-ci.org/pyca/cryptography/jobs/400673652#L680

alex:
@reaperhulk I added RSA, and it found a case where we hit an assertion error...

reaperhulk:
When we add the KWP vectors we should test to see if it would catch the 0 mod 8 bug we had.

reaperhulk:
The X25519 tests currently have 33 failures:

* 28 are InternalError. We should decide how to handle those InternalErrors, but they are all "small public key" or a public key of low order (both with flag and without).

The final 5 are more interesting. They result in a completed exchange, but the value OpenSSL provides does not match the test vector. Their comments/flags are labeled "edge case for shared secret" (one of which also has the flag "Twist"), and "public key >= p". At first glance these appear to be bugs in the OpenSSL implementation. I'm going to write a minimal C reproducer next.

reaperhulk:
The plot thickens. This C code, written based on the way we do it in cryptography, does not exhibit the bug.

```c
#include <stdio.h>
#include <string.h>
#include <openssl/crypto.h>
#include <openssl/evp.h>
#include <openssl/x509.h>


int main(void) {
    OPENSSL_init();
    int res;
    /* This is TcId: 50 from the X25519 wycheproof tests but with the X25519 PKCS8 prefix */
    /* added to the private key */
    const unsigned char public[32] = {0xf0,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0x7f};
    unsigned char private[48] = {0x30, 0x2e, 0x02, 0x01, 0x00, 0x30, 0x05, 0x06, 0x03, 0x2b, 0x65, 0x6e, 0x04, 0x22, 0x04, 0x20, 0x28, 0x87, 0x96, 0xbc, 0x5a, 0xff, 0x4b, 0x81, 0xa3, 0x75, 0x01, 0x75, 0x7b, 0xc0, 0x75, 0x3a, 0x3c, 0x21, 0x96, 0x47, 0x90, 0xd3, 0x86, 0x99, 0x30, 0x8d, 0xeb, 0xc1, 0x7a, 0x6e, 0xaf, 0x8d};
    unsigned char shared[32] = {0xb4, 0xe0, 0xdd, 0x76, 0xda, 0x7b, 0x07, 0x17, 0x28, 0xb6, 0x1f, 0x85, 0x67, 0x71, 0xaa, 0x35, 0x6e, 0x57, 0xed, 0xa7, 0x8a, 0x5b, 0x16, 0x55, 0xcc, 0x38, 0x20, 0xfb, 0x5f, 0x85, 0x4c, 0x5c};
    printf("Loading public key\n");
    EVP_PKEY *pkey = EVP_PKEY_new();
    if (pkey == NULL) {
        printf("Error allocating pkey");
        exit(1);
    }
    printf("Setting public key type\n");
    res = EVP_PKEY_set_type(pkey, NID_X25519);
    if(!res) {
        printf("Error setting type\n");
        exit(1);
    }
    printf("Setting public key value\n");
    res = EVP_PKEY_set1_tls_encodedpoint(pkey, public, 32);
    if(!res) {
        printf("Error setting public key value\n");
        exit(1);
    }
    printf("Loading the private key\n");
    BIO *private_bio = BIO_new_mem_buf(private, 48);
    if (private_bio == NULL) {
        printf("Error creating BIO\n");
        exit(1);
    }
    printf("d2i_PrivateKey_bio\n");
    EVP_PKEY *skey = d2i_PrivateKey_bio(private_bio, NULL);
    if (skey == NULL) {
        printf("Error loading private key\n");
        exit(1);
    }
    printf("Allocate EVP_PKEY_CTX\n");
    EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new(skey, NULL);
    if (ctx == NULL) {
        printf("Error allocating EVP_PKEY_CTX\n");
        exit(1);
    }
    printf("EVP_PKEY_derive_init\n");
    res = EVP_PKEY_derive_init(ctx);
    if (res != 1) {
        printf("Error in EVP_PKEY_derive_init\n");
        exit(1);
    }
    printf("EVP_PKEY_set_peer\n");
    res = EVP_PKEY_derive_set_peer(ctx, pkey);
    if (res != 1) {
        printf("Error setting peer\n");
        exit(1);
    }
    size_t keylen;
    printf("EVP_PKEY_derive (returns length of shared key)\n");
    res = EVP_PKEY_derive(ctx, NULL, &keylen);
    if (res != 1) {
        printf("Error calling derive to get shared key size\n");
        exit(1);
    }
    if (keylen != 32) {
        printf("Key length should have been 32\n");
        exit(1);
    }
    printf("EVP_PKEY_derive\n");
    unsigned char buf[32] = {0};
    res = EVP_PKEY_derive(ctx, buf, &keylen);
    if (res != 1) {
        printf("Error calling derive to get shared key\n");
        exit(1);
    }
    res = memcmp(buf, shared, 32);
    if (res != 0) {
        printf("Shared key does not match expectation\n");
        printf("computed: %s\n", OPENSSL_buf2hexstr(buf, 32));
        printf("expected: %s\n", OPENSSL_buf2hexstr(shared, 32));
        exit(1);
    } else {
        printf("All is well\n");
        exit(0);
    }
}
```

alex:
Summarizing from IRC: that bug does in fact reproduce, on OpenSSL 1.1.1 (master): https://github.com/openssl/openssl/issues/6687

thaidn:
> When we add the KWP vectors we should test to see if it would catch the 0 mod 8 bug we had.

Which bug is it? We'd love to add the test vectors to Wycheproof if we don't have them yet.

reaperhulk:
@thaidn I haven't looked at the KWP vectors in wycheproof yet, but https://github.com/pyca/cryptography/pull/4160 was our bug fix and it was reported in https://github.com/pyca/cryptography/issues/4156.

At the time the NIST vectors didn't contain any cases where the input was 0 mod 8 (bytes, not bits), so the logic error was not caught. They have since been updated (after @alex emailed them) so that they would catch this case, but it would probably be useful for wycheproof to contain examples of that length as well if it doesn't already.

thaidn:
Thanks! We have some test vectors whose lengths are multiple of 8. Please let us know if they don't trigger the bug.

alex:
@reaperhulk can you handle making `X25519` raise an exception (not `InternalError`) on error

tests/wycheproof/test_rsa.py
```diff
@@ -0,0 +1,82 @@
...
+            return False
+        return True
+
+    return False
+
+
+@pytest.mark.requires_backend_interface(interface=RSABackend)
+@pytest.mark.supported(
+    only_if=lambda backend: (
+        not backend._lib.CRYPTOGRAPHY_OPENSSL_LESS_THAN_102
```
alex: Sigh, this is skipping on LibreSSL, isn't it?

alex: Yes, it is. These tests don't pass ATM on LibreSSL, but we'd like to run them, at least on a future patched version of Libre...

opinions @reaperhulk ?

thaidn: @alex do you have the test results somewhere online? I want to see what the failures are.

alex: @thaidn https://travis-ci.org/pyca/cryptography/jobs/402496876#L664; we've also independently reached out to the LibreSSL team about these.

(If you notice anything that strikes you as particularly severe, maybe let's move that conversation to email)

reaperhulk: (As discussed on IRC we've decided to skip on libre right now)

reaperhulk:
This ready for review @alex?

alex:
@reaperhulk I feel like length errors on the `wrapped_key` in `unwrap` should be `InvalidUnwrap`, instead of `ValueError` -- wdyt?

reaperhulk:
Agreed, I merged the PR you sent for that

reaperhulk:
@thaidn I've confirmed that the wycheproof tests easily catch our old 0 mod 8 bug (33 different test cases fail).

tests/wycheproof/test_rsa.py
```diff
@@ -0,0 +1,85 @@
...
+@pytest.mark.supported(
+    only_if=lambda backend: (
+        # TODO: this also skips on LibreSSL, which is ok for now, since these
+        # don't pass on Libre, but we'll need to fix this after they resolve
+        # it.
+        not backend._lib.CRYPTOGRAPHY_OPENSSL_LESS_THAN_102
+    ),
+    skip_message=(
+        "Many of these tests fail on OpenSSL < 1.0.2 and since upstream isn't"
+        " maintaing it, they'll never be fixed."
```
reaperhulk: Typo

PR #4319, opened 2018-07-09
document that an ECPublicNumbers object has some unexpected properties

### LINKED ISSUE #4273: Where the EC point-on-curve check happens is not clearly documented
Whenever one imports an EC public key somewhere, it is critical to check the point is on the curve. The most natural place where you take in the serialized form (or otherwise get raw coordinates somewhere). A caller will naturally expect that deserialization can fail. But cryptography.io does not do this. It has *three* representations of public keys: bytes, `EllipticCurvePublicNumbers`, and `EllipticCurvePublicKey`, with the middle appearing to be a forced intermediate point (so users have to pay for more allocations). `EllipticCurvePublicNumbers.from_encoded_point` does *not* perform this check. Rather, cryptography.io relies on the backend to do it in `EllipticCurvePublicNumbers.public_key`. But this function does not document that it can fail.

This should be changed to either make the failure cases match the documentation by making `EllipticCurvePublicNumbers` only describe an actual point, or document clearly that an `EllipticCurvePublicNumbers` has performed no validation yet. It is just a three tuple of a curve and two numbers x and y, not necessarily actual affine coordinates of a point on the curve. That means that it needs to be documented that the caller must handle errors in *both* `from_encoded_point` and `public_key` because either function may fail in normal execution.

PS: https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicNumbers seems to be missing documentation on the constructor

reaperhulk:
It is not guaranteed to be a valid point on the curve as that is not checked until you convert it to a PublicKey object.

fixes #4273

### DIFF
--- a/docs/hazmat/primitives/asymmetric/ec.rst
+++ b/docs/hazmat/primitives/asymmetric/ec.rst
@@ -154,7 +154,10 @@ Elliptic Curve Signature Algorithms
 
     .. warning::
         The point represented by this object is not validated in any way until
-        :meth:`EllipticCurvePublicNumbers.public_key` is called.
+        :meth:`EllipticCurvePublicNumbers.public_key` is called and may not
+        represent a valid point on the curve. You should not attempt to perform
+        any computations using the values from this class until you have either
+        validated it yourself or called ``public_key()`` successfully.
 
     .. versionadded:: 0.5
 
--- a/docs/hazmat/primitives/asymmetric/ec.rst
+++ b/docs/hazmat/primitives/asymmetric/ec.rst
@@ -152,6 +152,10 @@ Elliptic Curve Signature Algorithms
 
 .. class:: EllipticCurvePublicNumbers(x, y, curve)
 
+    .. warning::
+        The point represented by this object is not validated in any way until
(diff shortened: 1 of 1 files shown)


alex:
I think more importantly, the numbers documentation should say that the point might not be on the curve, if it was created from untrusted data.

alex:
I think I was unclear, I'd specifically like language at the class-level, not specific to this method, highlighting that the constructor does not perform validation, that you need to use `public_key()` to know if you have a point on the curve.

docs/hazmat/primitives/asymmetric/ec.rst
```diff
@@ -152,6 +152,10 @@ Elliptic Curve Signature Algorithms
 
 .. class:: EllipticCurvePublicNumbers(x, y, curve)
 
+    .. warning::
+        The point represented by this object is not validated in any way until
```
alex: "is not validated in any way until [...], it may not represent a valid point on the curve. You should not attempt to perform any computations using the values from this class, until you have either validated it yourself or called `public_key()` successfully"

WDYT?

PR #4323, opened 2018-07-09
Don't initialize locks in OpenSSL 1.1.0+

### LINKED ISSUE #4315: Don't initialize locks in OpenSSL 1.1.0+
https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/bindings/openssl/binding.py#L143

We currently check if `CRYPTO_get_locking_callback()` is `NULL` to know if we need to set up our own locks. However, on OpenSSL 1.1.0, it's _always_ `NULL`, because the system provides it's own locking callback.

We can save some memory and speed up initialization a smidge by skipping this logic if we're on a version of OpenSSL that provide it's own locking.

tuxxy:
### What this does:
1. Checks if `Cryptography_HAS_LOCKING_CALLBACKS` is true before compiling `_ssl_thread_locking_functions` in `callbacks.py`.
2. Exposes `CRYPTOGRAPHY_OPENSSL_LESS_THAN_110` via types.
3. Checks if `Cryptography_HAS_LOCKING_CALLBACKS` is true when attempting to call `init_static_locks`; returns if it's `False`.
4. Adds better checks to `test_crypto_lock_init` for OpenSSL versions > 1.1.0 and < 1.1.0.
5. Closes #4315.

### DIFF
--- a/src/_cffi_src/openssl/callbacks.py
+++ b/src/_cffi_src/openssl/callbacks.py
@@ -47,6 +47,7 @@ CUSTOMIZATIONS = """
    using CPython APIs by Armin Rigo of the PyPy project.
 */
 
+#if Cryptography_HAS_LOCKING_CALLBACKS
 #ifdef _WIN32
 typedef CRITICAL_SECTION Cryptography_mutex;
 static __inline void cryptography_mutex_init(Cryptography_mutex *mutex) {
@@ -78,7 +79,6 @@ static inline void cryptography_mutex_unlock(Cryptography_mutex *mutex) {
     ASSERT_STATUS(pthread_mutex_unlock(mutex));
 }
 #endif
-
 
 
 static unsigned int _ssl_locks_count = 0;
@@ -135,6 +135,9 @@ int Cryptography_setup_ssl_threads(void) {
     }
     return 1;
 }
+#else
+int (*Cryptography_setup_ssl_threads)(void) = NULL;
+#endif
 
 typedef struct {
     char *password;
--- a/src/_cffi_src/openssl/callbacks.py
+++ b/src/_cffi_src/openssl/callbacks.py
@@ -47,7 +47,7 @@ CUSTOMIZATIONS = """
    using CPython APIs by Armin Rigo of the PyPy project.
 */
 
-#if Cryptography_HAS_LOCKING_CALLBACKS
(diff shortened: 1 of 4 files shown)


tuxxy:
I did the requested changes, but I'm not 100% if I did the conditional compile correctly. Would love some input on that.

reaperhulk:
We're close, but LibreSSL is dying so we need to handle that case.

src/_cffi_src/openssl/callbacks.py
```diff
@@ -47,6 +47,7 @@
    using CPython APIs by Armin Rigo of the PyPy project.
 */
 
+#if Cryptography_HAS_LOCKING_CALLBACKS
```
reaperhulk: This value is not actually a preprocessor value so we can't use the preprocessor to check it. If you change this to `CRYPTOGRAPHY_OPENSSL_LESS_THAN_110` (which is what sets `Cryptography_HAS_LOCKING_CALLBACKS`) it will work properly.

src/_cffi_src/openssl/crypto.py
```diff
@@ -94,7 +94,7 @@
 # define OPENSSL_PLATFORM        SSLEAY_PLATFORM
 # define OPENSSL_DIR             SSLEAY_DIR
 #endif
-#if CRYPTOGRAPHY_OPENSSL_LESS_THAN_110
+#if CRYPTOGRAPHY_OPENSSL_LESS_THAN_110 || CRYPTOGRAPHY_IS_LIBRESSL
```
reaperhulk: `CRYPTOGRAPHY_OPENSSL_LESS_THAN_110` already does the `CRYPTOGRAPHY_IS_LIBRESSL` check so we can revert this.

tuxxy: Awesome, thanks for your help!

reaperhulk:
@tuxxy it appears codecov is being stupid, could you either rebase or push an empty commit so we can re-trigger CI? Thanks!

tuxxy:
Ah, sorry. Just saw this. I've been traveling around a lot and completely forgot. Thanks for handling it @reaperhulk, sorry for not noticing. :/

reaperhulk:
No worries, we're all set now.

PR #4331, opened 2018-07-12
add crl.get_revoked_certificate method

### LINKED ISSUE #4330: Add method to check if cert is in CRL?
Hi,

I am using `cryptography` to check certificates against CRLs, to see if they are revoked, and I find the current CRL implementation a little lacking. To check if a certificate is revoked, I have to retrieve all the `RevokedCertificate` from the `CertificateRevocationList` and see if any of them has the same serial as the certificate in question (if there's a better way, please let me know!). This is time consuming on big CRLs.

I have played around with `X509_CRL_get0_by_serial()` and `X509_CRL_get0_by_cert()` from OpenSSL, and have found them to drastically improve performance.

So, what do you think about adding a function to `CertificateRevocationList` that checks if a specified certificate is in the CRL? I'm thinking `is_certificate_revoked` or `get_revoked_certificate` which can return a `RevokedCertificate`.

Thanks for making cryptography!

reaperhulk:
fixes #4330

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -463,6 +463,15 @@ X.509 CRL (Certificate Revocation List) Object
             >>> crl.fingerprint(hashes.SHA256())
             b'e\xcf.\xc4:\x83?1\xdc\xf3\xfc\x95\xd7\xb3\x87\xb3\x8e\xf8\xb93!\x87\x07\x9d\x1b\xb4!\xb9\xe4W\xf4\x1f'
 
+    .. method:: get_revoked_certificate(serial_number)
+
+        .. versionadded:: 2.3
+
+        :param serial_number: The serial as a Python integer.
+        :returns: :class:`~cryptography.x509.RevokedCertificate` if the
+            ``serial_number`` is present in the CRL or ``None`` if it
+            is not.
+
     .. attribute:: signature_hash_algorithm
 
         :type: :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm`
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -463,7 +463,7 @@ X.509 CRL (Certificate Revocation List) Object
             >>> crl.fingerprint(hashes.SHA256())
             b'e\xcf.\xc4:\x83?1\xdc\xf3\xfc\x95\xd7\xb3\x87\xb3\x8e\xf8\xb93!\x87\x07\x9d\x1b\xb4!\xb9\xe4W\xf4\x1f'
 
-    .. method:: get_revoked_certificate(serial_number)
+    .. method:: get_revoked_certificate_by_serial_number(serial_number)
 
         .. versionadded:: 2.3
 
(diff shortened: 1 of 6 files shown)


docs/x509/reference.rst
```diff
@@ -463,6 +463,15 @@ X.509 CRL (Certificate Revocation List) Object
             >>> crl.fingerprint(hashes.SHA256())
             b'e\xcf.\xc4:\x83?1\xdc\xf3\xfc\x95\xd7\xb3\x87\xb3\x8e\xf8\xb93!\x87\x07\x9d\x1b\xb4!\xb9\xe4W\xf4\x1f'
 
+    .. method:: get_revoked_certificate(serial_number)
```
alex: I think this should be `get_revoked_certificate_by_serial_number`?

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -235,6 +238,22 @@ def fingerprint(self, algorithm):
         h.update(der)
         return h.finalize()
 
+    def get_revoked_certificate(self, serial_number):
+        revoked = self._backend._ffi.new("X509_REVOKED **")
+        asn1_int = _encode_asn1_int_gc(self._backend, serial_number)
+        res = self._backend._lib.X509_CRL_get0_by_serial(
```
alex: Does this do something more efficient than just a linear search?

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -235,6 +238,22 @@ def fingerprint(self, algorithm):
...
+            self._x509_crl, revoked, asn1_int
+        )
+        if res == 0:
+            return None
```
alex: Does this push an error? Do we need to clear the stack?

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: Implementation here: https://github.com/openssl/openssl/blob/01ad66f85d22fd001582b5f2e6e18db8b820c550/crypto/x509/x_crl.c#L381-L414

Short answer: not really. It sorts and then does a linear asn1 integer comparison. But no copies across ffi boundary.

reaperhulk: Ah, actually the stack find uses a binary search once it sorts the list. So it is faster than a linear search (in most cases)

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: No errors are pushed onto the stack in the test cases we have, but you'll never get me to say I'm 100% confident there's no scenario under which OpenSSL won't push an error onto it from some other function it calls.

docs/x509/reference.rst
reaperhulk: Yeah that's more consistent with `get_values_for_type`, `get_extension_for_oid`, `get_extension_for_class`. Changing.

PR #4332, opened 2018-07-12
raise valueerror for null x25519 derived keys

reaperhulk:
OpenSSL errors when it hits this edge case and a null shared key is bad anyway so let's raise an error.

We can bikeshed on language now. Also, do we want to document this exception? `:raises ValueError: If derived key is null`?

refs #4310

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/x25519.py
+++ b/src/cryptography/hazmat/backends/openssl/x25519.py
@@ -71,5 +71,9 @@ class _X25519PrivateKey(object):
         self._backend.openssl_assert(keylen[0] > 0)
         buf = self._backend._ffi.new("unsigned char[]", keylen[0])
         res = self._backend._lib.EVP_PKEY_derive(ctx, buf, keylen)
-        self._backend.openssl_assert(res == 1)
+        if res != 1:
+            raise ValueError(
+                "Null shared key derived from public/private pair."
+            )
+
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/backends/openssl/x25519.py
```diff
@@ -71,5 +71,9 @@ def exchange(self, peer_public_key):
         self._backend.openssl_assert(keylen[0] > 0)
         buf = self._backend._ffi.new("unsigned char[]", keylen[0])
         res = self._backend._lib.EVP_PKEY_derive(ctx, buf, keylen)
-        self._backend.openssl_assert(res == 1)
+        if res != 1:
+            raise ValueError(
+                "Null shared key derived from public/private pair."
```
alex: `s/null/all zero/`?

alex:
I think I'm ok not documenting it.

src/cryptography/hazmat/backends/openssl/x25519.py
reaperhulk: That doesn't really seem clearer to me (is it actually zeroes or zero bits?), but I don't feel strongly so I'll change it if you care more than I do :)

alex:
I can live with null

reaperhulk:
Okay, then I might push an empty commit because codecov appears very confused.

alex:
@reaperhulk oh shoot, do we need to pop the error stack here?

PR #4342, opened 2018-07-17
disallow implicit tag truncation with finalize_with_tag

reaperhulk:


src/cryptography/hazmat/backends/openssl/ciphers.py
```diff
@@ -190,7 +190,7 @@ def finalize(self):
         self._backend.openssl_assert(res == 1)
         return self._backend._ffi.buffer(buf)[:outlen[0]]
 
-    def finalize_with_tag(self, tag):
+    def finalize_with_tag(self, tag, min_tag_length=16):
```
alex: Is there a reason we can't just use the value from teh `GCM` constructor?

reaperhulk: I think it's pretty ugly to say you need to put the `min_tag_length` into the GCM constructor, but not the tag (which you may not have yet), and then later on enforce that tag length when calling `finalize_with_tag`.

On the other hand it's also confusing and potentially error prone that `min_tag_length` is on the GCM constructor and also here. In an ideal world I'd make the "pass the tag in the constructor" API go away entirely as I think `finalize_with_tag` is a much saner API, but I'm not allowed to have nice things.

alex: I'd favor having us expose the API in one place; and I think it's somewhat natural to include tag length with the "algorithm configuration"

PR #4349, opened 2018-07-17
add wycheproof gcm tests

reaperhulk:
refs #3331


tests/wycheproof/test_aes.py
```diff
@@ -40,3 +40,34 @@ def test_aes_cbc_pkcs5(backend, wycheproof):
         assert computed_ct != ct
         with pytest.raises(ValueError):
             unpadder.update(padded_msg) + unpadder.finalize()
+
+
+@pytest.mark.requires_backend_interface(interface=CipherBackend)
+@pytest.mark.wycheproof_tests("aes_gcm_test.json")
```
alex: Should we also have a copy of this test that runs against the `AESGCM` aead?

reaperhulk: Ah good idea.

Issue #4088, opened 2018-01-18
SSH public key format unclear


ericwb:
After reading through the documentation and looking at the  cryptography.hazmat.primitives.serialization.load_ssh_public_key(data, backend) API, it's unclear exactly what format of public key is expected/allowed.

The docs state they are OpenSSH public keys, but then refers to RFC 4253 (which section is unknown).  But OpenSSH does not serialize its public keys according to the SSH standard spec.  The actual spec for serializing SSH keys is RFC 4716.  OpenSSH has its own format documented here: http://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/src/usr.bin/ssh/PROTOCOL.certkeys?content-type=text/plain

So I think the docs just need some tweaking.  Easiest would be to remove mention of RFC 4253 and replace with PROTOCOL.certkeys.  

But it would also be very useful to support the official spec for persisting public keys (RFC 4716 ) and not be OpenSSH specific.

reaperhulk:
Thanks for the report -- updating the docs sounds like a good idea. We could also potentially support RFC 4716, but I'd like to know what actually uses it before accepting a PR.

ronf:
Multiple SSH products support RFC 4716 format, but I believe this format was initially used by the commercial SSH2 product from ssh.com.

If you are looking for RFC 4716 support in Python, my AsyncSSH package supports several public key formats including PKCS#1 DER and PEM, PKCS#8 DER and PEM, OpenSSH, and RFC 4716. AsyncSSH makes use of cryptography for all its low-level crypto, but it doesn't currently rely on cryptography for SSH key serialization, as AsyncSSH's implementation of that predates the one currently in cryptography.

PR #4361, opened 2018-07-19
point to openssh docs and don't talk about an rfc that isn't relevant

reaperhulk:
As noted in #4088 our docs are wrong about the RFC for OpenSSH's key format. It's not an RFC. Let's link to the actual docs.


alex:
the format really is documented in the RFC, the on disk version is just a base64 encoded version of the on the wire IIRC

PR #4365, opened 2018-07-23
parametrize a bunch of tests

reaperhulk:
Eventually we'll figure out how to deprecate numbers, but until then let's clean up the tests.

### DIFF
--- a/tests/hazmat/primitives/test_rsa.py
+++ b/tests/hazmat/primitives/test_rsa.py
@@ -1853,304 +1853,81 @@ class TestRSANumbers(object):
         with pytest.raises(TypeError):
             rsa.RSAPublicNumbers(e=1, n=None)
 
-    def test_private_numbers_invalid_types(self):
-        public_numbers = rsa.RSAPublicNumbers(e=1, n=15)
-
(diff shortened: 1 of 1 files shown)


reaperhulk:
I don't think I believe codecov here since it can't show me actual lines.

tests/hazmat/primitives/test_rsa.py
```diff
@@ -1835,304 +1835,81 @@ def test_public_numbers_invalid_types(self):
...
-                dmp1=1,
-                dmq1=None,
-                iqmp=2,
-                public_numbers=public_numbers
-            )
-
+    @pytest.mark.parametrize(
+        ("p", "q", "d", "dmp1", "dmq1", "iqmp", "public_numbers"),
+        [
+            (None, 5, 1, 1, 1, 2, rsa.RSAPublicNumbers(e=1, n=15)),
```
earonesty: would be nice to add comments explaining "why" like the other tests have....

reaperhulk: Whys would be good in general for these. Maybe in a future PR 😁

PR #4372, opened 2018-07-26
Test for expected CryptographyDeprecationWarnings

### LINKED ISSUE #4311: Fix signer/verifier warnings in our tests
We have a lot of tests that call `signer()` and `verifier()`, even though they are deprecated in favor of `sign()` and `verify()` we should fix these.

tipabu:
The remaining calls to `signer()` and `verifier()` are exercising the deprecated API intentionally. Let's test that the deprecation warnings are being raised as expected.

Closes #4311; see also #4314.

### DIFF
--- a/tests/hazmat/primitives/test_rsa.py
+++ b/tests/hazmat/primitives/test_rsa.py
@@ -541,7 +541,8 @@ class TestRSASignature(object):
     )
     def test_use_after_finalize(self, backend):
         private_key = RSA_KEY_512.private_key(backend)
-        signer = private_key.signer(padding.PKCS1v15(), hashes.SHA1())
+        with pytest.warns(CryptographyDeprecationWarning):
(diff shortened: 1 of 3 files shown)


alex:
Thanks for contributing!

Looks like there's a PEP8 error on travis; can you please take a look at that, and then we can review!

tests/hazmat/primitives/test_rsa.py
```diff
@@ -963,8 +965,9 @@ def test_signature_not_bytes(self, backend):
         signature = 1234
 
         with pytest.raises(TypeError):
-            public_key.verifier(
+            public_key.verify(
```
tipabu: Looks like I went overboard here and lost some coverage: https://codecov.io/gh/pyca/cryptography/compare/e722e557e08171e265008711b824df44d4e9ec7d...9238a6369b657305706426ee24a8bb64d4024065/src/src/cryptography/hazmat/backends/openssl/rsa.py

PR #4374, opened 2018-07-30
Use A-label strings in tests

tipabu:
We already have coverage of the U-label deprecation warnings in test_x509_ext.py; see [TestRFC822Name.test_idna](https://github.com/pyca/cryptography/blob/2.3/tests/x509/test_x509_ext.py#L1790-L1797) and [TestUniformResourceIdentifier.test_idna_no_port](https://github.com/pyca/cryptography/blob/2.3/tests/x509/test_x509_ext.py#L1829-L1835)


alex:
Thanks!

Issue #4375, opened 2018-07-31
Add new wycheproof vectors


alex:
- [x] `aes_ccm_test.json` (#4379)
- [x] `ecdh_secp224r1_ecpoint_test.json` (#4384)
- [x] `ecdh_secp256r1_ecpoint_test.json` (#4384)
- [x] `ecdh_secp384r1_ecpoint_test.json` (#4384)
- [x] `ecdh_secp521r1_ecpoint_test.json` (#4384)
- [x] `rsa_pss_2048_sha1_mgf1_20_test.json` (#4381)
- [x] `rsa_pss_2048_sha256_mgf1_0_test.json` (#4381)
- [x] `rsa_pss_2048_sha256_mgf1_32_test.json` (#4381)
- [x] `rsa_pss_3072_sha256_mgf1_32_test.json` (#4381)
- [x] `rsa_pss_4096_sha256_mgf1_32_test.json` (#4381)
- [x] `rsa_pss_4096_sha512_mgf1_32_test.json` (#4381)
- [x] `rsa_pss_misc_test.json` (#4381)

PR #4384, opened 2018-08-05
Refs #4375 -- added ECDH EC point wycheproof tests

alex:


alex:
Landing is blocked on #4383

PR #4410, opened 2018-08-22
Fix encoding errors in RSA test keys.

davidben:
enc-rsa-pkcs8.pem and unenc-rsa-pkcs8.pem did not encode the RSA key
correctly. Per [RFC 8017, appendix A.1](https://tools.ietf.org/html/rfc8017#appendix-A.1):

>   The object identifier rsaEncryption identifies RSA public and private
>   keys as defined in Appendices A.1.1 and A.1.2.  The parameters field
>   has associated with this OID in a value of type AlgorithmIdentifier
>   SHALL have a value of type NULL.
>
>      rsaEncryption    OBJECT IDENTIFIER ::= { pkcs-1 1 }

unenc-rsa-pkcs8.pem, however, was missing that `NULL`, which was, in turn,
carried into the encrypted payload of enc-rsa-pkcs8.pem. The DER
version, enc-rsa-pkcs8.der, carries this mistake too. Interestingly,
unenc-rsa-pkcs8.der does *not* have it. I'm guessing it was converted
with the openssl command-line tool which fixed the encoding in
conversion.

Current versions of OpenSSL are lax and ignore the parameters field, but
it's best to test against spec-compliant inputs. Fix unenc-rsa-pkcs8.pem
to match unenc-rsa-pkcs8.der and then refresh enc-rsa-pkcs8.{der,pem}
with the new encoding but otherwise the same encryption parameters.

I've refreshed the `dumpasn1` (at least that's what it looks like)
preamble at the top of each file, but the current version of `dumpasn1`
appears to have changed the spacing slightly, so there's some whitespace
diff noise.


reaperhulk:
We pulled these vectors from the gnutls test suite. I have no objection to changing them, but is it useful to retain an invalid encoding test? (e.g. are they common enough that we should test our tolerance or should we move forward into a properly encoded asn1 world). We'll also want to update the test-vectors.rst to note that these were generated rather than pulled from another library's test suite.

davidben:
BoringSSL doesn't accept such keys and hasn't had problems. Note that, although OpenSSL accepts this encoding, it doesn't generate it. Skimming their code, neither does GnuTLS.

davidben:
Sure thing. How's this?

PR #4411, opened 2018-08-23
stop using coverage with pypy

reaperhulk:
pypy is slow with coverage and we don't need it to hit everything


alex:
codecov's website shows full coverage, going to land even though it hasn't propagated that to github

PR #4416, opened 2018-08-26
Don't clone wycheproof if we're doing a downstream test

alex:


.travis/install.sh
```diff
@@ -41,7 +41,9 @@ elif [ -n "${LIBRESSL}" ]; then
     fi
 fi
 
-git clone --depth=1 https://github.com/google/wycheproof $HOME/wycheproof
+if [ -z "${UPSTREAM}" ]; then
```
reaperhulk: This should be downstream right?

PR #4418, opened 2018-08-29
reorganize downstream tests to avoid dependency squashing

mattsb42-aws:
Related issue: #4415

## Background/Problem

When a downstream test is performed, if the downstream repository defines loose requirements (generally the most common pattern) when it installs requirements, we create the opportunity for a scenario where these downstream tests are testing the downstream consumer against dependencies that the consumer has not necessarily themselves tested against. This can cause issues if a change in those dependencies breaks the downstream consumer in ways that are unrelated to changes in `pyca/cryptography`.

Given that the intention of these downstream tests is to verify that changes made in `pyca/cryptography` are not going to break that consumer, this complicates matters because they are not actually testing just that. The simple solution to this is for the downstream consumers to provide a frozen requirements file that describes "this is a known good set of dependencies, use those to run your tests". However, this causes problems with the way that the downstream consumer test setup was being done because a frozen requirements file applied *after* installing the `pyca/cryptography` dev build would by definition remove the same dev build that we are wanting to test against.

## Solution

What this change does is separate the downstream consumer install and test steps into two separate stages so that we can install the `pyca/cryptography` dev build *after* installing the consumer's dependencies.

## Other potential issues

This does leave open the possible issue of `pyca/cryptography` removing a dependency and that dependency sticking around and masking failures that might otherwise be caused, but given current `pip` tooling this is a much simpler solution than trying to extract a `pyca/cryptography` dependency tree from a set of frozen dependencies.

### DIFF
--- a/.travis/downstream.d/_HOW_TO_ADD_DOWNSTREAM_TESTS_
+++ b/.travis/downstream.d/_HOW_TO_ADD_DOWNSTREAM_TESTS_
@@ -1,0 +1,8 @@
+In order to add downstream tests to be run in CI, first add a test handler for
+the downstream consumer that you want to test. This should be a file in this
+".travis/downstream.d/" directory and follow the form defined in the
+"downstream-template.sh" template file. The new test handler file should be
+named in the format of {downstream name}.sh where {downstream name} is the
+name that you wish to use to identify the consumer. Next, add an entry to
+the test matrix in ".travis.yml" that sets the DOWNSTREAM environment
+variable to the downstream name that you selected.
--- a/.travis/downstream.d/aws-encryption-sdk.sh
+++ b/.travis/downstream.d/aws-encryption-sdk.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+case "${1}" in
+    install)
+        git clone --depth=1 https://github.com/awslabs/aws-encryption-sdk-python
+        cd aws-encryption-sdk-python
+        pip install -r test/requirements.txt
+        pip install -e .
+        ;;
+    run)
+        cd aws-encryption-sdk-python
+        pytest -m local test/
+        ;;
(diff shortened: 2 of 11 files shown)


.travis/downstream.d/_HOW_TO_ADD_DOWNSTREAM_TESTS_
```diff
@@ -0,0 +1,8 @@
+In order to add downstream tests to be run in CI, first add a test handler for
```
alex: Please rename this file to `README.rst`

.travis/downstream.d/aws-encryption-sdk.sh
```diff
@@ -0,0 +1,17 @@
...
+        cd aws-encryption-sdk-python
+        pip install -r test/requirements.txt
+        pip install -e .
+        ;;
+    run)
+        cd aws-encryption-sdk-python
+        pytest -m local test/
+        ;;
+    *)
+        exit
```
alex: These should all be `exit 1`, so it's an error to pass some other argument

.travis/downstream.d/downstream-template.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
alex: Please delete this, the template is simple enough to intuit from the examples

mattsb42-aws:
Changes made.

One thing that I was toying with was just making the test handler file name the same as the downstream name. That would simplify the instruction for adding them and open the option for non-bash in the future (without forcing them to have a misleading extension) if that were to be desirable. It does have the drawback of relying on the shebang and contents to ID them as shell scripts, though, rather than being explicit with the extension. I can't decide if that would be making the whole thing simpler or more complicated. :/

PR #4427, opened 2018-08-30
Fixes #3460 -- deprecate OpenSSL 1.0.1

### LINKED ISSUE #3460: Decide when to deprecate 1.0.1
At some point we should start warning about users with OpenSSL 1.0.1, as we did with 0.9.8 and 1.0.0, now that upstream is no longer supporting it. We don't have to make a decision about when to actually drop it, but at some point we should be warning on it.

However, I don't think we should do it right now, as it'd be too easy to just contribute to warning fatigue, 71% of our downloads are currently using 1.0.1. Here's the last 11 months of progress: 

screen shot 2017-03-18 at 1 30 00 pm

And the SQL I used to generate that:

```sql
SELECT
  STRFTIME_UTC_USEC(timestamp, "%Y-%m") AS yyyymm,
  ROUND(100 * SUM(CASE WHEN SUBSTR(REGEXP_EXTRACT(details.openssl_version, r"^OpenSSL ([^ ]+) "), 0, 5) = "0.9.8" THEN 1 ELSE 0 END) / COUNT(*), 1) AS percent_098,
  ROUND(100 * SUM(CASE WHEN SUBSTR(REGEXP_EXTRACT(details.openssl_version, r"^OpenSSL ([^ ]+) "), 0, 5) = "1.0.0" THEN 1 ELSE 0 END) / COUNT(*), 1) AS percent_100,
  ROUND(100 * SUM(CASE WHEN SUBSTR(REGEXP_EXTRACT(details.openssl_version, r"^OpenSSL ([^ ]+) "), 0, 5) = "1.0.1" THEN 1 ELSE 0 END) / COUNT(*), 1) AS percent_101,
  ROUND(100 * SUM(CASE WHEN SUBSTR(REGEXP_EXTRACT(details.openssl_version, r"^OpenSSL ([^ ]+) "), 0, 5) = "1.0.2" THEN 1 ELSE 0 END) / COUNT(*), 1) AS percent_102,
  ROUND(100 * SUM(CASE WHEN SUBSTR(REGEXP_EXTRACT(details.openssl_version, r"^OpenSSL ([^ ]+) "), 0, 5) = "1.1.0" THEN 1 ELSE 0 END) / COUNT(*), 1) AS percent_110,
  COUNT(*) as download_count
FROM
  TABLE_DATE_RANGE(
    [the-psf:pypi.downloads],
    DATE_ADD(CURRENT_TIMESTAMP(), -1, "year"),
    CURRENT_TIMESTAMP()
  )
WHERE
  details.openssl_version IS NOT NULL AND file.project = 'cryptography' and details.system.name NOT IN ("Windows", "Darwin")
GROUP BY
  yyyymm
ORDER BY
  yyyymm DESC
LIMIT 100
```

alex:


src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -150,9 +151,24 @@ def init_static_locks(cls):
...
 
 
+def _verify_openssl_version(lib):
+    if (
+        lib.CRYPTOGRAPHY_OPENSSL_LESS_THAN_102 and
+        not lib.CRYPTOGRAPHY_IS_LIBRESSL
+    ):
+        warnings.warn(
+            "OpenSSL version 1.0.1 is no longer supported by the OpenSSL "
+            "project, please upgrade. The a future version of cryptography "
```
reaperhulk: English

reaperhulk:
@stevepeak Hey Steve, we're seeing a lot of commits that appear to be endlessly stuck in "processing" (https://codecov.io/gh/pyca/cryptography/commits) including the latest commit on this PR. Is there anything we can do or is there something you can do on your end?

CHANGELOG.rst
```diff
@@ -8,6 +8,9 @@ Changelog
 
 .. note:: This version is not yet released and is under active development.
 
+* Deprecated OpenSSL 1.0.1 support. OpenSSL 1.0.1 is no longer supported by
+  the OpenSSL project. Please upgrade to a newer version of OpenSSL, or
+  install ``cryptography`` from a wheel.
```
reaperhulk: At this time there is no time table for dropping support, however we strongly encourage all users to upgrade or install ``cryptography`` from a wheel.

PR #4431, opened 2018-08-30
OCSP response abstract base classes

reaperhulk:
For reference: https://tools.ietf.org/html/rfc6960#section-4.2.1

### DIFF
--- a/src/cryptography/x509/ocsp.py
+++ b/src/cryptography/x509/ocsp.py
@@ -5,6 +5,7 @@
 from __future__ import absolute_import, division, print_function
 
 import abc
+from enum import Enum
 
 import six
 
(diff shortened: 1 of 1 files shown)


src/cryptography/x509/ocsp.py
```diff
@@ -107,3 +123,128 @@ def serial_number(self):
...
+        The signature bytes
+        """
+
+    @abc.abstractproperty
+    def certs(self):
+        """
+        List of certs that may be used to help verify a response.
+        """
+
+    @abc.abstractproperty
```
alex: remove this

src/cryptography/x509/ocsp.py
```diff
@@ -107,3 +123,128 @@ def serial_number(self):
...
+    def version(self):
+        """
+        The version
+        """
+
+    @abc.abstractproperty
+    def responder_id(self):
+        """
+        The responder's key hash or Name
```
alex: I don't like this

src/cryptography/x509/ocsp.py
```diff
@@ -107,3 +123,102 @@ def serial_number(self):
...
+        """
+        The date of when the certificate was revoked or None if not
+        revoked.
+        """
+
+    @abc.abstractproperty
+    def revocation_reason(self):
+        """
+        The reason the certificate was revoked or None if not specified or
+        notrevoked.
```
alex: space

PR #4435, opened 2018-08-30
add EC OIDs

reaperhulk:
refs #3345


docs/hazmat/primitives/asymmetric/ec.rst
```diff
@@ -501,6 +501,37 @@ All named curves are instances of :class:`EllipticCurve`.
 
 
 
+Elliptic Curve Object Identifiers
```
alex: Please move this to the bottom of the page

PR #4441, opened 2018-08-30
Lock aws-encryption-sdk and dynamodb-encryption-sdk downstream test dependencies

mattsb42-aws:
To avoid future cases of #4415, lock `aws-encryption-sdk` and `dynamodb-encryption-sdk` downstream tests to using the frozen dependency set that we prepare and test in each of those packages.


mattsb42-aws:
Travis errored out trying to start one of the unrelated tests. :(

PR #4449, opened 2018-09-01
OCSP bindings

reaperhulk:
Much of OCSP was opaqued in 1.1.0 so this also adds a bunch of getters for older OpenSSL. However, 1.1.0 itself made it impossible to access certain fields in the opaque struct, so we're forced to de-opaque them for 1.1.0 through 1.1.0i as well as 1.1.1-pre1 through 1.1.1-pre9. There is a patch (openssl/openssl#7082) that fixes this and should be in 1.1.0j and 1.1.1-pre10 (or 1.1.1 final, whichever they choose to issue)


src/_cffi_src/openssl/cryptography.py
```diff
@@ -44,13 +44,18 @@
     (OPENSSL_VERSION_NUMBER >= 0x10100000 && !CRYPTOGRAPHY_IS_LIBRESSL)
 #define CRYPTOGRAPHY_OPENSSL_110F_OR_GREATER \
     (OPENSSL_VERSION_NUMBER >= 0x1010006f && !CRYPTOGRAPHY_IS_LIBRESSL)
+#define CRYPTOGRAPHY_OPENSSL_BETWEEN_111_and_111PRE9 \
+    (OPENSSL_VERSION_NUMBER >= 0x10101000L  && \
```
alex: We don'y use the `L` suffix

src/_cffi_src/openssl/ocsp.py
```diff
@@ -62,7 +73,85 @@
...
 int i2d_OCSP_RESPONSE_bio(BIO *, OCSP_RESPONSE *);
+int i2d_OCSP_RESPDATA(OCSP_RESPDATA *, unsigned char **);
 """
 
 CUSTOMIZATIONS = """
+#if ( \
+    CRYPTOGRAPHY_OPENSSL_110_OR_GREATER && \
+    CRYPTOGRAPHY_OPENSSL_LESS_THAN_110J \
+    ) || CRYPTOGRAPHY_OPENSSL_BETWEEN_111_and_111PRE9
+struct ocsp_responder_id_st {
```
alex: A comment might be nice

PR #4452, opened 2018-09-01
support OCSP response parsing

reaperhulk:


src/cryptography/hazmat/backends/openssl/ocsp.py
```diff
@@ -4,13 +4,32 @@
...
 from cryptography.hazmat.primitives import serialization
-from cryptography.x509.ocsp import OCSPRequest, _OIDS_TO_HASH
+from cryptography.x509.ocsp import (
+    OCSPCertStatus, OCSPRequest, OCSPResponse, OCSPResponseStatus,
+    _CERT_STATUS_TO_ENUM, _OIDS_TO_HASH, _RESPONSE_STATUS_TO_ENUM,
+)
+
+
+def _response_check(func):
+    def wrapper(*args):
```
alex: Please write this as `(self, *args)` so the `response_status` check is easier to read

alex: Please use `functools.wraps` so that the metadata on `func` is correct

src/cryptography/hazmat/backends/openssl/ocsp.py
```diff
@@ -4,13 +4,32 @@
...
+from cryptography.hazmat.backends.openssl.x509 import _Certificate
 from cryptography.hazmat.primitives import serialization
-from cryptography.x509.ocsp import OCSPRequest, _OIDS_TO_HASH
+from cryptography.x509.ocsp import (
+    OCSPCertStatus, OCSPRequest, OCSPResponse, OCSPResponseStatus,
+    _CERT_STATUS_TO_ENUM, _OIDS_TO_HASH, _RESPONSE_STATUS_TO_ENUM,
+)
+
+
+def _response_check(func):
```
alex: Let's give this a better name `_requires_successful_response` or something

src/cryptography/hazmat/backends/openssl/ocsp.py
```diff
@@ -63,6 +82,225 @@ def _hash_algorithm(backend, cert_id):
...
+    @property
+    @_response_check
+    def certificates(self):
+        sk_x509 = self._backend._lib.OCSP_resp_get0_certs(self._basic)
+        num = self._backend._lib.sk_X509_num(sk_x509)
+        certs = []
+        for i in range(num):
+            x509 = self._backend._lib.sk_X509_value(sk_x509, i)
+            self._backend.openssl_assert(x509 != self._backend._ffi.NULL)
+            certs.append(_Certificate(self._backend, x509))
```
alex: This permits a UAF, doesn't it?

src/cryptography/hazmat/backends/openssl/ocsp.py
```diff
@@ -63,6 +82,225 @@ def _hash_algorithm(backend, cert_id):
...
+
+    @property
+    @_response_check
+    def responder_name(self):
+        x509_name, _ = self._responder_key_name()
+        if x509_name == self._backend._ffi.NULL:
+            return None
+        else:
+            # This name is not upref'd so we dupe it to control object
+            # lifetime
```
alex: I don't understand this.

src/cryptography/hazmat/backends/openssl/ocsp.py
reaperhulk: Oof, yes. I'll assign the OCSPResponse to a private attribute (like we do for RevokedCertificate with the parent CRL) to prevent this.

src/cryptography/hazmat/backends/openssl/ocsp.py
```diff
@@ -63,6 +85,231 @@ def _hash_algorithm(backend, cert_id):
...
+    @property
+    @_requires_successful_response
+    def responder_name(self):
+        x509_name, _ = self._responder_key_name()
+        if x509_name == self._backend._ffi.NULL:
+            return None
+        else:
+            # x509_name is obtained via a get0 call so we do not control
+            # its lifetime. To ensure the underlying memory is not freed we
+            # duplicate it to a new X509_NAME *.
```
alex: The part I don't understand is why you can't just pass the original instance to `_decode_x509_name`, which doesn't appear to manipulate the ref count in any way?

src/cryptography/hazmat/backends/openssl/ocsp.py
```diff
@@ -63,6 +85,231 @@ def _hash_algorithm(backend, cert_id):
...
+
+        reason_ptr = self._backend._ffi.new("int *")
+        self._backend._lib.OCSP_single_get0_status(
+            self._single,
+            reason_ptr,
+            self._backend._ffi.NULL,
+            self._backend._ffi.NULL,
+            self._backend._ffi.NULL,
+        )
+        if reason_ptr[0] == -1:
```
alex: Can you please add a comment explaining what the deal with this is?

src/cryptography/hazmat/backends/openssl/ocsp.py
reaperhulk: Oh I understand. And the answer is we totally don't have to because we don't need to keep the `X509_NAME *` alive since we decode it and discard it. I'll fix that.

PR #4455, opened 2018-09-02
Add flags to error on compile with incompatible pointer type 

reaperhulk:
good chance this will cause failures on openssl 1.0.2


reaperhulk:
@alex not really sure what we can do here. To catch the bug that #4454 fixes we need to be able to compile against OpenSSL < 1.1.0 and error on incompatible pointer types but NOT on incompatible pointer types that discard qualifiers. Travis can't do that. It may be possible in Jenkins, but not without making the CFLAGS conditional on which distro the job is running against, which is bleh.

PR #4457, opened 2018-09-03
don't sort the serial numbers in a parsed CRL

### LINKED ISSUE #4456: get_revoked_certificate_by_serial_number breaks signature validation on CertificateRevocationList
Hi,

Using the `get_revoked_certificate_by_serial_number` method on CertificateRevocationList will make OpenSSL sort the content of the CRL, and thus changing it.

This can be demonstrated with the following snippet:

```
def main():
    with open("crl.crl", "rb") as open_file:
        crl = x509.load_der_x509_crl(open_file.read(), default_backend())

    print(base64.b64encode(hashlib.sha256(crl.tbs_certlist_bytes).digest()).decode())
    crl.get_revoked_certificate_by_serial_number(123)
    print(base64.b64encode(hashlib.sha256(crl.tbs_certlist_bytes).digest()).decode())
```

This will return two different hashes. This then makes the signature validation on the CRL fail (tried using both `tbs_certlist_bytes` and `is_signature_valid`).

Not a big deal (I'm guessing most will verify the signature on the CRL before checking if the certificate is in it), but I'm thinking a note in the documentation would be nice. I scratched my head quite a bit before understanding why a previously good CRL suddenly failed signature validation :-)

Thanks!

reaperhulk:
OpenSSL sorts them in place and this breaks the signature and more.

fixes #4456

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -216,6 +216,7 @@ class _CertificateRevocationList(object):
     def __init__(self, backend, x509_crl):
         self._backend = backend
         self._x509_crl = x509_crl
+        self._sorted_crl = None
 
     def __eq__(self, other):
         if not isinstance(other, x509.CertificateRevocationList):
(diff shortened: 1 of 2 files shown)


alex:
So, this is dumb. There's no reason to use the sorted version if we're going to dup the CRL; it results in sorting every single time, making this less efficient.

We need to either cache the copied-and-sorted CRL, or we should just write a simple for-loop in Python.

alex:
You can save a few lines of code by splitting this out into a seperate method and using the `cached_property` decorator.

reaperhulk:
I consistently forget that we have that decorator.

alex:
How do you feel about issuing a point release for this?

alex:
Something broken with the josepy tests

reaperhulk:
It looks like our coverage pin is incompatible with pytest_cov at this point.

PR #4458, opened 2018-09-04
unpin coverage

reaperhulk:
of course, if this works it might just be luck


alex:
Think we should land this?

reaperhulk:
I'm concerned about the ordering problem but right now we can't run the josepy downstreams due to the coverage conflict. So either we try this or we have to write some bash to bypass josepy's default test infra

PR #4464, opened 2018-09-07
OCSP request extension parsing

reaperhulk:
This only parses `requestExtensions`. There is no support for `singleRequestExtensions`.

By flattening our OCSPRequest and Request into a single object we do have a situation where there are two different kinds of extensions that can be present. My current thinking is implementing this as `extensions` and doing `singleRequestExtensions` as `single_request_extensions`. Of course I need to find/generate vectors that contain `singleRequestExtensions` first.

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -2432,6 +2432,30 @@ These extensions are only valid within a :class:`RevokedCertificate` object.
 
         :type: :class:`datetime.datetime`
 
+OCSP Extensions
+~~~~~~~~~~~~~~~
+
+.. class:: OCSPNonce(nonce)
+
+    .. versionadded:: 2.4
+
+    OCSP nonce is an extension that is only valid inside
+    :class:`~cryptography.x509.ocsp.OCSPRequest` and
+    :class:`~cryptography.x509.ocsp.OCSPResponse` objects. The nonce
+    cryptographically binds a request and a response to prevent replay attacks.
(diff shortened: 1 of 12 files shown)


docs/x509/reference.rst
```diff
@@ -2854,6 +2878,15 @@ instances. The following common OIDs are available as constants.
 
         Corresponds to the dotted string ``"2.5.29.24"``.
 
+
+.. class:: OCSPExtensionOID
```
alex: Should we differentiate between `OCSPRequestExtension` and `OCSPResponseExtension`?

src/cryptography/hazmat/backends/openssl/ocsp.py
```diff
@@ -95,6 +95,10 @@ def serial_number(self):
     def hash_algorithm(self):
         return _hash_algorithm(self._backend, self._cert_id)
 
+    @property
```
alex: Should this be a `@cached_property`? IIRC the other extensions are

docs/x509/reference.rst
reaperhulk: Nonce, for example, is the same extension in both response and request right down to the OID so I think differentiating is probably undesirable.

src/cryptography/hazmat/backends/openssl/ocsp.py
reaperhulk: Sure, updating :)

Issue #3526, opened 2017-04-29
Add binding for OP_NO_TLSv1_3


tiran:
OpenSSL 1.1.1 will introduce ```OP_NO_TLSv1_3```. Let's add the flag to cryptography now so that modules like PyOpenSSL can expose it.

see https://github.com/pyca/pyopenssl/issues/624 and https://bugs.python.org/issue29136

alex:
We don't generally add bindings until they're in a released version of OpenSSL.

tiran:
It may make sense to deviate from the rule and make an exception. If OpenSSL 1.1.1 will come with TLS 1.3 enabled, libraries are going to run into all sorts of trouble. For example if both sides have TLS 1.3 enabled but the library does not enable TLS 1.3 cipher suites, the handshake will prefer TLS 1.3 but fail with no matching ciphers.

With ```SSL_OP_NO_TLSv1_3``` available now, libraries can opt out from TLS 1.3 more easily than ```getattr(module, 'SSL_OP_NO_TLSv1_3', 0)```.

alex:
They're going to need that anyways, since our bindings don't expose non-existent values.

I'm against making this constant special; if we want to provide a prettier API, it can be at the pyOpenSSL layer, not cryptography.

reaperhulk:
This will be landed pretty much the moment 1.1.1 is released anyway (whenever that is). Just to be clear: the risk of not doing this right now is that users consuming an unreleased OpenSSL linked against pyca/cryptography and then consumed via pyOpenSSL will encounter problems if they attempt to communicate with servers that are TLS 1.3 intolerant?

tiran:
@reaperhulk that's one possible issue. An even less likely issue involves TLS termination devices and firewalls that inspect traffic on higher levels.

The biggest issue are the new TLS 1.3 cipher suites. If an application overrides the default set of cipher suites and does not include TLS 1.3 suites, then handshake between TLS 1.3 enabled servers and clients will fail. It's pretty common to replace OpenSSL's cipher suite selection with a custom list of trusted ciphers.

Applications and libraries like PyOpenSSL and Twisted have to change for TLS 1.3 no matter what. :/

Issue #4197, opened 2018-04-16
OpenSSL 1.1.1 Features/Tasks


reaperhulk:
This is a meta bug to track the various things we should do once 1.1.1 is released.

- [x] Add a 1.1.1 builder to CI (#4112)
- [x] ed25519 signing (#3509) (#4114)
- [x] SHA3 digests (#4573)
- [x] SHAKE (#4611)
- [x] SHA512-224, SHA512-256 (#4575)
- [x] X448 kex (#4136)
- [x] ed448 (#4610)
- [x] Add binding for SSL_CTX_set_ciphersuites (#4255)

The last published OpenSSL release schedule (https://www.openssl.org/policies/releasestrat.html) targets a release around May 15, 2018 so we're around a month away from final.

I'm tentatively 👍 on shipping a wheel with 1.1.1 as soon as we have some features that can use it, but if others feel strongly about waiting a bit to see if there are bug fix releases we can discuss it.

maqp:
OpenSSL 1.1.1 has finally been [released](https://www.openssl.org/blog/blog/2018/09/11/release111/).

reaperhulk:
This was finished with 2.6

PR #4476, opened 2018-09-26
add ed25519 bindings

reaperhulk:
Refs #4114

PR #4477, opened 2018-09-26
Bind SSL_OP_NO_TLSv1_3

tiran:
See: https://github.com/pyca/cryptography/issues/4197
Fixes: #3526


src/_cffi_src/openssl/cryptography.py
```diff
@@ -44,6 +44,8 @@
     (OPENSSL_VERSION_NUMBER >= 0x10100000 && !CRYPTOGRAPHY_IS_LIBRESSL)
 #define CRYPTOGRAPHY_OPENSSL_110F_OR_GREATER \
     (OPENSSL_VERSION_NUMBER >= 0x1010006f && !CRYPTOGRAPHY_IS_LIBRESSL)
+#define CRYPTOGRAPHY_OPENSSL_111_OR_GREATER \
```
alex: This isn't actually used, please remove it.

src/_cffi_src/openssl/ssl.py
```diff
@@ -806,4 +807,8 @@
 static const long Cryptography_HAS_CIPHER_DETAILS = 1;
 #endif
 
+#if CRYPTOGRAPHY_OPENSSL_LESS_THAN_111
```
alex: This attribute needs to be available only when it's really present in OpenSSL. See #4476 for all the places that need editing.

reaperhulk:
Replaced by #4555

PR #4478, opened 2018-10-04
Add a tutorial explianing how to check the exactness of a certificate

g-goessel:
I wanted to check if a certificate was correct but failed at finding a documentation. After many attempts I found out that I had to use the verify method on the public key in order to manually check the signature.

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -473,17 +473,13 @@ X.509 Certificate Object
            >>> from cryptography.hazmat.primitives.asymmetric import padding
            >>> issuer_public_key = load_pem_public_key(pem_issuer_public_key, default_backend())
            >>> cert_to_check = x509.load_pem_x509_certificate(pem_data_to_check, default_backend())
-           >>> try:
-           ...     issuer_public_key.verify(
-           ...       cert_to_check.signature,
-           ...       cert_to_check.tbs_certificate_bytes,
-           ...       #Depends on the algorithm used to create the certificate
-           ...       padding.PKCS1v15(),
-           ...       cert_to_check.signature_hash_algorithm,
-           ...     )
-           ...     print(True)
-           ... except InvalidSignature:
-           ...     print(False)
+           >>> issuer_public_key.verify(
+           ...   cert_to_check.signature,
+           ...   cert_to_check.tbs_certificate_bytes,
+           ...   #Depends on the algorithm used to create the certificate
+           ...   padding.PKCS1v15(),
+           ...   cert_to_check.signature_hash_algorithm,
+           ... )
 
 
     .. method:: public_bytes(encoding)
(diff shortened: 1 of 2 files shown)


g-goessel:
Thank you for your explanation.
I know this is not all we have to check but I wanted to do the steps manually (for a school project actually).

reaperhulk:
If you'd like to add the example code under the `tbs_certificate_bytes` section we'd be willing to take that as a PR. It would need to be done with `doctest` rather than `pycon` as well since we want to run our code during doc generation to ensure we always have functional examples.

docs/x509/reference.rst
```diff
@@ -415,6 +463,29 @@ X.509 Certificate Object
...
+        on the public key (this does not mean that the cert should be truster):
+
+        .. doctest::
+
+           >>> from cryptography.hazmat.primitives.serialization import load_pem_public_key
+           >>> from cryptography.exceptions import InvalidSignature
+           >>> from cryptography.hazmat.primitives.asymmetric import padding
+           >>> issuer_public_key = load_pem_public_key(pem_issuer_public_key, default_backend())
+           >>> cert_to_check = x509.load_pem_x509_certificate(pem_data_to_check, default_backend())
+           >>> try:
```
reaperhulk: For the doctest we can just call the verify call directly, no need to pu tit in a try block.

g-goessel: hum ok, will change it
sorry for the failed CI, I made a copy/paste error...

docs/x509/reference.rst
```diff
@@ -415,6 +462,29 @@ X.509 Certificate Object
         certificate validation is a complex problem that involves much more
         than just signature checks.
 
+        An example on how to check if a certificate signature is correct with a method
+        on the public key (this does not mean that the cert should be trusted):
+
+        .. doctest::
+
+           >>> from cryptography.hazmat.primitives.serialization import load_pem_public_key
+           >>> from cryptography.exceptions import InvalidSignature
```
reaperhulk: This import is no longer required

docs/x509/reference.rst
```diff
@@ -415,6 +462,29 @@ X.509 Certificate Object
...
+           >>> from cryptography.hazmat.primitives.asymmetric import padding
+           >>> issuer_public_key = load_pem_public_key(pem_issuer_public_key, default_backend())
+           >>> cert_to_check = x509.load_pem_x509_certificate(pem_data_to_check, default_backend())
+           >>> # verify() raises a
```
reaperhulk: This comment can be removed (or made into a complete sentence)

docs/x509/reference.rst
```diff
@@ -415,6 +462,29 @@ X.509 Certificate Object
...
+           ...   cert_to_check.signature,
+           ...   cert_to_check.tbs_certificate_bytes,
+           ...   #Depends on the algorithm used to create the certificate
+           ...   padding.PKCS1v15(),
+           ...   cert_to_check.signature_hash_algorithm,
+           ... )
+
+           An
+           :class:`~cryptography.exceptions.InvalidSignature`
+           exception will be raised otherwise.
```
reaperhulk: "will be raised if the signature fails to verify"

g-goessel:
Oops made a mistake while rebasing, please stash the commits together when merging :smile:

PR #4486, opened 2018-10-07
Fixes #4459 -- switch all our builds to xenial on travis

### LINKED ISSUE #4459: Investigate migrating all travis builders to xenial
This would give us a newer compiler so we could add the flags in #4455 to all builds. We would need to compile our own OpenSSL though (or add a trusty builder to jenkins) to continue covering 1.0.1.

alex:


alex:
Issues:

- [ ]  pypy-5.3 is not available for xenial
- [ ] 1.0.1 doesn't have a `install_ssldirs` `make` target -- we use this one instead of `make install` to avoid installing docs (purely for speed), is that right?
- [ ] urllib3 tests are failing
- [ ] dynamodb-encryption-sdk tests are failing

I can resolve three of these by simply running those builds on trusty still. Or I can dive in and investigate them. Opinions, @reaperhulk ?

reaperhulk:
We do use `install_ssldirs` solely for speed, yes.

I'd like to know why we're seeing failures on urllib3 and dynamo (which a quick glance shows `ImportError: No module named google_compute_engine`, huh), but I'd be okay with just leaving those on trusty for the moment.

Also, what about migrating to bionic? Does Travis have that yet?

.travis.yml
```diff
@@ -80,12 +76,17 @@ matrix:
           env: DOWNSTREAM=aws-encryption-sdk
         - python: 2.7
           env: DOWNSTREAM=dynamodb-encryption-sdk OPENSSL=1.1.0i
+          # dynamodb-encryption-sdk tests fail on xenial for whatever reason
+          dist: trust
```
reaperhulk: This...shouldn't work?

.travis.yml
```diff
@@ -80,12 +76,17 @@ matrix:
...
         - python: 2.7
           env: DOWNSTREAM=certbot OPENSSL=1.1.0i
         - python: 2.7
           env: DOWNSTREAM=certbot-josepy
         - python: 2.7
           env: DOWNSTREAM=urllib3
+          # urllib3 tests fail on xenial for whatever reason
+          dist: trusty
+          sudo: false
```
reaperhulk: This one says sudo: false but the other trusty builds don't. I assume we should put it everywhere?

.travis/install.sh
```diff
@@ -23,9 +23,14 @@ if [ -n "${OPENSSL}" ]; then
         shlib_sed
         make depend
         make -j"$(nproc)"
-        # avoid installing the docs
-        # https://github.com/openssl/openssl/issues/6685#issuecomment-403838728
-        make install_sw install_ssldirs
+        if [[ "${OPENSSL}" =~ "1.0.1" ]]; then
+            # OpenSSL 1.0.1 doesn't support installing without the docs.
```
reaperhulk: 😆 one day we'll get to drop OpenSSL 1.0.1 support.

.travis.yml
alex: I bet trusty is the default on parse failure

.travis.yml
alex: urllib3's tests fail on `sudo: true` for whatever reason, no one else needs it

.travis/install.sh
alex: some day

PR #4492, opened 2018-10-09
Add definitions that help with hostname checking

kaie:
This is related to https://github.com/pyca/pyopenssl/issues/795

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -626,9 +626,7 @@ static const long Cryptography_HAS_NEXTPROTONEG = 1;
 /* SSL_get0_param was added in OpenSSL 1.0.2. */
 #if CRYPTOGRAPHY_OPENSSL_LESS_THAN_102 && !CRYPTOGRAPHY_LIBRESSL_27_OR_GREATER
 X509_VERIFY_PARAM *(*SSL_get0_param)(SSL *) = NULL;
-static const long Cryptography_HAS_SSL_GET0_PARAM = 0;
 #else
-static const long Cryptography_HAS_SSL_GET0_PARAM = 1;
 #endif
 
 /* ALPN was added in OpenSSL 1.0.2. */
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -185,6 +185,8 @@ X509 *SSL_get_certificate(const SSL *);
 X509 *SSL_get_peer_certificate(const SSL *);
 int SSL_get_ex_data_X509_STORE_CTX_idx(void);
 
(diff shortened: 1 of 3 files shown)


src/_cffi_src/openssl/ssl.py
```diff
@@ -185,6 +185,8 @@
 X509 *SSL_get_peer_certificate(const SSL *);
 int SSL_get_ex_data_X509_STORE_CTX_idx(void);
 
+X509_VERIFY_PARAM *SSL_get0_param(SSL *ssl);
```
alex: Please remove the `ssl` name, we don't include parameter names.

alex:
It looks like some of these symbols do not exist in older OpenSSLs, so you'll need to make them conditional bindings.

You can see https://github.com/pyca/cryptography/pull/4476 for an example of a PR adding optional bindings

reaperhulk:
This LGTM other than the need for conditionally binding it so the symbols aren't exposed on < 1.0.2.

kaie:
> As @tiran pointed out on the pyopenssl PR we'll also need to make sure the check covers LibreSSL < 2.7.1 as `X509_VERIFY_PARAM_set1_host` is not available until 2.7.1.

x509_vfy.py already contains:
```
#if CRYPTOGRAPHY_OPENSSL_102_OR_GREATER
#else
#if !CRYPTOGRAPHY_LIBRESSL_27_OR_GREATER
int (*X509_VERIFY_PARAM_set1_host)(X509_VERIFY_PARAM *, const char *,
                                   size_t) = NULL;
#endif
#endif
```
I assume you want the equivalent for SSL_get0_param, as it seems to be a direct helper function for the set1_host API.

Besides X509_CHECK_FLAG_NO_PARTIAL_WILDCARDS, I'm adding a couple of additional related X509_CHECK_FLAG_* flags.

I noticed you have a couple of #ifndef checks like the following:
```
/* These 3 defines are unavailable in LibreSSL 2.5.x, but may be added
   in the future... */
#ifndef X509_V_ERR_HOSTNAME_MISMATCH
static const long X509_V_ERR_HOSTNAME_MISMATCH = 0;
#endif

```
I see the X509_CHECK_FLAG_* aren't available in old libressel either, so you probably want to protect these flags in the same way.

kaie:
I notice that _conditional.py needs to get updated, too. Is it acceptable to extend existing cryptography_has_102_verification_params?

reaperhulk:
Yes, `cryptography_has_102_verification_params` is appropriate for this binding so you can reuse that!

alex:
`X509_CHECK_FLAG_NEVER_CHECK_SUBJECT` is new in 1.1.0, so it'll need slightly seperate handling.

PR #4495, opened 2018-10-10
simplify some language

reaperhulk:


docs/x509/reference.rst
```diff
@@ -462,8 +462,7 @@ X.509 Certificate Object
         certificate validation is a complex problem that involves much more
         than just signature checks.
 
-        An example on how to check if a certificate signature is correct with a method
-        on the public key (this does not mean that the cert should be trusted):
+        To validate the signature on a certificate:
```
alex: This really needs the parenthetical indicating it's only part of the full validation you ought to do

reaperhulk: I originally had that but removed it because that exact admonition is the sentence directly above it.

alex: Change "does not validate anything else" to "and does not perform any of the other checks needed for secure certificate validation. Further, this code will only work for RSA public keys with PKCS1v15 signatures, and so it can't be used for general purpose signature verification." or something?

PR #4504, opened 2018-10-19
Adds a more descriptive error msg for wrong wrapping

### LINKED ISSUE #4498: More descriptive error msg for certificate loading
When we fail to load a PEM certificate we do this: https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/backends/openssl/backend.py#L1151

Given that people occasionally attempt to load valid certs that don't have proper line wrapping (see: https://github.com/pyca/cryptography/issues/4487 and https://github.com/pyca/cryptography/issues/4085) should we alter the text to say something about how what a valid PEM looks like and that they should check their cert?

andrealmeid:
I made a proof of concept of a function that should fix #4498. I addressed my code to check specifically for a bad wrapped RSA 512 certificate, verifying positions to ensure that it has a `\n` there. I have some questions about my code:

- This function is OK like that or should it be more generic than this?
- Is my logic right? 
- Where should I place it? I think the code would be cleaner if it has a `pem_check_wrap()`  function.
- There are performance issue in this code? I could remove the `chr()` cast for example. I inserted it to make the logic more obvious. 

I also want to write a unit test in the future to match this  specific case. Thanks!

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -123,7 +123,7 @@ the following rules:
 
 * has a one-line header like this: ``-----BEGIN [FILE TYPE]-----``
   (where ``[FILE TYPE]`` is ``CERTIFICATE``, ``PUBLIC KEY``, ``PRIVATE KEY``,
-  et cetera)
+  etc.)
 
 * has a one-line footer like this: ``-----END [FILE TYPE]-----``
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -111,28 +111,22 @@ not yet possible you  can also install ``cryptography`` with
 dependency. This workaround will be available until the feature is fully
 removed.
 
-.. _`NaCl`: https://nacl.cr.yp.to/
-.. _`PyNaCl`: https://pynacl.readthedocs.io
-.. _`WSGIApplicationGroup`: https://modwsgi.readthedocs.io/en/develop/configuration-directives/WSGIApplicationGroup.html
-.. _`issue`: https://github.com/pyca/cryptography/issues
-
-Why I can't import my PEM file?
+Why can't I import my PEM file?
 -------------------------------
 
-PEM is a file format to store keys, certificates and others cryptographic data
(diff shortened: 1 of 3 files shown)


reaperhulk:
Hi @andrealmeid, thanks for taking this task on. I think this approach is more complex than we want though -- rather than try to determine why the key isn't loading via checking what byte values are at specific offsets we should just change the ValueError text to something like "Unable to load certificate, is it a valid PEM?". We could add more there (like that PEM files have BEGIN and END lines and no more than 64 base64 encoded characters per line between them) or we could potentially write up a longer explanation of what a PEM is in our docs and then put a link to that in the error msg.

andrealmeid:
Ok @reaperhulk! I think will take the docs approach, filled with some examples. May I change `ValueError` messages for every function that imports a PEM file to link that docs page? This would address issues involving PEM in certificates, certificates requests, keys, ...

reaperhulk:
If you want to do the docs approach that's fine. It might be best to add it as a FAQ entry.

alex:
Looks like there's a flake8 issue.

andrealmeid:
Thanks @alex, should be fixed now.

docs/faq.rst
```diff
@@ -111,6 +111,37 @@ not yet possible you  can also install ``cryptography`` with
...
+
+PEM is a format (defined by several RFCs, but originally :rfc:`1421`) for
+encoding keys, certificates and others cryptographic data into a regular form.
+The data is encoded as base64 and wrapped with a header and footer.
+
+If you are having trouble importing PEM files, make sure your file fits
+the following rules:
+
+* has a one-line header like this: ``-----BEGIN [FILE TYPE]-----``
+  (where ``[FILE TYPE]`` is certificate, public key, et cetera)
```
alex: Please use real examples here, ``CERTIFICATE``, ``PUBLIC KEY``.

PR #4519, opened 2018-10-24
Updated BLAKE2s and BLAKE2b error messages from unsupportedalgorithm …

colinmetcalf:
…exception to an explicit error.

The error is now "ValueError: Digest size must be 32" (or 64 for BLAKE2b)
This was done to give a more contextual error message and should be in place until OpenSSL supports variable lengths.

### DIFF
--- a/src/cryptography/hazmat/primitives/hashes.py
+++ b/src/cryptography/hazmat/primitives/hashes.py
@@ -151,6 +151,12 @@ class BLAKE2b(object):
     block_size = 128
 
     def __init__(self, digest_size):
+
+        if (
+            digest_size != 64
+        ):
+            raise ValueError("Digest size must be 64")
+
         if (
             digest_size > self._max_digest_size or
             digest_size < self._min_digest_size
@@ -172,6 +178,12 @@ class BLAKE2s(object):
     _min_digest_size = 1
 
     def __init__(self, digest_size):
+
+        if (
+            digest_size != 32
+        ):
+            raise ValueError("Digest size must be 32")
+
         if (
             digest_size > self._max_digest_size or
             digest_size < self._min_digest_size


colinmetcalf:
Sorry for delay, got a bit hectic. For the test you referring too, do you mean using pytest with the /tests directory or would you like me too add tests specifically for this change in either the /test_hashes.py file or the evptests.txt. Sorry for my ignorance, thanks for the help!

reaperhulk:
Add tests that cover this specific change in `test_hashes.py`.

colinmetcalf:
@reaperhulk I've addressed the requested changes, thanks for your help!

reaperhulk:
Tests are failing due to mismatched tabs and spaces, but there are two other issues here. One is that I hadn't looked closely enough to realize we already had tests for disallowed lengths because we already had code verifying a range of acceptable inputs. This means when you put the new checks in place the old checks are now irrelevant and should be removed (we can't execute those lines with tests because the new conditional will always catch it first). Second, the added test cases you have shouldn't work because passing the correct value won't raise a ValueError. Since test cases already exist for invalid values (which I didn't realize, sorry) you can just remove those lines.

reaperhulk:
https://travis-ci.org/pyca/cryptography/jobs/451149438?utm_source=github_status&utm_medium=notification needs to be fixed. You're still going to have a coverage problem once the tests pass though. Your new init function looks like this (the BLAKE2s one looks similar):

```python
        if (digest_size != 64):
            raise ValueError("Digest size must be 64")

        if (
            digest_size > self._max_digest_size or
            digest_size < self._min_digest_size
        ):
            raise ValueError("Digest size must be {0}-{1}".format(
                self._min_digest_size, self._max_digest_size)
            )
```

You can see that the second if statement can never actually raise a ValueError since any invalid value is caught by the first check. Since we require 100% coverage this means CI will fail since no test covers those branches. The fix is to remove the second if from both the classes since it is now dead code.

colinmetcalf:
Dead code has been removed. flake8 still fails, but I believe that is still an error unrelated to this change.
Thanks for the clarification!

tests/hazmat/primitives/test_hashes.py
```diff
@@ -161,7 +161,6 @@ def test_invalid_digest_size(self, backend):
         with pytest.raises(ValueError):
             hashes.BLAKE2s(digest_size=-1)
 
-
```
reaperhulk: Pep8 failing because of this line removal. It is not an unrelated error.

src/cryptography/hazmat/primitives/hashes.py
```diff
@@ -151,13 +151,9 @@ class BLAKE2b(object):
...
     def __init__(self, digest_size):
-        if (
-            digest_size > self._max_digest_size or
-            digest_size < self._min_digest_size
-        ):
-            raise ValueError("Digest size must be {0}-{1}".format(
-                self._min_digest_size, self._max_digest_size)
-            )
+
+        if (digest_size != 64):
```
alex: Drop the parens please

src/cryptography/hazmat/primitives/hashes.py
```diff
@@ -172,13 +168,9 @@ class BLAKE2s(object):
...
     def __init__(self, digest_size):
-        if (
-            digest_size > self._max_digest_size or
-            digest_size < self._min_digest_size
-        ):
-            raise ValueError("Digest size must be {0}-{1}".format(
-                self._min_digest_size, self._max_digest_size)
-            )
+
+        if (digest_size != 32):
```
alex: Same

PR #4525, opened 2018-10-25
don't limit hypothesis's time to generate valid test permutations

reaperhulk:
It's important that CI find a new way to be flaky every single day.

### DIFF
--- a/tests/hypothesis/test_fernet.py
+++ b/tests/hypothesis/test_fernet.py
@@ -10,8 +10,8 @@ from cryptography.fernet import Fernet
 
 # Unlimited timeout will become the default in the future. When it does
 # we should remove this. See:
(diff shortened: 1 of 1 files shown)


tests/hypothesis/test_fernet.py
```diff
@@ -8,6 +8,10 @@
 from cryptography.fernet import Fernet
 
 
+# Unlimited timeout will become the default in the future. When it does
+# we should remove this. See:
+# https://hypothesis.readthedocs.io/en/latest/settings.html?highlight=
+# timeout#hypothesis.settings.timeout
```
alex: Tiny nit: please don't include the highlight querystring

reaperhulk: tbf without it there's no way to pick out the timeout section (there's no anchor).

PR #4526, opened 2018-10-25
docker experiment in travis

reaperhulk:


reaperhulk:
Everything works here except coverage is being written as root inside the container so when coverage is submitted outside of it nothing is found. https://github.com/pyca/infra/pull/190 will make it so we can just use `-u travis` and get the same uid/gid travis uses and bypass this whole complicated web of permissions shenanigans.

.travis/install.sh
```diff
@@ -46,6 +46,10 @@ elif [ -n "${LIBRESSL}" ]; then
     fi
 fi
 
+if [ -n "${DOCKER}" ]; then
+    docker pull "$DOCKER"
```
alex: Can we add an assert that there's no `$OPENSSL` or `$LIBRESSL` when `$DOCKER` is set?

.travis/run.sh
```diff
@@ -22,7 +22,11 @@ fi
 
 source ~/.venv/bin/activate
 
-if [ -n "${TOXENV}" ]; then
+if [ -n "${DOCKER}" ]; then
+    # We will be able to drop the -u once we switch the default container user in the
+    # dockerfiles.
+    docker run -u 2000:2000 -v "${TRAVIS_BUILD_DIR}":"${TRAVIS_BUILD_DIR}" -v "${HOME}/wycheproof":/wycheproof -e TOXENV "${DOCKER}" /bin/sh -c "cd ${TRAVIS_BUILD_DIR};tox -- --wycheproof-root='/wycheproof'"
```
alex: Can you add some line breaks so this is readable

.travis.yml
```diff
@@ -55,6 +55,56 @@ matrix:
...
+        - python: 3.6
+          services: docker
+          env: TOXENV=py36 DOCKER=pyca/cryptography-runner-sid
+        - python: 2.7
+          services: docker
+          env: TOXENV=py27 DOCKER=pyca/cryptography-runner-ubuntu-rolling
+        - python: 3.6
+          services: docker
+          env: TOXENV=py36 DOCKER=pyca/cryptography-runner-ubuntu-rolling
+        - python: 3.6
```
alex: Is this actually 3.6 in any sense?

reaperhulk: It's whatever `python` is. Just confirmed that it is 2.7. We could alter `randomorder` to ask for `python3` and then it'd be 3.6, but as is I'll change it to say 2.7.

.travis/install.sh
reaperhulk: So something like `if [ -n "${OPENSSL}" ] || [ -n "${LIBRESSL}" ]; then echo "Can't set those things"; exit 1; fi`

.travis/run.sh
reaperhulk: Precious bytes.

alex: You know I love wantonly wasting bytes

.travis/install.sh
alex: sure, that's what passes for an assert in this godforesaken language

.travis/run.sh
```diff
@@ -25,7 +25,10 @@ source ~/.venv/bin/activate
 if [ -n "${DOCKER}" ]; then
     # We will be able to drop the -u once we switch the default container user in the
     # dockerfiles.
-    docker run -u 2000:2000 -v "${TRAVIS_BUILD_DIR}":"${TRAVIS_BUILD_DIR}" -v "${HOME}/wycheproof":/wycheproof -e TOXENV "${DOCKER}" /bin/sh -c "cd ${TRAVIS_BUILD_DIR};tox -- --wycheproof-root='/wycheproof'"
+    docker run -u 2000:2000 -v "${TRAVIS_BUILD_DIR}":"${TRAVIS_BUILD_DIR}" \
+        -v "${HOME}/wycheproof":/wycheproof \
+        -e TOXENV "${DOCKER}" \
+        /bin/sh -c "cd ${TRAVIS_BUILD_DIR};tox -- --wycheproof-root='/wycheproof'"
```
alex: space after `;` please

.travis/run.sh
```diff
@@ -25,7 +25,10 @@ source ~/.venv/bin/activate
 if [ -n "${DOCKER}" ]; then
     # We will be able to drop the -u once we switch the default container user in the
     # dockerfiles.
-    docker run -u 2000:2000 -v "${TRAVIS_BUILD_DIR}":"${TRAVIS_BUILD_DIR}" -v "${HOME}/wycheproof":/wycheproof -e TOXENV "${DOCKER}" /bin/sh -c "cd ${TRAVIS_BUILD_DIR};tox -- --wycheproof-root='/wycheproof'"
+    docker run -u 2000:2000 -v "${TRAVIS_BUILD_DIR}":"${TRAVIS_BUILD_DIR}" \
```
alex: can you move the volume onto it's own line

.travis/run.sh
```diff
@@ -22,7 +22,15 @@ fi
...
 
-if [ -n "${TOXENV}" ]; then
+if [ -n "${DOCKER}" ]; then
+    # We will be able to drop the -u once we switch the default container user in the
+    # dockerfiles.
+    docker run --rm -u 2000:2000 \
+        -v "${TRAVIS_BUILD_DIR}":"${TRAVIS_BUILD_DIR}" \
+        -v "${HOME}/wycheproof":/wycheproof \
+        -e TOXENV "${DOCKER}" \
+        /bin/sh -c "cd ${TRAVIS_BUILD_DIR}; tox -- --wycheproof-root='/wycheproof'"
```
alex: Can you use `-w` to docker instead of `cd`

PR #4533, opened 2018-10-28
add IDP OID and docs

reaperhulk:
refs #2546


reaperhulk:
I put this OID in the wrong section (CRLEntryExtensionOID is just for extensions on RevokedCertificate objects inside a CRL). Moving it in a second.

reaperhulk:
Fixed

PR #4534, opened 2018-10-28
Add eq/ne/hash to PrecertificateSignedCertificateTimestamps

### LINKED ISSUE #2578: Implement __hash__ on all extensions we've implemented __eq__ on

reaperhulk:
This requires adding it to SignedCertificateTimestamps as well.

fixes #2578


src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -534,3 +534,22 @@ def entry_type(self):
         # we only have precerts.
         assert entry_type == self._backend._lib.CT_LOG_ENTRY_TYPE_PRECERT
         return x509.certificate_transparency.LogEntryType.PRE_CERTIFICATE
+
+    def __hash__(self):
+        return hash(
+            (self.version, self.log_id, self.timestamp, self.entry_type)
```
alex: This isn't right, we don't expose enough info ATM to reliably compare these.

reaperhulk: If we bound `SCT_get0_signature` we could use signature bytes as a comparator, right?

alex: I guess signature would probably be enough

tests/x509/test_x509_ext.py
```diff
@@ -4469,6 +4469,85 @@ def test_generate(self, backend):
...
+        skip_message="Requires OpenSSL 1.1.0f+",
+    )
+    def test_eq(self, backend):
+        scts = _load_cert(
+            os.path.join("x509", "badssl-sct.pem"),
+            x509.load_pem_x509_certificate,
+            backend
+        ).extensions.get_extension_for_class(
+            x509.PrecertificateSignedCertificateTimestamps
+        ).value._signed_certificate_timestamps
```
alex: Uhh, why do you need to access a private attribute?

PR #4535, opened 2018-10-28
add pkcs12 test vectors

reaperhulk:
refs #2860


reaperhulk:
It's related to the PKCS12 vectors in that that cert + key is inside every PKCS12 file, so a test might want to confirm that what is parsed out of the PKCS12 is identical to the cert + key pair. Looking at the docs it appears I forgot to actually state in the prose that each vector contains that cert+key.

PR #4545, opened 2018-10-29
do not pass conversion flags on Solaris

gaborbernat:
The package works on Solaris, but the Solaris compiler does not understand these flags.


gaborbernat:
@reaperhulk thanks a lot, any chance for a patch release soon-ish? :+1:

reaperhulk:
The plan is to release 2.4 in the next week or so.

PR #4552, opened 2018-11-03
IssuingDistributionPoint support (parse only)

reaperhulk:
h/t to Irina Renteria for the initial work here

refs #2546


docs/x509/reference.rst
```diff
@@ -2319,6 +2319,76 @@ X.509 Extensions
 
         :type: int
 
+.. class:: IssuingDistributionPoint(only_contains_user_certs,\
+           only_contains_ca_certs, indirect_crl, only_contains_attribute_certs,\
+           only_some_reasons=None, full_name=None, relative_name=None)
```
alex: We don't generally use defaults for these, why here?

docs/x509/reference.rst
```diff
@@ -2319,6 +2319,76 @@ X.509 Extensions
...
+        :type: bool
+
+        Set to ``True`` if the CRL this extension is embedded within only
+        contains information about attribute certificates.
+
+    .. attribute:: only_some_reasons
+
+        :type: frozenset of :class:`ReasonFlags` or None
+
+        The reasons for which the issuing distribution point is valid.
```
alex: "None indicates that it's valid for ..."

src/cryptography/hazmat/backends/openssl/decode_asn1.py
```diff
@@ -464,6 +464,28 @@ def _decode_general_subtrees(backend, stack_subtrees):
     return subtrees
 
 
+def _decode_issuing_dist_point(backend, idp):
+    idp = backend._ffi.cast("ISSUING_DIST_POINT *", idp)
+    idp = backend._ffi.gc(idp, backend._lib.ISSUING_DIST_POINT_free)
+    full_name = None
+    relative_name = None
```
alex: Assign these in the else please

src/cryptography/hazmat/backends/openssl/decode_asn1.py
```diff
@@ -464,6 +464,28 @@ def _decode_general_subtrees(backend, stack_subtrees):
...
+    if idp.distpoint != backend._ffi.NULL:
+        full_name, relative_name = _decode_distpoint(backend, idp.distpoint)
+
+    only_user = idp.onlyuser == 255
+    only_ca = idp.onlyCA == 255
+    indirect_crl = idp.indirectCRL == 255
+    only_attr = idp.onlyattr == 255
+    only_some_reasons = None
```
alex: same

src/cryptography/x509/__init__.py
```diff
@@ -51,6 +51,7 @@
 OID_FRESHEST_CRL = ExtensionOID.FRESHEST_CRL
 OID_INHIBIT_ANY_POLICY = ExtensionOID.INHIBIT_ANY_POLICY
 OID_ISSUER_ALTERNATIVE_NAME = ExtensionOID.ISSUER_ALTERNATIVE_NAME
+OID_ISSUING_DISTRIBUTION_POINT = ExtensionOID.ISSUING_DISTRIBUTION_POINT
```
alex: Are we exposing these? I thought this was just for backwards compat, and new ones should be on `ExtensionOID` only?

src/cryptography/x509/extensions.py
```diff
@@ -1446,6 +1446,138 @@ def __repr__(self):
...
+                "unspecified and remove_from_crl are not valid reasons in an "
+                "IssuingDistributionPoint"
+            )
+
+        bools = [
+            only_contains_user_certs, only_contains_ca_certs,
+            indirect_crl, only_contains_attribute_certs
+        ]
+
+        if not all([isinstance(x, bool) for x in bools]):
```
alex: Drop the list-comprehension. Also I think this is easier to read if you just write out the `and`s

tests/x509/test_x509_ext.py
```diff
@@ -4440,6 +4440,210 @@ def test_inhibit_any_policy(self, backend):
...
+class TestIssuingDistributionPointExtension(object):
+    @pytest.mark.parametrize(
+        ("filename", "expected"),
+        [
+            [
+                "crl_idp_fullname_indirect_crl.pem",
+                x509.IssuingDistributionPoint(
+                    False, False, True, False, None, [
+                        x509.UniformResourceIdentifier(
+                            u"http://myhost.com/myca.crl")], None)
```
alex: Please wrap this different so we don't have this rightward drift.

tests/x509/test_x509_ext.py
```diff
@@ -4440,6 +4440,210 @@ def test_inhibit_any_policy(self, backend):
...
+            [TypeError, ('notabool', False, False, False, None, None, None)],
+            [TypeError, (False, 'notabool', False, False, None, None, None)],
+            [TypeError, (False, False, 'notabool', False, None, None, None)],
+            [TypeError, (False, False, False, 'notabool', None, None, None)],
+            [ValueError, (True, True, False, False, None, None, None)],
+            [ValueError, (False, False, True, True, None, None, None)],
+            [ValueError, (False, False, False, False, None, None, None)],
+        ]
+    )
+    def test_invalid_init(self, error, args):
```
alex: Let's name the args, for easier reading

docs/x509/reference.rst
```diff
@@ -2319,6 +2319,77 @@ X.509 Extensions
 
         :type: int
 
+.. class:: IssuingDistributionPoint(only_contains_user_certs,\
+           only_contains_ca_certs, indirect_crl, only_contains_attribute_certs,\
+           only_some_reasons, full_name, relative_name)
```
alex: These should be in a different order, either the one from the RFC, or all the only "only"s togther, or something else, but right now they look basically random.

docs/x509/reference.rst
```diff
@@ -2319,6 +2319,77 @@ X.509 Extensions
...
+
+    .. versionadded:: 2.5
+
+    Issuing distribution point is a CRL extension that identifies the CRL
+    distribution point and scope for a particular CRL, and it indicates whether
```
alex: Break the sentence at ", and", it's a run on ATM.

src/cryptography/hazmat/backends/openssl/decode_asn1.py
```diff
@@ -464,6 +464,30 @@ def _decode_general_subtrees(backend, stack_subtrees):
...
+def _decode_issuing_dist_point(backend, idp):
+    idp = backend._ffi.cast("ISSUING_DIST_POINT *", idp)
+    idp = backend._ffi.gc(idp, backend._lib.ISSUING_DIST_POINT_free)
+    if idp.distpoint != backend._ffi.NULL:
+        full_name, relative_name = _decode_distpoint(backend, idp.distpoint)
+    else:
+        full_name = None
+        relative_name = None
+
+    only_user = idp.onlyuser == 255
```
alex: oh my god, is this seriously how this works?

src/cryptography/x509/extensions.py
```diff
@@ -1446,6 +1446,145 @@ def __repr__(self):
...
+                "indirect_crl and only_contains_attribute_certs "
+                "must all be boolean."
+            )
+
+        bools = [
+            only_contains_user_certs, only_contains_ca_certs,
+            indirect_crl, only_contains_attribute_certs
+        ]
+
+        if len([x for x in bools if x is True]) > 1:
```
alex: I think this is actually more clearly written as `if only_contains_user_certs + only_contains_ca_certs + indirect_crl + only_contains_attribute_certs > 1`?

src/cryptography/x509/extensions.py
```diff
@@ -1446,6 +1446,145 @@ def __repr__(self):
...
+            raise ValueError(
+                "Only one of the following can be set to True: "
+                "only_contains_user_certs, only_contains_ca_certs, "
+                "indirect_crl, only_contains_attribute_certs"
+            )
+
+        if (
+            not any([
```
alex: can this just be a chain of `or`?

tests/x509/test_x509_ext.py
```diff
@@ -4440,6 +4440,229 @@ def test_inhibit_any_policy(self, backend):
...
+    @pytest.mark.parametrize(
+        ("filename", "expected"),
+        [
+            [
+                "crl_idp_fullname_indirect_crl.pem",
+                x509.IssuingDistributionPoint(
+                    False, False, True, False, None, [
+                        x509.UniformResourceIdentifier(
+                            u"http://myhost.com/myca.crl")
+                    ], None
```
alex: The indentation on all these is super weird

alex: And please use kwargs, they're impossible to review without refereing to the docs 100 times ATM.

tests/x509/test_x509_ext.py
```diff
@@ -4440,6 +4440,229 @@ def test_inhibit_any_policy(self, backend):
...
+        assert idp == expected
+
+    @pytest.mark.parametrize(
+        (
+            "error", "only_contains_user_certs", "only_contains_ca_certs",
+            "indirect_crl", "only_contains_attribute_certs",
+            "only_some_reasons", "full_name", "relative_name"
+        ),
+        [
+            [
```
alex: These should be tuples, not lists

src/cryptography/x509/extensions.py
reaperhulk: Yes, do you think that's more clear than `any`?

alex: I think so. Do you disagree?

reaperhulk: I find it easier to quickly see that something must be truthy when `any` is present, but like most stylistic things I don't hold a strong opinion here.

src/cryptography/x509/extensions.py
reaperhulk: Isn't that abusing the fact that a bool is secretly an int? `True + True + False == 2` is kind of a Python WTF

alex: Yes, it is. If you don't like that, at least drop the `is True` and just make this `if x`. And rename `bools` to something semantically meaningful.

src/cryptography/x509/extensions.py
```diff
@@ -1446,6 +1446,136 @@ def __repr__(self):
...
+            isinstance(indirect_crl, bool) and
+            isinstance(only_contains_attribute_certs, bool)
+        ):
+            raise TypeError(
+                "only_contains_user_certs, only_contains_ca_certs, "
+                "indirect_crl and only_contains_attribute_certs "
+                "must all be boolean."
+            )
+
+        boolean_arguments = [
```
alex: Even more semantic! "crl_constraints" or something

Issue #4440, opened 2018-08-30
Deprecate and remove SECT* curves


alex:
See mailing list

reaperhulk:
Having thought about this some more I wonder if the answer is actually to move these to `cryptography.hazmat.primitives.asymmetric.ec.obscure` so that they're accessible for research purposes (and our hazmat layer definitely is friendly to researchers at this time) but out of the way of people who are looking for curves they should actually use.

One argument for leaving them where they are is that we've already improved this situation by moving the docs for the curves such that the binary curves are at the bottom and are labeled with "These binary curves are discouraged for new systems".

PR #4556, opened 2018-11-03
Add the rest of the curve OIDs, make oid a part of the EC interface

### LINKED ISSUE #3345: Add OID to asymmetric.ec.EllipticCurve
Class asymmetric.ec.EllipticCurve has a "name" that is the symbolic name used for the ASN.1 OID.
For example, OID 1.3.132.0.34 has the name "secp384r1" and the "name" provides this. Some protocols (e.g. XML Signature) need the OID value, which is currently not provided. An additional "oid" that returns this would be useful. The values can be taken from the first two columns of the "children" section in http://oidref.com/1.3.132.0

reaperhulk:
fixes #3345 

We're obviously talking about deprecating the binary curves (#4440), but we let ourselves get gridlocked too easily. This is a thing that makes sense to add, and if we need to add some more OIDs to make it happen immediately then fine. (I have some other thoughts on the SECT curves in the other issue as well)

### DIFF
--- a/docs/hazmat/primitives/asymmetric/ec.rst
+++ b/docs/hazmat/primitives/asymmetric/ec.rst
@@ -529,6 +529,12 @@ Key Interfaces
         Size (in :term:`bits`) of a secret scalar for the curve (as generated
         by :func:`generate_private_key`).
 
+    .. attribute:: oid
+
+        :type: :class:`~cryptography.x509.ObjectIdentifier`
+
+        Returns the object identifier assigned to the curve.
+
 
 .. class:: EllipticCurveSignatureAlgorithm
 
--- a/src/cryptography/hazmat/primitives/asymmetric/ec.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/ec.py
@@ -38,6 +38,12 @@ class EllipticCurveOID(object):
 @six.add_metaclass(abc.ABCMeta)
 class EllipticCurve(object):
     @abc.abstractproperty
+    def oid(self):
+        """
+        The ObjectIdentifier associated with the curve.
+        """
+
+    @abc.abstractproperty
     def name(self):
         """
         The name of the curve. e.g. secp256r1.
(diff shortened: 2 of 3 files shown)


alex:
I'm ok adding the additional constants, but I'm still not comfortable adding the `oid` property on EC.

alex:
@ronf We're looking back into this. How much does the addition of `EllipticCurveOID` solve your problems? Does adding them as attributes add a lot of additional value?

reaperhulk:
@ronf so in the final case those are explicit parameter encoded keys, but they still correspond to a named curve and not some crazy custom curve, correct?

alex:
Thanks for writing this up Ron. I'm going to put the point values and other metadata out of scope for this conversation and just focus on OIDs.

From what you said, it sounds like eliminating that second mapping would be nice, but relatively minor compared to all the other ones you're carrying, is that fair?

ronf:
@reaperhulk That's right. SSH only supports named curves. It does have a way in theory to construct an SSH curve name for an arbitrary ECDH curve OID, but both sides would still need to know the curve parameters corresponding to the curve that an OID represents for it to be usable, whether the key was encoded using a named curve OID or explicit parameters. There's no way to use a custom curve based on just its parameters.

@alex Yes. To give you a feel for it, here's a snippet of how I do the registration today:

```python
for _curve_id, _oid in ((b'nistp521', '1.3.132.0.35'),
                        (b'nistp384', '1.3.132.0.34'),
                        (b'nistp256', '1.2.840.10045.3.1.7'),
                        (b'1.3.132.0.10', '1.3.132.0.10')):
    _algorithm = b'ecdsa-sha2-' + _curve_id
    _cert_algorithm = _algorithm + b'-cert-v01@openssh.com'
    _x509_algorithm = b'x509v3-' + _algorithm

    _oid = ObjectIdentifier(_oid)
    _alg_oids[_curve_id] = _oid
    _alg_oid_map[_oid] = _curve_id

    register_public_key_alg(_algorithm, _ECKey, (_algorithm,))
    register_certificate_alg(1, _algorithm, _cert_algorithm,
                             _ECKey, SSHOpenSSHCertificateV01)
    register_x509_certificate_alg(_x509_algorithm)
```

The table at the top is a list of SSH names for each curve and the corresponding OID, and I construct the full names and set up all the bidirectional mappings I need from that for the first three cases. The conversion to PyCA keys happens elsewhere and is a simple dictionary:

```python
_curves = {b'1.3.132.0.10': (ec.SECP256K1, SHA256),
           b'nistp256':     (ec.SECP256R1, SHA256),
           b'nistp384':     (ec.SECP384R1, SHA384),
           b'nistp521':     (ec.SECP521R1, SHA512)}
```

The real win would be handling the fourth case, but even that's not a lot of code. I just hate duplicating values that OpenSSL must already know. For that, check out https://github.com/ronf/asyncssh/blob/master/asyncssh/crypto/ec_params.py.

reaperhulk:
Okay, so to cover case 3 we could provide a public dictionary that maps OID -> curve class.

ronf:
Yes - that could be useful. I'd need to figure out the hash function to use, but I should be able to do that as long as the curve provides a function to return its key size, which I believe is already there. RFC 5656 mandates the following mapping for ecdh-sha2-* and ecdsa-sha2-* SSH algorithms:

    The Elliptic Curve Digital Signature Algorithm (ECDSA) is specified
    for use with the SSH ECC public key algorithm.

    The hashing algorithm defined by this family of method names is the
    SHA2 family of hashing algorithms [FIPS-180-3].  The algorithm from
    the SHA2 family that will be used is chosen based on the size of the
    named curve specified in the public key:

                    +----------------+----------------+
                    |   Curve Size   | Hash Algorithm |
                    +----------------+----------------+
                    |    b <= 256    |     SHA-256    |
                    |                |                |
                    | 256 < b <= 384 |     SHA-384    |
                    |                |                |
                    |     384 < b    |     SHA-512    |
                    +----------------+----------------+

With that, I could in theory support any OID in PyCA's mapping, as the RFC states:

    The concatenation of any so encoded ASN.1 OID specifying a set of
    elliptic curve domain parameters with "ecdh-sha2-" is implicitly
    registered under this specification.

I would still want to continue to support the standard algorithm names for the three NIST curves listed in the RFC, but I could dynamically add other entries for any EC OID supported by PyCA.

ronf:
Thanks, I appreciate it!

reaperhulk:
closing in favor of #4759

PR #4573, opened 2018-11-12
add sha3 support

reaperhulk:
refs #4197

### DIFF
--- a/docs/hazmat/primitives/cryptographic-hashes.rst
+++ b/docs/hazmat/primitives/cryptographic-hashes.rst
@@ -151,6 +151,12 @@ SHA-family of hashes.
 SHA-3 family
 ~~~~~~~~~~~~
 
+SHA-3 is the most recent NIST secure hash algorithm standard. Despite the
+larger number SHA-3 is not considered to be better than SHA-2. Instead, it uses
+a significantly different internal structure so that **if** an attack appears
+against SHA-2 it is unlikely to apply to SHA-3. SHA-3 is significantly slower
+than SHA-2 so at this time most users should choose SHA-2.
+
 .. class:: SHA3_224()
 
     .. versionadded:: 2.5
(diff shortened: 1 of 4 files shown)


src/cryptography/hazmat/primitives/hashes.py
```diff
@@ -136,6 +136,34 @@ class SHA512(object):
     block_size = 128
 
 
+@utils.register_interface(HashAlgorithm)
+class SHA3_224(object):  # noqa: N801
+    name = "sha3-224"
+    digest_size = 28
+    block_size = None
```
reaperhulk: block_size isn't really a relevant value for a sponge function. It has a state size, but not a block.

docs/hazmat/primitives/cryptographic-hashes.rst
```diff
@@ -148,6 +148,37 @@ SHA-family of hashes.
 
     :raises ValueError: If the ``digest_size`` is invalid.
 
+SHA-3 family
+~~~~~~~~~~~~
+
```
alex: A couple of prose sentences about what the hell a SHA3 is would be good. it's secure, but slow, it's a new thing, probably you should just use it for compat where requierd, that sort of thing. with citations :-)

src/cryptography/hazmat/primitives/hashes.py
alex: We removed block_size from the interface I thought? This shouldn't be a problem

docs/hazmat/primitives/cryptographic-hashes.rst
reaperhulk: SHA-3 is the most recent NIST secure hash algorithm standard. Despite the 
larger number SHA-3 is not considered to be better than SHA-2. Instead, it uses
a significantly different internal structure so that **if** an attack appears 
against SHA-2 it is unlikely to apply to SHA-3. SHA-3 is significantly slower 
than SHA-2 so at this time most users should choose SHA-2.

PR #4577, opened 2018-11-12
x448 bindings

reaperhulk:
refs #4215

### DIFF
--- a/src/_cffi_src/openssl/evp.py
+++ b/src/_cffi_src/openssl/evp.py
@@ -22,6 +22,7 @@ static const int EVP_PKEY_DH;
 static const int EVP_PKEY_DHX;
 static const int EVP_PKEY_EC;
 static const int EVP_PKEY_X25519;
+static const int EVP_PKEY_X448;
 static const int EVP_MAX_MD_SIZE;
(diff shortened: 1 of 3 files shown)


src/_cffi_src/openssl/evp.py
```diff
@@ -272,4 +273,11 @@
 #ifndef EVP_PKEY_X25519
 #define EVP_PKEY_X25519 NID_X25519
 #endif
+
+/* This is tied to X448 support so we reuse the Cryptography_HAS_X448
+   conditional to remove it. OpenSSL 1.1.1 adds this define.  We can remove
+   this in the distant future when we drop 1.1.0 support. */
+#ifndef EVP_PKEY_X448
+#define EVP_PKEY_X448 NID_X448
```
alex: This feels like gibberish, why do we define it to `NID_X448`, instead of `0`?

reaperhulk: This is what we did for X25519 because that's what OpenSSL actually defines the value to be. It just so happens that it simplifies our removal code because we define NID_X448 as well if it doesn't exist. We can obviously choose to do explicitly do 0 (although that it is what it will be if we define NID_X448) as well.

PR #4592, opened 2018-11-15
Raise MemoryError when backend.derive_scrypt can't malloc enough

### LINKED ISSUE #4591: Raise MemoryError when Scrypt can't malloc enough memory
When someone makes a Scrypt derivation call and the system doesn't have enough memory to `malloc`, it raises quite a noisey OpenSSL `InternalError`.

```
InternalError: Unknown OpenSSL error. This error is commonly encountered when another library is not cleaning up the OpenSSL error stack. If you are using cryptography with another library that uses OpenSSL try disabling it before reporting a bug. Otherwise please file an issue at https://github.com/pyca/cryptography/issues with information on how to reproduce this. ([_OpenSSLErrorWithText(code=101404737, lib=6, func=181, reason=65, reason_text=b'error:060B5041:digital envelope routines:EVP_PBE_scrypt:malloc failure')])
(14 additional frame(s) were not displayed)
...
  File "nucypher/config/keyring.py", line 191, in _derive_key_material_from_passphrase
    ).derive(passphrase.encode())
  File "cryptography/hazmat/primitives/kdf/scrypt.py", line 60, in derive
    key_material, self._salt, self._length, self._n, self._r, self._p
  File "cryptography/hazmat/backends/openssl/backend.py", line 1950, in derive_scrypt
    self.openssl_assert(res == 1)
  File "cryptography/hazmat/backends/openssl/backend.py", line 105, in openssl_assert
    return binding._openssl_assert(self._lib, ok)
  File "cryptography/hazmat/bindings/openssl/binding.py", line 76, in _openssl_assert
    errors_with_text
```

Would it make sense to catch that and raise `MemoryError`, instead?

tuxxy:
### What this does:
1. Catches when an OpenSSL error with a reason of `ERR_R_MALLOC_FAILURE` and raises a `MemoryError` with helpful info instead.
2. Adds a test for `test_scrypt_malloc_failure`.
3. Resolves #4591

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -2088,13 +2088,18 @@ class Backend(object):
         )
         if res != 1:
             errors = self._consume_errors()
-            if errors[0].reason == 65:
-                min_memory = 128 * n * r // (10**6)
-                raise MemoryError(
-                    "Not enough memory to derive key. These parameters require"
-                    " {} MB of memory.".format(min_memory)
+            self.openssl_assert(
+                errors[0]._lib_reason_match(
+                    self._lib.ERR_LIB_EVP,
+                    self._lib.ERR_R_MALLOC_FAILURE
                 )
-            self.openssl_assert(res == 1)
+            )
+
+            min_memory = 128 * n * r // (1024**2)
+            raise MemoryError(
+                "Not enough memory to derive key. These parameters require"
+                " {} MB of memory.".format(min_memory)
+            )
         return self._ffi.buffer(buf)[:]
 
     def aead_cipher_supported(self, cipher):
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -2097,6 +2097,8 @@ class Backend(object):
                     )
                 )
 
+            # memory required formula explained here:
+            # https://blog.filippo.io/the-scrypt-parameters/
             min_memory = 128 * n * r // (1024**2)
             raise MemoryError(
                 "Not enough memory to derive key. These parameters require"
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -2086,7 +2086,15 @@ class Backend(object):
             key_material, len(key_material), salt, len(salt), n, r, p,
             scrypt._MEM_LIMIT, buf, length
         )
-        self.openssl_assert(res == 1)
+        if res != 1:
+            errors = self._consume_errors()
+            if errors[0].reason == 65:
+                min_memory = 128 * n * r // (10**6)
+                raise MemoryError(
+                    "Not enough memory to derive key. These parameters require"
+                    " {} MB of memory.".format(min_memory)
+                )
+            self.openssl_assert(res == 1)
         return self._ffi.buffer(buf)[:]
 
     def aead_cipher_supported(self, cipher):
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -2088,12 +2088,14 @@ class Backend(object):
         )
         if res != 1:
             errors = self._consume_errors()
-            self.openssl_assert(
-                errors[0]._lib_reason_match(
-                    self._lib.ERR_LIB_EVP,
-                    self._lib.ERR_R_MALLOC_FAILURE
+            if not self._lib.CRYPTOGRAPHY_OPENSSL_LESS_THAN_111:
+                # This error is only added to the stack in 1.1.1+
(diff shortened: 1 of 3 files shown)


src/cryptography/hazmat/primitives/kdf/scrypt.py
```diff
@@ -56,9 +56,20 @@ def derive(self, key_material):
...
-        return self._backend.derive_scrypt(
-            key_material, self._salt, self._length, self._n, self._r, self._p
-        )
+
+        try:
+            return self._backend.derive_scrypt(
+                key_material, self._salt, self._length,
+                self._n, self._r, self._p
+            )
+        except InternalError as e:
```
alex: This is being handled in the wrong place, we should handle this inside `derive_scrypt` instead of raising an `InternalError.`

tuxxy: Ah, I see. Thanks!

tuxxy:
@alex Moved this to `backend.derive_scrypt`. I saw other methods using `errors[0]._lib_reason_match`, but unfortunately I'm hitting a limit of my OpenSSL knowledge in how to check that it can't malloc enough. Is that the preferred way, or is the way I've done it good enough?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2086,7 +2086,15 @@ def derive_scrypt(self, key_material, salt, length, n, r, p):
             key_material, len(key_material), salt, len(salt), n, r, p,
             scrypt._MEM_LIMIT, buf, length
         )
-        self.openssl_assert(res == 1)
+        if res != 1:
+            errors = self._consume_errors()
+            if errors[0].reason == 65:
```
alex: You'll need to find out what name this constant has and add it to the bindings

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2086,7 +2086,15 @@ def derive_scrypt(self, key_material, salt, length, n, r, p):
...
+                min_memory = 128 * n * r // (10**6)
+                raise MemoryError(
+                    "Not enough memory to derive key. These parameters require"
+                    " {} MB of memory.".format(min_memory)
+                )
+            self.openssl_assert(res == 1)
```
alex: This should always fail, you're in ` res != ` block

tuxxy:
I changed the approach a bit to match what other methods were doing. Seems a bit more elegant after exposing the `ERR_R_MALLOC_FAILURE`.

alex:
Looks like there'a flake8 issue, and some of the builders (specific openssl versions?) aren't happy with it.

tuxxy:
Fixed the PEP8 problem, but I'm unsure on these other errors. Attempting to figure out now...

reaperhulk:
A quick glance at the stretch builders seems to indicate that failure isn't pushing errors on the stack. The macOS failures are likely because those machines have virtual memory so it's just thrashing them rather than actually failing on malloc. The errors on windows just show the process suddenly stopping, which could maybe be Windows handling out of memory in a different fashion?

reaperhulk:
Raising the block_size to 589824 will cause allocation to be >64TB (which should cause a malloc failure on any platform using 48-bit addressing). Of course, that might overflow and cause a different error on 32-bit platforms, so ¯\_(ツ)_/¯

tuxxy:
I've reduced the work factor in the test to see if that has any effects. Unfortunately, I don't have any 32-bit systems to test this on.

reaperhulk:
The memory handling of scrypt changed a great deal between 1.1.0 and 1.1.1 so I think you're going to need to carefully read EVP_pbe_scrypt to determine what can and can't raise errors. For example, 1.1.0 has no malloc error on stack so that assertion will always fail on stretch.

We also ideally don't want to make the allocation smaller because as I stated before we need alloc to fail on the macOS builders. Those builders have virtual memory so the numbers you're currently using are not causing alloc failure. Instead it's thrashing the disk so hard the machines become unusable.

If you use the values I stated then it might properly fail to alloc on all our test infra, but I'm not sure what 1.1.1 will do with it (the macOS and windows builders are not 1.1.1 yet, but will be converted soon).

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2086,7 +2086,22 @@ def derive_scrypt(self, key_material, salt, length, n, r, p):
...
+            if not self._lib.CRYPTOGRAPHY_OPENSSL_LESS_THAN_111:
+                # This error is only added to the stack in 1.1.1+
+                self.openssl_assert(
+                    errors[0]._lib_reason_match(
+                        self._lib.ERR_LIB_EVP,
+                        self._lib.ERR_R_MALLOC_FAILURE
+                    )
+                )
+
+            min_memory = 128 * n * r // (1024**2)
```
alex: Please add a comment with a cite to this formula; hopefully there's a clear reference?

PR #4593, opened 2018-11-18
Adjust DTLS check

neheb:
OpenSSL defines these even with OPENSSL_NO_DTLS.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -714,7 +714,7 @@ static const long TLS_ST_BEFORE = 0;
 static const long TLS_ST_OK = 0;
 #endif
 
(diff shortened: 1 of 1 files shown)


reaperhulk:
We don't test against NO_DTLS in CI so have you tested this against an OpenSSL built in this fashion? Also, what does a function like `DTLS_set_link_mtu` do when NO_DTLS is set?

neheb:
Yes. In OpenWrt we currently patch out the DTLS entries as OpenSSL is compiled without DTLS support by default.

The same thing. OPENSSL_NO_DTLS is only for the applications, not for the library. Checking for it as done here results in redefinition errors that break the build.

See: https://github.com/openwrt/packages/blob/master/lang/python/python-cryptography/patches/002-remove-undefined-dtls-methods.patch

reaperhulk:
Okay, this appears to be a bug we should fix then. 👍 from me

PR #4606, opened 2018-11-23
refactor serialization module into package

reaperhulk:


alex:
We ought to move the SSH stuff to it's own file, but that can be a follow up commit.

PR #4553, opened 2018-11-03
PKCS12 Basic Parsing

reaperhulk:
Add support for parsing a basic PKCS12. This assumes one cert and one key and no CA certificates (which matches Go's pkcs12 interface, but is far less than the complete spec). I suspect this is going to cover almost every use case for parsing so we'll want to wait for user requests before we add anything more.

PR #4612, opened 2018-11-26
PKCS12 interface

reaperhulk:
refs #4553

### DIFF
--- a/src/cryptography/hazmat/primitives/serialization/pkcs12.py
+++ b/src/cryptography/hazmat/primitives/serialization/pkcs12.py
@@ -11,6 +11,7 @@ import six
 
 @six.add_metaclass(abc.ABCMeta)
 class PKCS12(object):
+    @abc.abstractmethod
     def __init__(self, certificate, private_key, additional_certificates):
         """
         Create a PKCS12 object with a certificate, private key, and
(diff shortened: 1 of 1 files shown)


reaperhulk:
I suppose this fails coverage because nothing imports it. Well, we can still discuss.

src/cryptography/hazmat/primitives/serialization/pkcs12.py
```diff
@@ -0,0 +1,37 @@
...
+from __future__ import absolute_import, division, print_function
+
+import abc
+
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class PKCS12(object):
+    @abc.abstractmethod
```
alex: I don't think this should be on the interface

reaperhulk:
I haven't been able to come up with a great solution. The PKCS12 spec is just a pile of bags. These getters are a subset (e.g. `private_key` is really Union(KeyBag, PKCS8ShroudedKeyBag), `certificate` is a CertBag that matches the private key, and the additional certificates are any other CertBag), so it's possible we could just evolve the interface in the future we choose to support more.

Do we know of anything that actually implements a significantly larger chunk of this spec? Go supports just cert + key for reference.

reaperhulk:
(We're going to go with a function returning a tuple to avoid creating an interface we don't need yet)

PR #4614, opened 2018-11-29
fix certbot downstream, show commit hash for all downstreams

reaperhulk:


.travis/downstream.d/aws-encryption-sdk.sh
```diff
@@ -3,6 +3,7 @@
 case "${1}" in
     install)
         git clone --depth=1 https://github.com/awslabs/aws-encryption-sdk-python
+        git rev-parse HEAD
```
alex: gotta go below the cd

PR #4616, opened 2018-12-01
allow multi-valued RDNs

reaperhulk:
RDNs can have multiple values. This allows them in FreshestCRL and upcoming IssuingDistributionPoint encoding support.


reaperhulk:
The previous behavior here was raising an InternalError because we wrongly asserted that the result would always be 1

Issue #4409, opened 2018-08-20
Allow key material to use bytearray instead of bytes


earonesty:
Currently, this library requires key material to be "immutable" `bytes` type for most functions.  This means that it is, by design, difficult to clear the key material from RAM after it is used.  memset can be used on immutable `bytes` sequences to clear it... but this is fraught with potential issues.

It would be both fairly trivial and a big improvement for all private and symmetric keys to be the `bytearray` type.   Currently keys are left in RAM.   Using a `bytearray` allows keys to be cleaned up, removed from caches, etc.

reaperhulk:
Sorry for the slow response. We agree that allowing bytearrays as well as bytestrings is the right move. There are a lot of call sites that check for bytes (or six.binary_type) so we'll need to update all of those and decide if we want to add tests that use bytearrays in at least some places.

reaperhulk:
Just as an update here, in addition to the byte checks we need to fix we'll also need to call `ffi.from_buffer` on the bytearrays before passing them to the C layer. As of cffi 1.8 you can also call that function on byte strings, which would greatly simplify supporting this, although we'd need to move our minimum pypy to 5.4 for that.

Since

reaperhulk:
and as of #4621 we allow this

PR #4621, opened 2018-12-02
allow bytes-like for key/iv/data for symmetric encryption

reaperhulk:
* [x] Update and centralize our bytes check so that it also permits bytearray/memoryview.
* [x] Update calls through cffi to use `from_buffer`. We can call this unconditionally as of cffi 1.8. This will require us increasing our minimum PyPy version to 5.4 from 5.3 (5.3 did not allow the call on `bytes`). 5.4 was released over 2 years ago so I'm fine with that.
* [x] Update docs to state that bytes-like is allowable.

A future PR can do this for asymmetric keys as well.

refs #4409

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/ciphers.py
+++ b/src/cryptography/hazmat/backends/openssl/ciphers.py
@@ -219,7 +219,8 @@ class _CipherContext(object):
     def authenticate_additional_data(self, data):
         outlen = self._backend._ffi.new("int *")
         res = self._backend._lib.EVP_CipherUpdate(
-            self._ctx, self._backend._ffi.NULL, outlen, data, len(data)
+            self._ctx, self._backend._ffi.NULL, outlen,
+            self._backend._ffi.from_buffer(data), len(data)
         )
         self._backend.openssl_assert(res != 0)
 
--- a/src/cryptography/hazmat/backends/openssl/ciphers.py
+++ b/src/cryptography/hazmat/backends/openssl/ciphers.py
@@ -56,13 +56,15 @@ class _CipherContext(object):
             )
 
         if isinstance(mode, modes.ModeWithInitializationVector):
-            iv_nonce = mode.initialization_vector
+            iv_nonce = self._backend._ffi.from_buffer(
+                mode.initialization_vector
+            )
         elif isinstance(mode, modes.ModeWithTweak):
-            iv_nonce = mode.tweak
+            iv_nonce = self._backend._ffi.from_buffer(mode.tweak)
         elif isinstance(mode, modes.ModeWithNonce):
-            iv_nonce = mode.nonce
+            iv_nonce = self._backend._ffi.from_buffer(mode.nonce)
         elif isinstance(cipher, modes.ModeWithNonce):
-            iv_nonce = cipher.nonce
+            iv_nonce = self._backend._ffi.from_buffer(cipher.nonce)
         else:
             iv_nonce = self._backend._ffi.NULL
         # begin init with cipher and operation type
@@ -105,7 +107,7 @@ class _CipherContext(object):
             ctx,
             self._backend._ffi.NULL,
             self._backend._ffi.NULL,
-            cipher.key,
+            self._backend._ffi.from_buffer(cipher.key),
             iv_nonce,
             operation
         )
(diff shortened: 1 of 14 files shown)


h-vetinari:
@reaperhulk Guess I am a bit late, but I had wanted to write the following after reading
> Figure out a sane way to test both bytes and bytearray instantiation without doubling our tests.

which you have now edited out of the OP already... In any case (since I wrote the answer already before realizing): this sounds like a prime example of using fixtures: https://docs.pytest.org/en/latest/fixture.html

You could add the following in the toplevel `conftest.py`:
```python
@pytest.fixture(params=[bytes, bytearray])
def bytes_klazz(request):
    """
    Fixture for the allowed classes of bytes.

    Returns one of:
    * bytes
    * bytearray
    """
    return request.param
```

Then, this object is available to every test that pytest collects, which will then be run twice for all different values of the fixture (this becomes a full cartesian product if there are more fixtures).

In most cases, I presume the type of the bytes_class should be preserved (and since `bytes(b'1') == bytearray(b'1')`), so you should maybe have a util-method along the following lines
```python
def assert_bytez_equal(left, right):
    assert type(left) == type(right)
    assert left == right
```

Taking a random example from the tests, this could then be used as follows:
```diff
+ from util import assert_bytez_equal

+    def test_timestamp_ignored_no_ttl(self, monkeypatch, backend, bytes_klazz):
+        f = Fernet(base64.urlsafe_b64encode(bytes_klazz(b"\x00" * 32)), backend=backend)
+        pt = bytes_klazz(b"encrypt me")
-    def test_timestamp_ignored_no_ttl(self, monkeypatch, backend):
-        f = Fernet(base64.urlsafe_b64encode(b"\x00" * 32), backend=backend)
-        pt = b"encrypt me"
         token = f.encrypt(pt)
         ts = "1985-10-26T01:20:01-07:00"
         current_time = calendar.timegm(iso8601.parse_date(ts).utctimetuple())
         monkeypatch.setattr(time, "time", lambda: current_time)
+        assert_bytez_equal(f.decrypt(token, ttl=None), pt)
-        assert f.decrypt(token, ttl=None) == pt
```

h-vetinari:
@reaperhulk 
> [...] but the logical application of them (run every test both ways!) results in our test suite bloating significantly, which causes issues both for memory consumption and time of execution for the tests.

You don't have to use it everywhere - it's only injected where you add `bytes_klazz` to the signature of the test-function. So you can use it as sparingly as you like, but I'd rather err on the side of more coverage...

src/cryptography/utils.py
```diff
@@ -30,6 +30,10 @@ def _check_bytes(name, value):
         raise TypeError("{0} must be bytes".format(name))
 
 
+def _check_byteslike(name, value):
+    memoryview(value)
```
alex: Hmm, so this results in a worse error message than previously, do we care?

src/cryptography/hazmat/backends/openssl/ciphers.py
```diff
@@ -131,8 +133,10 @@ def update_into(self, data, buf):
             "unsigned char *", self._backend._ffi.from_buffer(buf)
         )
         outlen = self._backend._ffi.new("int *")
-        res = self._backend._lib.EVP_CipherUpdate(self._ctx, buf, outlen,
-                                                  data, len(data))
+        res = self._backend._lib.EVP_CipherUpdate(
+            self._ctx, buf, outlen,
+            self._backend._ffi.from_buffer(data), len(data)
```
alex: I'm sure I'll regret asking, but what happens if you pass the same buffer for `data` and `buf`?

reaperhulk: It will fail with a ValueError because we force buffer size to be `block_size - 1` larger than the data passed so that we can always write everything back to the supplied buffer. If we were to remove that restriction (which I am not proposing we do since it exists to prevent segfaults in cases where users are not careful about block alignment) then it would actually properly do an in place update from plaintext to ciphertext (and vice versa on decrypt).

src/cryptography/utils.py
reaperhulk: I think I'm okay with it. That said, we could do:

```python
try:
    memoryview(value)
except TypeError:
    raise TypeError("{0} must be bytes-like.".format(name))
```

tests/hazmat/primitives/test_ciphers.py
```diff
@@ -184,7 +184,7 @@ def test_invalid_key_size(self):
             SEED(b"\x00" * 17)
 
     def test_invalid_key_type(self):
-        with pytest.raises(TypeError, match="key must be bytes"):
```
alex: Can all of these be put back?

reaperhulk: Yep, fixed.

PR #4624, opened 2018-12-05
remove idna as a primary dependency

reaperhulk:
An idea for moving the idna deprecation towards conclusion. Requires another travis job (bleh), but hopefully only for a year or so?

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -99,6 +99,18 @@ support multiple versions of Python. The Python 3.4 ``abi3`` wheel can be used
 with any version of Python greater than or equal to 3.4. Recent versions of
 ``pip`` will automatically install ``abi3`` wheels.
 
+``ModuleNotFoundError``: No module named ``'idna'``
+---------------------------------------------------
+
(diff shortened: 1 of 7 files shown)


tox.ini
```diff
@@ -19,6 +19,12 @@ commands =
     coverage combine
     coverage report -m
 
+[testenv:idna]
```
alex: This should not be a stand-alone `idna` testenv, it should be a modified on the existing pyXX testenvs. See pyOpenSSL's for an example of that syntax

docs/faq.rst
```diff
@@ -99,6 +99,18 @@ support multiple versions of Python. The Python 3.4 ``abi3`` wheel can be used
 with any version of Python greater than or equal to 3.4. Recent versions of
 ``pip`` will automatically install ``abi3`` wheels.
 
+``ModuleNotFoundError``: No module named ``'idna'``
```
alex: Instead of, or in addition to, this FAQ, please make the lazy `import idna` raise a better error message on `ImportError`

PR #4626, opened 2018-12-06
temporarily allow dynamo to fail

reaperhulk:
@mattsb42-aws this doesn't really look like it's your fault but could you do us a favor and take a look at the failure here? Example: https://travis-ci.org/pyca/cryptography/jobs/464137535

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -134,6 +134,8 @@ matrix:
           # urllib3 tests fail on xenial for whatever reason
           dist: trusty
           sudo: false
+    allow_failures:
+      - env: DOWNSTREAM=dynamodb-encryption-sdk OPENSSL=1.1.0j
 
 install:
     - ./.travis/install.sh


mattsb42-aws:
Taking a look. That is strange...that shouldn't be able to happen with the frozen requirements...

reaperhulk:
I investigated a bit yesterday and couldn't even replicate it locally so I have no idea what's going on.

mattsb42-aws:
Confirmed this is not an ephemeral Travis issue, as it is happening on my fork as well. Looking to see how the environment being built is different than the environment where we test to make sure that this dependency set works properly[1] (they *should* be identical except for the `cryptography` build and any of its new dependencies).

[1] https://travis-ci.org/aws/aws-dynamodb-encryption-python/jobs/464195089

mattsb42-aws:
Testing with identical environments changed nothing. The only changes in installed dependencies are unrelated.

It looks like this might be an old issue that may have resurfaced recently[1]. Testing this in my fork to see if the `BOTO_CONFIG` hack fixes it.

...still no idea why this isn't failing anywhere else...

[1] https://github.com/boto/boto/issues/3717

mattsb42-aws:
Yup, that was it...that is annoying. Fortunately, it's an easy fix.

I vaguely remember seeing this happen before somewhere. I suspect that the difference might be that our tests run in a virtualenv and yours do not. ...I'll think on what we can do to replicate that.

reaperhulk:
We do run in a virtualenv but critically we do *not* use tox. And tox does env variable isolation unless you whitelist, so if the issue is that travis is setting `BOTO_CONFIG` then that variable wouldn't make it into the execution env inside tox.

reaperhulk:
Closing this since we have a real solution thanks to Matt!

PR #4627, opened 2018-12-07
Fix dynamodb-encryption-sdk downstream tests

mattsb42-aws:
Fix as discussed in #4626.

I also went ahead and normalized its build back to xenial, since it seems to be working just fine. I'm guessing that was modified trying to fix this issue?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -122,7 +122,6 @@ matrix:
           env: DOWNSTREAM=aws-encryption-sdk
         - python: 2.7
           env: DOWNSTREAM=dynamodb-encryption-sdk OPENSSL=1.1.0j BOTO_CONFIG=/dev/null
-          sudo: false
         - python: 2.7
           env: DOWNSTREAM=certbot OPENSSL=1.1.0j
         - python: 2.7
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -121,9 +121,7 @@ matrix:
         - python: 2.7
           env: DOWNSTREAM=aws-encryption-sdk
         - python: 2.7
-          env: DOWNSTREAM=dynamodb-encryption-sdk OPENSSL=1.1.0j
-          # dynamodb-encryption-sdk tests fail on xenial for whatever reason
-          dist: trusty
+          env: DOWNSTREAM=dynamodb-encryption-sdk OPENSSL=1.1.0j BOTO_CONFIG=/dev/null
           sudo: false
```
reaperhulk: You can remove this line now too

reaperhulk:
Thanks matt. Having it search dist-packages outside of the venv is some interesting behavior.

.travis.yml
mattsb42-aws: Good call. I left it in because https://github.com/boto/boto/issues/3717 seemed to indicate that it was also needed, but testing on my fork confirms it's fine without it.

alex:
Merged via that PR. Thanks.

PR #4628, opened 2018-12-08
test on latest pypy

reaperhulk:
this should work on xenial


reaperhulk:
pypy2.7-6.0 hangs on travis xenial builders, but not locally. Could be any number of issues, none of which I'm interested in chasing down right now soooo no upgrade for now!

PR #4636, opened 2018-12-11
convert some asserts to function calls

reaperhulk:


alex:
Coverage numbers are transparently a bug.

PR #4640, opened 2018-12-12
deprecate old from_encoded_point

reaperhulk:


ronf:
If from_encoded_point() is moving from EllipticCurvePublicNumbers to EllipticCurvePublicKey, shouldn't that also be done for encode_point()? I need both of these functions in AsyncSSH, as SSH serializes the EC public key as a "point" byte string when doing things like ECDH.

reaperhulk:
That work is in https://github.com/pyca/cryptography/pull/4638

ronf:
I see. Thanks.

After sleeping on this, I was going to suggest that eliminate these two methods instead of moving them and replace them with public_bytes()/from_public_bytes() to a new serialization encoding, but I'll add my comment on that over in #4638.

reaperhulk:
@ronf #4638 is almost certainly going to become a new serialization in `public_bytes` when I have time to finish the PR. You can roughly see the shape of what is likely to happen by looking at the serialization work in https://github.com/pyca/cryptography/pull/4653.

Do you envision `from_public_bytes` being capable of loading PKCS8 keys as well? We discussed that in #4653 but have been leaning away from it for no particularly strong reason.

reaperhulk:
Oops I see you responded on the other issue already! I'll go over there

ronf:
For X25519/X448 DH, I haven't found a need for any encoding other then the raw values or the conversion to/from compressed & uncompressed point values for the other EC curves. However, for Ed25519/Ed448, I think there would be value in support other forms of serialization. Specifically, OpenSSH supports Ed25519 private and public keys in its "new" key format. It doesn't yet support PKCS8, but if OpenSSL supports that for X25519/X448 keys when doing a TLS handshake, I think it would be worth supporting that serialization here.

I will probably make both the OpenSSH format and PKCS8 (DER and PEM) format available in AsyncSSH for these key types once I find a spec for what PKCS8 encoding of these key types looks like.

PR #4642, opened 2018-12-15
allow paramiko to fail temporarily while its test suite is broken

reaperhulk:
This can be reverted when https://github.com/paramiko/paramiko/pull/1354 is merged (or anything that fixes the paramiko CI)


reaperhulk:
Jeff already merged a fix so this is unnecessary!

PR #4649, opened 2018-12-18
handle empty byte string in from_encoded_point

reaperhulk:
Previously raised an index out of range error since we try to check the first byte to learn the encoding type.

### DIFF
--- a/src/cryptography/hazmat/primitives/asymmetric/ec.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/ec.py
@@ -155,6 +155,9 @@ class EllipticCurvePublicKey(object):
     @classmethod
     def from_encoded_point(cls, curve, data):
         utils._check_bytes("data", data)
+        if len(data) == 0:
+            raise ValueError("data must not be an empty byte string")
+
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/primitives/asymmetric/ec.py
```diff
@@ -155,6 +155,9 @@ def verify(self, signature, data, signature_algorithm):
     @classmethod
     def from_encoded_point(cls, curve, data):
         utils._check_bytes("data", data)
+        if len(data) == 0:
+            raise ValueError("data must not be an empty byte string")
```
alex: Nit, but can you put this check after the `curve` one?

reaperhulk: I can move it, but why? It seems more logical to group "is this bytes" and "is it non-empty" together to me.

alex: Because if you pass a non-curve and empty bytes, I want you to get the type error.

PR #4651, opened 2018-12-20
WIP: Add helper method to verify a certificate signature against an issuer

intgr:
Just a proof of concept for now, how would you feel about adding this kind functionality to cryptography?

Doing the verification and handling different key types and padding is nontrivial, would be great if application authors didn't have to worry about this stuff.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/x509.py
+++ b/src/cryptography/hazmat/backends/openssl/x509.py
@@ -9,7 +9,7 @@ import operator
 import warnings
 
 from cryptography import utils, x509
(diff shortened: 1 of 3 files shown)


reaperhulk:
I think a helper like this has a high likelihood of being misused by people who don't realize that verifying a signature is insufficient to trust a certificate. I'd prefer we finish and merge https://github.com/alex/x509-validator (but obviously that's a much larger job, which is why it's not done)

reaperhulk:
Closing since we're not likely to merge this. Anyone who's interested here should consider assisting on the x509-validator.

Issue #4497, opened 2018-10-10
Rethink public_bytes/private_bytes APIs


reaperhulk:
Right now our public_bytes and private_bytes APIs look like:

`public_bytes(encoding, format) -> bytes`
`private_bytes(encoding, format, encryption) -> bytes`

Where `encoding` is the `Encoding` enum and `format` is either the `PrivateFormat` or `PublicFormat` enum depending on key type. This works, but when adding a new format (e.g. `raw` for x25519/ed25519 keys) it's problematic to update all call sites.

As discussed in IRC we're probably going to migrate to this API:

`public_bytes(encode_format) -> bytes`
`private_bytes(encode_format, encryption) -> bytes`

where `encode_format` is a composition of encoding and format. So where before you might do

```
private_key.private_bytes(Encoding.PEM, PrivateFormat.PKCS8, NoEncryption())
```

in the new API this would become

```
private_key.private_bytes(PEM(PrivateFormat.PKCS8), NoEncryption())
```

Similarly for X25519 it would look like:

```
private_key.private_bytes(Raw(), NoEncryption())
# or
private_key.private_bytes(DER(PrivateFormat.PKCS8), BestAvailableEncryption())
```
## Migration

Since we already have an existing API we need a migration and deprecation plan. For `public_bytes` we could do `public_bytes(encoding=None, format=None)`, do an `isinstance` check on `encoding` to see if it's actually an instance of one of our new compositional classes and raise a deprecation warning if it's not.

If we want to use that approach for `private_bytes` we'd need to make it `private_bytes(encoding=None, format=None, encryption=None)`, which would make composition style calls look pretty ugly during the transition with a superfluous `None` in the middle:

```
private_key.private_bytes(PEM(PrivateFormat.PKCS8), None, NoEncryption())
```

We could play games with args such that we check to see if the second arg is actually an instance of an encryption provider, which would solve the extra None, but will give us quite a bit of complexity in these methods while the deprecation period is in effect.

Finally, these are widely used methods and a deprecation window for this is likely no less than ~3 years, maybe longer? Thoughts?

reaperhulk:
The existing API is not ideal but blowing it up doesn't give us enough advantages to be worth the pain. API stability is also a feature, so we'll choose stability here.

Issue #4386, opened 2018-08-07
[Feature request] Allow to import/export a X25519 private key


simo5:
X25519 private keys can be generated but cannot currently be exported.
This prevents passing them between process or saving to disk for whatever reason.

simo5:
Also allow to import it [edited subject to reflect]

evilaliv3:
This would be really useful also in relation to ECIES encryption

Many libraries like libsodium and opennpgp.js enable this, and many software like signal/whatapp already uses constructions where x25519 is adopted for encryption with preference over NIST curves.

It would be really important for many interoperability use cases to expose:
- private_bytes()
- encoding in PKCS8

ronf:
Curve25519 is also used in SSH, for both an ECDH key exchange and as an asymmetric encryption/signing mechanism via Ed25519. I haven’t had a chance to look closely at the implementation in Cryptography yet, but here are the primitives I'm currently using from the libnacl/libsodium implementation to implement Curve25519 DH:

crypto_scalarmult_curve25519_bytes: Return the size of a Curve25519 public key (point value)
crypto_scalarmult_curve25519_scalarbytes: Return the size of a Curve25519 private key (scalar value)
crypto_scalarmult_curve25519: Compute a shared key from a private key and a peer public key
crypto_scalarmult_curve25519_base: Compute the public key from a private key

The shim I put around this starts by generating a random private key and has a get_public() call to return the public key to share with the peer and a get_shared(peer_public) call that takes a peer public key and returns a shared key. You can see the code for this at https://github.com/ronf/asyncssh/blob/master/asyncssh/crypto/curve25519.py.

In addition, I use the following from libnacl/libsodium to implement Ed25519:

crypto_sign_keypair: Generate a new key pair
crypto_sign: Sign a message with a private key
crypto_sign_open: Verify a message signature with a public key

Here, crypto_sign_keypair() is used to generate a new key pair, crypto_sign(msg, private) is used in the sign operation, and crypto_sign_open(sig + data, public) is used in the verify operation.

This is a bit more involved since there's key serialization to deal with in various formats in addition to the sign & verify operations. This code can be found at https://github.com/ronf/asyncssh/blob/master/asyncssh/ed25519.py.

[...]

PR #4653, opened 2018-12-22
support x448 public/private serialization both raw and pkcs8

reaperhulk:
I'm putting this up so we can discuss the API and move forward on this.

This approach adds `Encoding.Raw` to the `serialization.Encoding` enum. This allows us to retain the method signatures we've used on every other asymmetric key type: `public_bytes(encoding, format)` and `private_bytes(encoding, format, encryption_algorithm)`. However, with this API `format` has no meaning with a raw encoding so `None` must be passed. Fortunately, the existing methods will already reject raw as a valid encoding (but tests need to be added to make sure we don't regress in the future)

I think this approach represents a reasonable compromise to extend our existing API without completely blowing up everything and introducing another multi-year deprecation (https://github.com/pyca/cryptography/issues/4497).

There are a few outstanding questions here though:

* Do we want `from_public_bytes` and `from_private_bytes` to be able to handle x448 pkcs8 or just raw?
* Do we want `load_{pem,der}_{public,private}_key` to understand x448 pkcs8? I think the answer here probably needs to be yes.

When we agree on the approach then this can be extended to x25519, ed25519, and ed448.

Outstanding work:
- [x] Add tests to confirm raw encodings are rejected for all other asym key types
- [x] Add tests to serialize keys to the PKCS8 forms and test roundtripping as well
- [x] Depending on what we choose to support from the questions above, complete that work and add testing around it.

refs #4386

### DIFF
--- a/docs/hazmat/primitives/asymmetric/x448.rst
+++ b/docs/hazmat/primitives/asymmetric/x448.rst
@@ -77,7 +77,11 @@ Key interfaces
             >>> from cryptography.hazmat.primitives import serialization
             >>> from cryptography.hazmat.primitives.asymmetric import x448
             >>> private_key = x448.X448PrivateKey.generate()
-            >>> private_bytes = private_key.private_bytes(serialization.Encoding.Raw, serialization.PrivateFormat.Raw, serialization.NoEncryption())
+            >>> private_bytes = private_key.private_bytes(
+            ...     encoding=serialization.Encoding.Raw,
+            ...     format=serialization.PrivateFormat.Raw,
+            ...     encryption_algorithm=serialization.NoEncryption()
+            ... )
             >>> loaded_private_key = x448.X448PrivateKey.from_private_bytes(private_bytes)
 
     .. method:: public_key()
@@ -108,7 +112,7 @@ Key interfaces
 
         :param format: A value from the
             :class:`~cryptography.hazmat.primitives.serialization.PrivateFormat`
-            enum or ``None``. If the ``encoding`` is
+            enum. If the ``encoding`` is
             :attr:`~cryptography.hazmat.primitives.serialization.Encoding.Raw`
             then ``format`` must be
             :attr:`~cryptography.hazmat.primitives.serialization.PrivateFormat.Raw`
@@ -137,7 +141,10 @@ Key interfaces
             >>> from cryptography.hazmat.primitives.asymmetric import x448
             >>> private_key = x448.X448PrivateKey.generate()
             >>> public_key = private_key.public_key()
-            >>> public_bytes = public_key.public_bytes(serialization.Encoding.Raw, serialization.PublicFormat.Raw)
+            >>> public_bytes = public_key.public_bytes(
+            ...     encoding=serialization.Encoding.Raw,
+            ...     format=serialization.PublicFormat.Raw
+            ... )
             >>> loaded_public_key = x448.X448PublicKey.from_public_bytes(public_bytes)
 
     .. method:: public_bytes(encoding, format)
@@ -157,7 +164,7 @@ Key interfaces
 
         :param format: A value from the
             :class:`~cryptography.hazmat.primitives.serialization.PublicFormat`
-            enum or ``None``. If the ``encoding`` is
+            enum. If the ``encoding`` is
             :attr:`~cryptography.hazmat.primitives.serialization.Encoding.Raw`
             then ``format`` must be
             :attr:`~cryptography.hazmat.primitives.serialization.PublicFormat.Raw`
--- a/docs/hazmat/primitives/asymmetric/x448.rst
+++ b/docs/hazmat/primitives/asymmetric/x448.rst
@@ -66,6 +66,20 @@ Key interfaces
 
         :returns: :class:`X448PrivateKey`
 
+    .. classmethod:: from_private_bytes(data)
+
+        :param bytes data: 56 byte private key.
+
+        :returns: :class:`X448PrivateKey`
+
+        .. doctest::
+
+            >>> from cryptography.hazmat.primitives import serialization
+            >>> from cryptography.hazmat.primitives.asymmetric import x448
+            >>> private_key = x448.X448PrivateKey.generate()
+            >>> private_bytes = private_key.private_bytes(serialization.Encoding.Raw, None, serialization.NoEncryption())
+            >>> loaded_private_key = x448.X448PrivateKey.from_private_bytes(private_bytes)
+
     .. method:: public_key()
 
         :returns: :class:`X448PublicKey`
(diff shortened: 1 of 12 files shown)


reaperhulk:
Revisiting this question "Do we want from_public_bytes and from_private_bytes to be able to handle x448 pkcs8 or just raw?"

If we want them to be able to handle PKCS8 then we'll have to add a password argument to `from_private_bytes`. I'm inclined to say no to this right now although I can't really articulate a great reason why.

alex:
I'm sure I'll regret asking this, but surely PKCS8 should go through the `serialization.load_{per,der}_{public,private}_key` functions?

reaperhulk:
That question was asked in the first post 😄. I think we do pkcs8 loading there, but will have to refactor a bit since the EVP symbols aren’t guaranteed to be available for the elif chain in _load_key. A bit weird to have pkcs8 there but raw loading as class methods but oh well

alex:
Mistakes were made 🤷‍♂️

docs/hazmat/primitives/asymmetric/x448.rst
```diff
@@ -66,6 +66,20 @@ Key interfaces
...
+        :param bytes data: 56 byte private key.
+
+        :returns: :class:`X448PrivateKey`
+
+        .. doctest::
+
+            >>> from cryptography.hazmat.primitives import serialization
+            >>> from cryptography.hazmat.primitives.asymmetric import x448
+            >>> private_key = x448.X448PrivateKey.generate()
+            >>> private_bytes = private_key.private_bytes(serialization.Encoding.Raw, None, serialization.NoEncryption())
```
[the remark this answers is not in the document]
reaperhulk: We can do that, but then we'll also need serialization.PublicFormat.Raw and accompanying tests that those things are rejected by the other methods properly.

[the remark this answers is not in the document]
alex: I think it's worth it, more readable. Agree/disasgree?

src/cryptography/hazmat/backends/openssl/x448.py
```diff
@@ -53,3 +77,38 @@ def exchange(self, peer_public_key):
...
+            return self._raw_private_bytes()
+
+        if (
+            encoding in _PEM_DER and
+            format is not serialization.PrivateFormat.PKCS8
+        ):
+            raise ValueError(
+                "format must be PKCS8 when encoding is PEM or DER"
+            )
+
```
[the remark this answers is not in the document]
reaperhulk: `_public_key_bytes` handles that case

reaperhulk:
x25519 will be converted in a separate PR. To maintain API compatibility we'll have to make `public_bytes` arguments be optional with a deprecation cycle.

docs/hazmat/primitives/asymmetric/x448.rst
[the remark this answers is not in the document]
reaperhulk: I'm fine with it, although we should think about how to have these sorts of conversations without completely implementing every iteration as we work towards an API we're happy with.

docs/hazmat/primitives/asymmetric/x448.rst
```diff
@@ -66,6 +66,20 @@ Key interfaces
...
+        :param bytes data: 56 byte private key.
+
+        :returns: :class:`X448PrivateKey`
+
+        .. doctest::
+
+            >>> from cryptography.hazmat.primitives import serialization
+            >>> from cryptography.hazmat.primitives.asymmetric import x448
+            >>> private_key = x448.X448PrivateKey.generate()
+            >>> private_bytes = private_key.private_bytes(serialization.Encoding.Raw, serialization.PrivateFormat.Raw, serialization.NoEncryption())
```
alex: Can you line wrap this?

docs/hazmat/primitives/asymmetric/x448.rst
```diff
@@ -77,6 +91,36 @@ Key interfaces
...
+        or
+        :attr:`~cryptography.hazmat.primitives.serialization.PrivateFormat.Raw`
+        ) are chosen to define the exact serialization.
+
+        :param encoding: A value from the
+            :class:`~cryptography.hazmat.primitives.serialization.Encoding` enum.
+
+        :param format: A value from the
+            :class:`~cryptography.hazmat.primitives.serialization.PrivateFormat`
+            enum or ``None``. If the ``encoding`` is
```
alex: This needs to be updated.

docs/hazmat/primitives/asymmetric/x448.rst
```diff
@@ -89,15 +133,38 @@ Key interfaces
...
+        or
+        :attr:`~cryptography.hazmat.primitives.serialization.PublicFormat.Raw`
+        ) are chosen to define the exact serialization.
+
+        :param encoding: A value from the
+            :class:`~cryptography.hazmat.primitives.serialization.Encoding` enum.
+
+        :param format: A value from the
+            :class:`~cryptography.hazmat.primitives.serialization.PublicFormat`
+            enum or ``None``. If the ``encoding`` is
```
alex: Update

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -508,6 +508,9 @@ def _evp_pkey_to_private_key(self, evp_pkey):
             self.openssl_assert(dh_cdata != self._ffi.NULL)
             dh_cdata = self._ffi.gc(dh_cdata, self._lib.DH_free)
             return _DHPrivateKey(self, dh_cdata, evp_pkey)
+        elif key_type == getattr(self._lib, "EVP_PKEY_X448", None):
+            # EVP_PKEY_X448 is not present in OpenSSL < 1.1.1
+            return _X448PrivateKey(self, evp_pkey)
```
alex: GC?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -539,6 +542,9 @@ def _evp_pkey_to_public_key(self, evp_pkey):
             self.openssl_assert(dh_cdata != self._ffi.NULL)
             dh_cdata = self._ffi.gc(dh_cdata, self._lib.DH_free)
             return _DHPublicKey(self, dh_cdata, evp_pkey)
+        elif key_type == getattr(self._lib, "EVP_PKEY_X448", None):
+            # EVP_PKEY_X448 is not present in OpenSSL < 1.1.1
+            return _X448PublicKey(self, evp_pkey)
```
alex: GC?

tests/hazmat/primitives/test_serialization.py
```diff
@@ -1231,3 +1233,68 @@ def test_non_bytes_password(self):
...
+            lambda derfile: derfile.read(),
+            mode="rb"
+        )
+        key = load_der_private_key(data, b"password", backend)
+        unencrypted = load_vectors_from_file(
+            os.path.join("asymmetric", "X448", "x448-pkcs8.der"),
+            lambda derfile: derfile.read(),
+            mode="rb"
+        )
+        key = load_der_private_key(data, b"password", backend)
```
alex: Duplicate of previous line?

tests/hazmat/primitives/test_x448.py
```diff
@@ -103,11 +104,59 @@ def test_rfc7748_1000_iteration(self, backend):
...
+                None,
+                serialization.load_der_private_key
+            ),
+        ]
+    )
+    def test_round_trip_private_serialization(self, encoding, fmt, encryption,
+                                              passwd, load_func, backend):
+        key = X448PrivateKey.generate()
+        serialized = key.private_bytes(encoding, fmt, encryption)
+        load_func(serialized, passwd, backend)
```
alex: assert something about this?

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: Any evp_pkey passed to this function has already been gc'd

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: Same

PR #4656, opened 2018-12-26
Fixes #4645 -- poll() on /dev/random before reading from /dev/urandom on Linux

### LINKED ISSUE #4645: Before using /dev/urandom verify CSPRNG has been seeded with select
Right now if `getrandom` isn't available on Linux, we fall back to `/dev/urandom`. This is good and proper. However, we can make this safer. Right now if you're running Python at early boot, when the CSPRNG isn't seeded, and you don't have `getrandom`, we'll read unseeded crap from `/dev/urandom`.

We can solve this by doing `select([open("/dev/random")], [], [])` first. This will ensure that `/dev/random` is ready (that is, has been seeded and has entropy available): https://github.com/torvalds/linux/blob/master/drivers/char/random.c#L1905

This is slightly overly aggressive, in that if some other process is sucking all the juice out of `/dev/random`, we'll block excessively, but there's no way around that.

alex:


src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -90,9 +94,54 @@ static struct {
...
+        return -1;
+    }
+    return 0;
+}
+
+static void close_fd(int fd) {
+    int n;
+    do {
+        n = close(fd);
+    } while (n < 0 && errno == EINTR);
```
nelhage: This is not thread-safe (maybe not a problem in Python?) and is unnecessary on Linux. See http://ewontfix.com/4/

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -90,9 +94,54 @@ static struct {
...
+}
+
+#ifdef __linux__
+/* On Linux, we open("/dev/random") and use select() to wait until it's readable
```
nelhage: pedantic nitpick: comment says `select`, code says `poll`

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -90,9 +94,54 @@ static struct {
...
+ * before we read from /dev/urandom, this ensures that we don't read from
+ * /dev/urandom before the kernel CSPRNG is initialized. This isn't necessary on
+ * other platforms because they don't have the same _bug_ as Linux does with
+ * /dev/urandom and early boot. */
+static int wait_on_devrandom(void) {
+    struct pollfd pfd = { 0 };
```
nelhage: `= {}` suffices to zero-initialize, and is IMO less confusing than `= { 0 }` since `struct pollfd` has three members.

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -90,9 +94,54 @@ static struct {
...
+    int ret = 0;
+    int random_fd = open("/dev/random", O_RDONLY);
+    if (random_fd < 0) {
+        return -1;
+    }
+    if (set_cloexec(random_fd) < 0) {
```
nelhage: `O_CLOEXEC` can be passed directly to the flags in `open`, making this both simpler and less racy (I suspect the race doesn't matter here, but it's the principle of the thing).

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -106,25 +155,25 @@ static int dev_urandom_fd(void) {
         }
     }
     if (urandom_cache.fd < 0) {
+#ifdef __linux__
+        if (wait_on_devrandom() < 0) {
```
nelhage: This is a stupid microoptimization, but I'd have `wait_on_devrandom` return the fd (or negative on error), and avoid having to double-open it. Since this only happens on startup it really doesn't matter.

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -106,25 +155,25 @@ static int dev_urandom_fd(void) {
...
-        /* set CLOEXEC flag */
-        flags = fcntl(fd, F_GETFD);
-        if (flags == -1) {
-            goto error;
-        } else if (fcntl(fd, F_SETFD, flags | FD_CLOEXEC) == -1) {
+        if (fstat(fd, &st)) {
             goto error;
         }
         /* Another thread initialized the fd */
         if (urandom_cache.fd >= 0) {
```
nelhage: You're probably aware and don't care, but this double-checked idiom is racy: http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

src/_cffi_src/openssl/src/osrandom_engine.c
alex: lol #4570 -- we tried this, and the environment we compile our manylinux1 wheels in has such an old libc that we can't have nice things.

src/_cffi_src/openssl/src/osrandom_engine.c
alex: they're different fds, one is urandom and the other is random

src/_cffi_src/openssl/src/osrandom_engine.c
alex: this was the pre-existing race condition I mentioned, I'm not dealing with it in this PR :-)

src/_cffi_src/openssl/src/osrandom_engine.c
nelhage: :joy:

src/_cffi_src/openssl/src/osrandom_engine.c
```diff
@@ -135,9 +177,7 @@ static int dev_urandom_fd(void) {
 
   error:
     if (fd != -1) {
-        do {
-            n = close(fd);
-        } while (n < 0 && errno == EINTR);
+        close(fd);
```
reaperhulk: This code used to assume `close` could fail and if it was `EINTR` it would retry. Is that foolish and it's safe to just assume `close` will always work/if it fails we don't care?

alex: Summarizing a conversation from IRC:

- On Linux the fd was generally closed if you get an EINTR, so this was wrong
- On other Unix the fd _may_ have been closed
- Signals are bad

The only time this can leak an fd now is if you use a weird unix and signals. I decided I was ok with this.

alex: Oh, also on linux the behavior varied between glibc and musl.

PR #4658, opened 2018-12-28
fix epoch overflow 32bit

luzgabriel:
https://en.wikipedia.org/wiki/Year_2038_problem

I've had a problem while generating certs with dates higher than 2038 while running it on a raspberry pi (32-bit architecture)

The openssl api has a function that takes the date as a string, so you don't have such problem when dealing with dates higher than 2038

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -5,7 +5,6 @@
 from __future__ import absolute_import, division, print_function
 
 import base64
-import calendar
 import collections
 import contextlib
 import itertools
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -836,8 +836,8 @@ class Backend(object):
         return _Certificate(self, x509_cert)
 
     def _set_asn1_time(self, asn1_time, time):
-        timestamp = calendar.timegm(time.timetuple())
-        res = self._lib.ASN1_TIME_set(asn1_time, timestamp)
+        timestamp = time.strftime('%Y%m%d%H%M%SZ').encode('ascii')
+        res = self._lib.ASN1_TIME_set_string(asn1_time, timestamp)
         if res == self._ffi.NULL:
             errors = self._consume_errors()
             self.openssl_assert(


alex:
Good find! Please add a test for this.

alex:
Looks like there's also soem windows tests that'll need updating.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -836,8 +835,8 @@ def create_x509_certificate(self, builder, private_key, algorithm):
         return _Certificate(self, x509_cert)
 
     def _set_asn1_time(self, asn1_time, time):
-        timestamp = calendar.timegm(time.timetuple())
-        res = self._lib.ASN1_TIME_set(asn1_time, timestamp)
+        timestamp = time.strftime('%Y%m%d%H%M%SZ').encode('ascii')
```
reaperhulk: Unfortunately this incorrectly encodes many values. RFC 5280 requires that times be encoded as UTCTime prior to 2050 and GeneralizedTime thereafter (https://tools.ietf.org/html/rfc5280#section-4.1.2.5). `ASN1_TIME_set_string_X509` does the right thing, but was only added in OpenSSL 1.1.1. If we want to change this we'll need to either replicate the logic in Python (along with plenty of tests to confirm that we're encoding the ASN.1 we expect) or create `Cryptography_ASN1_TIME_set_string_X509` which can duplicate the logic from OpenSSL 1.1.1.

luzgabriel:
```
    /*
     * Per RFC 5280 (section 4.1.2.5.), the valid input time
     * strings should be encoded with the following rules:
     *
     * 1. UTC: YYMMDDHHMMSSZ, if YY < 50 (20YY) --> UTC: YYMMDDHHMMSSZ
     * 2. UTC: YYMMDDHHMMSSZ, if YY >= 50 (19YY) --> UTC: YYMMDDHHMMSSZ
     * 3. G'd: YYYYMMDDHHMMSSZ, if YYYY >= 2050 --> G'd: YYYYMMDDHHMMSSZ
     * 4. G'd: YYYYMMDDHHMMSSZ, if YYYY < 2050 --> UTC: YYMMDDHHMMSSZ
     *
     * Only strings of the 4th rule should be reformatted, but since a
     * UTC can only present [1950, 2050), so if the given time string
     * is less than 1950 (e.g. 19230419000000Z), we do nothing...
     */
```

comment taken from the openssl code

njsmith:
Another manifestation of this bug: https://github.com/python-trio/trustme/issues/41

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: Yes, as you can see in that message any time of form `YYYYMMDDHHMMSSZ` must be reformatted to `YYMMDDHHMMSSZ` if `YYYY` < 2050. The PR here will unconditionally set it to `YYYYMMDDHHMMSSZ` because `ASN1_TIME_set_string` doesn't reformat (and the function that does reformat is not available in all supported versions of OpenSSL).

I think implementing this logic in Python and calling `ASN1_TIME_set_string` is a reasonable path forward, but tests will need to be added to check every case + verify that the resulting ASN.1 generated has the correct underlying string type (generalized vs utc).

alex:
fixed in #4727

PR #4667, opened 2019-01-01
add x448 private/public PEM/DER pkcs8/subjectpublickeyinfo vectors

reaperhulk:
Since we're going to support these types of serializations (both for the ed25519/ed448 keys as well as for X25519/X448 for ECIES purposes) we need some vectors!


alex:
Can you make the commit message accurate? This adds vectors, not serialization.

PR #4690, opened 2019-01-15
add support for byteslike on password and data for pkcs12 loading

reaperhulk:


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -450,13 +450,12 @@ def _bytes_to_bio(self, data):
         The char* is the storage for the BIO and it must stay alive until the
         BIO is finished with.
         """
-        data_char_p = self._ffi.new("char[]", data)
         bio = self._lib.BIO_new_mem_buf(
-            data_char_p, len(data)
+            self._ffi.from_buffer(data), len(data)
```
alex: You need to keep the `from_buffer` alive in a local var here as well I think?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2190,25 +2190,48 @@ def aead_cipher_supported(self, cipher):
...
         bio = self._bytes_to_bio(data)
         p12 = self._lib.d2i_PKCS12_bio(bio.bio, self._ffi.NULL)
         if p12 == self._ffi.NULL:
             self._consume_errors()
             raise ValueError("Could not deserialize PKCS12 data")
 
         p12 = self._ffi.gc(p12, self._lib.PKCS12_free)
+
+        if password is not None:
+            utils._check_byteslike("password", password)
```
alex: surely we should be checking the type before we try to parse the data?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2190,25 +2190,48 @@ def aead_cipher_supported(self, cipher):
             self._lib.EVP_get_cipherbyname(cipher_name) != self._ffi.NULL
         )
 
-    def load_key_and_certificates_from_pkcs12(self, data, password):
+    @contextlib.contextmanager
+    def _zeroed_null_terminated_pass_buf(self, password):
```
alex: drop pass from the name, nothing about the behavior is password specific

PR #4698, opened 2019-01-16
support bytes-like for X25519PrivateKey.from_private_bytes

reaperhulk:
yuck.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -14,6 +14,7 @@ from contextlib import contextmanager
 import asn1crypto.core
 
 import six
+from six.moves import range
 
 from cryptography import utils, x509
 from cryptography.exceptions import UnsupportedAlgorithm, _Reasons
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2109,8 +2111,12 @@ def x25519_load_private_bytes(self, data):
         # contains an OCTET STRING of length 32! So the last two bytes here
         # are \x04\x20, which is an OCTET STRING of length 32.
         pkcs8_prefix = b'0.\x02\x01\x000\x05\x06\x03+en\x04"\x04 '
-        bio = self._bytes_to_bio(pkcs8_prefix + data)
-        evp_pkey = backend._lib.d2i_PrivateKey_bio(bio.bio, self._ffi.NULL)
+        with self._zeroed_bytearray(48) as ba:
+            ba[0:16] = pkcs8_prefix
+            ba[16:] = data
```
alex: Is data's length actually checked before this?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2197,6 +2203,23 @@ def aead_cipher_supported(self, cipher):
...
+        zero when we're done.
+        """
+        ba = bytearray(length)
+        try:
+            yield ba
+        finally:
+            # We zero the bytearray this way because at the moment we're not
+            # sure of a better way that can guarantee it overwrites the
+            # memory and doesn't just replace the underlying char *.
+            for i in range(length):
```
alex: Please factor this into its own function and use it in the other place that we use `memmove` to zero

reaperhulk: We're doing `memmove` in the other place on a `char[]` so `memmove` is the right thing to use there.

alex: A loop also works though.

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: It was not being properly checked and #4703 adds the check (along with making x25519 and x448 consistent across their from_{public,private}_bytes methods

PR #4714, opened 2019-01-18
use OpenSSL 1.1.1's internal RNG + atfork handlers where possible

reaperhulk:
Well, I made this, but I'm not actually convinced we should merge it. This doesn't really gain us much until 1.1.1 is the lowest version we support (timeline on that: at least 5 years, probably more like 8-10?).


alex:
If I'm still maintaining this in 8-10 years, I reserve the right to burn it all down.

reaperhulk:
I am setting a reminder for 8 years from today on my calendar.

reaperhulk:
(Closing because we don't want to support two paths here and we can't get rid of our custom engine until we can drop support for all OpenSSL < 1.1.1. It'll be a number of years...)

PR #4728, opened 2019-01-22
allow asn1 times of 1950-01-01 and later.

reaperhulk:


alex:
How about at test that these encode correctly?

PR #4739, opened 2019-01-26
Rip out setup_requires from our setup.py

alex:


reaperhulk:
I'm going to close this since it's going to just be sitting here until I'm comfortable saying `pip 18+` is a reasonable requirement (which is probably like 2021...?). Note that `python-pip` in Ubuntu 18.10 is **still** 9.0.1.

alex:
@reaperhulk weak counter-point: you shouldn't install anything with `python-pip`, you should make a virtualenv.

PR #4747, opened 2019-02-03
concede to digicert's garbage CDN

reaperhulk:
Which, despite supporting HTTPS, is non-deterministically providing 404s and DigiCert has asserted that http is the only "supported" protocol.


alex:
The current breakage is not digicert, it's openssl

reaperhulk:
```
/home/travis/build/pyca/cryptography/docs/development/test-vectors.rst.rst:173:broken link: https://cacerts.digicert.com/DigiCertGlobalRootG3.crt (404 Client Error: Not Found for url: https://cacerts.digicert.com/DigiCertGlobalRootG3.crt)
```

reaperhulk:
To clarify, by that I mean the digi non-deterministic failure is *also* occurring and I'm unwilling to continuing to fight both the normal flakiness of the internet and DigiCert's explicitly bad and wrong CDN that they don't want to fix.

The OpenSSL issue is interesting in that the page in question has genuinely disappeared. The same text appears to be at `https://www.openssl.org/docs/man1.0.2/man3/CRYPTO_THREADID_set_callback.html` so maybe we should switch to that.

PR #4748, opened 2019-02-03
update the thread link

reaperhulk:
linkcheck doing its job!


alex:
Ugh, this paragraph is out of date, on modern OpenSSL's we use OpenSSL's default locking scheme. And there's no GIL issues since we rewrote the locking in C (booo!).

Do you mind updating this as well, or would you prefer to do that in a seperate PR?

PR #4750, opened 2019-02-07
expose and document aes gcm example, Fixes #4749

### LINKED ISSUE #4749: expose AES GCM example outside of hazmat
the Cryptography docs have a really nice example of how to correctly implement AES-GCM encrypt/decrypt it would be handy if this was exposed as a function outside of `hazmat` akin to Fernet.

https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#cryptography.hazmat.primitives.ciphers.modes.GCM

see also: https://github.com/pyca/cryptography/issues/2493

graingert:


reaperhulk:
Thanks for offering this contribution. However, we don't have an interest in exposing our example as an `encrypt` function. We already expose a method to (more safely) utilize GCM via the [aead](https://cryptography.io/en/latest/hazmat/primitives/aead/) module, which provides one-shot safe encryption semantics (and also concats the tag so that you have a single ciphertext blob).

PR #4753, opened 2019-02-14
add single_extensions to OCSPResponse

c0r0n3r:


alex:
Just FYI, the failing Windows tests in Jenkins are not your fault, there's an issue with our CI, so don't worry about those.

alex:
Failing travis tests look relevant though.

alex:
It looks like `test_single_extensions` is failing with some versions of OpenSSL.

reaperhulk:
Is this something you're still interested in working on?

c0r0n3r:
It is absolutely interesting from me. In my humble opinion the patch works well. The tests fail because they are run if OpenSSL version lesser than1.1.0f in spite of the fact that the test are marked to run only if OpenSSL version is greater or equal then 1.1.0f.

alex:
The tests need to be passing for us to consider this for merging. I believe the issue is that the test function doesn't take a `backend` argument, so the skip isn't invoked.

c0r0n3r:
Thanks for the very useful comment @alex

src/cryptography/hazmat/backends/openssl/decode_asn1.py
```diff
@@ -856,6 +871,10 @@ def _decode_nonce(backend, nonce):
     OCSPExtensionOID.NONCE: _decode_nonce,
 }
 
+_OCSP_SINGLERESP_EXTENSION_HANDLERS = {
```
alex: This doesn't appear to be used, so I'm pretty sure the extension parser below is being created incorrectly

c0r0n3r:
May I do anything to facilitate the review?

reaperhulk:
@c0r0n3r I'll be looking at this soon, sorry about the delay in review.

src/cryptography/hazmat/backends/openssl/decode_asn1.py
```diff
@@ -897,3 +897,15 @@ def _decode_nonce(backend, nonce):
     get_ext=lambda backend, x, i: backend._lib.OCSP_BASICRESP_get_ext(x, i),
     handlers=_OCSP_BASICRESP_EXTENSION_HANDLERS,
 )
+
+_OCSP_SINGLERESP_EXT_PARSER_NO_SCT = _X509ExtensionParser(
+    ext_count=lambda backend, x: backend._lib.OCSP_SINGLERESP_get_ext_count(x),
+    get_ext=lambda backend, x, i: backend._lib.OCSP_SINGLERESP_get_ext(x, i),
+    handlers=_EXTENSION_HANDLERS_NO_SCT
```
reaperhulk: What is the set of allowable extensions here? Looking at the RFC it appears that "any CRLEntry extension" from RFC 5280 section 5.3 is allowable + the set defined in RFC 6960. Are there any others? The set of handlers here doesn't look right based on that.

src/cryptography/hazmat/backends/openssl/ocsp.py
```diff
@@ -319,6 +321,18 @@ def serial_number(self):
...
 
+    @utils.cached_property
+    @_requires_successful_response
+    def single_extensions(self):
+        if self._backend._lib.CRYPTOGRAPHY_OPENSSL_110_OR_GREATER:
+            return _OCSP_SINGLERESP_EXT_PARSER.parse(
+                self._backend, self._single
+            )
+        else:
+            return _OCSP_SINGLERESP_EXT_PARSER_NO_SCT.parse(
```
reaperhulk: This branch isn't taken with current tests. We should add another test vector for an extension supported in < 1.1.0 (perhaps CRLReason) that will take this branch.

reaperhulk:
Continued in #5059

PR #4756, opened 2019-02-20
encode the package version in the shared object

### LINKED ISSUE #4754: Encode package version as a shared object symbol
One of the more common issues we see with users is an `AttributeError` where one of our `Cryptograpy_HAS_*` can't be found. This is pretty much always an issue where Python has loaded a newer version of cryptography's Python code, but for some reason loads an older shared object (which does not have the symbol the Python code is looking for).

This is a bad error message for a serious environment problem, and it doesn't always happen immediately. We should encode the version of cryptography in the shared object we build and then raise an `ImportError` on import if the shared object version and cryptography's version do not match.

refs https://github.com/certbot/certbot/issues/5651

reaperhulk:
fixes #4754

### DIFF
--- a/src/_cffi_src/openssl/cryptography.py
+++ b/src/_cffi_src/openssl/cryptography.py
@@ -4,6 +4,19 @@
 
 from __future__ import absolute_import, division, print_function
 
+import os
+
+
+# Load the cryptography __about__ to get the current package version
+base_src = os.path.dirname(
+    os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+)
+about = os.path.join(base_src, "cryptography", "__about__.py")
+__about__ = {}
+with open(about) as f:
+    exec(f.read(), __about__)
+
+
 INCLUDES = """
 /* define our OpenSSL API compatibility level to 1.0.1. Any symbols older than
    that will raise an error during compilation. We can raise this number again
@@ -62,7 +75,8 @@ INCLUDES = """
     (OPENSSL_VERSION_NUMBER < 0x101000af || CRYPTOGRAPHY_IS_LIBRESSL)
 #define CRYPTOGRAPHY_OPENSSL_LESS_THAN_111 \
     (OPENSSL_VERSION_NUMBER < 0x10101000 || CRYPTOGRAPHY_IS_LIBRESSL)
-"""
+#define CRYPTOGRAPHY_PACKAGE_VERSION "{0}"
+""".format(__about__["__version__"])
 
 TYPES = """
 static const int CRYPTOGRAPHY_OPENSSL_102L_OR_GREATER;
@@ -76,6 +90,8 @@ static const int CRYPTOGRAPHY_OPENSSL_LESS_THAN_111;
 static const int CRYPTOGRAPHY_IS_LIBRESSL;
 
 static const int CRYPTOGRAPHY_LIBRESSL_28_OR_GREATER;
+
+static const char *const CRYPTOGRAPHY_PACKAGE_VERSION;
 """
 
 FUNCTIONS = """
--- a/src/_cffi_src/openssl/cryptography.py
+++ b/src/_cffi_src/openssl/cryptography.py
@@ -4,18 +4,6 @@
 
 from __future__ import absolute_import, division, print_function
 
-import os
-
-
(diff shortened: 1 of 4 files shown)


src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -164,6 +165,29 @@ def _verify_openssl_version(lib):
...
+    so_package_version = Binding.ffi.string(
+        Binding.lib.CRYPTOGRAPHY_PACKAGE_VERSION
+    ).decode("ascii")
+    if version != so_package_version:
+        raise ImportError(
+            "The version of cryptography does not match with the loaded "
+            "shared object. This can happen if you have multiple copies of "
+            "cryptography installed in your Python path. Please try creating "
+            "a new virtual environment to resolve this issue. "
+            "Loaded python version: {0}, shared object version: {1}".format(
```
alex: Now that we've dropped py2.6 you can just use `{}` you don't need to number them

src/_cffi_src/openssl/cryptography.py
```diff
@@ -4,6 +4,19 @@
...
 
+import os
+
+
+# Load the cryptography __about__ to get the current package version
+base_src = os.path.dirname(
+    os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+)
+about = os.path.join(base_src, "cryptography", "__about__.py")
+__about__ = {}
```
alex: There's no reason for this to have a weird `__about__` name

src/_cffi_src/openssl/cryptography.py
```diff
@@ -62,7 +75,8 @@
     (OPENSSL_VERSION_NUMBER < 0x101000af || CRYPTOGRAPHY_IS_LIBRESSL)
 #define CRYPTOGRAPHY_OPENSSL_LESS_THAN_111 \
     (OPENSSL_VERSION_NUMBER < 0x10101000 || CRYPTOGRAPHY_IS_LIBRESSL)
-"""
+#define CRYPTOGRAPHY_PACKAGE_VERSION "{0}"
```
alex: same

src/_cffi_src/openssl/cryptography.py
```diff
@@ -62,7 +74,8 @@
     (OPENSSL_VERSION_NUMBER < 0x101000af || CRYPTOGRAPHY_IS_LIBRESSL)
 #define CRYPTOGRAPHY_OPENSSL_LESS_THAN_111 \
     (OPENSSL_VERSION_NUMBER < 0x10101000 || CRYPTOGRAPHY_IS_LIBRESSL)
-"""
+#define CRYPTOGRAPHY_PACKAGE_VERSION "{}"
+""".format(about["__version__"])
```
alex: Would it make more sense to do this in _all_ `build_ffi_for_binding`?

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -164,6 +165,29 @@ def _verify_openssl_version(lib):
...
+    # are installed and available in the python path. To avoid errors cropping
+    # up later this code checks that the currently imported package and the
+    # shared object that were loaded have the same version and raise an
+    # ImportError if they do not
+    so_package_version = Binding.ffi.string(
+        Binding.lib.CRYPTOGRAPHY_PACKAGE_VERSION
+    ).decode("ascii")
+    if version != so_package_version:
+        raise ImportError(
+            "The version of cryptography does not match with the loaded "
```
alex:
```suggestion
            "The version of cryptography does not match the loaded "
```

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -172,3 +196,4 @@ def _verify_openssl_version(lib):
 Binding.init_static_locks()
 
 _verify_openssl_version(Binding.lib)
+_verify_package_version(cryptography.__version__)
```
alex: Verify cryptography version first I think? Before the locks even.

bmw:
I think the `ImportError` being raised here will help us a lot when people hit this issue. Thank you for adding it!

PR #4758, opened 2019-02-20
docs: full state or province name when generating a sample certificate

itinerarium:
Generally, it appears that a state or province name is not abbreviated, but is rather the full name.

6.3.5 of ITU-T X.520 (10/2016) provides a spelled out sample state. In other contexts, hints suggest the "full name" of a state or province.

Accordingly, replacement of CA with California in the sample code might be more consistent with general usage.


reaperhulk:
Great catch, it is indeed true that for publicly trusted certificates the full state or province name (rather than an abbreviation) is needed.

PR #4760, opened 2019-02-22
Guard NPN functions in case they are unavailable

neheb:
NPN in OpenSSL can be compile-time disabled. In that case, this patch
fixes compilation.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -432,25 +432,9 @@ SRTP_PROTECTION_PROFILE *SSL_get_selected_srtp_profile(SSL *);
 
 long SSL_session_reused(SSL *);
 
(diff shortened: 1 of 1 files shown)


alex:
@reaperhulk can we just drop NPN entirely? I don't think anything should be using it at this point?

reaperhulk:
@alex we'd need to deprecate and remove it from pyopenssl first (hey we should do that)

alex:
Well, if we just set the disabled flag, it'll auto-disable in pyOpenSSL. But yes, we should properly deprecate it, I'll send a PR.

alex:
https://github.com/pyca/pyopenssl/pull/820

reaperhulk:
Okay, since we're deprecating on the pyOpenSSL side and it already handles the case where the symbols aren't present let's just go ahead and alter this PR to remove the bindings entirely.

reaperhulk:
#4765 replaces this

PR #4761, opened 2019-02-22
Fix compilation when missing ENGINE

neheb:
OpenWrt now disables ENGINE by default as most router platforms have no drivers or hardware for acceleration. Android doesn't have it either.

### DIFF
--- a/src/_cffi_src/openssl/src/osrandom_engine.c
+++ b/src/_cffi_src/openssl/src/osrandom_engine.c
@@ -496,6 +496,7 @@ static RAND_METHOD osrandom_rand = {
     osrandom_rand_status,
 };
 
+#ifndef OPENSSL_NO_ENGINE
 static const ENGINE_CMD_DEFN osrandom_cmd_defns[] = {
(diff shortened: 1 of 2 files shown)


reaperhulk:
Have you tested this on an OpenSSL with engine disabled? I would have expected it to fail since all of `src/_cffi_src/openssl/engine.py` is not declared conditionally. Additionally, if engine support isn't present we probably need to disable activation of the osrandom engine in  `src/cryptography/hazmat/backends/openssl/backend.py`.

Since OpenWRT has switched I think we probably need to support this config, but to do so we'll need to add an ENGINE-less builder to CI so that we can properly confirm that it's working as expected.

neheb:
@reaperhulk I patched out compilation of engine.py (src/cryptography.egg-info/SOURCES.txt) as it was easier than modifying engine.py.

Only compile tested.

neheb:
Closing in favor of https://github.com/pyca/cryptography/pull/4763

Turns out I was testing with ENGINE enabled on accident...

PR #4762, opened 2019-02-22
Fix compilation without deprecated OpenSSL 1.1 APIs

neheb:
OpenWrt has a compile-time option to disable deprecated APIs to save some space for embedded devices.

I put the replacements in the #else path.

Untested with LibreSSL.


reaperhulk:
Removing our dependency on deprecated APIs is very challenging as long as we support OpenSSL 1.0.2. To successfully do that would require creating wrappers for all our function calls, which would ultimately the equivalent of the deprecated API support in 1.1.0+.

PR #4763, opened 2019-02-24
support NO_ENGINE

reaperhulk:


src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -115,10 +115,13 @@ def _register_osrandom_engine(cls):
         # reliably clear the error queue. Once we clear it here we will
         # error on any subsequent unexpected item in the stack.
         cls.lib.ERR_clear_error()
-        cls._osrandom_engine_id = cls.lib.Cryptography_osrandom_engine_id
-        cls._osrandom_engine_name = cls.lib.Cryptography_osrandom_engine_name
-        result = cls.lib.Cryptography_add_osrandom_engine()
-        _openssl_assert(cls.lib, result in (1, 2))
+        if cls.lib.Cryptography_HAS_ENGINE:
+            cls._osrandom_engine_id = cls.lib.Cryptography_osrandom_engine_id
```
alex: Why do we do these assignments instead of just having the backend read the attrs from `lib`?

reaperhulk: Looks like we don't need it any more (if we ever did). We should do this in a separate PR though.

.travis/install.sh
```diff
@@ -14,12 +14,13 @@ shlib_sed() {
 # download, compile, and install if it's not already present via travis
 # cache
 if [ -n "${OPENSSL}" ]; then
-    OPENSSL_DIR="ossl-2/${OPENSSL}"
+    OPENSSL_CONFIG_PATH=$(dirname "${BASH_SOURCE[0]}")
```
alex: This is really like current location or something, and it can go at the top of the file.

.travis/run.sh
```diff
@@ -6,13 +6,15 @@ if [[ "${TOXENV}" == "pypy" ]]; then
     eval "$(pyenv init -)"
 fi
 if [ -n "${LIBRESSL}" ]; then
-    OPENSSL=$LIBRESSL
-    export CFLAGS="-Werror -Wno-error=deprecated-declarations -Wno-error=discarded-qualifiers -Wno-error=unused-function"
+    LIBRESSL_DIR="ossl-2/${LIBRESSL}"
```
alex: Why did this branch need to change?

.travis/run.sh
```diff
@@ -6,13 +6,15 @@ if [[ "${TOXENV}" == "pypy" ]]; then
...
+    export CFLAGS="-Werror -Wno-error=deprecated-declarations -Wno-error=discarded-qualifiers -Wno-error=unused-function -I$HOME/$LIBRESSL_DIR/include"
+    export PATH="$HOME/$LIBRESSL_DIR/bin:$PATH"
+    export LDFLAGS="-L$HOME/$LIBRESSL_DIR/lib -Wl,-rpath=$HOME/$LIBRESSL_DIR/lib"
 fi
 
 if [ -n "${OPENSSL}" ]; then
-    OPENSSL_DIR="ossl-2/${OPENSSL}"
-
+    OPENSSL_CONFIG_PATH=$(dirname "${BASH_SOURCE[0]}")
```
alex: Same

.travis/run.sh
reaperhulk: Because we were using the OpenSSL code path previously to do the directory path creation. Now that we build the directory with a hash based on the config args that no longer works for libre (where we don't have those config args).

PR #4765, opened 2019-02-25
remove NPN bindings -- you should be using ALPN!

reaperhulk:
pyOpenSSL consumed these, but we've marked it as deprecated and it already handles the case where the bindings are not available.

### DIFF
--- a/.travis/downstream.d/twisted.sh
+++ b/.travis/downstream.d/twisted.sh
@@ -9,7 +9,7 @@ case "${1}" in
         ;;
     run)
         cd twisted
-        python -m twisted.trial twisted.protocols.test.test_tls.TLSMemoryBIOTests.test_disorderlyShutdown
+        python -m twisted.trial src/twisted
         ;;
     *)
         exit 1
--- a/.travis/downstream.d/twisted.sh
+++ b/.travis/downstream.d/twisted.sh
@@ -9,7 +9,7 @@ case "${1}" in
         ;;
     run)
         cd twisted
-        python -m twisted.trial src/twisted
+        python -m twisted.trial twisted.protocols.test.test_tls.TLSMemoryBIOTests.test_disorderlyShutdown
         ;;
     *)
         exit 1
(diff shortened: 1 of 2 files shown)


alex:
We need to uncondtionally set `Cryptography_HAS_NEXTPROTONEG` to false if you do this

alex:
What about `SSL_select_next_proto` and `OPENSSL_NPN_NEGOTIATED`

reaperhulk:
`SSL_select_next_proto` is apparently also usable in an ALPN context: https://www.openssl.org/docs/man1.0.2/man3/SSL_select_next_proto.html

`OPENSSL_NPN_NEGOTIATED` is declared even if `OPENSSL_NO_NEXTPROTONEG` is declared, but there's no reason for it to exist so we can remove it.

reaperhulk:
Damn, the fix I put into twisted's tests doesn't entirely fix the test failures. One left. Dropping it here for later investigation, but this could be a real thing that needs fixing on the twisted side

```
[ERROR]
  File "/home/travis/build/pyca/cryptography/twisted/src/twisted/protocols/test/test_tls.py", line 1943, in checkNegotiatedProtocol
    self.assertEqual(client.negotiatedProtocol, b'h2')
  File "/home/travis/build/pyca/cryptography/twisted/src/twisted/trial/_synctest.py", line 432, in assertEqual
    super(_Assertions, self).assertEqual(first, second, msg)
  File "/opt/python/2.7.15/lib/python2.7/unittest/case.py", line 513, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/opt/python/2.7.15/lib/python2.7/unittest/case.py", line 506, in _baseAssertEqual
    raise self.failureException(msg)
twisted.trial.unittest.FailTest: None != 'h2'
```

reaperhulk:
On @glyph's recommendation I changed the test suite in the last push (a month ago) to just run the one test to see if it passed. It did. This is state leakage between tests, so twisted needs to sort that out before we can merge this.

alex:
Have we filed a bug with twisted for that?

reaperhulk:
Ping @glyph plz fix state leakage in your testing

alex:
I filed https://twistedmatrix.com/trac/ticket/9725 for this

PR #4776, opened 2019-02-26
remove unused x509 bindings

reaperhulk:
This may touch things pypy cares about...


reaperhulk:
This PR will fail until the EVP PR merges.

PR #4781, opened 2019-02-26
Mark bytes-like change as breaking

tiran:
The ffi.from_buffer() change is a breaking change for applications that
misused the APIs in the first place, e.g.
https://github.com/mitya57/secretstorage/commit/a334722a

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -36,9 +36,12 @@ Changelog
   well. If you still need this deprecated path please install cryptography
   with the ``idna`` extra: ``pip install cryptography[idna]``.
 * **BACKWARDS INCOMPATIBLE:** The minimum supported PyPy version is now 5.4.
-* Numerous classes and functions have been updated to allow :term:`bytes-like`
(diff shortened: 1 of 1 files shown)


reaperhulk:
I'm not sure how I feel about this. Most of these callsites enforced bytes via an `isinstance` check already and passing a list was definitely an unsupported and undocumented thing. I'm not a huge fan of calling this a breaking change when it only broke a tiny subset of people who were misusing the library to begin with, but what do you think @alex?

alex:
Is there an in-between solution here? Not sure what it'd be -- include the specific example that can cause breakage but not do the big bold?

reaperhulk:
In general what we commit to not breaking is our public API as documented. Since the change did break some users I’m okay with the cautionary language at the end, but I’d prefer to not set a precedent of everything being flagged as a breaking change whether or not it changes our documented behavior.

reaperhulk:
Do you still plan to update this Christian?

reaperhulk:
This can be reopened/rebased if desired.

PR #4783, opened 2019-02-26
Add PEP 517 to pyproject.toml

pganssle:
This puts a new minimum on the PEP 518 requirement for `setuptools` because older versions of setuptools' PEP 517 backend will fail to include `setup.py` in an sdist.

I will note that the presence of `pyproject.toml` alone is enough to trigger PEP 517 in the latest version of `pip`, using the default backend, so regardless of whether or not `build-backend` is added to `pyproject.toml`, the bounds on `setuptools` need to be bumped (in `pyproject.toml` only).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,10 @@
 [build-system]
-# Must be kept in sync with `setup_requirements` in `setup.py`
 requires = [
-    "setuptools>=18.5",
+    # The minimum setuptools version is specific to the PEP 517 backend,
+    # and may be stricter than the version required in `setup.py`
+    "setuptools>=40.6.0",
     "wheel",
+    # Must be kept in sync with the `setup_requirements` in `setup.py`
     "cffi>=1.8,!=1.11.3; python_implementation != 'PyPy'",
 ]
+build-backend = "setuptools.build_meta"


reaperhulk:
So without this patch if setuptools happens to be older than 40.6 when creating an sdist via `python setup.py sdist` the `setup.py` won't be included in the sdist?

It looks like the comment in `pyproject.toml` about keeping the info in setup.py and this file in sync is no longer accurate (at least for setuptools) though?

pganssle:
> So without this patch if setuptools happens to be older than 40.6 when creating an sdist via python setup.py sdist the setup.py won't be included in the sdist?

No, this is a bug with the PEP 517 build backend, so it will fail with `tox` or `pep517.build` or anything else that uses the PEP 517 build backend.

It's probably a pretty rare use case to build an sdist from source, but better safe than sorry.

reaperhulk:
Okay, got it. Could you update the comment about keeping the setup_requirements and this in sync that notes why setuptools versioning doesn't match? Other than that this LGTM.

pganssle:
@reaperhulk Updated, let me know if the comments need further changing.

reaperhulk:
For reference this PR was unrelated to the bug in cryptography 2.6 (which was an issue with wheel building resolved by re-releasing as 2.6.1 after correcting our wheel builder scripts)

PR #4785, opened 2019-02-27
support ed25519 openssh public keys

### LINKED ISSUE #4676: Add ed25519 support to openssh public key parser
Currently, ed25519 support is being added in the following pull request:

https://github.com/pyca/cryptography/pull/4114

After that lands, ed25519 support will then need to be added to the openssh public key parser. This issue is for tracking that work.

reaperhulk:
fixes #4676

### DIFF
--- a/src/cryptography/hazmat/primitives/serialization/ssh.py
+++ b/src/cryptography/hazmat/primitives/serialization/ssh.py
@@ -110,9 +110,6 @@ def _load_ssh_ed25519_public_key(expected_key_type, decoded_data, backend):
     if rest:
         raise ValueError('Key body contains extra bytes.')
 
-    if len(data) != ed25519._ED25519_KEY_SIZE:
-        raise ValueError("Public key is not 32 bytes")
-
     return ed25519.Ed25519PublicKey.from_public_bytes(data)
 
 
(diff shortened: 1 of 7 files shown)


src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -102,6 +104,15 @@ def _load_ssh_ecdsa_public_key(expected_key_type, decoded_data, backend):
     return ec.EllipticCurvePublicKey.from_encoded_point(curve, data)
 
 
+def _load_ssh_ed25519_public_key(expected_key_type, decoded_data, backend):
+    data, rest = _ssh_read_next_string(decoded_data)
```
alex: Need to verify rest is empty

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -102,6 +104,18 @@ def _load_ssh_ecdsa_public_key(expected_key_type, decoded_data, backend):
...
+
+    if rest:
+        raise ValueError('Key body contains extra bytes.')
+
+    if len(data) != ed25519._ED25519_KEY_SIZE:
+        raise ValueError("Public key is not 32 bytes")
```
alex: `from_public_bytes` already raises this same exception, do we need this check?

reaperhulk: Probably not. Since it's an invariant of the key type and not a particular quirk of ssh it makes sense for us to just use the enforcement at the key loading layer.

PR #4793, opened 2019-02-27
rfc4514_string: use oid._name before dotted_string

jcrowgey:
The oid._name propery is compliant with the LDAP registry:

```
FROM RFC4514:

   If the AttributeType is defined to have a short name (descriptor)
   [RFC4512] and that short name is known to be registered [REGISTRY]
   [RFC4520] as identifying the AttributeType, that short name, a
   <descr>, is used.
```

[REGISTRY]: https://www.iana.org/assignments/ldap-parameters/ldap-parameters.xhtml

So I believe it is compliant to use the oid._name values whenever they
are not equal to "Unknown OID".

This improves readability for many of the common attributes which do not appear in the short table.

### DIFF
--- a/src/cryptography/x509/name.py
+++ b/src/cryptography/x509/name.py
@@ -120,9 +120,12 @@ class NameAttribute(object):
         Format as RFC4514 Distinguished Name string.
 
         Use short attribute name if available, otherwise fall back to OID
-        dotted string.
+        name, and finally to dotted string.
         """
-        key = _NAMEOID_TO_NAME.get(self.oid, self.oid.dotted_string)
+        key = _NAMEOID_TO_NAME.get(self.oid, (
+            self.oid._name if self.oid._name != "Unknown OID"
+            else self.oid.dotted_string
+        ))
         return '%s=%s' % (key, _escape_dn_value(self.value))
 
     def __eq__(self, other):
--- a/tests/x509/test_x509.py
+++ b/tests/x509/test_x509.py
@@ -22,7 +22,6 @@ from cryptography.exceptions import UnsupportedAlgorithm
 from cryptography.hazmat.backends.interfaces import (
     DSABackend, EllipticCurveBackend, RSABackend, X509Backend
 )
-from cryptography.hazmat._oid import ObjectIdentifier
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import dsa, ec, padding, rsa
 from cryptography.hazmat.primitives.asymmetric.utils import (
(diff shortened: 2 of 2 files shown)


jcrowgey:
I didn't have everything necessary on my local box to run the tests, I pushed this PR in part so that I could see what travis would say. :)

I'll mark as WIP.

jcrowgey:
Unwipped. Respectfully awaiting feedback from the maintainers.

reaperhulk:
These RFCs are frustrating. 4514 references 4512 which says registration should be done according to 4520 (section 4: https://tools.ietf.org/html/rfc4520#section-4) which ultimately leads us to https://www.iana.org/assignments/ldap-parameters/ldap-parameters.xhtml#ldap-parameters-3 as the authoritative source of registered descriptors. Except while that may technically be the complete list, the reality is that far more than that are generally used. For example, the OID arc `1.3.6.1.4.1.311` is MSFT's PEN assignment and contains all the EV subject DN OIDs. They aren't registered as short names via the RFC 4520 process, but OpenSSL and others do use names like `jurisdictionStateOrProvinceName` in their text serialization.

I'm okay with the idea that we should extend our _NAMEOID_TO_NAME dictionary but I think falling back to calling `_name` is the wrong approach because it effectively makes that part of the public API (a thing we want to avoid for a variety of reasons).

Instead I'd rather we just add additional subject DN OIDs (like the EV OIDs) directly. This will result in some duplication but that's fine.

reaperhulk:
Are you interested in updating this PR with the proposed alternate approach?

jcrowgey:
Hi, yes, I definitely intended to follow up on your comments and I'm sorry I went MIA for so long.  I got pulled in several directions just shortly after you left this reply.  I hope it's not making things too cluttered up to have this open for a little while longer.

In terms of your specific points---I agree that chasing down the chain of references in these RFCs is a bit tedious and I wasn't sure that I had gotten things right even after several checks.  I agree with your analysis that the list of official names is many more than are used in practice---that was in part why I put the official OID for `favouriteDrink` in the test case I wrote---as a sort of easter egg for anyone that looks it up.

Here's a idea: what if the `._name` values were pulled from _NAMEOID_TO_NAME for those which exist (and maybe it would be okay to move most of the existing ones to that dictionary?).   That turns the dependency around---on the assumption that it's okay to expose the _NAMEOID_TO_NAME values (which happens through calls to repr --> rfc4514), and the private _name values just happen to use the same public ones.

My own interest in this is a json serialization library which wants to provide as many of the "friendly" names as possible.  I expose the dotted string values as well, but the names are certainly more memorable.  For my usage then, having names for the rare ones is actually quite an attractive thing.  I accept, however, that at the end of the day I may just need to record these constants in my own codebase, but once I realized how many of them you guys already have, wanted to try to take advantage of that.

In sum, if you only want the EV cert names in the NAMEOID_TO_NAME dict, I'll happy update this PR to fulfill that (and probably create some larger dictionary in my own codebase to accommodate as many of the missing friendly names as possible).  If you're interested at all in my idea of a refactor which turns the dependency around so that the private _name values come from that same dict, I'd be happy to looking into implementing that, instead.

Thanks again for all your work on pyca/cryptography---it's a very very useful package!

reaperhulk:
I'm going to go ahead and close this due to lack of activity, but if you're still interested feel free to submit again taking the feedback into account!

Issue #4798, opened 2019-03-07
add poly1305


reaperhulk:
OpenSSL 1.1.1 supports poly1305 as an EVP_PKEY now.

ronf:
I really appreciate you looking into this, and pointing out that vanilla ChaCha20 (without Poly1305) was previously added - I had missed that.

I actually have a local proof-of-concept of the AsyncSSH ChaCha20Poly1305 support working now using the existing PyCA ChaCha20() encryption support combined with direct calls to the internal libssl Poly1305_Init/Update/Final functions via ctypes. As soon as this is ready, I'll give it a shot.

One question: Is there any API for getting a raw ChaCha20 key stream, without XORing it against plain or cipher text? The way ChaCha20Poly1305 works is that it generates the Poly1305 key for each message by pulling out 256 bits of the raw ChaCha20 stream for a given key & nonce. I can accomplish this by passing 32 bytes of all-zero data into to a ChaCha20 encryptor() and using the resulting cipher text as my Poly1305 key, but I was wondering if there was a better way than this. In libnacl, there's a primitive which gets the ChaCha20 stream data directly, in addition to a primitive which lets you XOR it against something.

ronf:
Thanks very much - I can confirm that this is working for me. Using the raw ChaCha20 cipher and this feature, I'm able to implement an OpenSSH-compatible ChaCha20Poly1305 AEAD cipher (documented at https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.chacha20poly1305).

As an aside, have you considered supporting a one-shot API here (as well as possibly for other MAC functions) similar to what was done for asymmetric cipher sign/verify? I'm thinking of something like:

```python
    tag = Poly1305(key).sign(data)
```

This would be the equivalent of:

```python
    poly = Poly1305(key)
    poly.update(data)
    tag = poly.finalize()
```

Similarly, verify would look like:

```python
    Poly1305(key).verify(tag, data)
```

which would be the equivalent of:

```python
    poly = Poly1305(key)
    poly.update(data)
    poly.verify(tag)
```

The latter could be done by changing verify() to take an optional second argument which defaults to `None` that would do the update() call if present before performing the verify.

The finalize() function could also take an optional argument to pass to update() before doing the finalize, if you didn't want to add a new sign() function. I kind of like the symmetry of sign/verify better, though, and it matches what was done for asymmetric key signing.

reaperhulk:
[...]

PR #4802, opened 2019-03-08
poly1305 support

reaperhulk:
Update: discussion complete, we got rid of maccontext.

This PR marked as draft for now because I want to discuss whether we should implement `copy`. Poly1305 is a one-time authenticator so any key reuse for multiple tags can result in tag forgery. Accordingly, I suggest that `copy` not be a thing we implement. However, do we want to claim this conforms to `MACContext` then (and have `copy` raise `NotImplementedError`) or just not register it? And if we don't register it, should we consider just getting rid of MACContext and having HMAC and CMAC happen to use the same API?

refs #4798

### DIFF
--- a/docs/hazmat/primitives/mac/poly1305.rst
+++ b/docs/hazmat/primitives/mac/poly1305.rst
@@ -1,0 +1,85 @@
+.. hazmat::
+
+Poly1305
+========
+
+.. currentmodule:: cryptography.hazmat.primitives.poly1305
+
+.. testsetup::
+
+    key = b"\x01" * 32
+
+Poly1305 is a one-time authenticator that takes a 32-byte one-time key and a
+message and produces a 16-byte tag.  This tag is used to authenticate the
+message. It is described in :rfc:`7539`.
+
+.. class:: Poly1305(key)
+
+    .. versionadded:: 2.7
+
+    .. warning::
+
+        Reuse of a key potentially allows forgery of authentication tags. Always
+        create a new key for each tag. If you are using this as a MAC for
+        symmetric encryption please use
+        :class:`~cryptography.hazmat.primitives.ciphers.aead.ChaCha20Poly1305`
+        instead.
+
+    .. doctest::
+
+        >>> from cryptography.hazmat.primitives import poly1305
+        >>> p = poly1305.Poly1305(key)
+        >>> p.update(b"message to authenticate")
+        >>> p.finalize()
+        b'T\xae\xff3\xbdW\xef\xd5r\x01\xe2n=\xb7\xd2h'
+
+    To check that a given signature is correct use the :meth:`verify` method.
+    You will receive an exception if the signature is wrong:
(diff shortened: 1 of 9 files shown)


alex:
I would not object to dropping MACContext, it seems pretty dumb.

src/cryptography/hazmat/backends/openssl/poly1305.py
```diff
@@ -0,0 +1,58 @@
...
+
+_POLY1305_TAG_SIZE = 16
+_POLY1305_KEY_SIZE = 32
+
+
+class _Poly1305Context(object):
+    def __init__(self, backend, key):
+        self._backend = backend
+
+        key_ptr = self._backend._ffi.from_buffer(key)
```
alex: Don't you need to keep this alive? I suspect you can get a UAF right now with `k = Poly1305(bytearray(range(16)); k.finalize()`

reaperhulk: Ugh, I refactored this for something else and didn't store the key itself. Yeah this can definitely UAF

alex: add a test please

docs/hazmat/primitives/mac/poly1305.rst
```diff
@@ -0,0 +1,85 @@
...
+========
+
+.. currentmodule:: cryptography.hazmat.primitives.poly1305
+
+.. testsetup::
+
+    key = b"\x01" * 32
+
+Poly1305 is a one-time authenticator that takes a 32-byte one-time key and a
+message and produces a 16-byte tag.  This tag is used to authenticate the
```
alex: Define one-time right up front

docs/hazmat/primitives/mac/poly1305.rst
```diff
@@ -0,0 +1,85 @@
...
+message. It is described in :rfc:`7539`.
+
+.. class:: Poly1305(key)
+
+    .. versionadded:: 2.7
+
+    .. warning::
+
+        Reuse of a key potentially allows forgery of authentication tags. Always
```
alex: I think this reads more clearly: "Using the same key to sign multiple messages allows an attacker to forge future tags" or something

docs/hazmat/primitives/mac/poly1305.rst
```diff
@@ -0,0 +1,85 @@
...
+        :type data: :term:`bytes-like`
+        :raises cryptography.exceptions.AlreadyFinalized: See :meth:`finalize`
+        :raises TypeError: This exception is raised if ``data`` is not ``bytes``.
+
+    .. method:: verify(tag)
+
+        Finalize the current context and securely compare the MAC to
+        ``tag``.
+
+        :param bytes tag: The bytes to compare the current CMAC
```
alex: nope

PR #4803, opened 2019-03-08
remove maccontext

reaperhulk:
It's not really useful and we don't want Poly1305 to use it (#4802) so let's get rid of it.

This is technically backwards incompatible so I have documented it as such, but in reality it's very unlikely a single user will break due to this change.


reaperhulk:
Dunno why it didn't report, but https://codecov.io/gh/pyca/cryptography/tree/e76943fd8a882645b851aa43738915d22ce8fc8b shows 100% coverage

PR #4804, opened 2019-03-08
Improve deprecation warning to specify the release

jsoref:
https://github.com/certbot/certbot/issues/6827#issuecomment-470981623

The warning yielded by this message was very confusing in the context of certbot-auto. It seemed to indicate I should upgrade to python 2.7, but as I was already using python 2.7 (via redhat scl), I, and others misunderstood what it was trying to say.


alex:
The reason we didn't write an exact version there is that debian backported hmac.compare_digest to an earlier version iirc

alex:
josepy test failures aren't your fault, they're https://github.com/certbot/josepy/pull/47

alex:
Thanks!

PR #4806, opened 2019-03-10
Make Fernet.generate_key a staticmethod

bsolomon1124:
Change from `@classmethod` to `@staticmethod`.

Method contains no reference to the class (or class instance), so better fits the bill of a staticmethod, which also makes the docs clearer.

### DIFF
--- a/src/cryptography/fernet.py
+++ b/src/cryptography/fernet.py
@@ -42,8 +42,9 @@ class Fernet(object):
         self._encryption_key = key[16:]
         self._backend = backend
 
-    @classmethod
-    def generate_key(cls):
+    @staticmethod
+    def generate_key():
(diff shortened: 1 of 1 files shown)


alex:
Is there a functional reason to make this change?

bsolomon1124:
Frankly, no--it's grammatical, for lack of a better word.  Denoting it a staticmethod makes it more clear what it's designed to do and what it returns.  A classmethod (because it takes the class as its first argument implicitly) typically has a _reason_ to use that object within the method's scope, which is not the case here.

alex:
Ok. I think it's preferable to leave this as is -- it gives us the flexibility to do things like invoke other class methods or use class constants (and frankly factoring out key length into a constant wouldn't be the worst idea).

Thanks for the PR though!

bsolomon1124:
Okay, fair - your call!

Issue #4767, opened 2019-02-25
Need support for ECC Brainpool and Koblitz curves. 


deepakbairy:
New curve support required:
Brainpool160
 Brainpool192
 Brainpool224
 Brainpool320
 Secp160k1
 Secp192k1
 Secp224k1

reaperhulk:
What is the use case for the curves you list as "new curve support required"?

deepakbairy:
I use these curves for IOT security. Currently I am using this python cryptography module for loading ecc keys. SECP or NIST_P curves are working perfectly, but I need few more curves in brain pool and koblitz.

reaperhulk:
Could you give us some more details? What spec or product is utilizing these curves? And for what? ECDH?

deepakbairy:
I am using a hardware module where I am provisioning keys for the above mentioned curves. These keys are used during ECDH.

alex:
_Why_ did you select these curves though? Is there some external specification or interoeprability requirement to use them?

deepakbairy:
I am trying all the ECC curves on the hardware module. Where All other curves like NIST_P256, 384, 521, Brainpool 256, Koblitz 256 are working. Missing curves are listed in the requirement. 
 
https://datatracker.ietf.org/doc/rfc5639/?include_text=1 
refer section 5 table 1 for brain pool curves.

reaperhulk:
The brainpool (and koblitz) curves are not widely used nor significantly better than our existing curves and we have no plans to expand support for them outside of specific use cases. The set of brainpool curves we currently support is only exposed specifically to support German smart meter requirements. Similarly, we expose `secp256k1` specifically because of its importance in the cryptocurrency ecosystem.

For security purposes in environments without legacy interoperability requirements X25519, X448, Ed25519, and Ed448 are strongly preferred over older curves anyway.

Any additional curve support we choose to add is predicated on the existence of specifications that require those curves.

alex:
It's also worth mentioning that most of the curves you listed are <128 bit security.

deepakbairy:
I have updated cryptography\hazmat\primitives\asymmetric\ec.py file with adding curves. Brainpool160, Brainpool192, Brainpool224, Brainpool320 and Secp192k1 started working.
I would like to contribute upadted contents to cryptography. Is it possible to add that ?

reaperhulk:
[...]

PR #4832, opened 2019-04-05
Added support for Koblitz and Brainpool Curve

purnankNXP:
Support Koblitz and Brainpool Curve

### DIFF
--- a/src/cryptography/hazmat/primitives/asymmetric/ec.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/ec.py
@@ -14,13 +14,20 @@ from cryptography.hazmat._oid import ObjectIdentifier
 
 
 class EllipticCurveOID(object):
+    SECP160K1 = ObjectIdentifier("1.3.132.0.9")
     SECP192R1 = ObjectIdentifier("1.2.840.10045.3.1.1")
+    SECP192K1 = ObjectIdentifier("1.3.132.0.31")
     SECP224R1 = ObjectIdentifier("1.3.132.0.33")
+    SECP224K1 = ObjectIdentifier("1.3.132.0.32")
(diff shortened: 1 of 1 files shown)


reaperhulk:
Thanks for your effort here, but we recently discussed this in https://github.com/pyca/cryptography/issues/4767 and unless there's new information we're still not interested in exposing additional curves. Here's what I said in that issue:

The brainpool (and koblitz) curves are not widely used nor significantly better than our existing curves and we have no plans to expand support for them outside of specific use cases. The set of brainpool curves we currently support is only exposed specifically to support German smart meter requirements. Similarly, we expose secp256k1 specifically because of its importance in the cryptocurrency ecosystem.

For security purposes in environments without legacy interoperability requirements X25519, X448, Ed25519, and Ed448 are strongly preferred over older curves anyway.

Any additional curve support we choose to add is predicated on the existence of specifications that require those curves.

purnankNXP:
Sorry for replying late. These curves are used in small/micro embedded systems. 

As of now, for me, the `cryptography` python library serves two purposes.

1. Test counter part from PC/Linux/OS X in a high level language (python) to test cryptography of the counterparts small/micro embedded systems
2. Library for tools for cryptographic operation (inject/read keys) from PC/Linux/OS X into the small/micro embedded systems. 

With this patch, we are tying to expose the underlying `OpenSSL` implementation to the python library interface level.  If, in your opinion, it is not recommended to expose these curves to the core `cryptography` library, then I would try to investigate a way to handle this within my tools and test benches withe some other run time mechanism. 

What do you suggest?

reaperhulk:
Hmm, are these widely used embedded systems? At the moment I'd suggest just carrying this patch yourself as it will work fine for your purposes but won't expose curves that no one else needs, but if these are common systems we'd like to know so we can make a more informed choice.

PR #4857, opened 2019-04-25
Add add_attribute to CSR Builder

willemt:
Add add_attribute method to CertificateSigningRequestBuilder so that
it's possible to add attributes to CSRs.

Example usage:
```python
    from asn1crypto import x509 as asn1x509
    from cryptography import x509

    builder = x509.CertificateSigningRequestBuilder()
    builder = builder.add_attribute(
        x509.ObjectIdentifier('1.2.840.113549.1.9.7'),
        asn1x509.PrintableString(challenge))

```

Still needs a test. Want to please ask if the API is OK?

### DIFF
--- a/src/_cffi_src/openssl/x509.py
+++ b/src/_cffi_src/openssl/x509.py
@@ -212,6 +212,9 @@ X509_EXTENSION *sk_X509_EXTENSION_delete(X509_EXTENSIONS *, int);
 void sk_X509_EXTENSION_free(X509_EXTENSIONS *);
 void sk_X509_EXTENSION_pop_free(X509_EXTENSIONS *, sk_X509_EXTENSION_freefunc);
 
+int X509_REQ_add1_attr_by_txt(X509_REQ *, const char *, int,
+                              const unsigned char *, int);
+
 int sk_X509_REVOKED_num(Cryptography_STACK_OF_X509_REVOKED *);
 X509_REVOKED *sk_X509_REVOKED_value(Cryptography_STACK_OF_X509_REVOKED *, int);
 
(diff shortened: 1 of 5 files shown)


reaperhulk:
We don't want to tie our public APIs to objects constructed via asn1crypto. We need a better approach here. We already have a private enum called `_ASN1Type` https://github.com/pyca/cryptography/blob/9a22851fab924fd58482fdad3f8dd23dc3987f91/src/cryptography/x509/name.py#L15-L25
We could potentially make that public. However, we generally try to hide the need to explicitly set the underlying ASN1 string type since it's incredibly unfriendly. For `Name` we set a default type by OID:
https://github.com/pyca/cryptography/blob/9a22851fab924fd58482fdad3f8dd23dc3987f91/src/cryptography/x509/name.py#L30-L37

It would be very nice if we could make an API that looks similar to that or `add_extension`. What is typically encoded as an attribute? Are there common OIDs/objects such that we can make this simple in the common case and keep a private API (like `Name` has) for alternate asn1 string types?

darko1002001:
+1 on this one as well. I have the same use case, the ability to add a challengePassword to the CSR. In my use-case this is used as a shared secret nonce when requesting a certificate using a CSR from my backend. This seems it's quite near to being fully implemented, I can maybe help with the implementation if its not something the maintainers or the the original creator of the pull request is able to do in the near future.

willemt:
thanks @reaperhulk. I'll have a look and see if I can do it that way. I think 99% of the use of this feature is to add `challengePassword`, so that approach you mentioned makes sense.

@darko1002001 this is less than half way here. We need to add `get_attribute` (and friends possibly) to be able to add a test for this. I'm a little busy so can't work on this immediately

alex:
implemented in #5315

PR #4871, opened 2019-05-04
Replace debian wheezy with ubuntu trusty

alex:


alex:
This hopefully gets us a _slightly_ less EOL thing with a Python that doesn't have `hmac.compare_digest`

PR #4878, opened 2019-05-07
Update release.py to use azure for wheel building

alex:


alex:
Ok, this code is working now. This shouldn't be merged until we have wheel builders for all supported targets on azure though.

PR #4881, opened 2019-05-08
Add windows to azure wheel builder

alex:


alex:
Wheels built: https://dev.azure.com/pyca/cryptography/_build/results?buildId=360

PR #4913, opened 2019-06-03
fixed broken random order

alex:


reaperhulk:
So does randomly only reorder within a test file and not across all tests?

alex:
I believe it also reorders the files themselves

PR #4914, opened 2019-06-05
Switch to new notBefore/After APIs

neheb:
Introduced in OpenSSL 1.1. Added compatibility for older versions.


alex:
In order to not break pyOpenSSL, you'll need to add the new APIs, but not remove the old ones, and only remove the old ones after everything has had a release.

neheb:
Fixed

PR #4917, opened 2019-06-05
[WIP] Improve performance for repeated AEAD encryption / decryption

jlaine:
In many usecases it is possible to retain the cipher context between
calls to encrypt() and decrypt(), avoiding the cost of repeatedly
setting up and tearing down the context.


jlaine:
Indeed the AEAD can no longer be used concurrently.

I'm going to close this PR as my specialized use case is better served by tapping straight into OpenSSL.

PR #4931, opened 2019-06-28
Fix some backend feature checks in tests

davidben:
A bunch of tests have irrelevant `DHBackend` checks, and some EC and AES tests are missing checks for the curve and mode.

### DIFF
--- a/tests/hazmat/primitives/test_aes.py
+++ b/tests/hazmat/primitives/test_aes.py
@@ -13,6 +13,7 @@ from cryptography.hazmat.backends.interfaces import CipherBackend
 from cryptography.hazmat.primitives.ciphers import algorithms, base, modes
 
 from .utils import _load_all_params, generate_aead_test, generate_encrypt_test
+from ...doubles import DummyMode
 from ...utils import load_nist_vectors
 
 
@@ -484,6 +485,8 @@ class TestAESModeGCM(object):
         modes.CFB(bytearray(b"\x00" * 16)),
         modes.CFB8(bytearray(b"\x00" * 16)),
         modes.XTS(bytearray(b"\x00" * 16)),
+        # Add a dummy mode for coverage of the cipher_supported check.
+        DummyMode(),
     ]
 )
 @pytest.mark.requires_backend_interface(interface=CipherBackend)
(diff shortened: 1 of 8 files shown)


tests/hazmat/primitives/test_aes.py
```diff
@@ -489,9 +489,10 @@ def test_buffer_protocol(self, backend):
 @pytest.mark.requires_backend_interface(interface=CipherBackend)
 def test_buffer_protocol_alternate_modes(mode, backend):
     data = bytearray(b"sixteen_byte_msg")
-    cipher = base.Cipher(
-        algorithms.AES(bytearray(os.urandom(32))), mode, backend
-    )
+    key = algorithms.AES(bytearray(os.urandom(32)))
+    if not backend.cipher_supported(key, mode):
```
reaperhulk: This branch isn't covered in our CI (although codecov can't be bothered to actually report back to say that for some reason). Is this something you need?

davidben: It would make my life easier, but if it's not consistent with your goals, I can carry it on my end. I figured, based on every other test including cipher support checks, it was intentional to feature-guard most of everything. E.g., here's an existing one for AES modes:
https://github.com/pyca/cryptography/blob/31729119b5bb112bcc874c3086613c14397b84fc/tests/hazmat/primitives/test_aes.py#L152

Just that one, being a decorator, probably isn't going to read as missing line coverage. I wasn't sure how to do it as a decorator for a parameterized test.

alex: You should just be able to stack the decorators -- you can use a all-zero key in the decorator test, as we do in a bunch of other places.

davidben: How do I get the mode into the decorator? It looks like it's just passed the backend. (Or is there some magic introspecting parameter names thing going on? I don't really know how pytest works but I see adding parameters did weird things elsewhere.)

alex: Oh shoot. There's currently no way to get it in there. Hmmm

reaperhulk: We could just add the DummyCipherAlgorithm from `tests/doubles.py` to the list here and it would cover this branch.

reaperhulk: I meant DummyMode of course. This is a dumb hack, but it lets us carry the patch with effectively no cost and keeps us at full coverage so ¯\_(ツ)_/¯

davidben: Done.

PR #4941, opened 2019-07-07
Remove asn1crypto dependency

davidben:
cryptography.io actually contains two OpenSSL bindings right now, the expected cffi one, and an optional one [hidden in asn1crypto](https://github.com/wbond/asn1crypto/blob/master/asn1crypto/_perf/_big_num_ctypes.py). asn1crypto contains a lot of things that cryptography.io doesn't use, including a BER parser and a hand-rolled and (not constant-time) [EC implementation](https://github.com/wbond/asn1crypto/blob/master/asn1crypto/_elliptic_curve.py), which is where the bindings are used. (Though asn1crypto itself could likely avoid the bindings by using Jacobian coordinates, which reduces the sensitivity to mod inverse perf.)

Instead, check in a much small DER-only parser in cryptography/hazmat. A quick benchmark suggests this parser is also faster than asn1crypto:

```
import timeit

print(timeit.timeit(
    "decode_dss_signature(sig)",
    setup=r"""
from cryptography.hazmat.primitives.asymmetric.utils import decode_dss_signature
  sig=b"\x30\x2d\x02\x15\x00\xb5\xaf\x30\x78\x67\xfb\x8b\x54\x39\x00\x13\xcc\x67\x02\x0d\xdf\x1f\x2c\x0b\x81\x02\x14\x62\x0d\x3b\x22\xab\x50\x31\x44\x0c\x3e\x35\xea\xb6\xf4\x81\x29\x8f\x9e\x9f\x08"
""",
    number=10000))
```
Python 2.7:
* asn1crypto: 0.25
* _der.py: 0.098

Python 3.5:
* asn1crypto: 0.17
* _der.py: 0.10

**Notes**

The last time I wrote a significant amount of Python was before anyone used Python 3, so this is likely unidiomatic. Please feel free to thoroughly critique this on style, etc. I'm happy to rework this as needed. (`DERReader.read_any_element` is the only actually interesting part of this PR. The rest is window dressing.)

Additionally, I added support for negative integers in `DERReader.as_integer` and `encode_der_integer`, which required adding some features to `int_from_bytes` and mucking about with variable-length two's complement. I have never seen a negative ASN.1 INTEGER in my life. It is not actually necessary in any of these uses, except that `decode_dss_signature` is public API and has tests asserting support for negative integers. If you don't care about that, the INTEGER encoding can be simpler.

Finally, I believe I added tests to cover all the newly-added cases, but I wasn't sure how to trigger your coverage report, so I probably missed some. Point those out and I'll add the missing tests.

### DIFF
--- a/src/cryptography/hazmat/_der.py
+++ b/src/cryptography/hazmat/_der.py
@@ -1,0 +1,144 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+from __future__ import absolute_import, division, print_function
+
+import six
+
+from cryptography.utils import int_from_bytes, int_to_bytes
+
+
+# This module contains a lightweight DER encoder and decoder. See X.690 for the
+# specification. This module intentionally does not implement the more complex
+# BER encoding, only DER.
+#
+# Note this implementation treats an element's constructed bit as part of the
+# tag. This is fine for DER, where the bit is always computable from the type.
+
+
+CONSTRUCTED = 0x20
+CONTEXT_SPECIFIC = 0x80
+
+INTEGER = 0x02
+BIT_STRING = 0x03
+OCTET_STRING = 0x04
+NULL = 0x05
+OBJECT_IDENTIFIER = 0x06
+SEQUENCE = 0x10 | CONSTRUCTED
+SET = 0x11 | CONSTRUCTED
+PRINTABLE_STRING = 0x13
+UTC_TIME = 0x17
+GENERALIZED_TIME = 0x18
+
+
+class DERReader(object):
+    def __init__(self, data):
+        self.data = memoryview(data)
+
+    def is_empty(self):
+        return len(self.data) == 0
+
+    def check_empty(self):
+        if not self.is_empty():
+            raise ValueError("Invalid DER input: trailing data")
+
+    def read_byte(self):
+        if len(self.data) < 1:
+            raise ValueError("Invalid DER input: insufficient data")
+        ret = six.indexbytes(self.data, 0)
+        self.data = self.data[1:]
+        return ret
+
+    def read_bytes(self, n):
+        if len(self.data) < n:
+            raise ValueError("Invalid DER input: insufficient data")
+        ret = self.data[:n]
+        self.data = self.data[n:]
+        return ret
+
+    def read_any_element(self):
+        tag = self.read_byte()
+        # Tag numbers 31 or higher are stored in multiple bytes. No supported
+        # ASN.1 types use such tags, so reject these.
+        if tag & 0x1f == 0x1f:
+            raise ValueError("Invalid DER input: unexpected high tag number")
+        length_byte = self.read_byte()
+        if length_byte & 0x80 == 0:
+            # If the high bit is clear, the first length byte is the length.
+            length = length_byte
+        else:
+            # If the high bit is set, the first length byte encodes the length
+            # of the length.
+            length_byte &= 0x7f
+            if length_byte == 0:
+                raise ValueError(
+                    "Invalid DER input: indefinite length form is not allowed in DER"
+                )
+            length = 0
+            for i in range(length_byte):
+                length <<= 8
+                length |= self.read_byte()
+                if length == 0:
+                    raise ValueError(
+                        "Invalid DER input: length was not minimally-encoded"
+                    )
+            if length < 0x80:
+                # If the length could have been encoded in short form, it must
+                # not use long form.
+                raise ValueError(
+                    "Invalid DER input: length was not minimally-encoded"
+                )
+        body = self.read_bytes(length)
+        return tag, DERReader(body)
+
+    def read_element(self, expected_tag):
+        tag, body = self.read_any_element()
+        if tag != expected_tag:
+            raise ValueError("Invalid DER input: unexpected tag")
+        return body
+
+    def read_optional_element(self, expected_tag):
+        if len(self.data) > 0 and six.indexbytes(self.data, 0) == expected_tag:
+            return self.read_element(expected_tag)
+        return None
+
+    def as_integer(self):
+        if len(self.data) == 0:
+            raise ValueError("Invalid DER input: empty integer contents")
+        # The first 9 bits must not all be zero or all be ones. Otherwise, the
+        # encoding should have been one byte shorter.
+        if len(self.data) > 1:
+            first = six.indexbytes(self.data, 0)
+            second = six.indexbytes(self.data, 1)
+            if (
+                (first == 0 and second & 0x80 == 0) or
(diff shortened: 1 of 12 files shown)


davidben:
Updated to (hopefully) address flake8 warnings.

src/cryptography/hazmat/_der.py
```diff
@@ -0,0 +1,152 @@
...
+                length <<= 8
+                length |= self.read_byte()
+                if length == 0:
+                    raise ValueError(
+                        "Invalid DER input: length was not minimally-encoded"
+                    )
+            if length < 0x80:
+                # If the length could have been encoded in short form, it must
+                # not use long form.
+                raise ValueError(
```
reaperhulk: This branch isn't taken in a test.

davidben: Oh oops. I'd meant to cover that with this test, but since the length is zero, it hits line 183 instead. Will fix and remove the negative number support.
https://github.com/pyca/cryptography/pull/4941/files#diff-e31fedf020e23b010b695ba39b70342bR144

davidben: Done.

davidben:
One red bot, but it looks like it may be an infra issue?
https://travis-ci.org/pyca/cryptography/jobs/561915667

reaperhulk:
Looks like an ephemeral failure. I just rebuilt and it passed.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -994,11 +995,18 @@ def _create_x509_extension(self, handlers, extension):
...
+                    encode_der(INTEGER, encode_der_integer(x.value))
+                    for x in extension.value
+                ]
+            )
             value = _encode_asn1_str_gc(self, asn1)
             return self._create_raw_x509_extension(extension, value)
         elif isinstance(extension.value, x509.PrecertPoison):
-            asn1 = asn1crypto.core.Null().dump()
+            # The contents are an ASN.1 NULL.
+            asn1 = b'\x05\x00'
```
alex: Should this be `encode_der(NULL)`? It's slightly silly since it constant folds, on the other hand it makes the code self-documenting.

src/cryptography/hazmat/primitives/asymmetric/utils.py
```diff
@@ -4,27 +4,27 @@
...
-        not isinstance(r, six.integer_types) or
-        not isinstance(s, six.integer_types)
-    ):
-        raise ValueError("Both r and s must be integers")
-
-    return DSASignature({'r': r, 's': s}).dump()
+    return encode_der(
+        SEQUENCE,
+        encode_der(INTEGER, encode_der_integer(r)),
+        encode_der(INTEGER, encode_der_integer(s))
```
alex:
```suggestion
        encode_der(INTEGER, encode_der_integer(s)),
```

src/cryptography/x509/extensions.py
```diff
@@ -47,7 +48,20 @@ def _key_identifier_from_public_key(public_key):
...
         )
 
-        data = bytes(PublicKeyInfo.load(serialized)['public_key'])
+        public_key_info = DERReader(serialized).read_single_element(SEQUENCE)
+        algorithm = public_key_info.read_element(SEQUENCE)
+        public_key = public_key_info.read_element(BIT_STRING)
+        public_key_info.check_empty()
+
+        # Double-check the algorithm structure.
+        algorithm.read_element(OBJECT_IDENTIFIER)
```
alex: Should we `algorithm.check_empty()`?

tests/x509/test_x509.py
```diff
@@ -63,6 +65,47 @@ def _load_cert(filename, loader, backend):
...
+    _ = tbs_cert.read_optional_element(CONTEXT_SPECIFIC | CONSTRUCTED | 2)
+    # Skip extensions
+    _ = tbs_cert.read_optional_element(CONTEXT_SPECIFIC | CONSTRUCTED | 3)
+    tbs_cert.check_empty()
+
+    not_before_tag, _ = validity.read_any_element()
+    not_after_tag, _ = validity.read_any_element()
+    validity.check_empty()
+
+    return {
```
alex: Can we make a named tuple or something for this structure. Using dicts to represent objects is a big anti-pattern in my view (and why tons of Python code is slow).

src/cryptography/hazmat/backends/openssl/backend.py
davidben: Done.

src/cryptography/x509/extensions.py
davidben: AlgorithmIdentifiers end with an OPTIONAL parameters element. I've added some code to parse that down a later and then check_empty.

tests/x509/test_x509.py
davidben: Done.

src/cryptography/hazmat/backends/openssl/backend.py
davidben: Avoided hardcoding things in decode_asn1.py too, which is actually shorter anyway.

nickwilliams-eventbrite:
asn1crypto 1.0.0 was released today with this changelog:

https://github.com/wbond/asn1crypto/blob/master/changelog.md#100

They have removed several items from Asn1crypto prior to the release. Notably, I don't see `_perf/_big_num_ctypes.py` anymore. Since this pull request hasn't yet been released (the current Cryptography in PyPi still depends on `asn1crypto>=0.21.0`, which is pulling in 1.0.0 in our Continuous Integration builds), will the asn1crypto release break anything in Cryptography? Or should it still work with the new asn1crypto release?

If it's going to break, how soon can this be ready for release? If not soon, would it be possible to get a Cryptography 2.7.1 patch release that changes the dependency from `asn1crypto>=0.21.0` to `asn1crypto>=0.21.0,<1.0.0`?

nickwilliams-eventbrite:
Wanted to provide a little update. I cloned Cryptography and checked out 2.7 and run the Tox test suite. In the output it shows it installing `asn1crypto==1.0.0`, and every test passed. So I _think_ we're good, but if someone else sees a problem, by all means, let us know. :-)

davidben:
Hah. I guess if I'd just waited, asn1crypto would have removed the problematic bits on their own. Ah well. Now cryptography.io has one fewer dependency, so hopefully that's useful.

nickwilliams-eventbrite:
Quite :-) ... Out of curiosity, do you have an anticipated timeline for version 2.8 (is that what's next)?

reaperhulk:
The blocker is core developers are a bit burned out at the moment due to issues with the way openssl 1.1.1d broke some wycheproof tests. Decisions/fixes/bugs need to be filed and all of that takes investigation 😔. Ideally it will be in October.

wbond:
Just for historical reference, asn1crypto 1.0.0 removed the optional dependency on libcrypto and also removed the private key -> public key computation code.

PR #4943, opened 2019-07-08
add bindings to parse and create challenge passwords in X509 CSRs

reaperhulk:
We, unfortunately, have to partially de-opaque the ASN1_TYPE struct to read the type out of it because of the way the parsing works here. Alternately we could define our own getter, but that struct is still public even in the latest openssl so this approach seems fine.

### DIFF
--- a/src/_cffi_src/openssl/x509.py
+++ b/src/_cffi_src/openssl/x509.py
@@ -29,6 +29,7 @@ typedef struct {
     ...;
 } X509_ALGOR;
 
+typedef ... X509_ATTRIBUTE;
 typedef ... X509_EXTENSION;
 typedef ... X509_EXTENSIONS;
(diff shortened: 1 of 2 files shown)


src/_cffi_src/openssl/x509.py
```diff
@@ -258,6 +259,12 @@
 int i2d_re_X509_CRL_tbs(X509_CRL *, unsigned char **);
 void X509_REQ_get0_signature(const X509_REQ *, const ASN1_BIT_STRING **,
                              const X509_ALGOR **);
+X509_ATTRIBUTE *X509_REQ_get_attr(const X509_REQ *, int);
```
alex: You have these grouped with 1.1.0 functions. Since these aren't, please move or seperate.

PR #4955, opened 2019-07-27
fix osrandom/builtin switching methods for 1.1.0+

reaperhulk:
In 1.1.0 RAND_cleanup became a no-op. This broke changing to the builtin random engine via activate_builtin_random(). Fixed by directly calling RAND_set_rand_method. This works on 1.0.x and 1.1.x

Testing this is very, very painful. I believe to test this we need to create, register, and activate a bad RNG that we can use to detect whether it has actually switched between engine and builtin. This will require adding quite a bit of C code since the RAND_METHOD and callbacks are all opaque to the Python layer. I've written most of this, but it feels very bad to be adding an engine like this (even if we only invoke its addition during tests). Opinions?

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -170,6 +170,7 @@ class Backend(object):
                 self.openssl_assert(res == 1)
             # Reset the RNG to use the engine
             res = self._lib.RAND_set_rand_method(self._ffi.NULL)
+            self.openssl_assert(res == 1)
 
     def osrandom_engine_implementation(self):
         buf = self._ffi.new("char[]", 64)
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -167,8 +168,8 @@ def activate_osrandom_engine(self):
                 # Set the engine as the default RAND provider.
                 res = self._lib.ENGINE_set_default_RAND(e)
                 self.openssl_assert(res == 1)
-            # Reset the RNG to use the new engine.
-            self._lib.RAND_cleanup()
+            # Reset the RNG to use the engine
+            res = self._lib.RAND_set_rand_method(self._ffi.NULL)
```
alex: I do not understand this at all.

And unchecked `res`

reaperhulk: Take a look at the source code and it'll sort of make sense. Or, if not, look at the implementation of `RAND_cleanup` in 1.0.x and you'll see this was already called.

I'll add the assert momentarily.

PR #4963, opened 2019-08-01
parse SCTs in OCSPResponse

c0r0n3r:


src/cryptography/x509/extensions.py
```diff
@@ -1483,6 +1480,16 @@ def __ne__(self, other):
         return not self == other
 
 
+@utils.register_interface(ExtensionType)
+class PrecertificateSignedCertificateTimestamps(SCTBase):
```
reaperhulk: I realize these are only being used to change the name of the class and the OID associated, but despite that I'm not wild about having a subclass. We have been very successful at not subclassing anything in this project. @alex do you have an opinion?

tests/x509/test_ocsp.py
```diff
@@ -753,3 +753,26 @@ def test_invalid_serialize_encoding(self):
             resp.public_bytes("invalid")
         with pytest.raises(ValueError):
             resp.public_bytes(serialization.Encoding.PEM)
+
+    @pytest.mark.supported(
+        only_if=lambda backend: (
+            backend._lib.CRYPTOGRAPHY_OPENSSL_110F_OR_GREATER),
+        skip_message="Requires OpenSSL 1.1.0f+",
+    )
+    def test_single_extensions(self, backend):
```
reaperhulk: Could you add one more test that asserts on the result of parsing this response when on OpenSSL < 1.1.0f?

src/cryptography/x509/extensions.py
alex: I agree with you.

reaperhulk: @alex so what do we want to do here. Duplicate this class and its attendant tests?

alex: That's roughly what we do for all of the "this is an iterable" extensions, yes. If we want to think about reducing that duplication, there are things we could consider.

PR #4965, opened 2019-08-08
Add SSL_get0_verified_chain to cffi lib

arjenzorgdoc:
OpenSSL 1.1.0 supports SSL_get0_verified_chain. This gives the full chain from the peer cert including your trusted CA cert.


reaperhulk:
You can't put a conditional in the `FUNCTIONS` block unfortunately. Instead you'll need to follow the pattern we use for conditional bindings where we declare a NULL func pointer if the version is too old and then add it into `_conditional.py` to remove it from the runtime `lib` object. The file you modified has several examples of this.

PR #4966, opened 2019-08-08
Possibly build macOS + PyPy wheels.

Julian:


reaperhulk:
I know I asked you to submit this, but we're not likely to get to this any time soon. Next up on priority is probably aarch64 manylinux and then apple silicon. If a pypy enthusiast wants to help out here and also can confirm that pypy wheels are possible (stable ABI, tags set up with wheel and pip, pypi allowing uploads) we can revisit.

PR #4969, opened 2019-08-14
restore apis needed for a _ssl module when using cryptograhpy v2.7

mattip:
PyPy uses cryptography to implement a pure-python _ssl stdlib module. After upgrading to v2.7 of cryptography, it turns out some API wrappers were removed that PyPy needed.

@alex suggested we collaborate to ensure future compatibility. I am not sure you want all the cruft of the python _ssl implementation in this repo. Perhaps it makes more sense to provide a new pyca/_ssl repo that would include the tests from the cpython stdlib test suite?

### DIFF
--- a/src/_cffi_src/openssl/pem.py
+++ b/src/_cffi_src/openssl/pem.py
@@ -44,6 +44,8 @@ X509_REQ *PEM_read_bio_X509_REQ(BIO *, X509_REQ **, pem_password_cb *, void *);
 
 X509_CRL *PEM_read_bio_X509_CRL(BIO *, X509_CRL **, pem_password_cb *, void *);
 
+X509 *PEM_read_bio_X509_AUX(BIO *, X509 **, pem_password_cb *, void *);
+
 int PEM_write_bio_X509_CRL(BIO *, X509_CRL *);
 
 PKCS7 *PEM_read_bio_PKCS7(BIO *, PKCS7 **, pem_password_cb *, void *);
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -621,9 +621,9 @@ static const long Cryptography_HAS_SSL2 = 0;
 
 #ifdef OPENSSL_NO_SSL3_METHOD
 static const long Cryptography_HAS_SSL3_METHOD = 0;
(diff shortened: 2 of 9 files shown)


src/_cffi_src/openssl/ssl.py
```diff
@@ -621,9 +621,9 @@
 
 #ifdef OPENSSL_NO_SSL3_METHOD
 static const long Cryptography_HAS_SSL3_METHOD = 0;
-SSL_METHOD* (*SSLv3_method)(void) = NULL;
-SSL_METHOD* (*SSLv3_client_method)(void) = NULL;
-SSL_METHOD* (*SSLv3_server_method)(void) = NULL;
+const SSL_METHOD* (*SSLv3_method)(void) = NULL;
+const SSL_METHOD* (*SSLv3_client_method)(void) = NULL;
+const SSL_METHOD* (*SSLv3_server_method)(void) = NULL;
```
mattip: bug fix

src/_cffi_src/openssl/pem.py
```diff
@@ -44,6 +44,8 @@
 
 X509_CRL *PEM_read_bio_X509_CRL(BIO *, X509_CRL **, pem_password_cb *, void *);
 
+X509 *PEM_read_bio_X509_AUX(BIO *, X509 **, pem_password_cb *, void *);
```
alex: Can you drop this one. With https://github.com/python/cpython/pull/15303's merge, it should be possibel to use `PEM_read_bio_X509` in cryptography, which we already have.

mattip: Fixed in pypy 7aed2cf1c9c0.

mattip:
PyPy also has `pypy_win32_extra.py`. It seems this is still needed. Should I add it to this PR?

Also one of our buildbots is running Ubuntu 14.04, which still needs `static const int NID_X9_62_prime256v1` in nid.py.

src/_cffi_src/openssl/pem.py
mattip: I will update the PR once I get a successful pypy build

reaperhulk:
We're still willing to help with this but since there's been no activity I'll close this for now.

Issue #4279, opened 2018-06-12
Error while setting up developer environment: ImportError: The 'enchant' C library was not found


abhn:
* Versions of Python, ``cryptography``, ``cffi``, ``pip``, and ``setuptools``
  you're using
  - cffi version: 1.11.5
  - pip version: 10.0.1
  - setuptools version: 39.2.0
  - python: 3.6.5
  - cryptography: latest master branch

* How you installed ``cryptography``
  - cloned the repository
  - followed the build instructions listed here: https://cryptography.io/en/latest/installation/
  - followed the developer dependency setup listed here: https://cryptography.io/en/latest/development/getting-started/#development-dependencies

* Clear steps for reproducing your bug
  - on running `pip install --requirement dev-requirements.txt`, I get the error `ImportError: The 'enchant' C library was not found. Please install it via your OS package manager, or use a pre-built binary wheel from PyPI.` for the pyenchant package
  - I can confirm I have enchant installed as can be seen in the screenshots.

Please request any additional information that might be useful in solving this issue. The system is running Arch on Linux 4.16.13

![screenshot from 2018-06-12 20-08-14](https://user-images.githubusercontent.com/6270406/41299046-33adf3e8-6e80-11e8-99ab-a9ac4ac3e14c.png)
![screenshot from 2018-06-12 20-08-45](https://user-images.githubusercontent.com/6270406/41299047-33ef2822-6e80-11e8-8267-75b78678db11.png)

Can anyone guide me to fix this? Thanks in advance.

reaperhulk:
If you have `enchant` installed via your arch package manager (as it looks like you do) then you'll want to debug this with pyenchant.

abhn:
Okay, I'll check with that.

abhn:
For any fellow contributor who finds this issue, I fixed it by cloning the pychant repo (https://github.com/rfk/pyenchant), making this change: https://github.com/rfk/pyenchant/pull/128/commits/da45fcfbdda5a79e6c1cc3b41d8ec9f2fadc297a and then `pip setup.py install` from the directory. 

Pyenchant is unmaintained now, and hence this wasn't fixed. 

We can close this issue. Thanks.

Issue #4968, opened 2019-08-13
Documentation improvement for ECC verify function


boustrophedon:
[In the documentation for EllipticCurvePublicKey.verify](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey.verify) it isn't mentioned anywhere that the signature should be DER encoded. The only time it is mentioned in [the documentation for the ECDSA class](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#cryptography.hazmat.primitives.asymmetric.ec.ECDSA), where it says the output in the example code is DER encoded and can be decoded with the `asymmetric.utils.decode_dss_signature` function. This isn't mentioned in the documentation for [the output of the sign function either](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePrivateKey.sign)

I discovered this by [reading the test code](https://github.com/pyca/cryptography/blob/master/tests/hazmat/primitives/test_ec.py#L497), because I couldn't figure out why my signatures were valid with the Pycryptodome library but not the Cryptography library.

Proposals:
1. Update the documentation in `verify` to say that the signature should be DER encoded, and mention the utils.encode_dss_signature function
2. Update the documentation in `sign` to say that the output is DER encoded
3. Maybe have some more example code somewhere?

If these are okay then I'd be happy to submit a PR with the changes.

reaperhulk:
Yeah we'd be happy to accept a PR improving the docs to explain the signature output encoding (and maybe mentions `decode_dss_signature` a few more times)

boustrophedon:
I'll try to submit it tomorrow.

reaperhulk:
Resolved by #4970. Thanks again for the report.

PR #4970, opened 2019-08-14
Improve documentation for ECDSA sign and verify methods

boustrophedon:
See #4968 - the documentation doesn't mention that signatures produced and consumed are DER-encoded everywhere.

Also, I had a little trouble building the documentation because pyenchant is unmaintained and wasn't able to detect my system's enchant installation - I had to downgrade to enchant1.6. This is semi-documented in this closed issue: #4279


reaperhulk:
Thanks for the doc improvements!

Issue #4845, opened 2019-04-14
Support Ed25519 CSR building


thanatos:
I would like to build a CSR (Certificate Signing Request), and sign it w/ an Ed25519 key. Such keys are newly supported in TLS v1.3. It seems `cryptography` isn't able to support these yet. For example, if I follow the example in the docs on [building a CSR](https://cryptography.io/en/latest/x509/reference/#x-509-csr-certificate-signing-request-builder-object), but use an Ed25519 key instead:

```python
from cryptography import x509  
from cryptography.hazmat.backends import default_backend  
from cryptography.hazmat.primitives import hashes 
from cryptography.hazmat.primitives import serialization  
from cryptography.hazmat.primitives.asymmetric import ed25519  
from cryptography.x509.oid import NameOID  
private_key = ed25519.Ed25519PrivateKey.generate() 
builder = x509.CertificateSigningRequestBuilder()  
builder = builder.subject_name(x509.Name([  
    x509.NameAttribute(NameOID.COMMON_NAME, u'cryptography.io'),  
]))  
builder = builder.add_extension(  
    x509.BasicConstraints(ca=False, path_length=None), critical=True,  
)  
request = builder.sign(  
    private_key, hashes.SHA256(), default_backend()  
)  
isinstance(request, x509.CertificateSigningRequest)                        
```

I get,

```
InternalError                             Traceback (most recent call last)
<ipython-input-19-36fb6c2432d3> in <module>
     14 ) 
     15 request = builder.sign( 
---> 16     private_key, hashes.SHA256(), default_backend()
     17 ) 
     18 isinstance(request, x509.CertificateSigningRequest)

/usr/lib/python3.7/site-packages/cryptography/x509/base.py in sign(self, private_key, algorithm, backend)
    426         if self._subject_name is None:
    427             raise ValueError("A CertificateSigningRequest must have a subject")
--> 428         return backend.create_x509_csr(self, private_key, algorithm)
    429 
    430 

/usr/lib/python3.7/site-packages/cryptography/hazmat/backends/openssl/backend.py in create_x509_csr(self, builder, private_key, algorithm)
    782                 errors[0]._lib_reason_match(
    783                     self._lib.ERR_LIB_RSA,
--> 784                     self._lib.RSA_R_DIGEST_TOO_BIG_FOR_RSA_KEY
    785                 )
    786             )

[...]

PR #4972, opened 2019-08-17
Finish ed25519 and ed448 support in x509 module

markokr:
Currently only ed25519 cert generation works.  This series adds csr and crl support and also ed448 support for crt/csr/crl generation.

Fixes: #4845


markokr:
Pushed rebased v2: added OCSP support, minor fixes.

reaperhulk:
Thanks so much for your hard work here. I have a small cleanup PR I'll be sending but this was excellent. In the future if you wouldn't mind splitting these PRs up it definitely makes reviewing easier (although we know it also makes rebasing harder for the author).

Issue #4981, opened 2019-09-01
X25519 keys endianness incompatible with mbedTLS


Jongy:
I toyed with an X25519 demo of mbedTLS <-> cryptography.

**TL;DR** by reversing the output of `public_bytes()` before passing it to mbedTLS, by reversing the public part mbedTLS gives me before passing it to cryptography and by reversing the shared secret of either side, everything works.
_Why?_

In mbedTLS,  `mbedtls_mpi_write_binary` is used to "export" the bytes of the public part, as well as the final secret itself.
`mbedtls_mpi_write_binary` clearly states `Export X into unsigned binary data, big endian.`

In cryptography, when using `X25519PublicKey.public_bytes()`, it uses the OpenSSL backend (apparently hardcoded, at least as of ca723c42d0e48c4a17aaed2505c2a5be80e1165f) which ends up calling `EVP_PKEY_get1_tls_encodedpoint`.

I dived deeper and although I can't say I found the exact function doing the conversion (it's all rather messed up with lots of function tables), I did find a few functions implementing this conversion (perhaps not for Curve25519 though) and they all used the bignum API, by using `BN_bn2bin` which is apparently big endian as well, as seen [here](https://www.openssl.org/docs/man1.0.2/man3/BN_bn2bin.html).

I also tried with other curves (e.g SECP256R1) which allow accessing the point's X and Y via `public_numbers().X`. Then I used `public_bytes(Encoding.X962, PublicFormat.CompressedPoint)[1:]` and the resulting 32 bytes are indeed big endian: `public_numbers().X == int.from_bytes(public..., byteorder='big')`.

I also found [this](https://forums.mbed.com/t/incompatibility-between-x25519-key-created-with-openssl/3078) forum post in mbedTLS, explaining the same issue.

I think it should be documented in the X25519 docs, but it should be checked deeply to realize what's the root cause.

reaperhulk:
X25519 byte level encoding is defined as little-endian in [RFC 7748](https://www.ietf.org/rfc/rfc7748.txt). This is in contrast with many other things, which are typically encoded in "network byte order" aka big-endian. mbedTLS is wrong here if they want to be able to interop easily. We would, however, be happy to take a PR that explicitly notes that this encoding is little-endian for cases like this.

Jongy:
[...]

PR #4982, opened 2019-09-02
Explicitly state that X25519 uses little endian byteorder (#4981)

Jongy:
#4981

### DIFF
--- a/docs/hazmat/primitives/asymmetric/x25519.rst
+++ b/docs/hazmat/primitives/asymmetric/x25519.rst
@@ -56,6 +56,8 @@ present.
 Key interfaces
 ~~~~~~~~~~~~~~
 
+.. note:: This module uses *little endian* as its byte-level encoding, as per :rfc:`7748`.
+
 .. class:: X25519PrivateKey
 
     .. versionadded:: 2.0
--- a/docs/hazmat/primitives/asymmetric/x448.rst
+++ b/docs/hazmat/primitives/asymmetric/x448.rst
@@ -56,6 +56,8 @@ present.
 Key interfaces
 ~~~~~~~~~~~~~~
 
+.. note:: This module uses *little endian* as its byte-level encoding, as per :rfc:`7748`.
+
 .. class:: X448PrivateKey
 
     .. versionadded:: 2.5


Jongy:
Should I add the RFC for reference here? I thought it was superfluous.

reaperhulk:
Please do -- it helps to have the canonical reference easily at hand. (The test failure is a downstream regression unrelated to you so don't worry about that)

Jongy:
Done

reaperhulk:
Okay, looking at this a bit more I think we should put the notes under the `public_bytes` and `private_bytes` methods so when users look at the serialization options it calls it out. We should also do this for x25519, ed25519, ed448, and x448 so there will be a total of 8 notes saying something like "This method uses little-endian encoding as its byte level encoding as per RFC 7748". There is also a syntax for linking RFCs so you don't have to do a normal link. It looks like: 
```
:rfc:`7748`
```
and renders it as a link with the text "RFC 7748" (so don't write RFC again!).

Jongy:
> Okay, looking at this a bit more I think we should put the notes under the public_bytes and private_bytes methods so when users look at the serialization options it calls it out.

If placed under the method itself, then it should include `from_private_bytes`, `from_public_bytes` and `exchange` as well, and it becomes quite a lot, doesn't it? Too many identical notes.

I thought it's best put under the `Key interfaces` section, first, so readers will hit it first.

>  We should also do this for x25519, ed25519, ed448, and x448

You're probably right, but since I never used any of these, I don't know if it's actually correct. Actually it's not clear from the RFC whether the little-endian applies generally to the curve operations, or to the DHM parameters.

>  There is also a syntax for linking RFCs so you don't have to do a normal link.

Cool, will use it

reaperhulk:
Hmm, @alex what do you think? I'm reluctant to put a note at the top of the module since I think it just becomes noise. Do we need to therefore note that the bytes are little-endian in every method that deserializes or serializes?

alex:
I confess I don't understand the need to document this at all. The module follows everything about the RFC, picking out this one detail just seems odd to me.

Jongy:
> picking out this one detail just seems odd to me

I wouldn't call it "one detail" - after all, the API itself is quite clear, and I assume most people just try meddling with it without too much reading around - but the little-endian is unexpected.

Taking myself as an example user, I know the basics of ECDH (not pretending to know the maths behind it tho) and I took the mbedTLS `ecdh_curve25519` example and tried to modify it to work with cryptography. I didn't really have to read neither the docs nor the RFC, because by simply exploring the objects in IPython it was clear how the exchange should be done.

Now, I'm not saying it's wise to use primitives under `hazmat` without reading the docs at all (Well, it's not :sweat_smile:) but my point is that my very first stop once things came out broken, was the docs, and since the little-endian thing is a really peculiar detail, especially for those working with other cryptographic primitives, I think it should stand out. It would've saved me time, and I'm sure it'd be more helpful than noise for others as well.

And even if I was to read the RFC, the little-endian thing is only mentioned a few times in the whole document. Not so unplausible to miss that, and it's really one of the major things that can go wrong when attempting to interop between different crypto libraries. IMO, for the ease of developers, it should be bold in the API.

markokr:
Perhaps just:

> This curve uses custom byte-level encoding, specified in :rfc:`7748`.

for x25519/x448, and rfc8032 for ed25519/ed448?

The format differences are more that just endianess, also sign bit handling is different from ASN.1.
X448 uses sign bit as part of unsigned value, X25519 ignores it and allows anything there.

When somebody wants to read integer out from x25519/x448 blobs, they must follow all RFC details.  Same with Ed25519/Ed448, which are also different from each other.

Repeat that with secret scalars where reading slightly different in each four variants.

alex:
After discussing with @reaperhulk, we're going to close this -- since every element of this module follows the RFC, calling one specific element of the encoding doesn't make sense.

PR #4984, opened 2019-09-06
fix coverage by adding two artificial DSA public keys

### LINKED ISSUE #4961: Missing coverage
image

https://github.com/pyca/cryptography/blob/c7681e80a68a97ba56453e7fbb960f0e59f4acad/src/cryptography/x509/extensions.py#L56-L66

If this isn't reachable we should convert these into something assert-like.

reaperhulk:
One key removes the optional parameters from the structure to cover a branch conditional, and the other key has its BITSTRING padding value set to a non-zero value.

Fixes #4961


alex:
flake8

PR #4985, opened 2019-09-06
be clear that NoEncryption must be an instance in the exception

### LINKED ISSUE #4983: exception string wrong in Ed448
* Versions of Python, ``cryptography``, ``cffi``, ``pip``, and ``setuptools``
  you're using
cryptography   2.7

* How you installed ``cryptography``

pip install cryptography

* Clear steps for reproducing your bug

```
>>> from cryptography.hazmat.primitives.asymmetric.ed448 import Ed448PrivateKey
>>> private_key = Ed448PrivateKey.generate()
>>> from cryptography.hazmat.primitives.serialization import Encoding
>>> from cryptography.hazmat.primitives.serialization import PrivateFormat
>>> from cryptography.hazmat.primitives.serialization import NoEncryption
>>> private_key.private_bytes(Encoding.Raw, PrivateFormat.Raw, NoEncryption)                             
  File "/Users/jmg/git/medashare/ui/p/lib/python2.7/site-packages/cryptography/hazmat/backends/openssl/ed448.py", line 128, in private_bytes
    "When using Raw both encoding and format must be Raw "
ValueError: When using Raw both encoding and format must be Raw and encryption_algorithm must be NoEncryption
```

The issue is that at https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/backends/openssl/ed448.py#L125 it expects an *instance* of NoEncryption, and not NoEncryption the "constant" as described in the exception text.  Please update to include parenthesis, because as it stands, I called it exactly how the error message said I should call it and it didn't work.

P.S. It'd be nice to have high level API's for asymmetric crypto that isn't X.509.

reaperhulk:
We tell users to use NoEncryption in our [ed/x][25519/448] serialization APIs but don't make it clear that it needs to be an instance of the class and not the class itself. Simple enough to fix.

fixes #4983

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/ed25519.py
+++ b/src/cryptography/hazmat/backends/openssl/ed25519.py
@@ -123,7 +123,7 @@ class _Ed25519PrivateKey(object):
             ):
                 raise ValueError(
                     "When using Raw both encoding and format must be Raw "
-                    "and encryption_algorithm must be NoEncryption"
+                    "and encryption_algorithm must be NoEncryption()"
                 )
 
             return self._raw_private_bytes()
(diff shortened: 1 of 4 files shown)


alex:
@reaperhulk can you rebase this on master for clean CI?

PR #4987, opened 2019-09-07
Simplify implementing sequence methods

alex:


src/cryptography/x509/extensions.py
```diff
@@ -862,11 +844,7 @@ def __init__(self, usages):
 
         self._usages = usages
 
-    def __iter__(self):
-        return iter(self._usages)
-
-    def __len__(self):
-        return len(self._usages)
+    __len__, __iter__, __getitem__ = _make_sequence_methods("_usages")
```
reaperhulk: So we didn't previously implement `__getitem__` on EKU?

alex: Yeah :-/ I decided this was simpler.

PR #5024, opened 2019-10-18
Fix tests in FIPS enforcing mode

tiran:
Based on @reaperhulk original PR https://github.com/pyca/cryptography/pull/4960

``openssl-1.1.1c-2.el8.x86_64`` in FIPS mode requires the following tests to be disabled:

- ARC4, Blowfish, CAST5, Camellia, ChaCha20, IDEA, SEED
- MD5, blake2
- ChaCha20-Poly1305
- RSA key size < 2048 bits or RSA exponent < 65537
- DSA key size < 2048 bits
- finite field DH keysize < 2048 bits
- traditional OpenSSL format for encrypted PEM files
- memleak tests


tiran:
It should be possible to replace the hard-coded disallow lists and minimum key sizes with checks. For example FIPS approved EVP pkey algos have the ``EVP_PKEY_FLAG_FIPS`` flag set.

dsethi:
What is the core thing pending here? We really really need this for the work we are doing (or else would have to try and dig up some other module which is really last resort)

PR #5026, opened 2019-10-18
Correctly document `backend` argument of KBKDFHMAC

jlaine:
The documentation states that `backend` should be a `HashBackend`
instance when in fact it should be a `HMACBackend` instance.

This was discovered while working on https://github.com/python/typeshed/pull/3307


jlaine:
Thanks @alex!

PR #5028, opened 2019-10-18
Drop macOS < 10.12

alex:


alex:
For the historical record: dropping this because dropbox.

Issue #5002, opened 2019-09-15
Support for getting ASN.1 value for parsed extensions


felixfontein:
When parsing a certificate or CSR, it would be nice to be able to get the ASN.1 value for all extensions in a simple way without having to essentially re-create the certificate parsing code (i.e. iterating over the extensions by using `obj._backend` and `obj._x509` and FFI).

This data is available when reading certificates or CSRs, it is currently only stored for `UnrecognizedExtension`s. So storing it, and adding a property (maybe called `asn1_value`), which is either `None` (if ASN.1 value is not available) or the ASN.1 value itself, should be no problem.

I'd like the use this as follows:
 1) For code which dumps all extensions in a uniform way (i.e. as OID, critical flag, ASN.1 encoding tuples);
 2) For code which deals with extensions currently not supported by cryptography, but potentially later.

Let me expand on 2): say I want to add a certain extension using cryptography right now. For that, I can add a `UnrecognizedExtension` with the correct content. But if I want to check whether this extension is already there with the correct content, I have to assume that it will come as a `UnrecognizedExtension`. If cryptography ever supports this extension type directly, my code for detecting whether the correct content is there will stop working.

Would adding this to cryptography be acceptable? (I can create a PR for this if wanted.)

reaperhulk:
I'm interested in how you would implement this. I'm not opposed to this conceptually, but I think our willingness to accept a patch would be predicated on whether we can build a sane API for this.

felixfontein:
I created a WIP for this (#5025). In it, every extension type gets a new property `asn1value`, which can be `None` or a `bytes` string. (I've not added this property to `__eq__` and `__ne__` since that would break a lot of things.)

alex:
@reaperhulk and I discussed this a bit. One of the things that's on our roadmap for 2021 is to move our X.509 extension parsing away from using OpenSSL, and towards our own DER parser (see #5357). We think that's going to have profound implications for the API here, so we're not making any hard decisions until that happens.

PR #5030, opened 2019-10-19
[WIP] Allow to serialize X509 extension's value

felixfontein:
Another WIP for #5002, based on the discussion on IRC.

### DIFF
--- a/docs/hazmat/backends/interfaces.rst
+++ b/docs/hazmat/backends/interfaces.rst
@@ -608,6 +608,14 @@ A specific ``backend`` may provide one or more of these interfaces.
 
         :return bytes: The DER encoded bytes.
 
+    .. method:: serialize_x509_extension(extension)
+
+        .. versionadded:: 2.9
+
+        :param extension: An instance of :class:`~cryptography.x509.Extension`.
+
+        :return bytes: The DER encoded bytes.
+
 .. class:: DHBackend
 
     .. versionadded:: 0.9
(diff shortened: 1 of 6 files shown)


reaperhulk:
Right now when we serialize extensions we pass a set of handlers around and that mechanism is what maps a given extension to its private serialization method in `_encode_asn1`. If we attach a serialization method to the extension then we can have it invoke that method and remove a large pile of this plumbing. So instead of generating extension DER through that class, we could just iterate the list and call serialize on each one.

One complication here is our desire to remove `backend` as an arg. We could get around this by lazy-importing the backend though. So then we'd have classes with implementations that look like this:

```python
class IssuingDistributionPoint(object):
    ...

    def _der_bytes(self):
        from cryptography.hazmat.backends.openssl.encode_asn1 import _encode_issuing_dist_point
        return _encode_issuing_dist_point(backend, self)
```

If we want to go this path I'd say we should have the methods be private at first and expose them publicly after we're happy with how it's working. You could still call the private methods and would be able to follow any changes we might make without any problems :)

felixfontein:
From how I understood it, the `_encode_*` functions in `encode_asn1.py` do not return DER bytes, but (in general) structures which are encoded (indirectly) by OpenSSL's `X509V3_EXT_i2d` to DER. So this method `_der_bytes` would not return DER bytes, and would not really be usable except for passing the result to `X509V3_EXT_i2d` or similar functions.

PR #5032, opened 2019-10-20
Run py3.8 as our py3 in macOS CI

alex:


alex:
Going to be blocked on https://github.com/microsoft/azure-pipelines-image-generation/issues/1317

PR #5038, opened 2019-10-29
Fix #2358: implement streaming API

### LINKED ISSUE #2358: Implement (or document) streaming API
I'm looking at https://cryptography.io/en/latest/fernet/, and I'm wondering how to use this API if I want to encrypt *a lot* of data.

I hope that I can simply call encrypt in a loop:

```
with open('secret.txt', 'rb') as ifh:
    with open('encrypted.txt', 'wb') as ofh:
        f = Fernet(key)
        while True:
            buf = ifh.read(BUFSIZE)
            if not buf:
                break
            ofh.write(f.encrypt(buf)))
```

..but it would be nice to have this offically blessed (and implemented, if it doesn't work like that at the moment).

skvl:
Implement stream version of Fernet:
* The streamed ciphertext have magic _0x8a_.
* Use typing module.
* Tested with python 3.7.

### DIFF
--- a/src/cryptography/fernet.py
+++ b/src/cryptography/fernet.py
@@ -7,8 +7,10 @@ from __future__ import absolute_import, division, print_function
 import base64
 import binascii
 import os
+from pathlib import Path
 import struct
 import time
+from typing import Iterator
 
 import six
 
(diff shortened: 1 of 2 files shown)


alex:
There are two problems here:

a) We do not want to introduce recipes like this which are not supported by a specification which exists beyond this module.

b) The design you have here exposed unauthenticated plaintext on decryption. That is to say, it allows users to process data before the authentication tag is verified, which would be a major cryptographic failure. Any design with this property is unacceptable for a recipe in cryptography.

skvl:
@alex, thank you very much for review. Fill free to close it.

skvl:
@alex, would it be acceptable if:
a) I move new API into separate module?
b) I redesign it to encrypt/decrypt signed chunks of data and each chunk's signature would be checked before returning plaintext?

reaperhulk:
Chunked with a mechanism to prevent reordering would be a minimum, but ultimately you want an interoperable spec that others also choose to implement. I think the best way forward is to convince the Tink team to concretely specify their primitives (https://github.com/google/tink/blob/master/docs/PRIMITIVES.md#streaming-authenticated-encryption-with-associated-data). I have some plans to work on this in 2020, but it's potentially a slow process.

skvl:
> Chunked with a mechanism to prevent reordering would be a minimum
If we use AES in CTR mode then there is no need in such mechanism. Am I right?

PR #5040, opened 2019-10-30
Make more implementation details of bindings obvious private

alex:


src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -135,7 +135,7 @@ def _ensure_ffi_initialized(cls):
                 cls._register_osrandom_engine()
 
     @classmethod
-    def init_static_locks(cls):
+    def _init_static_locks(cls):
```
jlaine: Hm, that's an interesting change. I know of at least one project which will break but I'm happy to remove the call if it makes no sense:

https://github.com/aiortc/aiortc/blob/28cf22c2a82136784cc8ece6ca82143afdba5c86/aiortc/rtcdtlstransport.py#L40

alex: `init_static_locks()` is called at module scope, so there's no way to import `Binding` without it already having been called, so yeah, it doesn't really make sense

alex:
Oh what the hell, we also documented this...

alex:
Ugh, the hell with this, let's just plan to delete it all once we're at 1.1.0 and greater.

PR #5042, opened 2019-10-31
issue-5041: do not add extra flags when compiler or platform does not support them

aixtools:
Assuming that clang also supports these flags.

Please verify correct list of compilers that are known to support these flags.

### DIFF
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -6,7 +6,6 @@ from __future__ import absolute_import, division, print_function
 
 import os
 import sys
-
 from distutils import dist
 from distutils.ccompiler import get_default_compiler
 from distutils.command.config import config
@@ -56,7 +55,8 @@ def _extra_compile_args(platform):
         d = dist.Distribution()
         cmd = config(d)
         cmd._check_compiler()
-        is_gcc = cmd.compiler.compiler[0].startswith("gcc")
+        is_gcc = "gcc" in cmd.compiler.compiler[0] or
+        "clang" in cmd.compiler.compiler[0]
     if is_gcc or not (platform in ["win32", "hp-ux11", "sunos5"] or
                       platform.startswith("aix")):
         return ["-Wconversion", "-Wno-error=sign-conversion"]
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -11,8 +11,8 @@ from _cffi_src.utils import (
     build_ffi_for_binding, compiler_type, extra_link_args
 )
 
-from distutils.ccompiler import get_default_compiler
 from distutils import dist
+from distutils.ccompiler import get_default_compiler
 from distutils.command.config import config
 
 
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -7,13 +7,13 @@ from __future__ import absolute_import, division, print_function
 import os
 import sys
 
-from _cffi_src.utils import (
-    build_ffi_for_binding, compiler_type, extra_link_args
-)
-
 from distutils import dist
 from distutils.ccompiler import get_default_compiler
 from distutils.command.config import config
+
+from _cffi_src.utils import (
+    build_ffi_for_binding, compiler_type, extra_link_args
+)
 
 
 def _get_openssl_libraries(platform):
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -11,6 +11,10 @@ from _cffi_src.utils import (
     build_ffi_for_binding, compiler_type, extra_link_args
 )
 
+from distutils.ccompiler import get_default_compiler
+from distutils import dist
+from distutils.command.config import config
+
 
 def _get_openssl_libraries(platform):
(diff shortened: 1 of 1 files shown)


aixtools:
xlc_r is the definition of CC, but I see from the ci tests that not all distutils even have the variable 'CC', sigh.

This is a hard thing to catch - as not all compilers support these flags. 

In the past I have just editted the file and built it for myself, but I am trying to get more things done as wheels. And ansible has (in)directly a dependency on cryptography.

What do you suggest as the better approach?

alex:
Take a look at `extra_link_args`, that uses the compiler that distutils thinks it's using, I think that'll give you what you want.

aixtools:
The travis CI failed with - cannot connect to py27...

+git clone --depth=1 https://github.com/google/wycheproof /home/travis/wycheproof

Cloning into '/home/travis/wycheproof'...

fatal: unable to access 'https://github.com/google/wycheproof/': Failed to connect to github.com port 443: Connection timed out

The command "./.travis/install.sh" failed and exited with 128 during .

Your build has been stopped.

I'll test with close and reopen.

reaperhulk:
Surely there's a better way to get the compiler? This would ignore `CC` env var right?

aixtools:
I hope this is a "better" way. At least, it is always what I intended, but could not find.

FYI: it is extremely boring that this is broken. At least it seems that `tox -e py37` of ``pip`` depends on cryptography (it seems to want to build cryptography) - and without this, the build always fails.

alex:
The failing flake8 issues look real.

aixtools:
> 
> 
> The failing flake8 issues look real.

Yes - a second one. New to me, the order of the imports. Hope to have a new commit soon.

src/_cffi_src/build_openssl.py
```diff
@@ -46,7 +50,15 @@ def _extra_compile_args(platform):
...
     revisit this.
     """
-    if platform not in ["win32", "hp-ux11", "sunos5"]:
+    # make sure the compiler used supports the flags to be added
+    is_gcc = False
+    if get_default_compiler() == "unix":
+        d = dist.Distribution()
+        cmd = config(d)
+        cmd._check_compiler()
+        is_gcc = cmd.compiler.compiler[0].startswith("gcc")
```
alex:
```suggestion
        is_gcc = "gcc" in cmd.compiler.compiler[0] or "clang" in cmd.compiler.compiler[0]
```

Just to be safe.

aixtools: Made the change, and hope it passes the CI test that keeps failing on pep8. I do test locally, but what I get as a result is not what the CI is reporting - and getting really really tiresome to guess what next.

(tox) devpi@x064:[/data/prj/python/git/cryptography]pycodestyle src/_cffi_src/build_openssl.py

Nothing to report (and I do not like the formatting of the requested change, but tools report what tools report)

(tox) devpi@x064:[/data/prj/python/git/cryptography]flake8 src/_cffi_src/build_openssl.py
ksh: flake8:  not found.

Maybe there is additional stuff from "flake8" processing - but I do not know an easy way to do this locally.

Fingers crossed!

aixtools:
Would appreciate this being merged - even though not being patched does get me some activity on my forums from people requesting assistance :)

src/_cffi_src/build_openssl.py
```diff
@@ -46,7 +49,16 @@ def _extra_compile_args(platform):
...
     revisit this.
     """
-    if platform not in ["win32", "hp-ux11", "sunos5"]:
+    # make sure the compiler used supports the flags to be added
+    is_gcc = False
+    if get_default_compiler() == "unix":
+        d = dist.Distribution()
+        cmd = config(d)
+        cmd._check_compiler()
+        is_gcc = ("gcc" in cmd.compiler.compiler[0] or
```
reaperhulk: This can be written as `is_gcc = cmd.compiler.compiler[0] in ["gcc", "clang"]`.

src/_cffi_src/build_openssl.py
```diff
@@ -46,7 +49,16 @@ def _extra_compile_args(platform):
...
+                  "clang" in cmd.compiler.compiler[0])
+    if is_gcc or not (platform in ["win32", "hp-ux11", "sunos5"] or
+                      platform.startswith("aix")):
```
reaperhulk: Why does this use startswith? Different AIX releases have different suffixes?

src/_cffi_src/build_openssl.py
alex: No it can't.

reaperhulk: As noted in IRC: ugh, stupid string can be things like `clang-3.8` so my proposed switch won't work. C toolchains are such garbage.

aixtools:
Thanks!

src/_cffi_src/build_openssl.py
aixtools: Hi @reaperhulk . Sorry I did not reply much earlier. Answer is yes, until recently the value included the version (e..g., 5. 6 or 7) of the AIX system it was built on (not the version it is running on). iirc, starting with Python 3.8 it only returns "aix". So, for now, still have to look at the "startswith" part.

PR #5046, opened 2019-11-01
Drop a link because it's breaking linkcheck

alex:


alex:
Whoops, found a citation we can replace this with.

PR #5048, opened 2019-11-03
Deal with the 2.5 deprecations

alex:


tests/hazmat/primitives/test_x25519.py
```diff
@@ -95,16 +94,11 @@ def test_null_shared_key_raises_error(self, backend):
...
-        key = X25519PrivateKey.generate().public_key()
-        with pytest.warns(utils.DeprecatedIn25):
-            key.public_bytes()
-
     def test_public_bytes_bad_args(self, backend):
         key = X25519PrivateKey.generate().public_key()
         with pytest.raises(ValueError):
             key.public_bytes(None, serialization.PublicFormat.Raw)
-        with pytest.raises(ValueError):
+        with pytest.raises(TypeError):
```
reaperhulk: Is this test useful now?

tests/hazmat/primitives/test_ec.py
```diff
@@ -181,7 +181,7 @@ def test_encode_point():
         16
     )
     pn = ec.EllipticCurvePublicNumbers(x, y, ec.SECP256R1())
-    with pytest.warns(utils.DeprecatedIn25):
+    with pytest.warns(utils.CryptographyDeprecationWarning):
```
reaperhulk: Why not use PersistentlyDeprecatedIn2019 here?

tests/hazmat/primitives/test_x25519.py
alex: 🤷‍♂ it doesn't hurt.

tests/hazmat/primitives/test_ec.py
alex: Uhhh, wtf, I have no idea.

PR #5051, opened 2019-11-03
Run tests on centos8

alex:


alex:
Ahhh, https://github.com/openssl/openssl/issues/8450 is almost certainly the issue. So these tests need to be skipped. Great.

alex:
Err, maybe not, we don't enable FIPS mode yet...

@tiran can you think of any modifications that CentOS 8 does to OpenSSL that might cause it to have custom allocators not work?

reaperhulk:
This is because our centos8 container contains `/etc/system-fips` so we're getting some FIPS-y nonsense. We could remove this file and it would work, although I'm curious whether @tiran or another RH dev can explain why this is (at least partially) enabling FIPS without showing FIPS_mode of 1 in OpenSSL itself.

alex:
@reaperhulk should we kill the `/etc/fips` thing?

reaperhulk:
Probably the right way forward for now. FIPS is a separate PR with several other issues to solve before we could re-add the file to this container.

PR #5055, opened 2019-11-10
Fixed #5050 -- dropped support for two releases of LibreSSL

### LINKED ISSUE #5050: Drop support for some versions of LibreSSL
Currently our matrix of LibreSSL version support grows unbounded. We should drop support for one or two of them.

alex:


alex:
Alpine is still on 2.7... fml.

PR #5057, opened 2019-11-11
Fixes #4699 -- document that UnrecognizedExtension can be used in gen…

### LINKED ISSUE #4699: Document that UnrecognizedExtension can be used for custom extensions
We allow arbitrary data to be encoded into an extension via the cert builder and UnrecognizedExtension, but we don't document that fact. We should!

alex:
…eration

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -2460,7 +2460,9 @@ X.509 Extensions
     .. versionadded:: 1.2
 
     A generic extension class used to hold the raw value of extensions that
-    ``cryptography`` does not know how to parse.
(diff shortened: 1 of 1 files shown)


docs/x509/reference.rst
```diff
@@ -2460,7 +2460,9 @@ X.509 Extensions
     .. versionadded:: 1.2
 
     A generic extension class used to hold the raw value of extensions that
-    ``cryptography`` does not know how to parse.
+    ``cryptography`` does not know how to parse. This can also be used when
+    creating new certificates, CRLs, or OCSP requests and response to encode
```
reaperhulk: Oops, after approving I saw this should probably say "responses"

PR #5059, opened 2019-11-11
Parse single_extensions in OCSP responses

reaperhulk:
replaces #4753


src/cryptography/hazmat/backends/openssl/decode_asn1.py
```diff
@@ -857,6 +857,12 @@ def _decode_nonce(backend, nonce):
     OCSPExtensionOID.NONCE: _decode_nonce,
 }
 
+_OCSP_SINGLERESP_EXTENSION_HANDLERS_NO_SCT = _REVOKED_EXTENSION_HANDLERS.copy()
```
alex: Does the RFC say that anything valid for a revoked extension can be a single response extenion, or does it _happen_ to be that case?

src/cryptography/hazmat/backends/openssl/decode_asn1.py
```diff
@@ -857,6 +857,12 @@ def _decode_nonce(backend, nonce):
...
+_OCSP_SINGLERESP_EXTENSION_HANDLERS = _REVOKED_EXTENSION_HANDLERS.copy()
+_OCSP_SINGLERESP_EXTENSION_HANDLERS[
+    ExtensionOID.PRECERT_SIGNED_CERTIFICATE_TIMESTAMPS
+] = _decode_precert_signed_certificate_timestamps
```
alex: I don't think this is correct -- this shouldn't be a PrecertedSignedCerificateTimestamps, that's for placing in an X.509 certificate. The OCSP extension is different.

src/cryptography/hazmat/backends/openssl/decode_asn1.py
reaperhulk: It explicitly says it: https://tools.ietf.org/html/rfc6960#section-4.4.5

alex: Please add a comment

PR #5061, opened 2019-11-12
Windows Installation: fix link to binaries

mhils:
While https://github.com/pyca/infra/tree/master/windows/openssl points a trusted source (which would be preferable), pyca/infra currently does not provide any Windows binaries for download. Linking to the official OpenSSL wiki seems to be a reasonable compromise.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -47,12 +47,11 @@ just run
     $ pip install cryptography
 
 If you prefer to compile it yourself you'll need to have OpenSSL installed.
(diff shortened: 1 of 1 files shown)


alex:
Oops, looks like you've got some trailing whitespace. Please fix that up.

reaperhulk:
We should probably look into whether azure allows us to publish a public artifact now. We build a dumb container with it right now because public artifact feeds were nonsense early this year. Maybe they’re easier now?

PR #5062, opened 2019-11-12
OpenSSL: Expose SSL_CTX_clear_mode/SSL_clear_mode

mhils:
This PR adds `SSL_CTX_clear_mode` and `SSL_clear_mode` bindings, which were introduced in OpenSSL 1.1.1: https://www.openssl.org/docs/man1.1.1/man3/SSL_clear_mode.html

We have an ugly combination of `select()`, blocking I/O and `SSL_read()` in our old TCP proxying code for @mitmproxy, so we need a way to turn off  `SSL_MODE_AUTO_RETRY` and hide our atrocities until our sans-io proxying is ready.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -338,6 +338,7 @@ const COMP_METHOD *SSL_get_current_expansion(SSL *);
 const char *SSL_COMP_get_name(const COMP_METHOD *);
 
 unsigned long SSL_set_mode(SSL *, unsigned long);
+unsigned long SSL_clear_mode(SSL *, unsigned long);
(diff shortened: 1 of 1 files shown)


alex:
As far as I can tell these macros have existed since OpenSSL 0.9.8 -- is making them conditional really required?

mhils:
Ah, it looks like they were just not documented before 1.1.1. 
They are present in 1.0.2 [1], I will update the PR!

[1] https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/ssl/ssl.h

mhils:
Awesome. Thanks for the helpful pointer and the incredibly fast merge, @alex! 😃 🍰

PR #5073, opened 2019-11-24
Run codecov in required mode so it exits with non-zero on failure

alex:


alex:
Naturally on a PR to make codecov more correct, it's lost its mind and is not merging things correctly.

reaperhulk:
Merging despite non-100% coverage because while this touches the codecov upload `--required` isn't likely to be the reason. codecov has trained me to disbelieve it 😁

PR #5085, opened 2019-12-05
Make it easy to have custom extension handlers

uruz:
I want to have custom extension handler to read and write values for custom OID.
With the current implementation it's a bit hard to do: extension handlers are registered in the `encode_asn1.py` in a constant. So I thought it would be great to have the possibility to subclass a `Backend` class and override handler in the subclass. 
I am not very familiar with the source code of `cryptography` so the solution proposed maybe not the best one. It also introduces two more hash lookups. 
Please tell if the solution does not fit somehow.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -119,6 +119,11 @@ class Backend(object):
     OpenSSL API binding interfaces.
     """
     name = "openssl"
+    crl_entry_extension_encode_handlers = _CRL_ENTRY_EXTENSION_ENCODE_HANDLERS
+    crl_extension_encode_handlers = _CRL_EXTENSION_ENCODE_HANDLERS
+    extension_encode_handlers = _EXTENSION_ENCODE_HANDLERS
+    ocsp_basicresp_extension_encode_handlers = _OCSP_BASICRESP_EXTENSION_ENCODE_HANDLERS
+    ocsp_request_extension_encode_handlers = _OCSP_REQUEST_EXTENSION_ENCODE_HANDLERS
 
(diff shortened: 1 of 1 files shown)


reaperhulk:
I think registering custom OID handlers is an interesting idea, but I'm opposed to doing it via subclassing (cryptography has zero subclassing because we strongly prefer composition over inheritance). We'd be happy to talk about alternate proposals, although given that we're still planning to deprecate backends avoiding mutating global state may be very difficult...

uruz:
Alternative proposal would be something like registry above the handlers dictionary. Would that be more appropriate?

reaperhulk:
We're not going to make this public API for now, but #5405 and #5406 make this easier to do yourself now (just mutate the dicts, which are constructed per-backend).

dalai4git:
I tried to work with this, but then the call to `X509V3_EXT_d2i` fails (as expected?) in https://github.com/pyca/cryptography/blob/3.1/src/cryptography/hazmat/backends/openssl/decode_asn1.py#L248. 

My workaround was to patch `_X509ExtensionParser.parse()` and instead of raising a ValueError pass the DER data to the handler (similarly to what is done for an unrecognized extension). Obviously I would rather avoid patching around in library code, so I wanted to ask if there is some other way to do it.

Perhaps for custom extension handlers it would be possible to register them in an extra dict and then pass the DER data to the handler or, if no custom handler for that OID is present, create an `UnrecognizedExtension` by changing the code in https://github.com/pyca/cryptography/blob/3.1/src/cryptography/hazmat/backends/openssl/decode_asn1.py#L242-L246. Then reading is done and writing is already supported if one doesn't mind subclassing from `UnrecognizedExtension`.

PR #5096, opened 2020-01-07
[RFC] Can detect openssl libraries to use on Windows

AWhetter:
This is a proof of concept that makes the setup.py detect the names of the OpenSSL libraries to use, rather than relying on `CRYPTOGRAPHY_WINDOWS_LINK_LEGACY_OPENSSL` being set.

I wanted to get an idea of whether this is something that might be worth finishing off considering that 1.1.x support is going away in 2.9. It might be useful for backporting to 2.8 if that happens?

I don't know whether we need to worry about the fact that this could change behaviour in environments where both the new and legacy libraries exist, but the newer libraries are broken in some way. In this situation, cryptography will either build successfully but now against the legacy libraries, or the build will fail and with errors about the legacy libraries rather than the new libraries.
If this is a concern, then I think an environment variable would need to be set to allow detection of the library to use, in which case this change might not be worthwhile.

### DIFF
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -11,8 +11,33 @@ from distutils.ccompiler import get_default_compiler
 from distutils.command.config import config
 
 from _cffi_src.utils import (
-    build_ffi_for_binding, compiler_type, extra_link_args
+    build_ffi, build_ffi_for_binding, compiler_type, extra_link_args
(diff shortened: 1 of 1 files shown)


alex:
Hey, sorry we didn't get to this for a few months. Since upstream no longer supports 1.0.2, we don't really want to add extra code for supporting it.

PR #5104, opened 2020-01-16
Create simplified, default encryption classes category with RSA class

deadsmond:
Proposal of new category of objects in cryptography package - Simplified. 
The goal of cryptography.simplified is to provide ready-to-run encryption with default, community judged, settings, which would be used by people (like me), who do not need to deeply understand cryptography nuances. We would use methods provided in various examples from the Internet anyway, so why not create simplified objects that contain them, are verified and are ready-to-go?
Leave control over specific parameters to specialists and allow mortals to create encryption key, encrypt and decrypt message in maximum 5 lines of code.

### DIFF
--- a/src/cryptography/simplified/SimplifiedRSA.py
+++ b/src/cryptography/simplified/SimplifiedRSA.py
@@ -1,0 +1,161 @@
+import base64
+
+from cryptography.exceptions import InvalidSignature
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import hashes, serialization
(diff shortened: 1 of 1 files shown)


reaperhulk:
We do want higher level constructs (what we call recipes) in this library, but RSA encryption and signing are not good candidates for this for a variety of reasons. We also want those recipes to have concretely defined security properties (e.g. `IND-CCA2` under the following assumptions) and specifications that allow interoperable implementations. `tink` is planning to publish their specs soon so we may see a significant expansion of recipes soon, but classes like this (which likely solve a perfect use case for you!) aren't good candidates for merging into the main repository.

PR #5109, opened 2020-01-26
Allow NameAttribute.value to be an empty string

### LINKED ISSUE #5106: ValueError in .rfc4514_string()
Using `cryptography==2.8` and calling `.rfc4514_string()` on a subject/issuer containing empty strings raises a `ValueError` exception.

https://github.com/pyca/cryptography/blob/2.8/src/cryptography/x509/name.py#L96-L97

Is there any reason for doing so? I know this might not be good practice, but it prevents a quick and easy way to display some certificates. For example, this one from `code.jquery.com`:

https://censys.io/certificates/23723c81a723301885ba8825da77c145f84c1e181cc9ec94899b0f3dc33bbf3e

```
C=US, ST=, L=, O=jquery, OU=Certificate Authority, CN=code.jquery.com

adepasquale:
[RFC 4514](https://tools.ietf.org/html/rfc4514) does not explicitly mention that `AttributeValue` can not be an empty (zero-length) string.

This pull request fixes #5106 by allowing `NameAttribute.value` to be an empty string, but not `None`.

### DIFF
--- a/src/cryptography/x509/name.py
+++ b/src/cryptography/x509/name.py
@@ -54,6 +54,9 @@ _NAMEOID_TO_NAME = {
 def _escape_dn_value(val):
     """Escape special characters in RFC4514 Distinguished Name value."""
 
+    if not val:
+        return ''
+
     # See https://tools.ietf.org/html/rfc4514#section-2.4
     val = val.replace('\\', '\\\\')
     val = val.replace('"', '\\"')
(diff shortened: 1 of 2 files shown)


adepasquale:
Hello, can this be merged? Tests should be passing

alex:
@adepasquale Apologies we haven't gotten around to reviewing this yet. Can you please rebase/merge master into this PR? Our CI had some issues which we think are fixed now.

PR #5120, opened 2020-02-14
Reversed the order of RDNs in x509.Name.rfc4514_string()

### LINKED ISSUE #5119: x509.Name.rfc4514_string() returns RDNs in wrong order
Hi everyone,

I think the `rfc4514_string()` function of `x509.Name` should return the RDNs in reversed order compared to their 'real' order in i.e. a certificate, because RFC4514 states in section 2.1:

> Otherwise, the output consists of the string encodings of each
> RelativeDistinguishedName in the RDNSequence (according to Section
> 2.2), starting with the last element of the sequence and moving
> backwards toward the first.

I.e. I would expect that a x509.CertificateSigningRequest's subject translates into a string representation (according to RFC4514) like
`"CN=elephant,O=serengeti,C=TZ"`.

The CSR can i.e. be created using OpenSSL:
`openssl req -new -newkey rsa:1024 -keyout test.key -nodes -out test.csr -subj "/C=TZ/O=serengeti/CN=elephant"`.

However doing
```python
from cryptography import x509

with open('test.csr') as f:
    csr = f.read().encode('ascii')

csr = x509.load_pem_x509_csr(csr, default_backend())
print(csr.subject.rfc4514_string())
```
gives `C=TZ,O=serengeti,CN=elephant`.

The issue - if I understand the RFC correctly and it even is one - is probably located here: https://github.com/pyca/cryptography/blob/master/src/cryptography/x509/name.py#L221 . Inserting a `reversed(...)` should fix it.
Would be great if someone could share her/his opinion on this, and if we all agree I could create a PR. However there is not much point in creating a PR to add a single word, I guess ...

erbth:
When converting a RDNSequence to a string representation according to RFC 4514, one shall reverse the order of the RDNs. The RFC states this in section 2.1:

> Otherwise, the output consists of the string encodings of each
> RelativeDistinguishedName in the RDNSequence (according to Section
> 2.2), starting with the last element of the sequence and moving
> backwards toward the first.

I.e. if one creates a CSR using OpenSSL like `openssl req -new -newkey rsa:1024 -keyout test.key -nodes -out test.csr -subj "/C=TZ/O=serengeti/CN=elephant"` and reads this CSR using cryptography, `csr.subject.rfc4514_string()` should return `C=TZ,O=serengeti,CN=elephant`. This makes it consistent with, for example, `openssl req -noout -text -in test.csr -nameopt rfc2253` or what Apache2's mod_ssl returnes in client authentication meta-data. Note that RFC 2253 was the predecessor of RFC 4514, which did not change the conversion order for string representations of RDNSequences.

I've also updated the test for `x509.Name.rfc4514_string`. To not have to change more tests, and for backwards compatibility, I've kept the `x509.Name.__repr__` method's output the same though this required changing them (they used `rfc4514_string`). However I also think it is intuitive for `__repr__` to return something more close to the RDNSequence, since that is the way the name is stored in a binary representation. But I'm happy to update the PR if you think keeping `rfc4514_string` in `__repr__` would be better.

This PR should fix #5119.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,9 @@ Changelog
 * Added support for parsing
   :attr:`~cryptography.x509.ocsp.OCSPResponse.single_extensions` in an OCSP
   response.
+* **BACKWARDS INCOMPATIBLE:** Reversed the order in which
+  :meth:`~cryptography.x509.Name.rfc4514_string` returns the RDNs as required by
+  RFC4514.
 
 .. _v2-8:
 
--- a/src/cryptography/x509/name.py
+++ b/src/cryptography/x509/name.py
@@ -216,9 +216,11 @@ class Name(object):
         An X.509 name is a two-level structure: a list of sets of attributes.
         Each list element is separated by ',' and within each list element, set
         elements are separated by '+'. The latter is almost never used in
-        real world certificates.
+        real world certificates. According to RFC4514 section 2.1 the
+        RDNSequence must be reversed when converting to string representation.
         """
-        return ','.join(attr.rfc4514_string() for attr in self._attributes)
+        return ','.join(
+            attr.rfc4514_string() for attr in reversed(self._attributes))
 
     def get_attributes_for_oid(self, oid):
         return [i for i in self if i.oid == oid]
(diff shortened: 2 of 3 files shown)


erbth:
Sorry for the force push, had a type in the commit message that I'd missed before.

erbth:
Sorry for the delay, other things meant to be `more important' ...
Rebased to current master so that the travis check works.

erbth:
Does someone have an idea what could be wrong with the Travis build? It worked once (with the force push 2 days ago) and then never again, even with the same previously working changes in a different commit (to trigger a new build). If not I could debug this in the next few days ...

alex:
Ugh, that's an impressively useless error message.

erbth:
Thank You very much for the merge.

PR #5125, opened 2020-02-22
Hotfix: use enum34 != 1.1.8

gpotter2:
Cryptography is currently broken on Python 2 because of enum34 1.1.8. See
https://bitbucket.org/stoneleaf/enum34/issues/27/enum34-118-broken

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -236,7 +236,7 @@ setup(
         "six >= 1.4.1",
     ] + setup_requirements,
     extras_require={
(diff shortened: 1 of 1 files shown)


alex:
I don't think this will help you out very much -- we're not going to issue a release for this issue. You should be able to accomplish this by putting the bound in your own requirements.txt or similar.

gpotter2:
Makes sense, thanks for the note.

PR #5126, opened 2020-02-22
Added import for pkcs12

oliver-hasse:
Since v2.5 it's possible to load a pkcs12 file. The necessary import was not set within the __init__-file.


alex:
This import is not necessary. It's intentional that you access this with `from cryptography.hazmat.primitives.serialization import pkcs12`

PR #5136, opened 2020-03-12
Add MD4 hash algorithm to hashes

schwabe:
While nobody should be using this algorithm, there are some use cases
in which it is needed, like implementing CHAP2.

### DIFF
--- a/docs/hazmat/primitives/cryptographic-hashes.rst
+++ b/docs/hazmat/primitives/cryptographic-hashes.rst
@@ -231,6 +231,20 @@ SHA-1
     vulnerable to practical collision attacks, and collisions have been
     demonstrated.
 
+MD4
+~~~
+
+.. warning::
+
+    MD4 is a deprecated hash algorithm that has practical known collision
+    attacks. You are strongly discouraged from using it. Existing applications
+    should strongly consider moving away.
+
+.. class:: MD4()
+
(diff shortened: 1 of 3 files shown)


reaperhulk:
So you really need to implement CHAP2? That's such an old and terrible protocol 😢 

Is this blocking a move from another lib?

schwabe:
@reaperhulk Yeah, I really need it. It is the one you try to get rid of but Enterprise still use it (with unecrypted radius no less in this case).

But I later figured out that putting 

```
@utils.register_interface(HashAlgorithm)
class MD4(object):
    name = "md4"
    digest_size = 16
    block_size = 64
```

in my file that uses it instead of having it in cryptography, also works.

I am getting rid of [PyCrypto](https://www.pycrypto.org), so I needed a different way to get MD4

reaperhulk:
Yeah we’d definitely prefer not to expose this so if you have a good workaround that’s our preferred path. hashlib still has md4 as well 😄 Given the workaround you have I’ll go ahead and close this for now.

schwabe:
my hashlib doesn't:
```
>>> hashlib.md5
<built-in function openssl_md5>
>>> hashlib.md4
AttributeError: 'module' object has no attribute 'md4'
>>> 
```

But yes. I have a workaround.

alex:
`hashlib.new("md4")`

PR #5144, opened 2020-03-20
linkcheck: remove domain that fixed it's tls from ignore

alex:


reaperhulk:
See you in 90 days when this happens again

PR #5149, opened 2020-03-21
Cleanup serialize

markokr:
Changes to public_bytes() and private_bytes() are in separate commits, to better see external changes.

Diff for backend.py is not readable though, please just look at final code.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -1748,26 +1748,14 @@ class Backend(object):
         return ctx
 
     def _private_key_bytes(self, encoding, format, encryption_algorithm,
-                           evp_pkey, cdata):
+                           key, evp_pkey, cdata):
+        # validate argument types
+        if not isinstance(encoding, serialization.Encoding):
+            raise TypeError("encoding must be an item from the Encoding enum")
         if not isinstance(format, serialization.PrivateFormat):
             raise TypeError(
                 "format must be an item from the PrivateFormat enum"
             )
-
-        # X9.62 encoding is only valid for EC public keys
-        if encoding is serialization.Encoding.X962:
-            raise ValueError("X9.62 format is only valid for EC public keys")
-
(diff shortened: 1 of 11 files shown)


reaperhulk:
It looks like there may be an uncovered line here. It is line 1969 in backend.py according to codecov, but it's a bit hard to tell because codecov is broken garbage and only shows that in the diff and says it can't find the base commit otherwise. If you run coverage locally do you cover that line?

alex:
I think it's real -- that line should be hit by _every_ test run, so it's inconceivable it'd randomly be left out. Therefore, I think it really is uncovered.

markokr:
Fixed.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1748,174 +1748,164 @@ def _ec_key_set_public_key_affine_coordinates(self, ctx, x, y):
         return ctx
 
     def _private_key_bytes(self, encoding, format, encryption_algorithm,
-                           evp_pkey, cdata):
+                           key, evp_pkey, cdata):
```
reaperhulk, line 1751: You added key to args here but I can't find where it's used. Am I missing it?

markokr: It's used by OpenSSH PR...

markokr: Generally speaking, any serializer that does not use OpenSSL directly probably wants regular instances.

markokr:
Any trouble with this?  Do you want me to clarify or rewrite anything?

Main idea was to remove code and responsibility from algo-specific files to protect backend.py from unknown/new enum values.

Second idea was to clarify code in backend.py.  Old code was confusing and it was possible to trigger asserts which means there could be more interesting behaviour when running in Python with asserts disabled (-O).

reaperhulk:
@markokr Apologies for the slow reviews here, but thank you for your contribution!

markokr:
No problem, thanks for checking it out!

PR #5157, opened 2020-03-28
AES-CCM data length validation error message fix.

mjurczak:
Fixed error message in AES-CCM data length validation to reflect the true reason of error more accurately.

AES-CCM encrypted input data is limited by the length of provided nonce. Error message should indicate that the provided data input is too long for the nonce of given length.


reaperhulk:
Yeah unfortunately since it changes the travis config we can't just restart the job. If you could rebase or merge in master we'll be able to merge this fix. Thanks for the contribution!

alex:
Trying a magic thing.

PR #5174, opened 2020-04-02
nothing can ever work

reaperhulk:


reaperhulk:
https://dev.azure.com/pyca/cryptography/_build/results?buildId=1687&view=results demonstrates that this will work for releasing the 2.9 wheel for macos

PR #5178, opened 2020-04-02
Drop support for OpenSSL 1.0.1

alex:


src/_cffi_src/openssl/ssl.py
```diff
@@ -733,22 +680,6 @@
 static const long TLS_ST_OK = 0;
 #endif
 
-/* LibreSSL 2.9.1 added only the DTLS_*_method functions */
-#if CRYPTOGRAPHY_OPENSSL_LESS_THAN_102 && !CRYPTOGRAPHY_LIBRESSL_291_OR_GREATER
```
reaperhulk, line 737: Since we still support LibreSSL < 2.9.1 we need some logic here still right?

alex: Uhh, probably. I'm going to let the CI fail first though :D

tests/hazmat/backends/test_openssl.py
```diff
@@ -555,18 +555,9 @@ def test_numeric_string_x509_name_entry(self):
             x509.load_der_x509_certificate,
             backend
         )
-        if backend._lib.CRYPTOGRAPHY_OPENSSL_LESS_THAN_102I:
```
reaperhulk, line 558: Don't we still support < 1.0.2i? Xenial ships with 1.0.2g so we should still need this branch

src/cryptography/hazmat/primitives/constant_time.py
```diff
@@ -5,31 +5,10 @@
...
-    )
-
-    def bytes_eq(a, b):
-        if not isinstance(a, bytes) or not isinstance(b, bytes):
-            raise TypeError("a and b must be bytes.")
-
-        return lib.Cryptography_constant_time_bytes_eq(
-            a, len(a), b, len(b)
-        ) == 1
+    return hmac.compare_digest(a, b)
```
reaperhulk, line 14: Do we need to check in with certbot on this?

tests/wycheproof/test_ecdsa.py
```diff
@@ -62,9 +62,9 @@ def test_ecdsa_signature(backend, wycheproof):
             binascii.unhexlify(wycheproof.testgroup["keyDer"]), backend
         )
     except (UnsupportedAlgorithm, ValueError):
-        # In OpenSSL 1.0.1, some keys fail to load with ValueError, instead of
-        # Unsupported Algorithm. We can remove handling for that exception
-        # when we drop support.
+        # In some OpenSSL 1.0.2s, some keys fail to load with ValueError,
```
reaperhulk, line 65: God damn it.

src/cryptography/hazmat/primitives/constant_time.py
alex: Yeah. Probably: @bmw + @ohemorange -- Is dropping Pythons that don't have `hmac.compare_digest` (2.7.x where x < mumble mumble -- Debian Jessie / Ubuntu Trusty vintage) going to cause you trouble?

We're dropping OpenSSL 1.0.1, which I think we already confirmed with you, but without the distros that use OpenSSL 1.0.1, we don't have any Pythons old enough in our CI :-)

tests/wycheproof/test_ecdsa.py
alex: 💻

src/cryptography/hazmat/primitives/constant_time.py
bmw: I don't think this will cause us any trouble.

The only distro we still support that has older than Python 2.7.7 is CentOS/RHEL 7 where it appears RHEL backported `hmac.compare_digest`.

Thanks for checking!

alex: Outstanding, thanks!

PR #5179, opened 2020-04-03
Expose SSL_CTX_(get|set)_keylog_callback

mhils:
Thanks for the new cryptography release! 😃🍰 

We're currently upgrading @mitmproxy and it looks like our SSLKEYLOGFILE implementation breaks somewhere between OpenSSL 1.1.1d and 1.1.1f. This is something we can implement ugly workarounds for at the moment, but the long-term fix is to just use SSL_CTX_set_keylog_callback (thanks for building that, @Lukasa!). There will be a [follow-up PR](https://github.com/mhils/pyopenssl/tree/keylog) for pyOpenSSL once this one has landed.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -32,6 +32,7 @@ static const long Cryptography_HAS_SIGALGS;
 static const long Cryptography_HAS_PSK;
 static const long Cryptography_HAS_CIPHER_DETAILS;
 static const long Cryptography_HAS_VERIFIED_CHAIN;
+static const long Cryptography_HAS_KEYLOG;
 
 /* Internally invented symbol to tell us if SNI is supported */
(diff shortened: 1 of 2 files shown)


alex:
This appears to have broken tests on versions < 1.1.1

mhils:
I fixed the 1.1.1 builds and broke Travis in return. All jobs pass now, it just doesn't report back properly: https://travis-ci.org/github/pyca/cryptography/builds/670775392. Re-triggering doesn't help. Let me know if you need a new PR because of that. :)

alex:
Ugh, travis isn't reporting results. Doing the weird hack.

alex:
@mhils can you merge/rebase master or something here? Travis is rudely refusing to submit status.

mhils:
Already tried that. New PR it is. :-)

PR #5192, opened 2020-04-12
Run tests on aarch64, ppc, and s390x

alex:


alex:
Failures seem to be caused by caching here: https://travis-ci.community/t/permission-issue-while-building-wheels-for-various-python-packages/7822

alex:
This uses distro OpenSSL (1.0.2). Do we want to build our own here? In addition to or instead of the distro?

reaperhulk:
We should probably do both at least on aarch64 since we're likely to ship a wheel there. I don't much care about ppc64le or s390x but I guess they can stay?

alex:
Ok, added a 1.1.1f build for aarch64, left ppc and s390x as is.

PR #5197, opened 2020-04-17
Document that we test on bullseye

alex:


alex:
@mattsb42-aws it looks like something must have changed in the encryption SDK, because our tests are failing now :-) Do you have a suggestion for a fix? Should there be a fix on your end?

mattsb42-aws:
doh...I thought we had a CI check that should have caught this...let me check and see how involved a fix will be

alex:
thanks for the quick turnaround!

mattsb42-aws:
a patch-workaround would be to set `AWS_ENCRYPTION_SDK_PYTHON_INTEGRATION_TEST_AWS_KMS_KEY_ID` and `AWS_ENCRYPTION_SDK_PYTHON_INTEGRATION_TEST_AWS_KMS_KEY_ID_2` environment variables to something that starts with `arn:`, because this is just triggering on pytest test generation and the relevant tests aren't actually running...but we should be handling this on our end.

alex:
Ok, PR for that up here: https://github.com/pyca/cryptography/pull/5198 we can continue the discussion there

mattsb42-aws:
fix: https://github.com/aws/aws-encryption-sdk-python/pull/256

mattsb42-aws:
fix is merged; you should be gtg, but I'll stick around and keep an eye on that downstream check just in case

alex:
thanks, the travis queue is a bit backed up, but it should start going in a couple of minutes

alex:
Looks like we're all good. Thanks again for jumping on this.

reaperhulk:
A+ service would do downstream testing against AWS SDKs again 😁

PR #5198, opened 2020-04-17
Try to fix the aws-encryption-sdk tests

alex:


mattsb42-aws:
A proper fix is in process here: https://github.com/aws/aws-encryption-sdk-python/pull/256

alex:
You think that's likely to land this evening? I was about to ask if the office was already cleared out... but I guess that question isn't very relevant these days.

mattsb42-aws:
Yeah, I'll make sure it lands this evening; looks like I missed something, but I'll get that sorted asap.

alex:
Thanks. I'll go ahead and close this then.

PR #5202, opened 2020-04-19
Migrate the manylinux wheel builder to GHA

alex:


reaperhulk:
Please add a `name:` key, otherwise LGTM.

PR #5208, opened 2020-04-20
Add support for more AES key sizes

abarto:
Add support for all the supported AES key sizes under the current Fernet setup (128, 192 and 256 bits).


alex:
The fernet spec does not define support for any other AES key sizes, so no we can't do this.

abarto:
@alex No problem! Thanks for checking the PR.

Issue #5075, opened 2019-11-26
Add tests for new wycheproof vectors since c313761979d74b0417230eddd0f87d0cfab2b46b


alex:
Wycheproof just landed a bunch of new vectors! We need to add tests for them.

- [ ] ~~`testvectors/aead_aes_siv_cmac_test.json`~~
- [ ] ~~`testvectors/aegis128L_test.json`~~
- [ ] ~~`testvectors/aegis128_test.json`~~
- [ ] ~~`testvectors/aegis256_test.json`~~
- [ ] ~~`testvectors/dsa_2048_224_sha224_p1363_test.json`~~
- [x] `testvectors/dsa_2048_224_sha224_test.json` (#5098)
- [ ] ~~`testvectors/dsa_2048_224_sha256_p1363_test.json`~~
- [x] `testvectors/dsa_2048_224_sha256_test.json` (#5098)
- [ ] ~~`testvectors/dsa_2048_256_sha256_p1363_test.json`~~
- [x] `testvectors/dsa_2048_256_sha256_test.json` (#5098)
- [ ] ~~`testvectors/dsa_3072_256_sha256_p1363_test.json`~~
- [x] `testvectors/dsa_3072_256_sha256_test.json` (#5098)
- [ ] ~~`testvectors/ecdsa_brainpoolP224r1_sha224_p1363_test.json`~~
- [ ] ~~`testvectors/ecdsa_brainpoolP256r1_sha256_p1363_test.json`~~
- [ ] ~~`testvectors/ecdsa_brainpoolP320r1_sha384_p1363_test.json`~~
- [ ] ~~`testvectors/ecdsa_brainpoolP384r1_sha384_p1363_test.json`~~
- [ ] ~~`testvectors/ecdsa_brainpoolP512r1_sha512_p1363_test.json`~~
- [ ] ~~`testvectors/ecdsa_secp224r1_sha224_p1363_test.json`~~
- [ ] ~~`testvectors/ecdsa_secp224r1_sha256_p1363_test.json`~~
- [x] `testvectors/ecdsa_secp224r1_sha3_224_test.json` (#5099)
- [x] `testvectors/ecdsa_secp224r1_sha3_256_test.json` (#5099)
- [x] `testvectors/ecdsa_secp224r1_sha3_512_test.json` (#5099)
- [ ] ~~`testvectors/ecdsa_secp224r1_sha512_p1363_test.json`~~
[...]

PR #5237, opened 2020-04-26
Refs #5075 -- added the remainder of the wycheproof rsa tests

alex:


alex:
FWIW I filed https://github.com/openssl/openssl/issues/11648 about the lack of ability to use the truncated sha2 hashes with RSA

PR #5240, opened 2020-04-29
Update symmetric-encryption.rst

Yoshi325:
mis-named parameter for `cryptography.hazmat.primitives.ciphers.Cipher`


alex:
Thanks for the PR. Doing an open/close trick to try to convince codecov to fix itself.

PR #5246, opened 2020-05-11
Write a bunch of primitives tests as subtests for performance

alex:


.github/workflows/download_openssl.py
```diff
@@ -12,7 +12,7 @@ def get_response(session, url, token):
     response = session.get(url, headers={"Authorization": "token " + token})
     if response.status_code != 200:
         raise ValueError("Got HTTP {} fetching {}: ".format(
-            response.status_code, url, response.content
```
reaperhulk, line 15: juh?

alex: flake8 started complaining about this, it was always wrong.

alex:
19:09:24 <Alex_Gaynor> reaperhulk: ah fuck, there's a problem with just doing a loop. `pytest.skip()` raises an exception, which means all subsequent tests in the bundle get skipped, not just that test.
19:10:00 <Alex_Gaynor> reaperhulk: this means we're skipping a ton of tests, and probably significantl contributes to the perf difference (e.g. none of the kbkdf ones are running, because the first thing in that vector is skipped)
19:10:58 <Alex_Gaynor> reaperhulk: this is why we need real subtest support

Since pytest-subtests is py3 only, I'm not sure what to do. 25% of our downloads are still py2, so just drop it isn't on the table ATM.

alex:
Converted to a draft, since I don't think we have a way forward until the far future when we're py3 only.

setup.py
```diff
@@ -84,6 +84,7 @@
         extras_require={
             "test": [
                 "pytest>=3.6.0,!=3.9.0,!=3.9.1,!=3.9.2",
```
reaperhulk, line 86: Might as well bump pytest lower limit to at least 4.4 then since pytest-subtests doesn't run on any older version

reaperhulk (approved):

PR #5251, opened 2020-05-16
Deprecate support for Python 2

alex:
🍿

### DIFF
--- a/src/cryptography/__init__.py
+++ b/src/cryptography/__init__.py
@@ -4,13 +4,25 @@
 
 from __future__ import absolute_import, division, print_function
 
+import sys
+import warnings
+
 from cryptography.__about__ import (
     __author__, __copyright__, __email__, __license__, __summary__, __title__,
     __uri__, __version__
 )
+from cryptography.utils import CryptographyDeprecationWarning
 
 
 __all__ = [
     "__title__", "__summary__", "__uri__", "__version__", "__author__",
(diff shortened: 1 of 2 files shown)


src/cryptography/__init__.py
```diff
@@ -4,13 +4,25 @@
...
+from cryptography.utils import CryptographyDeprecationWarning
 
 
 __all__ = [
     "__title__", "__summary__", "__uri__", "__version__", "__author__",
     "__email__", "__license__", "__copyright__",
 ]
+
+if sys.version_info[0] == 2:
+    warnings.warn(
```
webknjaz, line 23: Raising it on `import` was a bad idea because it's now unignorable.
```console
$ python27 -W ignore::cryptography.utils.CryptographyDeprecationWarning
Invalid -W option ignored: invalid module name: 'cryptography.utils'
```

And using [pytest's filterwarnings](https://docs.pytest.org/en/latest/warnings.html) also does not help:
```ini
filterwarnings =
    error
    ignore:Python 2 is no longer supported by the Python core team. Support for it is now deprecated in cryptography, and will be removed in a future release.:cryptography.utils.CryptographyDeprecationWarning
```
The above snippet makes pytest explode during startup because `from cryptography.utils import CryptographyDeprecationWarning` causes the import machinery (internally) to first `import cryptography` and hit the `__init__` that explodes.

PR #5256, opened 2020-05-24
Add a way to pass current time to Fernet

jstasiak:
The motivation behind this is to be able to unit test code using Fernet
easily without having to monkey patch global state.

### DIFF
--- a/docs/fernet.rst
+++ b/docs/fernet.rst
@@ -55,6 +55,8 @@ has support for implementing key rotation via :class:`MultiFernet`.
 
     .. method:: encrypt_at_time(data, current_time)
 
+       .. versionadded:: 3.0
+
        Encrypts data passed using explicitly passed current time. See
        :meth:`encrypt` for the documentation of the ``data`` parameter, the
        return type and the exceptions raised.
@@ -101,6 +103,8 @@ has support for implementing key rotation via :class:`MultiFernet`.
                            ``bytes``.
 
     .. method:: decrypt_at_time(token, ttl, current_time)
+
+       .. versionadded:: 3.0
 
        Decrypts a token using explicitly passed current time. See
        :meth:`decrypt` for the documentation of the ``token`` and ``ttl``
--- a/docs/fernet.rst
+++ b/docs/fernet.rst
@@ -56,7 +56,7 @@ has support for implementing key rotation via :class:`MultiFernet`.
     .. method:: encrypt_at_time(data, current_time)
 
        Encrypts data passed using explicitly passed current time. See
-       :meth:`encrypt` for the documentaiton of the ``data`` parameter, the
+       :meth:`encrypt` for the documentation of the ``data`` parameter, the
        return type and the exceptions raised.
 
        The motivation behind this method is for the client code to be able to
--- a/docs/fernet.rst
+++ b/docs/fernet.rst
@@ -100,7 +100,7 @@ has support for implementing key rotation via :class:`MultiFernet`.
         :raises TypeError: This exception is raised if ``token`` is not
                            ``bytes``.
 
-    .. method:: dectypy_at_time(token, ttl, current_time)
+    .. method:: decrypt_at_time(token, ttl, current_time)
 
        Decrypts a token using explicitly passed current time. See
        :meth:`decrypt` for the documentation of the ``token`` and ``ttl``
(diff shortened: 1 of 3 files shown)


jstasiak:
FWIW fernet-rs (Rust implementation) has the following (from the [README](https://github.com/mozilla-services/fernet-rs#testing-token-expiry)):

> By default fernet wraps operations in an attempt to be safe - you should never be able to "hold it incorrectly". But we understand that sometimes you need to be able to do some more complicated operations.
> The major example of this is having your application test how it handles tokens that have expired past their ttl.
> To support this, we allow you to pass in timestamps to the encrypt_at_time and decrypt_at_time functions, but these are behind a feature gate.

And the API they expose is:

```rust
fn encrypt(&self, data: &[u8]) -> String
fn encrypt_at_time(&self, data: &[u8], current_time: u64) -> String
fn decrypt(&self, token: &str) -> Result<Vec<u8>, DecryptionError>
fn decrypt_at_time(
        &self,
        token: &str,
        ttl: Option<u64>,
        current_time: u64,
    ) -> Result<Vec<u8>, DecryptionError>
```

markokr:
Alternative API would be to move the time.time() call from the middle of bigger function to separate method so it can be overrided by subclassing.   Thus core class implementation can offer more guaranteed black-box api.

Just an idea, I guess core devs need to hint which way they would prefer.

jstasiak:
Oh, sure. Alternatively-alternatively a method to get current time could be passed to `Fernet`'s constructor, stored in an instance attribute and called when necessary.

jstasiak:
Is there something I can do to get this merged? I'm happy to work out any design concerns you may have.

docs/fernet.rst
```diff
@@ -35,12 +35,15 @@ has support for implementing key rotation via :class:`MultiFernet`.
         they'll also be able forge arbitrary messages that will be
         authenticated and decrypted.
 
-    .. method:: encrypt(data)
+    .. method:: encrypt(data, current_time=None)
```
alex, line 38: I think we'd rather have an `encrypt_at_time()` API, instead of an optional parameter, as a way of encouraging people not to use it except in test :-)

jstasiak:
I'm not sure if the single CI job failure is caused by me, I don't really recognize it and I'm not sure what's it about.

alex:
https://github.com/pyca/cryptography/pull/5272 is up to fix it

jstasiak:
Is there something I can actually do about that 99.99% coverage to make it hit 100% or is it some coverage-counting quirk?

src/cryptography/fernet.py
```diff
@@ -146,13 +150,16 @@ def __init__(self, fernets):
...
+
+    def encrypt_at_time(self, msg, current_time):
+        return self._fernets[0].encrypt_at_time(msg, current_time)
 
     def rotate(self, msg):
         timestamp, data = Fernet._get_unverified_token_data(msg)
         for f in self._fernets:
             try:
-                p = f._decrypt_data(data, timestamp, None)
+                p = f._decrypt_data(data, timestamp, None, None)
```
reaperhulk, line 162: The [partial coverage](https://codecov.io/gh/pyca/cryptography/compare/eceea9d912994dbd84bfc7b27031c1254e675e28...3d1fadd7ffd2490df956aca725971db623e0ac15/src/tests/test_fernet.py?before=tests/test_fernet.py) looks like it's occurring because `time.time` is no longer called when `rotate` is called. This is because we now pass `None` for current_time, which is fine since we also pass a ttl of `None` so no comparison ever occurs.

The purpose of that test is to confirm that a `rotate` doesn't touch the timestamp when re-encrypting, so we can just modify it to check that via a call to `encrypt_at_time` and not monkeypatch at all.

jstasiak: Hah, you're obviously right – I failed to notice that in the Codecov interface. This is addressed now.

docs/fernet.rst
```diff
@@ -53,6 +53,26 @@ has support for implementing key rotation via :class:`MultiFernet`.
             generated in *plaintext*, the time a message was created will
             therefore be visible to a possible attacker.
 
+    .. method:: encrypt_at_time(data, current_time)
```
reaperhulk, line 56: This needs a `.. versionadded:: 3.0` notation (as does `decrypt_at_time`). Otherwise this LGTM, thanks for working with us here.

jstasiak: Thanks for the review again, done.

reaperhulk:
Two more CI breakages later we can actually merge this. Thanks for the effort!

jstasiak:
Yeah, well, that's on me – could've avoided at least some of this easily. ;) I'm happy to have this merged, thanks for all the work on your side.

PR #5264, opened 2020-06-01
Pin Certbot dependencies during testing

bmw:
It appears the latest version of one of Certbot's dependencies is broken on Python 2.7. See https://github.com/bear/parsedatetime/issues/246. This PR fixes things by pinning Certbot's dependencies to known working versions.

A custom installation script is used because we have multiple requirements files you'd have to use to pin everything down and some of them contain hashes which causes `pip` to error when installing packages in editable mode. We could resolve these problems a different way here if you prefer, but the script we use upstream takes care of these problems.

### DIFF
--- a/.travis/downstream.d/certbot.sh
+++ b/.travis/downstream.d/certbot.sh
@@ -5,8 +5,8 @@ case "${1}" in
         git clone --depth=1 https://github.com/certbot/certbot
         cd certbot
         git rev-parse HEAD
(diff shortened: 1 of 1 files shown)


reaperhulk:
We're happy to use whatever script you use, so `pip_install_editable.py` is fine. It minimizes the probability of bit rot 😄

alex:
@bmw really appreciate how quickly y'all jumped on this!

PR #5265, opened 2020-06-03
Clarify that static linkage only applies to wheels

twosigmajab:
This recently confused a security reviewer with less familiarity with how this works, so I thought this change could help avoid potential confusion in the future. Hope this helps!

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -72,9 +72,9 @@ New releases for OpenSSL updates
 --------------------------------
 
 As of versions 0.5, 1.0.1, and 2.0.0, ``cryptography`` statically links OpenSSL
-on Windows, macOS, and Linux respectively, to ease installation. Due to this,
-``cryptography`` will release a new version whenever OpenSSL has a security or
-bug fix release to avoid shipping insecure software.
+in binary distributions for Windows, macOS, and Linux respectively, to ease
+installation. Due to this, ``cryptography`` will release a new version whenever
+OpenSSL has a security or bug fix release to avoid shipping insecure software.
 
 Like all our other releases, this will be announced on the mailing list and we
 strongly recommend that you upgrade as soon as possible.


docs/security.rst
```diff
@@ -72,6 +72,7 @@ New releases for OpenSSL updates
 --------------------------------
 
 As of versions 0.5, 1.0.1, and 2.0.0, ``cryptography`` statically links OpenSSL
+in :ref`binary distrubtions <installation:Static Wheels>`
```
twosigmajab, line 75: I attempted to link to the https://cryptography.io/en/latest/installation/#static-wheels section of the docs, since it's relevant, but I just realized the way I linked to it requires the `sphinx.ext.autosectionlabel` extension which it looks like we don't have installed. Would you like a PR to add that extension before merging this?

twosigmajab: Noticed I had a type too 🤦  - fixed and removed the link in the latest revision.

docs/security.rst
```diff
@@ -72,9 +72,9 @@ New releases for OpenSSL updates
 --------------------------------
 
 As of versions 0.5, 1.0.1, and 2.0.0, ``cryptography`` statically links OpenSSL
-on Windows, macOS, and Linux respectively, to ease installation. Due to this,
-``cryptography`` will release a new version whenever OpenSSL has a security or
-bug fix release to avoid shipping insecure software.
+in binary distributions for Windows, macOS, and Linux respectively, to ease
+installation. Due to this, ``cryptography`` will release a new version whenever
+OpenSSL has a security or bug fix release to avoid shipping insecure software.
```
twosigmajab, line 77: This just adds "in binary distributions for" (replacing "in" with "for"), but also rewraps at 79 characters to match existing style (though it makes it harder to see the changes, figuring the consistent wrapping was preferred here, but if not I'm happy to not rewrap).

reaperhulk:
This is a reasonable change and we would have asked for it to be wrapped. Thanks!

twosigmajab:
Awesome, thanks for merging this!

PR #5270, opened 2020-06-11
Add the ability to send alerts directly

tshmul:


reaperhulk:
`ssl3_send_alert` isn't a public function in OpenSSL so this won't work. You'll need to look at OpenSSL's APIs to determine how alerts should actually be sent on an open session.

tshmul:
Thanks @reaperhulk , I'll take a further look at this one (It worked on my windows env, so I didn't think there is an issue)

PR #5276, opened 2020-06-14
Unify X.509 signature algorithm validation

markokr:
- Use common implementation
- OCSP signing was missing validation

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -732,25 +732,32 @@ class Backend(object):
     def create_cmac_ctx(self, algorithm):
         return _CMACContext(self, algorithm)
 
-    def create_x509_csr(self, builder, private_key, algorithm):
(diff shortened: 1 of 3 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1559,6 +1536,10 @@ def create_ocsp_request(self, builder):
         return _OCSPRequest(self, ocsp_req)
 
     def _create_ocsp_basic_response(self, builder, private_key, algorithm):
+        if not isinstance(builder, ocsp.OCSPResponseBuilder):
```
reaperhulk, line 1539: The reason we don't have this check here is because our OCSP response builder doesn't take a `backend` (since we're tentatively planning to abandon `backend` at some point due to lack of use). Since only we ever call `_create_ocsp_basic_response` (and it is a private method) we don't need to have the friendlier TypeError.

markokr: Ok, will remove.

PR #5280, opened 2020-06-17
Disallow ttl=None in (Multi)Fernet.decrypt_at_time()

jstasiak:
Since the introduction of the _at_time() methods in #5256[1] there's
been this little voice in the back of my mind telling me that maybe it's
not the best idea to allow ttl=None in decrypt_at_time(). It's been like
this for convenience and code reuse reasons.

Then I submitted a patch for cryptography stubs in typeshed[2] and I had
to decide whether to define decrypt_at_time()'s ttl as int and be
incompatible with cryptography's behavior or Optional[int] and advertise
an API that can be misused much too easily. I went ahead with int.

Considering the above I decided to propose this patch. Some amount of
redundancy (and a new test to properly cover the
MultiFernet.decrypt_at_time() implementation) is a price to prevent
clients from shooting themselves in the foot with the tll=None gun since
setting ttl to None disabled timestamp checks even if current_time was
provided.

For your consideration @alex and @reaperhulk.

[1] https://github.com/pyca/cryptography/pull/5256
[2] https://github.com/python/typeshed/pull/4238

### DIFF
--- a/src/cryptography/fernet.py
+++ b/src/cryptography/fernet.py
@@ -73,9 +73,12 @@ class Fernet(object):
         return base64.urlsafe_b64encode(basic_parts + hmac)
 
     def decrypt(self, token, ttl=None):
-        return self.decrypt_at_time(token, ttl, int(time.time()))
+        timestamp, data = Fernet._get_unverified_token_data(token)
+        return self._decrypt_data(data, timestamp, ttl, int(time.time()))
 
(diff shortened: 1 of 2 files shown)


jstasiak:
The one build failure seems like a fluke: https://travis-ci.org/github/pyca/cryptography/jobs/699472890

PR #5288, opened 2020-06-26
constrain RSA key generation more heavily

reaperhulk:
OpenSSL 3.0.0 is going to enforce this but we might as well get more proscriptive now.

### DIFF
--- a/docs/hazmat/primitives/asymmetric/rsa.rst
+++ b/docs/hazmat/primitives/asymmetric/rsa.rst
@@ -20,6 +20,8 @@ mathematical properties`_.
 
     .. versionchanged:: 3.0
 
+        Tightened restrictions on ``public_exponent``.
+
     Generates a new RSA private key using the provided ``backend``.
     ``key_size`` describes how many :term:`bits` long the key should be. Larger
     keys provide more security; currently ``1024`` and below are considered
--- a/docs/hazmat/primitives/asymmetric/rsa.rst
+++ b/docs/hazmat/primitives/asymmetric/rsa.rst
@@ -18,6 +18,8 @@ mathematical properties`_.
 
     .. versionadded:: 0.5
 
+    .. versionchanged:: 3.0
+
     Generates a new RSA private key using the provided ``backend``.
     ``key_size`` describes how many :term:`bits` long the key should be. Larger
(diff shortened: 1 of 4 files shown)


src/cryptography/hazmat/primitives/asymmetric/rsa.py
```diff
@@ -120,8 +120,11 @@ def generate_private_key(public_exponent, key_size, backend):
 
 
 def _verify_rsa_parameters(public_exponent, key_size):
-    if public_exponent < 3:
-        raise ValueError("public_exponent must be >= 3.")
+    if public_exponent != 3 and public_exponent < 65537:
```
alex, line 123: Can this just be `public_expontent not in [3, 65537]`? Do we care about other values?

reaperhulk: We can constrain all the way to just those two. I was matching OpenSSL 3.0 but I'm fine with being more strict.

ronf: I'd be hesitant to restrict this more than OpenSSL does.

reaperhulk: We still parse any existing key, we just won't generate keys with weird values. I'm unaware of any production utility for large public exponents except for stunt public key tricks.

docs/hazmat/primitives/asymmetric/rsa.rst
```diff
@@ -18,6 +18,8 @@ mathematical properties`_.
 
     .. versionadded:: 0.5
 
+    .. versionchanged:: 3.0
```
alex, line 21: Explain how it was changed, "tightened restrictions on `public_expontent`

src/cryptography/hazmat/primitives/asymmetric/rsa.py
ronf: If it's only an issue when generating new keys from within PyCA, that's less of a concern, but I would still worry about someone porting OpenSSL-based code from another language or trying to build something using PyCA which interoperates with another OpenSSL-based application. If that application happened to require another exponent, you're making things hard for the developer using PyCA, and I'm not sure this stricter check is adding much value.

PR #5290, opened 2020-06-29
Pass test_invalid_version_cert if the certificate is rejected.

davidben:
OpenSSL currently accepts certificates with invalid versions, leaving
the `cert.version` getter to reject unknown ones. Stricter X.509
implementations may reject them at parse time. Tolerate this in the
tests.

### DIFF
--- a/tests/x509/test_x509.py
+++ b/tests/x509/test_x509.py
@@ -1000,11 +1000,18 @@ class TestRSACertificate(object):
         assert cert.version is x509.Version.v3
 
     def test_invalid_version_cert(self, backend):
-        cert = _load_cert(
-            os.path.join("x509", "custom", "invalid_version.pem"),
-            x509.load_pem_x509_certificate,
-            backend
-        )
+        try:
+            cert = _load_cert(
+                os.path.join("x509", "custom", "invalid_version.pem"),
+                x509.load_pem_x509_certificate,
+                backend
+            )
+        except ValueError:
+            # OpenSSL 1.1.1 accepts certificates with an invalid version, but
+            # future versions or other libraries may reject this. _load_cert
+            # will then fail to parse. Consider that a passing result.
+            return
+
         with pytest.raises(x509.InvalidVersion) as exc:
             cert.version
 


tests/x509/test_x509.py
```diff
@@ -1000,11 +1000,18 @@ def test_generalized_time_not_after_cert(self, backend):
...
-            x509.load_pem_x509_certificate,
-            backend
-        )
+        try:
+            cert = _load_cert(
+                os.path.join("x509", "custom", "invalid_version.pem"),
+                x509.load_pem_x509_certificate,
+                backend
+            )
+        except ValueError:
```
alex, line 1009: I don't suppose BoringSSL puts something on the error stack so that we can tell it's an `x509.InvalidVersion` and raise that?

davidben: It does, but given that the rest of cryptography.io doesn't build with BoringSSL anyway, I figured it wasn't worth doing anything more complicated there. (Also still TBD whether the change will stick. If I have to revert it and try later, the error code will disappear.)

Right now the errors are just dropped, for better or worse. Dunno if that's expected:
https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/backends/openssl/backend.py#L1194-L1199

davidben: (If you all don't want to support this, that's totally fine. I can carry a patch on my end easily enough.)

alex: Ooof. We need to figure out a strategy either way, since this results in a coverage issue (we enforce 100% coverage).

@reaperhulk any obvious theories of how to handle this?

alex:
Ok, here's a new proposal: Let's always raise on cert load for bad version, even on OpenSSL, instead of raising in the `version` accessor.

This will give us consistent behavior, and make testing easier.

reaperhulk:
I have no idea what the breakage is going to be here, but it seems unlikely to be high. You have to try to create a cert with an invalid version. Let's switch to erroring on parse and see what happens in the real world I guess.

reaperhulk:
#5299 continues this

PR #5298, opened 2020-07-05
Support parsing SCTs in OCSPResponse

reaperhulk:
replaces #4963 

Let's see what I missed after a ton of rebasing and modifications...

Note to self: we do not support generation of single_extensions via OCSPResponseBuilder at this time.


docs/x509/reference.rst
```diff
@@ -2099,6 +2099,33 @@ X.509 Extensions
...
 
+.. class:: SignedCertificateTimestamps(scts)
+
+    .. versionadded:: 3.0
+
+    This extension contains
+    :class:`~cryptography.x509.certificate_transparency.SignedCertificateTimestamp`
+    instances. These can be used to verify that the certificate is included
+    in a public Certificate Transparency log. This extension is typically found
+    in OCSP responses. For SCTs in an X.509 certificate see
```
alex, line 2110: Typically, or only?

reaperhulk: I actually don't know. I wrote typically as a weasel word 😄 Is it solely in OCSP responses?

alex:
Changelog as well

PR #5299, opened 2020-07-05
Enforce that X.509 version is valid on parse

alex:
Closes #5290

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,9 @@ Changelog
   :meth:`~cryptography.hazmat.primitives.asymmetric.rsa.generate_private_key`
   no longer accepts ``public_exponent`` values except 65537 and 3 (the latter
   for legacy purposes).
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -39,6 +39,9 @@ Changelog
   X.509 extension.
 * Added support for parsing
   :class:`~cryptography.x509.SignedCertificateTimestamps` in OCSP responses.
+* X.509 certificate parsing now enforces that the ``version`` field contains a
```
reaperhulk, line 42: We should probably flag this as backwards incompatible right? Also, if you put this with the other incompatible entries we'll avoid a conflict with another PR landing 😆

alex: My thought was that it's such a minor change, there was no need to call it out in bold. WDYT?

reaperhulk: I think people use us in parsers that scrape every cert they can find so this will potentially break them and we should call it out even though it's hypothetically a very rare case.

PR #5301, opened 2020-07-05
support 4096 bit DSA parsing from numbers classes

### LINKED ISSUE #5227: p must be exactly 1024, 2048, or 3072 bits long
Hi, I'm getting an error while downloading files from SFTP using pysftp. The source of the error is due to the limitation of the number of bits the parameters can take on the cryptography package.

The source of the issue in the below method in **cryptography\hazmat\primitives\asymmetric\dsa.py**

`def _check_dsa_parameters(parameters):
    if parameters.p.bit_length() not in [1024, 2048, 3072, **4096**]:
        raise ValueError("p must be exactly 1024, 2048, or 3072 bits long")
    if parameters.q.bit_length() not in [160, 224, 256]:
        raise ValueError("q must be exactly 160, 224, or 256 bits long")    
    if not (1 < parameters.g < parameters.p):
        raise ValueError("g, p don't satisfy 1 < g < p.")`

https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/primitives/asymmetric/dsa.py ... line no 131

By adding the 4096 bits in the bit length check fixed the issue. Can you please make this fix in the library.

python 3.6
cryptography 2.9.2
pip 20.0.2
setuptools 46.1.3

Used requirements.txt to install Cryptography
`cryptography==2.9.2`

Please let me know if you need any additional details.

reaperhulk:
fixes #5227 for bad legacy apps that want DSA but also are willing to use large key sizes. Adds a bunch of "don't use DSA" language to the docs since you should strive to not use DSA. ECDSA and EdDSA exist! Use those!


alex:
flake8

docs/hazmat/primitives/asymmetric/dsa.rst
```diff
@@ -5,6 +5,12 @@ DSA
 
 .. module:: cryptography.hazmat.primitives.asymmetric.dsa
 
+.. note::
+
+    DSA is a **legacy algorithm** and should generally be avoided in favor of
+    choices like :doc:`ECDSA</hazmat/primitives/asymmetric/ec>` or
+    :doc:`EdDSA using curve25519</hazmat/primitives/asymmetric/ed25519>`.
```
alex, line 12: Can you swap the order and list ed25519 first?

docs/hazmat/primitives/asymmetric/dsa.rst
```diff
@@ -14,15 +20,18 @@ Generation
 
     .. versionadded:: 0.5
 
+    .. versionchanged:: 3.0
+
+        Added support for 4096-bit keys for some legacy applications that
+        continue to use DSA despite the wider cryptographic community's
```
alex, line 26: Technically our docs style guide says we shoudl cite things like that. Can you cite fillipo's latest post or something?

PR #5312, opened 2020-07-18
fix indexing on X509 request attribute value

reaperhulk:
Previously this improperly used the position of the X509_ATTRIBUTE as the position of the type/value we wanted from the attribute. Now we assert that the supported attributes are single-valued, which is all we're planning to support at this time, and get the 0 index.


alex:
Test case?

PR #5315, opened 2020-07-18
support encoding attributes via CertificateSigningRequestBuilder

reaperhulk:
I think this probably needs some discussion. This API makes assumptions about what ASN.1 type the provided values should be encoded as. We previously made similar assumptions with `NameAttribute` and eventually added the `_ASN1Type` to allow for alternate string type encodings since that wasn't good enough.

I see two major options here:
1) Assume that we can get away with the encoding as described and leave the API simple until such time as we're proven wrong. At that point we can make a new object and figure out how to handle the legacy getters (the `get_attribute_for_oid` method should return an object if we change this) and setters.
2) Shoulder the complexity now. In that case we need to decide if we're building a new class called `Attribute` or trying to reuse `NameAttribute` (which is an attribute with some limited validation we have discussed moving elsewhere in #3857 and in https://github.com/pyca/cryptography/pull/3862).

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -812,7 +812,8 @@ class Backend(object):
         for attr_oid, attr_val in builder._attributes:
             obj = _txt2obj_gc(self, attr_oid.dotted_string)
             res = self._lib.X509_REQ_add1_attr_by_OBJ(
-                x509_req, obj, 12, attr_val, len(attr_val))
+                x509_req, obj, x509.name._ASN1Type.UTF8String.value,
+                attr_val, len(attr_val))
             self.openssl_assert(res == 1)
 
         # Sign the request using the requester's private key.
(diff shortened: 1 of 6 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -808,6 +808,13 @@ def create_x509_csr(self, builder, private_key, algorithm):
         res = self._lib.X509_REQ_add_extensions(x509_req, sk_extension)
         self.openssl_assert(res == 1)
 
+        # Add attributes (all bytes encoded as ASN1 UTF8_STRING)
+        for attr_oid, attr_val in builder._attributes:
+            obj = _txt2obj_gc(self, attr_oid.dotted_string)
+            res = self._lib.X509_REQ_add1_attr_by_OBJ(
+                x509_req, obj, 12, attr_val, len(attr_val))
```
alex, line 815: There has got to be a constant for this.

reaperhulk:
(It has been decided we're going to go for simple and cross our fingers)

PR #5325, opened 2020-07-19
PKCS12 support

reaperhulk:
continues #5308

Some outstanding questions still here but wanted to push it up

### DIFF
--- a/docs/hazmat/primitives/asymmetric/serialization.rst
+++ b/docs/hazmat/primitives/asymmetric/serialization.rst
@@ -489,6 +489,37 @@ file suffix.
         ``additional_certificates`` is a list of all other
         :class:`~cryptography.x509.Certificate` instances in the PKCS12 object.
 
+.. function:: generate_pkcs12(password, name, key, cert, cas, key_encryption,
+                              cert_encryption)
+
+    Serialize a PKCS12 blob.
+
+    :param name: The friendly name to use for the supplied certificate and key.
+    :type name: :term:`bytes-like`
+
+    :param password: The password to use to encrypt the data. ``None``
+        if the PKCS12 should not be encrypted.
+    :type password: :term:`bytes-like`
+
+    :param key: The private key to include in the structure.
+    :type key: :term:`bytes-like`
+
+    :param cert: The private-key's certificates.
+    :type cert: :class:`~cryptography.x509.Certificate`
+
+    :param cas: An optional set of certificates to also include in the structure.
+    :type cas: :term:`bytes-like`
+
+    :param key_encryption: The encryption algorithms that should be used for the key.
+        An instance of an object conforming to the :class:`~cryptography.hazmat.primitives.serialization.KeySerializationEncryption`
+        interface.
+
+    :param cert_encryption: The encryption algorithms that should be used for the certificate.
+        An instance of an object conforming to the :class:`~cryptography.hazmat.primitives.serialization.KeySerializationEncryption`
+        interface.
+
+    :return bytes: Serialized PKCS12.
+
 Serialization Formats
 ~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/hazmat/primitives/asymmetric/serialization.rst
+++ b/docs/hazmat/primitives/asymmetric/serialization.rst
@@ -514,7 +514,7 @@ file suffix.
         :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPrivateKeyWithSerialization`
         object.
 
-    :param cert: The private-key's certificates.
+    :param cert: The certificate associated with the private key.
     :type cert: :class:`~cryptography.x509.Certificate` or ``None``
 
     :param cas: An optional set of certificates to also include in the structure.
--- a/docs/hazmat/primitives/asymmetric/serialization.rst
+++ b/docs/hazmat/primitives/asymmetric/serialization.rst
@@ -489,8 +489,7 @@ file suffix.
         ``additional_certificates`` is a list of all other
         :class:`~cryptography.x509.Certificate` instances in the PKCS12 object.
 
-.. function:: serialize_key_and_certificates(name, key, cert, cas,
-                                             encryption_algorithm)
+.. function:: serialize_key_and_certificates(name, key, cert, cas, encryption_algorithm)
 
     .. versionadded:: 3.0
 
@@ -504,11 +503,11 @@ file suffix.
     Serialize a PKCS12 blob.
 
     :param name: The friendly name to use for the supplied certificate and key.
-    :type name: :term:`bytes`
+    :type name: bytes
 
     :param key: The private key to include in the structure.
     :type key: An
-        :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKeyWithSerialization
+        :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKeyWithSerialization`
         ,
         :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePrivateKeyWithSerialization`
         , or
--- a/docs/hazmat/primitives/asymmetric/serialization.rst
+++ b/docs/hazmat/primitives/asymmetric/serialization.rst
@@ -489,7 +489,7 @@ file suffix.
         ``additional_certificates`` is a list of all other
         :class:`~cryptography.x509.Certificate` instances in the PKCS12 object.
 
-.. function:: serialize_key_and_certificates(password, name, key, cert, cas,
+.. function:: serialize_key_and_certificates(name, key, cert, cas,
                                              key_encryption, cert_encryption)
 
     .. versionadded:: 3.0
@@ -497,30 +497,34 @@ file suffix.
     Serialize a PKCS12 blob.
 
     :param name: The friendly name to use for the supplied certificate and key.
-    :type name: :term:`bytes-like`
-
-    :param password: The password to use to encrypt the data. ``None``
-        if the PKCS12 should not be encrypted.
-    :type password: :term:`bytes-like`
+    :type name: :term:`bytes`
 
     :param key: The private key to include in the structure.
-    :type key: :term:`bytes-like`
+    :type key: An
+        :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKeyWithSerialization
+        ,
+        :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePrivateKeyWithSerialization`
+        , or
+        :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPrivateKeyWithSerialization`
+        object.
 
     :param cert: The private-key's certificates.
     :type cert: :class:`~cryptography.x509.Certificate`
 
     :param cas: An optional set of certificates to also include in the structure.
-    :type cas: :term:`bytes-like`
+    :type cas: list of :class:`~cryptography.x509.Certificate`
 
     :param key_encryption: The encryption algorithms that should be used for the
         key. An instance of an object conforming to the :
         class:`~cryptography.hazmat.primitives.serialization.KeySerializationEncryption`
-        interface.
+        interface. When using both ``key_encryption`` and ``cert_encryption``
+        the passwords **must** be the same.
 
     :param cert_encryption: The encryption algorithms that should be used for
         the certificate. An instance of an object conforming to the
         :class:`~cryptography.hazmat.primitives.serialization.KeySerializationEncryption`
-        interface.
+        interface. When using both ``key_encryption`` and ``cert_encryption``
+        the passwords **must** be the same.
 
     :return bytes: Serialized PKCS12.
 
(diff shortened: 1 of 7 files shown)


docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -489,6 +489,41 @@ file suffix.
...
+    :type password: :term:`bytes-like`
+
+    :param key: The private key to include in the structure.
+    :type key: :term:`bytes-like`
+
+    :param cert: The private-key's certificates.
+    :type cert: :class:`~cryptography.x509.Certificate`
+
+    :param cas: An optional set of certificates to also include in the structure.
+    :type cas: :term:`bytes-like`
```
alex, line 513: No

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -489,6 +489,41 @@ file suffix.
...
+
+    :param name: The friendly name to use for the supplied certificate and key.
+    :type name: :term:`bytes-like`
+
+    :param password: The password to use to encrypt the data. ``None``
+        if the PKCS12 should not be encrypted.
+    :type password: :term:`bytes-like`
+
+    :param key: The private key to include in the structure.
+    :type key: :term:`bytes-like`
```
alex, line 507: I don't think so?

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -489,6 +489,41 @@ file suffix.
...
+    .. versionadded:: 3.0
+
+    Serialize a PKCS12 blob.
+
+    :param name: The friendly name to use for the supplied certificate and key.
+    :type name: :term:`bytes-like`
+
+    :param password: The password to use to encrypt the data. ``None``
+        if the PKCS12 should not be encrypted.
+    :type password: :term:`bytes-like`
```
alex, line 504: This doesn't match the other in the signature.

Also the encryption algorithms have a password, so why does this have an additional password argument?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2397,6 +2397,88 @@ def load_key_and_certificates_from_pkcs12(self, data, password):
...
+                  dsa.DSAPrivateKeyWithSerialization,
+                  ec.EllipticCurvePrivateKeyWithSerialization)):
+            raise TypeError(
+                "Key must be RSA, DSA, or EllipticCurve private key."
+            )
+        if cert is not None and not isinstance(cert, x509.Certificate):
+            raise TypeError("cert must be a certificate")
+
+        if key is None:
+            nid_key = 0
```
alex, line 2418: Do we trust OpenSSL's default?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2397,6 +2397,88 @@ def load_key_and_certificates_from_pkcs12(self, data, password):
...
+                              serialization.KeySerializationEncryption):
+                raise TypeError(
+                    "Key encryption algorithm must be a "
+                    "KeySerializationEncryption instance"
+                )
+            if isinstance(key_encryption, serialization.NoEncryption):
+                nid_key = -1
+            elif isinstance(key_encryption,
+                            serialization.BestAvailableEncryption):
+                nid_key = 0  # sensible default will be used
```
alex, line 2430:
```suggestion
                nid_key = 0  # OpenSSL's default will be used
```

tests/hazmat/primitives/test_pkcs12.py
```diff
@@ -137,3 +139,138 @@ def test_buffer_protocol(self, backend):
         assert parsed_key is not None
         assert parsed_cert is not None
         assert parsed_more_certs == []
+
+    @pytest.fixture(scope='function')
```
alex, line 143: As I said on the other PR, I'm not wild about this fixtures pattern.

Even ignoring that, cert vs. cert2 is terrible naming.

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: No:

```
These defaults are: 40 bit RC2 encryption for certificates, triple DES encryption for private keys, a key iteration count of PKCS12_DEFAULT_ITER (currently 2048) and a MAC iteration count of 1.

The default MAC iteration count is 1 in order to retain compatibility with old software which did not interpret MAC iteration counts. If such compatibility is not required then mac_iter should be set to PKCS12_DEFAULT_ITER.

```
https://www.openssl.org/docs/man1.1.1/man3/PKCS12_create.html

Compatibility is likely to be a real problem here, but we should at least do AES before we merge this.

docs/hazmat/primitives/asymmetric/serialization.rst
reaperhulk: They should not, but also you can only have one password with OpenSSL's API so we'll need to enforce that both passwords are the same (if present).

reaperhulk: Err, it should not have this argument. Not sure why I wrote "They should not"

alex: Or we can only allow passing one parameter for the time being :-)

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -489,6 +489,45 @@ file suffix.
         ``additional_certificates`` is a list of all other
         :class:`~cryptography.x509.Certificate` instances in the PKCS12 object.
 
+.. function:: serialize_key_and_certificates(name, key, cert, cas,
+                                             key_encryption, cert_encryption)
```
alex, line 493: I think you removed `cert_encryption`?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2397,6 +2397,76 @@ def load_key_and_certificates_from_pkcs12(self, data, password):
...
+
+        if cas is None:
+            sk_x509 = self._ffi.NULL
+        else:
+            cas = list(cas)
+            if not all(isinstance(ca, x509.Certificate) for ca in cas):
+                raise TypeError("cert in cas must be a certificate")
+
+            sk_x509 = self._lib.sk_X509_new_null()
+            sk_x509 = self._ffi.gc(sk_x509, self._lib.sk_X509_free)
```
alex, line 2447: Does this need the `sk_x509_pop_free` thingy?

src/cryptography/hazmat/primitives/serialization/pkcs12.py
```diff
@@ -4,6 +4,17 @@
...
 from __future__ import absolute_import, division, print_function
 
+from cryptography.hazmat.backends import default_backend
+
 
 def load_key_and_certificates(data, password, backend):
     return backend.load_key_and_certificates_from_pkcs12(data, password)
+
+
+def serialize_key_and_certificates(name, key, cert, cas, key_encryption):
```
alex, line 14: It's not really `key_encryption`, it's just `encrytion_params`, right?

tests/hazmat/primitives/test_pkcs12.py
```diff
@@ -137,3 +139,113 @@ def test_buffer_protocol(self, backend):
...
+        assert parsed_cert == cert
+        assert parsed_key.private_numbers() == key.private_numbers()
+        assert parsed_more_certs == [cert2]
+
+        with pytest.raises(TypeError) as exc:
+            serialize_key_and_certificates(None, key, cert, cert2, encryption)
+
+        with pytest.raises(TypeError) as exc:
+            serialize_key_and_certificates(None, key, cert, [key], encryption)
+        assert str(exc.value) == 'cert in cas must be a certificate'
```
alex, line 205: We don't usually do these assertions, do we?

alex:
There's also a docs thing in teh changelog

docs/hazmat/primitives/asymmetric/serialization.rst
tomers: Paul wrote:
> I am open to opinions on whether we should have separate cert encryption and key encryption options or if we should merge the two. I think separation might be necessary for reasonable compatibility though,

https://github.com/pyca/cryptography/pull/5308#issuecomment-660531477

src/cryptography/hazmat/primitives/serialization/pkcs12.py
tomers: Agree

src/cryptography/hazmat/backends/openssl/backend.py
tomers: Not sure about it.

src/cryptography/hazmat/primitives/serialization/pkcs12.py
tomers, line 14: You removed cert_encryption. See above as to why both should exist (please discuss this with Paul)

tests/hazmat/primitives/test_pkcs12.py
tomers: What do you mean exactly? The check for assertion message?

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: (Defaults to aes-256-cbc now)

tests/hazmat/primitives/test_pkcs12.py
reaperhulk: Removed

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: It does not -- verified via a new memleak test I just pushed. We missed the PKCS12_free though 😄

tests/hazmat/primitives/test_pkcs12.py
reaperhulk: We're inconsistent about it, but I'm fine with leaving them here so they're still in the latest commit.

src/cryptography/hazmat/primitives/serialization/pkcs12.py
reaperhulk: I think we're going to see what happens with a relatively aggressive default and then we can expand support from there. One of the challenges in this space is that everyone leaves everything in terrible legacy modes even when it's not necessary because they're afraid of breakage. In this case we can be (relatively) aggressive at first and then expand compatibility (if necessary) without breaking existing consumers.

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -489,6 +489,46 @@ file suffix.
...
+        reasons only.
+
+    Serialize a PKCS12 blob.
+
+    :param name: The friendly name to use for the supplied certificate and key.
+    :type name: :term:`bytes`
+
+    :param key: The private key to include in the structure.
+    :type key: An
+        :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKeyWithSerialization
```
alex, line 511:
```suggestion
        :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKeyWithSerialization`
```

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -489,6 +489,46 @@ file suffix.
         ``additional_certificates`` is a list of all other
         :class:`~cryptography.x509.Certificate` instances in the PKCS12 object.
 
+.. function:: serialize_key_and_certificates(name, key, cert, cas,
+                                             encryption_algorithm)
```
alex, line 493:
```suggestion
.. function:: serialize_key_and_certificates(name, key, cert, cas, encryption_algorithm)
```

src/cryptography/hazmat/primitives/serialization/pkcs12.py
```diff
@@ -4,17 +4,43 @@
...
+
+    if not isinstance(
+        encryption_algorithm, serialization.KeySerializationEncryption
+    ):
+        raise TypeError(
+            "Key encryption algorithm must be a "
+            "KeySerializationEncryption instance"
+        )
+
+    if key is None and cert is None and (cas is None or not cas):
```
tomers, line 41: `cas is None or not cas`: isn't that equvalent to `not cas` since `None` is falsy?

tomers:
@reaperhulk and @alex thanks for your work! I'm really happy to see my PR matures into this :-)
Please update the release milestone with this new PR: https://github.com/pyca/cryptography/milestone/32

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -489,6 +489,45 @@ file suffix.
...
+    :param key: The private key to include in the structure.
+    :type key: An
+        :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKeyWithSerialization`
+        ,
+        :class:`~cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePrivateKeyWithSerialization`
+        , or
+        :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPrivateKeyWithSerialization`
+        object.
+
+    :param cert: The private-key's certificates.
```
reaperhulk, line 517: Should be singular

PR #5337, opened 2020-07-21
fixed copyright year

alex:


reaperhulk:
codecov never even reported on this one 😆

PR #5340, opened 2020-07-22
what happens if we do this

reaperhulk:


reaperhulk:
Didn't work :) (Worth a shot though!)

markokr:
It did not crash on arm64... Instead some amd64-s had random error.

reaperhulk:
Yeah it ignored the key and booted on an x86_64 builder. Presumably we have to be allowlisted to use that key (which is not publicly stated yet, I spotted it in a travis commit).

markokr:
Heh, that was indeed sneaky.

PR #5345, opened 2020-07-25
[WIP] make OCSP conditional

reaperhulk:
We probably want to have the builders and loaders properly fail saying OCSP support isn't present before we merge this.


alex:
this needs to rebase on master to pass

reaperhulk:
It needs more than that 😄 #5400 is also needed + some more skip logic/new paths to handle the idea that it may not be possible to encode an extension.

reaperhulk:
(This becomes possible when we finish oxidizing so just closing this)

PR #5351, opened 2020-07-26
remove idna support finally

reaperhulk:
Nearly 3 years after deprecating one of our greatest API mistakes we finally get to remove it.

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -432,9 +432,6 @@ X.509 Certificate Object
         :raises cryptography.x509.UnsupportedGeneralNameType: If an extension
             contains a general name that is not supported.
 
-        :raises UnicodeError: If an extension contains IDNA encoding that is
-            invalid or not compliant with IDNA 2008.
-
         .. doctest::
 
             >>> for ext in cert.extensions:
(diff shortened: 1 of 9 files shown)


docs/x509/reference.rst
```diff
@@ -432,9 +432,6 @@ X.509 Certificate Object
         :raises cryptography.x509.UnsupportedGeneralNameType: If an extension
             contains a general name that is not supported.
 
-        :raises UnicodeError: If an extension contains IDNA encoding that is
```
reaperhulk, line 435: This has been untrue since 2.1 and we never noticed. Since we don't compute U-labels we won't error.

docs/x509/reference.rst
```diff
@@ -1412,24 +1406,19 @@ General Name Classes
...
-        a U-label. This breaks backwards compatibility, but only for
-        internationalized domain names.
+    .. versionchanged:: 3.1
 
+        :term:`U-label` support has been removed. Encode them to
+        :term:`A-label` before use.
 
     This corresponds to an email address. For example, ``user@example.com``.
 
     :param value: The email address. If the address contains an
```
reaperhulk, line 1416: We require these to be text (which we encode to bytes via `.encode('ascii')`), but should we loosen these to allow bytes now?

alex: Nah, I don't see why.

alex:
flake8

PR #5352, opened 2020-07-26
start trying to make our error handling a bit more generic

reaperhulk:
This adds a some helpers for "friendly" openssl error messages and removes error matching in three places.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -2390,13 +2390,13 @@ class Backend(object):
             length,
         )
         if res != 1:
-            self._consume_errors()
+            errors = self._consume_errors_with_text()
             # memory required formula explained here:
             # https://blog.filippo.io/the-scrypt-parameters/
             min_memory = 128 * n * r // (1024 ** 2)
             raise MemoryError(
                 "Not enough memory to derive key. These parameters require"
-                " {} MB of memory.".format(min_memory)
+                " {} MB of memory.".format(min_memory), errors
             )
         return self._ffi.buffer(buf)[:]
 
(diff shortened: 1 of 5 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1473,18 +1476,7 @@ def elliptic_curve_supported(self, curve):
...
-                    self._lib.ERR_LIB_EC, self._lib.EC_R_UNKNOWN_GROUP
-                )
-                or
-                # This occurs in FIPS mode for unsupported curves on RHEL
-                errors[0]._lib_reason_match(
-                    self._lib.ERR_LIB_EC, self._lib.EC_R_NOT_A_NIST_PRIME
-                )
-            )
+            self._consume_errors()
             return False
```
reaperhulk, line 1480: This is a supported check and ultimately it doesn't matter why group is null, so we should just return false after consuming the error stack.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2398,19 +2390,7 @@ def derive_scrypt(self, key_material, salt, length, n, r, p):
             length,
         )
         if res != 1:
-            errors = self._consume_errors()
```
reaperhulk, line 2401: This is a bit more debatable. We will error with MemoryError no matter what so it won't proceed unsafely, but it is hypothetically possible this wouldn't be due to OOM. We could hedge by attaching the error stack here like we do in DH

src/cryptography/hazmat/backends/openssl/dh.py
```diff
@@ -149,8 +139,11 @@ def exchange(self, peer_public_key):
         )
 
         if res == -1:
-            errors = self._backend._consume_errors()
-            return _handle_dh_compute_key_error(errors, self._backend)
+            errors_with_text = self._backend._consume_errors_with_text()
```
reaperhulk, line 142: We only know of an invalid public key here, but we want to value error no matter what and this attaches errors_with_text to the ValueError to provide more context.

alex:
flake8 and libre are broken

src/cryptography/hazmat/backends/openssl/backend.py
alex: Attaching the error stack feels safe in basically all cases.

PR #5357, opened 2020-07-27
Port a tiny tiny bit of the ASN.1 parsing to Rust

alex:
Blocked on:
- [x] https://github.com/alex/rust-asn1/issues/81


alex:
We'll need https://github.com/PyO3/setuptools-rust/pull/68 before it's possible to consider something like this.

reaperhulk (approved):
Alex and I pair reviewed this (and rust-asn1) over a few hours, so the lack of comments before approval is not indicative of the level of review scrutiny.

PR #5358, opened 2020-07-28
Travis CI: test on a range of architectures

daxtens:
Travis CI supports arm64, s390x and ppc64le ("Power") now. All of
those platforms have big distros shipping Python and cryptography, so
it would be nice to test on them.

I've picked a small subset of tests to build: just Python 3.8 and the
most recent version of OpenSSL and LibreSSL. I think that provides
decent coverage without bloating the test matrix too much. (We could
also test on pypy, for example, but I'm not sure that provides enough
value to be worth the extra runtime.)

s390x currently fails due to something to do with cffi. Directly
installing cffi first works, but it should be automatically pulled in
already. It's all a bit weird. Testing on arm64 and ppc64le is still
valuable though, so this can be investigated later.

I'm happy to talk about how this might interact with #5341 and the
best way forward for testing on a broader range of platforms.

I have access to some Power systems at work so if there are any
issues in future feel free to tag me in and I can have a look.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -123,6 +123,19 @@ matrix:
           # Tests hangs when run under bionic
           dist: xenial
 
+        - python: 3.8
+          env: TOXENV=py38 OPENSSL=1.1.1g
+          arch: arm64
+        - python: 3.8
+          env: TOXENV=py38 OPENSSL=1.1.1g
+          arch: ppc64le
+        - python: 3.8
+          env: TOXENV=py38 LIBRESSL=3.2.0
+          arch: arm64
+        - python: 3.8
+          env: TOXENV=py38 LIBRESSL=3.2.0
+          arch: ppc64le
+
 install:
     - ./.travis/install.sh
 


reaperhulk:
We appreciate the contribution, but we did this on https://github.com/pyca/cryptography/pull/5192 and then reverted it in https://github.com/pyca/cryptography/pull/5230. The reality is that our CI is already too slow and this pushed it to unbearable levels. We're currently working with OpenDev in https://github.com/pyca/cryptography/pull/5341 to hopefully gain access to more performant AArch64 builders though.

daxtens:
OK, fair enough. The ppc64le builds seemed to run fairly quickly (about 11
mins in my tests, although I don't know if I used caching that needs to be
turned off). If that's too slow there's more direct ppc64le Jenkins access
we could get through https://osuosl.org/services/powerdev/request_powerci/.
I'm happy to work with you to get that set up if you'd be open to it.

>

reaperhulk:
Our CI is huge and obnoxiously unreliable due to the sheer number of moving parts so I'm very reluctant to add more pieces to it.

You're right though, the `ppc64le` builders on Travis are sufficiently fast. We could re-enable that, but I'd like to know a bit more about where that's seeing use before we test against it. Additionally, we're not likely to ship wheels for that config until such time as we redesign our wheel building infrastructure, although that is planned for "when we have the ambition" 😄

daxtens:
First up, I forgot to add last time: my apologies for not checking the
history more thoroughly first, that's my bad.

Your project gets shipped by a number of distros that provide ppc64le
binaries, including RHEL and Ubuntu. That's why I was hoping to see it
tested in CI on ppc64le. But I don't want it to be a burden to you folks:
if you want to wait and refactor your processes I can check back in a
couple of years :P

reaperhulk:
Haha, no we’re not offended. You provided a well done PR with reasoning behind it. We get that far less commonly than we’d prefer. I do hope that distros are running our test suite themselves, but I’m okay with re-adding the ppc64le builder for now. We won’t be shipping wheels just yet (at the very least aarch64 has priority there), but provided we don’t run into issues with CI it shouldn’t be a big deal to run a test suite there. If you want to open a new PR adding that we can move forward.

daxtens:
Thanks, will do. Feel free to tag me in if you do run into any issues with
ppc64le: I can poke some people if Travis on Power is unreliable, and I
also have access to ppc64le systems if there's weird non-CI-related
behaviour.

PR #5362, opened 2020-07-30
handle unnamed-but-really-named curves in 1.0.2u

reaperhulk:


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1580,6 +1580,13 @@ def _ec_key_new_by_curve(self, curve):
         curve_nid = self._elliptic_curve_to_nid(curve)
         ec_cdata = self._lib.EC_KEY_new_by_curve_name(curve_nid)
         self.openssl_assert(ec_cdata != self._ffi.NULL)
+        # Setting the ASN.1 flag to 1 (aka OPENSSL_EC_NAMED_CURVE) is
+        # only necessary on OpenSSL 1.0.2t and 1.0.2u. Once we drop support
+        # for 1.0.2 we can remove this and drop the binding for set_asn1_flag.
+        # So, so irritating.
```
alex, line 1586: Please use the magic comment syntax so we'll find this with grep.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1580,6 +1580,13 @@ def _ec_key_new_by_curve(self, curve):
...
+        group = self._lib.EC_KEY_get0_group(ec_cdata)
+        self.openssl_assert(group != self._ffi.NULL)
+        self._lib.EC_GROUP_set_asn1_flag(group, 1)
```
alex, line 1589: Can we bind `OPENSSL_EC_NAMED_CURVE` constant?

reaperhulk: Look at my commit msg

alex: ok i fight :-)

PR #5365, opened 2020-07-31
test: allow skip of hypothesis tests

jwslater0823:
The module hypothesis may not be available, so we conditionally
skip two tests that need it.

Signed-off-by: Joe Slater <joe.slater@windriver.com>

### DIFF
--- a/tests/hypothesis/__init__.py
+++ b/tests/hypothesis/__init__.py
@@ -3,3 +3,6 @@
 # for complete details.
 
 from __future__ import absolute_import, division, print_function
+
+import pytest
+hypothesis = pytest.importorskip("hypothesis")


alex:
Why? Our tox configuration states that hypothesis is a requirement to run our tests.

jwslater0823:
For openembedded systems, hypothesis for python3 is not yet available,   And, hypothesis seems to be used by only two tests, so it's a shame to have them blow up during collection if it is not around.

alex:
I'm not familiar with Openembedded, but they should either install the package from pip or they should package it in their own system.

We're not going to carry extra patches for folks who want to build alternate ecosystems.

PR #5366, opened 2020-07-31
try arm64-gravitron2?

alex:


alex:
How do I tell if this thing is building for arm64 or just randomly defaulting to amd64?

alex:
nope

PR #5367, opened 2020-08-01
Add support for X509_check_issued

tomers:


docs/x509/reference.rst
```diff
@@ -499,6 +499,16 @@ X.509 Certificate Object
         :return bytes: The data that can be written to a file or sent
             over the network to be verified by clients.
 
+    .. method:: check_issued(issuer)
+
+        .. versionadded:: 3.1
+
+        This function checks if certificate subject was issued using CA certificate issuer.
```
bluetech, line 506: Should the comment above starting with "To validate the signature on a certificate ..." be updated?

src/cryptography/x509/base.py
```diff
@@ -201,6 +201,13 @@ def public_bytes(self, encoding):
         Serializes the certificate to PEM or DER format.
         """
 
+    @abc.abstractmethod
+    def check_issued(self, issuer):
+        """
+        Verifies that the certificate was potentially issued by the given
```
bluetech, line 207: Why "potentially"?

docs/exceptions.rst
```diff
@@ -37,3 +37,8 @@ Exceptions
 
     This is raised when the verify method of a key derivation function's
     computed key does not match the expected key.
+
+
+.. class:: CheckIssuedFail
```
bluetech, line 42: Just as a user of `cryptography` I get the impression that "one-off" exceptions are not defined in `cryptography.exceptions` but near their use site, so probably in `cryptography.x509.base` in this case.

docs/x509/reference.rst
```diff
@@ -499,6 +499,16 @@ X.509 Certificate Object
         :return bytes: The data that can be written to a file or sent
             over the network to be verified by clients.
 
+    .. method:: check_issued(issuer)
```
bluetech, line 502: Perhaps `check_issued_by` or `verify_issued_by`? Just "check issued" seems incomplete.

docs/development/test-vectors.rst
```diff
@@ -409,6 +409,8 @@ Custom X.509 Vectors
 * ``ca/ca.pem`` - A self-signed certificate with ``basicConstraints`` set to
   true. Its private key is ``ca/ca_key.pem``. This certificate is encoded in
   several of the PKCS12 custom vectors.
+* ``ca/issuer.pem`` - A certificate signed by ``ca/ca.pem``.
```
bluetech, line 412: - This cert is not a CA, is it supposed to be under `ca/`?
- This cert is *issued* by `ca/ca.pem`, it is not an issuer itself, therefore would it not be better to call it `issued.pem` or `issued_by.pem`?

bluetech (commented):
I'm not a `cryptography` developer, but this functionality would have been useful to me, so leaving a few comments anyway.

docs/x509/reference.rst
tomers: I agree, but eventually this function calls `self._backend._lib.X509_check_issued`, and I would like to keep the naming convention as the library that we wrap. I see this in other places too.

docs/x509/reference.rst
tomers: Removed it and added a reference to the new validation function.

docs/exceptions.rst
tomers: Agree. Fixed.

docs/development/test-vectors.rst
tomers: Agree. Fixed.

src/cryptography/x509/base.py
tomers: Removed.

reaperhulk:
This method doesn't actually check the signature, but regardless #8011 supersedes this.

PR #5372, opened 2020-08-03
add docker builders, migrate travis to focal by default

reaperhulk:
This is 50% experiment since we've had some issues with some builders having trouble with newer distros. Let's see what happens now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,6 +118,8 @@ matrix:
 
         - python: 2.7
           env: DOWNSTREAM=pyopenssl
+          # Tests fail on OpenSSLs that enforce reasonable key length minimums
+          dist: bionic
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -82,6 +81,12 @@ matrix:
         - python: 3.8
           services: docker
           env: TOXENV=py38 DOCKER=pyca/cryptography-runner-sid
+        - python: 3.6
+          services: docker
+          env: TOXENV=py38 DOCKER=pyca/cryptography-runner-ubuntu-bionic
```
alex, line 86: versions don't match

reaperhulk: And indeed this won't even work because of that.

PR #5382, opened 2020-08-07
Sleep after errors to increase the chance of success

alex:


alex:
Yeah, but every line of software everywhere is itself a tale of woe.

PR #5387, opened 2020-08-14
deprecate support for python 3.5

reaperhulk:

PR #5388, opened 2020-08-14
Update README.rst

alanyee:
#5387


alex:
3.5 is still supported, it's just deprecated. We don't change this text until support is removed.

Issue #3881, opened 2017-08-31
Making passing a backend optional


reaperhulk:
We should make backends optional since 99.99% of the time the user wants the default.

FranklinYu:
Any progress? How about beginning with making the backend parameter optional?

Angelin01:
I agree with Franklin here. I have been using the library in recent projects and it's quite good, but could use a few usability updates. Finally starting the process of deprecating backend arguments would be a step in the right direction.

As a note: it's been more than 2 years since this issue was opened with no updates aside from the reference in Franklin's closed issue. Time to take a look at it again?

reaperhulk:
Well, this is done now 😄

PR #5389, opened 2020-08-15
Make backend optional part 1 of n

reaperhulk:
Refs #3881 

This makes all the KDF backend arguments optional. This is marked as WIP because we should talk about testing. If we want to continue to support hypothetical backends (which in this case are more like LibreSSL or BoringSSL) we need a way to mark these `no_backend` tests to be skipped.

This PR does not change the examples to not include the backend argument. That should be changed at a later date once we've had this optional for a year or so.

### DIFF
--- a/src/cryptography/hazmat/backends/__init__.py
+++ b/src/cryptography/hazmat/backends/__init__.py
@@ -17,3 +17,10 @@ def default_backend():
         _default_backend = backend
 
     return _default_backend
+
+
+def _get_backend(backend):
+    if backend is None:
+        return default_backend()
+    else:
+        return backend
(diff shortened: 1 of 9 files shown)


reaperhulk:
I'm not sure how I feel about just updating the docs to show that backends are optional. Since we generate docs for latest only this means users who look up information about these functions will get incorrect/confusing information during the transition.

src/cryptography/hazmat/backends/__init__.py
```diff
@@ -17,3 +17,10 @@ def default_backend():
         _default_backend = backend
 
     return _default_backend
+
+
+def _get_backend(backend):
+    if backend is None:
+        return default_backend()
+    else:
+        return backend
```
alex, line 26: Should this warn? We intend to deprecate this eventually, right?

reaperhulk: I have some ideas for backends that would be much easier if the concept still exists, so I don’t want to deprecate it just yet.

alex: Ok.

PR #5394, opened 2020-08-15
optional backend for hash, cipher, cmac, hmac, keywrap

reaperhulk:
refs #3881


reaperhulk:
codecov doing its thing

PR #5398, opened 2020-08-15
get rid of xenial builder on arm64

reaperhulk:
death to xenial


alex:
This removes the only py2 builder on arm64. Can we add one from focal or a modern distro?

PR #5400, opened 2020-08-15
generate unrecognizedextension when failing to parse

reaperhulk:
Rather than raising an exception we now return the DER payload inside an
UnrecognizedExtension. When you compile OpenSSL without support for
certain things (like no-ocsp) it doesn't know how to parse some
extensions. Rather than error, it makes more sense to return the DER to
the caller and let them sort it out.

refs #5345

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/decode_asn1.py
+++ b/src/cryptography/hazmat/backends/openssl/decode_asn1.py
@@ -233,28 +233,29 @@ class _X509ExtensionParser(object):
 
             try:
                 handler = self.handlers[oid]
-            except KeyError:
-                # Dump the DER payload into an UnrecognizedExtension object
-                data = backend._lib.X509_EXTENSION_get_data(ext)
-                backend.openssl_assert(data != backend._ffi.NULL)
-                der = backend._ffi.buffer(data.data, data.length)[:]
-                unrecognized = x509.UnrecognizedExtension(oid, der)
-                extensions.append(x509.Extension(oid, critical, unrecognized))
-            else:
                 ext_data = backend._lib.X509V3_EXT_d2i(ext)
                 if ext_data == backend._ffi.NULL:
                     backend._consume_errors()
-                    raise ValueError(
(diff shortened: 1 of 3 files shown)


src/cryptography/hazmat/backends/openssl/decode_asn1.py
```diff
@@ -233,30 +233,31 @@ def parse(self, backend, x509_obj):
...
-                data = backend._lib.X509_EXTENSION_get_data(ext)
-                backend.openssl_assert(data != backend._ffi.NULL)
-                der = backend._ffi.buffer(data.data, data.length)[:]
-                unrecognized = x509.UnrecognizedExtension(oid, der)
-                extensions.append(x509.Extension(oid, critical, unrecognized))
-            else:
                 ext_data = backend._lib.X509V3_EXT_d2i(ext)
                 if ext_data == backend._ffi.NULL:
                     backend._consume_errors()
-                    raise ValueError(
```
alex, line 247: As far as I can tell, with the new code, we _never_ raise ValueError, which is wrong :-) If you have a malformed extension, that's an error, not something that should result in an urecgonized error.

The way we fixed this with SCTs is we don't put it in the handler dict if it's not supported. Or we can implement the parsing in rust/python and always support it.

reaperhulk: We currently make decisions based on which list of extension parsers to use based on whether or not SCT support is present. To add another permutation without altering our approach we'd need 4 extension parsers (no-sct and no-ocsp, no-sct only, no-ocsp only, and all).

If we wanted to implement our own parsing the two extensions in question are OCSPNoCheck and InvalidityDate. InvalidityDate requires generating the DER for an ASN.1 GENERALIZEDTIME though.

alex: 4 permutations would be bad. surely there's a way we can structure this to get 2 ifs, instead of the cross product though?

PR #5407, opened 2020-08-16
update submitting patches docs to be a bit more true

reaperhulk:
For now best practice is that they should be optional.

### DIFF
--- a/docs/development/submitting-patches.rst
+++ b/docs/development/submitting-patches.rst
@@ -81,10 +81,9 @@ Every recipe should include a version or algorithmic marker of some sort in its
 output in order to allow transparent upgrading of the algorithms in use, as
 the algorithms or parameters needed to achieve a given security margin evolve.
 
-APIs at the :doc:`/hazmat/primitives/index` layer should always take an
-explicit backend, APIs at the recipes layer should automatically use the
-:func:`~cryptography.hazmat.backends.default_backend`, but optionally allow
-specifying a different backend.
+APIs at the :doc:`/hazmat/primitives/index` and recipes layer should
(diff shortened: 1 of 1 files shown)


reaperhulk:
recheck

docs/development/submitting-patches.rst
```diff
@@ -81,10 +81,9 @@ Every recipe should include a version or algorithmic marker of some sort in its
 output in order to allow transparent upgrading of the algorithms in use, as
 the algorithms or parameters needed to achieve a given security margin evolve.
 
-APIs at the :doc:`/hazmat/primitives/index` layer should always take an
-explicit backend, APIs at the recipes layer should automatically use the
-:func:`~cryptography.hazmat.backends.default_backend`, but optionally allow
-specifying a different backend.
+APIs at the :doc:`/hazmat/primitives/index` and recipes layer should
+automatically use the :func:`~cryptography.hazmat.backends.default_backend`,
+but optionally allow specifying a different backend.
```
alex, line 86: this isn't true for new APIs, we don't pass a backend at all.

reaperhulk: There is no succinct way to capture our current laissez faire attitude, although many of our current hardest feature requests (e.g. engines, pkcs11, yubikey signing, etc) are the most feasible via backends. Now that they're optional I'm much more interested in retaining the idea long term and hopefully being able to actually implement something useful soon-ish. That said, I won't be modifying the existing non-backend APIs to add an optional backend just yet. 😄 

I think this guidance for third parties makes sense. Add it optional and we'll make a game-time decision in a PR.

reaperhulk:
recheck

PR #5416, opened 2020-08-18
Travis graviton2

reaperhulk:
Hypothetically we’re migrated now. What happens when we do this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,12 @@ matrix:
           env: TOXENV=py38 LIBRESSL=3.1.3
         - python: 3.8
           env: TOXENV=py38 LIBRESSL=3.2.0
-
+        - python: 3.8
+          env: TOXENV=py38
+          dist: focal
+          arch: arm64-graviton2
+          virt: lxd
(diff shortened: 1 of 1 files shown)


geoffreyblake:
Appears to be working.  I've decided to take a stab at updating pyca/infra to get the proper docker containers built now that this project can do builds with graviton2.

geoffreyblake:
@reaperhulk,  once Travis-CI releases with their GA'ed version of Graviton2 based builders, it looks like your issues with timeouts on Travis for arm64 will be solved if you want to pivot that way in the future:

https://travis-ci.com/github/geoffreyblake/cryptography/builds/181680619

Based off this branch: https://github.com/geoffreyblake/cryptography/tree/graviton2_travisci_poc

Longest build is 25 minutes.  Still slower than x86 builds, but there's a fair amount of needing to build dependencies in there compared to x86 just grabbing binaries so should come down as the wheel community matures on arm.

reaperhulk:
(This won't be happening due to Travis's billing changes)

PR #5422, opened 2020-08-24
rephrase changelog entry

alex:


reaperhulk:
Trailing whitespace and line too long linting failures. Also this should mention Fernet.encrypt/decrypt uses those methods if you're going to explicitly mention what can now do >2GB :)

PR #5423, opened 2020-08-24
add some words about separating ECDH and ECDSA

### LINKED ISSUE #4369: API allows EC keys to be used for both signing and DH
This came up in a conversation today and it merits a discussion about whether this is something we want to deprecate (since it's generally poor cryptographic hygiene to use a single EC private key for both signing and DH) or just document that we strongly recommend against it.

We do talk about how ECDH keys should be ephemeral, but does it make sense to divorce them further and try to use the type system to encourage separation?

reaperhulk:
fixes #4369


reaperhulk:
Filippo's stuff would be okay. Thomas Pornin wrote a great explanation about this long ago on crypto stackexchange as well if we want to link that: https://crypto.stackexchange.com/a/3313

alex:
That's a good cite by me!

Issue #5379, opened 2020-08-06
FFI interface to SSL_set_min_proto_version etc


richvdh:
OpenSSL 1.1.0 added a number of new functions for controlling the acceptable TLS protocol versions:

```
 int SSL_CTX_set_min_proto_version(SSL_CTX *ctx, int version);
 int SSL_CTX_set_max_proto_version(SSL_CTX *ctx, int version);
 int SSL_set_min_proto_version(SSL *ssl, int version);
 int SSL_set_max_proto_version(SSL *ssl, int version);
```

Additionally, OpenSSL 1.1.1 added getters for the same information:
```
 int SSL_CTX_get_min_proto_version(SSL_CTX *ctx);
 int SSL_CTX_get_max_proto_version(SSL_CTX *ctx);
 int SSL_get_min_proto_version(SSL *ssl);
 int SSL_get_max_proto_version(SSL *ssl);
```

See https://www.openssl.org/docs/man1.1.1/man3/SSL_set_max_proto_version.html for more details of these functions. 

The intention, as documented at https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_new.html, is that these functions should be used instead of SSL_CTX_set_options and SSL_set_options for this purpose.

Anyway: it appears that these functions are not exposed via the `cryptography.hazmat.bindings.openssl.binding.Binding.lib` FFI interface. Could they be added?

richvdh:
While I'm on the subject, would it also be possible to add support for the `TLS_method()`, `TLS_server_method()` and `TLS_client_method()` methods which were introduced in OpenSSL 1.1.0:

```
 const SSL_METHOD *TLS_method(void);
 const SSL_METHOD *TLS_server_method(void);
 const SSL_METHOD *TLS_client_method(void);
```

(see https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_new.html)

They are the same as `SSLv23_method()`, `SSLv23_server_method()`, and `SSLv23_client_method()` respectively, but those three have very confusing names (which is why they were deprecated, I believe).

th3b0x:
I would like to second the update of the pyca-cryptography interface to provide the interfaces recommended by @richvdh  .

[...]

PR #5428, opened 2020-08-25
Th3b0x tls method patch

th3b0x:
This PR addresses the **TLS_method** issues presented in Issue #5379

This patch will allow developers to use the newer TLS_method interface name, allowing them to write code that is compatible with future iterations of the OpenSSL library and its interfaces/conventions as they emerge, and will reduce false positives in static analysis tools which flag deprecated SSL/TLS modes.

Added mapping of TLS_Method to SSLv23_method (the opposite of what OpenSSL project did for backwards compatibility). The rationale is that since SSLv23_method gets mapped to TLS_method during compiler preprocessing, this will not break newer versions of the Cryptography package that user newer versions of OpenSSL (because SSLv23_method = TLS_method), while simultaneously providing the TLS_method interface via CFFI for all versions of the Cryptography package (including versions that __only__ have SSLv23_method) (in this case, TLS_method = SSLv23_method)


reaperhulk:
(This isn't passing tests so it's not really reviewable yet)

th3b0x:
> (This isn't passing tests so it's not really reviewable yet)

Hi @reaperhulk I've been inactive over the past month.  I recently had the opportunity to return to this item.  __I will tag you in a reply to this ticket when the patch is fixed.__

As an update on progress, I've had to make some minor changes to my build system.  I have a few notes regarding updates that should be made to the pyca-cryptography Development and Installation instructions as a consequence of this activity.  Is there a convention for updating the docs, or do documentation updates work exactly the same as everything else?  I ask because some projects have designated parties for approving documentation updates.

reaperhulk:
Just submit a separate PR for any docs updates please. We use the same review and merge strategy for docs and code on this project.

th3b0x:
@reaperhulk uh, not sure what's going on here, I think some changes I made for PEP8 compliance (I edited a version check conditional into multiple lines) seem to be breaking the **codecov** tests, and the **docs** are failing to build for no apparent reason (made no changes there).

Other than that, everything _appears_ to be passing.  Any thoughts?

th3b0x:
Well, now that I've gone through this exercise, I now realize that I have far too much code from the Master branch for me to be bothered with Rebasing and Squashing, so I'm going to close this PR, put the _very simple_ patch in another branch, and submit a new PR with _just the relevant information_ included.  Lazy, sloppy, and confusing; I really should've seen this coming.

PR #5435, opened 2020-08-27
Alternative: Build Arm64 wheels via github wheel-builder action with docker buildx

geoffreyblake:
I know that the devs have invested quite a bit of time and effort with @ianw getting OpenDev up and running, and thanks a ton for getting Arm wheels building!

But, just wanted to present this as an alternative for wheel building so all the binaries drop into one place for x86 (win, mac, manylinux) and arm64 (manylinux) for simplified upstreaming.  What's in this PR:

- `manylinux_aarch64` job in wheel-builder.yml that builds the arm64 python3 wheels on manylinux2014_aarch64 using docker buildx, takes ~6 minutes.
- Dockerfile_aarch64 that builds the wheels and then extracts just the wheels to copy out
- build-wheels-aarch64.sh, bash script run inside docker that builds and tests the wheels in the same way as the x86 wheels.

At the very least, this can teach others some new tricks with using Docker and buildx to work around missing native platform support on the CI system of choice where applicable.  (Obviously emulation cannot replace native testing for this project's extensive unit testing!)

### DIFF
--- a/.github/workflows/Dockerfile_aarch64
+++ b/.github/workflows/Dockerfile_aarch64
@@ -1,0 +1,16 @@
+FROM pyca/cryptography-manylinux2014_aarch64 as build
+MAINTAINER Python Cryptographic Authority
+
+ARG PYTHON
+ARG CRYPTOGRAPHY_VERSION
+ENV PYTHON=${PYTHON}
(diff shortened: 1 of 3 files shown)


alex:
For the time being I think we're going to stick to using the real arm64 hardware for this. Thanks though.

PR #5438, opened 2020-08-27
Break users on OpenSSL 1.0.2

### LINKED ISSUE #5432: Make OpenSSL 1.0.2 error (+ env var fallback)

alex:
fixes #5432


reaperhulk:
Looks like urllib3 is having some problems. Are we going to need to drop that downstream soon since we can't get the tests to run >xenial anyway?

alex:
We should probably try to engage the urllib3 maintainers... but yes.

PR #5444, opened 2020-08-29
I reviewed some changes in the code.

fochoao:
In my repositry there are code changes, needed libraries in Lib Folder, in src the source of cryptography Library, documentation done, and latests synchronization with this project, as well, Python compiled in win32, there are many changes in there, but obviously I focused on the library.


alex:
I don't know what this is supposed to be. This is not a useful patch. Please stop submitting these.

reaperhulk:
You've been temporarily banned for bizarre spam-like responses before and now you're temporarily banned for multiple completely nonsensical PRs. I'm not sure what you're trying to accomplish here but none of this does *anything* that makes sense. We already build for Windows. We already have wheels. No changes are required for this, so why are you wasting everyone's time with this garbage?

PR #5448, opened 2020-09-01
Add a missing space to py35 deprecation warning

webknjaz:
This fixes a typo that's been introduced in #5387: `s/release ofcryptography/release of cryptography/`.

### DIFF
--- a/src/cryptography/__init__.py
+++ b/src/cryptography/__init__.py
@@ -41,7 +41,7 @@ if sys.version_info[0] == 2:
     )
 if sys.version_info[:2] == (3, 5):
     warnings.warn(
-        "Python 3.5 support will be dropped in the next release of"
+        "Python 3.5 support will be dropped in the next release of "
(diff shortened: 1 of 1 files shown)


src/cryptography/__init__.py
```diff
@@ -41,7 +41,7 @@
     )
 if sys.version_info[:2] == (3, 5):
     warnings.warn(
-        "Python 3.5 support will be dropped in the next release of"
+        "Python 3.5 support will be dropped in the next release of "
         "cryptography. Please upgrade your Python.",
```
webknjaz, line 45: @reaperhulk @alex have you considered including links to issues/PRs in deprecation warnings? I saw some projects doing this and I find it quite useful.

reaperhulk: That's a good idea -- we've done it in the past once or twice but we should do it for our future support deprecations.

Issue #4967, opened 2019-08-12
PKCS11 HSM support after removal of dependent OpenSSL binding


clover-bt:
https://github.com/space88man/cryptography_engine, prior to pyca/cryptography release 2.6, provided a relatively clean Python wrapper to access and utilize HSM objects through the PKCS11 interface.

https://github.com/pyca/cryptography/pull/4768/files removed the ENGINE_load_builtin_engines() binding, which appeared to be necessary for the wrapper's operation. OpenSSL still appears to have that function in its library. (Other required bindings may also have been removed; further research is needed.)

In https://github.com/pyca/cryptography/issues/4446, it was suggested that having HSM support and upstreaming the wrapper code might be possible.

What might be the next steps to towards getting stable PKCS11-based HSM support in pyca/cryptography?

reaperhulk:
Not knowing/remembering our binding consumers strikes again.

I think we need to know the set of bindings required for this to work here. And we should also think about whether it makes sense to do the PKCS11 support through OpenSSL or if some other idea makes more sense. (e.g. should we should resurrect the backend concept in cryptography to allow a separate PKCS11 backend to be passed in?0

clover-bt:
Were used and now missing:
```
ENGINE_ctrl_cmd_string
ENGINE_load_builtin_engines
ENGINE_load_private_key
ENGINE_load_public_key

EVP_PKEY_CTX_set_rsa_mgf1_md
EVP_PKEY_CTX_set_rsa_oaep_md
EVP_PKEY_CTX_set_rsa_padding
EVP_PKEY_CTX_set_rsa_pss_saltlen

EVP_md5
EVP_sha1
EVP_ripemd160
EVP_sha224
EVP_sha256
EVP_sha384
EVP_sha512
```
Removal of EVP_hashfunction entries seems to have occurred via https://github.com/pyca/cryptography/pull/4775/files#diff-04d869a327dc59acfd9b090d3a166467L100, also February 2019.

Used and still present:
```
_lib.ENGINE_by_id
_lib.ENGINE_finish
_lib.ENGINE_free
_lib.ENGINE_init

_lib.EVP_PKEY_CTX_free
_lib.EVP_PKEY_CTX_new
_lib.EVP_PKEY_CTX_set_signature_md
_lib.EVP_PKEY_decrypt
_lib.EVP_PKEY_decrypt_init
_lib.EVP_PKEY_encrypt
_lib.EVP_PKEY_encrypt_init
[...]

Issue #5186, opened 2020-04-06
Supporting OpenSSL Engines


torin-carey:
The OpenSSL library provides a rich and extensible interface for dealing with cryptographic tokens and dedicated accelerators through the engine API.

The vast majority of the cryptography library should already be able to transparently deal with engine supplied `EVP_PKEY` instances and OpenSSL already has functions for setting the default engines for various operations, therefore the implementation could be minimal.

The main difficulty would be deciding how an engine interface would look—a separate backend wouldn't make much sense and so there would be a decision between making a backend independent engine interface or a openssl-specific interface.

At the very least I think it would be useful to add some of the `ENGINE_` functions to the [CFFI bindings](https://github.com/pyca/cryptography/blob/master/src/_cffi_src/openssl/engine.py) so that modules could piggy-back off of cryptography's cffi bindings to provide the functionality.

alex:
What's your actual goal here? It sounds like you want to do asymmetric private keys ops on keys in hardware devices?

torin-carey:
That's what I'm mainly interested in. It's way outside the scope of cryptography to manage PKCS11 hardware tokens and there's ample implementation for openssl engines interfacing different hardware tokens. It would shift a lot of the issues of  "Can using X tokens be supported" over to openssl.

alex:
So, I think the actual API we want to figure out here is something like `load_pkcs11_key(...) -> returns an EC/RSA/DSA/x25519/etc. key` -- ENGINE can be the implementation, but the public API should be about the keys, not about ENGINE.

torin-carey:
I think even mention of pkcs11 is more low-level than needed since OpenSC provide an engine for pkcs11 interfacing.

We have the `ENGINE_load_private_key` function so at the very least, we could load a key by specifying the `engine_id` of the engine and the engine specific `key_id`. At that point the user can configure openssl outside of cryptography to load specific engines.

h-vetinari:
Note that a lot of this will change with openssl 3.0 which should land this year (the first alpha should be in a few weeks) - basically all things engine are getting (more or less [explicitly](https://github.com/openssl/openssl/pull/11082#issuecomment-585561943)) deprecated, and replaced with "providers".

I mean, it would be great to have an API for both, but if that's unrealistic, then perhaps focusing on providers is the more sustainable path.

paulidale:
[...]

PR #5449, opened 2020-09-01
re-add a few engine bindings for advanced users

reaperhulk:
For users who are capable of compiling cryptography against custom openssl and properly using these functions this hopefully allows PKCS11 usage through OpenSSL engines.

We've seen quite a few requests for this to be re-added over the past year, OpenSSL 3.0 is not coming out quickly (and providers won't replace engines immediately), and since we now support `no-engine` the maintenance burden here should be manageable.

Note that we are not directly supporting the use of this, merely providing the escape hatch for advanced users. If there are additional bindings that are missing please let us know, although we won't add functions that have equivalents already (e.g. `EVP_sha256() is not needed because you can get the `EVP_MD *` via a call to `EVP_get_digestbyname`).

### DIFF
--- a/src/_cffi_src/openssl/engine.py
+++ b/src/_cffi_src/openssl/engine.py
@@ -10,6 +10,7 @@ INCLUDES = """
 
 TYPES = """
 typedef ... ENGINE;
+typedef ... UI_METHOD;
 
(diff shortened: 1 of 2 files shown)


reaperhulk:
Fixes #4967 and #5186

PR #5451, opened 2020-09-02
Added the option for extending x905 parser with your own OIDs and handlers

dvaerum:
**Background for PR**  
I am using cryptography to handle Request Certificate from Puppet Agent and is in the process of creating a program for handle policy-based autosigning, but it is a little annoying that I cannot add (please tell me if I am wrong) my/our own OID and add handlers for them without getting an `UnrecognizedExtension`.

**What is the PR adding**  
This PR will add the option to extend  `_X509ExtensionParser` with your own OIDs and associated handlers, there the handlers are functions for processing the `ext_data`.
I have also added an Unittest to make sure that it keeps working 😉 

**Example**  
```python
from cryptography.x509 import ObjectIdentifier
from cryptography.x509.oid import _OID_NAMES

PP_ENVIRONMENT_OID = ObjectIdentifier("1.3.6.1.4.1.34380.1.1.12")
_OID_NAMES[PP_ENVIRONMENT_OID] = "Puppet node environment name"

def _pp_handler(backend, ext_data):
            return ext_data.decode('utf-8')

backend._certificate_extension_parser.add_extra_oid_handler(PP_ENVIRONMENT_OID, _pp_handler)

pem_req_data = b"""-----BEGIN CERTIFICATE REQUEST-----
PEM DATA ...
-----END CERTIFICATE REQUEST-----"""

csr = x509.load_pem_x509_csr(pem_req_data, default_backend())
pp_env = csr.extensions.get_extension_for_oid(PP_ENVIRONMENT_OID)
```

**FEEDBACK**  
If there is anything there need changing for this to be accepted, I will try to do it 😃

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/decode_asn1.py
+++ b/src/cryptography/hazmat/backends/openssl/decode_asn1.py
@@ -186,6 +186,17 @@ class _X509ExtensionParser(object):
         self.get_ext = get_ext
         self.handlers = handlers
         self._backend = backend
+        self.handlers_extra = {}
+
+    def add_extra_oid_handler(self, oid, func):
+        if not isinstance(oid, x509.ObjectIdentifier):
+            raise TypeError('The variable "oid" is not an ObjectIdentifier')
+        if not callable(func):
+            raise TypeError('The variable "func" have to be a callable object')
+        if func.__code__.co_argcount != 2:
+            raise KeyError('The callable object "func" have to take 2 arguments')
+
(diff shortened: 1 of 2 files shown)


alex:
The right way to handle this use case is to process the `UnsupportedExtension` that you get and parse the data they contain.  We don't want to add a register API (and as you've probably noticed, this API still requires going through a non-public API), see #2903 for a previous version of this thread.

PR #5452, opened 2020-09-04
Fixed typo in OpenSSL version

dwagin:
OpenSSL 1.1.1b version is 0x10101002


reaperhulk:
This is not correct. OpenSSL reserves the final nibble in its version number for development/betas and always uses f for the release. So the released version number of 1.1.1b is `0x1010102f`. We set it to `0x10101020` to cover dev versions as well.

dwagin:
https://github.com/openssl/openssl/blob/50eaac9f3337667259de725451f201e784599687/include/openssl/opensslv.h#L42

`# define OPENSSL_VERSION_NUMBER  0x1010102fL`

dwagin:
My system have `# define OPENSSL_VERSION_NUMBER  0x1010107fL` and `CRYPTOGRAPHY_OPENSSL_LESS_THAN_111B` not work

reaperhulk:
What do you mean when you say it doesn't work? That macro will return false for 1.1.1b or newer, which is what we expect. We have multiple jobs in our CI that also test against this, which is why your PR failed when you tried to make an incorrect change.

dwagin:
I'm sorry, I checked everything again, it worked. The problem was with Ed25519 support.

PR #5457, opened 2020-09-11
Make the OpenSSL signature recovery function available from Python

misterzed88:
There is a sometimes a need to recover the RSA signature data in order to parse its format. This is used with non-standard signature formats. Since this is an exotic case, the support was added only at the low level through the OpenSSL lib wrapper, without high level API support in hazmat.

### DIFF
--- a/src/_cffi_src/openssl/evp.py
+++ b/src/_cffi_src/openssl/evp.py
@@ -101,6 +101,9 @@ int EVP_PKEY_sign(EVP_PKEY_CTX *, unsigned char *, size_t *,
 int EVP_PKEY_verify_init(EVP_PKEY_CTX *);
 int EVP_PKEY_verify(EVP_PKEY_CTX *, const unsigned char *, size_t,
                     const unsigned char *, size_t);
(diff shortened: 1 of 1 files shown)


reaperhulk:
@misterzed88 When we don't have direct consumers in our bindings it becomes very difficult for us to maintain them effectively. If we're going to merge this without a hazmat layer API could you write a (pretty big) comment that explains the use case more thoroughly? You can put it directly above the bindings. Also, linking this PR in the comment would be good. Blame works, but only until we end up reformatting something in the future, so a URL that can point back here is also helpful.

Alternately, what does a hazmat API for this look like and how would someone document it to explain the (rare) use of it?

misterzed88:
You're right, exposing internal interfaces for external use is usually fragile. Thank you for even considering it.

Let's try to do it in the proper way. I will come back with an explanation of the use case and a proposed hazmat API for it. Then we'll see if it fits into the current API philosophy.

misterzed88:
I opened a separate issue (#5495) for this requirement to solve the problem in a more generic way than this pull request.

PR #5459, opened 2020-09-12
Clarify 'unsafe' argument for NIST curves

byrro:
The project guidelines state that "when there is real disagreement
between cryptographic experts, represent both sides of the argument
and describe the trade-offs clearly". Currently, the reference to
'SafeCurves' leaves off the other side of the argument and could
mislead uninformed readers. This edit clarifies the other side
and provides links to additional sources.

### DIFF
--- a/docs/hazmat/primitives/asymmetric/ec.rst
+++ b/docs/hazmat/primitives/asymmetric/ec.rst
@@ -369,7 +369,13 @@ binary field ("B") NIST curves may have been weakened during their generation.
 
 Currently `cryptography` only supports NIST curves, none of which are
 considered "safe" by the `SafeCurves`_ project run by Daniel J. Bernstein and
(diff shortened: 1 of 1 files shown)


reaperhulk:
I'm in favor of updating this, but I think maybe just the https://satoshinichi.gitlab.io/b/safecurves-scare.html and Pornin links are sufficient.

We should also update our text to note that we do support non-NIST curves (e.g. X25519/X448, ed25519/ed448 as well as brainpool since German smart meters turn out to require that). That's been out of date for quite some time now...

(Test failures are various unrelated CI issues that will hopefully clear up)

reaperhulk:
Just a friendly ping to see if there's still interest in continuing this PR.

reaperhulk:
Closing for now, but feel free to reopen if you get a chance to update it.

PR #5469, opened 2020-09-20
Remove bindings docs

alex:


alex:
For future us: this is in recognition that we make changes to these all the time to reflect changing OpenSSL versions, maintenance burdens, etc. and there's no meaningful way for us to keep these stable.

reaperhulk (approved):
Making explicit what has been implicitly true for years (and was very obvious with 3.1 when we removed quite a few bindings).

PR #5479, opened 2020-10-02
add seed

wellingtonf-souza:
In some situations, it is interesting that the same messages are encrypted equally. In my case, I needed to encrypt personal identifiers and these would be used to merge with other databases.


alex:
Sorry, no. Random IVs are a part of the specification of fernet and allowing users to set them makes it easy for them to break the cryptographyic guarantees. We will not be making this change.

PR #5483, opened 2020-10-06
Added TLS_method bindings for OpenSSL and LibreSSL

th3b0x:
Added bindings for TLS_method TLS_client_method and TLS_server_method with support for LibreSSL and OpenSSL
Added tests for TLS_method TLS_client_method and TLS_server_method with support for LibreSSL and OpenSSL


th3b0x:
Got distracted and had to make a new PR, because I borked my branch too much for a clean Rebase.  

Current PR in relation to the `TLS_method*` interfaces is now #5483

I will then be able to proceed to adding the new SSL_CTX option bindings, as requested by @richvdh 

Kindly requesting a review @reaperhulk  (this successfully passed all checks in https://github.com/pyca/cryptography/pull/5428 however, the commit log became very very mangled due to some careless mistakes I had made, so I have opted to sidestep Rebasing and simply create a new Branch and PR)

tests/hazmat/bindings/test_openssl.py
```diff
@@ -85,6 +85,61 @@ def test_ssl_mode(self):
         assert resp == expected_options
         assert b.lib.SSL_get_mode(ssl) == expected_options
 
+    @pytest.mark.skipif(
+        Binding.lib.Cryptography_HAS_TLS_METHOD == 0,
+        reason="TLS_method requires OpenSSL >= 1.1.0",
+    )
+    def test_tls_ctx_options(self):
```
reaperhulk, line 92: Duplicating these tests just to pass TLS_method isn't necessary.

th3b0x: Noted in the commit update (currently building).  Thanks for catching that.

src/_cffi_src/openssl/ssl.py
```diff
@@ -755,4 +762,13 @@
 #else
 static const long Cryptography_HAS_TLSv1_3 = 1;
 #endif
+
+#if CRYPTOGRAPHY_OPENSSL_LESS_THAN_110 && !CRYPTOGRAPHY_IS_LIBRESSL
+static const long Cryptography_HAS_TLS_METHOD = 0;
+const SSL_METHOD* (*TLS_method)(void) = NULL;
```
reaperhulk, line 768: Why isn't this just `#define TLS_method SSLv23_method` and then TLS_method is available in all cases?

th3b0x: Lack of attention on my part, I failed to pull this in from the prior PR.  Excellent catch and suggestion, I've committed and pushed.  Thank you, very much.

src/_cffi_src/openssl/ssl.py
```diff
@@ -755,4 +762,13 @@
 #else
 static const long Cryptography_HAS_TLSv1_3 = 1;
 #endif
+
+#if CRYPTOGRAPHY_OPENSSL_LESS_THAN_110 && !CRYPTOGRAPHY_IS_LIBRESSL
+static const long Cryptography_HAS_TLS_METHOD = 0;
```
reaperhulk, line 767: 1) When defining `TLS_method` to be `SSLv23_method` you no longer need to conditionally remove the function. It will always be available. So the else block, the `HAS_TLS_METHOD` variable, and the `cryptography_has_tls_method` func in `_conditional.py` are not required.
2) I'd prefer this to be done using #define rather than assigning func ptrs

src/_cffi_src/openssl/ssl.py
```diff
@@ -501,6 +506,7 @@
 """
 
 CUSTOMIZATIONS = """
+
```
reaperhulk, line 509: Extraneous newline. Otherwise this LGTM now.

src/_cffi_src/openssl/ssl.py
th3b0x: @reaperhulk Thank you for the very thorough and useful review (lots of great stuff that would've slipped through the cracks due to my inexperience with CFFI and lack of attention).  I'm reviewing one last issue here related to how this functionally modifies program behavior.  I suspect I'll have my final item ready for submission today or tomorrow.  Hold tight.

th3b0x: @reaperhulk 

Initially, I was in agreement on this, but now I am thinking that it might be unwise to implement the `TLS_method` interfaces in this manner, because it would allow a developer to invoke `TLS_method` on OpenSSL implementations that might not support **any version** of TLS.

Additionally, I am beginning to think that I might need to actually implement *more tests* in this PR before it's ready due to the kinds of problems described in this Code Review for Google's BoringSSL: https://boringssl-review.googlesource.com/c/boringssl/+/8513/   
I happen to know that the problems described by @davidben do happen to persist in OpenSSL 1.1.1g (because I performed an inspection earlier this year to identify potential avenues of exploitation related to the problems described in his BoringSSL commit comment).

As a specific test, I think it might be important to add a test related to whether `SSL_OP_NO*` causes "surprising" (dangerous) behavior as @davidben describes.  I am personally wondering if we should specifically retain conditional removal, but replace `cryptography_has_tls_method` with a flag similar to `CRYPTOGRAPHY_OPENSSL_LESS_THAN_110` that would indicate whether or not the library in use is a version of OpenSSL that retains the stupidly dangerous behavior of forcing the use of downgraded SSL/TLS methods when a user provides an `SSL_OP` bitmask that creates discontinuities in the SSL/TLS version ranges supported.  (I would provide code to demonstrate what I'm talking about, but I think @davidben  or yourself might already know what I mean).

th3b0x: Also, gonna tag @alex because I'm curious what someone in your location thinks about the problem described above.

I'm really trying to source ideas for tests, actually.

I am also thinking that I should revert [
a5f30c8](https://github.com/pyca/cryptography/pull/5483/commits/a5f30c8b88f95cb6694c58cdce583b975d2933ab).  The test suite seems a little thin around verifying functionality for more complex behaviors like the ones provided by `TLS_method` and `SSLv23_method`.  The missing parts of the notebook had some socket programming, but I think we can actually just directly handle this problem by using the **conditional loading** mechanism originally suggested, but I could really use a suggestion around what to rename the conditional load flag variable 

perhaps (as an example):
```
#if !CRYPTOGRAPHY_IS_LIBRESSL
static const long Cryptography_HAS_STUPID_SSL_OP_BEHAVIOR = 1;
```
Since this definitely applies to all of the OpenSSL versions I'm seeing as supported

th3b0x:
Converted to draft, since I don't want someone to accidentally approve a PR that might cause a security issue down the road.

src/_cffi_src/openssl/ssl.py
alex: Err, what is my location?

th3b0x: > Err, what is my location?

Showing as Washington DC (obviously, user accounts don't necessarily have accurate information).

alex: It's true I live in Washington, DC but I don't see the relevance?

reaperhulk: `TLS_method` is intended as a drop-in replacement for `SSLv23_method` and we should treat it as such...

cryptography's bindings layer cannot enforce sanity in OpenSSL's APIs. That way lies madness. We can only expose their API and attempt use it safely/sanely within our own code. We don't even consider our bindings public surface area any more (pyOpenSSL is a special case that we ensure does not break).

I'm opposed to trying to test for behavioral quirks (even dangerous ones) since it almost definitionally can't be useful to this project. That effort is much better expended upstream in conversations with OpenSSL about how to improve/replace dangerous APIs.

davidben: Yeah there's no difference in behavior between SSLv23_method and TLS_method. The version disabling flags do indeed behave a little oddly but I don't think it's related to this. The less surprising API is SSL_set_min/max_proto_version if you all want to switch to those, but they only exist starting 1.1.0.

th3b0x: Well, you're a major contributor to multiple important Python libraries for Cryptography and security.  And people in Washington DC are generally interested in that kind of thing (though, it would be presumptuous to think that **Paramiko**'s 2nd-most-frequent contributor this semester, and multi-library-contributor in general, might know people with inputs on this).  I suppose your library contributions are more relevant than your location, but people do seem quite politically interested in privacy as always (and DC gets political; I get that it can be exhausting).

**Jokes begin here**
Basically, I have to ask: do none of the other major contributors see this backporting behavior as problematic?  I would be allowing Pypi users (basically anyone using `pip` and trusting `DNS`) to invoke a `TLS_method` call in `OpenSSL 00.1.1hb` (it's the beta version of an OpenSSL so old that they still had romantic notions of releasing free cryptosystems in major revisions beyond the single-digits).  I mean, I feel like people forgot that OpenSSL 2.x.x, is a mature library actively being replaced (and this library kinda supports it, but not with a proper `TLS_method` interface) (though, once again, I'll point out that I'm just trusting `DNS` here, since the internet is regionally dictated by your local Network Operators).  For example: everyone knows that Dyna-Soars went extinct due to a refusal to upgrade cryptography to beyond OpenSSL 1.1.1;  Elon Musk, as a naturalized alien, doesn't even know about how Space Force 1 got taken out due to the use of an improperly implemented 3-DES (the keys were triple-wrapped instead of triple-encrypted; the distinctions are available in the lost-or-redacted "f*****g manuals" repeatedly referenced in the NTSB reports, which "got blacked" - a phrase which apparently meant "comprehensively redacted" back then; apparently "2-DES" would decrypt things, so you were never supposed to use that "logarithm").  Nor is it well known that Space Force 2 triggered it's Global ThermoNuclear War Live Fire Games programming due to a bunch of Phreaks that discovered the existence of the ("still-classified-as Top Secret Super Classified Information") `OpenSSL 2.x.x` libraries (tangentially, this incident later led to a great series of world wars, including one over whether we lived in Cyrodiil or Middle-Earth ¯\_(ツ)_/¯ ).  The real tragedy, of course, is that Space Force 4 might be using `OpenSSL 1.x.x` because they got pwned by open source software systems that many people use and trust (but typically don't inspect)

**End of Jokes**

@reaperhulk I want to know: do you want to enable older versions of OpenSSL because you have a secret double-digit version of OpenSSL, or are you one of those weird Austin types that loves using the prototype libraries because that's what Hipsters do? (You know, using crypto from the era before people knew how to read or count).

**Back to serious business**
I am conflicted between the nature of backporting `TLS_method` Python interfaces.  @alex , basically, what are the potential Politics of enabling PHP 5 to easily function on the modern internet, given the fact that Python provides the simplest mechanism for interfacing PHP to a modern OpenSSL implementation?  I'm inclined to actually withdraw this Pull Request upon consideration of what might crawl out of an era that makes Nazi 4chan look laughably innocent.

th3b0x: @reaperhulk  I'm satisfied with these answers.  I'll implement the final suggestions after examining why **paramiko** fails on Garbage Collection for this build in Travis CI, unless @alex beats me to it.

The real goal here is actually to get to the APIs suggested by @davidben 

I am mildly concerned about how well the modern culture would handle the re-emergence of content from the era when PHP5 was still a relevant language.  That seems like something worth discussing amongst people that remember what the internet was like back then.

th3b0x: @alex I couldn't replicate the **paramiko** build failure.

with respect to why I haven't fixed the extraneous new line yet see comment below (note, I'm going to let this one sit for a while in case anyone is maintaining versions of OpenSSL that don't implement the normal behavior):

I'm really trying to source ideas for tests, actually.

I am also thinking that I should revert [
a5f30c8](https://github.com/pyca/cryptography/pull/5483/commits/a5f30c8b88f95cb6694c58cdce583b975d2933ab).  The test suite seems a little thin around verifying functionality for more complex behaviors like the ones provided by `TLS_method` and `SSLv23_method`.  The missing parts of the notebook had some socket programming, but I think we can actually just directly handle this problem by using the **conditional loading** mechanism originally suggested, but I could really use a suggestion around what to rename the conditional load flag variable 

perhaps (as an example):
```
#if !CRYPTOGRAPHY_IS_LIBRESSL
static const long Cryptography_HAS_DANGEROUS_SSL_OP_BEHAVIOR = 1;
```
Since this definitely applies to all of the OpenSSL versions I'm seeing as supported

th3b0x: @alex I couldn't replicate the **paramiko** build failure.

note, I'm going to let this one sit for a while in case anyone is maintaining versions of OpenSSL that don't implement the normal behavior.

Meanwhile, I might try to push those interfaces from issue #5379

th3b0x:
@reaperhulk keep it pending as long as you want (assuming it doesn't fail on Travis CI again)

th3b0x:
@reaperhulk I'm going to move the rest of the fixes for #5379 into a separate PR.

Also, the Codecov check is failing on code unrelated to this PR, so I'm not sure what I ought to be doing with that.

neheb:
This looks like a better version of the one I have here: https://github.com/pyca/cryptography/pull/4920

PR #5498, opened 2020-10-25
allow additional certificates to be added to a pkcs7

### LINKED ISSUE #5492: Add additional x509 certificate to SMIMESignatureBuilder
As mentioned in #5433: It would be great to be able to add an additional certificate when using the new SMIMESignatureBuilder API. 

The background is that my x509 certificate is from a commercial CA. But - as it's commonly the case - it is not signed directly by the Root CA (instead it's signed from an intermediate/sub CA). So when I send a message to a new contact I would definitely want to include my certificate AND the certificate of the intermediate CA.

reaperhulk:
fixes #5492

### DIFF
--- a/docs/hazmat/primitives/asymmetric/serialization.rst
+++ b/docs/hazmat/primitives/asymmetric/serialization.rst
@@ -647,6 +647,13 @@ contain certificates, CRLs, and much more. PKCS7 files commonly have a ``p7b``,
             :class:`~cryptography.hazmat.primitives.hashes.SHA384`, or
             :class:`~cryptography.hazmat.primitives.hashes.SHA512`.
 
+    .. method:: add_certificate(certificate)
+
+        Add an additional certificate to the PKCS7 structure. This method may
+        be called multiple times to add as many certificates as desired.
+
+        :param certificate: The :class:`~cryptography.x509.Certificate` to add.
+
     .. method:: sign(encoding, options, backend=None)
 
         :param encoding: :attr:`~cryptography.hazmat.primitives.serialization.Encoding.PEM`,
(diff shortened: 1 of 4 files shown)


docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -647,6 +647,13 @@ contain certificates, CRLs, and much more. PKCS7 files commonly have a ``p7b``,
             :class:`~cryptography.hazmat.primitives.hashes.SHA384`, or
             :class:`~cryptography.hazmat.primitives.hashes.SHA512`.
 
+    .. method:: add_certificate(certificate)
+
+        Add an additional certificate to the PKCS7 structure. This method may
```
alex, line 652: Docs ought to describe what the certs are for :-)

Also, does it just put all of these in a ball? They don't need to be assosciated with a particular signer?

reaperhulk: If PKCS7 allows association it's not in the OpenSSL API. Big ball best ball. Let the verifier sort 'em out I always say.

alex (approved):

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -63,6 +64,14 @@ def add_signer(self, certificate, private_key, hash_algorithm):
             self._signers + [(certificate, private_key, hash_algorithm)],
         )
 
+    def add_certificate(self, certificate):
+        if not isinstance(certificate, x509.Certificate):
+            raise TypeError("certificate must be a x509.Certificate")
```
alex, line 69: untested path

alex (changes requested):
coverage

src/cryptography/hazmat/primitives/serialization/pkcs7.py
reaperhulk: fixed

alex (approved):

PR #5505, opened 2020-10-25
be more verbose in the 102 deprecation notice

reaperhulk:


src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -181,8 +181,11 @@ def _verify_openssl_version(lib):
         else:
             raise RuntimeError(
                 "You are linking against OpenSSL 1.0.2, which is no longer "
-                "supported by the OpenSSL project. You need to upgrade to a "
-                "newer version of OpenSSL."
+                "supported by the OpenSSL project. To use this version of "
+                "cryptography you need to upgrade to a newer version of "
+                "OpenSSL. For this version only you can also set the "
+                "environment variable CRYPTOGRAPHY_ALLOW_OPENSSL_102 to "
+                "allow OpenSSL 1.0.2."
```
alex, line 188: We have not included this in previous OpenSSL version droppings, and instead referred people to the docs. Why now?

reaperhulk: Because I've fielded at least one question on StackOverflow where it turns out they broke their pip (due to the pyOpenSSL + cryptography on older pip/Python). They couldn't find the docs because we only keep them around for one version and I'd rather make it obvious how to fix it directly in the error.

alex (approved):

PR #5511, opened 2020-10-26
Delete all the 1.0.2 code

alex:


src/_cffi_src/build_openssl.py
```diff
@@ -56,7 +54,7 @@ def _extra_compile_args(platform):
     function signatures in OpenSSL have changed from long -> unsigned long
     in the past. Since that isn't a precision issue we don't care.
     When we drop support for CRYPTOGRAPHY_OPENSSL_LESS_THAN_110 we can
-    revisit this.
+    revisit this. XXXX: Do we revisit now?
```
reaperhulk, line 57: Yes let’s file an issue to follow up on this and see if we can just pass this in CFLAGS/fix some definitions and remove this hack.

src/_cffi_src/openssl/x509.py
```diff
@@ -298,47 +298,5 @@
...
-{
-    return x->revocationDate;
-}
-const ASN1_INTEGER *X509_REVOKED_get0_serialNumber(const X509_REVOKED *x)
-{
-    return x->serialNumber;
-}
-
-#define X509_set1_notBefore X509_set_notBefore
-#define X509_set1_notAfter X509_set_notAfter
```
reaperhulk, line 339: Arguably these should stick around since only newer pyopenssl uses the new names. But they can upgrade pyopenssl so ¯\_(ツ)_/¯

reaperhulk (approved):
Hello 3.0.0 branch conflicts

PR #5517, opened 2020-10-26
fixes #5513 update comment on why we can't have nice things

### LINKED ISSUE #5513: Revisit _extra_compile_args in build_openssl.py
Now that we're dropping 1.0.2 in #5511, it's time to review that.

alex:


src/_cffi_src/build_openssl.py
```diff
@@ -117,13 +116,6 @@ def _extra_compile_args(platform):
         "callbacks",
     ],
     libraries=_get_openssl_libraries(sys.platform),
-    # These args are passed here so that we only do Wconversion checks on the
```
reaperhulk, line 120: Why remove this? It's still true that moving these args into CFLAGS would result in failures due to cffi generated code right?

alex: Because the comment is duplicated, and I didn't want to have to update it 2x every time.

reaperhulk (approved):

PR #5536, opened 2020-10-31
we have abi3 wheels for windows now

alex:
technically we still do have other wheels for windows

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -118,13 +118,13 @@ upstream, ``cryptography`` is also dropping support for them. To fix this issue
 you should upgrade to a newer version of OpenSSL (1.1.0 or later). This may
 require you to upgrade to a newer operating system.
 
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -118,13 +118,13 @@ upstream, ``cryptography`` is also dropping support for them. To fix this issue
 you should upgrade to a newer version of OpenSSL (1.1.0 or later). This may
 require you to upgrade to a newer operating system.
 
-Why are there no wheels for Python 3.6+ on Linux or macOS?
-----------------------------------------------------------
+Why are there no wheels for Python 3.6+?
```
reaperhulk, line 121: This should probably just be changes to say “why are no wheels for my python 3.x release” or something to make it more generic. We do have 36 now since that’s our base abi3.

alex: good call

reaperhulk (approved):

PR #5537, opened 2020-10-31
abi3 only on windows

alex:


.github/workflows/wheel-builder.yml
```diff
@@ -115,9 +115,6 @@ jobs:
           - {ARCH: 'x64', WINDOWS: 'win64'}
         PYTHON:
           - {VERSION: "2.7", MSVC_VERSION: "2010"}
-          - {VERSION: "3.6", MSVC_VERSION: "2019"}
-          - {VERSION: "3.7", MSVC_VERSION: "2019"}
-          - {VERSION: "3.8", MSVC_VERSION: "2019"}
           - {VERSION: "3.8", MSVC_VERSION: "2019", "USE_ABI3": "true", "ABI_VERSION": "cp36"}
```
reaperhulk, line 118: the USE_ABI3 flag can also be removed (and associated non-abi3 step below)

alex: py2still

reaperhulk: Could make that `if matrix.PYTHON.VERSION == '2.7'` but you're right, let's leave it as is until we drop 2.7 in Jan.

reaperhulk (approved):

PR #5543, opened 2020-11-01
Don't tell people to use PGP, it's not good

alex:


reaperhulk:
image

reaperhulk (approved):

martin-vi:
WTF! What about signatures and E2EE?

> You should verify that your MTA uses TLS to ensure the confidentiality of your message. 

*HAHA* good joke.

PR #5548, opened 2020-11-02
padding: Tip-toe around bytes subclasses.

benjaminp:
This change allows future's newbytes class to be padded again.

Fixes https://github.com/pyca/cryptography/issues/5547.

### DIFF
--- a/src/cryptography/hazmat/primitives/padding.py
+++ b/src/cryptography/hazmat/primitives/padding.py
@@ -42,7 +42,10 @@ def _byte_padding_update(buffer_, data, block_size):
 
     utils._check_byteslike("data", data)
 
-    buffer_ += bytes(data)
+    # six.PY2: Only coerce non-bytes objects to avoid triggering bad behavior
+    # of future's newbytes type. Unconditionally call bytes() after Python 2
+    # support is gone.
+    buffer_ += data if isinstance(data, bytes) else bytes(data)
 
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/primitives/padding.py
```diff
@@ -42,7 +42,9 @@ def _byte_padding_update(buffer_, data, block_size):
 
     utils._check_byteslike("data", data)
 
-    buffer_ += bytes(data)
+    # Only coerce non-bytes objects to avoid triggering bad behavior of future's
+    # newbytes type. Unconditionally call bytes() after Python 2 support is gone.
```
alex, line 46: Can you put words `six.PY2` in this comment so we'll find it with grep when teh time comes.

alex:
pep8/black errors.

reaperhulk (approved):
I hate this, but understand why we need it for now. Thanks for doing the work. How urgently do you need a release here?

benjaminp:
> I hate this, but understand why we need it for now. Thanks for doing the work.

Thank you very much for putting up with it.

> How urgently do you need a release here?

The regression is preventing me from going past 3.2, but that's not urgent for me at the moment. Just doing my regular update cadence for libraries. Before you drop Python 2 support would be good.

reaperhulk:
We will definitely do a 3.3 release with this. That release is likely to be our final 2.7 release.

PR #5552, opened 2020-11-05
Fix broken links

felixfontein:
The FAQ links result in a 404, `faq/` needs to be replaced with `faq.html` to fix that.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -1368,8 +1368,9 @@ class Backend(object):
         if x509 == self._ffi.NULL:
             self._consume_errors()
             raise ValueError(
-                "Unable to load certificate. See https://cryptography.io/en/la"
-                "test/faq.html#why-can-t-i-import-my-pem-file for more details."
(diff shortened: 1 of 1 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1369,7 +1369,7 @@ def load_pem_x509_certificate(self, data):
             self._consume_errors()
             raise ValueError(
                 "Unable to load certificate. See https://cryptography.io/en/la"
-                "test/faq/#why-can-t-i-import-my-pem-file for more details."
+                "test/faq.html#why-can-t-i-import-my-pem-file for more details."
```
reaperhulk, line 1372: All these lines exceed the char limit now so flake8 fails. 😢

felixfontein: Reminds me that I opened the failing travis result in a tab to look at what exactly was failing... Will fix in a minute!

felixfontein:
Failed CI seems to be unrelated this time.

reaperhulk (approved):

felixfontein:
@reaperhulk thanks!

PR #5561, opened 2020-11-11
Always rely on OpenSSL's builtin locking callbacks

alex:


reaperhulk:
`LICENSE` should probably be updated to note that the only code left is for the osrandom engine.

alex:
Good call

reaperhulk (approved):

PR #5563, opened 2020-11-11
port docs and docs-linkcheck

reaperhulk:


.github/workflows/ci.yml
```diff
@@ -230,3 +231,18 @@ jobs:
       - run: ./.github/downstream.d/${{ matrix.DOWNSTREAM }}.sh install
       - run: pip install .
       - run: ./.github/downstream.d/${{ matrix.DOWNSTREAM }}.sh run
+
+  docs-linkcheck:
+    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```
alex, line 236: Is there a way to check commit message, like we do on the other build?

.github/workflows/ci.yml
```diff
@@ -18,6 +18,7 @@ jobs:
           - {VERSION: "3.9", TOXENV: "pep8,packaging", COVERAGE: "false"}
           - {VERSION: "pypy2", TOXENV: "pypy-nocoverage", COVERAGE: "false"}
           - {VERSION: "pypy3", TOXENV: "pypy3-nocoverage", COVERAGE: "false"}
+          - {VERSION: "3.9", TOXENV: "docs", COVERAGE: "false"}
```
alex, line 21: Can we just run this in the ppe8/packaging build?

reaperhulk: We could. Do we want to have one job overloaded with three tasks? Is there a disadvantage to running docs separately?

alex: fewer builds, since there's a small overhead per build, and all 3 of these tox jobs are fast, make sense to bundle them

alex (approved):

PR #5585, opened 2020-11-25
Attempt to de-dupe logic from wheel builders

alex:


alex:
https://github.com/alex/cryptography/runs/1451375464?check_suite_focus=true

currently broken because `.venv/bin/pip` doesn't exist on Windows. Why is Windows different...

reaperhulk:
Windows requires .venv\Scripts\pip :trollface:

alex:
I'm giving up on this for now.

PR #5587, opened 2020-11-26
in OpenSSL 1.1.0+ error strings are automatically loaded

reaperhulk:


alex:
Can you point me at a doc or changelog reference on this? Deleting both the code and the test at the same time is a scary combo :-)

reaperhulk:
https://www.openssl.org/docs/man1.1.0/man3/SSL_load_error_strings.html

alex (approved):

PR #5591, opened 2020-11-27
remove a function that just invokes another function now

reaperhulk:


alex (changes requested):
pyOpenSSL calls this, if you want to do this you need to rename in the other direction

reaperhulk:
Not worth it.

PR #5592, opened 2020-11-29
disallow p less than 512-bit on DH

reaperhulk:
OpenSSL 3.0.0 enforces this so we'll go ahead and enforce it everywhere that's practical for us. (Note that we do not enforce on deserializing PKCS1/PKCS8 keys in < 3.0.0, but this PR adds a test so that in the 3.0.0 support branch we can test an error path)


CHANGELOG.rst
```diff
@@ -20,6 +20,10 @@ Changelog
   raise ``ValueError`` rather than ``UnsupportedAlgorithm`` when an
   unsupported cipher is used. This change is to conform with an upcoming
   OpenSSL release that will no longer distinguish between error types.
+* **BACKWARDS INCOMPATIBLE:** We no longer allow loading of finite field
+  Diffie-Hellman parameters of less than 512 bits in length. This change is to
+  conform with an upcoming OpenSSL release that no longer supports smaller
+  sizes.
```
alex, line 26: Maybe add: "These keys were already too small to be used in any secure application."

src/cryptography/hazmat/primitives/asymmetric/dh.py
```diff
@@ -95,6 +95,9 @@ def __init__(self, p, g, q=None):
         if g < 2:
             raise ValueError("DH generator must be 2 or greater")
 
+        if p.bit_length() < 512:
```
alex, line 98: Can we make `MIN_MODULUS_SIZE` constant?

alex (approved):

alex (approved):

PR #5594, opened 2020-11-30
fixes #4531 -- support encoding SCTs in certificates

### LINKED ISSUE #4531: Support encoding SCT data into an X.509 extension
Continued from #3461 since we've done everything but encoding them and it turns out that's more than a bit of a pain.

alex:


reaperhulk:
This API only allows encoding of SCTs that were previously parsed by OpenSSL. We have no concrete type for SCT at this time so I understand why it's structured that way, but I think we need a way to create SCTs that doesn't involve just passing through pre-existing pointers.

alex:
You're right that we need that API, but it's a whole separate thing -- we need an SCTBuilder or similar. I'd prefer to defer the implementation of that until we have Rust, since I'll implement it that way (and kill all of our usage of `SCT *`).

For now, I'd like to finish off the x.509 encode/decode parity.

reaperhulk (approved):

PR #5595, opened 2020-12-01
Added tls bindings for new OpenSSL APIs

alex:
fixes #5379
closes #5483

PR #5599, opened 2020-12-02
Added missing tls bindings for new OpenSSL APIs

th3b0x:
patches #5595 
re-adds setter interfaces not carried over from #5483


reaperhulk:
We don't support OpenSSL < 1.1.0 any more so these setters are always available.

th3b0x:
@reaperhulk ah, that makes sense.  I forgot that the project had dropped support for pre-1.1.0 OpenSSL.

PR #5621, opened 2020-12-09
xdist experiment

reaperhulk:


tox.ini
```diff
@@ -15,11 +15,7 @@ deps =
 passenv = ARCHFLAGS LDFLAGS CFLAGS INCLUDE LIB LD_LIBRARY_PATH USERNAME PYTHONIOENCODING OPENSSL_FORCE_FIPS_MODE
 commands =
     pip list
-    # We use parallel mode and then combine here so that coverage.py will take
-    # the paths like .tox/py38/lib/python3.8/site-packages/cryptography/__init__.py
-    # and collapse them into src/cryptography/__init__.py.
-    coverage run --parallel-mode -m pytest --capture=no --strict --durations=10 {posargs}
-    coverage combine
+    pytest -n 2 --cov=src/cryptography --cov=tests --capture=no --strict --durations=10 {posargs}
```
alex, line 18: Is there a `-n auto` or something?

tox.ini
```diff
@@ -15,11 +15,7 @@ deps =
 passenv = ARCHFLAGS LDFLAGS CFLAGS INCLUDE LIB LD_LIBRARY_PATH USERNAME PYTHONIOENCODING OPENSSL_FORCE_FIPS_MODE
```
alex, line 15: Can you delete line 12 here?

tests/hazmat/backends/test_openssl_memleak.py
```diff
@@ -126,6 +126,14 @@ def free(ptr, path, line):
 def assert_no_memory_leaks(s, argv=[]):
     env = os.environ.copy()
     env["PYTHONPATH"] = os.pathsep.join(sys.path)
+
+    # When using pytest-cov it attempts to instrument subprocesses. This
+    # causes the memleak tests to raise exceptions.
+    # we don't need coverage so we remove the env vars.
+    env.pop("COV_CORE_CONFIG")
+    env.pop("COV_CORE_DATAFILE")
+    env.pop("COV_CORE_SOURCE")
```
alex, line 135:
```suggestion
    env.pop("COV_CORE_CONFIG", None)
    env.pop("COV_CORE_DATAFILE", None)
    env.pop("COV_CORE_SOURCE", None)
```

tox.ini
alex, line 18: Based on production coverage results, `src/cryptography` doesn't look correct.

reaperhulk: There is, but I'm reluctant to use it because it will spawn more processes than are useful quite easily (e.g. on my laptop it spawns 12, which is slower than just doing 4 on this 6 core/12 thread machine).

alex: That's... interesting... do we have any idea why it gets slower beyond 4 cores?

reaperhulk: As best I can tell it's a function of the cost of collection and merging after test completion. I can switch to auto though because it doesn't really matter for our CI (and I don't run tox locally).

alex: SGTM

alex (approved):

PR #5635, opened 2020-12-11
remove windows builds and only test lowest/highest

reaperhulk:
same policy as our macOS builders


alex:
Not sure this helps much, we're not really constrained by concurrency here, just runtime.

reaperhulk:
@alex I'm not looking to make it faster, just consistent. It also minimizes future combinatorial problems as Python releases on a more rapid cadence.

alex (approved):
I am merging this under duress. I don't love it, and if anything I'd rather _expand_ our macOS CI now that it doesn't suck, but whatever.

PR #5643, opened 2020-12-13
add missing verification parameter flags

toidi:
Adding 2 missing verification parameter flags:

- `X509_V_FLAG_NO_ALT_CHAINS` If the initial chain is not trusted, do not attempt to build an alternative chain.
- `X509_V_FLAG_NO_CHECK_TIME` Do not check certificate/CRL validity against current time

Corresponding OpenSSL code:
https://github.com/openssl/openssl/blob/OpenSSL_1_1_0-stable/include/openssl/x509_vfy.h#L227-L234

### DIFF
--- a/src/_cffi_src/openssl/x509_vfy.py
+++ b/src/_cffi_src/openssl/x509_vfy.py
@@ -122,6 +122,8 @@ static const long X509_V_FLAG_SUITEB_128_LOS_ONLY;
 static const long X509_V_FLAG_SUITEB_192_LOS;
 static const long X509_V_FLAG_SUITEB_128_LOS;
 static const long X509_V_FLAG_PARTIAL_CHAIN;
+static const long X509_V_FLAG_NO_ALT_CHAINS;
+static const long X509_V_FLAG_NO_CHECK_TIME;
 
 static const long X509_LU_X509;
 static const long X509_LU_CRL;


alex:
Can you share what your use case for these is? We generally add things only as required, not merely for completeness.

toidi:
I want to incorporate efforts from https://github.com/pyca/pyopenssl/pull/948 into a tool for certificate checking.

For my special needs, I only want to test a certificate against the trust chain and ignore expiration checks.

What I have to do now is this:
```python
from OpenSSL import crypto

store = crypto.X509Store()

# Do not check certificate/CRL validity against current time
# https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/include/openssl/x509_vfy.h#L241-L242
X509_V_FLAG_NO_CHECK_TIME = 0x200000

store.load_locations(cafile=CA_FILE)
store.set_flags(X509_V_FLAG_NO_CHECK_TIME)
```

I'm aware contribution to pyOpenSSL is discouraged, so I try my luck here.

toidi:
@alex , I've updated my post several times.

Please let me know if the final version makes sense for you.

alex:
That's hlepful, thanks.

alex (approved):

PR #5647, opened 2020-12-15
feat: test on macos big sur

jeremad:
Add jobs to test on MacOS Big Sur

### DIFF
--- a/.github/workflows/wheel-builder.yml
+++ b/.github/workflows/wheel-builder.yml
@@ -55,7 +55,7 @@ jobs:
           path: cryptography-wheelhouse/
 
   macos:
-    runs-on: macos-latest
+    runs-on: ${{ matrix.os.VERSION }}
     strategy:
       matrix:
         PYTHON:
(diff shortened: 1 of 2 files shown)


.github/workflows/wheel-builder.yml
```diff
@@ -55,14 +55,17 @@ jobs:
           path: cryptography-wheelhouse/
 
   macos:
-    runs-on: macos-latest
+    runs-on: ${{ matrix.os }}
```
alex, line 58: There's no need to do change our wheel builder, the wheels we build already work with any macOS version 10.10 and greater.

jeremad: I should probably have opened an issue first, but the prebuilt-wheel does not work on mac 11, it tries to re-compile

alex: Trying upgrading to a newer pip. It sounds like you're hitting https://github.com/pypa/pip/issues/9138

jeremad: Fair enough, I'll remove the duplicated release job then!

jeremad: Actually the fixed is not released yet, it should come around pretty soon though

reaperhulk:
I appreciate the fact that this PR is well-structured (really!), but we won't merge this for somewhat counterintuitive reasons.

1) The concern you had (wheel not working with later macOS release) is not testable by our current CI since that requires packaging the wheel with our wheel builders and then downloading it onto a later OS release. This is doable, but complex for a case that has never caused an actual bug for us. In an ideal world we'd add that job anyway, but 7 years of constantly rebuilding our CI in new form has convinced me that we need to justify complexity given the very real costs (in both time and even more importantly reliability).
2) We used to test on multiple macOS releases, but our dependency on specific versions of macOS is effectively nothing at this point (we have a single code path that depends on version, which we do test but is not used in any binaries produced by our wheel builder). Since macOS concurrency is low (5 jobs max) and we've never had a bug related to macOS versioning the decision was made to test only on latest. With GHA this currently means Catalina but we're comfortable with it switching to Big Sur whenever they swap it.

All the previous being said, clearly our CI will expand when arm64 macOS becomes available to us since that different arch (and the need to build wheels against it!) is sufficiently different to warrant CI resourcing.

jeremad:
OK! I understand, thank you anyway!

PR #5652, opened 2020-12-17
Remove last large usage of parametrize with vectors

alex:


alex:
My claim of last is based on:

```
$ py.test --collect-only | grep -o -E '<Function [a-zA-Z0-9_]+' | sort | uniq -c | sort -n
```

reaperhulk (approved):

PR #5662, opened 2020-12-21
Add ssl version constants

mhils:
I'm upgrading mitmproxy to use `SSL_CTX_set_min_proto_version` (with the idea to get the changes into pyOpenSSL as well), and it turns out the version constants are not exposed at the moment. :)

`SSL_CTX_set_min_proto_version` landed in https://github.com/pyca/cryptography/pull/5595, see https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_min_proto_version.html for the OpenSSL docs.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -590,8 +590,11 @@ static const long TLS_ST_OK = 0;
 #if CRYPTOGRAPHY_IS_LIBRESSL
 static const long SSL_OP_NO_DTLSv1 = 0;
 static const long SSL_OP_NO_DTLSv1_2 = 0;
-/* it's not really clear why DTLS1_VERSION is defined, but it is */
-/* don't set this to 0, SSL_CTX_set_min_proto_version(DTLS1_2_VERSION) would silently allow all versions */
+/* it's not really clear why DTLS1_VERSION is defined, but it is. */
+/*
+don't set this to 0,
+SSL_CTX_set_min_proto_version(TLS1_3_VERSION) would silently allow all version
+*/
 static const long DTLS1_2_VERSION = -1;
 long (*DTLS_set_link_mtu)(SSL *, long) = NULL;
 long (*DTLS_get_link_min_mtu)(SSL *) = NULL;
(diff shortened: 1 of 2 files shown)


src/_cffi_src/openssl/ssl.py
```diff
@@ -677,6 +691,11 @@
 
 #if CRYPTOGRAPHY_OPENSSL_LESS_THAN_111
 static const long Cryptography_HAS_TLSv1_3 = 0;
+/*
+don't set this to 0,
+SSL_CTX_set_min_proto_version(TLS1_3_VERSION) would silently allow all version
+*/
```
alex, line 697: This doesn't matter, since we make the constant inaccessible when it's not available. Please make this consistent with all the other constants and remove the comment.

src/_cffi_src/openssl/ssl.py
```diff
@@ -582,6 +590,12 @@
 #if CRYPTOGRAPHY_IS_LIBRESSL
 static const long SSL_OP_NO_DTLSv1 = 0;
 static const long SSL_OP_NO_DTLSv1_2 = 0;
+/* it's not really clear why DTLS1_VERSION is defined, but it is. */
+/*
+don't set this to 0,
+SSL_CTX_set_min_proto_version(TLS1_3_VERSION) would silently allow all version
+*/
+static const long DTLS1_2_VERSION = -1;
```
alex, line 598: This also needs to be included in `_conditional.py`.

src/_cffi_src/openssl/ssl.py
```diff
@@ -582,6 +590,12 @@
 #if CRYPTOGRAPHY_IS_LIBRESSL
 static const long SSL_OP_NO_DTLSv1 = 0;
 static const long SSL_OP_NO_DTLSv1_2 = 0;
+/* it's not really clear why DTLS1_VERSION is defined, but it is. */
```
alex, line 593: What's `DTLS1_VERSION`? I don't see that anywhere in the code.

mhils: Defined above, it's one of the new OpenSSL version constants for DTLS. 

LibreSSL doesn't support DTLS 1.2 (https://github.com/libressl-portable/portable/issues/380), but it's a bit of a puzzle to me whether/how they do support DTLS 1.0. 

At the very least, they have DTLS1_VERSION defined (but apparently not SSL_OP_NO_DTLSv1 if I understand line 591 correctly?), and setting that to =0 here (as I do with DTLS1_2_VERSION ) would break the build.

src/_cffi_src/openssl/ssl.py
mhils: `SSL_OP_NO_DTLSv1` etc. aren't in `_conditional.py` yet either. Do you want me to add a new section there for everything in this block or is that intentional?

src/_cffi_src/openssl/ssl.py
alex: If there's no need for the constant, just don't expose it :-)

src/_cffi_src/openssl/ssl.py
mhils: This resolves itself by not including the DTLS constants. :-)

alex (approved):

mhils:
As always, thanks for the super super quick helpful feedback! ❤️

PR #5668, opened 2020-12-23
Convert unpadding code to Rust

alex:


src/rust/src/lib.rs
```diff
@@ -2,7 +2,72 @@
...
 // for complete details.
 
+/// Returns the value of the input with the most-significant-bit copied to all
+/// of the bits.
+fn duplicate_msb_to_all(a: u16) -> u16 {
+    (1 - (a >> 15)).wrapping_sub(1)
+}
+
+/// This returns 0xFFFF if a < b else 0x0000, but does so in a constant time
+/// fashion.
```
h-vetinari, line 12: Pretty sure that "docstring" is not accurate; here's how I believe it should look. I presume this is a known limitation, but the impact of it on the rest of the code here is not obvious to me.
```suggestion
/// This returns 0xFFFF if a < b (more specifically: 0 < b - a < 2^15 OR a - b >= 2^15),
/// else 0x0000, and does so in a constant-time fashion.
```
If I'm not gravely misunderstanding something, the function should look roughly like this:
```
b /\
  |          2^15        2^16-1
 _|___________|___________|
  |          /           /|
  | 0x0000 /           /  |
  |      /           /    |
  |    /           /      |
  |  /   0xFFFF  /        |
 _|/           /          |
  |          /           /|   
  |        /  0x0000   /  |
  |      /           /    |
  |    /           /      |
  |  /           / 0xFFFF |
 _|/__________ /__________|__\ a
  |           |           |  /
  0          2^15        2^16-1
```

alex: Are you saying there's a behvioral difference between the Rust and the C implementations, or that the C implementation also has this issue?

h-vetinari: The way I understand it, both implementations behave the same.

I was just trying to follow what the implementation was doing... Since the output of `duplicate_msb_to_all` (as well as `Cryptography_DUPLICATE_MSB_TO_ALL`) switches every 2^15 (i.e. whether `a >> 15` is 0 or 1, together with the fact that we wrap at 2^16), and the difference `a - b` can be anything in `[-(2^16-1), 2^16 -1]`, it cannot be constant for all possible combinations of `a < b`.

alex: Interesting, the origin of this code appears to have been derived from OpenSSL:

https://github.com/openssl/openssl/commit/5a3d21c0585064292bde5cd34089e120487ab687#diff-a1876514e966a072053ae160d38ab53d245761e72e32281ecb9d16e7a6fe1b0eL70

However, it looks like their modern lt logic is different: https://github.com/openssl/openssl/blob/master/include/internal/constant_time.h#L120

h-vetinari: Yeah, with those XORs & ORs it sounds more plausible (didn't verify though).

alex: Do you have a specific test case that this fails on, that we can add? AFACIT this issue isn't reachable in actual code, since one of the values is always known to be in `[0, 256)`, does that look right to you?

PS: Thanks so much for taking a look at this code!

alex: I've gone ahead and swithed this to use `u8` everywhere, which I think fixes the issue, it was a bug that `u16` was ever used in that code IMO -- it reflect a confusion between bits (which are >256) and bytes (which aren't).

h-vetinari: You're welcome - I was just checking this out of curiosity, and with some mathematical (but very little cryptographic) background. I believe the problem can still occur even if `a` or `b` are restricted to 0-256 (though the problematic area obviously gets much smaller, proportionally).

Still, I'd prefer the variant from openssl and in any case, would recommend not relying on my musings but rather double-checking with the openssl-people. 🙃

src/rust/src/lib.rs
```diff
@@ -2,7 +2,72 @@
...
 // for complete details.
 
+/// Returns the value of the input with the most-significant-bit copied to all
+/// of the bits.
+fn duplicate_msb_to_all(a: u8) -> u8 {
+    (1 - (a >> 7)).wrapping_sub(1)
+}
+
+/// This returns 0xFF if a < b else 0x0000, but does so in a constant time
+/// fashion.
```
h-vetinari, line 12: The problem is size-independent... Moving to `u8` just changes the constants, but the problem stays the same.

alex: Indeed, I'm adapting the logic from OpenSSL and adding a test that simply tries everything in the `(a, b)` state space, which is practical now that they're `u8`

src/rust/src/lib.rs
```diff
@@ -2,7 +2,84 @@
...
 // for complete details.
 
+/// Returns the value of the input with the most-significant-bit copied to all
+/// of the bits.
+fn duplicate_msb_to_all(a: u8) -> u8 {
+    0u8.wrapping_sub(a >> 7)
+}
+
+/// This returns 0xFF if a < b else 0x00, but does so in a constant time
+/// fashion.
```
h-vetinari, line 12: I'd leave a note to the source - a reference to an openssl implementation would have likely convinced me not to puzzle my way through whether this does what it says on the label. ;-)

```suggestion
/// fashion. The logic comes from openssl, compare
/// https://github.com/openssl/openssl/blob/OpenSSL_1_1_1i/include/internal/constant_time.h#L123
```

Also, are the triple-`///` intentional?

src/rust/src/lib.rs
```diff
@@ -2,7 +2,84 @@
...
+fn constant_time_lt(a: u8, b: u8) -> u8 {
+    duplicate_msb_to_all(a ^ ((a ^ b) | ((a.wrapping_sub(b)) ^ b)))
+}
```
h-vetinari, line 15: Nit: unnecessary braces
```suggestion
fn constant_time_lt(a: u8, b: u8) -> u8 {
    duplicate_msb_to_all(a ^ ((a ^ b) | (a.wrapping_sub(b) ^ b)))
}
```

h-vetinari (commented):
I think adopting the openssl-implementation is a good idea, and even better with the fully exhaustive test 👍

src/rust/src/lib.rs
alex: Triple `///` are doc-comments in Rust, meaning if you compiled this with rustdoc they'd show up as docs for hte function.

src/rust/src/lib.rs
```diff
@@ -2,7 +2,86 @@
...
+fn duplicate_msb_to_all(a: u8) -> u8 {
+    0u8.wrapping_sub(a >> 7)
+}
+
+/// This returns 0xFF if a < b else 0x00, but does so in a constant time
+/// fashion.
+fn constant_time_lt(a: u8, b: u8) -> u8 {
+    // Derived from:
+    // https://github.com/openssl/openssl/blob/OpenSSL_1_1_1i/include/internal/constant_time.h#L120
+    duplicate_msb_to_all(a ^ ((a ^ b) | (a.wrapping_sub(b)) ^ b))
```
h-vetinari, line 16:
```suggestion
    duplicate_msb_to_all(a ^ ((a ^ b) | (a.wrapping_sub(b) ^ b)))
```

armw4:
Awesome!!! Public Private key Pairs like Git and ssh-keygen support, etc.

src/rust/src/lib.rs
```diff
@@ -2,7 +2,86 @@
...
+    duplicate_msb_to_all(a ^ ((a ^ b) | (a.wrapping_sub(b) ^ b)))
+}
+
+#[pyo3::prelude::pyfunction]
+fn check_pkcs7_padding(data: &[u8], block_len: u8) -> bool {
+    let mut mismatch = 0;
+    let pad_size = data[(block_len - 1) as usize];
+    for i in 0..block_len {
+        let mask = constant_time_lt(i, pad_size);
+        let b = data[(block_len - 1 - i) as usize] as u8;
```
Stranger6667, line 25: It seems like `as u8` is unnecessary here as (as per `cargo check` with the `trivial_numeric_casts` check enabled)

src/rust/src/lib.rs
```diff
@@ -2,7 +2,86 @@
...
 
+/// Returns the value of the input with the most-significant-bit copied to all
+/// of the bits.
+fn duplicate_msb_to_all(a: u8) -> u8 {
+    0u8.wrapping_sub(a >> 7)
+}
+
+/// This returns 0xFF if a < b else 0x00, but does so in a constant time
+/// fashion.
+fn constant_time_lt(a: u8, b: u8) -> u8 {
```
dignifiedquire, line 13: any specific reason not to use subtle https://github.com/dalek-cryptography/subtle for constant time operations?

src/rust/src/lib.rs
```diff
@@ -2,7 +2,86 @@
...
+
+#[pyo3::prelude::pyfunction]
+fn check_pkcs7_padding(data: &[u8], block_len: u8) -> bool {
+    let mut mismatch = 0;
+    let pad_size = data[(block_len - 1) as usize];
+    for i in 0..block_len {
+        let mask = constant_time_lt(i, pad_size);
+        let b = data[(block_len - 1 - i) as usize] as u8;
+        mismatch |= mask & (pad_size ^ b);
+    }
```
LingMan, line 27: The assignment of `pad_size` will panic if `data` is empty or `block_len` is 0. I'd make that explicit with an 
```rust
assert!(data.len() >= usize::from(block_len) && block_len > 0);
```
as first line. Usually that also helps the optimizer avoid bounds checks but it's not enough in this case.

If you rewrite the loop the rusty way using iterators, the bounds checks disappear and the code gets vectorized:
```rust
// Is `block_len` some C interface remnant? Slices know their length,
// so if `block_len == data.len()` we could skip this.
let data = &data[..usize::from(block_len)];
// This unwrap will always succeed because above we asserted that data.len() > 0.
let pad_size = *data.last().unwrap();
// Usually I'd use `data.iter().enumerate()` here, but then `i` would be u32
// and need a cast to u8 for the call to `constant_time_lt. At least on the
// type level that would be a lossy cast, which I'd rather avoid.
//
// I did consider reversing the range with `rev()` instead of doing it manually
// with `block_len - 1 - i`, but the code looses the vectorizations then.
for (i, b) in (0..block_len).zip(data.iter()) {
    let mask = constant_time_lt(block_len - 1 - i, pad_size);
    mismatch |= mask & (pad_size ^ b);
}
```

PS: Not a crypto person, so take this with the appropriate amount of salt.

src/rust/src/lib.rs
LingMan: Note that there has been some [drama around maintainership](https://old.reddit.com/r/rust/comments/lcsg0u/ownership_and_maintenance_struggles_in/) in dalek-cryptography. There's a fork called [subtle-ng](https://github.com/zkcrypto/subtle-ng) now.

src/rust/src/lib.rs
```diff
@@ -2,9 +2,93 @@
...
+    let pad_size = data[data.len() - 1];
+    let len: u8 = data.len().try_into().expect("data too long");
+    for (i, b) in (0..len).zip(data.iter().rev()) {
+        let mask = constant_time_lt(i, pad_size);
+        mismatch |= mask & (pad_size ^ b);
+    }
+
+    // Check to make sure the pad_size was within the valid range.
+    mismatch |= !constant_time_lt(0, pad_size);
+    mismatch |= constant_time_lt(data.len() as u8, pad_size);
```
LingMan, line 35: We have `len` now, no need for another (theoretically) lossy cast.
```suggestion
    mismatch |= constant_time_lt(len, pad_size);
```

src/rust/src/lib.rs
```diff
@@ -2,9 +2,92 @@
...
+    // https://github.com/openssl/openssl/blob/OpenSSL_1_1_1i/include/internal/constant_time.h#L120
+    duplicate_msb_to_all(a ^ ((a ^ b) | (a.wrapping_sub(b) ^ b)))
+}
+
+#[pyo3::prelude::pyfunction]
+fn check_pkcs7_padding(data: &[u8]) -> bool {
+    assert!(!data.is_empty());
+
+    let mut mismatch = 0;
+    let pad_size = data[data.len() - 1];
```
LingMan, line 26: Now that `block_len` is gone this would be more natural than an assert:
```suggestion
    let mut mismatch = 0;
    let pad_size = *data.first().expect("data is empty");
```

src/rust/src/lib.rs
```diff
@@ -2,9 +2,92 @@
...
+    let len: u8 = data.len().try_into().expect("data too long");
+    // Skip the first one with the pad size
+    for (i, b) in (1..len).zip(data[..data.len() - 1].iter().rev()) {
+        let mask = constant_time_lt(i, pad_size);
+        mismatch |= mask & b;
+    }
+
+    // Check to make sure the pad_size was within the valid range.
+    mismatch |= !constant_time_lt(0, pad_size);
+    mismatch |= constant_time_lt(data.len() as u8, pad_size);
```
LingMan, line 61: Same as above.
```suggestion
    mismatch |= constant_time_lt(len, pad_size);
```

src/rust/src/lib.rs
```diff
@@ -2,9 +2,92 @@
...
+    // Now check the low bit to see if it's set
+    (mismatch & 1) == 0
+}
+
+#[pyo3::prelude::pyfunction]
+fn check_ansix923_padding(data: &[u8]) -> bool {
+    assert!(!data.is_empty());
+
+    let mut mismatch = 0;
+    let pad_size = data[data.len() - 1];
```
LingMan, line 51: Same as above.
```suggestion
    let mut mismatch = 0;
    let pad_size = *data.first().expect("data is empty");
```

src/rust/src/lib.rs
```diff
@@ -2,9 +2,92 @@
...
+    duplicate_msb_to_all(a ^ ((a ^ b) | (a.wrapping_sub(b) ^ b)))
+}
+
+#[pyo3::prelude::pyfunction]
+fn check_pkcs7_padding(data: &[u8]) -> bool {
+    assert!(!data.is_empty());
+
+    let mut mismatch = 0;
+    let pad_size = data[data.len() - 1];
+    let len: u8 = data.len().try_into().expect("data too long");
```
LingMan, line 27: Looks like the comment got lost when you inserted my iterator based loop.
```suggestion
    // Skip the first one with the pad size
    let len: u8 = data.len().try_into().expect("data too long");
```

reaperhulk (approved):

Issue #5670, opened 2020-12-25
Add inline type hints


reaperhulk:
Still to do:

- [ ] Some symmetric cipher base work
- [x] KDF

alex:
The one outstanding thing for symmetric ciphers is `encryptor()` + `decryptor()`. These are basically impossible to type correctly because of the screwed up way we handle AEADs. I think those may have to remain untyped.

PR #5709, opened 2021-01-29
Apply type annotations to x509 builders

alex:
Refs #5670

### DIFF
--- a/src/cryptography/utils.py
+++ b/src/cryptography/utils.py
@@ -6,6 +6,7 @@
 import abc
 import inspect
 import sys
+import typing
 import warnings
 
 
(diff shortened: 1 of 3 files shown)


src/cryptography/utils.py
```diff
@@ -23,19 +24,19 @@ class CryptographyDeprecationWarning(UserWarning):
...
 
 
-def _check_bytes(name, value):
+def _check_bytes(name: str, value: bytes):
     if not isinstance(value, bytes):
         raise TypeError("{} must be bytes".format(name))
 
 
-def _check_byteslike(name, value):
+def _check_byteslike(name: str, value: bytes):
```
reaperhulk, line 32: lol

alex: 🤷

src/cryptography/x509/base.py
```diff
@@ -652,14 +675,14 @@ def not_valid_after(self, time):
             self._extensions,
         )
 
-    def add_extension(self, extension, critical):
+    def add_extension(self, extval: ExtensionType, critical: bool):
```
reaperhulk, line 678: If we're going to change the name of this arg it needs to be updated in the docs (and this would break kwarg users right?)

alex: It's hard to know anything for certain, but I claim that there's no reason you'd ever pass this value as a kwarg. I'll go fix the docs.

reaperhulk (approved):

PR #5712, opened 2021-01-30
Apply type annotations to x509 ct and ocsp

alex:
Refs #5670

### DIFF
--- a/src/cryptography/fernet.py
+++ b/src/cryptography/fernet.py
@@ -157,8 +157,8 @@ class Fernet(object):
 
 
 class MultiFernet(object):
-    def __init__(self, fernets_it: typing.Iterator[Fernet]):
-        fernets = list(fernets_it)
+    def __init__(self, fernets: typing.Iterable[Fernet]):
+        fernets = list(fernets)
         if not fernets:
             raise ValueError(
                 "MultiFernet requires at least one Fernet instance"
--- a/src/cryptography/x509/ocsp.py
+++ b/src/cryptography/x509/ocsp.py
@@ -5,12 +5,14 @@
 
 import abc
 import datetime
+import typing
 from enum import Enum
 
 from cryptography import x509
-from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.x509.base import (
     _EARLIEST_UTC_TIME,
+    _PRIVATE_KEY_TYPES,
     _convert_to_naive_utc_time,
     _reject_duplicate_extension,
 )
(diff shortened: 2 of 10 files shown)


src/cryptography/fernet.py
```diff
@@ -157,8 +157,8 @@ def _decrypt_data(
 
 
 class MultiFernet(object):
-    def __init__(self, fernets_it: typing.Iterator[Fernet]):
-        fernets = list(fernets_it)
+    def __init__(self, fernets: typing.Iterable[Fernet]):
```
reaperhulk, line 160: Changing this back to match the docs?

alex: `list` isn't iterator, but it is iterable

reaperhulk (changes requested):
`_OCSPRequest` and `_OCSPResponse`, as well as `OCSPRequest` and `OCSPResponse` (the ABCs) are missing type information on quite a few properties. `Extensions` is as well.

`RFC822Name`, `DNSName`, `UniformResourceIdentifier`, `DirectoryName`, `RegisteredID`, `IPAddress`, `OtherName`,  `NameAttribute`, and more don't have typing on various dunder methods, while we *did* add typing to those methods in past PRs (e.g. https://github.com/pyca/cryptography/pull/5711/files#diff-c42cf4fc4d5c5eacf9b7cda43dbb0ab06335d0b9804814a20f488a827987d847R48).

Maybe we should just strip the typing on dunder methods?

src/cryptography/x509/ocsp.py
```diff
@@ -184,279 +137,339 @@ def __init__(
...
+class OCSPResponseBuilder(object):
+    def __init__(
+        self, response=None, responder_id=None, certs=None, extensions=[]
+    ):
+        self._response = response
+        self._responder_id = responder_id
+        self._certs = certs
+        self._extensions = extensions
+
+    def add_response(
```
reaperhulk, line 359: All these functions should have `-> "OCSPResponseBuilder"` for return type so mypy can do chaining.

src/cryptography/x509/ocsp.py
```diff
@@ -184,279 +137,339 @@ def __init__(
...
         The list of single response extensions. Not response extensions.
         """
+
+
+class OCSPRequestBuilder(object):
+    def __init__(self, request=None, extensions=[]):
+        self._request = request
+        self._extensions = extensions
+
+    def add_certificate(
```
reaperhulk, line 313: All these functions should have `-> "OCSPRequestBuilder"` for return type so mypy can do chaining.

reaperhulk (approved):

PR #5713, opened 2021-01-30
type hinting for hashes, hmac, and cmac

reaperhulk:


src/cryptography/hazmat/backends/openssl/cmac.py
```diff
@@ -65,7 +65,7 @@ def finalize(self):
 
         return self._backend._ffi.buffer(buf)[:]
 
-    def copy(self):
+    def copy(self) -> "_CMACContext":
```
alex, line 68: Should this be the interface, not the concrete type?

src/cryptography/hazmat/backends/openssl/hashes.py
```diff
@@ -37,7 +36,7 @@ def __init__(self, backend, algorithm, ctx=None):
 
     algorithm = utils.read_only_property("_algorithm")
 
-    def copy(self):
+    def copy(self) -> "_HashContext":
```
alex, line 39: same.

src/cryptography/hazmat/backends/openssl/hmac.py
```diff
@@ -41,7 +42,7 @@ def __init__(self, backend, key, algorithm, ctx=None):
 
     algorithm = utils.read_only_property("_algorithm")
 
-    def copy(self):
+    def copy(self) -> "_HMACContext":
```
alex, line 45: Same 😂

src/cryptography/hazmat/backends/openssl/cmac.py
reaperhulk: There's no CMAC interface 😄. For HMAC and Hash we also have a structure where we instantiate an object which in turn instantiates a context. Every user-facing operation goes through the primary object, which in turn invokes the underlying context. So the user will never actually interact with these context objects directly.

alex (approved):

PR #5714, opened 2021-01-30
add typing to padding and constant time modules

reaperhulk:


src/cryptography/hazmat/primitives/padding.py
```diff
@@ -48,15 +49,17 @@ def _byte_padding_update(buffer_, data, block_size):
     return buffer_, result
 
 
-def _byte_padding_pad(buffer_, block_size, paddingfn):
+def _byte_padding_pad(
+    buffer_: typing.Optional[bytes], block_size: int, paddingfn: typing.Callable
```
alex, line 53: You don't want to annotate the types on the callable?

src/cryptography/hazmat/primitives/padding.py
```diff
@@ -100,44 +105,42 @@ def unpadder(self):
...
 
-@utils.register_interface(PaddingContext)
-class _PKCS7PaddingContext(object):
-    def __init__(self, block_size):
+class _PKCS7PaddingContext(PaddingContext):
+    def __init__(self, block_size: int):
         self.block_size = block_size
         # TODO: more copies than necessary, we should use zero-buffer (#193)
-        self._buffer = b""
+        self._buffer: typing.Optional[bytes] = b""
```
alex, line 112: I think we should type attributes at the class level, not in `__init__`

alex (approved):

PR #5718, opened 2021-01-31
add serialization type hinting

reaperhulk:
It's not entirely clear to me how useful return types that are a big union really are. But I don't know how to do better, so ¯\_(ツ)_/¯

### DIFF
--- a/src/cryptography/hazmat/primitives/serialization/base.py
+++ b/src/cryptography/hazmat/primitives/serialization/base.py
@@ -4,38 +4,47 @@
 
 
 import abc
+import typing
 from enum import Enum
 
 from cryptography import utils
 from cryptography.hazmat.backends import _get_backend
+from cryptography.hazmat.primitives.asymmetric import dh
+
+if typing.TYPE_CHECKING:  # pragma: no cover
+    from cryptography.x509.base import _PRIVATE_KEY_TYPES, _PUBLIC_KEY_TYPES
 
 
-def load_pem_private_key(data, password, backend=None):
+def load_pem_private_key(
+    data: bytes, password: typing.Optional[bytes], backend=None
+) -> "_PRIVATE_KEY_TYPES":
     backend = _get_backend(backend)
     return backend.load_pem_private_key(data, password)
 
 
-def load_pem_public_key(data, backend=None):
+def load_pem_public_key(data: bytes, backend=None) -> "_PUBLIC_KEY_TYPES":
     backend = _get_backend(backend)
     return backend.load_pem_public_key(data)
 
 
-def load_pem_parameters(data, backend=None):
+def load_pem_parameters(data: bytes, backend=None) -> dh.DHParameters:
     backend = _get_backend(backend)
     return backend.load_pem_parameters(data)
 
 
-def load_der_private_key(data, password, backend=None):
+def load_der_private_key(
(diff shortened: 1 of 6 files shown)


src/cryptography/hazmat/primitives/serialization/base.py
```diff
@@ -4,38 +4,47 @@
...
 
 import abc
+import typing
 from enum import Enum
 
 from cryptography import utils
 from cryptography.hazmat.backends import _get_backend
+from cryptography.hazmat.primitives.asymmetric import dh
 
+if typing.TYPE_CHECKING:  # pragma: no cover
```
alex, line 14: I don't like this, this is making our code messier and more confusing for the type checker. Can we put these constants in some module that avoids the circular issue?

src/cryptography/hazmat/primitives/serialization/base.py
```diff
@@ -4,38 +4,44 @@
...
-from cryptography import utils
+from cryptography.hazmat._types import _PRIVATE_KEY_TYPES, _PUBLIC_KEY_TYPES
 from cryptography.hazmat.backends import _get_backend
+from cryptography.hazmat.primitives.asymmetric import dh
 
 
-def load_pem_private_key(data, password, backend=None):
+def load_pem_private_key(
+    data: bytes, password: typing.Optional[bytes], backend=None
+) -> "_PRIVATE_KEY_TYPES":
```
alex, line 17: No need for quotes

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -613,11 +627,21 @@ def serialize_ssh_private_key(private_key, password=None):
         ciph.encryptor().update_into(buf[ofs:mlen], buf[ofs:])
 
     txt = _ssh_pem_encode(buf[:mlen])
-    buf[ofs:mlen] = bytearray(slen)
+    # Ignore the following type because mypy wants
+    # Sequence[bytes] but what we're passing is fine.
```
alex, line 631: Err, mypy doesn't consider `bytearray` to be a `Sequence[bytes]`? Can we please file a bug with them on this.

reaperhulk: Correct:

```python
buf = memoryview(bytearray(b"a" * 500))
buf[3:4] = bytearray(1)
```
```
test.py:2: error: Incompatible types in assignment (expression has type "bytearray", target has type "Sequence[bytes]")
```

alex: Please file a mypy bug on this.

I suspect you can work around this with `slen_buf: Sequence[bytes] = bytearray(slen); buf[ofs:mlen] = slen_buf` though.

reaperhulk: That doesn't work

```
error: Incompatible types in assignment (expression has type "bytearray", variable has type "Sequence[bytes]")
```

alex: Welp. Anyways, mypy bug please.

alex (dismissed):

alex (dismissed):

alex (approved):

PR #5719, opened 2021-01-31
type hinting for symmetric ciphers

reaperhulk:


src/cryptography/hazmat/primitives/ciphers/modes.py
```diff
@@ -104,90 +109,78 @@ def __init__(self, tweak):
...
-    def validate_for_algorithm(self, algorithm):
+    def validate_for_algorithm(self, algorithm: CipherAlgorithm):
         _check_aes_key_length(self, algorithm)
-        if len(self.nonce) * 8 != algorithm.block_size:
-            raise ValueError(
-                "Invalid nonce size ({}) for {}.".format(
-                    len(self.nonce), self.name
-                )
-            )
+        _check_nonce_length(self.nonce, self.name, algorithm)
```
reaperhulk, line 170: This is a workaround. mypy doesn't support intersection types and in this case we want to have an algorithm that is both a `CipherAlgorithm` and a `BlockCipherAlgorithm`. Since we can't have that we hide the call that requires `BlockCipherAlgorithm` behind an untyped call.

There may be a better way to do this.

alex (dismissed):

alex:
Err, tests are failing.

alex (approved):

PR #5720, opened 2021-01-31
Types check untyped defs

reaperhulk:
Slowly converging on good types...

I have tests down to about 80 errors, but they're more obnoxious to fix. There are some points of discussion in this PR potentially as well.

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/binding.py
+++ b/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -91,7 +91,7 @@ def _openssl_assert(lib, ok, errors=None):
 
 def build_conditional_library(lib, conditional_names):
     conditional_lib = types.ModuleType("lib")
-    conditional_lib._original_lib = lib
+    conditional_lib._original_lib = lib  # type: ignore[attr-defined]
     excluded_names = set()
     for condition, names_cb in conditional_names.items():
         if not getattr(lib, condition):
--- a/src/cryptography/hazmat/primitives/twofactor/totp.py
+++ b/src/cryptography/hazmat/primitives/twofactor/totp.py
@@ -36,7 +36,7 @@ class TOTP(object):
         self._time_step = time_step
         self._hotp = HOTP(key, length, algorithm, backend, enforce_key_length)
 
-    def generate(self, time: int):
+    def generate(self, time: typing.Union[int, float]):
         counter = int(time / self._time_step)
         return self._hotp.generate(counter)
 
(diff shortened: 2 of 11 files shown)


src/cryptography/hazmat/_types.py
```diff
@@ -27,3 +27,10 @@
     dsa.DSAPrivateKey,
     ec.EllipticCurvePrivateKey,
 ]
+_PRIVATE_KEY_TYPES_WITH_SERIALIZATION = typing.Union[
```
reaperhulk, line 30: Should we consider just making PrivateKey and PrivateKeyWithSerialization the same thing? If we ever support opaque key references we can just make the private_bytes method raise an exception, and it would significantly simplify this typing challenge.

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -91,7 +91,7 @@ def _openssl_assert(lib, ok, errors=None):
 
 def build_conditional_library(lib, conditional_names):
     conditional_lib = types.ModuleType("lib")
-    conditional_lib._original_lib = lib
+    conditional_lib._original_lib = lib  # type: ignore[attr-defined]
```
reaperhulk, line 94: mypy allows you to ignore specific subsets of errors, so that's what these are.

src/cryptography/hazmat/primitives/twofactor/totp.py
```diff
@@ -36,7 +36,7 @@ def __init__(
         self._time_step = time_step
         self._hotp = HOTP(key, length, algorithm, backend, enforce_key_length)
 
-    def generate(self, time: int):
+    def generate(self, time: typing.Union[int, float]):
```
reaperhulk, line 39: We have a test that confirms floats are allowed. We document int only though. For compatibility I defined it this way.

src/cryptography/x509/ocsp.py
```diff
@@ -304,6 +304,12 @@ def single_extensions(self) -> x509.Extensions:
         The list of single response extensions. Not response extensions.
         """
 
+    @abc.abstractmethod
+    def public_bytes(self, encoding: serialization.Encoding) -> bytes:
```
reaperhulk, line 308: mypy caught that we entirely missed this.

src/cryptography/hazmat/_types.py
alex: Yeah, I'm pro merging them. Separate PR for that though, I suppose.

reaperhulk:
#5722 should merge first since this will change the type definitions.

alex:
rebase away!

alex (approved):

PR #5725, opened 2021-02-01
DH types

reaperhulk:
In this PR I reordered some things to use less quotes and it makes the diff much harder to read. Subsequent PRs will not do this.

### DIFF
--- a/src/cryptography/hazmat/primitives/asymmetric/dh.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/dh.py
@@ -4,9 +4,11 @@
 
 
 import abc
+import typing
 
 from cryptography import utils
 from cryptography.hazmat.backends import _get_backend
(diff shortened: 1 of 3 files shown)


src/cryptography/hazmat/primitives/asymmetric/dh.py
```diff
@@ -17,41 +19,47 @@ def generate_parameters(generator, key_size, backend=None):
...
-            self._x == other._x
-            and self._public_numbers == other._public_numbers
+            self._p == other._p and self._g == other._g and self._q == other._q
         )
 
     def __ne__(self, other):
         return not self == other
 
-    def private_key(self, backend=None):
+    def parameters(self, backend=None):
```
alex, line 52: type this?

alex (approved):

PR #5726, opened 2021-02-01
dsa type hinting

reaperhulk:
This reorganizes serialization to break an import cycle (and is used in all subsequent asym PRs)


alex:
Is it possible to do the re-org in its own PR?

alex (approved):

PR #5733, opened 2021-02-01
rsa type hinting

reaperhulk:


alex:
flake issue

src/cryptography/hazmat/backends/openssl/rsa.py
```diff
@@ -30,11 +32,17 @@
...
+    RSAPublicNumbers,
 )
 
 
-def _get_rsa_pss_salt_length(pss, key, hash_algorithm):
+def _get_rsa_pss_salt_length(
+    pss: PSS,
+    key: typing.Union[RSAPrivateKey, RSAPublicKey],
+    hash_algorithm: hashes.HashAlgorithm,
+):
```
alex, line 45: Return type?

src/cryptography/hazmat/backends/openssl/rsa.py
```diff
@@ -74,22 +87,32 @@ def _enc_dec_rsa(backend, key, data, padding):
...
 
 
-def _enc_dec_rsa_pkey_ctx(backend, key, data, padding_enum, padding):
+def _enc_dec_rsa_pkey_ctx(
+    backend,
+    key: typing.Union[RSAPrivateKey, RSAPublicKey],
+    data: bytes,
+    padding_enum: int,
+    padding: AsymmetricPadding,
+):
```
alex, line 96: Another return type missing?

src/cryptography/hazmat/backends/openssl/rsa.py
```diff
@@ -74,22 +87,32 @@ def _enc_dec_rsa(backend, key, data, padding):
     return _enc_dec_rsa_pkey_ctx(backend, key, data, padding_enum, padding)
 
 
-def _enc_dec_rsa_pkey_ctx(backend, key, data, padding_enum, padding):
+def _enc_dec_rsa_pkey_ctx(
+    backend,
+    key: typing.Union[RSAPrivateKey, RSAPublicKey],
```
alex, line 92: This should be typed with the concrete types, right?

reaperhulk: It should, yes.

src/cryptography/hazmat/backends/openssl/rsa.py
```diff
@@ -74,22 +87,32 @@ def _enc_dec_rsa(backend, key, data, padding):
...
     if isinstance(key, _RSAPublicKey):
         init = backend._lib.EVP_PKEY_encrypt_init
         crypt = backend._lib.EVP_PKEY_encrypt
     else:
         init = backend._lib.EVP_PKEY_decrypt_init
         crypt = backend._lib.EVP_PKEY_decrypt
 
-    pkey_ctx = backend._lib.EVP_PKEY_CTX_new(key._evp_pkey, backend._ffi.NULL)
+    pkey_ctx = backend._lib.EVP_PKEY_CTX_new(
+        key._evp_pkey, backend._ffi.NULL  # type: ignore[union-attr]
```
alex, line 105: Now that this is typed using the implementation, not the interface, I claim that this ignore shouldn't be required.

alex, line 105: I don't think the mypy ignore should be required now? Same for all the other places in this diff to be clear.

src/cryptography/hazmat/backends/openssl/rsa.py
```diff
@@ -43,7 +51,12 @@ def _get_rsa_pss_salt_length(pss, key, hash_algorithm):
         return salt
 
 
-def _enc_dec_rsa(backend, key, data, padding):
+def _enc_dec_rsa(
+    backend,
+    key: typing.Union["_RSAPrivateKey", "_RSAPublicKey"],
+    data: bytes,
+    padding: AsymmetricPadding,
+):
```
alex, line 59:
```suggestion
) -> bytes:
```

src/cryptography/hazmat/backends/openssl/rsa.py
```diff
@@ -74,22 +87,32 @@ def _enc_dec_rsa(backend, key, data, padding):
...
     backend.openssl_assert(pkey_ctx != backend._ffi.NULL)
     pkey_ctx = backend._ffi.gc(pkey_ctx, backend._lib.EVP_PKEY_CTX_free)
     res = init(pkey_ctx)
     backend.openssl_assert(res == 1)
     res = backend._lib.EVP_PKEY_CTX_set_rsa_padding(pkey_ctx, padding_enum)
     backend.openssl_assert(res > 0)
-    buf_size = backend._lib.EVP_PKEY_size(key._evp_pkey)
+    buf_size = backend._lib.EVP_PKEY_size(
+        key._evp_pkey
+    )
```
alex, line 115: Strongly suspect you need to `black` this back.

alex (approved):

PR #5735, opened 2021-02-01
KDF type hinting

reaperhulk:


src/cryptography/hazmat/primitives/kdf/concatkdf.py
```diff
@@ -19,11 +20,15 @@
 from cryptography.hazmat.primitives.kdf import KeyDerivationFunction
 
 
-def _int_to_u32be(n):
+def _int_to_u32be(n: int):
```
alex, line 23: return type?

src/cryptography/hazmat/primitives/kdf/concatkdf.py
```diff
@@ -33,7 +38,9 @@ def _common_args_checks(algorithm, length, otherinfo):
         utils._check_bytes("otherinfo", otherinfo)
 
 
-def _concatkdf_derive(key_material, length, auxfn, otherinfo):
+def _concatkdf_derive(
+    key_material: bytes, length: int, auxfn: typing.Callable, otherinfo: bytes
```
alex, line 42: type the function

src/cryptography/hazmat/primitives/kdf/concatkdf.py
```diff
@@ -71,36 +81,46 @@ def __init__(self, algorithm, length, otherinfo, backend=None):
...
 
         _common_args_checks(algorithm, length, otherinfo)
         self._algorithm = algorithm
         self._length = length
-        self._otherinfo = otherinfo
-        if self._otherinfo is None:
-            self._otherinfo = b""
+        self._otherinfo: bytes = otherinfo if otherinfo is not None else b""
+
+        block_size = getattr(algorithm, "block_size", None)
```
alex, line 116: Hmm?

reaperhulk: `block_size` isn't a part of the `HashAlgorithm` interface (and we don't publicly document it). Adding it arguably wouldn't make sense either since sponge constructions like SHA3 don't have an internal block size either. mypy made this bug obvious so I added a new branch and test to make sure only hashes with `block_size` could be used.

I'm open to other solutions here.

alex: `BlockHashAlgorithm`?

reaperhulk: Would need to be a subclass of `HashAlgorithm` for mypy to allow it because mypy doesn't have intersection types (so you can't have a multiply-inherited class that is both `HashAlgorithm` and `BlockHashAlgorithm`).

alex: That seems reasonable?

reaperhulk: Are we comfortable with changing our CipherAlgorithms in that fashion as well? (e.g. `BlockCipherAlgorithm` is a subclass of `CipherAlgorithm`, whereas now it is not). If so, then okay, I'll do that here too.

reaperhulk: Also, would we want to document these new properties in the docs?

alex: I'm... not sure. Thoughts?

reaperhulk: If they're exposed as part of the typing it seems like we need to document them. I guess it matters in this case since ConcatKDF would literally have failed with a TypeError (when it tried to multiply a `None`) if you passed SHA3 to it.

alex: If `ConcatKDF` has the property, but it's simply `None`, maybe we should add that to the base interface as `Optional[int]`, then simply check it's not none in the places we use it. That seems better than the `getattr` shenangins.

reaperhulk: We can do that here, yes. I'd be tempted to not document that property on `HashAlgorithm` though, since it's not generally useful to our users.

alex: That's fine with me.

Would you mind doing that bit in it's own PR for ease of review? I'm trying to push us towards the typing PRs not introducing substantive changes themselves.

reaperhulk: I had no intention of doing it in this PR. This is already a 290 line diff :)

alex (approved):

PR #5743, opened 2021-02-04
Added typing for a bunch of random stuff

alex:


src/cryptography/hazmat/primitives/asymmetric/padding.py
```diff
@@ -55,14 +61,17 @@ def __init__(self, mgf, algorithm, label):
...
+    def __init__(self, algorithm: hashes.HashAlgorithm):
         if not isinstance(algorithm, hashes.HashAlgorithm):
             raise TypeError("Expected instance of hashes.HashAlgorithm.")
 
         self._algorithm = algorithm
 
 
-def calculate_max_pss_salt_length(key, hash_algorithm):
+def calculate_max_pss_salt_length(
+    key: typing.Union["rsa.RSAPrivateKey", "rsa.RSAPublicKey"],
```
reaperhulk, line 72: Shouldn't need quotes here

alex: You, in fact, do because of recursive import stuff -- `padding.py` is imported by `rsa.py` so the module isn't fully initialized.

reaperhulk (approved):

PR #5751, opened 2021-02-07
fix macos target for wheel builder

alex:


alex:
https://github.com/pyca/cryptography/actions/runs/545928908

reaperhulk (approved):

PR #5758, opened 2021-02-07
fix import cycle with asymmetricpadding

reaperhulk:


src/cryptography/hazmat/primitives/_asymmetric.py
```diff
@@ -0,0 +1,17 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+import abc
+
+
+# This exists to break an import cycle. It is normally accessible from the
+# ciphers module.
```
alex, line 9: No it's not.

reaperhulk, line 9:
```suggestion
# asymmetric padding module.
```

alex (approved):

alex:
codecov results look like nonsense

Asvasilevs (commented):
fix-2361

PR #5763, opened 2021-02-07
include cargo in the docs for alpine

alex:


reaperhulk (approved):

reaperhulk:
(This will only work on 3.13+ Alpine. Alpine users should probably just use rustup)

reaperhulk (approved):

PR #5767, opened 2021-02-08
remove faq entry that is no longer relevant and reorder

reaperhulk:
rust info should be closer to the top

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -1,33 +1,6 @@
 Frequently asked questions
 ==========================
 
-.. _faq-howto-handle-deprecation-warning:
-
-I cannot suppress the deprecation warning that ``cryptography`` emits on import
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -1,33 +1,6 @@
 Frequently asked questions
 ==========================
 
-.. _faq-howto-handle-deprecation-warning:
-
-I cannot suppress the deprecation warning that ``cryptography`` emits on import
```
alex, line 6: Hmm, we'll emit other warnings and adding/removing this is a pain, I think we probably want to hold on to this, but maybe move it to the bottom?

webknjaz: Agreed. This will arise again for as long as the built-in deprecation warning is no used.

Issue #5774, opened 2021-02-08
Typing signature breaks register_interface() for custom implementations


tiran:
FreeIPA contains a custom implementatin of ``cryptography.x509.Certificate``. Our class is not a subclass of cryptography's Certificate class. Instead it wraps a certificate instance and provides all methods with compatible signatures. We use ``register_interface`` to register our implementation as a compatible implementation of the interface

```
@crypto_utils.register_interface(crypto_x509.Certificate)
class IPACertificate:
    """
    A proxy class wrapping a python-cryptography certificate representation for
    IPA purposes
    """
    def __init__(self, cert, backend=None):
        """
        :param cert: A python-cryptography Certificate object
        :param backend: A python-cryptography Backend object
        """
        self._cert = cert
```

The introduction of type hints break our code. ``verify_interface`` rejects our implementation because it does not have type hints. Because we have to support backwards compatibility with cryptograhy 3.3, we cannot add type hints to our implementation either. Adding type hints makes our class incompatible with 3.3.

```
py39 run-test: commands[0] | /home/heimes/redhat/freeipa/.tox/py39/bin/ipa --help
Traceback (most recent call last):
  File "/home/heimes/redhat/freeipa/.tox/py39/bin/ipa", line 5, in <module>
    from ipaclient.__main__ import main
  File "/home/heimes/redhat/freeipa/.tox/py39/lib/python3.9/site-packages/ipaclient/__main__.py", line 7, in <module>
    from ipalib import api, cli
  File "/home/heimes/redhat/freeipa/.tox/py39/lib/python3.9/site-packages/ipalib/__init__.py", line 921, in <module>
    from ipalib.frontend import Command, LocalOrRemote, Updater
  File "/home/heimes/redhat/freeipa/.tox/py39/lib/python3.9/site-packages/ipalib/frontend.py", line 31, in <module>
    from ipalib.parameters import create_param, Param, Str, Flag
  File "/home/heimes/redhat/freeipa/.tox/py39/lib/python3.9/site-packages/ipalib/parameters.py", line 125, in <module>
    from ipalib.x509 import (
  File "/home/heimes/redhat/freeipa/.tox/py39/lib/python3.9/site-packages/ipalib/x509.py", line 109, in <module>
    class IPACertificate:
  File "/home/heimes/redhat/freeipa/.tox/py39/lib64/python3.9/site-packages/cryptography/utils.py", line 45, in register_decorator
    verify_interface(iface, klass)
[...]

PR #5775, opened 2021-02-08
Interface: Make annotation check optional

tiran:
Fixes: https://github.com/pyca/cryptography/issues/5774

### DIFF
--- a/src/cryptography/utils.py
+++ b/src/cryptography/utils.py
@@ -41,8 +41,8 @@ def read_only_property(name: str):
 
 
 def register_interface(iface):
-    def register_decorator(klass):
-        verify_interface(iface, klass)
+    def register_decorator(klass, *, check_annotations=False):
+        verify_interface(iface, klass, check_annotations=check_annotations)
         iface.register(klass)
         return klass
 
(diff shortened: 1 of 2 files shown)


src/cryptography/utils.py
```diff
@@ -69,7 +69,21 @@ class InterfaceNotImplemented(Exception):
...
+        [
+            inspect.Parameter(
+                param.name,
+                param.kind,
+                default=param.default,
+                annotation=inspect.Parameter.empty,
+            )
+            for param in signature.parameters.values()
+        ]
+    )
```
stanislavlevin, line 83: What about `replace` methods?
for example,
```
return signature.replace(                                                                    
    parameters=[                                                                   
        param.replace(annotation=inspect.Parameter.empty)                          
        for param in signature.parameters.values()                                       
    ],                                                                             
    return_annotation=inspect.Signature.empty,                                     
) 
```

alex:
Interesting, I hadn't realized anyone was using `register_interface` outside of our tree. WDYT @reaperhulk ?

reaperhulk:
If we're breaking people on this we should definitely merge a fix. I'm fine with this approach.

alex:
Note that this does change the default on whether it verifies sigs, I suppose that's fine?

reaperhulk (approved):

reaperhulk:
Yeah I'm fine with that to be clear :)

PR #5777, opened 2021-02-08
Remove install_requires

felixonmars:
install_requires become runtime dependencies, which is not the case here.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,7 +102,6 @@ try:
         ),
         include_package_data=True,
         python_requires=">=3.6",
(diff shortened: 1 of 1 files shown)


alex:
Thanks for the PR! Tests seem to be failing -- I think the issue is that we do need cffi to be in `install_requires`

tiran:
This looks like a duplicate of my PR https://github.com/pyca/cryptography/pull/5779
Sorry, I didn't realize you were working on the same problem.

alex:
Just merged #5779. Thanks though!

Issue #5778, opened 2021-02-08
cryptography 3.4 should not have an install_requires on setuptools_rust


tiran:
``setup.py`` lists ``setuptools_rust`` as setup and as installation dependency. cryptography only needs ``setuptools_rust`` during setup. The additional dependency causes issues with Fedora's automatic dependency generator, see https://bugzilla.redhat.com/show_bug.cgi?id=1926181

```python

setup_requirements = ["cffi>=1.12", "setuptools-rust>=0.11.4"]


setup(
    ...
    install_requires=setup_requirements,
    setup_requires=setup_requirements,
)
```

tom--pollard:
Also hitting this issue when trying to pip install docker-compose

PR #5779, opened 2021-02-08
Remove setuptools_rust from install requirement

tiran:
setuptools_rust is only required for building cryptography.

Fixes: https://github.com/pyca/cryptography/issues/5778

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,8 +39,10 @@ with open(os.path.join(src_dir, "cryptography", "__about__.py")) as f:
     exec(f.read(), about)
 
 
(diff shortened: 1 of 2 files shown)


alex:
Heh, looks like this broke the docs `tox` build, which was relying on this being implicitly installed, Can you fix that and then we can merg?

tiran:
Suggestions? I could add ``setuptools_rust`` to ``tox.ini``, but feels like a hack.

alex:
You could add a special `extras_require` with it and specify that for that build. That seems good enough to me and reduces most of the duplication

alex (approved):

PR #5802, opened 2021-02-09
Added a py.typed so mypy prefers us to typeshed

alex:
closes #5796

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -7,6 +7,7 @@ include LICENSE.PSF
 include README.rst
 
 include pyproject.toml
+recursive-include src py.typed
 
 recursive-include docs *
 recursive-include src/_cffi_src *.py *.c *.h


reaperhulk (approved):

intgr:
Note that this change regresses the precision of type checkers significantly, at least in my use case.

See for example https://github.com/pyca/cryptography/blob/b24d522b05b2220fbcb74363e454c2997a0c7ab7/src/cryptography/x509/extensions.py, there are almost no type hints in this file.

Compare that to the typeshed version https://github.com/python/typeshed/blob/master/stubs/cryptography/cryptography/x509/__init__.pyi

Note that a dozen or so people contributed to the cryptography stubs in typeshed, me included. I think cryptography upstream should be more considerate that this work isn't lost without consideration.

But now that this is done, what are the options going forward? I can only think of
* (1) revert addition of `py.typed` in cryptography
* (2) import hints from typeshed

Unfortunately option (2) seems like significant work and not a short-term solution. I understand that (1) isn't great either. Any other options that I didn't think of?

reaperhulk:
@intgr Thanks for pointing out that we missed this! We've started the typing work on that module in #5812 and will get it done ASAP. If there are other areas we've missed please open a new issue so we can track it!

PR #5804, opened 2021-02-09
Added a unit test to protect against import cycles

alex:


reaperhulk:
Does this catch the latest cycle we fixed?

alex:
SUre does.

tiran:
```
flake run-test: commands[3] | mypy src/cryptography/ vectors/cryptography_vectors/ tests/
tests/test_meta.py:18: error: Module has no attribute "__path__"; maybe "__author__"?  [attr-defined]
Found 1 error in 1 file (checked 185 source files)
ERROR: InvocationError for command /home/runner/work/cryptography/cryptography/.tox/flake/bin/mypy src/cryptography/ vectors/cryptography_vectors/ tests/ (exited with code 1)
```

alex:
Grumble.

webknjaz:
_/me was about to file that issue as requested in https://github.com/pyca/cryptography/issues/5794#issuecomment-776062029 but guess there's no need anymore_

I'll just extract the pointers to where this idea is coming from (originally mentioned in https://github.com/pyca/cryptography/issues/5794#issuecomment-776053456 + https://github.com/pyca/cryptography/issues/5794#issuecomment-776054297):

- https://github.com/sanitizers/octomachinery/blob/master/tests/circular_imports_test.py
- https://github.com/pytest-dev/pytest/blob/master/testing/test_meta.py
- twitter.com/codewithanthony/status/1229445110510735361

This test may be rather slow (proportionally to the number of files in checked) so it may be helpful to use it with pytest-xdist.

alex:
We already use pytest-xdist locally, and also pytest-shard on windows to keep CI time down. This seems to take about 10 seconds to run, so not too bad.

tests/test_meta.py
```diff
@@ -0,0 +1,38 @@
...
+
+    # When using pytest-cov it attempts to instrument subprocesses. This
+    # causes the memleak tests to raise exceptions.
+    # we don't need coverage so we remove the env vars.
+    env.pop("COV_CORE_CONFIG", None)
+    env.pop("COV_CORE_DATAFILE", None)
+    env.pop("COV_CORE_SOURCE", None)
+
+    for module in find_all_modules():
+        with subtests.test():
```
webknjaz, line 36: Have you considered using `@pytest.mark.parametrize`? Using it allows to load-balance the subtests across workers with plugins like `xdist`. Having a loop in the test itself will probably make the testing less responsive.

tests/test_meta.py
```diff
@@ -0,0 +1,38 @@
...
+def test_no_circular_imports(subtests):
+    env = os.environ.copy()
+    env["PYTHONPATH"] = os.pathsep.join(sys.path)
+
+    # When using pytest-cov it attempts to instrument subprocesses. This
+    # causes the memleak tests to raise exceptions.
+    # we don't need coverage so we remove the env vars.
+    env.pop("COV_CORE_CONFIG", None)
+    env.pop("COV_CORE_DATAFILE", None)
+    env.pop("COV_CORE_SOURCE", None)
```
webknjaz, line 33: Dunno if it's preferable but the same can be achieved with the built-in `monkeypatch` fixture: https://docs.pytest.org/en/latest/monkeypatch.html.

tests/test_meta.py
alex: It's a trade off: parameterize allows splitting with xdist, however it also makes collection more expensive. Because collection is done by each process you can end up with _worse_ performance. In general we've moved away from parameterize for this reason.

In any event, we have enough tests that this happening entirely serially doesn't matter, it just runs in parallel with a test of RSA or something. Given enough cores this would matter again, but we don't run our CI on a threadripper :-(

tests/test_meta.py
alex: We don't want to mutate the parent process's env, we only want to mess with the child processes.

tests/test_meta.py
```diff
@@ -0,0 +1,38 @@
...
+
+    for module in find_all_modules():
+        with subtests.test():
+            argv = [sys.executable, "-c", f"__import__({module!r})"]
```
webknjaz, line 37: Was it a conscious decision not to add `-W error`?

alex: Yes. The goal of the test in pytest was to check for warnings, that wasn't our goal here :-)

We in fact _do_ have modules that emit warnings.

webknjaz: > We in fact _do_ have modules that emit warnings.

This is why in my implementation `-W error` is followed by a specific `-W ignore` :) But I get why you want to opt-out of this check...

Maybe you'd consider summarizing the considerations in a docstring?

reaperhulk (approved):

tests/test_meta.py
reaperhulk: I'm going to merge this but if you'd like to add a comment to this effect I'd be happy to review that PR too!

PR #5809, opened 2021-02-10
Name: update get_attributes_for_oid return type

dhalperi:
`List` gives more power to the caller, and matches the same function for
`RelativeDistinguishedName`.

This is a follow-up to #5712 (cc: @alex / @reaperhulk).

---
This change is [Reviewable](https://reviewable.io/reviews/pyca/cryptography/5809)

### DIFF
--- a/src/cryptography/x509/name.py
+++ b/src/cryptography/x509/name.py
@@ -216,7 +216,7 @@ class Name(object):
             attr.rfc4514_string() for attr in reversed(self._attributes)
         )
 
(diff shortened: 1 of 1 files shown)


dhalperi:
Is there a reason this was `Iterable` only for `Name`? If we don't want to
promise `List`, `Sequence` is another alternative.

alex (approved):
I don't know why I did it this way.Thanks!

PR #5823, opened 2021-02-12
Bump pyo3 and lower MSRV

alex:


reaperhulk:
Alpine msg needs update. Also since this will merge to our primary docs we might want to hold on this until we're ready for 3.4.5 so we don't confuse people. We do version the docs but most people will be looking at main.

reaperhulk (approved):

PR #5824, opened 2021-02-13
Start replacing read_only_property with dedicated functions

alex:
read_only_property is basically impossible for mypy to check

### DIFF
--- a/src/cryptography/hazmat/primitives/asymmetric/ec.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/ec.py
@@ -326,10 +326,17 @@ _CURVE_TYPES: typing.Dict[str, typing.Type[EllipticCurve]] = {
 
 
 class ECDSA(EllipticCurveSignatureAlgorithm):
-    def __init__(self, algorithm):
+    def __init__(
+        self,
+        algorithm: typing.Union[asym_utils.Prehashed, hashes.HashAlgorithm],
+    ):
         self._algorithm = algorithm
 
-    algorithm = utils.read_only_property("_algorithm")
+    @property
+    def algorithm(
+        self,
+    ) -> typing.Union[asym_utils.Prehashed, hashes.HashAlgorithm]:
+        return self._algorithm
 
 
 def generate_private_key(
@@ -415,9 +422,9 @@ class EllipticCurvePublicNumbers(object):
         else:
             raise ValueError("Unsupported elliptic curve point type")
 
-    curve = utils.read_only_property("_curve")
-    x = utils.read_only_property("_x")
-    y = utils.read_only_property("_y")
+    curve = property(lambda self: self._curve)
+    x = property(lambda self: self._x)
+    y = property(lambda self: self._y)
 
     def __eq__(self, other):
         if not isinstance(other, EllipticCurvePublicNumbers):
@@ -463,8 +470,8 @@ class EllipticCurvePrivateNumbers(object):
         backend = _get_backend(backend)
         return backend.load_elliptic_curve_private_numbers(self)
 
-    private_value = utils.read_only_property("_private_value")
-    public_numbers = utils.read_only_property("_public_numbers")
+    private_value = property(lambda self: self._private_value)
+    public_numbers = property(lambda self: self._public_numbers)
 
     def __eq__(self, other):
         if not isinstance(other, EllipticCurvePrivateNumbers):
--- a/src/cryptography/hazmat/primitives/asymmetric/dh.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/dh.py
@@ -6,7 +6,6 @@
 import abc
 import typing
 
-from cryptography import utils
 from cryptography.hazmat.backends import _get_backend
 from cryptography.hazmat.primitives import serialization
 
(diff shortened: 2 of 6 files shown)


src/cryptography/hazmat/primitives/asymmetric/ec.py
```diff
@@ -326,10 +326,17 @@ class BrainpoolP512R1(EllipticCurve):
...
 class ECDSA(EllipticCurveSignatureAlgorithm):
-    def __init__(self, algorithm):
+    def __init__(
+        self,
+        algorithm: typing.Union[asym_utils.Prehashed, hashes.HashAlgorithm],
+    ):
         self._algorithm = algorithm
 
-    algorithm = utils.read_only_property("_algorithm")
+    @property
```
reaperhulk, line 335: Is type information lost here if you just do `algorithm = property(lambda self: self._algorithm)`?

alex: `mypy` complained that it didn't match the parent declaration, yeah.

reaperhulk: Could we get a comment here explaining the need for this extra boilerplate and pointing at either an existing or new mypy bug?

alex: It's not a mypy bug -- this implements an interface and so the property needs to match teh types on that interface.

It's a bug of `read_only_property` that it goes to `Any` and thus disables the type checking.

I don't see a need for a comment on what amounts to a straightforward function.

reaperhulk: Are you saying that using `property` as a non-decorator also always returns `Any` rather than the underlying type of the property? I understand that `read_only_property` did that since we wrapped the property declaration ourselves and therefore it implicitly returned `Any`. That's unfortunate.

alex: I'm saying that `read_only_property` was a whole `Any` thing. This effectively disabled mypy.

Then `property(lambda self: self.foo)` had no type signature so `mypy` was like "hey, your parent class has a type signature so you need one too".

reaperhulk: Ah, I see. It's a shame there's no way to wrap a value to be read only and proxy the type information. Oh well.

reaperhulk (approved):

ronf:
I ran into this issue recently, with properties such as the various  PrivateNumbers/PublicNumbers attributes for encryption keys. When I looked more closely at the implementation to see if I could add type hints, I found things like the following in RSAPrivateNumbers:

```python
    p = utils.read_only_property("_p")
    q = utils.read_only_property("_q")
    d = utils.read_only_property("_d")
    dmp1 = utils.read_only_property("_dmp1")
    dmq1 = utils.read_only_property("_dmq1")
    iqmp = utils.read_only_property("_iqmp")
    public_numbers = utils.read_only_property("_public_numbers")
```

As a result, I get coverage misses in mypy for all of these.

ronf:
I upgraded to cryptography main tonight and noticed that it has replaced all the references to read_only_property() with lambda functions. However, I'm still not getting proper coverage when accessing public/private numbers and I'm not sure why. I added some reveal_type() calls, and saw the following...

Here's the code I tested:

```python
from typing import TYPE_CHECKING

priv_key = rsa.generate_private_key(65537, 2048)
priv = priv_key.private_numbers()
pub = priv.public_numbers

if TYPE_CHECKING:
    reveal_type(priv_key)
    reveal_type(priv)
    reveal_type(pub)
    reveal_type(pub.e)
```

Here's the output I got:

```
t.py:9: note: Revealed type is "cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKey"
t.py:10: note: Revealed type is "cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateNumbers"
t.py:11: note: Revealed type is "Any"
t.py:12: note: Revealed type is "Any"
```

So, the signatures on generate_rsa_key() and private_numbers() are fine, but the use of property() for public_numbers() and the various other members of RSAPrivateNumbers and RSAPublicNumbers is not working here.

I'm using the latest version of pyca/cryptography@main for this testing, with Python 3.10.0b2 and mypy 0.910 on macOS Mojave. I compiled PyCA with version 1.53.0 of rust & cargo.

ronf:
I was able to confirm that the assignment to property() like this isn't something mypy handles properly with the following test code:

```python
from typing import TYPE_CHECKING

class Foo:
    def __init__(self, x: int):
        self._x = x

    x = property(lambda self: self._x)

foo = Foo(5)

if TYPE_CHECKING:
    reveal_type(foo.x)

print(foo.x)
```

Changing it to a normal decorator call did work, though:

```python
from typing import TYPE_CHECKING

class Foo:
    def __init__(self, x: int):
        self._x = x

    @property
    def x(self) -> int:
        return self._x

foo = Foo(5)

if TYPE_CHECKING:
    reveal_type(foo.x)

print(foo.x)
```

So, this would be one possible solution (at the expense of being a bit more verbose)...

A more compact (but kind of ugly) alternative is:

```python
    x = cast(int, property(lambda self: self._x))
```

I tested this in my local copy of rsa.py, and it did fix my coverage issue.

reaperhulk:
@ronf Sorry for the slow response here. I'd be okay with moving to just `@property` decorators on getter functions to resolve this. mypy is a harsh mistress.

ronf:
Thanks, Paul.

I know what you mean about mypy - I've been wrestling with it for weeks now as I try to annotate AsyncSSH...

PR #5825, opened 2021-02-13
change to a new version scheme

### LINKED ISSUE #5801: Version Scheme Change
We've talked on and off for quite a few years about our versioning scheme for `cryptography`, but #5771 made it very clear that despite our [versioning documentation](https://cryptography.io/en/latest/api-stability.html#versioning) some users still assume it's [SemVer](https://semver.org) and are caught off guard by changes. I propose that we **switch to a [CalVer](https://calver.org) scheme** for the thirty fifth feature release (which we currently would call 3.5). This has the advantage of being unambiguously not semantic versioning but encoding some vaguely useful information (the year of release) in it.

### Alternate Choices
**Stay the course**
We've successfully used our versioning scheme for a bit over 7 years now and the probability of another monumental shift like this is low. There is, however, a constant (but low) background radiation of people who are tripped up by making assumptions about our versioning so I would argue against this choice.

**SemVer**
Switch to using an actual semantic versioning scheme. Without re-litigating years of conversations, I don't believe either @alex or myself are interested in this option. (See https://github.com/pyca/cryptography/issues/5801#issuecomment-776067787 for a bit of elaboration on our position)

**Firefox/Chrome Versioning**
(From @alex)
"Its merits are that it's technically semver compatible, it's fairly predictable, and it communicates at least a little info on the relative age of the project.

Its demerits are that it probably won't solve the real problem here, which was that folks were surprised a major change happened at all."

alex:
fixes #5801

### DIFF
--- a/docs/api-stability.rst
+++ b/docs/api-stability.rst
@@ -52,7 +52,7 @@ Given a version ``cryptography X.Y.Z``,
   releases (such as fixing security issues or correcting regressions in a major
   release).
 
-This scheme is compatible with `SemVer`_, though many major releases will *not*
+This scheme is compatible with `SemVer`_, though many major releases will **not**
 include any backwards-incompatible changes.
 
 Deprecation
--- a/docs/api-stability.rst
+++ b/docs/api-stability.rst
@@ -52,8 +52,8 @@ Given a version ``cryptography X.Y.Z``,
   releases (such as fixing security issues or correcting regressions in a major
   release).
 
-This scheme is compatible with `SemVer`_, though many major releases will **not**
-include any backwards-incompatible changes.
+This scheme is compatible with `SemVer`_, though many major releases will
+**not** include any backwards-incompatible changes.
 
 Deprecation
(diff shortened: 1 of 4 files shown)


docs/api-stability.rst
```diff
@@ -37,35 +37,60 @@ policy as necessary in order to resolve a security issue or harden
...
-    The dividing decimal point can effectively be ignored.
+* ``X`` indicates the major version number. This is incremented on any feature
+  release.
+* ``Y`` is always ``0``.
+* ``Z`` is an integer that is incremented for minor backward-compatible
+  releases (such as fixing security issues or correcting regressions in a major
+  release).
 
-* ``Z`` is an integer that is incremented for backward-compatible releases.
+This scheme is compatible with `SemVer`_, though many major releases will *not*
```
reaperhulk, line 55:
```suggestion
This scheme is compatible with `SemVer`_, though many major releases will **not**
```

reaperhulk (approved):

PR #5831, opened 2021-02-14
make Extension a generic class (fixes #5830)

### LINKED ISSUE #5830: type hints: make Extension generic?
In the official typeshed, `cryptography.x509.Extension` is a [generic class](https://mypy.readthedocs.io/en/stable/generics.html), but in the type hinting added in 3.4, it isn't. I couldn't find a discussion about this, if there was I apologize for the duplicate. If there wasn't, anything speaking against making that change (I would create a PR).

Making Extension a generic class would allow more precise type hints, for a (very simplified, just to get the point across) example:

```
def get_pathlen(ext: x509.Extension[x509.BasicConstraints]) -> int:
    return ext.value.path_length  # this works now
```

Like I said, if you think this is a good idea, I can definitely create a PR for this!

mathiasertl:
That was simpler then expected (I was expecting more mypy errors).

One thing to note: For the `Extensions` (plural) class, typeshed does this:

```
    def __init__(self, general_names: List[Extension[Any]]) -> None: ...
```

while I do the slightly more specific:

```
def __init__(self, extensions: typing.List["Extension[ExtensionType]"]):
```

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -33,6 +33,8 @@ from cryptography.x509.oid import (
     ObjectIdentifier,
 )
 
+ExtensionTypeVar = typing.TypeVar("ExtensionTypeVar", bound="ExtensionType")
+
 
 def _key_identifier_from_public_key(public_key: _PUBLIC_KEY_TYPES) -> bytes:
     if isinstance(public_key, RSAPublicKey):
@@ -108,17 +110,19 @@ class ExtensionType(metaclass=abc.ABCMeta):
 
 
 class Extensions(object):
-    def __init__(self, extensions: typing.List["Extension"]):
+    def __init__(self, extensions: typing.List["Extension[ExtensionType]"]):
         self._extensions = extensions
 
-    def get_extension_for_oid(self, oid: ObjectIdentifier) -> "Extension":
+    def get_extension_for_oid(
+        self, oid: ObjectIdentifier
+    ) -> "Extension[ExtensionType]":
         for ext in self:
             if ext.oid == oid:
                 return ext
 
         raise ExtensionNotFound("No {} extension was found".format(oid), oid)
 
-    def get_extension_for_class(self, extclass) -> "Extension":
+    def get_extension_for_class(self, extclass) -> "Extension[ExtensionType]":
         if extclass is UnrecognizedExtension:
             raise TypeError(
                 "UnrecognizedExtension can't be used with "
@@ -1220,9 +1224,9 @@ class NameConstraints(ExtensionType):
     excluded_subtrees = utils.read_only_property("_excluded_subtrees")
 
 
-class Extension(object):
+class Extension(typing.Generic[ExtensionTypeVar]):
     def __init__(
-        self, oid: ObjectIdentifier, critical: bool, value: ExtensionType
+        self, oid: ObjectIdentifier, critical: bool, value: ExtensionTypeVar
     ):
         if not isinstance(oid, ObjectIdentifier):
             raise TypeError(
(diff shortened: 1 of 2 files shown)


src/cryptography/x509/extensions.py
```diff
@@ -108,17 +110,19 @@ def oid(self) -> ObjectIdentifier:
...
+        self, oid: ObjectIdentifier
+    ) -> "Extension[ExtensionType]":
         for ext in self:
             if ext.oid == oid:
                 return ext
 
         raise ExtensionNotFound("No {} extension was found".format(oid), oid)
 
-    def get_extension_for_class(self, extclass) -> "Extension":
+    def get_extension_for_class(self, extclass) -> "Extension[ExtensionType]":
```
alex, line 125: Is there any way to type this so that it's known that `ExtensionType` here matches `extclass`?

mathiasertl: YES that is possible, like this:

```diff
diff --git a/src/cryptography/x509/extensions.py b/src/cryptography/x509/extensions.py
index 2858cef9..56f8b367 100644
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -122,7 +122,9 @@ class Extensions(object):
 
         raise ExtensionNotFound("No {} extension was found".format(oid), oid)
 
-    def get_extension_for_class(self, extclass) -> "Extension[ExtensionType]":
+    def get_extension_for_class(
+        self, extclass: typing.Type[ExtensionType]
+    ) -> "Extension[ExtensionType]":
         if extclass is UnrecognizedExtension:
             raise TypeError(
                 "UnrecognizedExtension can't be used with "
```

but this makes mypy reveal another issue: 

1. `ExtensionType.oid` is defined as abstract property. Meaning for example `BasicConstraints(...).oid` works, but `BasicConstraints.oid` does **not**.
2. That's of course not actually true, as every subclass defines `oid` as a class attribute.
3. However `ExtensionNotFound` raised at the end of `get_extension_for_class` accesses it as a class attribute, triggering a mypy error.

Since the issue is not related to the change done by this PR, the function argument was not annotated before this PR either, and annotating it would reveal this issue even without my PR, I propose we ignore this issue here.

mathiasertl: PS: python has no `@classproperty` decorator, neither does the `abc` module. The solution I found was defining oid like this:

```diff
diff --git a/src/cryptography/x509/extensions.py b/src/cryptography/x509/extensions.py
index 91e9a30d..ec4ae9fb 100644
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -102,11 +102,10 @@ class ExtensionNotFound(Exception):
 
 
 class ExtensionType(metaclass=abc.ABCMeta):
-    @abc.abstractproperty
-    def oid(self) -> ObjectIdentifier:
-        """
-        Returns the oid associated with the given extension type.
-        """
+    oid: typing.ClassVar[ObjectIdentifier]
+    """
+    Returns the oid associated with the given extension type.
+    """
 
 
 class Extensions(object):
```

... that works (test suite passes too). BUT since ExtensionType no longer has any abstract property (but maybe we could add/declare value to be an abstract property?), you now can actually create an ExtensionType instance.

mathiasertl: Thinking about this more closely, the diff from my initial comment won't do what you requested, it would only if this were a generic type parameter. I couldn't find anything to do that with concrete classes.

Not that the parameter shouldn't be annotated, but again I would do that in a separate PR, as it requires larger changes.

mathiasertl:
Anything wrong here? I'd love to have this in 3.5 so I can use type hinting properly.

alex (approved):

Issue #4468, opened 2018-09-13
cryptography for SM2


MasterMeng:
I want to add SM2 to cryptography for my work,but I can not found the API for add user-defined ECC

MasterMeng:
Thank you for your answer.
In our company, we need to build a CA system for our future project,and it needs support SM2 for some customers.

PR #5833, opened 2021-02-15
Add support for SM3 hash algorithm

tobyp:
This PR includes python classes for SM3, as well as test vectors and documentation.

SM3 is supported in [OpenSSL 1.1.1](https://www.openssl.org/docs/man1.1.0/man3/).

This is my first contribution, please let me know if anything requires further work :)


reaperhulk:
Thanks for the contribution! We've chosen not to add support for this previously while waiting for someone to provide a use case, but it seems likely your organization has one. Is it WAPI or something else?

tobyp:
We use the cryptography library for internal testing and validation because it is much easier than using OpenSSL directly. We particularly want to test SM3 and SM4, so we implemented these patches as part of our internal system.

We wanted to share this contribution with the community nevertheless, as there is a certain trend using these ciphers, WAPI being one of them.

reaperhulk:
Are there any other public specs you can share that need SM3/SM4? It’s extremely useful to us to know where these primitives are required so any detail you can provide would be great.

This PR also needs a rebase at this point.

tobyp:
SM3 in particular is most often used together with the SM2 signature and encryption schemes, where it is the default hash algorithm. SM2/3/4 are also sometimes mandatory for products to be distributed in China.

Besides WAPI and our own, I'm afraid I can't point out other concrete use cases, especially since I can't read Chinese and am limited to the RFC documents and English materials. There is however quite a bit of material about other products gaining SM2/3/4 capabilities (e.g. the kernel, libgcrypt and botan, as well as some crypto IP cores), so the trend is visible.

tobyp:
Though there is of course the past use case in #4468.

docs/hazmat/primitives/cryptographic-hashes.rst
```diff
@@ -245,6 +245,19 @@ MD5
     message digest and has practical known collision attacks.
 
 
+SM3
+~~~
+
+.. class:: SM3()
+
+    .. versionadded:: 3.5
```
reaperhulk, line 253:
```suggestion
    .. versionadded:: 35.0.0
```

reaperhulk (changes requested):

reaperhulk (approved):

PR #5834, opened 2021-02-15
Add support for SM4 symmetric block cipher

tobyp:
This PR includes python classes for SM4, as well as test vectors and documentation.

SM4 is supported in [OpenSSL 1.1.1](https://www.openssl.org/docs/man1.1.1/man3/EVP_sm4_ecb.html).

See also #5833.

### DIFF
--- a/src/cryptography/hazmat/primitives/ciphers/algorithms.py
+++ b/src/cryptography/hazmat/primitives/ciphers/algorithms.py
@@ -153,3 +153,16 @@ class ChaCha20(CipherAlgorithm, ModeWithNonce):
     @property
     def key_size(self) -> int:
         return len(self.key) * 8
+
+
+class SM4(CipherAlgorithm, BlockCipherAlgorithm):
+    name = "SM4"
+    block_size = 128
+    key_sizes = frozenset([128])
+
+    def __init__(self, key: bytes):
+        self.key = _verify_key_size(self, key)
+
+    @property
+    def key_size(self) -> int:
+        return len(self.key) * 8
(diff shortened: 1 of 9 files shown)


tests/hazmat/primitives/test_sm4.py
```diff
@@ -0,0 +1,101 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+from __future__ import absolute_import, division, print_function
```
reaperhulk, line 5: This import isn't required now that we only support 3.6+

reaperhulk (changes requested):

src/cryptography/hazmat/primitives/ciphers/algorithms.py
```diff
@@ -153,3 +153,16 @@ def nonce(self) -> bytes:
...
+class SM4(CipherAlgorithm, BlockCipherAlgorithm):
+    name = "SM4"
+    block_size = 128
+    key_sizes = frozenset([128])
+
+    def __init__(self, key):
+        self.key = _verify_key_size(self, key)
+
+    @property
+    def key_size(self):
```
reaperhulk, line 167:
```suggestion
    def key_size(self) -> int:
```

tobyp: Fixed, and the `__init__` parameters too.

reaperhulk (approved):

PR #5846, opened 2021-02-18
Strict is deprecated

reaperhulk:


tox.ini
```diff
@@ -14,7 +14,7 @@ deps =
 passenv = ARCHFLAGS LDFLAGS CFLAGS INCLUDE LIB LD_LIBRARY_PATH USERNAME PYTHONIOENCODING OPENSSL_FORCE_FIPS_MODE
 commands =
     pip list
-    pytest -n auto --cov=cryptography --cov=tests --capture=no --strict --durations=10 {posargs}
+    pytest -n auto --cov=cryptography --cov=tests --capture=no --strict-markers --durations=10 {posargs}
```
alex, line 17: What version of pytest was `strict-markers` added in, we should probably set that as the minimum.

reaperhulk: Looks like this happened with 6.2, which also dropped support for Python less than 3.6. Fortunately that’s fine for us

alex: is this by way of saying you'll change our setup.py, or you don't want to because python_requires already makes thie always true?

reaperhulk: I’ll change it when I’ve got a full computer available 😀

alex (approved):

PR #5851, opened 2021-02-20
Fix typo in error messages: can not => cannot

rbrtmrtn:
Hello,

First of all, thank you for all your hard work on this indispensable library! 😄 

This is a very minor PR, but just caught a few typos where the words "can not" should be "cannot". This was in a few error messages, and one code comment.

There was one more error I couldn't find, which was "Can not find Rust compiler". If someone happens to know where that message is located, I'd be happy to add it to the PR.

Thank you!


alex (approved):
thanks

PR #5852, opened 2021-02-20
add ObjectIdentifier.name property

mathiasertl:
Add a `name` property to ObjectIdentifiers.

I really don't see why this is just a private property. Both `ObjectIdentifier._name` and `cryptography.x509.oid._OID_NAMES` are private. This makes it impossible to display your own representation of an ObjectID without accessing private APIs.

I do propose leaving the old property in place, as lots of code probably still uses it, including my own code.

PS: Is the changelog really accurate? do you intend to switch to a new versioning scheme in the next release? See Version 35 vs. 3.5.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,8 @@ Changelog
 
  .. note:: This version is not yet released and is under active development.
 
+* Add a ``name`` property to :class:`cryptography.x509.ObjectIdentifier`.
+
 .. _v3-4-6:
 
 3.4.6 - 2021-02-16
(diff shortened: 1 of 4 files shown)


alex:
It's intentional that this attribute is not public -- it's because these names are not canonical and we don't want to commit to them as a public API surface. They exist for `repr()` debugging only.

Thanks for your contribution though!

mathiasertl:
Aha okay. I was arguing that I want to do a different representation, but okay.

Is it at least okay to typehint the property? Otherwise mypy will complain when using it.

alex:
Sure, a type hint on the existing thing would be fine.

On Sun, Feb 21, 2021 at 1:03 PM Mathias Ertl <notifications@github.com> wrote:
>
> Aha okay. I was arguing that I want to do a different representation, but okay.
>
> Is it at least okay to typehint the property? Otherwise mypy will complain when using it.
>
> —
> You are receiving this because you modified the open/close state.



-- 

PR #5856, opened 2021-02-21
add typehint for ObjectIdentifier._name

mathiasertl:
As discussed in #5852, at least add a typehint to `_name`.

Question: I did not want to add this here without asking, but is it wanted to add typehints to other private functions (such as `__repr__`) as well? mypy would warn about these in strict mode. If yes, I can add them to this PR, at least for this module (and would volunteer to add them elsewhere, too).


reaperhulk:
It is our goal to slowly turn on the stricter type checks in mypy, so typehinting repr, etc are acceptable, yes.

alex (approved):

PR #5857, opened 2021-02-21
test pypy3.6 and pypy3.7

reaperhulk:


reaperhulk:
For reference we previously just tested `pypy3` which was equivalent to `pypy-3.6`

alex (approved):

PR #5868, opened 2021-02-26
Remove unused X509 verification flags bindings

davidben:
These don't appear to be used in pyopenssl or cryptography.io. One less source of conditionals.

### DIFF
--- a/src/_cffi_src/openssl/x509_vfy.py
+++ b/src/_cffi_src/openssl/x509_vfy.py
@@ -18,7 +18,6 @@ typedef STACK_OF(X509_OBJECT) Cryptography_STACK_OF_X509_OBJECT;
 """
 
 TYPES = """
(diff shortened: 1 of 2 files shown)


alex (approved):
Will leave this open until @reaperhulk double checks me on if there's a user

alex (dismissed):
Ooops, these need to eb removed from the other place they're present in that file, around L103 or so.

davidben:
Oops, not sure how I missed that. Rebase mishap, maybe. Fixed, hopefully.

reaperhulk:
pyOpenSSL definitely never used these so I'm comfortable with removing them.

reaperhulk (approved):

reaperhulk:
To be clear I'm unaware of any consumers of this, but I'm very sure pyOpenSSL has never used these flags since I checked the entire git history.

PR #5870, opened 2021-02-26
Strict typehints for extensions and OIDs

mathiasertl:
Attempt to completely typehint extensions and Object Identifier modules.

This is a WIP and I'm seeking feedback on your preferences on a couple of issues. Please see comments to specific LOC I will to in a second...


src/cryptography/hazmat/_oid.py
```diff
@@ -2,9 +2,11 @@
 # 2.0, and the BSD License. See the LICENSE file in the root of this repository
 # for complete details.
 
+import typing
+
 
 class ObjectIdentifier(object):
-    def __init__(self, dotted_string: str):
+    def __init__(self, dotted_string: str) -> None:
```
mathiasertl, line 9: The [mypy cheatcheet|https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html] consistently annotates the return type of constructors, so I'm doing here as well. If you don't want this, please just say so.

src/cryptography/hazmat/_oid.py
```diff
@@ -46,21 +48,21 @@ def __init__(self, dotted_string: str):
                 % (self._dotted_string)
             )
 
-    def __eq__(self, other):
+    def __eq__(self, other: typing.Any) -> bool:
```
mathiasertl, line 51: If I don't annotate with `Any`, mypy will complain, yet the exception below still seems odd. I don't know of any other class where comparing to an instance of a _different_ class raises an exception. Why not just return `False` (same in many other places)?

src/cryptography/x509/extensions.py
```diff
@@ -76,8 +75,8 @@ def _key_identifier_from_public_key(public_key: _PUBLIC_KEY_TYPES) -> bytes:
     return hashlib.sha1(data).digest()
 
 
-def _make_sequence_methods(field_name):
-    def len_method(self):
+def _make_sequence_methods(field_name: str):
+    def len_method(self) -> int:
```
mathiasertl, line 79: This function is basically untypable, I think. Open to suggestions, of course.

Note that here `self` is not even bound, so even that doesn't check out. `iter_method` below would have to return `Iterator[WHAT?]`. And `_make_sequence_methods` would have to be returned with several callables, all with a type.

This could be done with a generic typevar. But this will add a lot of `typing.TypeVar` variables. Is this really what we want? Alternatively, we can also just implement this function, "the long way around", like I did with `Extensions` below.

src/cryptography/x509/extensions.py
```diff
@@ -134,38 +137,60 @@ def get_extension_for_class(self, extclass) -> "Extension[ExtensionType]":
             if isinstance(ext.value, extclass):
                 return ext
 
+        oid = typing.cast(ObjectIdentifier, extclass.oid)
```
mathiasertl, line 140: Same issue here as in #5830: ExtensionType.oid is an abstract property (= an abstract _instance_ property), but of course all classes define a class attribute. Should we replace with 

```
class ExtensionType:
     oid: ObjectIdentifier
```

?

src/cryptography/x509/extensions.py
```diff
@@ -1664,14 +1711,14 @@ class IssuingDistributionPoint(ExtensionType):
...
         self,
-        full_name,
-        relative_name,
-        only_contains_user_certs,
-        only_contains_ca_certs,
-        only_some_reasons,
-        indirect_crl,
-        only_contains_attribute_certs,
-    ):
+        full_name: typing.Optional[typing.List[GeneralName]],
```
mathiasertl, line 1714: Special note here: We never cast full_name to a list (as we do in many places). So you really should pass a list - at present. Should we do `list(full_name) and use `Iterable`  instead?

src/cryptography/x509/oid.py
```diff
@@ -4,7 +4,7 @@
 
 import typing
 
-from cryptography.hazmat._oid import ObjectIdentifier
+from cryptography.hazmat._oid import ObjectIdentifier as ObjectIdentifier
```
mathiasertl, line 7: See https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport - seems super-redundant here of course.

src/cryptography/x509/extensions.py
alex: Will doing `oid: ObjectIdentifier` work with the fact that it's a classvar on most `ExtensionType` implementors, but an instancevar on `UnknownExtension`?

src/cryptography/x509/extensions.py
```diff
@@ -134,38 +137,60 @@ def get_extension_for_class(self, extclass) -> "Extension[ExtensionType]":
...
                 return ext
 
+        oid = typing.cast(ObjectIdentifier, extclass.oid)
         raise ExtensionNotFound(
-            "No {} extension was found".format(extclass), extclass.oid
+            "No {} extension was found".format(extclass), oid
         )
 
-    __len__, __iter__, __getitem__ = _make_sequence_methods("_extensions")
+    @typing.overload
```
alex, line 145: Let's hold off on making this change. Everything else in this PR is unobjectionable and I'm happy to merge, but changing all of our extension types to be implemented like this is something I think we should hold off on.

src/cryptography/x509/oid.py
alex: I kind of hate this, can we not include it in this PR

alex (commented):
95% of this PR looks great -- would love to get those pieces merged ASAP and then we can mull through some of the harder cases

src/cryptography/x509/oid.py
mathiasertl: I think we can all agree that this looks very, very, very ugly.

For reference, here's the warning:

```
$ mypy --strict src/cryptography/x509/extensions.py 
src/cryptography/x509/extensions.py:28: error: Module 'cryptography.x509.oid' does not explicitly export attribute 'ObjectIdentifier'; implicit reexport disabled  [attr-defined]
```

I removed this in this PR, but one way or another this has to be solved - see the link I posted above.

src/cryptography/x509/extensions.py
mathiasertl: Not sure what _exactly_ you want reverted here? "manually" implementing the sequence methods? Just the overload?

alex: Stick with `_make_sequence_methods` and then a seperate PR to discuss these changes.

src/cryptography/x509/extensions.py
mathiasertl: Just tested this, and it seems so, at least from a mypy point of view, here's the diff I tested:

```diff
 class ExtensionType(metaclass=abc.ABCMeta):
-    @abc.abstractproperty
-    def oid(self) -> ObjectIdentifier:
-        """
-        Returns the oid associated with the given extension type.
-        """
+    oid: ObjectIdentifier
```

HOWEVER I must point out - from [abc](https://docs.python.org/3/library/abc.html):

>  A class that has a metaclass derived from ABCMeta cannot be instantiated unless all of its abstract methods and properties are overridden.

That means, the change would mean you can instantiate an ExtensionType, if you want to:

```
>>> x509.ExtensionType()
<cryptography.x509.extensions.ExtensionType object at 0x...>
```

alex: Unfortunate, but not a deal breaker. I guess it's do it!

src/cryptography/x509/extensions.py
mathiasertl: done!

mathiasertl:
Alright, I made oid a property and removed the "manual" implementation of sequence methods. Only thing you might still want to consider IMHO is this: https://github.com/pyca/cryptography/pull/5870#discussion_r583831414

Here's the current output of mypy, with comments from me as to why they're there:

```
$ mypy --strict src/cryptography/x509/extensions.py 
# what mypy calls an "implicit reexport", see comments above
src/cryptography/x509/extensions.py:28: error: Module 'cryptography.x509.oid' does not explicitly export attribute 'ObjectIdentifier'; implicit reexport disabled  [attr-defined]

# function in another module
src/cryptography/x509/extensions.py:56: error: Call to untyped function "DERReader" in typed context  [no-untyped-call]

# these and following are for _make_sequence_methods
src/cryptography/x509/extensions.py:78: error: Function is missing a return type annotation  [no-untyped-def]
src/cryptography/x509/extensions.py:79: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
src/cryptography/x509/extensions.py:82: error: Function is missing a return type annotation  [no-untyped-def]
src/cryptography/x509/extensions.py:85: error: Function is missing a type annotation  [no-untyped-def]

# These two are because __iter__ cannot be annotated. looping through self._extensions would also solve this warning:
src/cryptography/x509/extensions.py:118: error: Returning Any from function declared to return "Extension[ExtensionType]"  [no-any-return]
src/cryptography/x509/extensions.py:134: error: Returning Any from function declared to return "Extension[ExtensionType]"  [no-any-return]
```

mathiasertl:
I love how mypy shows an error, but does not show the same error with `--strict`. :facepalm: 

```
$ mypy  src/cryptography/x509/extensions.py 
src/cryptography/x509/extensions.py:1848: error: Signature of "oid" incompatible with supertype "ExtensionType"  [override]
```

Anyway, there's a warning here, and I'm unsure how to solve it. Strictly from an OO standpoint, we do have a violation with this change, as the definition in `ExtensionType` implies that you can update/delete the attribute. Even adding `oid: typing.ClassVar[...]` doesn't change the mypy warning. I've now done so anyway, as it's more accurate for all extensions _except_ UnrecognizedExtension.

Unsure what to do - please advise:

* If we revert that change, we have to add a typecast/ignore when accessing the oid property of an ExtensionType class.
* If we leave it as is, we also have to the same, but in the UnrecognizedExtension class.

My gut feeling would be that it's "cleaner" to add an override in the "this is an error and should not happen class" thas is UnrecognizedExtension, instead of the "normal" Extensions class. But of course, it's up to you.

alex:
Mmm, yes my instinct is also that `UnrecognizedExtension` is basically special and should get treated special. @reaperhulk do you have a thought?

mathiasertl:
Alright, added type ignore to `UnrecognizedExtension`. If @reaperhulk agrees, it could be merged, I think.

alex (approved):

src/cryptography/x509/extensions.py
reaperhulk: We should cast and use iterable! If mypy is helping us see inconsistency let's stamp it out 😄

reaperhulk:
One comment that isn't a blocker (we can do the iterable/cast in a separate PR). I'm okay with the contortion for `oid` in `UnrecognizedExtension`.

alex:
I'm going to merge as-is, let's do the iterable change in a follow up PR.

PR #5882, opened 2021-02-28
changes to support typing backend

reaperhulk:
This gets rid of the conditional backend type registration. We could therefore delete `register_interface_if` but I'm a bit concerned it might be used elsewhere. Perhaps we should deprecate it in a separate PR.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -161,22 +161,21 @@ class _RC2(object):
     pass
 
 
-@utils.register_interface(CipherBackend)
-@utils.register_interface(CMACBackend)
-@utils.register_interface(DERSerializationBackend)
-@utils.register_interface(DHBackend)
-@utils.register_interface(DSABackend)
-@utils.register_interface(EllipticCurveBackend)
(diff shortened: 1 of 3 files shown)


reaperhulk:
(These should maybe become protocols in the future but going for least invasive changes for now)

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -161,22 +161,21 @@ class _RC2(object):
...
-@utils.register_interface(EllipticCurveBackend)
-@utils.register_interface(HashBackend)
-@utils.register_interface(HMACBackend)
-@utils.register_interface(PBKDF2HMACBackend)
-@utils.register_interface(RSABackend)
-@utils.register_interface(PEMSerializationBackend)
-@utils.register_interface(X509Backend)
-@utils.register_interface_if(
-    binding.Binding().lib.Cryptography_HAS_SCRYPT, ScryptBackend
-)
```
alex, line 178: You can delete `register_interface_if` now

reaperhulk: Read initial PR comment plz

alex (approved):

PR #5883, opened 2021-03-01
Delete unused register_interface_if

alex:


reaperhulk (approved):
one day codecov will be good. today is not that day

PR #5887, opened 2021-03-02
more typing

reaperhulk:
This adds a bunch of typing for backend args. To do this it needed to define interfaces. Both ones that were incomplete and ones entirely missing.

We briefly talked about this in IRC but do we want to move from `PKCS12Backend` and `PKCS7Backend` to just `BaseBackend` and stick all future methods on that for now?

All of this helps get us closer to `disallow_incomplete_defs = True`.

### DIFF
--- a/src/cryptography/hazmat/backends/__init__.py
+++ b/src/cryptography/hazmat/backends/__init__.py
@@ -4,7 +4,9 @@
 
 import typing
 
-_default_backend: typing.Any = None
+from cryptography.hazmat.backends.interfaces import Backend
+
+_default_backend: typing.Optional[Backend] = None
 
 
 def default_backend():
@@ -18,10 +20,7 @@ def default_backend():
     return _default_backend
 
 
-T = typing.TypeVar("T")
-
-
-def _get_backend(backend: typing.Optional[T]) -> T:
+def _get_backend(backend: typing.Optional[Backend]) -> Backend:
     if backend is None:
         return default_backend()
     else:
--- a/src/cryptography/hazmat/backends/__init__.py
+++ b/src/cryptography/hazmat/backends/__init__.py
@@ -18,7 +18,10 @@ def default_backend():
     return _default_backend
 
 
-def _get_backend(backend):
+T = typing.TypeVar("T")
+
+
+def _get_backend(backend: typing.Optional[T]) -> T:
     if backend is None:
         return default_backend()
     else:
--- a/src/cryptography/hazmat/backends/interfaces.py
+++ b/src/cryptography/hazmat/backends/interfaces.py
@@ -317,6 +317,18 @@ class X509Backend(metaclass=abc.ABCMeta):
         Compute the DER encoded bytes of an X509 Name object.
         """
 
+    @abc.abstractmethod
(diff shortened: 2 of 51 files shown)


src/cryptography/hazmat/primitives/twofactor/hotp.py
```diff
@@ -24,7 +24,7 @@ def _generate_uri(
     type_name: str,
     account_name: str,
     issuer: typing.Optional[str],
-    extra_parameters,
+    extra_parameters: typing.List,
```
alex, line 27: Make this list type concrete.

alex (commented):
Yes, I think we should go with BaseBackend (or something even better named) rather thna adding the pkcs12 and 7 backends.

reaperhulk:
Sigh, can't use Protocol actually. Forgot that was 3.8+

alex:
> 2021-03-02T06:21:33.4386574Z ./src/cryptography/hazmat/primitives/ciphers/modes.py:10:1: F401 'cryptography.hazmat.primitives._cipheralgorithm.BlockCipherAlgorithm' imported but unused

src/cryptography/hazmat/backends/interfaces.py
```diff
@@ -317,6 +317,18 @@ def x509_name_bytes(self, name):
         Compute the DER encoded bytes of an X509 Name object.
         """
 
+    @abc.abstractmethod
+    def load_pem_x509_crl(self, data):
```
alex, line 321:
```suggestion
    def load_pem_x509_crl(self, data: bytes) -> "CertificateRevocationList":
```

src/cryptography/hazmat/backends/interfaces.py
```diff
@@ -317,6 +317,18 @@ def x509_name_bytes(self, name):
...
+        """
+        Load an X.509 CRL from PEM encoded data.
+        """
+
+    @abc.abstractmethod
+    def load_der_x509_crl(self, data):
```
alex, line 327: Same

src/cryptography/hazmat/backends/__init__.py
```diff
@@ -18,7 +18,10 @@ def default_backend():
     return _default_backend
 
 
-def _get_backend(backend):
+T = typing.TypeVar("T")
+
+
+def _get_backend(backend: typing.Optional[T]) -> T:
```
alex, line 24: Can this just be `typing.Optional[Backend] -> Backend` now?

src/cryptography/hazmat/backends/interfaces.py
reaperhulk: Doing this will cause an import loop so I'd prefer figuring out how to type it in a separate PR :)

src/cryptography/hazmat/backends/__init__.py
reaperhulk: Yeah that should be fine now since we're not unwrapping subtypes.

alex: Great, let's do that. Should also remove the need to type a bunch of local variables.

alex (approved):

PR #5894, opened 2021-03-04
[RFC/WIP] Add SM2 Elliptic Curve Signature and Encryption Algorithms

tobyp:
This PR is very much a [Request for Comments], not least because the test vectors are still WIP. I would like to discuss how SM2 could be contributed to cryptography unobtrusively, for compatibility purposes similar to the previous SM3/4 PRs [#5833 #5834]

The SM2 implementation in OpenSSL is a little unusual in 3 ways:
- It uses PKEY "alias types" (`EC_PKEY_set_alias_type` to `EVP_PKEY_SM2`). This is currently toggled during `_ec_sig_sign`, but could also be moved to the `EllipticCurve(Public|Private)Key` classes. This is a point worth discussing.
- The signature scheme prepends an extra hash to the message, over an extra "User ID" and some public key/curve parameters. This extra user ID is implemented as a field in the `SM2Sign` `EllipticCurveSignatureAlgorithm`, and to support this, the signature algorithm and public key are passed further down the call stack than before during signing/verifying. As far as I can tell this only touches private (underscore names) APIs.
- OpenSSL calculates this hash itself if `EVP_MD_DigestVerify` is used, but does *not* calculate it if `EVP_PKEY_sign` is used. The former precludes `Prehashed` though. I've avoided duplicating code paths here and implemented a `sm2_z_hash` utility function so `EVP_PKEY_sign` can be used exclusively. This z hash does not include any secret information.

Note this PR rewrites the EC signature generation/verification to use EVP_PKEY APIs, instead of the (deprecated) ECDSA APIs used currently. This might prove useful going towards OpenSSL 3.0, and makes the EC code more like RSA.

Concerning the test vectors, the official standards unfortunately use an older curve than the one that ended up standardized as SM2P256V1. The included set of vectors is generated using BouncyCastle C#, but the code for this includes large amounts of code copied from the original BC test suite (to make RNG output deterministic), which has a hybrid MIT/X license. I intend to generate vectors using botan instead and include that code, similar to how IDEA used botan, and then verify them with BC for good measure.

As an aside, it seems SM2Encrypt would be the first EC public key encryption scheme supported in cryptography. I hope to explore the state of ECIES support in OpenSSL and add it to cryptography as well, once this PR is complete.

Please let me know about any questions, concerns or comments for this PR while I continue to work on this. The WIP parts


reaperhulk:
There's a lot to take in here 😄 

We do have a 3.0 compatibility PR (#5250) which I've been keeping up-to-date with OpenSSL alphas (as well as using to build various test cases to file bugs). I'm not opposed to moving to using the EVP API, although it'd be interesting to see how painful that will be as long as we retain LibreSSL compatibility. I think if we want to do that we should split that effort out and get that landed separately.

For SM2 I'd prefer to just look at SM2Sign first and leave SM2Encrypt for a later ECIES discussion. And with SM2Sign we should talk about API designs a bit.

When we originally designed the EC API we made the actual signature algorithm pluggable on the assumption that a common EC interface was desirable. However, later curves (like curve25519) and signature algorithms like EdDSA caused us to revisit that decision. Given that SM2 signing isn't supported for other curves and requires a specific hash algorithm does it make sense to shoehorn it into our existing ec signature interface or should it have its own like the Ed25519 and Ed448 classes do?

tobyp:
> There's a lot to take in here :smile:

Indeed! So splitting it up sounds like an excellent idea. Splitting off SM2Encrypt for later is probably for the best, I'll move that to a branch for later.

Yes, the API is quite shoe-horned here, mainly to support `Prehashed`. Now that I've had a few days away from this code, I suspect this is actually a big footgun: it's inevitable that someone will eventually forget to prepend the Z-hash and end up with non-standard signatures (essientially ECDSA on the SM2 curve). Ed25519 and the others don't seem to support `Prehashed` either, and have a much cleaner API for it. Without that, SM2Sign could be a copy-paste of Ed25519, and to be honest that "set-and-restore" of the key type around the signing operation makes me uncomfortable. Side note: this would probably mean SM2Encrypt has to go in the new class as well, so there wouldn't be much to coordinate with ECIES.

How does that sound?

Also, removing `Prehashed` means `sm2_z_hash` and `elliptic_curve_parameters` could go. Would the latter be useful to keep?

reaperhulk:
I agree, let's go down the Ed25519 style path. We can look at signing/verification first and we don't need to support `Prehashed`. Let's keep this simple and drop `elliptic_curve_parameters` as well.

tobyp:
While rewriting SM2 to be more like the Ed and X curve keys, I ran into a few problems with serialization - namely that `EVP_PKEY_new|get_raw_private|public_key` don't support SM2. There is also no clear convention/default for serialization format, as far as I can find. SM2 keys are basically EVP_PKEY_EC, with an EVP_PKEY_SM2 alias type (which `PEM_read_bio_PrivateKey` did not automatically detect from the curve is a small experiment I did, so I'll try a cleaner test later).

I see a few options here:
- Ed/X-style `from_private_bytes` etc. with only one specific encoding. This would be an arbitrary constraint.
- Detect the curve and switch the key type in `serialization.load_*_key`. Prevents ECDSA/sm2p256v1 (if anyone wanted to use that...) but complicates compatibility (e.g. with 1.1.0 or LibreSSL). This seems to be similar to something OpenSSL discussed in [openssl#8435](https://github.com/openssl/openssl/issues/8435)
- An extra optional/keyword parameter on `serialization.load_*_key` could also facilitate an `EVP_PKEY_set_alias_type` with no further changes
- Give `SM2*Key` their own variants of `serialization.load_*_key` and share the EC import code
- Copy the EC import machinery, which is quite a lot of code.

Whichever path we take, it's probably always going to be possible to load a seralized SM2 key as a normal EC key and end up doing ECDSA with it because this technically *is* allowed.

I'm pushing a draft with the first alternative (using DER PKCS8) so I can continue with tests, but I'm not really happy with it and would welcome some discussion. I'd also like to take some time to read [openssl#8435](https://github.com/openssl/openssl/issues/8435) more carefully to make sure everything is up to date. I'm also in freenode#cryptography-dev (but unfortunately with quite some latency).

reaperhulk:
Closing due to lack of movement, but I'll be happy to review still if this gets dusted off again 😄

PR #5897, opened 2021-03-06
Generic extension value/typehint x509.Name

mathiasertl:
First, Make `Extension.value` return a generic value. This way mypy detects what concrete type the value is, not just the more abstract base class. Consider for example this code:

```python
def example(ext: x509.Extension[x509.FreshestCRL]) -> x509.FreshestCRL:
    return ext.value
```

*Without* the patch, the function would have to use `x509.ExtensionType` as return type annotation.

Second, typehint the `x509.Name()`. The `typing.overload` decorator allows mypy to know that the list is either all NameAttribute or all RelativeDistinguishedName, not mixed, see [Function overloading](https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading).

### DIFF
--- a/src/cryptography/x509/name.py
+++ b/src/cryptography/x509/name.py
@@ -196,14 +196,32 @@ class RelativeDistinguishedName(object):
 
 
 class Name(object):
-    def __init__(self, attributes):
+    @typing.overload
+    def __init__(self, attributes: typing.Iterable[NameAttribute]) -> None:
+        ...
+
+    @typing.overload
+    def __init__(
+        self, attributes: typing.Iterable[RelativeDistinguishedName]
+    ) -> None:
+        ...
+
+    def __init__(
+        self,
+        attributes: typing.Iterable[
+            typing.Union[NameAttribute, RelativeDistinguishedName]
+        ],
+    ) -> None:
         attributes = list(attributes)
         if all(isinstance(x, NameAttribute) for x in attributes):
             self._attributes = [
-                RelativeDistinguishedName([x]) for x in attributes
+                RelativeDistinguishedName([typing.cast(NameAttribute, x)])
+                for x in attributes
             ]
         elif all(isinstance(x, RelativeDistinguishedName) for x in attributes):
-            self._attributes = attributes
+            self._attributes = typing.cast(
+                typing.List[RelativeDistinguishedName], attributes
+            )
         else:
             raise TypeError(
                 "attributes must be a list of NameAttribute"
--- a/.coveragerc
+++ b/.coveragerc
@@ -14,3 +14,4 @@ source =
 exclude_lines =
     @abc.abstractmethod
     @abc.abstractproperty
+    @typing.overload
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -1350,7 +1350,7 @@ class Extension(typing.Generic[ExtensionTypeVar]):
         return self._critical
 
     @property
-    def value(self) -> ExtensionType:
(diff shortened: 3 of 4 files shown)


src/cryptography/x509/name.py
```diff
@@ -196,14 +196,32 @@ def __repr__(self) -> str:
...
+
+    @typing.overload
+    def __init__(
+        self, attributes: typing.Iterable[RelativeDistinguishedName]
+    ) -> None:
+        ...
+
+    def __init__(
+        self,
+        attributes: typing.Iterable[
```
alex, line 211: Would writing this as `typing.Union[typing.Iterable[NameAttribute], typing.Iterable[RelativeDistinguishedName]` allow us to avoid the need for the overloads?

mathiasertl: I did try this, and you get:

```
$ mypy src/cryptography/ vectors/cryptography_vectors/ tests/
src/cryptography/x509/name.py:206: error: Incompatible types in assignment (expression has type "List[object]", variable has type "Union[Iterable[NameAttribute], Iterable[RelativeDistinguishedName]]")
```

When I typehint the attributes cast, I get:

```
$ mypy src/cryptography/ vectors/cryptography_vectors/ tests/
src/cryptography/x509/name.py:206: error: Name 'attributes' already defined on line 199  [no-redef]
```

Also when using a different name in the assignment I again get "Incompatible types in assingment". So... I'm open to any way to improve that. 

For the record, here is the full change I tried:

```diff
diff --git a/src/cryptography/x509/name.py b/src/cryptography/x509/name.py
index 0bfc2d94..f5be2b06 100644
--- a/src/cryptography/x509/name.py
+++ b/src/cryptography/x509/name.py
@@ -196,31 +196,25 @@ class RelativeDistinguishedName(object):
 
 
 class Name(object):
-    @typing.overload
-    def __init__(self, attributes: typing.Iterable[NameAttribute]) -> None:
-        ...
-
-    @typing.overload
-    def __init__(
-        self, attributes: typing.Iterable[RelativeDistinguishedName]
-    ) -> None:
-        ...
-
     def __init__(
         self,
-        attributes: typing.Iterable[
-            typing.Union[NameAttribute, RelativeDistinguishedName]
+        attributes: typing.Union[
+            typing.Iterable[NameAttribute],
+            typing.Iterable[RelativeDistinguishedName],
         ],
     ) -> None:
-        attributes = list(attributes)
-        if all(isinstance(x, NameAttribute) for x in attributes):
+        attrs: typing.Union[
+            typing.List[NameAttribute],
+            typing.List[RelativeDistinguishedName],
+        ] = list(attributes)
+        if all(isinstance(x, NameAttribute) for x in attrs):
             self._attributes = [
                 RelativeDistinguishedName([typing.cast(NameAttribute, x)])
-                for x in attributes
+                for x in attrs
             ]
-        elif all(isinstance(x, RelativeDistinguishedName) for x in attributes):
+        elif all(isinstance(x, RelativeDistinguishedName) for x in attrs):
             self._attributes = typing.cast(
-                typing.List[RelativeDistinguishedName], attributes
+                typing.List[RelativeDistinguishedName], attrs
             )
         else:
             raise TypeError(
```

alex: What a mess, thanks.

alex (approved):

src/cryptography/x509/name.py
mathiasertl: Noticed before you really don't like these overloads. They look quite okay to me.

reaperhulk (approved):

PR #5899, opened 2021-03-07
typehint x509.base

mathiasertl:
Add typehints to `cryptography.x509.base` and also at least typehint the interfaces used by this module:

```
$ mypy --strict src/cryptography/x509/base.py 
Success: no issues found in 1 source file
```

**Request for comments:** There are still some mypy errors, even in normal mode, (almost) all of this nature:

```
$ mypy src/cryptography/ vectors/cryptography_vectors/ tests/
src/cryptography/hazmat/backends/openssl/backend.py:931: error: Item "Ed25519PublicKey" of "Union[Ed25519PublicKey, Ed448PublicKey, RSAPublicKey, DSAPublicKey, EllipticCurvePublicKey]" has no attribute "_evp_pkey"  [union-attr]
...
```

The issue here is (as far as I could reverse-engineer):

1. `x509.CertificateSigningRequestBuilder.private_key` (that line 931) is now type hinted to `cryptography.hazmat._types._PRIVATE_KEY_TYPES`, which is the _interfaces_ in `cryptography.hazmat.primitives.asymmetric` (that's the whole purpose of that PR, after all).
2. The `public_key()` method also returns the _interfaces_ of the corresponding public keys.
3. So `public_key` in line 930 is typehinted to that interface class.
4. But the interface class does _not_ define the `_evp_pkey` property, the concrete implementation for OpenSSL does.

We could typehint the `private_key` parameter in the concrete implementation to the concrete classes - but that would violated parameter type contravariance and mypy would again complain. We could simply cast the variable to the concrete implementation (with an actual typecheck before, maybe). Or maybe something completely else?

### DIFF
--- a/src/cryptography/hazmat/backends/interfaces.py
+++ b/src/cryptography/hazmat/backends/interfaces.py
@@ -8,14 +8,16 @@ import typing
 
 from cryptography.hazmat._types import _PRIVATE_KEY_TYPES
 from cryptography.hazmat.primitives import hashes
-from cryptography.x509.base import Certificate
-from cryptography.x509.base import CertificateBuilder
-from cryptography.x509.base import CertificateRevocationList
-from cryptography.x509.base import CertificateRevocationListBuilder
-from cryptography.x509.base import CertificateSigningRequest
-from cryptography.x509.base import CertificateSigningRequestBuilder
-from cryptography.x509.base import RevokedCertificate
-from cryptography.x509.base import RevokedCertificateBuilder
+from cryptography.x509.base import (
+    Certificate,
+    CertificateBuilder,
+    CertificateRevocationList,
+    CertificateRevocationListBuilder,
+    CertificateSigningRequest,
+    CertificateSigningRequestBuilder,
+    RevokedCertificate,
+    RevokedCertificateBuilder,
+)
 from cryptography.x509.name import Name
 
 
--- a/src/cryptography/hazmat/backends/interfaces.py
+++ b/src/cryptography/hazmat/backends/interfaces.py
@@ -6,19 +6,21 @@
 import abc
 import typing
 
-from cryptography.hazmat._types import _PRIVATE_KEY_TYPES
-from cryptography.hazmat.primitives import hashes
-from cryptography.x509.base import (
-    Certificate,
-    CertificateBuilder,
-    CertificateRevocationList,
-    CertificateRevocationListBuilder,
-    CertificateSigningRequest,
-    CertificateSigningRequestBuilder,
-    RevokedCertificate,
-    RevokedCertificateBuilder,
-)
-from cryptography.x509.name import Name
+
+if typing.TYPE_CHECKING:
+    from cryptography.hazmat._types import _PRIVATE_KEY_TYPES
+    from cryptography.hazmat.primitives import hashes
+    from cryptography.x509.base import (
+        Certificate,
+        CertificateBuilder,
+        CertificateRevocationList,
+        CertificateRevocationListBuilder,
+        CertificateSigningRequest,
+        CertificateSigningRequestBuilder,
+        RevokedCertificate,
+        RevokedCertificateBuilder,
+    )
+    from cryptography.x509.name import Name
 
 
 class CipherBackend(metaclass=abc.ABCMeta):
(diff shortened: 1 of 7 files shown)


src/cryptography/hazmat/backends/interfaces.py
```diff
@@ -4,6 +4,19 @@
...
+from cryptography.hazmat.primitives import hashes
+from cryptography.x509.base import Certificate
+from cryptography.x509.base import CertificateBuilder
+from cryptography.x509.base import CertificateRevocationList
+from cryptography.x509.base import CertificateRevocationListBuilder
+from cryptography.x509.base import CertificateSigningRequest
+from cryptography.x509.base import CertificateSigningRequestBuilder
+from cryptography.x509.base import RevokedCertificate
+from cryptography.x509.base import RevokedCertificateBuilder
+from cryptography.x509.name import Name
```
alex, line 19: We prefer to have a single `from cryptography.x509.base import (A, b, C, D)` with newlines.

src/cryptography/x509/base.py
```diff
@@ -73,7 +74,7 @@ class Version(Enum):
 
 
 class InvalidVersion(Exception):
-    def __init__(self, msg, parsed_version):
+    def __init__(self, msg: str, parsed_version: typing.Any):
```
alex, line 77: Is this really an `Any`?  Looks like it should be an `int`.

src/cryptography/x509/base.py
```diff
@@ -438,10 +453,28 @@ def load_der_x509_crl(
...
         Creates an empty X.509 certificate request (v1).
         """
+        if extensions is None:
+            extensions = []
+        else:
+            extensions = list(extensions)
+        if attributes is None:
+            attributes = []
+        else:
+            attributes = list(attributes)
```
alex, line 476: We don't need the `list` wrapping here -- this function isn't a public API and our internals only ever call it with a `list`. For the same reason it's fine to just use the `[]` default.

src/cryptography/hazmat/backends/interfaces.py
mathiasertl: fixed

src/cryptography/x509/base.py
mathiasertl: The exception is only raised in one place (`src/cryptography/hazmat/backends/openssl/x509.py`), and there it's the return value of an internal function `X509_get_version()`.  The exception itself does not do anything with the value.

So I presume it's de facto always an int, and I've adjusted the code accordingly.

src/cryptography/x509/base.py
mathiasertl: The class is documented right here: https://cryptography.io/en/latest/x509/reference.html. While constructor arguments are admittedly not documented, it's not hard to find out about them, and the names are not prefixed with a `_` either. So I really feel we should leave this in. It's not like it hurts anyone.

mathiasertl:
I was about to say that the whole layout of code just screams "circular imports", but it seems unit tests beat me to it. 

Either we do some more significant refactoring here, or we add some special handing when type hinting is enabled. *shrug*

alex:
Can we start by splitting this into two pieces? I think the backend piece is the really hard part, so let's start by doing everything-but?

mathiasertl:
Hmm, fair point. I'll look into it tomorrow. 

But I'm not 100% sure there is a good sweet spot here: IIRC, type hinting x509.base immediately triggers mypy showing errors in the interfaces part, so it would be impossible to do the "easy" part first with a PR where myp still runs through. I'll report back.

reaperhulk:
Yeah we are into the parts of the code that we deliberately haven’t finished typing due to these challenges. From my perspective I don’t really care about typing inside our private interfaces, but I’d like to make sure all our public APIs are fully typed.

mathiasertl:
Alright, solved the circular import issue with a conditional import for mypy and added ignore-comments to some accessors to internal API.

alex:
I think there's two coverage issues:

1. You added a new branch on `_private_key is None` and that's not covered
2. You added a coverage ignore for the `if typing.TYPE_CHECKING` line, but that _whole block_ needs to be ignored. I'm not sure how to do that.

mathiasertl:
@alex Thanks for the pointer, I totally overlooked that if-statement.

Checks pass now with added test. the `if typing.TYPE_CHECKING` actually covers the whole block.

alex (approved):

PR #5900, opened 2021-03-07
Bugfix/issue 5889 typehint get values for types

### LINKED ISSUE #5889: Typing for get_values_for_type is incorrect
https://github.com/pyca/cryptography/blob/032a7f3cdca5207d3b1e2a5eed14897d88f78378/src/cryptography/x509/extensions.py#L1392-L1401
This is currently defined as returning a list of GeneralName objects, but it actually returns a list of strings *or* a list of OtherName objects (which is a type of GeneralName). However, `typing.Union[typing.List[str], typing.List[OtherName]]` causes mypy to get confused as well. We should figure out how to make this (and the other instances of this same signature in `extensions.py`) correct.

mathiasertl:
Fixes #5889.

Properly typehint `get_values_for_type`. I also added a type alias for the different ip address/network types.

One RFC definitely here: note the `__all__` in `x509.oid`. The issue here is (again) the implicit re-export of `ObjectIdentifier` that mypy doesn't like, breaking type hinting for OIDs:

```
from cryptography.x509.oid import ObjectIdentifier
oid = ObjectIdentifier("1.2.3")
reveal_type(oid)  # that's Any!
```

The same happens when you import from `cryptography.x509.ObjectIdentifier` - the canonical location - because __init__.py also imports from the "wrong" location, as do a few other places in the code.

The other way to do this would be to consistently import from `cryptography.hazmat._oid` throughout the source code.

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -23,8 +23,18 @@ from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey
 from cryptography.x509.certificate_transparency import (
     SignedCertificateTimestamp,
 )
-from cryptography.x509.general_name import GeneralName, IPAddress, OtherName
-from cryptography.x509.name import RelativeDistinguishedName
+from cryptography.x509.general_name import (
+    DNSName,
+    DirectoryName,
+    GeneralName,
+    IPAddress,
+    OtherName,
+    RFC822Name,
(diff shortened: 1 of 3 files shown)


reaperhulk:
I'm fine with doing `__all__`. There are probably other locations where this is happening since we did this pattern a few times to break import cycles.

reaperhulk (approved):

PR #5925, opened 2021-03-21
fix XTS less than one block length. fixes #5885

### LINKED ISSUE #5885: AES-XTS failing for plaintext less then 16 bytes
I am using version 2.6.1 (Python 2)
Encrypting plaintext less then 16 bytes in AES XTS mode throws the following exception:

> "_cryptography.exceptions.InternalError: Unknown OpenSSL error. This error is commonly encountered when another library is not cleaning up the OpenSSL error stack. If you are using cryptography with another library that uses OpenSSL try disabling it before reporting a bug. Otherwise please file an issue at https://github.com/pyca/cryptography/issues with information on how to reproduce this."

See code snippet below
Thanks

`from cryptography.hazmat.primitives.ciphers import (
            Cipher, algorithms, modes
    )

    limit = 10000
    failures =  []
    for i in range(limit):
        key = os.urandom(64)
        tweak = os.urandom(16)
        
        txtLen = i
        plaintext = os.urandom(txtLen)
        
        encryptor = Cipher(
            algorithms.AES(key),
            modes.XTS(tweak),
            backend=default_backend()
        ).encryptor()
        
        ciphertext = encryptor.update(plaintext) + encryptor.finalize()
        try:
            ciphertext = encryptor.update(plaintext) + encryptor.finalize()
        except Exception as e:
            if (not txtLen in failures):
                failures.append(txtLen)
            
        
    for i in failures:
        print i,
        `

reaperhulk:
Note: since we don't track total bytes processed this approach assumes any error in `update` with XTS is the < 1 block issue. If we want to be more granular we need to start tracking.

### DIFF
--- a/tests/hazmat/primitives/test_aes.py
+++ b/tests/hazmat/primitives/test_aes.py
@@ -52,6 +52,14 @@ class TestAESModeXTS(object):
                 computed_pt = dec.update(ct) + dec.finalize()
                 assert computed_pt == pt
 
+    def test_xts_too_short(self):
+        key = os.urandom(32)
+        tweak = os.urandom(16)
+        cipher = base.Cipher(algorithms.AES(key), modes.XTS(tweak))
+        enc = cipher.encryptor()
+        with pytest.raises(ValueError):
+            enc.update(b"0" * 15)
+
 
 @pytest.mark.supported(
     only_if=lambda backend: backend.cipher_supported(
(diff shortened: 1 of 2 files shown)


tests/hazmat/primitives/test_aes.py
```diff
@@ -52,6 +52,14 @@ def test_xts_vectors(self, backend, subtests):
                 computed_pt = dec.update(ct) + dec.finalize()
                 assert computed_pt == pt
 
+    def test_xts_too_short(self):
+        key = os.urandom(32)
+        tweak = os.urandom(16)
```
alex, line 57:
```suggestion
        key = b"\x00" * 32
        tweak = b"\x00" * 16
```

alex:
lol, this key is unloved

alex (approved):

Issue #5935, opened 2021-03-26
Build wheels for more python versions


ShadowJonathan:
According to the [wheelodex](https://www.wheelodex.org/projects/cryptography/), `cryptography` currently does not publish wheels for any other version except 3.6 (`cp36`)

I suggest generating wheels for all versions [currently officially supported](https://endoflife.date/python), which means versions 3.6 through 3.9.

This greatly lifts the dependency burden of `cryptography`, especially with the recent addition of a rust compiler dependency (#5771).

I personally suggest using [cibuildwheel](https://github.com/joerick/cibuildwheel/) for something like this, but i understand if using a tool like that is not possible due to security considerations.

Nontheless, I think `cryptography` should have wheels built for more versions than `3.6`

callahad:
Cryptography's wheels work on 3.6+, per the FAQ: https://cryptography.io/en/latest/faq.html#why-are-there-no-wheels-for-my-python3-x-version

ShadowJonathan:
...huh, i personally didn't know about that.

PR #5936, opened 2021-03-26
Add version specifier to "recent versions of pip"

ShadowJonathan:
This confused me when I read the FAQ mentioned in #5935, I went to look at the changelogs and [`20.0`](https://pip.pypa.io/en/stable/news/#id247) seems to add the support for `abi3`, so I'm adding that here for future reference.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -138,7 +138,7 @@ Why are there no wheels for my Python3.x version?
 Our Python3 wheels are ``abi3`` wheels. This means they support multiple
 versions of Python. The ``abi3`` wheel can be used with any version of Python
 greater than or equal to the version it specifies. Recent versions of ``pip``
-will automatically install ``abi3`` wheels.
+(versions ``20.0`` and greater) will automatically install ``abi3`` wheels.
 
 Why can't I import my PEM file?
 -------------------------------


tiran:
The PR is not entirely correct. ABI3 builds work with earlier versions of pip, too. 

* pip < 19.0 does not recognize abi3 tag
* pip 19.0 recognizes the abi3 tag on Linux and downloads the correct binary wheel
* pip 19.3 and 20.0 contain some fixes for abi3 tag for some platforms, Python versions, and build flavors.

ShadowJonathan:
Ah, thanks for correcting, so it has to be `19.0` instead of `20.0`?

tiran:
> Ah, thanks for correcting, so it has to be `19.0` instead of `20.0`?

The answer is: it's complicated :)

It has to be at least ``>=19.0`` but may require ``>=20.0`` on some platforms or builds.

ShadowJonathan:
So think then the current comment would suffice? It would be simpler to say "use 20.0 and greater" for wider support than to say "use 19.0 on some platforms, and 20.0 on others".

alex:
I don't think 19.0 is the correct minimum version -- we've had this note in our FAQ since 2018, which is older that pip 19!

reaperhulk:
Making this more specific is likely to just cause additional confusion since the various constraints change based on platform and tag. abi3 itself has been recognized for a long time on non-Windows platforms. manylinux1/2010/2014 have different dates, and newer versions of pip also change other aspects of installation. Our recommendation is always to upgrade to latest.

ShadowJonathan:
I think i'll close this PR then, seeing as how nailing the recommended version is pretty hard to do.

PR #5950, opened 2021-04-02
WIP: Add method Certificate._has_signature_of()

KonstantinShemyak:
This PR is the first step in implementing #5116 . Here, we only check the validity of the cryptographic signature. As this method is likely to be used only inside `Certificate.is_issued_by()`, it is declared as internal.
Huge line diff is caused by added test certificates, real code change is small.


KonstantinShemyak:
CI step `pyca/check` has been stuck for two days 🤔

alex:
Yes, it's a known issue. For now there's nothing to do but ignore that one job.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -166,6 +166,58 @@ def public_bytes(self, encoding: serialization.Encoding) -> bytes:
...
+            )
+
+        signature = self.signature
+        data = self.tbs_certificate_bytes
+        if isinstance(pubkey, asymmetric.rsa.RSAPublicKeyWithSerialization):
+            pubkey.verify(
+                signature,
+                data,
+                padding=asymmetric.padding.PKCS1v15(),
+                algorithm=self.signature_hash_algorithm,  # type: ignore
```
reaperhulk, line 185: Why do these need ignores?

src/cryptography/x509/base.py
```diff
@@ -104,6 +104,13 @@ def public_key(self) -> _PUBLIC_KEY_TYPES:
         Returns the public key
         """
 
+    @abc.abstractmethod
```
reaperhulk, line 107: As an internal method we don't need to add it to our interface (or document it publicly)

src/cryptography/hazmat/backends/openssl/x509.py
KonstantinShemyak: @reaperhulk `Certificate.signature_hash_algorithm` [is of type](https://github.com/pyca/cryptography/blob/3.4.7/src/cryptography/hazmat/backends/openssl/x509.py#L115) `Optional[hashes.HashAlgorithm]` , while `pubkey.verify` [expects](https://github.com/pyca/cryptography/blob/3.4.7/src/cryptography/hazmat/primitives/asymmetric/dsa.py#L139) its `algorithm` argument to be `Union[asym_utils.Prehashed, hashes.HashAlgorithm]`.

If the signer's public key is, say, RSA, then signature with it has to have some not-`None` hash algorithm, so we are safe to pass it to `verify()`. I don't see a way how to express this better in the code.

src/cryptography/x509/base.py
KonstantinShemyak: Oho! This makes the PR much shorter.
Branch force-pushed (also updated on `main`).

KonstantinShemyak:
I have a question myself: wouldn't it be more appropriate to move this method under `x509.base.Certificate`? The method has nothing specific to the OpenSSL backend.

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: If you assert on its type then mypy will unwrap the optional and this will work without type ignores

KonstantinShemyak:
Force-push: implemented [suggestion to assert type](https://github.com/pyca/cryptography/pull/5950#discussion_r612036700).

KonstantinShemyak:
Force-push: the new method is implemented completely under `x509`; no new code under `hazmat` anymore.

To remind, the final goal of this work is to implement certificate validation (#2381). This PR is the first step out of three. The second step is #5116, I have pushed a WIP branch (missing tests and documentation) showing how I plan to do it. The final step is doing checks as in #5116 along all computed chains of trust from the leaf to any of the provided trusted CAs. 

GitHub seems to start requiring approval from project maintainers to run workflows :smiling_imp:

KonstantinShemyak:
Force-push: rebase on updated `main`.

KonstantinShemyak:
^ rebase on updated `main`.

@reaperhulk , @alex - what is your view on this approach?
While this is only first step out of three, I believe it's also useful by itself, as now the user of the library [needs non-trivial code](https://github.com/pyca/cryptography/blame/main/docs/x509/reference.rst#L462) to just verify the cryptographic part of the signature.

alex:
First, I'm sorry we haven't given this PR much attention.

Second, I think I'm conceptually ok with this, but I'm on the fence about having it as a private method. I realize the goal is for it to be ultimately used by a more powerful (and safe) public API, but I do wonder if this shouldn't just be public as well.

I haven't reviewed the implementation of the PR yet. I also think we might want to iterate on the name a bit.

alex:
https://github.com/pyca/cryptography/commit/db7dd61de3c6f7c8d66d5615cbfbcf5c085c4448 incorporates this

PR #5951, opened 2021-04-02
Update annotation for key

Dreamsorcerer:
The key argument is passed immediately into `urlsafe_b64decode()`, which accepts `str` or `bytes`.

> Decode bytes-like object or ASCII string s...
https://docs.python.org/3/library/base64.html#base64.urlsafe_b64decode

### DIFF
--- a/src/cryptography/fernet.py
+++ b/src/cryptography/fernet.py
@@ -27,7 +27,7 @@ _MAX_CLOCK_SKEW = 60
 
 
 class Fernet(object):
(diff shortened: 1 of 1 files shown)


alex (approved):
This is also what our docs say: https://github.com/pyca/cryptography/blob/main/docs/fernet.rst#fernet-symmetric-encryption

reaperhulk:
Closing/reopening because the opendev builder is hanging in a way that precludes restarting it.

PR #5952, opened 2021-04-02
Build manylinux_2_24 wheels

alex:


alex:
arm64 wheel builds, but not x86 ones are failing, and I have no idea why.

reaperhulk (approved):

PR #5959, opened 2021-04-08
Textual OID representation

mhils:
We'd like to have a textual representation of OIDs in mitmproxy. Would you folks be willing to either publicly expose `_NAMEOID_TO_NAME` (as done in this PR), or add a function/property that exposes it (e.g. `NameAttribute.oid_str()` or `NameAttribute.name`)? Either change should be trivial, I'd be happy to send a PR.

Short background: We previously used `.rfc4514_string()`, but that adds undesired escaping (`DigiCert\, Inc.` instead of `DigiCert, Inc.`, mitmproxy/mitmproxy#4555). We could use `NameAttribute.rfc4514_string().partition("=")[0]`, but that feels unnecessarily complex to me.

I'm also happy to take "no" as an answer here, but I figured we can't be the only ones. 😃 

/cc @Prinzhorn

### DIFF
--- a/src/rust/Cargo.lock
+++ b/src/rust/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "asn1"
-version = "0.3.1"
+version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


alex:
I think an attribute named `rfc4514_attribute_name` would be an acceptable public API here.

reaperhulk:
That's fine by me as well.

PR #5963, opened 2021-04-11
Converted DER parsing of SPKIs to rust

alex:


reaperhulk:
Do we have tests that verify that this generates the same values? Looking at `from_public_key` it doesn't look like our test for this asserts on a specific value.

alex:
There's a handful of tests -- e.g. `test_from_dsa_public_key` in `test_x509_ext.py` checks that it matches what's in a cert. That's against a fixed value.

alex:
There's also versions of that for ed25519 and ed448

reaperhulk:
Ah, yes, okay. 👍

reaperhulk (approved):

PR #5965, opened 2021-04-13
Convert encode/decode of DSS signatures to Rust

alex:
Before this merges, I'd like to benchmark this to make sure we're going zoom zoom.

### DIFF
--- a/src/rust/src/asn1.rs
+++ b/src/rust/src/asn1.rs
@@ -120,9 +120,8 @@ fn py_uint_to_big_endian_bytes<'p>(
     py: pyo3::Python<'p>,
     v: &'p pyo3::types::PyLong,
 ) -> pyo3::PyResult<&'p [u8]> {
-    if v.rich_compare((0).to_object(py), CompareOp::Lt)?
-        .is_true()?
-    {
+    let zero = (0).to_object(py);
+    if v.rich_compare(zero, CompareOp::Lt)?.is_true()? {
         return Err(pyo3::exceptions::PyValueError::new_err(
             "Negative integers are not supported",
         ));
--- a/src/rust/src/asn1.rs
+++ b/src/rust/src/asn1.rs
@@ -2,6 +2,7 @@
 // 2.0, and the BSD License. See the LICENSE file in the root of this repository
 // for complete details.
 
+use pyo3::class::basic::CompareOp;
 use pyo3::conversion::ToPyObject;
 
 enum PyAsn1Error {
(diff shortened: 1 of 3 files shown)


alex:
Benchmark:

```
python -m timeit -s "from cryptography.hazmat.primitives.asymmetric.utils import decode_dss_signature" "decode_dss_signature(b'0-\x02\x15\x00\xb5\xaf0xg\xfb\x8bT9\x00\x13\xccg\x02\r\xdf\x1f,\x0b\x81\x02\x14b\r;\"\xabP1D\x0c>5\xea\xb6\xf4\x81)\x8f\x9e\x9f\x08')"
```

Before: 6.4 usec per loop
After: 730 nsec per loop

Seems good

alex:
Benchmark:

```
(.venv) alex@media-01:/tmp/x$ python -m timeit -s "from cryptography.hazmat.primitives.asymmetric.utils import encode_dss_signature" "encode_dss_signature(21312412321, 32143214321)"
```

Before: 2.6 usec per loop
After: 1 usec per loop

Seems good

src/rust/src/asn1.rs
```diff
@@ -95,11 +96,71 @@ fn parse_spki_for_data(py: pyo3::Python<'_>, data: &[u8]) -> pyo3::PyResult<pyo3
...
+    v.call_method1("to_bytes", (n, "big"))?.extract()
+}
+
+#[pyo3::prelude::pyfunction]
+fn encode_dss_signature(
+    py: pyo3::Python<'_>,
+    r: &pyo3::types::PyLong,
+    s: &pyo3::types::PyLong,
+) -> pyo3::PyResult<pyo3::PyObject> {
+    if r.rich_compare((0).to_object(py), CompareOp::Lt)?
```
reaperhulk, line 134: How do we ensure coverage when hoisting these types of branches into rust?

alex: Well, we could figure out Rust coverage in CI? Do you want that to be a pre-req to this? (Is this literally the first branch in the rust code? Neat!)

reaperhulk: `parse_spki_for_data` has a branch too, but it slipped my mind in that review 😄

Yes I think we should figure that out, although I'm a bit worried given that our hosted coverage service doesn't handle a single language well...

alex: Yeah, tooling here will be... fascinating. I'll investigate.

reaperhulk:
This should be rebase-able now 😄

reaperhulk (approved):

PR #5969, opened 2021-04-13
Add x509.Name.rfc4514_attribute_name

mhils:
Follow up to #5959, which for some reason decided to self-close when I force-pushed. 🤷😃 

In short, this PR adds `x509.Name.rfc4514_attribute_name` to expose a (short-form) textual OID representation. I'm happy to change things around, let me know what you think.

### DIFF
--- a/src/cryptography/x509/name.py
+++ b/src/cryptography/x509/name.py
@@ -118,6 +118,13 @@ class NameAttribute(object):
     def value(self) -> str:
         return self._value
 
+    @property
+    def rfc4514_attribute_name(self) -> str:
+        """
+        The short attribute name (for example "CN") if available, otherwise the OID dotted string.
+        """
+        return _NAMEOID_TO_NAME.get(self.oid, self.oid.dotted_string)
+
     def rfc4514_string(self) -> str:
         """
         Format as RFC4514 Distinguished Name string.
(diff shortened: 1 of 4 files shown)


alex (commented):
Please add this to `docs/x509/reference.rst` and add a test for it. Thanks.

alex (approved):

src/cryptography/x509/name.py
```diff
@@ -118,15 +118,21 @@ def oid(self) -> ObjectIdentifier:
     def value(self) -> str:
         return self._value
 
+    @property
+    def rfc4514_attribute_name(self) -> str:
+        """
+        The short attribute name (for example "CN") if available, otherwise the OID dotted string.
```
alex, line 124: Line is too long, please wrap at 80 cols.

src/cryptography/x509/name.py
```diff
@@ -118,15 +118,21 @@ def oid(self) -> ObjectIdentifier:
...
     def rfc4514_string(self) -> str:
         """
         Format as RFC4514 Distinguished Name string.
 
         Use short attribute name if available, otherwise fall back to OID
         dotted string.
         """
-        key = _NAMEOID_TO_NAME.get(self.oid, self.oid.dotted_string)
-        return "%s=%s" % (key, _escape_dn_value(self.value))
+        return "%s=%s" % (self.rfc4514_attribute_name, _escape_dn_value(self.value))
```
alex, line 135: This will also need to be line-wrapped. (Running black should do it automatically!)

mhils: Fixed, thanks! And sorry for the repeated turnarounds. Seems like I really need to get used to more zealous CI checks :)

alex (approved):

PR #5972, opened 2021-04-14
switch to using EVP_PKEY_derive instead of DH_compute_key in DH

reaperhulk:
Where the DH key checks are occurring is changing in OpenSSL 3.0 and this makes it easier to be consistent (and is the API we should be using anyway). The tests change because EVP_PKEY_derive now verifies that we have shared parameters, which the test previously only verified by asserting that the derived keys didn't match.

Let's see if this breaks Libre.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/dh.py
+++ b/src/cryptography/hazmat/backends/openssl/dh.py
@@ -127,35 +127,45 @@ class _DHPrivateKey(dh.DHPrivateKey):
         )
 
     def exchange(self, peer_public_key: dh.DHPublicKey) -> bytes:
-        buf = self._backend._ffi.new("unsigned char[]", self._key_size_bytes)
-        pub_key = self._backend._ffi.new("BIGNUM **")
-        self._backend._lib.DH_get0_key(
-            peer_public_key._dh_cdata,  # type: ignore[attr-defined]
-            pub_key,
-            self._backend._ffi.NULL,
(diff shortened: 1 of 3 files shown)


src/cryptography/hazmat/backends/openssl/dh.py
```diff
@@ -127,35 +127,45 @@ def private_numbers(self) -> dh.DHPrivateNumbers:
...
         )
-        self._backend.openssl_assert(pub_key[0] != self._backend._ffi.NULL)
-        res = self._backend._lib.DH_compute_key(
-            buf, pub_key[0], self._dh_cdata
+        self._backend.openssl_assert(ctx != self._backend._ffi.NULL)
+        ctx = self._backend._ffi.gc(ctx, self._backend._lib.EVP_PKEY_CTX_free)
+        res = self._backend._lib.EVP_PKEY_derive_init(ctx)
+        self._backend.openssl_assert(res == 1)
+        res = self._backend._lib.EVP_PKEY_derive_set_peer(
+            ctx, peer_public_key._evp_pkey  # type:ignore[attr-defined]
```
alex, line 138: Ok I think this function actually needs a type check, rather than a type:ignore. WHat'd happen here if you passed some other type with an `_evp_key` (i.e. any other pkey type)

alex:
CI issue

alex (approved):

PR #5974, opened 2021-04-15
remove bionic and centos 8 arm64 jobs

reaperhulk:
Reliability and speed is insufficient and we get almost the same CI
bug finding value with just one arm64 job for now.


alex (commented):
You need to remove these in `project.yaml`.

alex (approved):

PR #5976, opened 2021-04-15
fixes #5975 -- added typing stubs for rust module

### LINKED ISSUE #5975: Add type stubs for rust module
In https://github.com/pyca/cryptography/pull/5965 we replaced two Python functions with functions from our new rust bindings. However, we lost type hinting with that. We should add type stubs for the methods we're adding.

We should also try to figure out how to enforce that we keep public methods in the module and type signatures in sync. This may not be feasible until we can add more mypy flags though.

alex:


reaperhulk:
Can we remove the `_rust` section in `mypy.ini` with this?

reaperhulk (approved):

reaperhulk (approved):

PR #5978, opened 2021-04-16
Corrected test parsing of certificates

alex:
RFC 5280 says these fields are both `IMPLICIT` tagging, and `IMPLICIT` tagging is only `CONSTRUCTED` if the inner type is, and the inner type is `BIT STRING` here, which is not `CONSTRUCTED`.

In practice this is irrelevant since _no one_ ever uses these fields


davidben (approved):
Oops yeah I must have missed the implicit here.

reaperhulk (approved):

reaperhulk:
standard codecov is full of it comment.

PR #5979, opened 2021-04-16
Port X.509 test parser to Rust

alex:


tests/x509/test_x509.py
```diff
@@ -1971,9 +1902,13 @@ def test_extreme_times(self, not_valid_before, not_valid_after, backend):
...
         assert cert.not_valid_before == not_valid_before
         assert cert.not_valid_after == not_valid_after
-        parsed = _parse_cert(cert.public_bytes(serialization.Encoding.DER))
-        assert parsed.not_before_tag == UTC_TIME
-        assert parsed.not_after_tag == GENERALIZED_TIME
+        parsed = asn1.test_parse_certificate(
+            cert.public_bytes(serialization.Encoding.DER)
+        )
+        # UTC TIME
+        assert parsed.not_before_tag == 0x17
```
reaperhulk, line 1909: Should the rust side expose constants for these tags?

src/rust/src/asn1.rs
```diff
@@ -179,6 +179,93 @@ fn encode_dss_signature(
...
+                        let not_before_tag = p.read_element::<asn1::Tlv>()?.tag();
+                        let not_after_tag = p.read_element::<asn1::Tlv>()?.tag();
+                        Ok((not_before_tag, not_after_tag))
+                    })?;
+                    // Subject
+                    let subject_value_tags = parse_name_value_tags(p)?;
+
+                    // Subject public key info
+                    p.read_element::<asn1::Sequence>()?;
+                    // Issuer unique ID
```
reaperhulk, line 244: Given what you learned here it's probably worth putting a comment that says "these are optional but literally do not exist in the real world".

tests/x509/test_x509.py
alex: Don't really have an opinion on that.

src/rust/src/asn1.rs
alex: Sure, I can add a comment.

tests/x509/test_x509.py
reaperhulk: I'm comfortable with saying we can revisit this as we progress. It's not like this is a breaking API change so we can add it at any time.

reaperhulk (approved):

PR #5982, opened 2021-04-17
[wip] attempt to use GHA for generating coverage reports

alex:


alex:
I think this is currently blocked on https://github.com/nedbat/coveragepy/issues/1147 -- or on me coming up with an even more deranged workaround.

PR #5983, opened 2021-04-18
Added more paths to coverage path combiner

alex:


.coveragerc
```diff
@@ -8,6 +8,8 @@ source =
 source =
    src/cryptography
    .tox/*/lib/python*/site-packages/cryptography
+   .tox/*/lib64/python*/site-packages/cryptography
```
reaperhulk, line 11: Could we just do lib*?

alex: We can, would you prefer that?

reaperhulk: Please

alex: done

reaperhulk (approved):

PR #5988, opened 2021-04-19
Fix build with LibreSSL 3.3.2

vishwin:
LibreSSL 3.3.2 added support for `SSL_OP_NO_DTLS*`. Not opposed to using `LESS_THAN` instead of `OR_GREATER` for the preprocessor guards.


src/_cffi_src/openssl/cryptography.py
```diff
@@ -32,6 +32,13 @@
 #include <Winsock2.h>
 #endif
 
+#if CRYPTOGRAPHY_IS_LIBRESSL
+    #define CRYPTOGRAPHY_LIBRESSL_332_OR_GREATER \
+        (LIBRESSL_VERSION_NUMBER >= 0x3030200f)
```
alex, line 37: Please don't indent defines like this.

src/_cffi_src/openssl/ssl.py
```diff
@@ -586,8 +586,10 @@
 #endif
 
 #if CRYPTOGRAPHY_IS_LIBRESSL
+#if !CRYPTOGRAPHY_LIBRESSL_332_OR_GREATER
```
alex, line 589: Given teh single user, it might be cleaner if the context were `LESS_THAN`

alex (commented):
Thanks!

src/_cffi_src/openssl/cryptography.py
```diff
@@ -59,6 +66,8 @@
 static const int CRYPTOGRAPHY_NEEDS_OSRANDOM_ENGINE;
 
 static const int CRYPTOGRAPHY_IS_LIBRESSL;
+
+static const int CRYPTOGRAPHY_LIBRESSL_LESS_THAN_332;
```
alex, line 70: This can be removed, it's not actually used from Python.

alex (approved):

PR #6001, opened 2021-04-23
make PRIVATE/PUBLIC_KEY_TYPES a public API

mathiasertl:
With this PR I propose to move  the type aliases in `hazmat._types` to `hazmat.primitives.asymmetric.types` and make them public (= remove the `_` at the start).

Rationale: Without public access, it becomes impossible to typehint functions that use cryptography without either duplicating the type aliases or accessing internal members. Making them public does not hurt maintainability of the code base, as the `Union` would have to updated with any new key types anyway to make mypy happy.

Consider this (not too outlandish) code example in a downstream project:

```python
# Either that import or create your own type alias that is just a duplicate
from cryptography.hazmat._types import _PRIVATE_KEY_TYPES
# This would be much better:
#from cryptography.hazmat.primitives.asymmetric import types

def read_private_key(path) -> _PRIVATE_KEY_TYPES:
    """Reads a private key but throw an error if it is an RSA key with key_size < 2048."""
    with open("path/to/key.pem", "rb") as key_file:
        key = serialization.load_pem_private_key(key_file.read(), password=None)
    
    if isinstance(key, rsa.RSAPrivateKey) and key.key_size < 2048:
        raise ValueError("Sorry, RSA keys must be at least 2048 bytes")
    return key
```

### DIFF
--- a/src/cryptography/hazmat/backends/interfaces.py
+++ b/src/cryptography/hazmat/backends/interfaces.py
@@ -8,7 +8,9 @@ import typing
 
 
 if typing.TYPE_CHECKING:
-    from cryptography.hazmat._types import _PRIVATE_KEY_TYPES
+    from cryptography.hazmat.primitives.asymmetric.types import (
+        PRIVATE_KEY_TYPES,
+    )
     from cryptography.hazmat.primitives import hashes
     from cryptography.x509.base import (
         Certificate,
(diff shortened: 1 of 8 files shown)


reaperhulk (approved):

reaperhulk:
You're right, the modifications to this type alias will always be things our users would want anyway, so 👍

PR #6008, opened 2021-04-25
update the tests to use sha2 in more places

reaperhulk:
Also fix two places we were passing a class instead of an instance.


alex:
You broke CI.

reaperhulk:
I wondered if I would

alex (approved):

PR #6012, opened 2021-04-25
FIPS 3.0.0 support

reaperhulk:


reaperhulk:
(This is blocked on a bug that was fixed in master post-alpha15)

tests/conftest.py
```diff
@@ -11,6 +11,11 @@
 
 
 def pytest_report_header(config):
+    # Performed in this function to enable it before any test collection
```
reaperhulk, line 14: Is there a better function to call this in? I couldn't find any pytest documentation about the order of hooks, but this one definitely occurs before any collection.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -311,17 +328,31 @@ def hash_supported(self, algorithm):
         return evp_md != self._ffi.NULL
 
     def scrypt_supported(self):
-        return self._lib.Cryptography_HAS_SCRYPT == 1
+        if self._fips_enabled:
+            return False
```
reaperhulk, line 332: Insert salty comment here since I definitely disagree with OpenSSL 3.0.0's position on this.

.github/workflows/ci.yml
```diff
@@ -31,6 +31,8 @@ jobs:
           - {VERSION: "3.9", TOXENV: "py39-ssh", OPENSSL: {TYPE: "openssl", VERSION: "1.1.1k"}}
           - {VERSION: "3.9", TOXENV: "py39", OPENSSL: {TYPE: "openssl", VERSION: "1.1.1k", CONFIG_FLAGS: "no-engine no-rc2 no-srtp no-ct"}}
           - {VERSION: "3.9", TOXENV: "py39", OPENSSL: {TYPE: "openssl", VERSION: "3.0.0-beta2"}}
+          - {VERSION: "3.9", TOXENV: "py39", OPENSSL: {TYPE: "openssl", VERSION: "3.0.0-beta2"}}
```
alex, line 34: Is this a dupe of above?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -200,12 +207,22 @@ def __init__(self):
         if self._lib.Cryptography_HAS_EVP_PKEY_DHX:
             self._dh_types.append(self._lib.EVP_PKEY_DHX)
 
+    def __repr__(self):
+        return "backend: {}, FIPS: {}".format(
```
alex, line 211: Can we format this more repr like `<OpenSSLBackend(...)>` or so

.github/workflows/ci.yml
reaperhulk: yep, whoops

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -306,17 +323,31 @@ def hash_supported(self, algorithm):
...
 
     def cipher_supported(self, cipher, mode):
-        if self._fips_enabled and not isinstance(cipher, self._fips_ciphers):
-            return False
+        if self._fips_enabled:
+            # FIPS mode requires AES or TripleDES, but only CBC/ECB allowed
+            # in TripleDES mode.
+            if not isinstance(cipher, self._fips_ciphers) or (
+                isinstance(cipher, TripleDES)
+                and not isinstance(mode, (CBC, ECB))
```
alex, line 347: What on god's green earth...

tests/conftest.py
```diff
@@ -42,3 +48,31 @@ def disable_rsa_checks(backend):
     backend._rsa_skip_check_key = True
     yield
     backend._rsa_skip_check_key = False
+
+
+def enable_fips():
```
alex, line 53: Is there no way to enable this external to the process?

reaperhulk: This can be done via config but to do that we would need to modify our initialization to read the config by default (it currently does not). That would possibly change behavior for users who don't use our wheels.

tests/conftest.py
alex: https://docs.pytest.org/en/6.2.x/reference.html#initialization-hooks strongly suggests to me that you want `pytest_configure`

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -116,6 +116,22 @@ class Binding(object):
...
+        # have the FIPS provider installed properly.
+        _openssl_assert(self.lib, self.lib.CRYPTOGRAPHY_OPENSSL_300_OR_GREATER)
+        self._base_provider = self.lib.OSSL_PROVIDER_load(
+            self.ffi.NULL, b"base"
+        )
+        _openssl_assert(self.lib, self._base_provider != self.ffi.NULL)
+        self.lib._fips_provider = self.lib.OSSL_PROVIDER_load(
+            self.ffi.NULL, b"fips"
+        )
+        _openssl_assert(self.lib, self.lib._fips_provider != self.ffi.NULL)
```
alex, line 130:
```suggestion
        self._fips_provider = self.lib.OSSL_PROVIDER_load(
            self.ffi.NULL, b"fips"
        )
        _openssl_assert(self.lib, self._fips_provider != self.ffi.NULL)
```

alex (approved):

alex (approved):

PR #6024, opened 2021-04-29
Fix version in changelog

bagerard:
Minor edit but I noticed the changelog was mentioning 35.0.0 instead of (I guess 3.5.0)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,9 +1,9 @@
 Changelog
 =========
 
-.. _v35-0-0:
+.. _v3-5-0:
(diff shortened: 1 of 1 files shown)


alex:
This is intentional: https://cryptography.io/en/latest/api-stability/#versioning

I guess we can add a changelog entry that we have changed our version scheme. Would you like to do that?

bagerard:
OK I'll do that shortly

bagerard:
I've opened another MR here https://github.com/pyca/cryptography/pull/6027

PR #6027, opened 2021-05-01
Document the version scheme change in changelog

bagerard:
Update changelog to document the version scheme to avoid possible  confusion


reaperhulk (approved):
Thanks for the contribution!

PR #6051, opened 2021-05-16
Allow to build wheels in Docker containers

d9pouces:
cryptography currently has wheels for a few Linux architectures: pypy3.6, pypy3.7 and cpython3.6.
Installing cryptography with other Python versions is harder (due to the rust dependency).
This patch allows to build wheels in Docker containers, for easily building wheels in other cpython versions and uploading them to pypi.  For example, if you target 3.5, 3.6, 3.7, 3.8, 3.9, 3.10 and Pypy3.7:

```bash
python release.py 3.4.7 python:3.5-slim-buster python:3.6-slim-buster python:3.7-slim-buster python:3.8-slim-buster python:3.9-slim-buster python:rc-slim-buster pypy:3.7-slim-buster
```

### DIFF
--- a/docker_build.py
+++ b/docker_build.py
@@ -1,0 +1,21 @@
+import argparse
+import subprocess
+import sys
+from pathlib import Path
+
+
+def build(workdir):
+    python_deps = ["wheel", "setuptools_rust", "click", "requests"]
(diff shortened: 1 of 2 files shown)


alex:
https://cryptography.io/en/latest/faq/#why-are-there-no-wheels-for-my-python3-x-version

PR #6054, opened 2021-05-19
Point folks towards irc.libera.chat

alex:


reaperhulk:
Do we want to migrate to #pyca as a channel name at the same time?

alex:
No opinion. Do you want to?

reaperhulk:
Yeah I think we should. Not because we intend to overly expand the umbrella of topics, but because I'd like to minimize confusion about us being a generic cryptography channel.

reaperhulk (approved):

PR #6079, opened 2021-05-31
oxidize crlnumber

reaperhulk:


src/rust/src/x509.rs
```diff
@@ -74,11 +75,31 @@ fn parse_crl_entry_extension(
...
+    py: pyo3::Python<'_>,
+    der_oid: &[u8],
+    ext_data: &[u8],
+) -> Result<pyo3::PyObject, PyAsn1Error> {
+    let oid = asn1::ObjectIdentifier::from_der(der_oid).unwrap();
+
+    let x509_module = py.import("cryptography.x509")?;
+    if oid == *CRL_NUMBER_OID {
+        let bignum = asn1::parse_single::<asn1::BigUint>(ext_data)?;
+        let pynum = big_asn1_uint_to_py(py, bignum)?;
```
alex, line 89: Does this need to be a bignum, or can it be a fixed-width integer? I have no idea what kind of values go in here.

alex: Eh, from teh RFC:

>    Given the requirements above, CRL numbers can be expected to contain
>   long integers.  CRL verifiers MUST be able to handle CRLNumber values
>   up to 20 octets.  Conforming CRL issuers MUST NOT use CRLNumber
>   values longer than 20 octets.

alex (dismissed):

alex:
`cargo fmt` required

alex (approved):

PR #6089, opened 2021-06-06
oxidize SubjectAltName

reaperhulk:
This encompasses oxidizing parsing of GeneralName, which most of the remaining extensions need.

### DIFF
--- a/src/rust/src/x509.rs
+++ b/src/rust/src/x509.rs
@@ -2,7 +2,7 @@
 // 2.0, and the BSD License. See the LICENSE file in the root of this repository
 // for complete details.
 
-use crate::asn1::{big_asn1_uint_to_py, PyAsn1Error};
+use crate::asn1::{big_asn1_uint_to_py, AttributeTypeValue, Name, PyAsn1Error};
 use pyo3::conversion::ToPyObject;
 
 lazy_static::lazy_static! {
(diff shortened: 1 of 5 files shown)


src/rust/src/x509.rs
```diff
@@ -55,7 +92,100 @@ fn parse_x509_extension(
...
+                }
+                GeneralName::RegisteredID(data) => {
+                    let oid = x509_module
+                        .call_method1("ObjectIdentifier", (data.to_string(),))?
+                        .to_object(py);
+                    x509_module
+                        .call_method1("RegisteredID", (oid,))?
+                        .to_object(py)
+                }
+                // TODO: can we combine this with the EDIPartyName path?
```
reaperhulk, line 164: Right now I've got two separate paths for X400 and EDIPartyName because it's not clear to me if there's a way to get the implicit tag value with a `_` catch-all. However, this duplication won't have full coverage because we have no EDIPartyName vectors (and I don't want to create one!)

alex: Since these integers are kind of gibberish, I don't see a way to do this without paths for each. Or not providing the integer values.

src/rust/src/x509.rs
```diff
@@ -55,7 +92,100 @@ fn parse_x509_extension(
     let oid = asn1::ObjectIdentifier::from_der(der_oid).unwrap();
 
     let x509_module = py.import("cryptography.x509")?;
-    if oid == *TLS_FEATURE_OID {
+    if oid == *SUBJECT_ALTERNATIVE_NAME_OID {
+        let sans = pyo3::types::PyList::empty(py);
```
alex, line 96: Can we package general-name-to-python up in a function so it can be reused.

src/rust/src/x509.rs
```diff
@@ -55,7 +92,100 @@ fn parse_x509_extension(
...
+                    .getattr("RFC822Name")?
+                    .call_method1("_init_without_validation", (data.0,))?
+                    .to_object(py),
+                GeneralName::DNSName(data) => x509_module
+                    .getattr("DNSName")?
+                    .call_method1("_init_without_validation", (data.0,))?
+                    .to_object(py),
+                GeneralName::DirectoryName(data) => {
+                    let py_list = pyo3::types::PyList::empty(py);
+                    for rdn in data {
```
alex, line 117: same with name-to-python

src/rust/src/x509.rs
```diff
@@ -55,7 +92,100 @@ fn parse_x509_extension(
...
+                        let py_set = pyo3::types::PySet::empty(py)?;
+                        for attribute in rdn {
+                            let oid = x509_module
+                                .call_method1("ObjectIdentifier", (attribute.type_id.to_string(),))?
+                                .to_object(py);
+                            let tag_enum = py
+                                .import("cryptography.x509.name")?
+                                .getattr("_ASN1_TYPE_TO_ENUM")?;
+                            let py_tag = tag_enum.get_item(attribute.value.tag().to_object(py))?;
+                            // TODO: is it safe to assume utf8? Current Python does.
```
tiran, line 127: Pretty much, https://datatracker.ietf.org/doc/html/rfc4514#section-2.4

As you all know, DN's are uncommon and the rare cases follow LDAP syntax. LDAP implementations use UTF-8 in RDN attribute values.

src/rust/src/x509.rs
```diff
@@ -19,6 +19,43 @@ lazy_static::lazy_static! {
...
+            std::str::from_utf8(data).map_err(|_| asn1::ParseError::InvalidValue)?,
+        ))
+    }
+}
+
+#[derive(asn1::Asn1Read)]
+enum GeneralName<'a> {
+    #[implicit(0)]
+    OtherName(AttributeTypeValue<'a>),
+    #[implicit(1)]
```
alex, line 40: nit: can you put a newline between each of the options? I think it'll make it more readable

src/rust/src/x509.rs
```diff
@@ -46,6 +83,115 @@ fn get_bit(input: &[u8], n: usize) -> bool {
...
+) -> Result<pyo3::PyObject, PyAsn1Error> {
+    let x509_module = py.import("cryptography.x509")?;
+    let oid = x509_module
+        .call_method1("ObjectIdentifier", (attribute.type_id.to_string(),))?
+        .to_object(py);
+    let tag_enum = py
+        .import("cryptography.x509.name")?
+        .getattr("_ASN1_TYPE_TO_ENUM")?;
+    let py_tag = tag_enum.get_item(attribute.value.tag().to_object(py))?;
+    // TODO: is it safe to assume utf8? Current Python does.
```
alex, line 98: If teh current behaviors assumes this stuff is utf-8 it seems fine to me. We can always change this later.

src/rust/src/x509.rs
```diff
@@ -46,6 +83,115 @@ fn get_bit(input: &[u8], n: usize) -> bool {
...
+    let py_data =
+        std::str::from_utf8(attribute.value.data()).map_err(|_| asn1::ParseError::InvalidValue)?;
+    Ok(x509_module
+        .call_method1("NameAttribute", (oid, py_data, py_tag))?
+        .to_object(py))
+}
+
+fn parse_name(py: pyo3::Python<'_>, name: Name) -> Result<pyo3::PyObject, PyAsn1Error> {
+    let x509_module = py.import("cryptography.x509")?;
+    let py_list = pyo3::types::PyList::empty(py);
```
alex, line 108: Can we give these semantically meaningful names? Just calling itlist tells us nothing.

src/rust/src/x509.rs
```diff
@@ -46,6 +83,115 @@ fn get_bit(input: &[u8], n: usize) -> bool {
...
+    Ok(pyobj)
+}
+
+fn parse_general_names(
+    py: pyo3::Python<'_>,
+    ext_data: &[u8],
+) -> Result<pyo3::PyObject, PyAsn1Error> {
+    let gns = pyo3::types::PyList::empty(py);
+    for gn in asn1::parse_single::<asn1::SequenceOf<GeneralName>>(ext_data)? {
+        let pyobj = parse_general_name(py, gn)?;
```
alex, line 189: same

alex (approved):

alex:
Pretty sure codecov is nonsense here.

PR #6100, opened 2021-06-09
Add RDN for emailAddress

divinehawk:
This patch adds in the short name for an RDN "E". This matches the behavior for Windows.

It's deprecated but still in many of the certs I encounter --1.2.840.113549.1.9.1 in the subject.


reaperhulk:
The function that uses that dictionary is `Name.rfc4514_string()`. Based on my reading of section 2 of that RFC I would be okay with adding this if we can document where `E` is officially designated as the short name. Is there an RFC or LDAP spec that does so?

reaperhulk:
This change also breaks a test, which assumes that there is no short name, so that will need to be resolved as well.

tests/x509/test_x509.py
```diff
@@ -4690,7 +4690,7 @@ def test_distinugished_name(self):
         # Nonstandard attribute OID
         na = x509.NameAttribute(NameOID.EMAIL_ADDRESS, "somebody@example.com")
         assert (
-            na.rfc4514_string() == "1.2.840.113549.1.9.1=somebody@example.com"
+            na.rfc4514_string() == "E=somebody@example.com"
```
alex, line 4693: The goal of this test is to see what happens with an OID not in that dict, so we need to use a different OID here, not change the expectations

divinehawk: Thank you for the explanation. Hopefully this works better.

divinehawk:
Unfortunately I could not find an office spec for that. Several of the LDAP providers use E:

> https://www.ibm.com/docs/en/ibm-mq/7.5?topic=certificates-distinguished-names
> https://docs.oracle.com/cd/E19146-01/821-1828/geezw/index.html
> https://docs.microsoft.com/en-us/openspecs/sharepoint_protocols/ms-osco/dbdc3411-ed0a-4713-a01b-1ae0da5e75d4

Interestingly enough, openssl uses "emailAddress". 

I recognize it is cosmetic. So I can tweak my app to show it differently if needed. Thanks for humoring me!

reaperhulk:
Do you mean you're okay with it being either value, or remaining an OID, or...? OpenSSL does indeed use `emailAddress` which is part of why I wanted to dig a bit more.

divinehawk:
> Do you mean you're okay with it being either value, or remaining an OID, or...? OpenSSL does indeed use `emailAddress` which is part of why I wanted to dig a bit more.

"E" would be preferable since it's the closest to being the "de-facto" standard. I think the OpenSSL output is the outlier.

reaperhulk:
@divinehawk are you still interested in this PR? I'm happy with this if you can rebase and run black against it (which is why the flake job failed).

divinehawk:
> @divinehawk are you still interested in this PR? I'm happy with this if you can rebase and run black against it (which is why the flake job failed).

@reaperhulk Done! I think I did that right.

reaperhulk (approved):

PR #6108, opened 2021-06-11
oxidize issuing distribution point

reaperhulk:


src/rust/src/x509.rs
```diff
@@ -633,6 +686,29 @@ fn parse_crl_extension(
             .to_object(py))
     } else if oid == *AUTHORITY_KEY_IDENTIFIER_OID {
         Ok(parse_authority_key_identifier(py, ext_data)?)
+    } else if oid == *ISSUING_DISTRIBUTION_POINT_OID {
+        let reason_bit_mapping = py
+            .import("cryptography.x509.extensions")?
+            .getattr("_REASON_BIT_MAPPING")?;
```
alex, line 692: This logic is already in `parse_crl_entry_extension`, just use that.

src/rust/src/x509.rs
```diff
@@ -633,6 +686,29 @@ fn parse_crl_extension(
...
+        let idp = asn1::parse_single::<IssuingDistributionPoint<'_>>(ext_data)?;
+        let dp = match idp.distribution_point {
```
alex, line 694: Use tuple unpacking here `let (f1, f2) = match` rather than access by `.0` below.

alex:
Almost certainly needs a cargo fmt.

alex (approved):

PR #6109, opened 2021-06-11
oxidize CRLDP and FreshestCRL

reaperhulk:


alex (dismissed):

alex:
We need a vector for `parse_crl_extension` with unknown extension, it's an uncovered path.

src/rust/src/x509.rs
```diff
@@ -587,6 +657,17 @@ fn parse_x509_extension(
             .to_object(py))
     } else if oid == *AUTHORITY_KEY_IDENTIFIER_OID {
         Ok(parse_authority_key_identifier(py, ext_data)?)
+    } else if oid == *CRL_DISTRIBUTION_POINTS_OID {
+        Ok(x509_module
+            .call1(
+                "CRLDistributionPoints",
+                (parse_distribution_points(py, ext_data)?,),
+            )?
```
alex, line 665: This is showing as uncovered for stupid reasons.

Should be fixable by:

```
let dp = parse_distribution_points(py, ext_data)?;
Ok(x509_module.call1("CRLDistributoinPoints, (dp,))?.to_object)
```

alex (approved):

alex:
Need to remove `_decode_general_names` it's no longer used.

alex (approved):

PR #6113, opened 2021-06-12
oxidize nameconstraints

reaperhulk:


alex:
wtf, leading ones was really only stabalized in 1.46...

reaperhulk:
Not sure what the right way forward is here 😂

alex:
Write count-leading-ones/count-trailing-zeros functions in rust I think

alex:
Should be easy to implement and unit test.

src/rust/src/x509.rs
```diff
@@ -48,6 +49,39 @@ impl<'a> asn1::SimpleAsn1Readable<'a> for UnvalidatedIA5String<'a> {
...
+    excluded_subtrees: Option<asn1::SequenceOf<'a, GeneralSubtree<'a>>>,
+}
+
+#[derive(asn1::Asn1Read)]
+struct GeneralSubtree<'a> {
+    base: GeneralName<'a>,
+
+    #[implicit(0)]
+    #[default(0)]
+    // TODO: doesn't like u64
```
alex, line 67: What does "doesn't like" mean?

src/rust/src/x509.rs
```diff
@@ -291,6 +325,55 @@ fn parse_name(py: pyo3::Python<'_>, name: Name<'_>) -> Result<pyo3::PyObject, Py
     Ok(py_name)
 }
 
+fn ipv4_netmask(data: &[u8]) -> Result<u32, PyAsn1Error> {
```
alex, line 328: Make this take a `u32` and do the conversion from `&[u8]` in the caller so it's obviously correct

src/rust/src/x509.rs
```diff
@@ -291,6 +325,55 @@ fn parse_name(py: pyo3::Python<'_>, name: Name<'_>) -> Result<pyo3::PyObject, Py
...
+    if (!num).leading_zeros() + num.trailing_zeros() != 32 {
+        return Err(PyAsn1Error::from(pyo3::exceptions::PyValueError::new_err(
+            "Invalid netmask",
+        )));
+    }
+    Ok((!num).leading_zeros())
+}
+
+fn ipv6_netmask(data: &[u8]) -> Result<u32, PyAsn1Error> {
+    let num = u128::from_be_bytes(data[16..].try_into().unwrap());
```
alex, line 341: Same

alex (approved):

PR #6116, opened 2021-06-13
Adding mypy configuration to pyproject

tanvimoharir:
Refers #6104


alex:
It looks like mypy no longer runs succesfully with this configuration, unfortunately. Perhaps some of the option's names changed?

tanvimoharir:
> It looks like mypy no longer runs succesfully with this configuration, unfortunately. Perhaps some of the option's names changed?

Yeah seems like it doesn't, I will check if it's an issue with option's names.

alex (approved):

PR #6119, opened 2021-06-15
oxidize certificate policies

reaperhulk:


alex:
clippy warnings

src/cryptography/hazmat/backends/openssl/decode_asn1.py
```diff
@@ -111,91 +104,19 @@ def parse(self, x509_obj):
...
-                        "parsed".format(oid)
-                    )
-
-                value = handler(self._backend, ext_data)
-                extensions.append(x509.Extension(oid, critical, value))
+            # Dump the DER payload into an UnrecognizedExtension object
+            # The rust code covers every other extension
+            data = self._backend._lib.X509_EXTENSION_get_data(ext)
+            self._backend.openssl_assert(data != self._backend._ffi.NULL)
+            der = self._backend._ffi.buffer(data.data, data.length)[:]
```
alex, line 111: This is  just `data_bytes` from above

src/rust/src/x509.rs
```diff
@@ -29,13 +29,141 @@ lazy_static::lazy_static! {
...
+    organization: DisplayText<'a>,
+    notice_numbers: asn1::SequenceOf<'a, asn1::BigUint<'a>>,
+}
+
+#[derive(asn1::Asn1Read)]
+enum DisplayText<'a> {
+    IA5String(asn1::IA5String<'a>),
+    UTF8String(asn1::Utf8String<'a>),
+    VisibleString(asn1::VisibleString<'a>),
+}
```
alex, line 80: Comment about BMP string

src/rust/src/x509.rs
```diff
@@ -29,13 +29,147 @@ lazy_static::lazy_static! {
...
+                                Some(DisplayText::IA5String(data)) => {
+                                    pyo3::types::PyString::new(py, data.as_str()).to_object(py)
+                                }
+                                Some(DisplayText::UTF8String(data)) => {
+                                    pyo3::types::PyString::new(py, data.as_str()).to_object(py)
+                                }
+                                Some(DisplayText::VisibleString(data)) => {
+                                    pyo3::types::PyString::new(py, data.as_str()).to_object(py)
+                                }
+                                None => py.None(),
```
alex, line 129: Factor this into a function so that we don't need coverage for each string type in both paths

src/rust/src/x509.rs
```diff
@@ -29,13 +29,137 @@ lazy_static::lazy_static! {
...
+                }
+                py_pq.to_object(py)
+            }
+            None => py.None(),
+        };
+        certificate_policies.append(
+            x509_module
+                .call_method1("PolicyInformation", (pi, py_pqis))?
+                .to_object(py),
+        )?;
```
alex, line 160: I think you can fix this by changing it to:

```rust
let py_pi = x509_module.call_method1("PolicyInformation", (pi, py_pqis))?                .to_object(py);
certificate_policies.append(py_pi)?;
```

alex (commented):
Code LGTM, though I think if this were split into more functions it might be easier to read.

It looks like the next step is a rust-asn1 release at any rate though!

alex (approved):

PR #6123, opened 2021-06-16
 type annotate cryptography.utils:register_interface 

graingert:


alex:
What's the motivation for this? I'm not sure that in practice this could ever catch anything.

graingert:
Without this the decorator strips the type of any class it decorated

alex (approved):

PR #6134, opened 2021-06-21
Add logo

Itotutona:
# New Logo
Here is a new logo for the Pyca organization. What do you think ?? 

pyca

## Info
This is my first contribution to an open source project. If it's not like that we do. Can you tell me? Thank you.


reaperhulk:
Thanks, but we’re not looking for a logo at this time. And if we were, a PR adding a PNG to the root of one of our repositories would not be the correct approach.

Itotutona-old:
Ok

PR #6138, opened 2021-06-26
Expose a few more OpenSSL functions that are useful for DTLS support

njsmith:
Also `OP_NO_RENEGOTIATION` I guess, which is useful because renegotiation is the worst.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -90,6 +90,7 @@ static const long SSL_OP_ALL;
 static const long SSL_OP_SINGLE_ECDH_USE;
 static const long SSL_OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION;
 static const long SSL_OP_LEGACY_SERVER_CONNECT;
+static const long SSL_OP_NO_RENEGOTIATION;
 static const long SSL_VERIFY_PEER;
 static const long SSL_VERIFY_FAIL_IF_NO_PEER_CERT;
 static const long SSL_VERIFY_CLIENT_ONCE;
@@ -225,6 +226,13 @@ void SSL_CTX_set_cookie_generate_cb(SSL_CTX *,
                                         unsigned char *,
                                         unsigned int *
                                     ));
+void SSL_CTX_set_cookie_verify_cb(SSL_CTX *,
+                                    int (*)(
+                                        SSL *,
+                                        unsigned char *,
+                                        unsigned int
+                                    ));
+
 long SSL_CTX_get_read_ahead(SSL_CTX *);
 long SSL_CTX_set_read_ahead(SSL_CTX *, long);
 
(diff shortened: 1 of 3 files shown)


src/_cffi_src/openssl/ssl.py
```diff
@@ -468,6 +476,13 @@
 long DTLSv1_handle_timeout(SSL *);
 long DTLS_set_link_mtu(SSL *, long);
 long DTLS_get_link_min_mtu(SSL *);
+size_t DTLS_get_data_mtu(SSL *);
+long SSL_set_mtu(SSL *, long);
+typedef ... BIO_ADDR;
```
reaperhulk, line 481: This should be in types

reaperhulk: Where are these BIO_ADDR things originally defined? If bio.h they should live in bio.py

njsmith: Fixed

njsmith:
...This thing where each CI run has to be re-approved again is pretty annoying. I guess feel free to add me to the org temporarily? I promise that the only way I will abuse my power would be to kick myself out again ASAP.

alex:
An easier move, more compatible with how we manage permissions generally, would be if you could send some other small PR (docs fix, typo, etc.) and then we could merge that, freeing this up to be run CI jobs without our intervention.

njsmith:
Well, I guess I'll keep an eye out for typos then? :-) In the mean time, I *think* it should pass now, so probably it will only take another 3 attempts...

njsmith:
Well would you look at that. I guess this is ready for review then.

src/_cffi_src/openssl/ssl.py
```diff
@@ -594,6 +614,10 @@
 long (*DTLS_get_link_min_mtu)(SSL *) = NULL;
 #endif
 
+#if CRYPTOGRAPHY_OPENSSL_LESS_THAN_111
+size_t (*DTLS_get_data_mtu)(SSL *) = NULL;
```
reaperhulk, line 618: Take a look in `_conditional.py` to see what we do when a symbol isn't actually available.

src/_cffi_src/openssl/ssl.py
```diff
@@ -556,6 +569,13 @@
 static const long Cryptography_HAS_NEXTPROTONEG = 0;
 static const long Cryptography_HAS_ALPN = 1;
 
+#ifdef SSL_OP_NO_RENEGOTIATION
+static const long Cryptography_HAS_OP_NO_RENEGOTIATION = 1;
+#else
+static const long Cryptography_HAS_OP_NO_RENEGOTIATION = 0;
```
reaperhulk, line 575: This should be used in `_conditional.py` to remove the value when it doesn't exist.

src/_cffi_src/openssl/ssl.py
njsmith: It looks like currently, `_conditional.py` doesn't actually remove any conditionally-available functions (e.g. `DTLS_get_link_min_mtu`, visible just above my new code in the diff, doesn't appear there). So I made my best guess at how it should work for `DTLS_get_data_mtu` -- take a look.

reaperhulk: Ha, I would argue that is a bug (we don't want callables that are null func ptrs!). Looks like that's a thing we need to separately fix for the DTLS bindings though...

alex:
Looks like there's some flake8 issues.

njsmith:
Huh. Codecov output is a bit mangled, but I *think* it might be saying that in `_comparisons.py`, we're never detecting `DTLS_get_data_mtu` as being falsey? That doesn't make sense to me -- we're definitely setting it to NULL on libressl and openssl 1.1.0, and I confirmed locally that CFFI treats null function pointers as falsey (`bool(ffi.cast("void (*)(void)", 0)) == False`).

njsmith:
image

PR #6144, opened 2021-06-26
Remove unused gf2m bindings

alex:


src/_cffi_src/openssl/ec.py
```diff
@@ -68,15 +68,9 @@
 int EC_POINT_set_compressed_coordinates_GFp(const EC_GROUP *, EC_POINT *,
```
reaperhulk, line 68: kill this too?

reaperhulk (approved):

PR #6147, opened 2021-06-28
parse certificates with pure rust

reaperhulk:


CHANGELOG.rst
```diff
@@ -11,6 +11,13 @@ Changelog
 * Changed the :ref:`version scheme <api-stability:versioning>`. This will
   result in us incrementing the major version more frequently, but does not
   change our existing backwards compatibility policy.
+* **BACKWARDS INCOMPATIBLE:** The X.509 certificate parser no longer allows
+  negative serial numbers.
```
alex, line 15: "The RFC has always prohibited these".

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -957,7 +956,7 @@ def create_x509_certificate(
         builder: x509.CertificateBuilder,
         private_key: PRIVATE_KEY_TYPES,
         algorithm: typing.Optional[hashes.HashAlgorithm],
-    ) -> _Certificate:
+    ) -> rust_x509.Certificate:
```
alex, line 959: Should this just be `x509.Certificate` (that is, the interface)?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1333,31 +1332,29 @@ def load_der_parameters(self, data):
...
+    def load_pem_x509_certificate(self, data: bytes) -> x509.Certificate:
+        return rust_x509.load_pem_x509_certificate(data)
 
-        x509 = self._ffi.gc(x509, self._lib.X509_free)
-        return _Certificate(self, x509)
+    def load_der_x509_certificate(self, data: bytes) -> x509.Certificate:
+        return rust_x509.load_der_x509_certificate(data)
 
-    def load_der_x509_certificate(self, data: bytes) -> _Certificate:
+    def _x509_rust2openssl(self, cert):
```
alex, line 1341: I don't think rust shoul be in this method name -- rust is merely an implementation detail.

src/cryptography/hazmat/bindings/_rust/x509.pyi
```diff
@@ -1,15 +1,55 @@
...
+    der_oid: bytes, ext_data: bytes
+) -> x509.ExtensionType: ...
 def encode_precertificate_signed_certificate_timestamps(
-    extension: PrecertificateSignedCertificateTimestamps,
+    extension: x509.PrecertificateSignedCertificateTimestamps,
 ) -> bytes: ...
 
 class Sct: ...
+
+class Certificate(x509.Certificate):
```
alex, line 25: Err, why is this required? Just define the functions as returning `x509.Certificate`

src/rust/Cargo.toml
```diff
@@ -7,8 +7,10 @@ publish = false
 
 [dependencies]
 lazy_static = "1"
+pem = { version = "0.8" }
```
alex, line 10:
```suggestion
pem = "0.8"
```

src/rust/src/x509.rs
```diff
@@ -43,6 +43,454 @@ lazy_static::lazy_static! {
...
+        let mut vec = Vec::new();
+        for rdn in self.raw.borrow_value().tbs_cert.subject.clone() {
+            for attribute in rdn {
+                let mut attr = attribute.type_id.to_string();
+                attr.push('=');
+                attr.push_str(std::str::from_utf8(attribute.value.data()).unwrap());
+                vec.push(attr);
+            }
+        }
+        repr.push_str(&vec.join(", "));
```
alex, line 145: Why make a vec instead of just pushing ino `repr` as you go?

src/rust/src/x509.rs
```diff
@@ -43,6 +43,454 @@ lazy_static::lazy_static! {
...
+        slf: pyo3::pycell::PyRef<'_, Self>,
+        _memo: pyo3::PyObject,
+    ) -> pyo3::pycell::PyRef<'_, Self> {
+        slf
+    }
+
+    fn public_key<'p>(&self, py: pyo3::Python<'p>) -> pyo3::PyResult<&'p pyo3::PyAny> {
+        let serialized = pyo3::types::PyBytes::new(
+            py,
+            &asn1::write_single(&self.raw.borrow_value().tbs_cert.spki),
```
alex, line 163: Can you put a TODO comment on here admonishing us that this performs a copy that shouldn't be required?

src/rust/src/x509.rs
```diff
@@ -43,6 +43,454 @@ lazy_static::lazy_static! {
...
+    fn version<'p>(&self, py: pyo3::Python<'p>) -> Result<&'p pyo3::PyAny, PyAsn1Error> {
+        let x509_module = py.import("cryptography.x509")?;
+        let version = self.raw.borrow_value().tbs_cert.version;
+        if version == 0 {
+            Ok(x509_module.getattr("Version")?.get_item("v1")?)
+        } else {
+            // in the load function we prohibit anything but 0 or 2
+            assert!(version == 2);
+            Ok(x509_module.getattr("Version")?.get_item("v3")?)
+        }
```
alex, line 229: Can you rewrite this as:

```rust
// comment
match version {
    0 => blah,
    2 => blah,
    _ => unreachable!()
```

src/rust/src/x509.rs
```diff
@@ -43,6 +43,454 @@ lazy_static::lazy_static! {
...
+        let time = match &self.raw.borrow_value().tbs_cert.validity.not_before {
+            Asn1Time::UtcTime(data) => data.as_chrono(),
+            Asn1Time::GeneralizedTime(data) => data.as_chrono(),
+        };
+        chrono_to_py(py, time)
+    }
+
+    #[getter]
+    fn not_valid_after<'p>(&self, py: pyo3::Python<'p>) -> pyo3::PyResult<&'p pyo3::PyAny> {
+        let time = match &self.raw.borrow_value().tbs_cert.validity.not_after {
```
alex, line 270: Perhaps an `as_chrono` method on `Time` with this match, since we do it in a few places?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1333,31 +1331,25 @@ def load_der_parameters(self, data):
...
-            mem_bio.bio, self._ffi.NULL, self._ffi.NULL, self._ffi.NULL
-        )
-        if x509 == self._ffi.NULL:
-            self._consume_errors()
-            raise ValueError(
-                "Unable to load certificate. See https://cryptography.io/en/"
-                "latest/faq.html#why-can-t-i-import-my-pem-file for more"
-                " details."
-            )
+    def load_pem_x509_certificate(self, data: bytes) -> x509.Certificate:
```
alex, line 1334: Hmm, why change them at the backend layer vs. doing so at the x509 layer and removing these methods?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1333,31 +1331,25 @@ def load_der_parameters(self, data):
...
+        return rust_x509.load_pem_x509_certificate(data)
 
-        x509 = self._ffi.gc(x509, self._lib.X509_free)
-        return _Certificate(self, x509)
+    def load_der_x509_certificate(self, data: bytes) -> x509.Certificate:
+        return rust_x509.load_der_x509_certificate(data)
 
-    def load_der_x509_certificate(self, data: bytes) -> _Certificate:
+    def _cert2ossl(self, cert):
```
alex, line 1340: typing

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -21,150 +22,24 @@
...
-            raise TypeError("encoding must be an item from the Encoding enum")
-
-        self._backend.openssl_assert(res == 1)
-        return self._backend._read_mem_bio(bio)
+# This exists for pyOpenSSL compatibility and SHOULD NOT BE USED
+# WE WILL REMOVE THIS VERY SOON.
+def _Certificate(backend, x509) -> x509.Certificate:  # noqa: N802
+    warnings.warn(
+        "This version of cryptography contains a temporary pyOpenSSL "
+        "fallback path. Upgrade pyOpenSSL now.",
```
alex, line 39: Do we have a PR up on the pyOpenSSL side?

src/rust/src/x509.rs
```diff
@@ -43,6 +43,505 @@ lazy_static::lazy_static! {
...
+pub(crate) type Name<'a> = asn1::SequenceOf<'a, asn1::SetOf<'a, AttributeTypeValue<'a>>>;
+
+#[derive(asn1::Asn1Read, asn1::Asn1Write)]
+pub(crate) struct AttributeTypeValue<'a> {
+    pub(crate) type_id: asn1::ObjectIdentifier<'a>,
+    pub(crate) value: asn1::Tlv<'a>,
+}
+
+#[derive(asn1::Asn1Read, asn1::Asn1Write)]
+enum Asn1Time {
```
alex, line 83: Why not just `Time`?

src/rust/src/x509.rs
alex: Question still here.

src/rust/src/x509.rs
```diff
@@ -43,6 +43,505 @@ lazy_static::lazy_static! {
...
+            ),
+        )?;
+        let backend = py
+            .import("cryptography.hazmat.backends.openssl.backend")?
+            .getattr("backend")?;
+        Ok(backend.call_method1("_cert2ossl", (slf,))?)
+    }
+}
+
+fn cert_version(py: pyo3::Python<'_>, version: u8) -> Result<&'_ pyo3::PyAny, PyAsn1Error> {
```
alex, line 510:
```suggestion
fn cert_version(py: pyo3::Python<'_>, version: u8) -> Result<&pyo3::PyAny, PyAsn1Error> {
```

Should be sufficient?

tests/hazmat/backends/test_openssl.py
```diff
@@ -609,30 +609,6 @@ def test_password_length_limit(self):
             )
 
 
-class TestGOSTCertificate(object):
```
alex, line 612: What changed here?

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -21,150 +22,24 @@
...
+        utils.CryptographyDeprecationWarning,
```
alex, line 40: This should be a versioned warning!

tests/hazmat/backends/test_openssl.py
reaperhulk: We support this string type on all paths and it's not an openssl specific test (so it's now tested in test_x509)

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: Not yet, no.

src/cryptography/hazmat/backends/openssl/x509.py
```diff
@@ -21,150 +22,24 @@
     _encode_asn1_int_gc,
     _txt2obj_gc,
 )
+from cryptography.hazmat.bindings._rust import x509 as rust_x509
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.x509.base import PUBLIC_KEY_TYPES
 from cryptography.x509.name import _ASN1Type
 
+# Runtime isinstance checks need this since the rust class is not a subclass.
+x509.Certificate.register(rust_x509.Certificate)
```
alex, line 31: I think maybe this registration should be in the x509 module where `Certificate` is defined?

src/rust/src/x509.rs
```diff
@@ -43,6 +43,502 @@ lazy_static::lazy_static! {
...
+    }
+
+    fn __richcmp__(
+        &self,
+        other: pyo3::pycell::PyRef<Certificate>,
+        op: pyo3::class::basic::CompareOp,
+    ) -> pyo3::PyResult<bool> {
+        match op {
+            pyo3::class::basic::CompareOp::Eq => {
+                Ok(self.raw.borrow_data() == other.raw.borrow_data())
```
alex, line 132: Mmm, is comparing by bytes always correct, or are there ever cases where different bytes are really the same?

src/rust/src/x509.rs
```diff
@@ -43,6 +43,502 @@ lazy_static::lazy_static! {
...
+        )));
+    }
+    load_der_x509_certificate(py, &parsed.contents)
+}
+
+#[pyo3::prelude::pyfunction]
+fn load_der_x509_certificate(py: pyo3::Python<'_>, data: &[u8]) -> PyAsn1Result<Certificate> {
+    let raw = OwnedRawCertificate::try_new(data.to_vec(), |data| asn1::parse_single(data))?;
+    // Parse cert version immediately so we can raise error on parse if it is invalid.
+    cert_version(py, raw.borrow_value().tbs_cert.version)?;
```
alex, line 535: Should we also check that there's no extensions if version is 1?

src/rust/src/x509.rs
```diff
@@ -945,9 +1441,6 @@ fn parse_x509_extension(py: pyo3::Python<'_>, der_oid: &[u8], ext_data: &[u8]) -
             .getattr("PolicyConstraints")?
             .call1((pc.require_explicit_policy, pc.inhibit_policy_mapping))?
             .to_object(py))
-    } else if oid == *PRECERT_POISON_OID {
```
alex, line 948: Should this function be renamed `parse_csr_extension` given that that's not what it's for?

src/cryptography/hazmat/backends/openssl/x509.py
reaperhulk: https://github.com/pyca/pyopenssl/pull/1030

src/rust/src/x509.rs
reaperhulk: We can rename it, sure. As we convert CSR we'll rip this out entirely but can't do that yet so might as well name it accurately.

src/rust/src/x509.rs
reaperhulk: I have no opinion. We did not do that check previously.

alex: Did OpenSSL not do that check previously either? If not, I guess I don't care.

src/rust/src/x509.rs
reaperhulk: https://github.com/openssl/openssl/blob/10af976962b2383bb3044120a764037361b8bff7/crypto/x509/x509_cmp.c#L155

OpenSSL generates a hash and verifies that the "encoding" hasn't changed. Digging through the layers it looks like byte for byte comparison probably is fine? It's also the case that the only bytes that could ever be different would be the ones outside of tbsCertificateBytes of course.

alex (approved):

intgr:
Hi! I since the changelog drew attention to it, I just happened to read the [API stability and versioning doc](https://github.com/pyca/cryptography/blob/4aabad3dbf101810323a0d19d71a6970203375f6/docs/api-stability.rst):
> * The behavior of existing APIs will not change.
>   * One exception to our API stability policy is for security.
>
> Deprecation
> * In cryptography X.0.0 the feature exists.
> * In cryptography (X + 1).0.0 using that feature will emit a UserWarning.

However, this change breaks "the behavior of existing APIs" with no warning. And I don't think it's a "security" fix either -- while stricter validation can improve security, that reasoning would be a slippery slope by which all sorts of undue API breaks could be justified.

What's the cryptography project's opinion on this? It seems to me that the document is simply overpromising on stability. Maybe it's time to change the stability claims?

reaperhulk:
Stricter and looser validation are properties that we have historically flexed on (sometimes due to our own dependency changing its behavior, sometimes for other reasons). I'm not sure how to capture the reality of our policy without making it sound like we break APIs frequently rather than "we have changed an unusual parsing/serialization edge case that 99.99% of our callers will never see".

PR #6166, opened 2021-07-11
Add RFC4514 parser

### LINKED ISSUE #6165: RFC4514 parser?
I'm wondering why there is no RFC4514 parser in cryptography to parse str represenations of a DN into `x509.Name`, e.g.:

```
>>> x509.Name.from_rfc4514_string("CN=example.com,C=AT")
>>> <Name(C=AT,CN=example.com)>
```

I would volunteer for a PR if you're open to adding that functionality to cryptography. (I just don't want to put in all that work and then you say you don't want it for some reason ;-)).

mathiasertl:
This PR adds a RFC4514 parser.

The implementation uses [shlex.shlex](https://docs.python.org/3/library/shlex.html) parsers to take care of quoting and escaping issues. I've added quite a few tests and I think the approach is maybe not perfect, but reasonably safe.

Known limitations:

1. The parser does not understand unicode characters (like in the example given in RFC 4514, section 4 in the last example). However, neither does the serialization already implemented, so I think it's a fair limitation that may be fixed in a separate pull request.
2. Using `shlex` makes it possible to quote strings, so `CN="Foo,Bar"` will work as a user would probably expect (with the CN being `Foo,Bar` without quotes), but the string is not strictly RFC4514 compatible anyway. The standard would mandate `"` to be escaped, but this would also change the semantics, as it would make them part of the CN.

Fixes #6165.


src/cryptography/x509/name.py
```diff
@@ -138,6 +164,23 @@ def rfc4514_string(self) -> str:
...
+        """
+        attr_type, attr_value = value.split("=", 1)
+
+        # if a descr, it's case insensitive (RFC 4512, section 1.4)
+        attr_type = attr_type.upper()
+        attr_value = _parse_dn_value(attr_value)
+
+        if attr_type in _NAME_TO_NAMEOID:
+            return NameAttribute(_NAME_TO_NAMEOID[attr_type], attr_value)
+        elif re.match(r"[0-9]+(\.[0-9]+)*$", attr_type):
```
alex, line 180: All regexps should be made into constants with `re.compile`.

src/cryptography/x509/name.py
```diff
@@ -138,6 +164,23 @@ def rfc4514_string(self) -> str:
...
+            return NameAttribute(ObjectIdentifier(attr_type), attr_value)
+        raise ValueError("%s: Could not parse attribute" % value)
```
alex, line 182: In general we use `str.format` for formatting values, not the `%` formatting.

src/cryptography/x509/name.py
```diff
@@ -253,6 +304,45 @@ def rfc4514_string(self) -> str:
...
+        Parse a RFC4514 formatted Distinguished Name string.
+        """
+        # value must start with a attributeType, which is either a descr or a
+        # numericoid. A descr is defined in RFC4512 as starting with a
+        # alphabetic char (case insensitive), numeric OIDs start with a digit.
+        if not re.match("[A-Za-z0-9]", value):
+            raise ValueError("Value does not conform to RFC 4514")
+
+        # Split the value by commas into a list of names
+        lex = shlex.shlex(value, posix=True)
```
alex, line 319: I'm nervous about using `shlex` here. For obvious reasons this isn't what it's designed for, so I fear there's likely to be behavioral differences in places.

alex (commented):
A few initial comments, but hardly a comprehensive review.

src/cryptography/x509/name.py
mathiasertl: fixed. By the way, the serialization function just above also uses `%`.

src/cryptography/x509/name.py
mathiasertl: agreed, fixed.

src/cryptography/x509/name.py
mathiasertl: Well, there's quite a few test cases added in this PR. `shlex` docs btw state:

> This will often be useful for writing minilanguages, (for example, in run control files for Python applications) or for parsing quoted strings.

So that seems fair enough. A more sophisticated parser (duplicating code from shlex) is certainly doable, but a lot more code.

mathiasertl:
Hi @alex, 

Thanks for the review so far. I solved the two minor issues you mentioned (string formatting, compiled regex).

I don't know how "nervous" you're about using `shlex`. I could implement a manual parsing solution, but I'm sure it would be significantly more code.

kr, Mat

src/cryptography/x509/name.py
```diff
@@ -47,6 +49,12 @@ class _ASN1Type(utils.Enum):
     NameOID.DOMAIN_COMPONENT: "DC",
     NameOID.USER_ID: "UID",
 }
+#: Mapping of short names to from RFC 4514 to NameOIDs
```
pprindeville, line 52: "to from"?  Which is it?

src/cryptography/x509/name.py
```diff
@@ -73,6 +81,28 @@ def _escape_dn_value(val: str) -> str:
     return val
 
 
+def _parse_dn_value(val: str) -> str:
+    if not val:
+        return ""
+
+    # See https://tools.ietf.org/html/rfc4514#section-2.4
+    val = val.replace("\\\\", "\\")
```
pprindeville, line 89: Shouldn't this be last?

src/cryptography/x509/name.py
```diff
@@ -138,6 +168,23 @@ def rfc4514_string(self) -> str:
             _escape_dn_value(self.value),
         )
 
+    @classmethod
+    def from_rfc4514_string(cls, value: str) -> "NameAttribute":
+        """
+        Parse a RFC4514 formatted Distinguished Name string.
```
pprindeville, line 174: "Parse an RFC 4514 ..."

src/cryptography/x509/name.py
```diff
@@ -138,6 +168,23 @@ def rfc4514_string(self) -> str:
...
+        """
+        attr_type, attr_value = value.split("=", 1)
```
pprindeville, line 176: What if the `=` is inside a quoted string like `CN="Joe = Public",...`

src/cryptography/x509/name.py
```diff
@@ -138,6 +168,23 @@ def rfc4514_string(self) -> str:
...
+
+        # if a descr, it's case insensitive (RFC 4512, section 1.4)
+        attr_type = attr_type.upper()
+        attr_value = _parse_dn_value(attr_value)
+
+        if attr_type in _NAME_TO_NAMEOID:
+            return NameAttribute(_NAME_TO_NAMEOID[attr_type], attr_value)
+        elif _OID_RC.match(attr_type):
+            return NameAttribute(ObjectIdentifier(attr_type), attr_value)
+        raise ValueError("{0}: Could not parse attribute".format(value))
```
pprindeville, line 186: I'd insert a blank link before here so this stands out... or else insert `else:`

tests/x509/test_x509.py
```diff
@@ -4740,6 +4765,70 @@ def test_rfc4514_string_empty_values(self):
...
+            [
+                x509.NameAttribute(NameOID.DOMAIN_COMPONENT, "example"),
+                x509.NameAttribute(NameOID.COMMON_NAME, 'Jane "J,S" Smith'),
+                x509.NameAttribute(NameOID.COUNTRY_NAME, "US"),
+            ]
+        )
+        assert n.rfc4514_string() == name
+
+        # Not strictly part of the spec, but simple quotes work too:
+        name = 'C=US,CN="Jane J,S Smith",DC=example'
```
pprindeville, line 4795: Test also for `=` inside quotes...

pprindeville (changes requested):
Minor comments...

src/cryptography/x509/name.py
mathiasertl: fixed grammar, thanks!

src/cryptography/x509/name.py
mathiasertl: No, however the link above should link to section 3, not 2.4, which I've updated. In there, it states:

> If in <string> form, a LDAP string representation asserted value can
> be obtained by replacing (left to right, non-recursively) each <pair>
> appearing in the <string> as follows:
> 
>     replace <ESC><ESC> with <ESC>;
>     replace <ESC><special> with <special>;
>     replace <ESC><hexpair> with the octet indicated by the <hexpair>.

src/cryptography/x509/name.py
mathiasertl: fixed, thanks

src/cryptography/x509/name.py
mathiasertl: Not a problem due to the `maxsplit` argument. However it will raise an unhandled ValueError if `value` happens to contain no `=` at all. I cought the error and raised an appropriate error.

tests/x509/test_x509.py
mathiasertl: Done, and also for the case with no `=`, no surprises here.

src/cryptography/x509/name.py
pprindeville: That's counter to the way that escaping is handled in most protocols.

alex:
First, I'm sorry we've sat on this for so long without a forward direction :-(

After long review, I've a) gotten comfortable with the idea of us including this API (despite how much I hate the format :-)), b) I'm not comfortable with `shlex` and would prefer a dedicated parser for this format that's easier to ensure matches the RFC.

If you're still up for working on this, awesome, otherwise I can port this away from `shlex`. I'm hoping we can include this in our next release.

PR #6173, opened 2021-07-14
[hack] see if putting this on one line fixes coverage

alex:


src/rust/src/asn1.rs
```diff
@@ -34,6 +34,8 @@ impl From<PyAsn1Error> for pyo3::PyErr {
     }
 }
 
+pub(crate) type PyAsn1Result<T = pyo3::PyObject> = Result<T, PyAsn1Error>;
```
reaperhulk, line 37: Could you add a comment noting that this type alias exists solely to shorten lines so we can avoid a rust coverage bug? I'd like to eventually be able to remove this.

alex: Done

reaperhulk (approved):

PR #6180, opened 2021-07-27
build macos and windows pypy wheels

reaperhulk:


.github/workflows/wheel-builder.yml
```diff
@@ -24,7 +25,11 @@ jobs:
     steps:
       - run: ${{ matrix.PYTHON.PATH }} -m venv .venv
       - name: Install Python dependencies
-        run: .venv/bin/pip install -U pip wheel cffi setuptools-rust
+        run: .venv/bin/pip install -U pip wheel setuptools-rust
+      - name: Install cffi
+        run: .venv/bin/pip install -U cffi
+        if: contains(matrix.PYTHON.VERSION, 'pypy') == false
```
alex, line 31: Why is this required?

.github/workflows/wheel-builder.yml
```diff
@@ -121,6 +137,10 @@ jobs:
           - {ARCH: 'x64', WINDOWS: 'win64', RUST_TRIPLE: 'x86_64-pc-windows-msvc'}
         PYTHON:
           - {VERSION: "3.8", MSVC_VERSION: "2019", "ABI_VERSION": "cp36"}
+          - {VERSION: "pypy-3.7", MSVC_VERSION: "2019"}
+        exclude:
```
alex, line 141: Comment explaining why

.github/workflows/wheel-builder.yml
reaperhulk: It may not be but cffi is still built into pypy so installing another version is, at best, a no op right?

alex: Yes, but we were producing pypy wheels already here :-)

reaperhulk: Just because it was working doesn't make it correct 😄. I can revert this if you prefer, but unless things have changed with cffi (which maybe it has!) installing cffi on pypy historically could potentially break the built-in cffi.

alex: I have a soft preference for reverting, on readability grounds.

alex (approved):

PR #6181, opened 2021-07-28
Implement KBKDFCMAC

kjackiewicz:


kjackiewicz:
I didn't change the line below:
https://github.com/pyca/cryptography/pull/6181/checks?check_run_id=3183309153 -> CHANGELOG.rst:29: D001 Line too long

kjackiewicz:
recheck

src/cryptography/hazmat/primitives/kdf/kbkdf.py
```diff
@@ -27,10 +37,9 @@ class CounterLocation(utils.Enum):
     AfterFixed = "after_fixed"
 
 
-class KBKDFHMAC(KeyDerivationFunction):
+class _KBKDFBase(KeyDerivationFunction):
```
reaperhulk, line 40: We strongly attempt to avoid subclassing and inheritance (we had none at all until mypy forced us to stop using ABC registration). There's a lot shared here, but I think we should avoid calls to `super()`.

Let's move the shared logic from this into functions and then have two separate classes `KBKDFHMAC(KeyDerivationFunction)` and `KBKDFCMAC(KeyDerivationFunction)` that call those functions to reduce duplication.

There will be some duplication in `__init__` with this approach, but overall I'd prefer it.

reaperhulk (changes requested):
Thanks for working on this!

src/cryptography/hazmat/primitives/kdf/kbkdf.py
kjackiewicz: Ok, I'll rewrite it as you suggest. I have some other stuff to handle next week but after that I can get back to this PR. If you have any other comments please share them.

tests/hazmat/primitives/utils.py
```diff
@@ -422,24 +423,9 @@ def kbkdf_counter_mode_test(backend, params):
...
 
-    supported_counter_locations = {
-        "before_fixed": CounterLocation.BeforeFixed,
-        "after_fixed": CounterLocation.AfterFixed,
-    }
-
-    algorithm = supported_algorithms.get(params.get("prf"))
-    if algorithm is None or not backend.hmac_supported(algorithm()):
-        pytest.skip(
-            "KBKDF does not support algorithm: {}".format(params.get("prf"))
```
reaperhulk, line 433: I don't entirely follow why this is no longer a relevant skip. What captures this now?

reaperhulk (commented):
Could you also rebase this against the latest main? Getting close!

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -755,6 +755,143 @@ KBKDF
         ``key_material`` generates the same key as the ``expected_key``, and
         raises an exception if they do not match.
 
+.. class:: KBKDFCMAC(algorithm, mode, length, rlen, llen, location,\
+           label, context, fixed, backend=None)
+
+    .. versionadded:: 35.0
```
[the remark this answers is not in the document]
reaperhulk: Nope, 35.0 is the next release so if this lands before we release this is the correct value.

src/cryptography/hazmat/primitives/kdf/kbkdf.py
reaperhulk: I'm okay with this approach.

src/cryptography/hazmat/primitives/kdf/kbkdf.py
```diff
@@ -153,6 +144,129 @@ def _generate_fixed_input(self) -> bytes:
...
+        backend: typing.Optional[Backend] = None,
+    ):
+        backend = _get_backend(backend)
+        if not isinstance(backend, CMACBackend):
+            raise UnsupportedAlgorithm(
+                "Backend object does not implement CMACBackend.",
+                _Reasons.BACKEND_MISSING_INTERFACE,
+            )
+
+        if not issubclass(
```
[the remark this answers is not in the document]
kjackiewicz: In case of CMAC we won't have any instance until we get the key material. Here, the algorithm is a class as in KBKDFCMAC(algorithms.AES, ... )

src/cryptography/hazmat/primitives/kdf/kbkdf.py
```diff
@@ -88,19 +79,19 @@ def __init__(
...
         self._llen = llen
         self._location = location
         self._label = label
         self._context = context
-        self._backend = backend
         self._used = False
         self._fixed_data = fixed
 
-    def _valid_byte_length(self, value: int) -> bool:
+    @staticmethod
```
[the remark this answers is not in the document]
kjackiewicz: No. That was the only reason.

tests/hazmat/primitives/utils.py
kjackiewicz: Test vectors contain only KBKDFHMAC and KBKDFCMAC test data. Before this PR, the latter triggered this condition. Now, that all algorithms are covered, "prf" checks have to be replaced by asserts (lines 427, 428, 459, 500). Otherwise, codecov would complain.

kjackiewicz:
> Could you also rebase this against the latest main? Getting close!

Done.

reaperhulk (approved):

PR #6186, opened 2021-07-31
Reuse the ctx with ChaCha20Poly1305

bdraco:
- When profiling camera snapshots from `HAP-python`, which splits
  each frame into 1024 encrypted blocks, I noticed ~30% of the time
  was spent in `_aead_setup`. ChaCha20Poly1305 can reuse the ctx
  so we only have to set the `nonce` which 1/3 the time of `_aead_setup`.
  The overall speed up is roughly 20%
  
Before:
Screen Shot 2021-07-31 at 12 38 24 PM

After:
Screen Shot 2021-07-31 at 12 38 19 PM

### DIFF
--- a/src/cryptography/hazmat/primitives/ciphers/aead.py
+++ b/src/cryptography/hazmat/primitives/ciphers/aead.py
@@ -13,6 +13,9 @@ from cryptography.hazmat.backends.openssl.backend import backend
 
 class ChaCha20Poly1305(object):
     _MAX_SIZE = 2 ** 32
+    _KEY_LEN = 32
+    _NONCE_LEN = 12
+    _TAG_LENGTH = 16
 
     def __init__(self, key: bytes):
         if not backend.aead_cipher_supported(self):
@@ -22,14 +25,17 @@ class ChaCha20Poly1305(object):
             )
         utils._check_byteslike("key", key)
 
-        if len(key) != 32:
+        if len(key) != self._KEY_LEN:
             raise ValueError("ChaCha20Poly1305 key must be 32 bytes.")
 
+        self._cipher_name = aead._aead_cipher_name(self)
         self._key = key
+        self._decrypt_ctx = None
+        self._encrypt_ctx = None
 
     @classmethod
     def generate_key(cls) -> bytes:
-        return os.urandom(32)
+        return os.urandom(ChaCha20Poly1305._KEY_LEN)
 
     def encrypt(
         self,
@@ -37,6 +43,15 @@ class ChaCha20Poly1305(object):
         data: bytes,
         associated_data: typing.Optional[bytes],
     ) -> bytes:
+        if not self._encrypt_ctx:
+            self._encrypt_ctx = aead._aead_setup_with_fixed_nonce_len(
+                backend,
+                self._cipher_name,
+                self._key,
+                self._NONCE_LEN,
+                aead._ENCRYPT,
+            )
+
         if associated_data is None:
             associated_data = b""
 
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/primitives/ciphers/aead.py
```diff
@@ -22,21 +25,33 @@ def __init__(self, key: bytes):
...
         associated_data: typing.Optional[bytes],
     ) -> bytes:
+        if not self._encrypt_ctx:
+            self._encrypt_ctx = aead._aead_setup_with_fixed_nonce_len(
+                backend,
+                self._cipher_name,
+                self._key,
+                self._NONCE_LEN,
+                aead._ENCRYPT,
+            )
```
alex, line 53: This is not safe in the face of concurrent users of it. This doesn't even require multi-threading for it to occur. All it woul take is:


```py
e1 = ctx.encryptor()
e2 = ctx.encryptor()
e1.update(data1)
e2.update(data2)
```

bdraco: Thanks for the quick feedback. The specific use case I've mentioned in the PR is from `asyncio` and the `nonce` is being incremented each time so there isn't a concurrency concern.

I'm not very familiar with this code base. Hopefully you can help me out with suggestion on how to proceed? Thanks!

reaperhulk:
So if I understand correctly the usage pattern here is something like:
```python
# Pseudocode obviously
chacha = ChaCha20Poly1305(key)
for pt, nonce in blocks:
   chacha.encrypt(pt, nonce)
```
With the desire being to speed up repeated invocations of encrypt or decrypt against the same ChaCha20Poly1305 object. Is that correct?

Reuse of an underlying cipher context is possible in this scenario, but it is difficult to manage it on the caller's behalf. If we're going to do that, I think the only safe place would be to initialize it in `__init__`, although perhaps others can think of something more elegant.

We could also have optional args to our APIs here that allow for passing a ctx as part of an encrypt call, but I wouldn't want to do that unless there's no other choice.

bdraco:
> So if I understand correctly the usage pattern here is something like:
> 
> ```python
> # Pseudocode obviously
> chacha = ChaCha20Poly1305(key)
> for pt, nonce in blocks:
>    chacha.encrypt(pt, nonce)
> ```

That's exactly right.  For reference https://github.com/ikalchev/HAP-python/blob/dev/pyhap/hap_crypto.py#L108

> 
> With the desire being to speed up repeated invocations of encrypt or decrypt against the same ChaCha20Poly1305 object. Is that correct?

Yes
> 
> Reuse of an underlying cipher context is possible in this scenario, but it is difficult to manage it on the caller's behalf. If we're going to do that, I think the only safe place would be to initialize it in `__init__`, although perhaps others can think of something more elegant.

I was thinking the same thing, but I usually bend over backwards to avoid any side effects in `__init__`. It may not be possible otherwise here.

> 
> We could also have optional args to our APIs here that allow for passing a ctx as part of an encrypt call, but I wouldn't want to do that unless there's no other choice.

reaperhulk:
We could add a new API specifically for performance here, but I think the `__init__` path is our best bet. That said, it's maybe worth experimenting with a few different approaches and use your benchmark (could you share that?) to make some decisions with data.

For example, if you wrapped the lazy context initialization in a threading.Lock context manager to ensure it is thread safe, what does that do to perf?

reaperhulk:
Since there's been no progress here for quite some time I'm going to close this, but for anyone who sees this we're happy to take PRs for performance improvements, but we'll want to make sure it doesn't cause other bugs and also have some metrics to define the improvement.

reaperhulk:
#7601 is the successor to tha

PR #6187, opened 2021-07-31
add PKCS7_get0_signers(), X509_STORE_set_purpose(), and X509_PURPOSE_*

panzi:
Add functions and constants referenced in issue https://github.com/pyca/pyopenssl/issues/1031

### DIFF
--- a/src/_cffi_src/openssl/pkcs7.py
+++ b/src/_cffi_src/openssl/pkcs7.py
@@ -68,6 +68,9 @@ int PKCS7_final(PKCS7 *, BIO *, int);
 int PKCS7_verify(PKCS7 *, Cryptography_STACK_OF_X509 *, X509_STORE *, BIO *,
                  BIO *, int);
 PKCS7 *SMIME_read_PKCS7(BIO *, BIO **);
+Cryptography_STACK_OF_X509 *PKCS7_get0_signers(PKCS7 *p7,
+                                               Cryptography_STACK_OF_X509 *certs,
+                                               int flags);
 
 int PKCS7_type_is_signed(PKCS7 *);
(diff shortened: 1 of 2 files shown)


src/_cffi_src/openssl/pkcs7.py
```diff
@@ -68,6 +68,9 @@
 int PKCS7_verify(PKCS7 *, Cryptography_STACK_OF_X509 *, X509_STORE *, BIO *,
                  BIO *, int);
 PKCS7 *SMIME_read_PKCS7(BIO *, BIO **);
+Cryptography_STACK_OF_X509 *PKCS7_get0_signers(PKCS7 *p7,
```
alex, line 71: Functions should have their names removed.

alex (dismissed):

reaperhulk:
If you wouldn't mind adding a comment above each of these blocks noting what this is for that would be helpful. When we curate our bindings (for compatibility with various versions of OpenSSL and its forks) we might accidentally remove these if we don't understand the consumers. Once that's done I think this is ready for merge.

reaperhulk:
(Reminder that this PR needs the comments I requested and then can be merged)

alex (approved):

schwabe:
As a side note if you need that feature in an earlier version of cryptography you can use  `X509_VERIFY_PARAM_set_purpose` instead if you hardcode the ints of the purpose and only need to set the purpose:

        param = _lib.X509_VERIFY_PARAM_new()
        param = _ffi.gc(param, _lib.X509_VERIFY_PARAM_free)
        _lib.X509_VERIFY_PARAM_set_purpose(param, purpose)
        _openssl_assert(_lib.X509_STORE_set1_param(self._store, param) != 0)

PR #6190, opened 2021-08-05
build wheel for pyston

aisk:
[pyston](https://github.com/pyston/pyston) is an alternative python implementation like pypy, it's a fork of cpython but have changed the ABI, so we need build an alternative wheel.

The pyston binary is installed via this PR: https://github.com/pyca/infra/pull/364

### DIFF
--- a/.github/workflows/wheel-builder.yml
+++ b/.github/workflows/wheel-builder.yml
@@ -16,6 +16,7 @@ jobs:
           - { VERSION: "cp36-cp36m", PATH: "/opt/python/cp36-cp36m/bin/python", ABI_VERSION: 'cp36' }
           - { VERSION: "pypy3.6", PATH: "/opt/pypy3.6/bin/pypy" }
           - { VERSION: "pypy3.7", PATH: "/opt/pypy3.7/bin/pypy" }
+          - { VERSION: "pyston3.8", PATH: "/opt/pyston2.3/pyston" }
         MANYLINUX:
           - { NAME: "manylinux2010_x86_64", CONTAINER: "cryptography-manylinux2010:x86_64" }
           - { NAME: "manylinux2014_x86_64", CONTAINER: "cryptography-manylinux2014:x86_64" }


reaperhulk:
We won't build a wheel for platforms we don't test. A prerequisite for building wheels here is going to be adding this to our CI matrix first.

However, one of the biggest challenges in this project is that our CI cycle time can be problematic. We've spent an immense amount of effort keeping the cycle time low, and adding additional CI jobs is something we do only with an eye towards the tradeoffs.

I'm reluctant to add Pyston here since, to my knowledge, it is extremely niche at this time. Does PyPI even allow uploading of Pyston tags? What's usage look like right now?

aisk:
> Does PyPI even allow uploading of Pyston tags?

Yes, I created a test package on test pypi server: https://test.pypi.org/project/lunzi/0.0.1/#files . Upload and install it just works like normal python's wheel packages.

We're trying to using pyston in a large company which have thousands of projects. Pyston is highly compatible with CPython, but the bigest challenge is the installation compality. Without wheel packges, pip will build packages from souce, this works fine, but many packages have third party dependencies, like cryptography need rust toolchain. This caused many manual works for developers.

What's more, build packages from source will increase build time, like cryptography need minutes to be build, instead of seconds when having a prebuild wheel binary.

> We won't build a wheel for platforms we don't test. A prerequisite for building wheels here is going to be adding this to our CI matrix first.

I would like to finish the work, if it's ok to add pyston wheel packages on pypi.

reaperhulk:
Is your organization not running its own package repository? That would be the logical place to upload pyston packages. Your org can then build whatever packages are needed and upload them to your internal system.

I appreciate that you want to contribute this work upstream so that others can benefit, but the additional burden on this project is more than the value gained. As Pyston becomes more popular we can potentially revisit the choice, but for now we aren't going to add this.

aisk:
Thanks for suggestion!

PR #6192, opened 2021-08-08
Test with 3.0.0-beta2

tiran:
- OpenSSL 3.0.0-beta2 now uses lib64 on X86_64
- fail on implicit function definition


reaperhulk:
We don't want to modify the wheel builder yet since we're not shipping 3.0.0, but we'll have to remember to do the lib -> lib64 change when we do move to 3.0.0 in our release binaries.

reaperhulk (approved):

PR #6193, opened 2021-08-08
remove twisted from downstream testing for now

reaperhulk:
spurious failures and no resolution at this time.


alex (approved):
Please file a follow up ticket about reenabling

PR #6204, opened 2021-08-10
support arm64 in GHA

reaperhulk:


.github/workflows/wheel-builder.yml
```diff
@@ -58,6 +58,46 @@ jobs:
           name: "cryptography-${{ github.event.inputs.version }}-${{ matrix.MANYLINUX.NAME }}-${{ matrix.PYTHON.VERSION }}"
           path: cryptography-wheelhouse/
 
+  manylinux_arm64:
```
alex, line 61: I don't think we should build wheels from this, since it's a stateful machine with no configuration management.

.github/workflows/ci.yml
```diff
@@ -158,6 +158,42 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        IMAGE:
+          - {IMAGE: "ubuntu-focal:aarch64", TOXENV: "py38"}
+    name: "${{ matrix.IMAGE.TOXENV }} on ${{ matrix.IMAGE.IMAGE }}"
+    timeout-minutes: 20
+    steps:
+      - name: "Delete workspace" # self-hosted runners need this, sigh
+        run: find ! -name '.' ! -name '..' -delete
```
alex, line 173: Are there any docs on the need for this?

reaperhulk: There's a variety of support threads and actions people have built for this, but https://github.community/t/how-to-properly-clean-up-self-hosted-runners/128909 is an example. GitHub's docs don't bother mentioning it, although they say self-hosted runners "Don't need to have a clean instance for every job execution." (https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners)

.github/workflows/ci.yml
```diff
@@ -158,6 +158,42 @@ jobs:
...
+  linux-arm64:
+    runs-on: [self-hosted, linux, ARM64]
+    container: ghcr.io/pyca/cryptography-runner-${{ matrix.IMAGE.IMAGE }}
+    strategy:
+      fail-fast: false
+      matrix:
+        IMAGE:
+          - {IMAGE: "ubuntu-focal:aarch64", TOXENV: "py38"}
+    name: "${{ matrix.IMAGE.TOXENV }} on ${{ matrix.IMAGE.IMAGE }}"
+    timeout-minutes: 20
```
alex, line 170: So nothing here is conditional on it being you or I :-) What's the security story with this running for everyone?

The answer is that anyone could run a malicious payload and escape the container. What do they get access to if they do that? Do they get a priviledged github token?

reaperhulk: They definitely get a GH token. It's not clear to me what permissions it has. Does `permissions: read-all` constrain it usefully? Can any user just change that as part of the PR? Do we consider that risk low given the workflow approval requirement for new users?

alex: The approval requirement only applies to brand new github accounts, so it's a very weak control.

I'm super nervous because I do not understand the security properties of this.

As an alternative idea, what do you think about making this a 2nd workflow file which only runs on `main`, just so we can see it working?

reaperhulk: I originally had it that way actually. I can also add github.actor restrictions (`if: github.actor == "alex" || github.actor == "reaperhulk"`), but that restriction is in the yaml itself so it seems highly likely that a user can just...remove it.

alex (approved):

PR #6206, opened 2021-08-10
Speed up RSA tests in 3.0.0

reaperhulk:
`RSA_check_key` is slower in OpenSSL 3.0.0 due to improved primality checking. In normal use this is unlikely to be a problem since users don't load new keys constantly, but we do in our tests. This adds some private flags to allow skipping those checks for performance reasons.

On my laptop with this patch it takes 16s to run test_rsa.py. The previous commit takes 72s.

### DIFF
--- a/tests/hazmat/primitives/test_rsa.py
+++ b/tests/hazmat/primitives/test_rsa.py
@@ -67,15 +67,6 @@ class DummyMGF(object):
     _salt_length = 0
 
 
-@pytest.fixture
-def disable_rsa_checks(backend):
-    # Use this fixture to skip RSA key checks in tests that need the
-    # performance.
-    backend._rsa_skip_check_key = True
-    yield
-    backend._rsa_skip_check_key = False
-
-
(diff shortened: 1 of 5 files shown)


tests/hazmat/primitives/test_rsa.py
```diff
@@ -67,6 +67,15 @@ class DummyMGF(object):
     _salt_length = 0
 
 
+@pytest.fixture
+def disable_rsa_checks(backend):
+    # Use this fixture to skip RSA key checks in tests that need the
+    # performance.
+    backend._rsa_skip_check_key = True
+    yield
+    backend._rsa_skip_check_key = False
```
alex, line 76:
```suggestion
    try:
        yield
    finally:
        backend._rsa_skip_check_key = False
```

reaperhulk: pytest does magic here and exceptions inside the yield are caught and the statements after are executed, so there's no need for a try/finally here.

alex (approved):

PR #6207, opened 2021-08-10
[Docs] Replace broken viewcode with linkcode

gpotter2:
As of right now, the `viewcode` extension (which this PR replaces) does nothing in cryptography's doc (should be adding `[source]` tags), probably because it tries to import an unbuilt cryptography (which won't work without the rust bindings..).
This PR tries to fix it, and also use the `linkcode` extension instead to link to github rather than a copy of the source code.

- Tries to compile cryptography during the `readthedocs` install
- Moves documentation extensions to a `_ext` folder because there are now 2.
- Uses a cool trick (heavily inspired by numpy, modified to fit cryptography) to display inline links directly to the source code:

Clicking the `[source]` button on this:
image

Redirects to https://github.com/pyca/cryptography/blob/3.3.1/src/cryptography/hazmat/primitives/asymmetric/dh.py#L18-L20


gpotter2:
I expected this to work just fine on readthedocs but it doesn't compile for some reason... Any suggestions would be appreciated

reaperhulk:
The failing build is due to a redefinition of `ocsp_responder_id_st`. We only define that struct if the OpenSSL headers claim to be 1.1.0 through 1.1.0i. This suggests a header mismatch problem. I think you can locally test by pulling the rtd docker images to experiment: https://github.com/readthedocs/readthedocs-docker-images

0xaead:
> The failing build is due to a redefinition of `ocsp_responder_id_st`. We only define that struct if the OpenSSL headers claim to be 1.1.0 through 1.1.0i. This suggests a header mismatch problem. I think you can locally test by pulling the rtd docker images to experiment: https://github.com/readthedocs/readthedocs-docker-images

It seems quite odd for `typedef` causes a redefinition conflict with a definition.
>  /usr/include/openssl/ossl_typ.h:208:16: note: originally defined here
   typedef struct ocsp_responder_id_st OCSP_RESPID;

Since this wasn't an issue before, I wildly guess the issue comes from `gcc` shipped by doc-builder docker. Will install it in my PC and take a look why.

0xaead:
> The failing build is due to a redefinition of `ocsp_responder_id_st`. We only define that struct if the OpenSSL headers claim to be 1.1.0 through 1.1.0i. This suggests a header mismatch problem. I think you can locally test by pulling the rtd docker images to experiment: https://github.com/readthedocs/readthedocs-docker-images

@reaperhulk 
I pulled `latest` {readthedocs} docker image on Ubuntu 18.04 but above failing can not be reproduced by exactly executing all the commands shown in https://readthedocs.org/projects/cryptography/builds/14446988/.
The last command ends with success:

> Collecting docutils<0.18,>=0.14
  Downloading docutils-0.16-py2.py3-none-any.whl (548 kB)
     |████████████████████████████████| 548 kB 5.0 MB/s
Building wheels for collected packages: cryptography
  Building wheel for cryptography (PEP 517) ... done
  Created wheel for cryptography: filename=cryptography-35.0.0.dev1-cp38-cp38-linux_x86_64.whl size=2651049 sha256=3fcc7abaa04c01c5feda70c164f356b6d9d883cec5be536d3287467d920cfe9c
  Stored in directory: /tmp/pip-ephem-wheel-cache-_4xu2z16/wheels/bb/ac/14/9c4a5862a13d5501b11042a0d7a0a4743272c169451b432418
Successfully built cryptography
Installing collected packages: sphinxcontrib-qthelp, sphinxcontrib-jsmath, sphinxcontrib-htmlhelp, sphinxcontrib-devhelp, sphinxcontrib-applehelp, pycparser, docutils, sphinx, cffi, sphinx-rtd-theme, cryptography
  Attempting uninstall: docutils
    Found existing installation: docutils 0.17.1
    Uninstalling docutils-0.17.1:
      Successfully uninstalled docutils-0.17.1
  Attempting uninstall: sphinx
    Found existing installation: Sphinx 1.8.5
    Uninstalling Sphinx-1.8.5:
      Successfully uninstalled Sphinx-1.8.5
  Attempting uninstall: sphinx-rtd-theme
    Found existing installation: sphinx-rtd-theme 0.4.3
    Uninstalling sphinx-rtd-theme-0.4.3:
      Successfully uninstalled sphinx-rtd-theme-0.4.3
Successfully installed cffi-1.14.6 cryptography-35.0.0.dev1 docutils-0.16 pycparser-2.20 sphinx-4.1.2 sphinx-rtd-theme-0.5.2 sphinxcontrib-applehelp-1.0.2 sphinxcontrib-devhelp-1.0.2 sphinxcontrib-htmlhelp-2.0.0 sphinxcontrib-jsmath-1.0.1 sphinxcontrib-qthelp-1.0.3

It seems test server is not using Ubuntu 18.04, at least some version newer than. I would presume the issue might be triggered by some newer version of toolchain e.g. GCC there?
GCC on 18.04 is:
> gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Another difference is that I am using Python3.8.0 shipped to 18.04 instead of 3.9.0. But I wonder how could python impacts on this build failure.

Any suggestion to Linux distro or something else to try next?

reaperhulk:
We use the 7.0 image, which https://github.com/readthedocs/readthedocs-docker-images claims is "testing" while 6.0 is "latest". So you may not be trying the correct image.

0xaead:
> We use the 7.0 image, which https://github.com/readthedocs/readthedocs-docker-images claims is "testing" while 6.0 is "latest". So you may not be trying the correct image.

I will try 7.0 to see. Pardon me if I am too nosy... May I ask the reason to use 7.0 image, instead of a stable 6.0 which works fine to generate docs?

gpotter2:
7.0 is the only one with rust 👍 Thanks a lot for all the help

0xaead:
> 7.0 is the only one with rust 👍 Thanks a lot for all the help

Ah, yeah, that makes sense. My pleasure.

0xaead:
> 7.0 is the only one with rust 👍 Thanks a lot for all the help
@gpotter2  and @reaperhulk 

OK, now I see why this build failed... The openssl headers linked are *too* old:

In 7.0 container, if we check /usr/include/openssl/opensslv.h:

\# define OPENSSL_VERSION_NUMBER  0x100020efL
\# ifdef OPENSSL_FIPS
\#  define OPENSSL_VERSION_TEXT    "OpenSSL 1.0.2n-fips  7 Dec 2017"
\# else
\#  define OPENSSL_VERSION_TEXT    "OpenSSL 1.0.2n  7 Dec 2017"
\# endif

That is why `ocsp_responder_id_st` was added:
>#if ( \
    !CRYPTOGRAPHY_IS_LIBRESSL && \
    CRYPTOGRAPHY_OPENSSL_LESS_THAN_110J \
    )
/* These structs come from ocsp_lcl.h and are needed to de-opaque the struct
   for the getters in OpenSSL 1.1.0 through 1.1.0i */
struct ocsp_responder_id_st {
...
};

Since we have claimed
> The OpenSSL C library. Cryptography supports OpenSSL version 1.1.0 and greater.

Seems pyca/cryptography doesn't consider to be linked against openssl earlier than 1.1.0, even done by mistake. Since the error message is quite misleading, and I saw a few similar cases reported by some parties tried to build cryptography wheels. I filled a new PR regarding to enforce this case.

And apparently, the 7.0 docker container is having some problem with openssl installation. I will fill another issue to them later on.
> $ openssl version
OpenSSL 1.1.1  11 Sep 2018

But:
> find /usr -name libssl*
/usr/lib/x86_64-linux-gnu/pkgconfig/libssl.pc
/usr/lib/x86_64-linux-gnu/libssl.a
/usr/lib/x86_64-linux-gnu/libssl.so
/usr/lib/x86_64-linux-gnu/libssl3.so
/usr/lib/x86_64-linux-gnu/libssl.so.1.0.0
/usr/lib/x86_64-linux-gnu/libssl.so.1.1
/usr/share/doc/libssl1.0-dev
/usr/share/doc/libssl1.0.0
/usr/share/doc/libssl1.1
/usr/share/lintian/overrides/libssl1.0.0

0xaead:
> And apparently, the 7.0 docker container is having some problem with openssl installation. I will fill another issue to them later on.
> ...

https://github.com/readthedocs/readthedocs-docker-images/issues/176#issue-991911490

reaperhulk:
Looks like this could possibly be rebased now that https://github.com/pyca/cryptography/pull/6331 has merged.

0xaead:
@gpotter2 Well, finally the `20.04` with proper openssl has been verified and merged. Could you rebase this change on latest HEAD?

gpotter2:
Hmm, it would appear that this broke support for https://docs.readthedocs.io/en/stable/config-file/v2.html or that the docs isn't up to date :/

reaperhulk:
Latest error will be resolved with: https://github.com/pyca/cryptography/pull/6331#issuecomment-930858037

The separate python section probably isn't needed either?

gpotter2:
> The separate python section probably isn't needed either?

It is actually needed to have readthedocs build the module (which is necessary to inspect it)

reaperhulk (approved):

PR #6215, opened 2021-08-20
Handle exception in x509.extensions.py to parse invalid certificates with ca:false path_length:0,1  (Closes #3856)

### LINKED ISSUE #3856: Some leaf certificates have path_length = 0 and CA = false
Most cases are path_length = None and CA = false, but I have seen some leaf (end entity) code signing certificates have path length = 0 even though the CA is False. Because the exception is raised, the extension information is no longer parsed.

https://github.com/pyca/cryptography/blob/0d6aaf49c1890378ce7ecf741a3a40c859d3b9fb/src/cryptography/x509/extensions.py#L333

ahmadmkhan1:
The exception handling only prompts regarding ca:false path_length not None, and allows parsing of invalid certificates

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -414,8 +414,11 @@ class BasicConstraints(ExtensionType):
         if not isinstance(ca, bool):
             raise TypeError("ca must be a boolean value")
 
-        if path_length is not None and not ca:
-            raise ValueError("path_length must be None when ca is False")
+        try:
+            if path_length is not None and not ca:
+                raise ValueError("path_length must be None when ca is False")
+        except Exception as E:
+            print(E)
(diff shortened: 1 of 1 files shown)


reaperhulk:
Unfortunately this approach entirely eliminates the check (which is a behavioral change) and also causes an exception to be printed in cases where it would have failed. Libraries like cryptography should never have `print` statements as part of their normal flow. We've discussed approaches to resolve this in the past (https://github.com/pyca/cryptography/pull/3862), but there's been no consensus on the proper way forward since correctness also matters.

The likely best approach is to split our x509 layer into a high level API (the existing one) and a lower level hazmat API that allows parsing of bad things, but that is a large amount of work that no one has attempted.

PR #6233, opened 2021-08-26
Explicit reexport of AsymmetricPadding

Dreamsorcerer:
Based on the comments and other code in the repo, this should be exported so that it can be imported by other code.

Even if you don't check against explicit exports in your codebase, it will causes errors for users who do have the option enabled:
https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport

### DIFF
--- a/src/cryptography/hazmat/primitives/asymmetric/padding.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/padding.py
@@ -7,7 +7,8 @@ import typing
 
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives._asymmetric import (
-    AsymmetricPadding as AsymmetricPadding)
+    AsymmetricPadding as AsymmetricPadding
+)
 from cryptography.hazmat.primitives.asymmetric import rsa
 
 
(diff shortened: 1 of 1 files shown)


reaperhulk:
If we're going to fix this I'd prefer to fix it systemically. Adding `no_implicit_reexport = true` to our mypy config shows quite a few errors we should correct. If you'd like to volunteer for this I'm happy to review an expanded PR, but otherwise you can file an issue and we'll treat it as a future mypy enhancement. Thanks for bringing this to our attention.

Dreamsorcerer:
This was really just a 5 second fix without cloning and running mypy.

If the concern is only where it affects users, then you probably want to enable `no_implicit_reexport` just on the examples/tests, which should catch the cases where things are exposed to the user without bothering you about imports that are internal to the library.

I don't see a mypy config file in the repo, but it'd basically just be a section like:
```
[mypy-tests.*]
no_implicit_reexport = True
```

Probably not going to have time to do this myself, but I can do a quick review if desired.

reaperhulk:
This is now done in #6240, thanks for pointing out this issue to us!

PR #6234, opened 2021-08-27
Add more demonstrative code to examples

0xaead:
1. In signature generation code example, add a preferred way to create 1key'. Also enables this code example to be run solely.
2. In verify( ) code example, add a positive case before the negative one. Also add usage of copy( ) which is mandatory to do self authentication.

Tested with `Python 3.8.1 (tags/v3.8.1:1b293b6, Dec 18 2019, 22:39:24) [MSC v.1916 32 bit (Intel)] on win32'

### DIFF
--- a/docs/hazmat/primitives/mac/hmac.rst
+++ b/docs/hazmat/primitives/mac/hmac.rst
@@ -28,10 +28,12 @@ of a message.
     .. doctest::
 
         >>> from cryptography.hazmat.primitives import hashes, hmac
+        >>> key = b'test key. Beware! A real key should use os.urandom or TRNG to generate'
         >>> h = hmac.HMAC(key, hashes.SHA256())
         >>> h.update(b"message to hash")
-        >>> h.finalize()
-        b'#F\xdaI\x8b"e\xc4\xf1\xbb\x9a\x8fc\xff\xf5\xdex.\xbc\xcd/+\x8a\x86\x1d\x84\'\xc3\xa6\x1d\xd8J'
+        >>> signature = h.finalize()
+        >>> signature
+        b'k\xd9\xb29\xefS\xf8\xcf\xec\xed\xbf\x95\xe6\x97X\x18\x9e%\x11DU1\x9fq}\x9a\x9c\xe0)y`='
 
     If the backend doesn't support the requested ``algorithm`` an
     :class:`~cryptography.exceptions.UnsupportedAlgorithm` exception will be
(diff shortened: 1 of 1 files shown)


alex:
Please take a look at the failing tests

0xaead:
> Please take a look at the failing tests

It turns out the failure comes from inspection to example output of `print(signature)`, which is unpredictable since introducing of a randomized key. Now change it to a comment afterwards.

docs/hazmat/primitives/mac/hmac.rst
```diff
@@ -28,10 +28,13 @@ of a message.
...
         >>> from cryptography.hazmat.primitives import hashes, hmac
+        >>> import os
+        >>> key = os.urandom(32)
         >>> h = hmac.HMAC(key, hashes.SHA256())
         >>> h.update(b"message to hash")
-        >>> h.finalize()
-        b'#F\xdaI\x8b"e\xc4\xf1\xbb\x9a\x8fc\xff\xf5\xdex.\xbc\xcd/+\x8a\x86\x1d\x84\'\xc3\xa6\x1d\xd8J'
+        >>> signature = h.finalize()
+        >>> # A signature is a 256-bit digest. E.g.
+        >>> # b'\xcb\x07\xe7.\x86&\xc3\t\xf8w\xb7\x8a1\xbe3\xa5JQH*5\x93\xf7H\xa0\x10\x80\xa8\x89\xf9vY'
```
reaperhulk, line 37: We want our doctests to actually test against values to confirm that they work and we haven't broken the API without realizing we need to update the docs. In this case you'll need to provide a fixed key for that, but you can add a comment explaining that users should generate and store their own key using `os.urandom` and load that in their own code.

0xaead: @reaperhulk I totally agree with you. And now a fixed key is added to code example. Also, I guess there is no need to add a comment about `os.urandom` since being addressed in previous instruction.

docs/hazmat/primitives/mac/hmac.rst
```diff
@@ -28,10 +28,12 @@ of a message.
     .. doctest::
 
         >>> from cryptography.hazmat.primitives import hashes, hmac
+        >>> key = b'\xb4/\xc4\xbb/\xe3\xe8\x15\xc9\xd0\x80H \xe6\x10O\xf8\xcdI\xc4\xf8GN\xbd\xe9\x84\xa8e\xed\x0c&\xd0'
```
reaperhulk, line 31: I don't love this because based on past experience I suspect people will copy/paste this and start using the key. This is part of why we originally didn't directly show where the key was coming from. If we want to change that I'd suggest making a key that is very clearly non-random `key = b'test key. a real key should use os.urandom'` or something like that.

0xaead: OK, this way is somehow better than adding a comment that might be neglected by a copy-paste "programmer"  ;)

0xaead:
The codecov failure on 98.93% is irrelevant to the change, no idea why it pops up here.

reaperhulk (approved):

PR #6236, opened 2021-08-29
preparations for musllinux

Liath:
Just added musllinux to the wheel builder and tweaked the execstack call to readelf which happens to be present in both pypa manylinux and musllinux images.

This should just work once the following are resolved:
- [x] https://github.com/pypa/manylinux/pull/1135
- [x] https://github.com/pypa/auditwheel/issues/327
- [x] https://github.com/pyca/infra/pull/367
- [x] https://github.com/pypa/warehouse/pull/9982

If you'd like to test that whole chain out, there's instructions in the last PR https://github.com/pyca/infra/pull/367

### DIFF
--- a/.github/workflows/wheel-builder.yml
+++ b/.github/workflows/wheel-builder.yml
@@ -22,6 +22,12 @@ jobs:
           - { NAME: "manylinux2014_x86_64", CONTAINER: "cryptography-manylinux2014:x86_64" }
           - { name: "manylinux_2_24_x86_64", CONTAINER: "cryptography-manylinux_2_24:x86_64"}
           - { name: "musllinux_1_1_x86_64", CONTAINER: "cryptography-musllinux_1_1:x86_64"}
+        exclude:
+          # There are no readily available musllinux PyPy distributions
+          - PYTHON: { VERSION: "pypy3.6", PATH: "/opt/pypy3.6/bin/pypy" }
+            MANYLINUX: { name: "musllinux_1_1_x86_64", CONTAINER: "cryptography-musllinux_1_1:x86_64" }
+          - PYTHON: { VERSION: "pypy3.7", PATH: "/opt/pypy3.7/bin/pypy" }
+            MANYLINUX: { name: "musllinux_1_1_x86_64", CONTAINER: "cryptography-musllinux_1_1:x86_64"}
     name: "${{ matrix.PYTHON.VERSION }} for ${{ matrix.MANYLINUX.NAME }}"
     steps:
       - run: ${{ matrix.PYTHON.PATH }} -m venv .venv
--- a/.github/workflows/wheel-builder.yml
+++ b/.github/workflows/wheel-builder.yml
@@ -21,6 +21,7 @@ jobs:
           - { NAME: "manylinux2010_x86_64", CONTAINER: "cryptography-manylinux2010:x86_64" }
           - { NAME: "manylinux2014_x86_64", CONTAINER: "cryptography-manylinux2014:x86_64" }
           - { name: "manylinux_2_24_x86_64", CONTAINER: "cryptography-manylinux_2_24:x86_64"}
+          - { name: "musllinux_1_1_x86_64", CONTAINER: "cryptography-musllinux_1_1:x86_64"}
     name: "${{ matrix.PYTHON.VERSION }} for ${{ matrix.MANYLINUX.NAME }}"
     steps:
       - run: ${{ matrix.PYTHON.PATH }} -m venv .venv
@@ -41,8 +42,8 @@ jobs:
       - run: auditwheel repair --plat ${{ matrix.MANYLINUX.NAME }} tmpwheelhouse/cryptograph*.whl -w wheelhouse/
       - run: unzip wheelhouse/*.whl -d execstack.check
       - run: |
-          results=$(execstack execstack.check/cryptography/hazmat/bindings/*.so)
-          count=$(echo "$results" | grep -c '^X' || true)
+          results=$(readelf -lW execstack.check/cryptography/hazmat/bindings/*.so)
+          count=$(echo "$results" | grep -c 'GNU_STACK.*[R ][W ]E' || true)
           if [ "$count" -ne 0 ]; then
             exit 1
           else


reaperhulk:
First of all, thank you very much for contributing! I’ll review the infra PR shortly, but just one question here: what version of `pip` adds support for the `musllinux` tag?

Liath:
No clue tbh. In a very confirmation-bias-y way I want to believe this comment is saying that pip won't actually need to do anything to support them: https://github.com/pypa/pip/pull/10088#issuecomment-864801361

Liath:
Oh, or maybe it's v21.2.4, because that's where packaging got upgraded to a version that handles them:
https://github.com/pypa/pip/commit/76cd70ac42cbbe6a51d83bd100f500ac4c21f26b

reaperhulk:
Great. The one other thing we'll want to do when landing this is updating the installation documentation to note that Alpine can use wheels as well (but requires pip >= 21.2.4). (Once this lands I will immediately run the wheel builder and upload the new musl wheels. We can put the doc work in a separate PR but I'd like to get that updated ASAP since it will be very helpful to users when v35.0 comes out and has a hard requirement on `rustc`)

Liath:
How about something like
```diff
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -78,9 +78,8 @@ OpenSSL. If you see this error it is likely because your copy of ``pip`` is too
 old to find our wheel files. Upgrade your ``pip`` with ``pip install -U pip``
 and then try to install ``cryptography`` again.

-Users on PyPy, unusual CPU architectures, or distributions of Linux using
-``musl`` (like Alpine) will need to compile ``cryptography`` themselves. Please
-view our :doc:`/installation` documentation.
+Users on PyPy and unusual CPU architectures will need to compile ``cryptography``
+themselves. Please view our :doc:`/installation` documentation.

 ``cryptography`` raised an ``InternalError`` and I'm not sure what to do?
 -------------------------------------------------------------------------
diff --git a/docs/installation.rst b/docs/installation.rst
index a9f40bf9..27fa6fec 100644
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -85,17 +85,16 @@ Building cryptography on Linux

 ``cryptography`` ships ``manylinux`` wheels (as of 2.0) so all dependencies
 are included. For users on **pip 19.0** or above running on a ``manylinux2010``
-(or greater) compatible distribution (almost everything **except Alpine**) all
+(or greater) compatible distribution (or **pip 21.2.4** for ``musllinux``) all
 you should need to do is:

 .. code-block:: console

     $ pip install cryptography

-If you are on Alpine or just want to compile it yourself then
-``cryptography`` requires a C compiler, a Rust compiler, headers for Python (if
-you're not using ``pypy``), and headers for the OpenSSL and ``libffi`` libraries
-available on your system.
+If you want to compile ``cryptography`` yourself you'll need a C compiler, a
+Rust compiler, headers for Python (if you're not using ``pypy``), and headers
+for the OpenSSL and ``libffi`` libraries available on your system.

 On all Linux distributions you will need to have :ref:`Rust installed and
 available<installation:Rust>`.
@@ -196,10 +195,10 @@ Static Wheels
 ~~~~~~~~~~~~~

 Cryptography ships statically-linked wheels for macOS, Windows, and Linux (via
-``manylinux``). This allows compatible environments to use the most recent
-OpenSSL, regardless of what is shipped by default on those platforms. Some
-Linux distributions (most notably Alpine) are not ``manylinux`` compatible so
-we cannot distribute wheels for them.
+``manylinux`` and ``musllinux``). This allows compatible environments to use
+the most recent OpenSSL, regardless of what is shipped by default on those
+platforms. Some Linux distributions are not ``manylinux`` or ``musllinux``
+compatible so we cannot distribute wheels for them.

 However, you can build your own statically-linked wheels that will work on your
 own systems. This will allow you to continue to use relatively old Linux
```

reaperhulk:
That looks like a good draft we can work with. It looks like we're out of date in some other ways with that documentation (notably: we ship PyPy wheels on x86_64 for Windows, macOS, and manylinux), but we can fix that separately.

Liath:
Regarding this bit
```
Some Linux distributions are not ``manylinux`` or ``musllinux`` compatible so we cannot distribute wheels for them.
```
Are there any other common distros that don't use one or the other? This might be clearer pushed into the following paragraph to get something like:
```diff
@@ -196,15 +195,15 @@ Static Wheels
 ~~~~~~~~~~~~~

 Cryptography ships statically-linked wheels for macOS, Windows, and Linux (via
-``manylinux``). This allows compatible environments to use the most recent
-OpenSSL, regardless of what is shipped by default on those platforms. Some
-Linux distributions (most notably Alpine) are not ``manylinux`` compatible so
-we cannot distribute wheels for them.
-
-However, you can build your own statically-linked wheels that will work on your
-own systems. This will allow you to continue to use relatively old Linux
-distributions (such as LTS releases), while making sure you have the most
-recent OpenSSL available to your Python programs.
+``manylinux`` and ``musllinux``). This allows compatible environments to use
+the most recent OpenSSL, regardless of what is shipped by default on those
+platforms.
+
+In the off chance you use a platform outside of that list, you can build your
+own statically-linked wheels that will work on your own systems. This will
+allow you to continue to use relatively old Linux distributions (such as LTS
+releases), while making sure you have the most recent OpenSSL available to
+your Python programs.

 To do so, you should find yourself a machine that is as similar as possible to
 your target environment (e.g. your production environment): for example, spin
```

reaperhulk:
I'm not aware of any significant distributions that remain uncovered with manylinux + musllinux so that language is definitely better. The primary point of failure for wheels at that point will be users on distributions older than the wheel compatibility (e.g. CentOS 6), architectures we don't produce wheels for (e.g. i686, s390x, ppc64le, etc), or alternate Python runtimes (e.g. Pyston).

I'd probably change `In the off chance you use a platform outside of that list` to `If you are using a platform not covered by our wheels, you can build your own...`, but it would be easier to provide feedback on language if we just put it into a PR 😄

messense:
FYI, looks like PyPI does not allow uploading musllinux wheels yet.

https://github.com/pypa/warehouse/blob/3943226cf1168f5cead40913d42603e9d1f25010/warehouse/forklift/legacy.py#L92-L173

Liath:
@messense Thanks for the heads up! I opened a PR for it here https://github.com/pypa/warehouse/pull/9982

reaperhulk:
For future reference #6280 covers the arm64 side of this as well (along with https://github.com/pyca/infra/pull/369).

reaperhulk (approved):

Liath:
With this in, would it make sense to add a comment to [this discussion](https://github.com/pyca/cryptography/issues/5771)? This and #6280 should, afaict, resolve the need to have Rust installed for most people.

reaperhulk:
Good idea. I just put an update under alpine in the final post.

Issue #6232, opened 2021-08-26
Migrate most of setup.py to setup.cfg


reaperhulk:
We apparently require a new enough setuptools, etc so we should move to the more modern solution. #6228 added the beginnings of a cfg.

PR #6239, opened 2021-08-29
Port most of setup.py to setup.cfg

reaperhulk:
Best modern practices and all that.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,87 @@
 [metadata]
+name = cryptography
+version = attr: cryptography.__version__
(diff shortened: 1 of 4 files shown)


alex (commented):
I think you can use `file:` for `long_description`

reaperhulk:
This fixes #6232

reaperhulk:
This is now ready for review. As best I can tell this does the same thing. I generated an `sdist` in both `main` and this branch and diff'd the resulting `PKG-INFO` file and they were identical.

alex (approved):

PR #6255, opened 2021-09-01
Make OAEP test vector generating works on python3

0xaead:
The script seems to be run with Python 2 originally some time ago and it is not compatible to Python 3 since its mixed usage of bytes and string. This change does a minimum change to make it work on python3.

TL;DR:
I was originally trying to make the script to work on both 2 and 3, so all changes are made in the way to do so.
But when testing with Python 2 I met backend importing error, then I found that "latest" pyca/cryptography already dropped the support to python2:
 >3.4 - 2021-02-07
 > BACKWARDS INCOMPATIBLE: Support for Python 2 has been removed.

 Anyways, if someone really wants to run _this_ script on Python 2 with an old version of pyca/cryptography, it should be doable.

I agree that RSA OAEP encrypt and decrypt are really rare use cases if compared to RSA sign/verify. But from a cryptographic worker's aspect, it is *as* important as others. So I listed some open topics to this script:
1. If consider _functionality_ of this script, it doesn't generate test vector for 1K, 3K and 4K key size. Ofc it can be easily added. 
2. Also, my test to use truncated SHA512 as `mgf1( )` and digest algorithm was rejected by openssl backend. Though they are legitimate at least by definition of https://datatracker.ietf.org/doc/html/rfc8017#page-19. But this is more like a case for openssl after checking its support to OAEP...

### DIFF
--- a/docs/development/custom-vectors/rsa-oaep-sha2/generate_rsa_oaep_sha2.py
+++ b/docs/development/custom-vectors/rsa-oaep-sha2/generate_rsa_oaep_sha2.py
@@ -27,29 +27,29 @@ def build_vectors(mgf1alg, hashalg, filename):
         skey = rsa.generate_private_key(65537, 2048, backend)
         pn = skey.private_numbers()
         examples = private["examples"]
-        output.append(b"# =============================================")
-        output.append(b"# Example")
-        output.append(b"# Public key")
-        output.append(b"# Modulus:")
+        output.append("# =============================================")
+        output.append("# Example")
+        output.append("# Public key")
+        output.append("# Modulus:")
         output.append(format(pn.public_numbers.n, "x"))
-        output.append(b"# Exponent:")
+        output.append("# Exponent:")
         output.append(format(pn.public_numbers.e, "x"))
-        output.append(b"# Private key")
-        output.append(b"# Modulus:")
+        output.append("# Private key")
+        output.append("# Modulus:")
         output.append(format(pn.public_numbers.n, "x"))
-        output.append(b"# Public exponent:")
+        output.append("# Public exponent:")
         output.append(format(pn.public_numbers.e, "x"))
-        output.append(b"# Exponent:")
+        output.append("# Exponent:")
         output.append(format(pn.d, "x"))
-        output.append(b"# Prime 1:")
+        output.append("# Prime 1:")
         output.append(format(pn.p, "x"))
-        output.append(b"# Prime 2:")
+        output.append("# Prime 2:")
         output.append(format(pn.q, "x"))
-        output.append(b"# Prime exponent 1:")
+        output.append("# Prime exponent 1:")
         output.append(format(pn.dmp1, "x"))
-        output.append(b"# Prime exponent 2:")
+        output.append("# Prime exponent 2:")
         output.append(format(pn.dmq1, "x"))
-        output.append(b"# Coefficient:")
+        output.append("# Coefficient:")
         output.append(format(pn.iqmp, "x"))
         pkey = skey.public_key()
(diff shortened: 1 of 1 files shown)


docs/development/custom-vectors/rsa-oaep-sha2/generate_rsa_oaep_sha2.py
```diff
@@ -94,11 +94,14 @@ def build_vectors(mgf1alg, hashalg, filename):
             output.append(b"# Encryption:")
             output.append(binascii.hexlify(ct))
 
+    for i, o in enumerate(output):
+        if type(o) != bytes:
+            output[i] = bytes(str(o).encode("utf-8"))
```
alex, line 99: This isn't the right way to do this. We need to ensure we're putting the same type of objects in the output list in the first place.

0xaead: OK, now simplify the logic to output `str` only.

0xaead:
```
would reformat docs/development/custom-vectors/rsa-oaep-sha2/generate_rsa_oaep_sha2.py
Oh no! 💥 💔 💥
1 file would be reformatted, 241 files would be left unchanged.
ERROR: InvocationError for command /home/runner/work/cryptography/cryptography/.tox/flake/bin/black --check . (exited with code 1)
___________________________________ summary ____________________________________
ERROR:   flake: commands failed
Error: Process completed with exit code 1.
```
flake installation unstable?

alex:
Running with this patch errors for me:

```
(.venv) ~/p/cryptography ❯❯❯ env PYTHONPATH=. python docs/development/custom-vectors/rsa-oaep-sha2/generate_rsa_oaep_sha2.p
y
  File "docs/development/custom-vectors/rsa-oaep-sha2/generate_rsa_oaep_sha2.py", line 123, in <module>
    "oaep-{0}-{1}.txt".format(hashtuple[0].name, hashtuple[1].name),
  File "docs/development/custom-vectors/rsa-oaep-sha2/generate_rsa_oaep_sha2.py", line 102, in write_file
    f.write(data)
TypeError: a bytes-like object is required, not 'str'
```

docs/development/custom-vectors/rsa-oaep-sha2/generate_rsa_oaep_sha2.py
alex: This is saying that `black` needs this file to be reformatted.

0xaead:
> Running with this patch errors for me:
> 
> ```
> (.venv) ~/p/cryptography ❯❯❯ env PYTHONPATH=. python docs/development/custom-vectors/rsa-oaep-sha2/generate_rsa_oaep_sha2.p
> y
> Traceback (most recent call last):
>   File "docs/development/custom-vectors/rsa-oaep-sha2/generate_rsa_oaep_sha2.py", line 123, in <module>
>     "oaep-{0}-{1}.txt".format(hashtuple[0].name, hashtuple[1].name),
>   File "docs/development/custom-vectors/rsa-oaep-sha2/generate_rsa_oaep_sha2.py", line 102, in write_file
>     f.write(data)
> TypeError: a bytes-like object is required, not 'str'
> ```

Sorry, it is my bad to allow a typo in. Fixed now.

alex:
This needs to be formatted with `black` that's why CI is failing.

0xaead:
> This needs to be formatted with `black` that's why CI is failing.

Yes, just installed and get it formatted.

alex (approved):

Issue #6267, opened 2021-09-09
_ModuleWithDeprecations should inherit from types.ModuleType


SlaushVunter:
Steps for reproducing bug:
```python
import inspect
import cryptography.utils
inspect.getsourcefile(cryptography.utils)
```
This will produce the following error:
> TypeError: module, class, method, function, traceback, frame, or code object was expected, got _ModuleWithDeprecations

The fix looks straightforward in cryptography.utils:
```python
class _ModuleWithDeprecations(types.ModuleType):
    def __init__(self, module):
        super().__init__(module.__name__)
        self.__dict__["_module"] = module
```
Used versions:
* ``Python`` 3.8.5
* ``cryptography`` 3.4.7
* ``cffi`` 1.14.0
* ``pip`` 20.2.2
* ``setuptools`` 49.2.1

* ``cryptography`` was installed by conda

PR #6268, opened 2021-09-10
_ModuleWithDeprecations should inherit from types.ModuleType (#6267)

SlaushVunter:
Update utils.py

I created a unit test for this simple change, but sadly I could not run the tests, because I could not install all dependencies (e.g. Rust as I am behind a corporate firewall). Someone please run the unit tests if they are not run automatically.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -4,6 +4,7 @@
 
 
 import binascii
+import inspect
 import os
 import textwrap
 
@@ -12,6 +13,7 @@ import pretend
 import pytest
 
 import cryptography
+import cryptography.utils
 from cryptography.exceptions import UnsupportedAlgorithm, _Reasons
 
 import cryptography_vectors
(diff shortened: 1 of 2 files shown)


alex:
`flake` test is failing due to a need to run `black`

tests/test_utils.py
```diff
@@ -4444,3 +4446,10 @@ def test_raises_unsupported_algorithm():
             "An error.", _Reasons.BACKEND_MISSING_INTERFACE
         )
     assert exc_info.type is UnsupportedAlgorithm
+
+
+def test_inspect_deprecated_module():
+    # Check if inspection is supported by _ModuleWithDeprecations.
```
alex, line 4452: Can you add an assertion that `isinstance` check that the module is an `_ModuleWithDeprecations`? Otherwise this test becomes pointless when we removed `int_from_bytes`.

reaperhulk (approved):

PR #6273, opened 2021-09-14
Allow pkcs12 to be importable

zmartin-impinj:
Add pkcs12.py's methods to its parent `__init__.py`.


reaperhulk:
Why do you want this? We kept pkcs12 in a separate namespace deliberately and documented it as such. With this it's possible to import without `.pkcs12` but then it's not clear what `load_key_and_certificates` really means.

alex:
Yes, it's intentional that these methods are accessed thrugh the `pkcs12` module.

zmartin-impinj:
Apologize. I was using the top level `serialization` and was expecting to just be able to `serialization.pkcs12..`.

PR #6276, opened 2021-09-18
Migrate CRL implementation to Rust

alex:


alex:
Need to write a test case for:

- [ ] `next_update` is null

src/rust/src/x509.rs
```diff
@@ -539,6 +540,509 @@ fn load_der_x509_certificate(py: pyo3::Python<'_>, data: &[u8]) -> PyAsn1Result<
...
+                },
+                pem::EncodeConfig {
+                    line_ending: pem::LineEnding::LF,
+                },
+            )
+            .into_bytes();
+            Ok(pyo3::types::PyBytes::new(py, &pem))
+        } else {
+            Err(pyo3::exceptions::PyTypeError::new_err(
+                "encoding must be an item from the Encoding enum",
```
reaperhulk, line 762: The encoding enum has tons of stuff, we should probably just say it must be DER or PEM from encoding.

reaperhulk: (This should be a followup PR)

reaperhulk (approved):

PR #6282, opened 2021-09-19
Don't internal error on CRL with no nextUpdate value

alex:


reaperhulk:
Does this have a valid or invalid sig? If invalid please note that in the docs

alex:
done

reaperhulk (approved):

reaperhulk:
We should add a changelog entry stating that next update is now correctly parsed as optional.

alex:
Do we normally note bug fixes in teh changelog?

reaperhulk:
We're not consistent, so we could go either way. I'm good with it being in a separate PR though if we decide to do it.

PR #6292, opened 2021-09-21
Expose X509_verify for X.509 certificate signature verification

torin-carey:
Currently, certificate signature verification involves verifying the signature on the cert tbs data. This is already more involved than it should be as users will have to handle the possible values of `Certificate.signature_algorithm_oid` for the different algorithm and hash pairings.
The `Certificate` class only exposes the `ObjectIdentifier` of the signature's ASN.1 `AlgorithmIdentifier` field, which isn't so bad if we're dealing with RSA PKCS#1 v1.5 signatures, but is an issue if we're dealing with RSA PSS signatures, where the `AlgorithmIdentifier` may store other parameters. Handling these signatures isn't currently easy.

To solve this, we can ~~shamelessly copy~~ adapt the existing code used for verifying CRL signatures. We simply serialise the certificate and let OpenSSL check.

### DIFF
--- a/src/_cffi_src/openssl/x509.py
+++ b/src/_cffi_src/openssl/x509.py
@@ -59,6 +59,7 @@ int X509_set_pubkey(X509 *, EVP_PKEY *);
 
 unsigned char *X509_alias_get0(X509 *, int *);
 int X509_sign(X509 *, EVP_PKEY *, const EVP_MD *);
+int X509_verify(X509 *, EVP_PKEY *);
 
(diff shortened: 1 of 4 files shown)


reaperhulk:
While we do expose signature checking on CSR and CRL, we've been reluctant to do so in this fashion for certificates because we don't want people to be confused about the difference between a valid signature and a trustable chain from that valid signature. #5093, #5116, #5367, and #5950 are all examples of attempts to tackle this, but no one has come up with a comprehensive implementation for doing this safely and rigorously as yet.

PR #6297, opened 2021-09-22
Convert OCSPResponse to be Rust

alex:


alex:
Missing test cases:

- [ ] `response_type` that's not `BASIC_RESPONSE_OID`
- [ ] `response_bytes` that aren't present even though the response is SUCCESSFUL
- [ ] `response_status` that's an unknown value

reaperhulk (approved):

PR #6312, opened 2021-09-27
Parse CSRs in pure rust

reaperhulk:


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1401,31 +1409,24 @@ def _crl_is_signature_valid(
...
-            self._consume_errors()
-            raise ValueError(
-                "Unable to load request. See https://cryptography.io/en/"
-                "latest/faq.html#why-can-t-i-import-my-pem-file for more"
-                " details."
-            )
-
+        x509_req = self._lib.d2i_X509_REQ_bio(mem_bio.bio, self._ffi.NULL)
+        self.openssl_assert(x509_req != self._ffi.NULL)
         x509_req = self._ffi.gc(x509_req, self._lib.X509_REQ_free)
```
alex, line 1419: Why not use `_csr2ossl` here?

src/rust/src/x509.rs
```diff
@@ -544,6 +547,441 @@ fn load_der_x509_certificate(py: pyo3::Python<'_>, data: &[u8]) -> PyAsn1Result<
...
+}
+
+#[pyo3::prelude::pyclass]
+struct CertificateSigningRequest {
+    raw: OwnedRawCsr,
+    cached_extensions: Option<pyo3::PyObject>,
+}
+
+#[pyo3::prelude::pyproto]
+impl pyo3::class::basic::PyObjectProtocol for CertificateSigningRequest {
```
alex, line 615: There's a whole bunch of names that can be shortened: none of the `::class::` are required nor are the `::pycell::`

src/rust/src/x509.rs
```diff
@@ -544,6 +547,441 @@ fn load_der_x509_certificate(py: pyo3::Python<'_>, data: &[u8]) -> PyAsn1Result<
...
+            .downcast::<pyo3::types::PyString>()?
+            .to_string();
+        let rust_oid = asn1::ObjectIdentifier::from_string(&oid_str).unwrap();
+        // TODO: no more unwrap once attributes isn't an Optional
+        for attribute in self.raw.borrow_value().csr_info.attributes.clone().unwrap() {
+            if rust_oid == attribute.type_id {
+                check_attribute_length(attribute.values.clone())?;
+                if let Some(val) = attribute.values.clone().next() {
+                    // We allow utf8string, printablestring, and ia5string at this time
+                    if val.tag() == 0x0c || val.tag() == 0x13 || val.tag() == 0x16 {
```
alex, line 755: Is it possible to use `asn1::UTF8String::TAG` (etc.) here?

src/rust/src/x509.rs
```diff
@@ -544,6 +547,441 @@ fn load_der_x509_certificate(py: pyo3::Python<'_>, data: &[u8]) -> PyAsn1Result<
...
+                ),
+            )?,
+        ))
+    }
+
+    #[getter]
+    fn extensions(&mut self, py: pyo3::Python<'_>) -> pyo3::PyResult<pyo3::PyObject> {
+        let x509_module = py.import("cryptography.x509")?;
+        let exts = self.raw.borrow_value().csr_info.get_extension_attribute()?;
+        parse_and_cache_extensions(py, &mut self.cached_extensions, &exts, |oid, ext_data| {
```
alex, line 782: This duplicates a ton of cert parsing, can we put it in a function that we reuse, as we do for OCSP/CRLEntry

alex: (PS: Put that refactoring in it's own PR for ease of review plz)

reaperhulk: This is an overlapping set but there are two more extensions in cert via csr so we'll have to handle that small variance

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -967,7 +956,10 @@ def create_x509_csr(
             errors = self._consume_errors_with_text()
             raise ValueError("Signing failed", errors)
 
-        return _CertificateSigningRequest(self, x509_req)
+        bio = self._create_mem_bio_gc()
+        res = self._lib.i2d_X509_REQ_bio(bio, x509_req)
+        self.openssl_assert(res == 1)
+        return rust_x509.load_der_x509_csr(self._read_mem_bio(bio))
```
alex, line 962:
```suggestion
        return self._ossl2csr(x509_req)
```

src/rust/src/x509.rs
```diff
@@ -391,6 +411,296 @@ fn load_der_x509_certificate(py: pyo3::Python<'_>, data: &[u8]) -> PyAsn1Result<
...
+        // TODO: Convert ObjectIdentifier in python to a rust class so this can be more efficient
+        let oid_str = oid
+            .getattr("dotted_string")?
+            .downcast::<pyo3::types::PyString>()?
+            .to_string();
+        let rust_oid = asn1::ObjectIdentifier::from_string(&oid_str).unwrap();
+        for attribute in self.raw.borrow_value().csr_info.attributes.clone() {
+            if rust_oid == attribute.type_id {
+                check_attribute_length(attribute.values.clone())?;
+                if let Some(val) = attribute.values.clone().next() {
```
alex, line 609: This can just be `val = attributes.values.clone().next().unwrap()` -- there's always an element in the iterator due to teh `check_attribute_length`

src/rust/src/x509.rs
```diff
@@ -391,6 +411,296 @@ fn load_der_x509_certificate(py: pyo3::Python<'_>, data: &[u8]) -> PyAsn1Result<
...
+    fn get_attribute_for_oid<'p>(
+        &self,
+        py: pyo3::Python<'p>,
+        oid: &pyo3::PyAny,
+    ) -> pyo3::PyResult<&'p pyo3::PyAny> {
+        // TODO: Convert ObjectIdentifier in python to a rust class so this can be more efficient
+        let oid_str = oid
+            .getattr("dotted_string")?
+            .downcast::<pyo3::types::PyString>()?
+            .to_string();
```
alex, line 604: Does `.extract::<&str>` not work?

src/rust/src/x509.rs
```diff
@@ -391,6 +411,296 @@ fn load_der_x509_certificate(py: pyo3::Python<'_>, data: &[u8]) -> PyAsn1Result<
...
+                        )));
+                    }
+                }
+            }
+        }
+        Err(pyo3::PyErr::from_instance(
+            py.import("cryptography.x509")?.call_method1(
+                "AttributeNotFound",
+                (
+                    format!("No {} attribute was found", rust_oid.to_string()),
```
alex, line 630: If the extract thing works, you can ust use that value instead of converting rust_oid back to a string.

alex (dismissed):

alex:
There's some coverage issue in the rust code, I don't think it's codecov but it might be a weird rust coverage issue.

alex (approved):

Issue #5124, opened 2020-02-22
openssl.Backend.load_der_ocsp_response() failing to load complex response


HQJaTu:
* Versions of Python, ``cryptography``, ``cffi``, ``pip``, and ``setuptools``
  you're using
```
Name: cryptography
Version: 2.8
---
Name: cffi
Version: 1.11.5
---
Name: pip
Version: 18.1
---
Name: setuptools
Version: 40.8.0
```

* How you installed ``cryptography``
`pip install cryptography --upgrade`

* Clear steps for reproducing your bug
Code:
```python
from cryptography.hazmat.backends.openssl.backend import backend as x509_openssl_backend
ocsp_response_der = open('ocsp-army.deps.mil-resp.der', 'rb').read()
x509_openssl_backend.load_der_ocsp_response(ocsp_response_der )
```

Error:
```
    ocsp_resp = x509_openssl_backend.load_der_ocsp_response(r.content)
  File "/usr/local/lib64/python3.7/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 1531, in load_der_ocsp_response
    return _OCSPResponse(self, response)
  File "/usr/local/lib64/python3.7/site-packages/cryptography/hazmat/backends/openssl/ocsp.py", line 106, in __init__
    self._backend._lib.OCSP_resp_count(self._basic) == 1
  File "/usr/local/lib64/python3.7/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 131, in openssl_assert
    return binding._openssl_assert(self._lib, ok)
  File "/usr/local/lib64/python3.7/site-packages/cryptography/hazmat/bindings/openssl/binding.py", line 78, in _openssl_assert
    errors_with_text
cryptography.exceptions.InternalError: Unknown OpenSSL error. This error is commonly encountered when another library is not cleaning up the OpenSSL error stack.
```

The specific OCSP-response file is attached as a Zip-file.
[ocsp-army.deps.mil-resp.zip](https://github.com/pyca/cryptography/files/4239418/ocsp-army.deps.mil-resp.zip)

* Additional information:
Running:
```
$ openssl ocsp -respin ocsp-army.deps.mil-resp.der -text
```
will produce correct and expected output without errors.

This error is highly data-dependant and very reproducible. For some other data, the Python code works. For this particular data everything explodes. My thinking is, this data is worst-case-scenario. The OCSP-response has all kinds of extensions in it.

reaperhulk:
[...]

PR #6320, opened 2021-09-28
Support for multiple SINGLERESP messages in one OCSP response

turettn:
#5124 identified an issue where OCSP responses with multiple certificate blocks were rejected violently.  #5316 cleaned this up with an intelligent error message.

This commit allows the somewhat common case described in the original issue to be properly handled. `load_der_ocsp_response` now takes an extra parameter - the OCSP request the user is interested in.  If this is specified, the certificates are iterated through until one is found where the hash algorithm, issuer key and name hashes, and serial number all match.  If found, it is returned.

If the request parameter is not specified, it throws `cryptography.exceptions.MultiOCSPResponseWIthoutReq`.  This descends from ValueError, so old code should continue to work without modification.

If the request parameter is specified, but the a corresponding certificate is not found, it throws `cryptography.exceptions.MultiOCSPResponseNoMatchReq`.  There is an annoying case where the user submits request using hash algorithm X, but the server returns the response using hash algorithm Y.  This prevents positive matching between request and response.  The exception contains the hash algorithms used in the response in the event an ambitious user wants to retry the request.

Tests are included, as well as test vectors in the form of requests corresponding to certificates in the existing `ocsp-army.deps.mil-resp.der` vector.

This is the farthest I've gone down the pyca/cryptography or OpenSSL rabbit-hole, and am happy to revise this as necessary.

Thanks for taking the time to review this.


reaperhulk:
I haven't looked at what this attempts to do yet so I can't comment on whether the approach is something we want to pursue, but unfortunately you picked a component of the source code that we replaced entirely with rust-based parsing just three days ago (https://github.com/pyca/cryptography/pull/6297). If you're interested in learning some Rust this could definitely be implemented on top of the new code as well, although it might be worth having a discussion about the motivation and usage you envision here. 

My concern at the moment is that we consider an OCSPResponse to contain a complete structure with a valid signature, but with this model you'd be able to have "views" into that structure where, while it still holds all the bytes, it only looks at parts of them depending upon the way in which you constructed it. Is that really the ideal way to do this or would a new API make more sense? For example, `response.single_responses` where it's an iterable with an object containing all the getter methods. We can leave the top level getters, which will be documented as aliases for `response.single_responses[0].some_getter` and then we have both the typical single response case and also support for multiple.

turettn:
@reaperhulk Thanks for the extremely prompt feedback!

Motivation-wise, there's at least one commonly used OCSP server that pre-generates OCSP responses in batches of 20, then when it receives a OCSP request it simply finds the correct response file and fires it back.  These servers seem to be mostly used in environments with extremely large client certificate roll-outs (e.g. the [NIPRNet](https://en.wikipedia.org/wiki/NIPRNet), where every user has a [physical token](https://en.wikipedia.org/wiki/Common_Access_Card) with a client certificate).  On these networks, servers are constantly checking client certificate validity due to the high rate of revocations (e.g. cards are lost\stolen, eligibility for access changes, etc...).  The original issue (#5124) referenced above provided the `ocsp-army.deps.mil-resp.der` test vector, which appears to have come from one such network.

In this situation, where the client asks to validate serial number 12345 and the server returns 12340-12359, the user likely doesn't care about the extra responses - they're completely irrelevant to the request, but are sent as a performance optimization on the server side.

I can understand not wanting to build "views" into the OCSPResponse.  In this case, I would be tempted to implement the iterable as you suggested, then also provide a function that could be called to match based on a OCSP Request object.

I am by no means a Rust expert, but I've written some before, and am happy to take a stab at it.  Let me know what you think.

reaperhulk:
I'm happy to review a PR since there exists a concrete use case for this. I would definitely prefer the iterable model over views, but @alex may have his own opinions.

alex:
Yes, I think an approach where you load an OCSPResponse and then there's a way to access individual members makes more sense (and probably most methods on OCSPResponse give an error if there's multiple responses?)

turettn:
Sounds good - thanks for the feedback.  I'll close out this PR, and will try to get a new one submitted sometime next week.

PR #6325, opened 2021-09-28
add CRL pyopenssl fallback

reaperhulk:


.github/downstream.d/pyopenssl-release.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -ex
+
+case "${1}" in
+    install)
+        pip install "pyopenssl[test]"
+        git clone https://github.com/pyca/pyopenssl
+        cd pyopenssl
+        VERSION=$(python -c "import OpenSSL;print(OpenSSL.__version__)")
+        git checkout "$VERSION"
```
alex, line 9: This seems like a not-amazing way to do this, and I don't even see how it works? The current `__version__` is `21.0.0.dev` which isn't a tag?

Can we not use `pip download` or something?

alex, line 9:
```suggestion
        VERSION=$(curl https://pypi.org/pypi/pyOpenSSL/json | jq -r .info.version)
        git clone https://github.com/pyca/pyopenssl
        cd pyopenssl
        git checkout "$VERSION"
```

alex (approved):

PR #6328, opened 2021-09-29
build our wheels from tag so we can upload sdist last

reaperhulk:
This tries to preserve the capability of building arbitrary versions but now also building automatically when a tag is pushed.


reaperhulk:
Example building via the workflow dispatch: https://github.com/pyca/cryptography/actions/runs/1285264228

alex (approved):
Approving for now, but `release.py` needs to be updated before we can actually release.

PR #6331, opened 2021-09-29
docs: upgrade image to latest 20.04

0xaead:
openssl version is messed in `image 7', which causes error e.g.
https://readthedocs.org/projects/cryptography/builds/14446988/

Signed-off-by: Match Man <ppucoder@gmail.com>

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -6,6 +6,8 @@ sphinx:
   builder: dirhtml
 
 build:
-  # First RTD build env which includes a rust toolchain
-  image: "7.0"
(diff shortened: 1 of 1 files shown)


alex:
Just to confirm, 20.04 has rust installed?

alex (approved):

humitos:
> Just to confirm, 20.04 has rust installed?

Note that Ubuntu 20.04 LTS does not have installed Rust by default. You have to declare it by using:

```yaml
# .readthedocs.yaml
build:
  tools:
    rust: "1.55"
```

Let me know if that works.

PR #6332, opened 2021-09-30
update release.py

reaperhulk:


release.py
```diff
@@ -79,27 +79,13 @@ def download_artifacts_github_actions(session, token, run_url):
...
-    )
-    response.raise_for_status()
-
-    # Give it a few seconds for the run to kick off.
-    time.sleep(5)
     response = session.get(
         (
             "https://api.github.com/repos/pyca/cryptography/actions/workflows/"
-            "wheel-builder.yml/runs?event=workflow_dispatch"
+            "wheel-builder.yml/runs?event=create"
```
reaperhulk, line 88: I think this is the right event type but https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows and https://docs.github.com/en/rest/reference/actions#workflow-runs are the docs to check.

alex (approved):

alex (approved):

PR #6348, opened 2021-10-01
PKCS12: return 'friendly name' with new load_key_and_certificates_with_name API

### LINKED ISSUE #5760: Being able to read PKCS12 'friendly name'
Right now, the interface for loading PKCS12 files (https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization.html?highlight=pkcs12#cryptography.hazmat.primitives.serialization.pkcs12.load_key_and_certificates) returns a tuple `(private_key, certificate, additional_certificates)`. This tuple does not contain the 'friendly name', and assuming you don't want to have a different tuple length depending on some function argument, this cannot be properly extended.

Goal: make it possible to load the PKCS12 friendly name.

Idea: how about creating a small immutable object for PKCS12 content (i.e. it contains private key, certificate, list of additional certificates, friendly name - and maybe more in the future?) that can be serialized and deserialized by a new set of functions?

Does this sounds good? Does anyone have a better idea? I'm hesistant of suggestion a completely new interface, now that we "just" got PKCS12 support a few versions ago. But I can't think of a much better way to be able to return more data (friendly name now) while keeping backwards compatibility.

felixfontein:
This PR provides a new API asked for in #5760 to return the 'friendly name' from a PKCS12 file, which happens to be the name of the "main" certificate with private key that was provided to the `serialize_key_and_certificates` API when creating a PKCS12. This API brings cryptography's API more to par with the API provided by PyOpenSSL. (Now the only remaining differences I'm aware of are related to the encryption algorithm / configuration.)

As suggested by @reaperhulk in https://github.com/pyca/cryptography/issues/5760#issuecomment-813112084, adds a new API which also returns the 'friendly name'. I've adjusted the name slightly to `load_key_and_certificates_with_name` since the storing API (`serialize_key_and_certificates`) also uses `name`. I've put `name` as the first return value of the new API since that's the order of parameters `serialize_key_and_certificates` uses. (If these two things should change, I'm happy to do that :) )

The old API `load_key_and_certificates` now basically calls the new API and ignores the name.

Fixes #5760.


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2559,6 +2571,9 @@ def load_key_and_certificates_from_pkcs12(self, data, password):
         if x509_ptr[0] != self._ffi.NULL:
             x509 = self._ffi.gc(x509_ptr[0], self._lib.X509_free)
             cert = self._ossl2cert(x509)
+            maybe_name = self._lib.X509_alias_get0(x509_ptr[0], self._ffi.NULL)
```
reaperhulk, line 2574: I asked this previously, but can there be multiple friendly names in a PKCS12 (e.g. one per additional certificate)? If yes then this gets a bit trickier because we should have an API that can return them all. You could imagine that being `(key, (name, cert), [(name, cert), ...])` I suppose, although it's worth asking ourselves if we've reached the threshold of needing a real PKCS12 object at that point.

reaperhulk (commented):
This needs docs, changelog, and tests that exercise the new codepath (codecov not reporting is a common problem, but if it did report this would fail coverage).

src/cryptography/hazmat/backends/openssl/backend.py
felixfontein: I think there can be, but right now it's not possible to create such a PKCS12 file with cryptography or PyOpenSSL. The OpenSSL API used by both is a simplified API (`PKCS12_create`) which accepts one single friendly name (which is applied to the 'main' certificate). I guess we can return names we find for other certificates as well, but I'm not sure how useful this would be if it's not possible to create such a certificate without using third-party code.

(I'll start with tests, changelog and docs once it's clear how the interface should be.)

felixfontein: I implemented what you suggested in 4dfb7a68f866ddd9579e12b5347b9f7cbbfafd15 together with an API to create such PKCS#12 files (`serialize_key_and_certificates_with_names()`), and added some basic tests.

felixfontein: (I'm still hesitant at adding a `PKCS12` object at this step, since that kind of fixes the structure of PKCS#12 files that cryptography can handle for a longer time.)

reaperhulk:
Two big comments:
1) Let's not make a new serialization API for now. Let's just add a test vector that has multiple aliases and be able to parse that.
2) I think we need to return a PKCS12 object now, this multiply-nested tuple API is too much.

A true PKCS12 object is ridiculous (see: https://www.bouncycastle.org/docs/pkixdocs1.5on/org/bouncycastle/pkcs/PKCS12PfxPduBuilder.html) so maybe what we want is more along the lines of

```python
class PKCS12KeyAndCertificates:
    @getter
    def key -> PRIVATE_KEY_TYPES: ...

    @getter
    def cert -> PKCS12Certificate: ....

    @getter
    def additional_certs -> List[PKCS12Certificate]: ...
```

```
class PKCS12Certificate:
    @getter
    def friendly_name -> bytes: ...

    @getter
    def certificate -> x509.Certificate: ...
```

felixfontein:
@reaperhulk I guess `key` and `cert` in `PKCS12KeyAndCertificates` should be `typing.Optional[]`?

reaperhulk:
Yeah they're not required so they should be optionals. That's what I get from writing an interface in a GitHub comment 😄

reaperhulk (changes requested):
All the new test vectors should be documented in `test-vectors.rst` describing what they contain. We have a section named `Custom PKCS12 Test Vectors` you can add them to.

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -499,6 +499,24 @@ file suffix.
         ``additional_certificates`` is a list of all other
         :class:`~cryptography.x509.Certificate` instances in the PKCS12 object.
 
+.. function:: load_key_and_certificates_object(data, password, backend=None)
```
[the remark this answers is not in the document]
felixfontein: I agree; I don't have a good idea though. Maybe `load_key_and_certificates_pkcs12_object`, `load_pkcs12_key_and_certificates`? If you have better suggestions I'm happy to implement them :)

docs/x509/reference.rst
```diff
@@ -3262,6 +3262,47 @@ instances. The following common OIDs are available as constants.
...
+
+        A :class:`~cryptography.x509.Certificate` instance.
+
+    .. attribute:: friendly_name
+
+        :type: bytes
+
+        An optional byte string containing the friendly name of the certificate.
+
+.. class:: PKCS12KeyAndCertificates
```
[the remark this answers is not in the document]
felixfontein: Implemented 27ca1d9ead0e74d40dd17409eb697f4205163e74.

docs/hazmat/primitives/asymmetric/serialization.rst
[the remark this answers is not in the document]
reaperhulk: cc @alex who may have an idea here. We could just do `load_binary` (a PKCS12 is not strict DER so we can't say it's DER) I suppose, although that makes the naming for a future more-complex API more difficult.

docs/development/test-vectors.rst
```diff
@@ -625,6 +625,108 @@ Custom PKCS12 Test Vectors
 * ``pkcs12/cert-aes256cbc-no-key.p12`` - A PKCS12 file containing a cert
   (``x509/custom/ca/ca.pem``) encrypted via AES 256 CBC with the
   password ``cryptography`` and no private key.
+* ``pkcs12/no-name-no-pwd.p12`` - A PKCS12 file containing a cert
```
reaperhulk, line 628: I can review these vectors and land them as a separate PR if you are willing to split them out and then rebase this PR against main once it merges.

felixfontein: @reaperhulk that would be great, thank you!

docs/hazmat/primitives/asymmetric/serialization.rst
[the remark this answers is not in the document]
reaperhulk: Okay, talked it over with Alex and while we came up with two examples, one is too horrifying (we modify `load_key_and_certificates` to return this class but add tuple-unpacking to it to remain API compatible), so let's just name this `load_pkcs12` and we'll cross the bridge of figuring out how to expand to a more complete API if and when we need to.

reaperhulk:
Ready for rebase

felixfontein:
I don't think the cancelled test is related to this PR.

src/cryptography/hazmat/primitives/serialization/pkcs12.py
```diff
@@ -18,6 +18,114 @@
...
+class PKCS12Certificate:
+    def __init__(
+        self,
+        cert: x509.Certificate,
+        friendly_name: typing.Optional[bytes],
+    ):
+        if not isinstance(cert, x509.Certificate):
+            raise TypeError("Expecting x509.Certificate object")
+        if friendly_name is not None and not isinstance(friendly_name, bytes):
+            raise TypeError("friendly_name must be bytes")
```
reaperhulk, line 30: bytes or None

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -543,6 +564,45 @@ file suffix.
...
+
+    Represents additional data provided for a certificate in a PKCS12 file.
+
+    .. attribute:: certificate
+
+        A :class:`~cryptography.x509.Certificate` instance.
+
+    .. attribute:: friendly_name
+
+        :type: bytes
```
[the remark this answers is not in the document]
felixfontein: Implemented in 41b2d6f63e860c5598c73543e50148317cf60df1.

src/cryptography/hazmat/primitives/serialization/pkcs12.py
felixfontein: Implemented in 41b2d6f63e860c5598c73543e50148317cf60df1.

src/cryptography/hazmat/primitives/serialization/pkcs12.py
```diff
@@ -18,6 +18,114 @@
...
+        if key is not None and not isinstance(
+            key,
+            (
+                rsa.RSAPrivateKey,
+                dsa.DSAPrivateKey,
+                ec.EllipticCurvePrivateKey,
+            ),
+        ):
+            raise TypeError(
+                "Key must be RSA, DSA, or EllipticCurve private key."
```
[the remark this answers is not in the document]
felixfontein: Implemented in a5b2648cdf18e04cf89b4d777561437f4d182d42.

tests/hazmat/primitives/test_pkcs12.py
```diff
@@ -335,3 +440,210 @@ def make_cert(name):
...
+class TestPKCS12Objects(object):
+    def test_certificate_constructor(self, backend):
+        with pytest.raises(TypeError):
+            PKCS12Certificate(None, None)
+        with pytest.raises(TypeError):
+            PKCS12Certificate("hello", None)
+        with pytest.raises(TypeError):
+            PKCS12Certificate(None, "hello")
+        with pytest.raises(TypeError):
+            PKCS12Certificate(None, 42)
```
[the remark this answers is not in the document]
felixfontein: I think that was what I intended to do anyway... I've fixed it in 1546d844c0be0ceadc3c8361678111daad528fb3.

reaperhulk (approved):

PR #6365, opened 2021-10-03
Accept combined PEM files with multiple sections

mhils:
This PR adds a `find_in_pem` on the Rust side, which filters out unrelated PEM sections and returns the matching one. This makes it possible to read combined key-and-cert PEM files again (#6351).

If multiple sections are found that match our criteria, we error. I'm not sure if this matches the pre-35.0 behavior, but I feel it's not a particulary good idea to let that go through unnoticed. :)


alex (commented):
Behavior looks good.

Unfortunately you're about to discover that coverage is unhappy, since we don't excersise the error handling in all the `load_pem_*` functions.

I'd suggest having `find_in_pem` accept the error strings for `NoMatchingSectino` and `MultipleMatchingSections` as `&str` and then having it generate hte python exceptions, this way we don't nee to repeat the branching.

mhils:
Great suggestion, that actually makes things quite simpler. 👍

alex (approved):

niander:
When will this be released? This change on behavior breaks [azure-identity](https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/identity/azure-identity). It has been since expected that PEM can have private keys and certificate sections in any order. It seems that since v35.0.0, only the first header is taken into account.

Issue #6368, opened 2021-10-04
Cryptography 35.0.0 fails to parse FreeIPA server CSR


tiran:
Cryptograhy 35.0.0 fails to parse a CSR generated by FreIPA 4.9 / certmonger. The certificate contains three additional fields that are uncommon:

- SAN OtherName szOID_NT_PRINCIPAL_NAME (1.3.6.1.4.1.311.20.2.3)
- SAN OtherName Kerberos V5 pkinit (1.3.6.1.5.2.2)
- Microsoft szOID_ENROLL_CERTTYPE (1.3.6.1.4.1.311.20.2) extension

While writing the bug report, Alex pointed out that the issue is caused by the critical field. The critical field is encoded although the value is equal to the default value. In DER fields should not be encoded if the value matches the default value.

```
$ openssl asn1parse -inform PEM -in freeipa.csr
...
  421:d=7  hl=2 l=   3 prim: OBJECT            :X509v3 Subject Alternative Name
  426:d=7  hl=2 l=   1 prim: BOOLEAN           :0
  429:d=7  hl=3 l= 135 prim: OCTET STRING      [HEX DUMP]:...
...
```

```
   Extension  ::=  SEQUENCE  {
        extnID      OBJECT IDENTIFIER,
        critical    BOOLEAN DEFAULT FALSE,
        extnValue   OCTET STRING
                    -- contains the DER encoding of an ASN.1 value
                    -- corresponding to the extension type identified
                    -- by extnID
        }
```

```
Certificate Request:
    Data:
        Version: 1 (0x0)
        Subject: O = IPA.TEST, CN = replica1.ipa.test
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                RSA Public-Key: (2048 bit)
                Modulus:
                    00:e5:3c:d0:66:92:a0:5a:ee:2f:e0:b9:97:e3:8c:
                    51:0a:00:36:0d:cb:d8:e0:3e:cb:c3:f6:72:af:7b:
                    9c:87:9e:20:bc:05:16:8b:6a:22:fc:f6:15:8b:12:
                    50:35:51:df:64:87:24:82:b3:d1:3b:e8:2d:31:a1:
                    6c:5e:e2:07:81:61:e3:cb:64:dd:14:d3:95:e5:46:
                    af:dd:6d:3b:7e:5d:bc:9e:78:26:01:67:7c:4a:ae:
[...]

PR #6370, opened 2021-10-04
Add test case for certmonger CSR issue

tiran:
Certmonger used to generate FreeIPA server CSRs with invalid DER.
The CSR encodes a critical=FALSE component in X509v3 extensions although
FALSE is a default value. DER encoding requires that any component value
equal to default is not included.

See: https://github.com/pyca/cryptography/issues/6368
See: https://pagure.io/certmonger/issue/223
See: https://pagure.io/freeipa/issue/9005


tiran:
The test cases passes with 3.4.8 and fails with 35.0.0.

alex:
Ok, I've taken a provisional look at this, and I have a few questions/comments:

- Does RHEL plan to backport the certmonger fix to current releases so that broken CSRs leave the ecosystem in a timely fashion?
- If we do this, we're going to want to do it on a temporary basis, how long would you want us to maintain this behavior for?
- This PR, as written, changes the `Extension` struct that's also used by certificates, CRLs, OCSP, etc. We do not want to change the behavior for those, we're only willing to consider changing it for CSRs. Can you show what it'd look like to change just CSRs?

tiran:
> I think you'll need something like a `CsrExtension` struct which has the alternate definition.

The struct was the easy part. I'm struggling with Rust's trait system. I cannot figure out how to write a From/Into trait that converts from a sequence of CsrExtensions to a sequence of Extensions.

reaperhulk:
We're getting close-ish to 36.0. What are we planning to do here?

tiran:
> We're getting close-ish to 36.0. What are we planning to do here?

Alex wants the workaround only for CSR parsing. My Rust-fu is not strong enough yet to develop a fix on my own. I tried and failed to implement a trait that converts from CsrExtension to general extension.

PR #6373, opened 2021-10-04
Point at 3.10 now that it's released

alex:


alex:
Looks like I'm faster than GHA. Leaving this open and will re-run in a bit.

alex:
closing, @reaperhulk has a better version of this

PR #6383, opened 2021-10-05
Add new PKCS12 test vectors

felixfontein:
Part of #6348; see https://github.com/pyca/cryptography/pull/6348#discussion_r721883333.

CC @reaperhulk

### DIFF
--- a/docs/development/test-vectors.rst
+++ b/docs/development/test-vectors.rst
@@ -627,104 +627,104 @@ Custom PKCS12 Test Vectors
   password ``cryptography`` and no private key.
 * ``pkcs12/no-name-no-pwd.p12`` - A PKCS12 file containing a cert
   (``x509/custom/ca/ca.pem``) and key (``x509/custom/ca/ca_key.pem``),
-  as well as two additional certificates (``x509/custom/dsa_selfsigned_ca.pem``
+  as well as two additional certificates (``x509/cryptography.io.pem``
   and ``x509/letsencryptx3.pem``).
 * ``pkcs12/name-all-no-pwd.p12`` - A PKCS12 file containing a cert
   (``x509/custom/ca/ca.pem``) and key (``x509/custom/ca/ca_key.pem``)
   with friendly name ``name``, as well as two additional certificates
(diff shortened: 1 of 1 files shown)


reaperhulk:
One big question here, why use a DSA cert? Nobody uses those and some day we may want to deprecate support, which these vectors will make more of a challenge 😄

felixfontein:
I basically used the same certs that the other PKCS12 tests were using, and cert2 happens to use a DSA key. Should I replace them by something else (like `x509/cryptography.io.pem`)?

reaperhulk:
Yeah please do, sorry about that 😢

felixfontein:
I'm using the following script to create the vectors: https://gist.github.com/felixfontein/f750763fd5773fbf0ab22af87239aab0 It's based on an earlier version of #6348 which had serialization code.

(Unfortunately it's not idempotent, I guess that is because PKCS12 serialization includes some randomness.)

docs/development/test-vectors.rst
```diff
@@ -625,6 +625,108 @@ Custom PKCS12 Test Vectors
 * ``pkcs12/cert-aes256cbc-no-key.p12`` - A PKCS12 file containing a cert
   (``x509/custom/ca/ca.pem``) encrypted via AES 256 CBC with the
   password ``cryptography`` and no private key.
+* ``pkcs12/no-name-no-pwd.p12`` - A PKCS12 file containing a cert
+  (``x509/custom/ca/ca.pem``) and key (``x509/custom/ca/ca_key.pem``),
+  as well as two additional certificates (``x509/custom/dsa_selfsigned_ca.pem``
```
reaperhulk, line 630: Looks like the docs need updating still?

felixfontein: :laughing: it's early in the morning for me... :)

felixfontein: Fixed in cdc2cd7201df408427cd92d1eddb1329b9b56799.

reaperhulk (approved):

felixfontein:
@reaperhulk thanks for reviewing and merging!

PR #6403, opened 2021-10-11
Specify the out length when obtaining the tag for poly1305

alex:


alex:
See https://github.com/openssl/openssl/pull/16789

PR #6512, opened 2021-10-31
make all-green only pass if the tests actually pass

reaperhulk:
In GitHub Actions if a job needs a previous job (which fails) then the job gets skipped. If a required status is skipped then it's considered passing. This confluence is not good when we want to have an aggregate job we use as a required status to gate for auto-merge queue.

To fix this, we always run the job and then explicitly check the result of every dependent job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -498,5 +498,8 @@ jobs:
     # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
     runs-on: ubuntu-latest
     needs: [linux, linux-distros, linux-rust, linux-rust-coverage, macos, windows, linux-downstream]
+    if: ${{ always() }}
     steps:
+    - run: echo "😢"; exit 1
+      if: ${{ needs.linux.result != 'success' || needs.linux-distros.result != 'success' || needs.linux-rust.result != 'success' || needs.linux-rust-coverage.result != 'success' || needs.macos.result != 'success' || needs.windows.result != 'success' || needs.linux-downstream.result != 'success' }}
     - run: echo "🎉"


.github/workflows/ci.yml
```diff
@@ -498,5 +498,10 @@ jobs:
     # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
     runs-on: ubuntu-latest
     needs: [linux, linux-distros, linux-rust, linux-rust-coverage, macos, windows, linux-downstream]
+    if: ${{ always() }}
     steps:
+    - run: echo "😢";exit 1
+      if: ${{ needs.linux.result == 'failure' || needs.linux-distros.result == 'failure' || needs.linux-rust.result == 'failure' || needs.linux-rust-coverage.result == 'failure' || needs.macos.result == 'failure' || needs.windows.result == 'failure' || needs.linux-downstream.result == 'failure' }}
+      # this step will run if the previous step is skipped because there is an implicit
+      # if: ${{ success() }} on all steps
```
alex, line 506:
```suggestion
    # this step will run if the previous step is skipped because there is an implicit
    # if: ${{ success() }} on all steps
```

or delete, not sure how useful this comment is

.github/workflows/ci.yml
```diff
@@ -498,5 +498,10 @@ jobs:
     # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
     runs-on: ubuntu-latest
     needs: [linux, linux-distros, linux-rust, linux-rust-coverage, macos, windows, linux-downstream]
+    if: ${{ always() }}
     steps:
+    - run: echo "😢";exit 1
```
alex, line 503:
```suggestion
    - run: echo "😢"; exit 1
```

alex (approved):

.github/workflows/ci.yml
```diff
@@ -498,5 +498,8 @@ jobs:
     # https://github.community/t/is-it-possible-to-require-all-github-actions-tasks-to-pass-without-enumerating-them/117957/4?u=graingert
     runs-on: ubuntu-latest
     needs: [linux, linux-distros, linux-rust, linux-rust-coverage, macos, windows, linux-downstream]
+    if: ${{ always() }}
     steps:
+    - run: echo "😢"; exit 1
+      if: ${{ needs.linux.result != 'success' || needs.linux-distros.result != 'success' || needs.linux-rust.result != 'success' || needs.linux-rust-coverage.result != 'success' || needs.macos.result != 'success' || needs.windows.result != 'success' || needs.linux-downstream.result != 'success' }}
```
webknjaz, line 504: Woah! This is mind-blowing! We've hit the same problem in aiohttp and I wasn't sure where the bug was.
Do you think it could be easier to maintain if "check" was two jobs? The dependency chain would be `[test matrix] -> pre-check -> check`. Then this check would become `if: ${{ needs.pre-check.result != 'success' }}`. Otherwise, it's easy to forget additionally added dependencies to this check also...

reaperhulk: Sounds like that path could work too, at the cost of one more serial job.

webknjaz: Yeah, so it's like a maintenance burden with possible security-related consequences vs. extra 10 seconds of CPU time in GHA.

webknjaz: On the second thought, I think I have an idea on how to iterate over the context values w/o needing an extra job. I'll need to experiment with that first...

webknjaz: Here's a demo with a Python oneliner: https://github.com/webknjaz/check-gha-test/actions/runs/1514147440.
I guess I'll probably make it into action since I'll need it in many places...

webknjaz: @reaperhulk so I ended up making this https://github.com/marketplace/actions/alls-green / https://github.com/re-actors/alls-green and I'm starting to integrate it in my repos (currently added to CherryPy and aiohttp). If you want, I can send a PR.
Here's a demo for when I manually canceled a workflow via GH UI: https://github.com/cherrypy/cheroot/runs/4414994316?check_suite_focus=true#step:2:113.
Successful outcome example: https://github.com/aio-libs/aiohttp/runs/4414927258?check_suite_focus=true#step:2:102.
I've also added the ability to allow failures for some of the needed jobs: https://github.com/webknjaz/check-gha-test/runs/4404086833?check_suite_focus=true#step:3:3.

webknjaz: @reaperhulk give it a try? https://github.com/pyca/cryptography/pull/7643

Issue #6499, opened 2021-10-30
More explicitly document backend deprecation


reaperhulk:
We have previously made it so `backend` isn't a required argument, but we've referred to them as `optional`. We should update our docs to note that backends are not only not required, they're not recommended. Since a huge number of callers pass backends these functions will all continue to take the argument for the foreseeable future (many years!), but it's probably time to officially deprecate `backend`.

As part of this do we also want to remove the backend section of our docs?

alex:
I'd go so far as to say we should delete the docs for them!

reaperhulk:
I think we can call this complete for the moment. At some point we'll want to revisit the entire way we parametrize over backend in the tests but I don't have the ambition for it at the moment.

PR #6518, opened 2021-11-02
deprecate backend part 2 of n

reaperhulk:
proposed faq entry and removal on a single file

refs #6499

### DIFF
--- a/docs/hazmat/primitives/asymmetric/ec.rst
+++ b/docs/hazmat/primitives/asymmetric/ec.rst
@@ -6,33 +6,26 @@ Elliptic curve cryptography
 .. module:: cryptography.hazmat.primitives.asymmetric.ec
 
 
-.. function:: generate_private_key(curve, backend=None)
+.. function:: generate_private_key(curve)
 
     .. versionadded:: 0.5
 
-    Generate a new private key on ``curve`` for use with ``backend``.
+    Generate a new private key on ``curve``.
 
     :param curve: An instance of :class:`EllipticCurve`.
-
-    :param backend: An optional instance of
-        :class:`~cryptography.hazmat.backends.interfaces.EllipticCurveBackend`.
 
     :returns: A new instance of :class:`EllipticCurvePrivateKey`.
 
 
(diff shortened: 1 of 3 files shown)


docs/hazmat/primitives/asymmetric/ec.rst
```diff
@@ -6,34 +6,33 @@ Elliptic curve cryptography
 .. module:: cryptography.hazmat.primitives.asymmetric.ec
 
 
-.. function:: generate_private_key(curve, backend=None)
+.. function:: generate_private_key(curve)
 
     .. versionadded:: 0.5
 
-    Generate a new private key on ``curve`` for use with ``backend``.
+    .. versionchanged:: 36.0
```
alex, line 13: Is putting one of these on every single function useful?

reaperhulk: I think that's worth some discussion, which is why I stopped at this PR 😄 

Let me give the primary case: we know that users search and use whatever docs pop up via the search engine. By putting a note on (many) functions noting that the signature has changed we can improve the experience for those people by directing them to the proper docs and lowering confusion. The cost is a bit of noise in the docs for people who are less confused, but this is a big change to people who have expected to pass `backend` or `default_backend()` for the past 8 years (even though we made it optional over a year ago).

alex: My arguments against documenting it everywhere:

- It's very noisy, and this may lead to it getting ignored even by people who need it!
- The most useful place for this information is the changelog
- It's annoying to write/review each of these

I could go either way though, truth be told that 3rd one is most motivating

reaperhulk: If we choose not to document it everywhere, should we still add `.. versionchanged:: 36.0` and just rely on the changelog? Or should we keep the FAQ entry but not link it?

alex: Keep FAQ, no vesion changed marker on all functions is my idea.

alex (approved):

Issue #2972, opened 2016-06-09
Add OPENSSL_config binding


agladkov:
By default openssl does not load config, so there are no ability to use additional engine like "gost".
Need to call `OPENSSL_config(NULL)` or build bindings with `OPENSSL_LOAD_CONF` defined ([man OPENSSL_config](https://www.openssl.org/docs/manmaster/crypto/OPENSSL_config.html))

reaperhulk:
We're happy to take a PR for this. `OPENSSL_config` has been deprecated in 1.1.0 but is still present so we can add this binding while preserving compatibility.

agladkov:
Since 1.1.0 there are `OPENSSL_init_crypto` for this purpose.
I'll prepare PR for `OPENSSL_config` and check existed PR for 1.1.0 (#2826)

PR #6536, opened 2021-11-03
remove OPENSSL_no_config

reaperhulk:


reaperhulk:
This was originally added in https://github.com/pyca/cryptography/pull/2974 but the ask in https://github.com/pyca/cryptography/issues/2972 was for the load config binding, not the no config one.

alex (approved):

PR #4089, opened 2018-01-19
Add bindings for DTLS support

gesslerpd:
Added binding options for BIO datagram control and DTLS HelloVerifyRequest cookies.

PR #6554, opened 2021-11-06
Begin building with BoringSSL in our CI, no tests yet

alex:


src/_cffi_src/openssl/bio.py
```diff
@@ -57,4 +59,13 @@
     free(ptr);
 }
 #endif
+
+#if CRYPTOGRAPHY_IS_BORINGSSL
+static const long Cryptography_HAS_BIO_DGRAM = 0;
+
+BIO *(*BIO_new_dgram)(int, int) = NULL;
```
reaperhulk, line 66: Do we actually need these? They were landed in https://github.com/pyca/cryptography/pull/4089 for DTLS support, but https://github.com/pyca/pyopenssl/pull/1026 added DTLS support to pyOpenSSL and explicitly does not use the dgram BIOs.

alex: I'd be delgihted to delete them! Separate PR incoming.

reaperhulk (approved):

PR #6565, opened 2021-11-08
build a universal2 wheel

reaperhulk:
This builds a universal2 wheel in our CI in addition to an x86_64 wheel. You can see it succeed here: https://github.com/pyca/cryptography/actions/runs/1433759288 and I have tested the resulting wheel on arm64 to confirm it works as expected.

We should determine if we want to ship wheels like this while we wait for macOS arm64 CI to exist.

### DIFF
--- a/.github/workflows/wheel-builder.yml
+++ b/.github/workflows/wheel-builder.yml
@@ -76,18 +76,39 @@ jobs:
           path: cryptography-wheelhouse/
 
   macos:
-    runs-on: macos-latest
+    runs-on: macos-11
     strategy:
       fail-fast: false
       matrix:
         PYTHON:
-          - VERSION: '3.8'
+          - VERSION: '3.10'
             ABI_VERSION: 'cp36'
-            DOWNLOAD_URL: 'https://www.python.org/ftp/python/3.8.10/python-3.8.10-macosx10.9.pkg'
-            BIN_PATH: '/Library/Frameworks/Python.framework/Versions/3.8/bin/python3'
+            # Despite the name, this is built for the macOS 11 SDK on arm64 and 10.9+ on intel
+            DOWNLOAD_URL: 'https://www.python.org/ftp/python/3.10.0/python-3.10.0post2-macos11.pkg'
+            BIN_PATH: '/Library/Frameworks/Python.framework/Versions/3.10/bin/python3'
+            DEPLOYMENT_TARGET: '10.10'
+            # This archflags is default, but let's be explicit
+            ARCHFLAGS: '-arch x86_64 -arch arm64'
+            # See https://github.com/pypa/cibuildwheel/blob/c8876b5c54a6c6b08de5d4b1586906b56203bd9e/cibuildwheel/macos.py#L257-L269
+            # This will change in the future as we change the base Python we
+            # build against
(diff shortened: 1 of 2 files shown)


reaperhulk:
One additional outstanding question: Is `-march=core2` necessary for x86_64 builds? I think we're fine since Python will build the module with the set of flags it was built with itself, and we've built OpenSSL itself with a conservative set of flags already as well.

reaperhulk:
Fixes #5918

reaperhulk:
@alex and I had a sidebar discussion and we agreed that the path forward here is to add an m1 builder to our CI so that we have some basic assurance. This will likely take the form of a new runner that only runs on merges to `main`, in the same style as our linux aarch64 runners (https://github.com/pyca/cryptography/blob/55ec7169c702da6a426879a7c8cdc91fe3953117/.github/workflows/arm64.yml). 

Building this will require an extra M1 mac dedicated solely to our CI for now. Always be spending money on open source projects.

kainjow:
Is the need to test on Apple Silicon? If so it seems there isn’t much progress that on that yet: https://github.com/actions/virtual-environments/issues/2187

Building can be done on Intel and the project could just say it’s experimental and not fully supported.

Otherwise it seems the way to support native Apple Silicon for cryptography is manually combining the binaries with lipo.

alex:
Yes, we need to test on this architecture in order to ship wheels for it.

@reaperhulk owns an M1 Mac Mini and we'll eventually incorporate that into our CI.

reaperhulk:
We expect to land this work once I'm able to repurpose the M1 mini to our CI. That's likely to be early Jan 2022.

kainjow:
Great! Have you considered MacStadium? https://www.macstadium.com/opensource

reaperhulk:
Do you mean applying through https://www.macstadium.com/opensource-application for free resources? That might not be a bad idea actually.

The current plan is just to repurpose an M1 mini I already own. Actions doesn't natively support arm64 but `arch -arm64` can be used to invoke native binaries from the rosetta shell. Unfortunately, the security model for self-hosted actions runners isn't very good unless you use their new [autoscaling](https://docs.github.com/en/actions/hosting-your-own-runners/autoscaling-with-self-hosted-runners#using-ephemeral-runners-for-autoscaling) feature...but to use that effectively on the M1 we'd need to write software integrating with Hypervisor.framework to manage VMs. That's not likely to happen (although never say never), so instead we'll probably just limit runs to merges to `main`, which is significantly inferior but does give us reasonable ongoing assurance that things are working as expected.

All of that can happen with a MacStadium based M1 as well, with the added benefit of not being yet another computer I have to run in my house.

reaperhulk:
This will be reviewable once #6672 lands

reaperhulk:
Just as an update here we now have an M1 mini thanks to the generosity of MacStadium. Thanks for suggesting it @kainjow

alex (approved):

kainjow:
What’s the expected release date for 37.0.0?

alex:
None planned. If history is any indicator, in the next few months.

בתאריך יום ו׳, 3 בדצמ׳ 2021, 10:31, מאת Kevin Wojniak ‏<
***@***.***>:

reaperhulk:
If you are asking because you want a wheel uploaded we actually uploaded a universal2 wheel for 36 as well.

PR #6569, opened 2021-11-08
Add Encryption2021 serialization encryption type

tiran:
``BestAvailableEncryption`` provides a curated selection of encryption
algorithms that can change over time. While a curated selection is best
for majority of users, some users in a LTS/Enterprise environment need
more stability.

I propose the addition of a curated and versioned selection of
encryption types. Every time the algorithm selection of
``BestAvailableEncryption``, a new ``Encryption$VERSION`` is added. Once
the algorithms of an old ``Encryption$VERSION`` are considered too weak
(for some definition of "weak"), the type is deprecated. When the
algorithm is broken, it is dropped.

In practice this will happen very rarely. The current selection uses
AES-256.

### DIFF
--- a/docs/hazmat/primitives/asymmetric/serialization.rst
+++ b/docs/hazmat/primitives/asymmetric/serialization.rst
@@ -978,6 +978,17 @@ Serialization Encryption Types
 
     :param bytes password: The password to use for encryption.
 
+.. class:: Encryption2021(password)
+
+    Encrypt using the best available encryption for a given key in year 2021.
+    The algorithm will never change. In the future, encryption may fail or the
+    class may be removed if the encryption algorithm is considered too weak.
+
+    As of now it uses the same algorithms as
+    :class:`~cryptography.hazmat.primitives.serialization.BestAvailableEncryption`.
+
+    :param bytes password: The password to use for encryption.
+
 .. class:: NoEncryption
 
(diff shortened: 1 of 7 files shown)


alex:
I am not a fan of this API, and before I'd be willing to consider it I need to hear the motivation fleshed out in more detail.

`BestAvailableEncryption` hasn't changed in the lifetime of this project for PEM serialization. This makes the motivation here very theoretical and hard to assess what compatibility constraints would lead to us needing to make changes here. This is further compounded by the fact that right now 2021 is behaviorally identical!

My other concern is future planning for what it would mean to update BestAvailableEncryption but not immediately need to remove 2021 for security issues. To me a scenario where we allowed these to diverge is a very "bug fixing is how you improve security" mindset, which I think is wrong (cf https://alexgaynor.net/2019/mar/07/chrome-windows-exploit-security-beyond-bugfixes/) -- security features and improvements are as important as bug fixes.

Finally, the existing PEM encryption is awful, the KDF is terrible and it doesn't authenticate ciphertext. If there was a better option I'd want to switch to it immediately, but there isn't AFAIK.

reaperhulk:
@tiran do you intend to revisit this or can we close?

reaperhulk:
Closing, but if you want to discuss again feel free to reopen.

PR #6571, opened 2021-11-08
Move extension encoding code to new crate x509::extensions

felixfontein:
See https://github.com/pyca/cryptography/pull/6346#discussion_r739917601.

### DIFF
--- a/src/rust/src/x509/crl.rs
+++ b/src/rust/src/x509/crl.rs
@@ -9,19 +9,6 @@ use pyo3::ToPyObject;
 use std::convert::TryInto;
 use std::sync::Arc;
 
-lazy_static::lazy_static! {
-    pub(crate) static ref CRL_NUMBER_OID: asn1::ObjectIdentifier<'static> = asn1::ObjectIdentifier::from_string("2.5.29.20").unwrap();
-    pub(crate) static ref DELTA_CRL_INDICATOR_OID: asn1::ObjectIdentifier<'static> = asn1::ObjectIdentifier::from_string("2.5.29.27").unwrap();
-    pub(crate) static ref ISSUER_ALTERNATIVE_NAME_OID: asn1::ObjectIdentifier<'static> = asn1::ObjectIdentifier::from_string("2.5.29.18").unwrap();
-    pub(crate) static ref AUTHORITY_INFORMATION_ACCESS_OID: asn1::ObjectIdentifier<'static> = asn1::ObjectIdentifier::from_string("1.3.6.1.5.5.7.1.1").unwrap();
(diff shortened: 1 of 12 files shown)


src/rust/src/x509/crl.rs
```diff
@@ -4,22 +4,22 @@
...
-    static ref AUTHORITY_INFORMATION_ACCESS_OID: asn1::ObjectIdentifier<'static> = asn1::ObjectIdentifier::from_string("1.3.6.1.5.5.7.1.1").unwrap();
-    static ref AUTHORITY_KEY_IDENTIFIER_OID: asn1::ObjectIdentifier<'static> = asn1::ObjectIdentifier::from_string("2.5.29.35").unwrap();
-    static ref ISSUING_DISTRIBUTION_POINT_OID: asn1::ObjectIdentifier<'static> = asn1::ObjectIdentifier::from_string("2.5.29.28").unwrap();
-    static ref FRESHEST_CRL_OID: asn1::ObjectIdentifier<'static> = asn1::ObjectIdentifier::from_string("2.5.29.46").unwrap();
-    static ref CRL_REASON_OID: asn1::ObjectIdentifier<'static> = asn1::ObjectIdentifier::from_string("2.5.29.21").unwrap();
-    static ref CERTIFICATE_ISSUER_OID: asn1::ObjectIdentifier<'static> = asn1::ObjectIdentifier::from_string("2.5.29.29").unwrap();
-    static ref INVALIDITY_DATE_OID: asn1::ObjectIdentifier<'static> = asn1::ObjectIdentifier::from_string("2.5.29.24").unwrap();
+    pub(crate) static ref CRL_NUMBER_OID: asn1::ObjectIdentifier<'static> = asn1::ObjectIdentifier::from_string("2.5.29.20").unwrap();
```
reaperhulk, line 13: Let's just consolidate all our lazy_static OIDs for extensions in the extensions module rather than pub(crate)'ing them here. (If they still need to be public to the crate that's fine, but since they're all extension OIDs it makes sense to put them all in extensions.rs)

reaperhulk (commented):
Only one real comment here other than that we may want to reorganize our extension parsing a bit in a future PR as well.

src/rust/src/x509/crl.rs
felixfontein: Implemented in 8ce7433051cc49982af6e9f1c62e1d8518ae322f.

reaperhulk:
The more I look at this the more I'm sure we need to also merge the parsers if this is the direction we're going, but I think that should be a subsequent PR.

What do you think about putting all the OIDs in one spot? Maybe we need an `oid.rs` module for this just so we don't have them scattered everywhere...

reaperhulk (approved):

felixfontein:
@reaperhulk thanks for reviewing and merging!

PR #6572, opened 2021-11-08
retry paramiko tests if they fail

reaperhulk:
flaky tests are so annoying

### DIFF
--- a/.github/downstream.d/paramiko.sh
+++ b/.github/downstream.d/paramiko.sh
@@ -10,6 +10,7 @@ case "${1}" in
         ;;
     run)
         cd paramiko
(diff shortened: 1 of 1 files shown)


alex:
a) Have we reported this upstream, I think that's a pre-req
b) I don't think this works due to the `-ex`?

.github/downstream.d/paramiko.sh
```diff
@@ -10,7 +10,7 @@ case "${1}" in
         ;;
     run)
         cd paramiko
-        inv test
+        inv test || inv test
```
alex, line 13: plz add a comment with link to upstream bug

alex (approved):

PR #6593, opened 2021-11-12
encode KeyUsage with unused bits

reaperhulk:


alex (approved):
Does this apply to `encode_distribution_point_reasons` as well?

PR #6606, opened 2021-11-15
add AES OCB3 test vectors

reaperhulk:


alex:
Please don't mix multiple vector sources in a single file, it makes it incredibly difficult to review.

reaperhulk:
This is actually all one source (the RFC). I should remove that other info and rename the file.

alex (approved):

PR #6607, opened 2021-11-15
OCB3 support

### LINKED ISSUE #6136: [Feature request] AES-OCB
Old issues: https://github.com/pyca/cryptography/issues/1770 https://github.com/pyca/cryptography/issues/2968

The latter mentions patent issues. There is a license issued to OpenSSL specifically, so I wouldn't expect that to be a problem. However note that as of 2021 the patents have not been renewed, thereby allowing the IP to move into public domain: https://mailarchive.ietf.org/arch/msg/cfrg/qLTveWOdTJcLn4HP3ev-vrj05Vg/. Therefore there should be no more blockers to expose that particular API.

reaperhulk:
fixes #6136

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/aead.py
+++ b/src/cryptography/hazmat/backends/openssl/aead.py
@@ -116,7 +116,9 @@ def _encrypt(backend, cipher, nonce, data, associated_data, tag_length):
     _process_aad(backend, ctx, associated_data)
     processed_data = _process_data(backend, ctx, data)
     outlen = backend._ffi.new("int *")
-    # OCB can return up to 15 bytes (16 byte block - 1) in finalization
+    # All AEADs we support besides OCB are streaming so they return nothing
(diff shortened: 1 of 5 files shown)


src/cryptography/hazmat/backends/openssl/aead.py
```diff
@@ -113,9 +116,11 @@ def _encrypt(backend, cipher, nonce, data, associated_data, tag_length):
     _process_aad(backend, ctx, associated_data)
     processed_data = _process_data(backend, ctx, data)
     outlen = backend._ffi.new("int *")
-    res = backend._lib.EVP_CipherFinal_ex(ctx, backend._ffi.NULL, outlen)
+    # OCB can return up to 15 bytes (16 byte block - 1) in finalization
```
alex, line 119: Say that all other modes are streaming and thus don't return anything.

alex (approved):

PR #6614, opened 2021-11-17
type some more things in the tests

reaperhulk:
This PR defines a return type for `load_vectors_from_file`, which in turn expands typing significantly through our asymmetric tests. A followup PR extends the return type through `_load_cert`, which significantly expands type information in x509 tests.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -38,7 +38,12 @@ def raises_unsupported_algorithm(reason):
     assert exc_info.value._reason is reason
 
 
-def load_vectors_from_file(filename, loader, mode="r"):
+T = typing.TypeVar("T")
+
+
+def load_vectors_from_file(
+    filename, loader: typing.Callable[..., T], mode="r"
+) -> T:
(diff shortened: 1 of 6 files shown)


tests/utils.py
```diff
@@ -38,7 +38,12 @@ def raises_unsupported_algorithm(reason):
     assert exc_info.value._reason is reason
 
 
-def load_vectors_from_file(filename, loader, mode="r"):
+T = typing.TypeVar("T")
+
+
+def load_vectors_from_file(
+    filename, loader: typing.Callable[..., T], mode="r"
```
alex, line 45: Going to merge this for now, but this shouldn't be a `...`, it should be an IO, right?

alex (approved):

tests/utils.py
reaperhulk: Probably, I basically just skipped typing everything other than the return because that’s 99% of the value of the type here. We can type the rest in the future

PR #6616, opened 2021-11-17
more typing in the tests

reaperhulk:
This adds typing to `_load_cert` and now that it returns a real type we can improve our tests (and switch to using our better APIs that do more effective type narrowing in many cases).

One nuance here is that now that we allow x25519/x448 as public key types we have an odd asymmetry where it's sane to encode that in SKI but nonsensical to have it in AKI. I've left the mypy types such that this is the case (although you can defeat the typing by generating an SKI and then create the AKI from SKI. Sorry, Python typing isn't expressive enough to prevent that).

### DIFF
--- a/tests/x509/test_x509.py
+++ b/tests/x509/test_x509.py
@@ -62,7 +62,10 @@ class FakeGeneralName(object):
     value = utils.read_only_property("_value")
 
 
-def _load_cert(filename, loader, backend=None):
+T = typing.TypeVar("T")
+
+
+def _load_cert(filename, loader: typing.Callable[..., T], backend=None) -> T:
     cert = load_vectors_from_file(
         filename=filename,
         loader=lambda pemfile: loader(pemfile.read(), backend),
@@ -200,7 +203,7 @@ class TestCertificateRevocationList(object):
             backend,
         )
         with pytest.raises(TypeError):
-            crl1 < crl1
+            crl1 < crl1  # type: ignore[operator]
 
     def test_update_dates(self, backend):
         crl = _load_cert(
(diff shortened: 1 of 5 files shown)


tests/x509/test_x509.py
```diff
@@ -62,7 +62,10 @@ def __init__(self, value):
     value = utils.read_only_property("_value")
 
 
-def _load_cert(filename, loader, backend=None):
+T = typing.TypeVar("T")
+
+
+def _load_cert(filename, loader: typing.Callable[..., T], backend=None) -> T:
```
alex, line 68: Same deal, this shouldn't be `...`, this should just be `bytes`?

tests/x509/test_x509.py
```diff
@@ -263,6 +266,7 @@ def test_get_revoked_certificate_by_serial_number(self, backend):
         )
         serial_number = 725064303890588110203033396814564464046290047507
         revoked = crl.get_revoked_certificate_by_serial_number(serial_number)
+        assert isinstance(revoked, x509.RevokedCertificate)
```
alex, line 269: Why is this required?

alex (approved):

tests/x509/test_x509.py
reaperhulk: Yes (although it'd be `[bytes]` I suppose). I skipped it for the same reason I skipped the last one: I cared exclusively about the return type and figured we could come back around and completely type it later.

tests/x509/test_x509.py
reaperhulk: `get_revoked_certificate_by_serial_number ` returns `Optional[RevokedCertificate]` so we need to assert on it. We can do isinstance or we could just assert not None, but otherwise the next line will fail

PR #6626, opened 2021-11-20
support negative serials in certificate parsing

reaperhulk:
but raise a warning every time we see it. also proactively raise on
initial parse of the certificate, not just when accessing the
serial_number attribute

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,7 +31,7 @@ Changelog
 * Allow parsing of negative serial numbers in certificates. Negative serial
   numbers are prohibited by :rfc:`5280` so a deprecation warning will be
   raised whenever they are encountered. A future version of ``cryptography``
-  will drop support.
+  will drop support for parsing them.
 * Added support for parsing PKCS12 files with friendly names for all
   certificates with
   :func:`~cryptography.hazmat.primitives.serialization.pkcs12.load_pkcs12`,
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -28,6 +28,10 @@ Changelog
   in the same file.
 * Fixed parsing of :class:`~cryptography.x509.CertificatePolicies` extensions
   containing legacy ``BMPString`` values in their ``explicitText``.
+* Allow parsing of negative serial numbers in certificates. Negative serial
+  numbers are prohibited by :rfc:`5280` so a deprecation warning will be
+  raised whenever they are encountered. A future version of ``cryptography``
+  will drop support.
 * Added support for parsing PKCS12 files with friendly names for all
   certificates with
   :func:`~cryptography.hazmat.primitives.serialization.pkcs12.load_pkcs12`,
--- a/src/rust/src/asn1.rs
+++ b/src/rust/src/asn1.rs
@@ -5,6 +5,7 @@
 use crate::x509::Name;
 use pyo3::basic::CompareOp;
 use pyo3::ToPyObject;
+use pyo3::types::IntoPyDict;
 
 pub enum PyAsn1Error {
     Asn1(asn1::ParseError),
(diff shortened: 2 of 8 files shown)


src/rust/src/asn1.rs
```diff
@@ -77,21 +78,42 @@ struct DssSignature<'a> {
     s: asn1::BigUint<'a>,
 }
 
-pub(crate) fn big_asn1_uint_to_py<'p>(
+pub(crate) fn big_byte_slice_to_py_int<'p>(
     py: pyo3::Python<'p>,
-    v: asn1::BigUint<'_>,
+    v: &'_ [u8],
 ) -> pyo3::PyResult<&'p pyo3::PyAny> {
+    let signed = big_byte_slice_is_negative(py, v)?;
```
alex, line 85: This seems like a weird way to solve this, why not just take `signed` as a parameter that teh caller specifies based on context?

src/rust/src/asn1.rs
```diff
@@ -77,21 +78,42 @@ struct DssSignature<'a> {
...
+    let kwargs = [("signed", signed)].into_py_dict(py);
+    int_type.call_method("from_bytes", (v, "big"), Some(kwargs))
+}
+
+pub(crate) fn big_byte_slice_is_negative(
+    py: pyo3::Python<'_>,
+    bytes: &'_ [u8],
+) -> pyo3::PyResult<bool> {
+    if bytes[0] & 0x80 != 0 {
+        let cryptography_warning = py.import("cryptography.utils")?.getattr("DeprecatedIn36")?;
```
alex, line 96: This is not the right place for this warning. This function has an ostensibly generic purpose and instead it's got behavior that's totally specific to serials.

src/rust/src/x509/certificate.rs
```diff
@@ -350,6 +353,10 @@ fn load_der_x509_certificate(py: pyo3::Python<'_>, data: &[u8]) -> PyAsn1Result<
     let raw = OwnedRawCertificate::try_new(Arc::from(data), |data| asn1::parse_single(data))?;
     // Parse cert version immediately so we can raise error on parse if it is invalid.
     cert_version(py, raw.borrow_value().tbs_cert.version)?;
+    // determine if the serial is negative and raise a warning if it is. We want to drop support
+    // for this sort of invalid encoding eventually.
+    big_byte_slice_is_negative(py, raw.borrow_value().tbs_cert.serial.as_bytes())?;
```
alex, line 358: See! This functions behavior in no way matches it's name.

CHANGELOG.rst
```diff
@@ -28,6 +28,10 @@ Changelog
   in the same file.
 * Fixed parsing of :class:`~cryptography.x509.CertificatePolicies` extensions
   containing legacy ``BMPString`` values in their ``explicitText``.
+* Allow parsing of negative serial numbers in certificates. Negative serial
+  numbers are prohibited by :rfc:`5280` so a deprecation warning will be
+  raised whenever they are encountered. A future version of ``cryptography``
+  will drop support.
```
alex, line 34:
```suggestion
  will drop support for parsing them.
```

src/rust/src/asn1.rs
```diff
@@ -77,21 +78,38 @@ struct DssSignature<'a> {
     s: asn1::BigUint<'a>,
 }
 
-pub(crate) fn big_asn1_uint_to_py<'p>(
+pub(crate) fn big_byte_slice_to_py_int<'p>(
     py: pyo3::Python<'p>,
-    v: asn1::BigUint<'_>,
+    v: &'_ [u8],
 ) -> pyo3::PyResult<&'p pyo3::PyAny> {
+    warn_if_negative(py, v)?;
```
alex, line 85: I still object to this being part of the behavior of `big_byte_slice_to_py_int.

Functions behavior should be predictable and well factored -- generic functions like type conversion shouldn't have context specific warnings like this. Just put `warn_if_negative` calls in the serial accessors. And please also name it `warn_if_negative_serial` based on the warning message.

alex (dismissed):

src/rust/src/x509/ocsp_req.rs
```diff
@@ -95,7 +95,10 @@ impl OCSPRequest {
 
     #[getter]
     fn serial_number<'p>(&self, py: pyo3::Python<'p>) -> Result<&'p pyo3::PyAny, PyAsn1Error> {
-        Ok(big_asn1_uint_to_py(py, self.cert_id()?.serial_number)?)
+        Ok(big_byte_slice_to_py_int(
+            py,
+            self.cert_id()?.serial_number.as_bytes(),
+        )?)
```
alex, line 101: This is uncovered because of silliness. assign `self.cert_id()?.serial_number.as_bytes()` to a local var

alex (approved):

PR #6629, opened 2021-11-20
support parsing bitstring values in DNs

### LINKED ISSUE #3542: X.509 OID 2.5.4.45 cannot be read because it's a bit string, not text
\_\_repr\_\_ of the certificate object for the attached X.509 certificate fails with the exception
      ValueError: Unsupported ASN1 string type. Type: 3

I tracked it down to OID 2.5.4.45, which is of type "BIT STRING (64 bit)", according to https://lapo.it/asn1js/, which is the correct type for the OID if we are to believe http://www.alvestrand.no/objectid/2.5.4.45.html.

It looks like 'cryptography' is insisting name attributes must be text, and using decode_asn._asn1_string_to_utf8 accordingly, but in this case there is no text conversion and decode_asn1._asn1_string_to_bytes should have been used instead.

Attached, [sup_ds.zip](https://github.com/pyca/cryptography/files/1004488/sup_ds.zip) containing:
sup_ds.crt: A sample offending certificate.
sup_ds.py: Reproducing code.
sup_ds.out: Traceback when run against cryptography 1.8.1, openssl 1.1.0e.

reaperhulk:
fixes #3542


src/cryptography/x509/name.py
```diff
@@ -79,20 +86,31 @@ def _escape_dn_value(val: str) -> str:
...
+                raise TypeError("value must be bytes for BitString")
+        else:
+            if not isinstance(value, str):
+                raise TypeError("value argument must be a str")
 
         if (
             oid == NameOID.COUNTRY_NAME
             or oid == NameOID.JURISDICTION_COUNTRY_NAME
         ):
+            if not isinstance(value, str):
```
alex, line 109: There's a lot more than just these two that need to be strings, right?

src/cryptography/x509/name.py
```diff
@@ -91,14 +98,22 @@ def __init__(
             raise TypeError(
                 "oid argument must be an ObjectIdentifier instance."
             )
-
-        if not isinstance(value, str):
-            raise TypeError("value argument must be a str.")
+        if _type == _ASN1Type.BitString:
+            if not oid == NameOID.X500_UNIQUE_IDENTIFIER:
```
alex, line 102:
```suggestion
            if oid != NameOID.X500_UNIQUE_IDENTIFIER:
```

alex (approved):

PR #6630, opened 2021-11-20
add very limited RSASSAPSS x509 certificate support

### LINKED ISSUE #2850: Support PSS X509 Signatures
This will require an API change to `sign`.

### LINKED ISSUE #4858: x509._Certificate.public_key does not work with RSASSA_PSS certs
Cause: in `cryptography/hazmat/backends/openssl/backend.py` `_evp_pkey_to_public_key`
cryptography is only testing for `self._lib.EVP_PKEY_RSA` and not `EVP_PKEY_RSA_PSS` as well.

I guess the same problem will occur with RSA OAEP certificates.

* Versions of Python, ``cryptography``, ``cffi``, ``pip``, and ``setuptools``
  you're using
    * Python 3.7.3
    * cryptography 2.6.1
    * cffi 1.11.5
    * pip 19.0.3
    * setuptools 40.4.3
* How you installed ``cryptography`` : pip

1. Create a self-signed certificate with RSA 2048 RSASSA_PSS; OpenSSL can self-verify this
example cert, i.e., `openssl verify -CAfile  sample.pem sample.pem` gives `OK`.
```
-----BEGIN CERTIFICATE-----
MIIDLTCCAhWgAwIBAgIUCD6ZpVYd5vCgzRh5ZIqYupE2VKIwDQYJKoZIhvcNAQEK
MAAwJTESMBAGA1UECgwJYm9vdHN0cmFwMQ8wDQYDVQQDDAZyb290Y2EwHhcNMTYw
MTAxMDAwMDAwWhcNNDAwMTAxMDAwMDAwWjAlMRIwEAYDVQQKDAlib290c3RyYXAx
DzANBgNVBAMMBnJvb3RjYTCCAVYwQQYJKoZIhvcNAQEKMDSgDzANBglghkgBZQME
AgEFAKEcMBoGCSqGSIb3DQEBCDANBglghkgBZQMEAgEFAKIDAgEgA4IBDwAwggEK
AoIBAQCuXwLURTDREKWTeBxUQWQvj/hDVc0+PruZtBF5voNAZCjKSOKHFLasmCDw
JuEHjj7ZHDF1JWZVGUbz3p5P+TiFmO/c1Wgb5IyAxdiDUGZvSVX3uC/X8EG/1MQz
bwcDpqiadoIjL59jUJ0g2BJnx81NvpNgpe0rmK7aU52sPKJme31Ttd/lO8VJ3Mps
lzpH0qzDJEcE3+lBF+AOJf2XDbPTFlbuPvDZHE5tVmdYC2IOad8U0Q/FLOhpMFOX
j0n6tKf+Z14+7+xu7RV3gj/NMm0CXWG3ibTSOSrbyvreI0dHgZL57RdqCSE9scjA
/1tD0a7UINhPBDZc6HaUqQUsQCPhAgMBAAGjITAfMB0GA1UdDgQWBBSpiALqV+wo
zwcvMEvwYFLKe/vDPzANBgkqhkiG9w0BAQowAAOCAQEAEdp7sFyQA9g3Vk1KsrAB
UFKqEe1a0azE4TRz2SRktRCswgv7iae0CiBGtPrzBNS6MlketixTfF1npEi7wuDn
/00XRdgHCBIRGvemATx8oSP4qVrHUud2y/DLZOZBGYiasSHHgybsnikFZppGFp7m
1D3Tti+GEsaANwRH5GW7h8f9hTMluwXlnNwyT/83yKq9Uih0eFEWHtf8hFswpCEK
4swEBwBHUiCZs1O702H36xEnoayOWnIWAkV8ZccEjfbCHs+rnU+nGI5UfXg86VRb
9iqcII7xyzPxb/HZRBusI9Uu/xiiqLjWqmx/ZMeadhiQndCoOj2yR8YT+G1BqcLa
pA==
-----END CERTIFICATE-----
```

2. Load this cert into cryptography
```
cert = x509.load_pem_x509_certificate(b'-----BEGIN ...', backend=default_backend())
cert.public_key()
  File "/home/Projects/venv37/lib/python3.7/site-packages/cryptography/hazmat/backends/openssl/x509.py", line 89, in public_key
    return self._backend._evp_pkey_to_public_key(pkey)
  File "/home/Projects/venv37/lib/python3.7/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 556, in _evp_pkey_to_public_key
    raise UnsupportedAlgorithm("Unsupported key type.")
cryptography.exceptions.UnsupportedAlgorithm: Unsupported key type.
```

reaperhulk:
fixes #4858

also fixes #2850, albeit not completely. I’d rather see if we get future requests for encode and parameters with new issues though.


src/rust/src/x509/certificate.rs
```diff
@@ -126,6 +126,18 @@ impl Certificate {
             py,
             &asn1::write_single(&self.raw.borrow_value().tbs_cert.spki),
         );
+        let backend = py.import("cryptography.hazmat.backends.openssl.backend")?;
+        let lib = backend.getattr("backend")?.getattr("_lib")?;
+        let has_pss = lib
+            .getattr("Cryptography_HAS_EVP_PKEY_RSA_PSS")?
+            .extract::<u8>()?;
+        if self.raw.borrow_value().signature_alg.oid == *x509::oid::RSASSA_PSS_OID && has_pss == 0 {
```
alex, line 134: Err, is it really the signature alg we need to check ehre? Surely it's the spki algorithm?

tests/x509/test_x509.py
```diff
@@ -734,6 +734,35 @@ def test_load_multiple_sections(self, backend):
...
+        cert = _load_cert(
+            os.path.join("x509", "custom", "rsa_pss.pem"),
+            x509.load_pem_x509_certificate,
+            backend,
+        )
+        assert cert.signature_algorithm_oid == SignatureAlgorithmOID.RSASSA_PSS
+        public_key = cert.public_key()
+        assert isinstance(public_key, rsa.RSAPublicKey)
+        with pytest.raises(UnsupportedAlgorithm):
+            cert.signature_hash_algorithm
```
alex, line 751: There's no way here to actually get the PSS parameters, which would be required to actually verify the signature, is there?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -633,7 +639,7 @@ def _evp_pkey_to_public_key(self, evp_pkey):
 
         key_type = self._lib.EVP_PKEY_id(evp_pkey)
 
-        if key_type == self._lib.EVP_PKEY_RSA:
+        if key_type in self._supported_rsa_key_types:
```
alex, line 642: What happens if you use a key that's got a type of `EVP_PKEY_RSA_PSS`? Does it do something weird?

alex:
How does this address #2850?

tests/x509/test_x509.py
reaperhulk: If we want to get them we'll need to parse this object ourselves (and also figure out an API for providing them, seems like overloading signature_hash_algorithm is probably not a good idea)

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: If it has PSS parameter restrictions OpenSSL will enforce them (e.g. hash alg, mask alg, salt length). Otherwise it behaves as a normal key.

src/rust/src/x509/certificate.rs
reaperhulk: Ugh, quite correct.

src/cryptography/hazmat/backends/openssl/backend.py
alex: That sounds awful? It means you have to know where your key came from to know how `.sign()` is going to behave.

This doesn't seem like a good way to go.

reaperhulk:
It doesn't fully address #2850, but I'm curious if anyone actually needs more than the ability to fetch the public key. Maybe they have out of band methods of knowing the PSS information in a signature and can therefore reliably verify as long as they have the key available? This is basically the worst possible PSS support put up as a basis for discussion to see if it's worth pursuing. Or, what we need to implement in Rust to get to the minimum viable.

reaperhulk:
Sitting on this a while now I believe we need to implement rust parsing for the PSS data to make this useful. Barring someone being willing to do that work (at the moment I am not interested) this will have to wait.

PR #6632, opened 2021-11-20
Try removing line whose purpose I don't understand

alex:


reaperhulk:
We don’t need this any more because we no longer support versions old enough to have that backport

reaperhulk (approved):

PR #6660, opened 2021-11-26
Drop supported for older LibreSSL

alex:
3.1 covers the oldest version used by versions of Alpine and OpenBSD that are supported by their upstreams.


CHANGELOG.rst
```diff
@@ -8,6 +8,9 @@ Changelog
 
 .. note:: This version is not yet released and is under active development.
 
+* **BACKWARDS INCOMPATIBLE:** Support for LibreSSL 2.9.x and 3.0.x. The new
```
reaperhulk, line 11:
```suggestion
* **BACKWARDS INCOMPATIBLE:** Dropped support for LibreSSL 2.9.x and 3.0.x. The new
```

reaperhulk (approved):

PR #6662, opened 2021-11-27
fix: fixed confusing typo on DSA docs

taliamax:
From the looks of it, this is just a typo but it was a bit confusing when reading it since the bit length of `q` is not any of the numbers listed, but the bit length of `p` is supposed to be


alex (approved):
thanks!

PR #6665, opened 2021-11-27
See if x25519_supported can be true on LibreSSL

alex:


alex:
Hahaha, absolutely not apparently. Wild.

PR #6672, opened 2021-12-02
add macos arm64 self-hosted CI

reaperhulk:
Sample run here: https://github.com/pyca/cryptography/actions/runs/1530356575

Doesn't run on PRs just like our arm64 linux runners. Coverage disabled because I didn't feel like figuring out why CTracer wasn't building (incidentally it takes ~4.5 min to run the test suite even with pure Python coverage on an M1).

All the `arch -arm64` prefixes are because the GHA runner is running under rosetta2 so we need to explicitly invoke the binaries as arm64 instead.

### DIFF
--- a/.github/workflows/macarm64.yml
+++ b/.github/workflows/macarm64.yml
@@ -1,0 +1,69 @@
+name: macOS arm64 CI
+on:
+  push:
+    branches:
+      - main
+      - '*.*.x'
+    tags:
+      - '*.*'
+      - '*.*.*'
+
+permissions: read-all
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  macos-arm64:
+    if: github.repository_owner == 'pyca'
+    runs-on: [self-hosted, macos, ARM64]
+    name: "macOS arm64"
+    strategy:
+      fail-fast: false
+      matrix:
+        PYTHON:
+          - {TOXENV: "py310", BIN_PATH: '/Library/Frameworks/Python.framework/Versions/3.10/bin/python3'}
+    timeout-minutes: 10
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/macarm64.yml
```diff
@@ -0,0 +1,62 @@
...
+          persist-credentials: false
+      - uses: actions/cache@v2.1.7
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            src/rust/target/
+          key: ${{ runner.os }}-${{ matrix.IMAGE.IMAGE }}-cargo-2-${{ hashFiles('**/Cargo.lock') }}
```
alex, line 37: There is no image, this cache key is wrong.

.github/workflows/macarm64.yml
```diff
@@ -0,0 +1,62 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  macos-arm64:
+    if: github.repository_owner == 'pyca'
+    runs-on: [self-hosted, macos, ARM64]
+    name: "macOS arm64"
+    timeout-minutes: 10
```
alex, line 22: This should have a matrix for python version, even if there's only one thing in it ATM.

tox.ini
```diff
@@ -23,6 +23,14 @@ commands =
     pip list
     pytest -n auto --durations=10 {posargs} tests/
 
+# Temp coverage disable for macOS arm64 because CTracer isn't compiling
+# probably due to the crazy rosetta arch -arm64 shenanigans
```
alex, line 27: Is there a bug that should be filed somewhere?

alex (approved):

PR #6695, opened 2021-12-12
PoC: Make it possible to write code that interacts with OpenSSL in Rust

messense:
Implements #6634

TODO:

- [x] Bump MSRV to 1.48.0 for the openssl crate.
- [ ] Fix tests on PyPy
- [x] Deal with `Py_LIMITED_API`


src/_cffi_src/build_openssl.py
```diff
@@ -112,3 +112,21 @@ def _extra_compile_args(platform):
...
+    import sysconfig
+    from cffi import recompiler
+
+    module_name, source, source_extension, kwds = ffi._assigned_source
+    c_file = os.path.join(os.path.dirname(os.path.dirname(__file__)), "rust", module_name + source_extension)
+    recompiler.make_c_source(ffi, module_name, source, c_file)
+
+    libraries = _get_openssl_libraries(sys.platform)
+    for lib in libraries:
+        print(f"cargo:rustc-link-lib={lib}")
```
alex, line 126: I think rust-openssl will also emit `rustc-link-lib` lines, which will (hopefully!) be the same as these. Is this required?

src/rust/.cargo/config.toml
```diff
@@ -0,0 +1,11 @@
+[target.x86_64-apple-darwin]
```
alex, line 1: Why is this required? So we can build without going through `setuptools-rust`?

src/rust/build.rs
```diff
@@ -0,0 +1,27 @@
...
+        .arg("../_cffi_src/build_openssl.py")
+        .output()
+        .expect("failed to execute build_openssl.py");
+    let stdout = String::from_utf8(output.stdout).unwrap();
+    let mut include = String::new();
+    for line in stdout.lines() {
+        if line.starts_with("cargo:") {
+            println!("{}", line);
+        } else if line.starts_with("include:") {
+            include = line.replace("include:", "");
```
messense, line 18: I'm thinking the `PYO3_PYTHON` above and this `include` from `sysconfig` should be obtained from `pyo3-build-config` crate.

src/rust/build.rs
```diff
@@ -0,0 +1,27 @@
...
+        }
+    }
+    cc::Build::new()
+        .file("_openssl.c")
+        .include(include)
+        // FIXME: should not hardcode this
+        .include("/opt/homebrew/opt/openssl/include")
```
alex, line 25: It looks like `openssl-sys` exposes some details via `DEP_` vars (https://docs.rs/openssl/latest/openssl/#feature-detection). Should we request a var from them that says "where did you find OpenSSL's headers"?

src/_cffi_src/build_openssl.py
```diff
@@ -112,3 +112,21 @@ def _extra_compile_args(platform):
...
     extra_compile_args=_extra_compile_args(sys.platform),
 )
+
+if __name__ == "__main__":
+    import sysconfig
+    from cffi import recompiler
+
+    module_name, source, source_extension, kwds = ffi._assigned_source
+    c_file = os.path.join(os.path.dirname(os.path.dirname(__file__)), "rust", module_name + source_extension)
+    recompiler.make_c_source(ffi, module_name, source, c_file)
```
messense, line 122: This generates the `_openssl.c` file for later consumption in Rust side.

src/rust/src/lib.rs
```diff
@@ -91,6 +96,11 @@ fn _rust(py: pyo3::Python<'_>, m: &pyo3::types::PyModule) -> pyo3::PyResult<()>
     crate::x509::ocsp_resp::add_to_module(ocsp_mod)?;
     m.add_submodule(ocsp_mod)?;
 
+    let openssl_mod = unsafe {
+        let ptr = PyInit__openssl();
+        pyo3::types::PyModule::from_owned_ptr(py, ptr)
+    };
+    m.add_submodule(openssl_mod)?;
```
alex, line 103: Instead of making `openssl` a sub-module of `_rust`, can we instead make the heirarchy the same as it is now. That is:

- Change `src/rust/` to be `cryptography.hazmat.bindings`
- `m.add_submodule(openssl_mod)`
- `m.add_submodule(rust_mod)`, where `rust_mod` is a new module which has all the current sub-modules as sub-modules of _it_.

src/rust/build.rs
alex: This sounds good to me!

src/rust/.cargo/config.toml
messense: It's not required. It  just makes it easier to test the build script locally with `cargo build`. I accidentally committed it. Will remove it later.

src/rust/build.rs
messense: That'd be great.

alex: Issue filed: https://github.com/sfackler/rust-openssl/issues/1579 it turns out it already exists!

messense: Yeah, found a doc in https://doc.rust-lang.org/cargo/reference/build-script-examples.html#using-another-sys-crate

messense:
> running: "cc" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "-Werror=implicit-function-declaration" "-I/home/runner/work/cryptography/cryptography/osslcache/openssl-1.1.0l-1c9f97a0fce5ec7feed4e4c4944a9d9cca01bcfc/include" "-I" "" "-I" "/usr/include" "-o" "/home/runner/work/cryptography/cryptography/src/rust/target/release/build/cryptography-rust-c7d21142bca98d61/out/_openssl.o" "-c" "/home/runner/work/cryptography/cryptography/src/rust/target/release/build/cryptography-rust-c7d21142bca98d61/out/_openssl.c"
>    cargo:warning=cc: error: /home/runner/work/cryptography/cryptography/src/rust/target/release/build/cryptography-rust-c7d21142bca98d61/out/_openssl.c: No such file or directory
>    cargo:warning=cc: fatal error: no input files

I'm having trouble understand why this fails with tox. Does tox have a cli option to ask it not to delete everything after failure?

alex:
Hmm, I don't think tox should be deleting anything, all the files should still be in `.tox/<env name>/`. Is it possibly some other file getting deleted?

messense:
Here is the same error on my local machine, looks like tox put files a in temp dir.

>    running: "cc" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-arch" "arm64" "-I" "" "-I" "/opt/homebrew/opt/openssl@1.1/include" "-Wall" "-Wextra" "-o" "/private/var/folders/0r/4fqyt_zs2tj4zvglsb7f4zjw0000gn/T/pip-req-build-9o4tjl27/src/rust/target/release/build/cryptography-rust-c81c254f7cb1dd51/out/_openssl.o" "-c" "/private/var/folders/0r/4fqyt_zs2tj4zvglsb7f4zjw0000gn/T/pip-req-build-9o4tjl27/src/rust/target/release/build/cryptography-rust-c81c254f7cb1dd51/out/_openssl.c"
>    cargo:warning=clang: error: no such file or directory: '/private/var/folders/0r/4fqyt_zs2tj4zvglsb7f4zjw0000gn/T/pip-req-build-9o4tjl27/src/rust/target/release/build/cryptography-rust-c81c254f7cb1dd51/out/_openssl.c'
>    cargo:warning=clang: error: no input files

alex:
Oh no :-/ this isn't tox itself, it's pip's `pep517` support. I think you want `--no-clean` on `pip`, but I'm not sure how to pass it down :-/

messense:
I also tried `python -m build` and it works for me. Not sure why tox doesn't like it.

alex:
I'm very confused by what's happening with `tox`, if I put `1 / 0` in the code it still doesn't seem to fail.

alex:
Ok, this might be because we need `    println!("cargo:rerun-if-changed=../_cffi_src/build_openssl.py");`. Still poking though.

alex:
Ok, this fixes it, it's because of `PYTHONPATH`:

```diff
diff --git a/src/rust/build.rs b/src/rust/build.rs
index 60d28131..0dfc2534 100644
--- a/src/rust/build.rs
+++ b/src/rust/build.rs
@@ -6,12 +6,24 @@ fn main() {
     let out_dir = env::var("OUT_DIR").unwrap();
     // FIXME: maybe pyo3-build-config should provide a way to do this?
     let python = env::var("PYO3_PYTHON").unwrap_or("python3".to_string());
+    println!("cargo:rerun-if-changed=../_cffi_src/");
+    let python_path = match env::var("PYTHONPATH") {
+        Ok(mut val) => {
+            val.push_str(":../");
+            val
+        },
+        Err(_) => {
+            "../".to_string()
+        }
+    };
     let output = Command::new(&python)
-        .env("PYTHONPATH", "../")
+        .env("PYTHONPATH", python_path)
         .env("OUT_DIR", &out_dir)
         .arg("../_cffi_src/build_openssl.py")
         .output()
         .expect("failed to execute build_openssl.py");
+    assert!(output.status.success());
+
     let stdout = String::from_utf8(output.stdout).unwrap();
     let mut include = String::new();
     for line in stdout.lines() {
```

alex:
Right now we specify our OpenSSL with CFLAGS/LDFLAGS, I guess we'll need to update to use `openssl-sys`'s env vars. See `.github/workflows/ci.yml`.

messense:
Some of the test matrix passed, hooray!

Debugging unfamiliar CI failure is time consuming, so please feel free to push to my branch to experiment.

src/rust/src/lib.rs
messense: I can do this after building and testing on CI works. But I'm not sure how to deal with these Python files in  https://github.com/pyca/cryptography/tree/main/src/cryptography/hazmat/bindings/openssl

alex:
Sent https://github.com/pyca/infra/pull/380 to install pkg-config
which should help on the linux distros.

On Sun, Dec 12, 2021 at 9:17 PM messense ***@***.***> wrote:
>
> @messense commented on this pull request.
>
> ________________________________
>
> In src/rust/src/lib.rs:
>
> > @@ -91,6 +96,11 @@ fn _rust(py: pyo3::Python<'_>, m: &pyo3::types::PyModule) -> pyo3::PyResult<()>
>      crate::x509::ocsp_resp::add_to_module(ocsp_mod)?;
>      m.add_submodule(ocsp_mod)?;
>
> +    let openssl_mod = unsafe {
> +        let ptr = PyInit__openssl();
> +        pyo3::types::PyModule::from_owned_ptr(py, ptr)
> +    };
> +    m.add_submodule(openssl_mod)?;
>
> I can do this after building and testing on CI works. But I'm not sure how to deal with these Python files in https://github.com/pyca/cryptography/tree/main/src/cryptography/hazmat/bindings/openssl
>
> —
> You are receiving this because you commented.
> Triage notifications on the go with GitHub Mobile for iOS or Android.



-- 

alex:
Ok, next build should reflect that.

On Sun, Dec 12, 2021 at 9:21 PM Alex Gaynor ***@***.***> wrote:
>
> Sent https://github.com/pyca/infra/pull/380 to install pkg-config
> which should help on the linux distros.
>
> On Sun, Dec 12, 2021 at 9:17 PM messense ***@***.***> wrote:
> >
> > @messense commented on this pull request.
> >
> > ________________________________
> >
> > In src/rust/src/lib.rs:
> >
> > > @@ -91,6 +96,11 @@ fn _rust(py: pyo3::Python<'_>, m: &pyo3::types::PyModule) -> pyo3::PyResult<()>
> >      crate::x509::ocsp_resp::add_to_module(ocsp_mod)?;
> >      m.add_submodule(ocsp_mod)?;
> >
> > +    let openssl_mod = unsafe {
> > +        let ptr = PyInit__openssl();
> > +        pyo3::types::PyModule::from_owned_ptr(py, ptr)
> > +    };
> > +    m.add_submodule(openssl_mod)?;
> >
> > I can do this after building and testing on CI works. But I'm not sure how to deal with these Python files in https://github.com/pyca/cryptography/tree/main/src/cryptography/hazmat/bindings/openssl
> >
> > —
> > You are receiving this because you commented.
> > Reply to this email directly, view it on GitHub, or unsubscribe.
> > Triage notifications on the go with GitHub Mobile for iOS or Android.
>
>
>
> --
> All that is necessary for evil to succeed is for good people to do nothing.



-- 

messense:
> thread 'main' panicked at 'OpenSSL library directory does not exist: /home/runner/work/cryptography/cryptography/osslcache/openssl-3.0.0-1c9f97a0fce5ec7feed4e4c4944a9d9cca01bcfc/lib', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-sys-0.9.72/build/main.rs:68:9

Where can I download this `openssl-3.0.0` cached build?

alex:
We don't put them anywhere, they're just built ephemerally in CI.

messense:
Looks like openssl-sys is expecting libs in `lib/` not `lib64` when using the `OPENSSL_DIR` env var.
```
/home/runner/work/cryptography/cryptography/osslcache/openssl-3.0.0-1c9f97a0fce5ec7feed4e4c4944a9d9cca01bcfc
├── bin
│   ├── c_rehash
│   └── openssl
├── include
│   └── openssl
├── lib64
│   ├── engines-3
│   ├── libcrypto.a
│   ├── libcrypto.so -> libcrypto.so.3
│   ├── libcrypto.so.3
│   ├── libssl.a
│   ├── libssl.so -> libssl.so.3
│   ├── libssl.so.3
│   ├── ossl-modules
│   └── pkgconfig
└── ssl
    ├── certs
    ├── ct_log_list.cnf
    ├── ct_log_list.cnf.dist
    ├── misc
    ├── openssl.cnf
    ├── openssl.cnf.dist
    └── private
```

src/rust/build.rs
```diff
@@ -0,0 +1,85 @@
...
+use std::io::Write;
+use std::path::Path;
+use std::process::{Command, Stdio};
+
+fn main() {
+    let out_dir = env::var("OUT_DIR").unwrap();
+    // FIXME: maybe pyo3-build-config should provide a way to do this?
+    let python = env::var("PYO3_PYTHON").unwrap_or_else(|_| "python3".to_string());
+    println!("cargo:rerun-if-changed=../_cffi_src/");
+    let python_path = match env::var("PYTHONPATH") {
```
messense, line 11: This isn't working on Windows

```
      File "../_cffi_src/build_openssl.py", line 12, in <module>
        from _cffi_src.utils import build_ffi_for_binding, compiler_type
    ModuleNotFoundError: No module named '_cffi_src'
```

alex:
How fun, putting it in `lib64` is new in OpenSSL 3.0.

src/rust/build.rs
alex: :-/ I wonder why.

reaperhulk:
Yes OpenSSL switched to using `lib64` by default in 3.0.0-beta2 (https://github.com/pyca/cryptography/pull/6192).

src/rust/build.rs
messense: Windows use `;` instead of `:`.

messense:
Opened https://github.com/sfackler/rust-openssl/issues/1581

messense:
There are two errors I don't quite understand:

1. `undefined symbol: PyInit__openssl` on [PyPy](https://github.com/pyca/cryptography/runs/4502687798?check_suite_focus=true)
2. `Symbol not found: ___isPlatformVersionAtLeast` on [macOS Python 3.10](https://github.com/pyca/cryptography/runs/4502688656?check_suite_focus=true)

messense:
> Symbol not found: ___isPlatformVersionAtLeast

Could this be that the pre-compiled OpenSSL is built on macOS 11 but we're using it on macOS 10.15?

messense:
> undefined symbol: PyInit__openssl on PyPy

This is interesting.

```bash
root@5b52dbd4d5c2:/io# readelf -s /io/src/cryptography/hazmat/bindings/_rust.pypy38-pp73-aarch64-linux-gnu.so | grep PyInit
    11: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND PyInit__openssl
   148: 0000000000115290    32 FUNC    GLOBAL DEFAULT   11 PyInit__rust
 52405: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND PyInit__openssl
 52520: 0000000000115290    32 FUNC    GLOBAL DEFAULT   11 PyInit__rust
```

V.S. the `_openssl.*.so` on main branch for PyPy

```bash
root@5b52dbd4d5c2:/io# readelf -s src/cryptography/hazmat/bindings/_openssl.pypy38-pp73-aarch64-linux-gnu.so | grep PyInit
root@5b52dbd4d5c2:/io# readelf -s src/cryptography/hazmat/bindings/_openssl.pypy38-pp73-aarch64-linux-gnu.so | grep _cffi_pypyinit
   709: 0000000000025fa0    32 FUNC    GLOBAL DEFAULT   11 _cffi_pypyinit__openssl
  1407: 0000000000025fa0    32 FUNC    GLOBAL DEFAULT   11 _cffi_pypyinit__openssl
```

The only useful information I can find is https://github.com/pypa/auditwheel/issues/93

Looking at the generated `_openssl.c`

```c
#ifdef PYPY_VERSION
PyMODINIT_FUNC
_cffi_pypyinit__openssl(const void *p[])
{
    p[0] = (const void *)0x2601;
    p[1] = &_cffi_type_context;
#if PY_MAJOR_VERSION >= 3
    return NULL;
#endif
}
#  ifdef _MSC_VER
     PyMODINIT_FUNC
#  if PY_MAJOR_VERSION >= 3
     PyInit__openssl(void) { return NULL; }
#  else
     init_openssl(void) { }
#  endif
#  endif
#elif PY_MAJOR_VERSION >= 3
PyMODINIT_FUNC
PyInit__openssl(void)
{
  return _cffi_init("_openssl", 0x2601, &_cffi_type_context);
}
#else
PyMODINIT_FUNC
init_openssl(void)
{
  _cffi_init("_openssl", 0x2601, &_cffi_type_context);
}
#endif
```

It seems that we need to define `PYPY_VERSION` for it when building on PyPy. `PyInit__openssl(void) { return NULL; }` is also interesting.

alex:
My read of the PyPy source code is that they have a completely different signature and there's not really a C-API that gives you a `PyObject *` for a module: https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/pypy/module/_cffi_backend/cffi1_module.py#L20

I think we may need to ask for a new feature from the cffi/PyPy team. I'll send a mail to their team asking for it.

messense:
I'll try define `PY_MAJOR_VERSION` for it to force it use cpyext API. Let's see if it works.

src/_cffi_src/build_openssl.py
```diff
@@ -112,3 +112,15 @@ def _extra_compile_args(platform):
...
+    from cffi import recompiler
+
+    out_dir = os.getenv("OUT_DIR")
+    module_name, source, source_extension, kwds = ffi._assigned_source
+    c_file = os.path.join(out_dir, module_name + source_extension)
+    recompiler.make_c_source(ffi, module_name, source, c_file)
+
+    extra_compile_args = _extra_compile_args(sys.platform)
+    for arg in extra_compile_args:
+        print(f"cargo:rustc-link-arg={arg}")
```
messense, line 126: Is it OK to move `-Wconversion` and `-Wno-error=sign-conversion` to Rust side using only [flag_if_supported](https://docs.rs/cc/latest/cc/struct.Build.html#method.flag_if_supported)?

alex: Seems fine to me.

messense:
Force `#undef PYPY_VERSION` doesn't work either. Doing it will call `_cffi_init` on extension import which try to call [`_cffi_backend._init_cffi_1_0_external_module`](https://foss.heptapod.net/pypy/cffi/-/blob/9b931c3e2e03b0b471822fb38460c1dc66928a5c/c/_cffi_backend.c#L7729) method, but this method seems only available on the PyPI `cffi` package not on PyPy's builtin `_cffi_backend` module.

alex:
For macOS, I believe this is the solution: https://github.com/alexcrichton/curl-rust/pull/283 (see the assosciated issue for more details, it has the exact same symptoms)

alex:
PR here for lib64: https://github.com/sfackler/rust-openssl/pull/1582

src/rust/src/lib.rs
messense:
```diff
diff --git a/setup.cfg b/setup.cfg
index 23a6a55b..45ad765f 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,7 +43,7 @@ zip_safe = False
 package_dir =
     =src
 packages = find:
-ext_package = cryptography.hazmat.bindings
+ext_package = cryptography.hazmat
 # `install_requires` must be kept in sync with `pyproject.toml`
 install_requires =
     cffi >=1.12
diff --git a/setup.py b/setup.py
index 8f781a6a..8a6e6433 100644
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ try:
     setup(
         rust_extensions=[
             RustExtension(
-                "_rust",
+                "bindings",
                 "src/rust/Cargo.toml",
                 py_limited_api=True,
                 # Enable abi3 mode if we're not using PyPy.
diff --git a/src/cryptography/hazmat/bindings/openssl/binding.py b/src/cryptography/hazmat/bindings/openssl/binding.py
index 5ad08189..21daadff 100644
--- a/src/cryptography/hazmat/bindings/openssl/binding.py
+++ b/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -12,11 +12,11 @@ import warnings
 import cryptography
 from cryptography import utils
 from cryptography.exceptions import InternalError
-from cryptography.hazmat.bindings._rust import _openssl
+from cryptography.hazmat.bindings._openssl import ffi, lib
 from cryptography.hazmat.bindings.openssl._conditional import CONDITIONAL_NAMES

-ffi = _openssl.ffi
-lib = _openssl.lib
+# ffi = _openssl.ffi
+# lib = _openssl.lib
 _OpenSSLErrorWithText = collections.namedtuple(
     "_OpenSSLErrorWithText", ["code", "lib", "reason", "reason_text"]
 )
diff --git a/src/rust/src/lib.rs b/src/rust/src/lib.rs
index 0b7c98f1..c91aaa7a 100644
--- a/src/rust/src/lib.rs
+++ b/src/rust/src/lib.rs
@@ -100,11 +100,18 @@ fn _rust(py: pyo3::Python<'_>, m: &pyo3::types::PyModule) -> pyo3::PyResult<()>
     crate::x509::ocsp_resp::add_to_module(ocsp_mod)?;
     m.add_submodule(ocsp_mod)?;

+    Ok(())
+}
+
+#[pyo3::prelude::pymodule]
+fn bindings(py: pyo3::Python<'_>, m: &pyo3::types::PyModule) -> pyo3::PyResult<()> {
     let openssl_mod = unsafe {
         let ptr = PyInit__openssl();
         pyo3::types::PyModule::from_owned_ptr(py, ptr)
     };
     m.add_submodule(openssl_mod)?;
+
+    m.add_wrapped(pyo3::wrap_pymodule!(_rust))?;
     Ok(())
 }

diff --git a/tests/hazmat/backends/test_openssl_memleak.py b/tests/hazmat/backends/test_openssl_memleak.py
index d1d672a4..3bae0243 100644
--- a/tests/hazmat/backends/test_openssl_memleak.py
+++ b/tests/hazmat/backends/test_openssl_memleak.py
@@ -24,10 +24,10 @@ def main(argv):

     import cffi

-    from cryptography.hazmat.bindings._rust import _openssl
+    from cryptography.hazmat.bindings._openssl import ffi, lib

-    ffi = _openssl.ffi
-    lib = _openssl.lib
+    # ffi = _openssl.ffi
+    # lib = _openssl.lib

     heap = {}
```

This does not work 

```python
ImportError while loading conftest '/Users/messense/Projects/cryptography/tests/conftest.py'.
tests/conftest.py:8: in <module>
    from cryptography.hazmat.backends.openssl import backend as openssl_backend
src/cryptography/hazmat/backends/openssl/__init__.py:6: in <module>
src/cryptography/hazmat/backends/openssl/backend.py:13: in <module>
    from cryptography import utils, x509
src/cryptography/x509/__init__.py:6: in <module>
    from cryptography.x509 import certificate_transparency
src/cryptography/x509/certificate_transparency.py:10: in <module>
    from cryptography.hazmat.bindings._rust import x509 as rust_x509
E   ImportError: cannot import name 'x509' from 'cryptography.hazmat.bindings._rust' (unknown location)
```

messense: @alex what do you think?

messense:
> My read of the PyPy source code is that they have a completely different signature and there's not really a C-API that gives you a `PyObject *` for a module: https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/pypy/module/_cffi_backend/cffi1_module.py#L20
> 
> I think we may need to ask for a new feature from the cffi/PyPy team. I'll send a mail to their team asking for it.

Let me know if there is any progress on this, thanks.

src/rust/src/lib.rs
alex: I think let's not worry about including it in this PR, it'd be cool to have the module names stay teh same, but not worth the headache!

messense:
> flake run-test: commands[3] | mypy src/cryptography/ vectors/cryptography_vectors/ tests/
src/cryptography/hazmat/bindings/openssl/binding.py:15: error: Module "cryptography.hazmat.bindings._rust" has no attribute "_openssl"  [attr-defined]
Found 1 error in 1 file (checked 186 source files)

Not sure how to fix this mypy error.

alex:
Probably the `[[tool.mypy.overrides]]` section in `pyproject.toml`.

alex:
Discussion about PyPy is here: https://foss.heptapod.net/pypy/pypy/-/issues/3619

alex:
PR to `openssl-sys` for `lib64/` has landed. It's not in a release yet, but it should be possible to point `Cargo.toml` at git to see if it works.

alex:
GHA is pretty busted ATM, unfortunately. Tons of networking errors + timeouts.

src/rust/build.rs
```diff
@@ -0,0 +1,139 @@
...
+        .file(openssl_c)
+        .include(python_include)
+        .include(openssl_include)
+        .flag_if_supported("-Wconversion")
+        .flag_if_supported("-Wno-error=sign-conversion");
+
+    // Enable abi3 mode if we're not using PyPy.
+    if python_impl != "PyPy" {
+        // cp36
+        // build.define("Py_LIMITED_API", "0x030600f0");
```
messense, line 77: Unable to get cffi embedding to compile with Py_LIMITED_API.

src/rust/src/lib.rs
```diff
@@ -91,6 +99,13 @@ fn _rust(py: pyo3::Python<'_>, m: &pyo3::types::PyModule) -> pyo3::PyResult<()>
     crate::x509::ocsp_resp::add_to_module(ocsp_mod)?;
     m.add_submodule(ocsp_mod)?;
 
+    let openssl_mod = unsafe {
+        // let ptr = PyInit__openssl();
+        // pyo3::types::PyModule::from_owned_ptr(py, ptr)
+        make_cryptography_openssl_module();
+        pyo3::types::PyModule::import(py, "_openssl")?
```
messense, line 106: This also makes `import _openssl` work after importing cryptography, could this be a problem for end users?

messense:
It's not building on Windows, I'm not sure why and I don't have a Windows machine, help appreciated!

messense:
Unfortunately looks like PyPy tests are hanging with the cffi embedding approach.

alex:
Looks like that issue is already closed, I'd suggest either opening a new one or including a comment on the cffi thread we have open. (Sorry I was slow to look at this!)

src/_cffi_src/build_openssl.py
```diff
@@ -112,3 +112,15 @@ def _extra_compile_args(platform):
...
     extra_compile_args=_extra_compile_args(sys.platform),
 )
+
+if __name__ == "__main__":
+    out_dir = os.getenv("OUT_DIR")
+    module_name, source, source_extension, kwds = ffi._assigned_source
+    c_file = os.path.join(out_dir, module_name + source_extension)
+    ffi.embedding_api(
+        """
+        extern "Python" void _this_is_not_used(void);
```
alex, line 122: Can we add a comment explaining what this is for?

src/_cffi_src/utils.py
```diff
@@ -73,6 +73,12 @@ def build_ffi(
     verify_source += '\n#define CRYPTOGRAPHY_PACKAGE_VERSION "{}"'.format(
         about["__version__"]
     )
+    verify_source += r"""
+
+int make_cryptography_openssl_module(void) {
```
alex, line 78: Let's name this `Cryptography_make_openssl_module()`, which is the prefix we generally use.

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -11,9 +11,11 @@
 import cryptography
 from cryptography import utils
 from cryptography.exceptions import InternalError
-from cryptography.hazmat.bindings._openssl import ffi, lib
+from cryptography.hazmat.bindings._rust import _openssl
 from cryptography.hazmat.bindings.openssl._conditional import CONDITIONAL_NAMES
 
+ffi = _openssl.ffi  # type: ignore
+lib = _openssl.lib  # type: ignore
```
alex, line 18: What actually uses these? Theoretically everything should go through a `Binding` instance.

src/rust/build.rs
alex: Oof :-/ Maybe we need to do the embedding way on PyPy and the "normal" way on CPython? PyPy already doesn't support `Py_LIMITED_API`.

src/rust/src/lib.rs
alex: Probably not. Maybe we should name it `_cryptography_openssl` or something to ensure no one accidentally uses it?

messense:
Sorry I'm busy recently and unable to make progress on this. I'll come back to this maybe after Chinese lunar new year. In the meantime, feel free to take over.

alex:
Thank you for all your work on this!

On Thu, Jan 20, 2022 at 9:11 AM messense ***@***.***> wrote:
>
> Sorry I'm busy recently and unable to make progress on this. I'll come back to this maybe after Chinese lunar new year. In the meantime, feel free to take over.
>
> —
> Triage notifications on the go with GitHub Mobile for iOS or Android.



-- 

alex:
I'm attempting to take over this PR in #7164. Thanks for all your work on this!

Still stuck on the PyPy hangs!

PR #6699, opened 2021-12-13
Added init defaults to classes in x509.extensions

andrewjroth:
I added several defaults to class initialization methods for parameters which were already considered to be optional, but were created as positional (required) arguments.  With this change, users creating instances of these classes do not have to add several `None` or `False` arguments for the sake of satisfying the initialization, which only makes the code more confusing.  Instead, by changing them to keyword arguments, with defaults, code can be more succinct. 

Example Before:
```python
from cryptography import x509
x509.KeyUsage(True, False, False, False, False, True, True, False, False)  # Unclear what these mean
```

Example After:
```python
from cryptography import x509
x509.KeyUsage(digital_signature=True, key_cert_sign=True, crl_sign=True)  # Clear what usage is being configured
```

Because python allows keyword arguments to be set using positional arguments, this should be a non-breaking change.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,7 @@ Changelog
   and ``x86_64`` architectures. Users on macOS should upgrade to the latest
   ``pip`` to ensure they can use this wheel, although we will continue to
   ship ``x86_64`` specific wheels for now to ease the transition.
+* Added defaults to several class __init__ to make code more succinct.
 
 .. _v36-0-0:
 
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -183,9 +183,9 @@ class AuthorityKeyIdentifier(ExtensionType):
 
     def __init__(
         self,
-        key_identifier: typing.Optional[bytes],
-        authority_cert_issuer: typing.Optional[typing.Iterable[GeneralName]],
-        authority_cert_serial_number: typing.Optional[int],
+        key_identifier: typing.Optional[bytes] = None,
+        authority_cert_issuer: typing.Optional[typing.Iterable[GeneralName]] = None,
+        authority_cert_serial_number: typing.Optional[int] = None,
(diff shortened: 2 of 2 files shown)


alex:
Hi,

Not having defaults here is intentional, we want to require folks to engage with all the realities of X.509, rather than accidentally end up with things they didn't intend.

The solution to the readability problem is to always use keyword arguments for all of the parameters -- and if we have any examples where we don't use them, please point them out so we can fix that. If we were designing the APIs today, we'd make them keyword-only arguments, however these APIs were originally designed when python2 was still a thing, alas.

Thanks for submitting this PR though!

andrewjroth:
Hello,

Thank you for responding so quickly and considering this change.  I'm not sure I understand or agree with the reasoning.  From my perspective, having all parameters as positional only makes things more confusing without really understanding X.509 at all.  I still believe this would be an improvement and hope it could be considered in the future.  However, it's just a "convenience" improvement and ultimately doesn't impact the usability of the package.

Thank you!

reaperhulk:
One of the tenets of cryptography is that we want to make auditing of a codebase as unambiguous as possible. This generally means avoiding defaults, which a security auditor has to look up (and also confirm that they haven't changed in various versions). Defaulting to False on all KeyUsages would generally be safe (as long as we never changed our mind, which we could say we wouldn't but an auditor would always have to confirm!), but in my opinion the verbosity of this is a fair tradeoff for keeping that consistency across our APIs.

It would be nice to have a means by which we could long-term transition these to kwargs only though. Maybe a PR that starts a long term deprecation (probably 5+ years) on positional args is a path to force readability improvements? Not sure.

Even though we're not going to accept this contribution, I do want to say thank you for engaging us in conversation like this. Ideas on improvement from the community are important and we'd love to see you contribute again in the future. If you think it's possible something will be controversial feel free to file an issue with a sketch of the proposed solution so we can discuss it!

PR #6710, opened 2021-12-16
Increased minimum macOS version to 10.11

alex:
Both Firefox and Chrome have dropped support for 10.10, and apple hasn't supported it in 4 years. Dropbox however, does still seem to support it.


alex:
@maxbelanger are y'all still supporting 10.10? Do you have a timeline to drop?

maxbelanger:
@alex We do still support 10.10 and I don't know of a timeline for deprecating support. I wouldn't object to this change though, we can make do. I'll make sure we consider this in our planning :).

alex:
Ok, I'm going to wontfix this for now then. Thanks.

Issue #5669, opened 2020-12-24
Python 3.7 & 3.8 wheels missing from pypi


bryan-hunt:
It looks like some build system changes and/or failures must have occurred with the 3.3 release (see [3.3](https://pypi.org/project/cryptography/3.3/#files) and [3.3.1](https://pypi.org/project/cryptography/3.3.1/#files) ) as the last version with python 3.7 & 3.8 wheels is version [3.2.1](https://pypi.org/project/cryptography/3.2.1/#files).

I looked around for an issue already describing this but I didn't see one - so obviously close if this is a duplicate

alex:
https://cryptography.io/en/latest/faq.html#why-are-there-no-wheels-for-my-python3-x-version

PR #6715, opened 2021-12-21
Docs: clarify min pip version which understands abi3

sparrowt:
Be more specific than just "recent versions of pip" regarding which versions support selecting 'cp3X-abi3' wheels on python 3.Y where Y > X.

Aside from 'recent' not aging well, this should also help people who may have pip itself pinned & need to know how far they have to upgrade to benefit from this.

Having found #5669 when searching for the issue at hand, I ended up on the linked https://cryptography.io/en/latest/faq/#why-are-there-no-wheels-for-my-python3-x-version and had to figure out by trial and error which pip version is sufficient to get this behaviour.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -155,7 +155,7 @@ Why are there no wheels for my Python3.x version?
 Our Python3 wheels are ``abi3`` wheels. This means they support multiple
 versions of Python. The ``abi3`` wheel can be used with any version of Python
 greater than or equal to the version it specifies. Recent versions of ``pip``
-will automatically install ``abi3`` wheels.
+(20.0.1+) will automatically install ``abi3`` wheels.
 
 Why can't I import my PEM file?
 -------------------------------


docs/faq.rst
```diff
@@ -155,7 +155,7 @@ Why are there no wheels for my Python3.x version?
 Our Python3 wheels are ``abi3`` wheels. This means they support multiple
 versions of Python. The ``abi3`` wheel can be used with any version of Python
 greater than or equal to the version it specifies. Recent versions of ``pip``
-will automatically install ``abi3`` wheels.
+(20.0.1+) will automatically install ``abi3`` wheels.
```
alex, line 158: Why 20.0.1? There's nothing in the changelog that obviously indicates that's when abi3 was added https://pip.pypa.io/en/stable/news/#v20-0-1

reaperhulk: This is around when pip added abi3 for windows, but other platforms had support far earlier. Trying to specify an exact version here won't be simple because the required pip version is platform and cryptography version dependent. This is generally why all of our installation error messages suggest upgrading pip to latest.

sparrowt: Indeed I also looked at the changelog in vain - so sadly had to resort to binary search. I discovered that `19.3.1` doesn't handle abi3 correctly and `20.0.1` does. Per https://pypi.org/project/pip/#history `20.0` was yanked and there are no other released versions of pip inbetween.
image

It must have been fixed somewhere in the 16kloc diff between https://github.com/pypa/pip/compare/19.3.1...20.0.1 but I didn't have the time or motivation to pin it down further I'm afraid (!)

reaperhulk: 20.0.1 probably is correct, but that's strictly for Windows. Since this is different for every platform and it is our general advice that users upgrade to the latest `pip` I'm not sure what better language is available than `recent` for this.

sparrowt: Fair enough, I didn't realise pip gained this support for different platforms at different times so I figured stating an exact version would be non-controversial and might help some.

"recent" will of course get less true the longer it has been since 20.0.1 but I can't think of a better non-specific word without listing a version number so if you'd rather not do so that's your call of course :)

PR #6725, opened 2021-12-22
type some of the private interfaces of poly1305

reaperhulk:


src/cryptography/hazmat/primitives/poly1305.py
```diff
@@ -2,16 +2,20 @@
...
 
+import typing
 
 from cryptography import utils
 from cryptography.exceptions import (
     AlreadyFinalized,
     UnsupportedAlgorithm,
     _Reasons,
 )
+from cryptography.hazmat.backends.openssl.poly1305 import _Poly1305Context
```
reaperhulk, line 13: We could potentially remove this import if we changed our model of finalization to not check whether `_ctx` is `None` or not.

alex (approved):

PR #6726, opened 2021-12-22
Improve typing of x509.name module

alex:


src/cryptography/x509/name.py
```diff
@@ -29,7 +29,7 @@ class _ASN1Type(utils.Enum):
 
 _ASN1_TYPE_TO_ENUM = {i.value: i for i in _ASN1Type}
 _SENTINEL = object()
```
reaperhulk, line 31: You should remove this if we're not going to explicitly use the sentinel any more

alex: done

reaperhulk (dismissed):

reaperhulk (approved):

PR #6734, opened 2021-12-22
improvements to public/private key types and misc backend typing

reaperhulk:


src/cryptography/x509/base.py
```diff
@@ -21,9 +21,9 @@
     x448,
 )
 from cryptography.hazmat.primitives.asymmetric.types import (
+    CERTIFICATE_ISSUER_PUBLIC_KEY_TYPES,
+    CERTIFICATE_PRIVATE_KEY_TYPES,
```
reaperhulk, line 25: This doesn't re-export so may break people. Do we want to re-export these symbols with the old names?

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2151,7 +2206,11 @@ def serialize_key_and_certificates_to_pkcs12(
         with self._zeroed_null_terminated_buf(password) as password_buf:
             with self._zeroed_null_terminated_buf(name) as name_buf:
                 ossl_cert = self._cert2ossl(cert) if cert else self._ffi.NULL
-                evp_pkey = key._evp_pkey if key else self._ffi.NULL
+                if key is not None:
+                    evp_pkey = key._evp_pkey  # type: ignore[union-attr]
+                else:
+                    evp_pkey = self._ffi.NULL
```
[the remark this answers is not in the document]
reaperhulk: Only the error suggested any of the three was acceptable. I've pushed an update that has a more specific error msg for one case, although the logic is probably more convoluted than it needs to be since I've been doing this too much today 😁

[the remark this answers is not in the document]
reaperhulk: Okay, mitmproxy relies on the existing behavior (where passing `cert` without `key` makes it show up in `cas`), so I'll just add tests for this behavior, document it a bit better, and move on.

tests/hazmat/primitives/test_pkcs12.py
```diff
@@ -363,10 +363,53 @@ def test_generate_no_cert(self, backend):
             p12, None, backend
         )
         assert parsed_cert is None
-        assert parsed_key is not None
+        assert isinstance(parsed_key, ec.EllipticCurvePrivateKey)
         assert parsed_key.private_numbers() == key.private_numbers()
         assert parsed_more_certs == []
 
+    @pytest.mark.parametrize(
```
alex, line 370: Please put this in a separate PR.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -952,7 +964,7 @@ def load_der_public_key(self, data):
             else:
                 self._handle_key_loading_error()
 
-    def load_der_parameters(self, data):
+    def load_der_parameters(self, data: bytes) -> _DHParameters:
```
alex, line 967: Why private?

reaperhulk: ugh, I fixed this on pem but missed der.

alex (approved):

PR #6738, opened 2021-12-22
more cipher typing

reaperhulk:
bytes is an alias that includes bytearray and memoryview so we can type
update_into's buf

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/ciphers.py
+++ b/src/cryptography/hazmat/backends/openssl/ciphers.py
@@ -149,7 +149,7 @@ class _CipherContext(object):
         n = self.update_into(data, buf)
         return bytes(buf[:n])
 
-    def update_into(self, data: bytes, buf) -> int:
+    def update_into(self, data: bytes, buf: bytes) -> int:
         total_data_len = len(data)
         if len(buf) < (total_data_len + self._block_size_bytes - 1):
             raise ValueError(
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/backends/openssl/ciphers.py
```diff
@@ -149,7 +149,7 @@ def update(self, data: bytes) -> bytes:
         n = self.update_into(data, buf)
         return bytes(buf[:n])
 
-    def update_into(self, data: bytes, buf) -> int:
+    def update_into(self, data: bytes, buf: bytes) -> int:
```
alex, line 152: `bytes` isn't really correct here, why does mypy even allow it?

reaperhulk: https://docs.python.org/3/library/typing.html#typing.ByteString

```
This type represents the types bytes, bytearray, and memoryview of byte sequences.

As a shorthand for this type, bytes can be used to annotate arguments of any of the types mentioned above.
```

alex: It's not correct to pass any of those types though, only a subset of them!

reaperhulk: Which we enforce at runtime. mypy doesn’t distinguish in a useful manner for us though 😔

alex: Right, but the point of typing is indicate invalid types are compile time. Is there really no way in mypy to express "a writable buffer type"!?

reaperhulk: Sounds like somebody volunteered to determine the answer to that once and for all.

alex (dismissed):

alex (approved):

PR #6739, opened 2021-12-22
verify typing works on py36

reaperhulk:


reaperhulk:
Not sure what to make of this...

alex:
what is that test trying to test even? We don't usually test that "omitting required args raises a type error" since that's handled entirely inside python itself.

PR #6740, opened 2021-12-22
type encryptor/decryptor

reaperhulk:
These methods are dynamic so the type of return you get depends on the
cipher/mode you passed. Users should type narrow to CipherContext,
AEADEncryptionContext, or AEADDecryptionContext as needed.

This commit significantly modifies the way we declare these interfaces
and duplicates the methods between AEADCipherContext and CipherContext.

If we didn't do that then type narrowing like
isinstance(encryptor, CipherContext) would be ambiguous since both
AEADEncryptionContext and AEADDecryptionContext would be descendents of
CipherContext.

Removing the indirection (as in #6642) doesn't provide much performance benefit and this approach allows a reasonable path forward on typing.

### DIFF
--- a/src/cryptography/hazmat/primitives/ciphers/base.py
+++ b/src/cryptography/hazmat/primitives/ciphers/base.py
@@ -6,7 +6,6 @@
 import abc
 import typing
 
-from cryptography import utils
 from cryptography.exceptions import (
     AlreadyFinalized,
     AlreadyUpdated,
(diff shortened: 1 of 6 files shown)


alex (commented):
This looks like it's basically correct, but the ergonomics suck :-)

My last idea in terms of getting this to work inuitive is "something with generics :handwave:". Specifically if `Cipher` were generic over the mode type, could we make the return type of `encryptor`/`decryptor` change based on the type parameter?

PR #6741, opened 2021-12-22
Try including python version in the cache

alex:


alex:
Doesn't fix the segfault, but probably a good idea.

reaperhulk (approved):

PR #6742, opened 2021-12-22
Handle large OIDs cleanly

### LINKED ISSUE #6573: CSR parsing error with >28 bits oids
`cryptography` is capable of producing PEM-encoded CSRs containing >28bits OIDs, but fails when parsing them with the error (here, the OID is in a policy identifier):
```
ValueError: error parsing asn1 value: ParseError { kind: InvalidValue, location: ["0", "PolicyInformation::policy_identifier"] }
```

This makes the UUID OID subtree (2.25) unusable, as it requires 128-bits OIDs.

Now, I guess this could be reasonably closed as wontfix because there are many other implementations out there which do not (and some will likely never) support >64bits OIDs (ex: golang's). This is a messy topic at he very least:
- because there are many implementations incomplete in such a way out there, relying on >64bits OIDs is a bad idea
- ...so fixing it here is unlikely to do much good
- ...except for existing implementations which were not aware of this lack of support, in which case propagating the lack of support (=breaking >64bits oids in `cryptography`) will complicate the upgrade path

For my personal use cases, I was kind of lucky that I discovered the issue before support was dropped in `cryptography`, and I believe all impacted certificates have been long migrated. It is my automated testing of this very migration codepath which detected the breakage in `cryptography`, so I guess I have to at least skip this test when parsing fails.

If this bug was at 63 or 64bits, I would sugest resolving it by just (retroactively ?) listing this breakage in the changelog. But seeing that this happens with OIDs well under 64bits makes me think a fix is needed.

Of course, if such fix can also add support for 128bits OIDs, it is even better, but I do not know if it is necessary.

Versions:
- Python: 3.9.2
- cryptography: 35.0.0 (installed with pip in a virtualenv)
- cffi: 1.15.0
- pip: 21.3.1
- setuptools: 58.5.3

Steps to reproduce:

```python
from cryptography import x509
from cryptography.hazmat.primitives import serialization, hashes

LONG_UUID = 2**128 - 1
LONG_OID = '2.25.%i' % (LONG_UUID, )
SHORT_OID = '2.25.%i' % (LONG_UUID & (2**28 - 1), )
test_private_key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
def makeAndLoadCSR(oid):
    extension = x509.CertificatePolicies([
        x509.PolicyInformation(x509.oid.ObjectIdentifier(oid), None),
    ])
    csr_pem = x509.CertificateSigningRequestBuilder(
        subject_name=x509.Name([x509.NameAttribute(oid=x509.oid.NameOID.COMMON_NAME, value='test')]),
    ).add_extension(extension, critical=False).sign(
        private_key=test_private_key,
        algorithm=hashes.SHA256(),
    ).public_bytes(encoding=serialization.Encoding.PEM)
    _ = x509.load_pem_x509_csr(csr_pem).extensions

# this succeeds
makeAndLoadCSR(SHORT_OID)
# this fails
makeAndLoadCSR(LONG_OID)
```

Output of the last statement above when the issue is present:
```
  File "<stdin>", line 14, in makeAndLoadCSR
ValueError: error parsing asn1 value: ParseError { kind: InvalidValue, location: ["0", "PolicyInformation::policy_identifier"] }
```

alex:
closes #6573

### DIFF
--- a/src/rust/src/x509/csr.rs
+++ b/src/rust/src/x509/csr.rs
@@ -2,7 +2,7 @@
 // 2.0, and the BSD License. See the LICENSE file in the root of this repository
 // for complete details.
 
-use crate::asn1::{PyAsn1Error, PyAsn1Result};
+use crate::asn1::{py_oid_to_oid, PyAsn1Error, PyAsn1Result};
 use crate::x509;
 use crate::x509::{certificate, oid};
 use asn1::SimpleAsn1Readable;
@@ -224,8 +224,7 @@ impl CertificateSigningRequest {
                 cryptography_warning,
             ),
         )?;
-        let oid_str = oid.getattr("dotted_string")?.extract::<&str>()?;
-        let rust_oid = asn1::ObjectIdentifier::from_string(oid_str).unwrap();
+        let rust_oid = py_oid_to_oid(oid)?;
         for attribute in self
             .raw
             .borrow_value()
@@ -255,7 +254,7 @@ impl CertificateSigningRequest {
         Err(pyo3::PyErr::from_instance(
             py.import("cryptography.x509")?.call_method1(
                 "AttributeNotFound",
-                (format!("No {} attribute was found", oid_str), oid),
+                (format!("No {} attribute was found", oid), oid),
             )?,
         ))
     }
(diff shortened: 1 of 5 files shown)


src/rust/src/x509/csr.rs
```diff
@@ -255,7 +254,7 @@ impl CertificateSigningRequest {
         Err(pyo3::PyErr::from_instance(
             py.import("cryptography.x509")?.call_method1(
                 "AttributeNotFound",
-                (format!("No {} attribute was found", oid_str), oid),
+                (format!("No {} attribute was found", oid), oid),
```
reaperhulk, line 257: what does format! do with a PyAny?

alex: `__repr__` I believe

reaperhulk (approved):

PR #6743, opened 2021-12-23
Don't test with coverage on py311 due to segfaults

alex:


reaperhulk (approved):

reaperhulk:
https://bugs.python.org/issue46159 for tracking purposes

pquentin:
The ABI between alpha2 and alpha3 changed, see https://bugs.python.org/issue46320. If pip built (say) cffi in CI for alpha2 and put it in the cache, then it reused that wheel for alpha3 and the segfault is expected. I wonder if there's a way to tell GitHub's pip cache to not cache 3.11 wheels, as "The ABI is generally not stable until the first beta (and sometimes a bit after that)."

Issue #6531, opened 2021-11-03
Simplify Cipher


reaperhulk:
We have an extra layer of indirection in cipher that we probably don't need. This issue can be closed if we either fix that or decide we do want to keep it.

reaperhulk:
I did a bunch of experimentation here 4 months ago. It didn't really help perf and it wasn't less confusing, so I'm going to close this.

PR #6744, opened 2021-12-23
type encryptor/decryptor

### LINKED ISSUE #6083: Lack of stubs in Cipher.encryptor and Cipher.decryptor
the Cipher.encryptor() and Cipher.decryptor() functions provide no static type hints, this limits the amount of static typing that can be applied around the encryptor and decryptor objects

example line missing type hints:
https://github.com/pyca/cryptography/blob/686185003b03c9611fd31ba409d90d8080130f36/src/cryptography/hazmat/primitives/ciphers/base.py#L103

reaperhulk:
This makes Cipher generic over mode and then uses that generic type to determine the return type of encryptor/decryptor. This makes the type signature look weird for the constructor with tools like pylance, but mypy itself enforces the typing properly and pylance correctly shows the resulting encryptor/decryptor return type. All hail abusing generics.

replaces #6740 

fixes #6083 

refs #6531

### DIFF
--- a/src/cryptography/hazmat/primitives/ciphers/base.py
+++ b/src/cryptography/hazmat/primitives/ciphers/base.py
@@ -15,6 +15,12 @@ from cryptography.hazmat.primitives._cipheralgorithm import CipherAlgorithm
 from cryptography.hazmat.primitives.ciphers import modes
 
 
+if typing.TYPE_CHECKING:
+    from cryptography.hazmat.backends.openssl.ciphers import (
+        _CipherContext as _BackendCipherContext,
+    )
+
+
 class CipherContext(metaclass=abc.ABCMeta):
     @abc.abstractmethod
     def update(self, data: bytes) -> bytes:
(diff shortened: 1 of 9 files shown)


src/cryptography/hazmat/primitives/ciphers/base.py
```diff
@@ -64,23 +83,59 @@ def tag(self) -> bytes:
...
+            # mypy needs this assert to narrow the type from our generic
+            # type. Maybe it won't some time in the future.
+            assert isinstance(mode, modes.Mode)
             mode.validate_for_algorithm(algorithm)
 
         self.algorithm = algorithm
         self.mode = mode
 
+    @typing.overload
+    def encryptor(self: "Cipher[modes.ModeWithNonce]") -> CipherContext:
```
alex, line 112: Can this be a union of all the values that would work? either `Cipher[Union[...]]` or `Union[Cipher[X], Cipher[Y], ...]` to reduce verbosity?

alex (approved):

PR #6750, opened 2021-12-23
remove the backend interfaces

reaperhulk:
they have no purpose since we will only have one backend forevermore


src/cryptography/hazmat/backends/__init__.py
```diff
@@ -2,10 +2,8 @@
 # 2.0, and the BSD License. See the LICENSE file in the root of this repository
 # for complete details.
 
-from cryptography.hazmat.backends.interfaces import Backend
 
-
-def default_backend() -> Backend:
+def default_backend():
```
alex, line 6: Ok but why not type this as returning the concrete class?

reaperhulk: Anywhere a user would pass this we type the backend argument typing.Any so the real type is WhoCares, but the typing module doesn’t have that one yet 😁

alex (approved):

PR #6755, opened 2021-12-24
remove a method that only calls another method

reaperhulk:
init_static_locks used to do more, but now we don't need it


reaperhulk:
sigh, external callers.

PR #6756, opened 2021-12-26
consistently use object for other in eq/ne types

reaperhulk:
This appears to be best practice and also makes the mypy type-coverage analysis happier. We were already doing this in various places so now we're consistent.


alex (approved):
not going to pretend to understand the why of this

PR #6763, opened 2021-12-27
add circleci for arm64 linux

reaperhulk:
this runs tests and also builds all three wheels

### DIFF
--- a/.circleci/build-wheel.sh
+++ b/.circleci/build-wheel.sh
@@ -1,0 +1,35 @@
+#!/bin/bash -ex
+
+cd /test
+
+echo "Building for ${PLATFORM}"
+
+PYBIN=/opt/python/cp36-cp36m/bin
+
+mkdir -p /test/wheelhouse.final
+
+"${PYBIN}"/python -m venv .venv
+
+.venv/bin/pip install -U pip wheel cffi setuptools-rust
+
+.venv/bin/python setup.py sdist
+cd dist
+tar zxf cryptography*.tar.gz
+rm -rf cryptograph*.tar.gz
+cd cryptography*
+
+REGEX="cp3([0-9])*"
+if [[ "${PYBIN}" =~ $REGEX ]]; then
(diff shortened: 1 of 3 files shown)


.circleci/build-wheel.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash -ex
```
alex, line 1: We now have like 3 copies of roughly the same script, should we be consolidating?

.circleci/build-wheel.sh
```diff
@@ -0,0 +1,35 @@
...
+
+cd /test
+
+echo "Building for ${PLATFORM}"
+
+PYBIN=/opt/python/cp36-cp36m/bin
```
alex, line 7: This should be a parameter, I'd think?

.circleci/config.yml
```diff
@@ -0,0 +1,91 @@
...
+        type: string
+      command:
+        type: string
+    steps:
+      - run: docker run -e PLATFORM -v $(pwd):/test <<parameters.image>>  /bin/bash -c 'cd /test;<<parameters.command>>'
+
+jobs:
+  linux-arm64:
+    machine:
+      image: ubuntu-2004:202111-01
```
alex, line 22: What is this image?

reaperhulk: circleci releases images quarterly. This is their equivalent of ubuntu-latest, but they don't have an alias that we can use to always stay on latest.

alex: Ok. Follow up task to file an issue to write a GHA job to check for new images? the way we do for boringssl

reaperhulk: Since they release these quarterly the task maybe is a GHA job to file an issue to remind us every 3 months? 😂 I don't think they have any API to fetch these, but whoever tackles this should determine that first

alex (approved):

.circleci/config.yml
alex: Sure, that's fine :-) As a starting point, let's just file an issue to figure out how.

PR #6768, opened 2021-12-27
Informative error on incompatible Fernet key

bharel:
Current error throws a random `binascii.Error` which [causes confusion](https://stackoverflow.com/questions/70499901/binascii-error-incorrect-padding-when-making-an-encrypted-password). I've changed it to be a `ValueError`, same as thrown by an incorrect length.

Don't think there's any backwards compatibility concern in here. A different error will be thrown but it's an erroneous input and the wrong error.

### DIFF
--- a/src/cryptography/fernet.py
+++ b/src/cryptography/fernet.py
@@ -34,8 +34,7 @@ class Fernet(object):
             key = base64.urlsafe_b64decode(key)
         except binascii.Error as exc:
             raise ValueError(
-                f"Fernet key must be 32 url-safe base64-encoded bytes, "
-                f"not {key!r}."
+                f"Fernet key must be 32 url-safe base64-encoded bytes."
                 ) from exc
         if len(key) != 32:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


src/cryptography/fernet.py
```diff
@@ -30,7 +30,13 @@ def __init__(
         key: typing.Union[bytes, str],
         backend: typing.Any = None,
     ):
-        key = base64.urlsafe_b64decode(key)
+        try:
+            key = base64.urlsafe_b64decode(key)
+        except binascii.Error as exc:
+            raise ValueError(
+                f"Fernet key must be 32 url-safe base64-encoded bytes, "
+                f"not {key!r}."
```
reaperhulk, line 38: Putting the key value in the exception is a dangerous thing to do since there it dramatically increases the odds sensitive material will get accidentally logged at some point. Otherwise I'm okay with this, but you'll need to add a test that exercises this path or else coverage will fail.

bharel: I like your thinking. Fixing as we speak.

alex:
There's also a flake8 issue, please format your code with `black`. Thanks for contributing!

alex (approved):

bharel:
Was the quickest turnaround I've ever seen for any project. Unbelievable. Well done @alex & @reaperhulk, I highly respect it.

PR #6778, opened 2022-01-02
Remove Zuul from our CI

alex:
It was providing arm64 builds, but we now do those on CircleCI


alex:
Before this can be merged we'll need to remove it as a required job from github.

reaperhulk (approved):

PR #6781, opened 2022-01-03
type a context manager and its helper

reaperhulk:


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2035,7 +2035,7 @@ def _zeroed_bytearray(self, length):
         finally:
             self._zero_data(ba, length)
 
-    def _zero_data(self, data, length):
+    def _zero_data(self, data: bytearray, length: int) -> None:
```
alex, line 2038: It's not actually bytearray for both callers.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2023,7 +2023,7 @@ def aead_cipher_supported(self, cipher) -> bool:
         return self._lib.EVP_get_cipherbyname(cipher_name) != self._ffi.NULL
 
     @contextlib.contextmanager
-    def _zeroed_bytearray(self, length):
+    def _zeroed_bytearray(self, length: int) -> typing.Iterator[bytearray]:
```
alex, line 2026: I don't understand how typing works with `contextlib.contextmanager`. I guess this accurately describes the inner function?

reaperhulk: I guess so? I just tried to crib out of typeshed here honestly

alex (approved):

PR #6787, opened 2022-01-05
what if we use lld on windows

reaperhulk:


src/rust/Cargo.toml
```diff
@@ -24,3 +24,6 @@ crate-type = ["cdylib"]
 [profile.release]
 lto = "thin"
 overflow-checks = true
+
+[target.x86_64-pc-windows-msvc]
```
alex, line 28: If we do this, we probably also want the i686 variant

reaperhulk:
This doesn't want to work and I'm skeptical it's of value anyway.

PR #6792, opened 2022-01-06
remove unneeded cache entry

reaperhulk:


alex (approved):

alex:
we actually do use this for:
```
      - uses: actions-rs/install@v0.1.2
        with:
          crate: cargo-binutils
          version: latest
        if: steps.cargo-cache.outputs.cache-hit != 'true'
```

you can either remove it from the others but not this one, or kill the PR

PR #6793, opened 2022-01-06
make the pip cache work in docker containers

reaperhulk:


.github/workflows/ci.yml
```diff
@@ -168,6 +168,9 @@ jobs:
           repository: "google/wycheproof"
           path: "wycheproof"
           ref: "master"
+      # When run in a docker container the home directory doesn't have the same owner as the apparent user so pip refuses to create a cache dir
+      - name: create pip cache dir
+        run: mkdir -p /github/home/.cache/pip
```
alex, line 173: a) is there an env var, `$HOME` or something for the path?
b) please line wrap the comment
c) does this actually make anything faster, given that we don't store this dir in the actions cache?

alex (approved):

PR #6800, opened 2022-01-06
Don't declare static functions in .h

alex:
The point of static functions is that they're not exported and the point of a .h file is to export things.


alex:
despite working locally, its not working in CI.

PR #6803, opened 2022-01-09
Document cross compilation target == host bugfix.

jameshilliard:
This fix is needed for example when cross compiling for a `x86_64` target from a `x86_64` build host.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -160,6 +160,18 @@ use ``--no-binary``.
 
     $ pip install cryptography --no-binary cryptography
 
+Cross Compiling
+~~~~~~~~~~~~~~~
+
+If you are cross compiling for a target system that has the same target triple
+as the build host you should add these variables in your environment to prevent
+cargo from linking host build-scripts against the target libraries.
+
+.. code-block:: console
+
+    __CARGO_TEST_CHANNEL_OVERRIDE_DO_NOT_USE_THIS="nightly"
+    CARGO_UNSTABLE_TARGET_APPLIES_TO_HOST="true"
+    CARGO_TARGET_APPLIES_TO_HOST="false"
 
 Using your own OpenSSL on Linux
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


docs/installation.rst
```diff
@@ -160,6 +160,18 @@ use ``--no-binary``.
...
+Cross Compiling
+~~~~~~~~~~~~~~~
+
+If you are cross compiling for a target system that has the same target triple
+as the build host you should add these variables in your environment to prevent
+cargo from linking host build-scripts against the target libraries.
+
+.. code-block:: console
+
+    __CARGO_TEST_CHANNEL_OVERRIDE_DO_NOT_USE_THIS="nightly"
```
alex, line 172: The `DO_NOT_USE_THIS` is kind of a red-flag that this probably isn't the right way to do this 😂

I'm not sure I understand the use case here, what does it mean to cross-compile for a target with the same triple as the host?

jameshilliard: > The `DO_NOT_USE_THIS` is kind of a red-flag that this probably isn't the right way to do this 😂

It's the correct way to do this on a stable rust toolchain, because the fix is behind a nightly flag and that's the only way to enable a nightly flag on the stable toolchain. By policy nightly flags are not supported/documented as available on stable toolchains but there's not technical reason that they don't won't work(they work effectively the same as they do on a nightly toolchain if you set `__CARGO_TEST_CHANNEL_OVERRIDE_DO_NOT_USE_THIS="nightly"`), this particular nightly only feature is needed to fix a critical cross compilation toolchain bug. I've [been trying to stabilize this for a while](https://github.com/rust-lang/cargo/pull/9753) but that's stalled for unclear reasons.

> I'm not sure I understand the use case here, what does it mean to cross-compile for a target with the same triple as the host?

For example if the host triple(ie `x86_64-unknown-linux-gnu`) is identical to the target triple(ie `x86_64-unknown-linux-gnu`) then you need to set this in the env when cross compiling otherwise the build will likely fail with cryptic build script linker errors due to cargo using the wrong linker for build scripts.

reaperhulk: Why do you want to cross compile at all if the host triple is the target triple? Wouldn't you just compile it? I'm assuming there's a nuance I'm missing here 😄

jameshilliard: > Why do you want to cross compile at all if the host triple is the target triple? Wouldn't you just compile it? I'm assuming there's a nuance I'm missing here 😄

For example I'm trying to cross compile a rootfs image for an embedded `x86_64` board using [buildroot](https://buildroot.org/).

jameshilliard: We are currently using [this fix with our infrastructure](https://github.com/buildroot/buildroot/blob/ac573c55aac3cbb4257f5388c91321c81095c654/package/pkg-cargo.mk#L26-L50) to build [python-cryptography](https://github.com/buildroot/buildroot/blob/01764876e3d4e7ab95c8365f1c83e6c12d49757c/package/python-cryptography/python-cryptography.mk#L17).

alex: (Apologies for not following up here more promptly.)

I guess the thing I don't understand is why this is considered cross-compilation at all? If the host and target are both "the same", what's being solved here?

reaperhulk:
Apologies for letting this sit around so long, but reading through this I'm still confused. Is this some unusual thing where you're cross compiling for another rust triple but both the host triple and the target triple have the same architecture? You're definitely an expert on cross compilation compared to us and I think it's useful for this project to have more help on cross compiling in the docs, but we need to understand it before we can merge it. 😄

alex:
Going to close this since it's been quiet for a month -- we're still happy to merge if you can help us understand the use case, so please leave a comment if you're interested!

PR #6827, opened 2022-01-29
Remove a bunch of bindings now that there's been a pyOpenSSL release

alex:


alex:
mitmproxy appears to still pin to an older pyopenssl, so we can't actually merge this.

PR #6832, opened 2022-01-31
point to the AEAD version of GCM in the FAQ

alex:


reaperhulk (approved):
Should we consider updating this for modern reasons or entirely removing it given our position in the ecosystem now?

alex:
I don't feel strongly about it.

PR #6836, opened 2022-02-02
attempt to disable spotlight

alex:


alex:
apparently we also need to kill the running process, which I don't know how to do

reaperhulk:
Maybe killall mdworker?

alex:
no evident time savings

PR #6845, opened 2022-02-06
Timeout fetching from cache

alex:
Occasionally it randomly fails


reaperhulk:
Any reason this shouldn’t just be 1?

alex:
I figure this saves 1/3 time in the (rare) failure case, and has a 0% chance of being a false positive. 5 minutes is about the same amount of time a regular job takes, so there's not actually much savings in going with something shorter since you can't restart until all jobs are done.

reaperhulk (approved):

PR #6852, opened 2022-02-08
fixes #6835 -- add signatures with sha3 for rsa and ecdsa

### LINKED ISSUE #6835: sha3 hash algorithms not supported for signatures
Hi, I found possible issue. I was trying to use sha3 hash algorithm to create certificate request, but it is no longer supported.
It was working on cryptography version 3.4.8.

OpenSSL 1.1.1m  14 Dec 2021 (tested also on OpenSSL 1.1.1f  31 Mar 2020)
Python 3.8.10
pip 20.0.2
cryptography==36.0.1
cffi==1.15.0
setuptools 60.2.0

Cryptography installed with pip install

Code to reproduce the issue:
```
csr = x509.CertificateSigningRequestBuilder().subject_name(x509.Name(name_attributes)).sign(pkey,
                                                                                                hash_digest, <cryptography.hazmat.primitives.hashes.SHA3_512>
                                                                                                default_backend())
```
Error msg: Hash algorithm "sha3-512" not supported for signatures
Same happens for other sha3-xxx algorithms.
sign () Call works for "sha512",  "md5", "sha1", "sha224".

I guess error msg comes from: https://github.com/pyca/cryptography/blob/main/src/rust/src/x509/sign.rs#L97

alex:


reaperhulk:
I think we need a changelog note to say we re-added support for this that was broken in 36.x

reaperhulk (approved):

PR #6865, opened 2022-02-13
allow ed pk12, better tests

### LINKED ISSUE #6854: Can load PKCS12 with ED25519 Keys but cannot Serialize them
Why does the pkcs12.serialize_key_and_certificates() still sanitize against ed private keys?  cryptography has no problem loading pkcs12 files which contain ed25519 private keys and related certificates.

whiteowl3:
fixes #6854

### DIFF
--- a/tests/hazmat/primitives/test_pkcs12.py
+++ b/tests/hazmat/primitives/test_pkcs12.py
@@ -35,6 +35,15 @@ from ...doubles import DummyKeySerializationEncryption
 from ...utils import load_vectors_from_file
 
 
+def _skip_curve_unsupported(backend, curve):
+    if not backend.elliptic_curve_supported(curve):
+        pytest.skip(
+            "Curve {} is not supported by this backend {}".format(
+                curve.name, backend
+            )
+        )
+
+
 @pytest.mark.skip_fips(
     reason="PKCS12 unsupported in FIPS mode. So much bad crypto in it."
 )
@@ -293,7 +302,11 @@ def _load_ca(backend):
 )
 class TestPKCS12Creation:
     @pytest.mark.parametrize(
-        ("kgenerator", "ktype", "kparam"),
+        (
+            "kgenerator",
+            "ktype",
+            "kparam",
+        ),
         [
             pytest.param(
                 ed448.Ed448PrivateKey.generate,
@@ -318,19 +331,9 @@ class TestPKCS12Creation:
         ]
         + [
             pytest.param(
-                ec.generate_private_key,
-                ec.EllipticCurvePrivateKey,
-                [curve],
-                marks=pytest.mark.supported(
-                    only_if=lambda backend: backend.elliptic_curve_supported(
-                        curve
-                    ),
-                    skip_message=(
-                        "Requires OpenSSL with " + curve.name + " support"
-                    ),
-                ),
+                ec.generate_private_key, ec.EllipticCurvePrivateKey, [curve]
             )
-            for curve in [ec._CURVE_TYPES[i]() for i in ec._CURVE_TYPES]
+            for curve in ec._CURVE_TYPES.values()
         ],
     )
     @pytest.mark.parametrize("name", [None, b"name"])
@@ -344,6 +347,9 @@ class TestPKCS12Creation:
     def test_generate_each_supported_keytype(
         self, backend, kgenerator, ktype, kparam, name, algorithm, password
     ):
+        if ktype == ec.EllipticCurvePrivateKey:
+            _skip_curve_unsupported(backend, *kparam)
+
         key = kgenerator(*kparam)
 
         assert isinstance(key, ktype)


whiteowl3:
i suppose now that the test functions have all been combined into one, its no longer necessary to use the cert function to prevent duplication.... le sigh.

whiteowl3:
updated documentation,
added WithSerialization aliases for ed448 and ed25519 (not sure why tho, just trying to fit in)

docs/hazmat/primitives/asymmetric/ed25519.rst
```diff
@@ -103,6 +103,12 @@ Key interfaces
 
         :return bytes: Serialized key.
 
+.. class:: Ed25519PrivateKeyWithSerialization
```
reaperhulk, line 106: We don't need these (they're legacy we've deprecated on our other classes)

whiteowl3: that leaves a question of how you want the serialization docs to look.  would you rather i change all the allowed types to remove the "withserialization" aliases or just leave it so that ed448/ed25519 private keys are referenced directly and Ec/RSA/DSA keys are still referred to by their alias? i just think it would look weird for some private key types to be "with serialization" and some not.

whiteowl3: ok this is the way that makes the most sense to me, given that these are the types that are actually checked for, but it does rather broaden the scope of the pr to change them here. please advise at your convenience.

tests/hazmat/primitives/test_pkcs12.py
```diff
@@ -290,43 +336,44 @@ class TestPKCS12Creation:
             (serialization.NoEncryption(), None),
         ],
     )
-    def test_generate(self, backend, name, encryption_algorithm, password):
-        cert, key = _load_ca(backend)
+    def test_generate_each_supported_keytype(
+        self, backend, ktype, kparam, name, encryption_algorithm, password
+    ):
+        if ktype.__module__.rsplit(".", 1)[-1].startswith("ed"):
```
reaperhulk, line 342: I don't think this is the right approach here at all. If you're looking to resist duplication then why not just make it parametrize over the callable that is the generation function and pass parameters if kparam is not None? Using `__module__` and `__import__` is definitely not something we want to land.

tests/hazmat/primitives/test_pkcs12.py
```diff
@@ -282,6 +292,42 @@ def _load_ca(backend):
...
+            ),
+            (rsa.RSAPrivateKey, (65537, 1024)),
+            # (rsa.RSAPrivateKey, (65537, 2048)),
+            # (rsa.RSAPrivateKey, (65537, 3072)),
+            # (rsa.RSAPrivateKey, (65537, 4096)),
+            (dsa.DSAPrivateKey, (1024,)),
+            # (dsa.DSAPrivateKey, (2048,)),
+            # (dsa.DSAPrivateKey, (3072,)),
+            # (dsa.DSAPrivateKey, (4096,)),
+            (ec.EllipticCurvePrivateKey, (ec.SECP256R1,)),
```
reaperhulk, line 322: I don't think we really need every possible type here, but there is a curve supported check: https://github.com/pyca/cryptography/blob/main/src/cryptography/hazmat/backends/openssl/backend.py#L1183

tests/hazmat/primitives/test_pkcs12.py
```diff
@@ -282,6 +292,42 @@ def _load_ca(backend):
...
+            pytest.param(
+                ed25519.Ed25519PrivateKey,
+                None,
+                marks=pytest.mark.supported(
+                    only_if=lambda backend: backend.ed25519_supported(),
+                    skip_message="Requires OpenSSL with Ed25519 support",
+                ),
+            ),
+            (rsa.RSAPrivateKey, (65537, 1024)),
+            # (rsa.RSAPrivateKey, (65537, 2048)),
```
reaperhulk, line 315: We can drop all the commented out RSA/DSA options -- the core test here is whether it can serialize the key types.

tests/hazmat/primitives/test_pkcs12.py
whiteowl3: doing it this way prevents the use of a third parameter, iiuc.  we pass the ktype both to this line and to the isinstance check on the key.  do i understand correctly that doing it the way you suggest would require parametrizing both the generator and the key type? can you help me understand your resistance to ```__module__``` and ```__import__``` in this context?

tests/hazmat/primitives/test_pkcs12.py
whiteowl3: my instinct was to test all curves which the docs don't recommend against, but ill follow your lead.  which should be tested?

tests/hazmat/primitives/test_pkcs12.py
reaperhulk: The readability of this is lower than simply passing the callable and type. Readers need to go look up the module that a given ktype is a member of so they can see what the substring match does, as well as understand what `__import__` does and how it returns. And if that module name changes this would break in an unexpected way. That module name **won't** change, but we shouldn't depend on it this way.

Another argument for isinstance is fine :)

tests/hazmat/primitives/test_pkcs12.py
reaperhulk: If you want to add them all I have no objection, you just need to do a curve supported check.

tests/hazmat/primitives/test_pkcs12.py
whiteowl3: this comment kind of addresses both points.  I wasnt able to do this before because I didn't notice the curve support check, but one of the reasons I like this code is because it paves the way to do nastily exhaustive things like this:
```
    @pytest.mark.parametrize(
        ("ktype", "kparam"),
        [
            pytest.param(
                ed448.Ed448PrivateKey,
                None,
                marks=pytest.mark.supported(
                    only_if=lambda backend: backend.ed448_supported(),
                    skip_message="Requires OpenSSL with Ed448 support",
                ),
            ),
            pytest.param(
                ed25519.Ed25519PrivateKey,
                None,
                marks=pytest.mark.supported(
                    only_if=lambda backend: backend.ed25519_supported(),
                    skip_message="Requires OpenSSL with Ed25519 support",
                ),
            ),
            (rsa.RSAPrivateKey, (65537, 1024)),
            (dsa.DSAPrivateKey, (1024,)),
        ]
        + [
            pytest.param(
                ec.EllipticCurvePrivateKey,
                (curve,),
                marks=pytest.mark.supported(
                    only_if=lambda backend: backend.elliptic_curve_supported(
                        curve
                    ),
                    skip_message=(
                        "Requires OpenSSL with " + curve.name + " support"
                    ),
                ),
            )
            for curve in [ec._CURVE_TYPES[i]() for i in ec._CURVE_TYPES]
        ],
    )
```

whiteowl3:
Something very wonky is up that makes the alpine tests fail when using the ```marks=pytest.supported(only_if=lambda  backend: backend.thingsupported())``` syntax, so i pulled the logic out of test_ec.py and that seems to make it happy.  __i think it would be very worthwhile to understand why only alpine passes or fails based on the difference between these two commits.__
if it is obvious to you please explain, otherwise, i think this addresses all your reviews.

alex:
The problem is with how lambdas capture variables in loops:

```
>>> vals = [lambda: i for i in range(10)]
>>> [v() for v in vals]
[9, 9, 9, 9, 9, 9, 9, 9, 9, 9]
```

whiteowl3:
> The problem is with how lambdas capture variables in loops:
I appreciate this explanation. why it affects only alpine still doesn't make sense to me.  the loop works on all other test systems.

whiteowl3:
oh. because it makes it switch the entire suite on brainpoolP512r1. lame.

reaperhulk (approved):

PR #6913, opened 2022-02-28
Begin deprecation process for Python 3.6

alex:


reaperhulk (approved):
This is likely to be a very long deprecation process, but as @alex and I discussed in IRC, we have a precedent for long deprecation with various OpenSSL releases. We won't be removing 3.6 support until it drops much, much lower than our current download numbers show.

Issue #6880, opened 2022-02-17
DSA/DSS will be disallowed in FIPS mode


tiran:
[FIPS PUB 186-5](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5-draft.pdf) (Draft) paragraph 4 *The Digital Signature Algorithm (DSA)* lists DSA as no longer approved. The document is still a draft and not yet final. RHEL downstream engineers do not expect any changes to DSA disapproval and blocked DSA in C9S / RHEL 9.

Like #6879 the algorithm is blocked by downstream patch https://gitlab.com/redhat/centos-stream/rpms/openssl/-/blob/c9s/0045-FIPS-services-minimize.patch . The patch causes calls to ``EVP_PKEY_get1_DSA`` to fail with error ``error:030000CD:digital envelope routines::keymgmt export failure``.

reaperhulk:
Do we need to just skip DSA in FIPS tests and quietly wish it never existed?

tiran:
Yes, I wish DSA never existed in the first place. (That's the typical third wish in every fairy tale. Undo all horrible events.)

* DSA key generation, signature generation, and signature verification should raise UnsupportedAlgorithm error
* DSA tests should be skipped

reaperhulk:
Item two is the easy one, so let's do that first. Friendly errors when trying to access disallowed algorithms in FIPS is currently not something `cryptography` even tries to do, and making it work reliably across Red Hat's patches as well as the OpenSSL 3 FIPS provider is likely to be quite a bit of work. 😄

alex:
fixed by #6916

PR #6916, opened 2022-03-01
Disable DSA tests in FIPS mode

tiran:
See: #6880

### DIFF
--- a/tests/hazmat/primitives/test_dsa.py
+++ b/tests/hazmat/primitives/test_dsa.py
@@ -55,6 +55,10 @@ def test_skip_if_dsa_not_supported(backend):
         _skip_if_dsa_not_supported(backend, DummyHashAlgorithm(), 1, 1, 1)
 
 
+@pytest.mark.supported(
+    only_if=lambda backend: backend.dsa_supported(),
+    skip_message="Does not support DSA.",
+)
 class TestDSA:
     def test_generate_dsa_parameters(self, backend):
         parameters = dsa.generate_parameters(2048, backend)
@@ -385,6 +389,10 @@ class TestDSA:
         ).private_key(backend)
 
 
+@pytest.mark.supported(
+    only_if=lambda backend: backend.dsa_supported(),
+    skip_message="Does not support DSA.",
+)
 class TestDSAVerification:
     def test_dsa_verification(self, backend, subtests):
         vectors = load_vectors_from_file(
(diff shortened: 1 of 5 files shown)


reaperhulk:
Could you push an empty commit here? codecov is being dumb.

reaperhulk:
Coverage issue looks legit now (some FIPS checks that now don't take the branches since FIPS entirely skips them)

tests/x509/test_x509.py
```diff
@@ -2646,7 +2646,9 @@ def test_sign_ec_with_md5(self, backend):
     def test_build_cert_with_dsa_private_key(
         self, hashalg, hashalg_oid, backend
     ):
-        if backend._fips_enabled and hashalg is hashes.SHA1:
+        if (
+            backend._fips_enabled and hashalg is hashes.SHA1
+        ):  # pragma: no cover
```
reaperhulk, line 2651: If we always skip FIPS due to the dsa_supported check returning False when FIPS is enabled, this entire branch can be removed, right?

tiran:
I don't understand why CodeCov is complaining about skip class decorator in two out of four places. https://app.codecov.io/gh/pyca/cryptography/compare/6916/tree/tests/hazmat/primitives/test_dsa.py#D1L605

tests/hazmat/primitives/test_dsa.py
```diff
@@ -511,6 +518,10 @@ def test_prehashed_digest_mismatch(self, backend):
             private_key.sign(digest, prehashed_alg)
 
 
+@pytest.mark.supported(
+    only_if=lambda backend: backend.dsa_supported(),
+    skip_message="Does not support DSA.",
+)
```
alex, line 524: This isn't necessary, none of these tests touch the backend.

tests/hazmat/primitives/test_dsa.py
```diff
@@ -591,6 +602,10 @@ def test_repr(self):
         )
 
 
+@pytest.mark.supported(
```
alex, line 605: Same

tiran: Fixed

tests/hazmat/primitives/test_dsa.py
tiran: Fixed

reaperhulk (approved):

PR #6922, opened 2022-03-02
fixes #6809 -- deprecate 4 legacy ciphers

### LINKED ISSUE #6809: Deprecate/remove legacy symmetric algorithms
`cryptography` ships numerous legacy symmetric algorithms that, anecdotally, have very low use. These algorithms have a variety of security drawbacks when compared to more modern equivalents (namely AES or ChaCha20) and generally shouldn't be used. We plan to deprecate them on a long cycle such that users have a chance to speak up if they have a use case that we should consider.

The algorithms we'd like to deprecate and remove are:
* CAST5
* SEED
* IDEA
* Blowfish

3DES and ARC4 are also legacy (and ARC4 has serious security issues), but their common use in many scenarios means they will not be deprecated.

alex:
all the `_Internal` contortions are to keep our tests warnings free

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,13 @@ Changelog
 * Deprecated Python 3.6 support. Python 3.6 is no longer supported by the
   Python core team. Support for Python 3.6 will be removed in a future
   ``cryptography`` release.
+* Deprecated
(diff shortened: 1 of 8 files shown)


alex:
ok, https://github.com/pyca/cryptography/pull/6923 should help here

reaperhulk (approved):

PR #6928, opened 2022-03-03
fixes #6927 -- handle negative return values from openssl

### LINKED ISSUE #6927: Insufficient error check in RSA: EVP_PKEY_CTX_set_signature_md() can return < 0
The latest downstream build of OpenSSL 3.0.1 in RHEL 9 has started to block SHA-1 signatures with the ``DEFAULT`` crypto policy. That means that SHA-1 signatures are blocked even in non-FIPS mode.

During testing we found a problem in the error check of ``EVP_PKEY_CTX_set_signature_md()``. https://github.com/pyca/cryptography/blob/3fd9d5a8424eef3e08260a622ae3644777ac5339/src/cryptography/hazmat/backends/openssl/rsa.py#L198-L213

``_rsa_sig_setup`` checks for ``res == 0`` but ``EVP_PKEY_CTX_set_signature_md`` can [return a negative value](https://www.openssl.org/docs/man3.0/man3/EVP_PKEY_CTX_set_signature_md.html):

> All other functions described on this page return a positive value for success and 0 or a negative value for failure. In particular a return value of -2 indicates the operation is not supported by the public key algorithm.

Tested with 3.4.7 and 36.0.1.

Downstream bug: https://bugzilla.redhat.com/show_bug.cgi?id=2060343

alex:


tiran:
Patch works for me. Thanks!

```
  File "/tmp/testrsa.py", line 51, in <module>
    signature = private_key.sign(
  File "/usr/lib64/python3.9/site-packages/cryptography/hazmat/backends/openssl/rsa.py", line 492, in sign
    return _rsa_sig_sign(self._backend, padding, algorithm, self, data)
  File "/usr/lib64/python3.9/site-packages/cryptography/hazmat/backends/openssl/rsa.py", line 241, in _rsa_sig_sign
    pkey_ctx = _rsa_sig_setup(
  File "/usr/lib64/python3.9/site-packages/cryptography/hazmat/backends/openssl/rsa.py", line 210, in _rsa_sig_setup
    raise UnsupportedAlgorithm(
cryptography.exceptions.UnsupportedAlgorithm: sha1 is not supported by this backend for RSA signing.
```

tiran (approved):

reaperhulk (approved):

PR #6931, opened 2022-03-03
Dedicated check for signature hash algorithms

tiran:
Move the check for FIPS mode and blocked SHA1 signature algorithm
into the backend code. Some distros will block SHA1 for RSA signatures
in the near future. The new ``signature_hash_supported()`` method will
allow us to flip the switch in one place.

Note: The ban of SHA1 signatures does not affect MGF1's inner hash algorithm.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -326,6 +326,15 @@ class Backend:
         evp_md = self._evp_md_from_algorithm(algorithm)
         return evp_md != self._ffi.NULL
 
+    def signature_hash_supported(
+        self, algorithm: hashes.HashAlgorithm
+    ) -> bool:
+        # Dedicated check for hashing algorithm use in message digest for
+        # signatures, e.g. RSA PKCS#1 v1.5 SHA1 (sha1WithRSAEncryption).
+        if self._fips_enabled and isinstance(algorithm, hashes.SHA1):
+            return False
+        return self.hash_supported(algorithm)
+
     def scrypt_supported(self) -> bool:
         if self._fips_enabled:
             return False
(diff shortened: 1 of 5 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -723,7 +732,8 @@ def rsa_padding_supported(self, padding: AsymmetricPadding) -> bool:
         if isinstance(padding, PKCS1v15):
             return True
         elif isinstance(padding, PSS) and isinstance(padding._mgf, MGF1):
-            # SHA1 is permissible in MGF1 in FIPS
+            # SHA1 is permissible in MGF1 in FIPS and when SHA1 is blocked
```
reaperhulk, line 735: is this supposed to say "even when" rather than "and when"?

reaperhulk (commented):
this lgtm other than my one question

src/cryptography/hazmat/backends/openssl/backend.py
tiran: Yes ...

tiran, line 735:
```suggestion
            # SHA1 is permissible in MGF1 in FIPS even when SHA1 is blocked
```

reaperhulk:
(closing/reopening to give codecov a chance to actually respond)

reaperhulk:
Can you push an empty commit to this? GitHub's status is broken (and Codecov may be unhappy, but unclear since the status is broken)

alex:
Bleh, needs another rebase due to changes in CI infra.

tiran:
I have rebased my PR onto main.

reaperhulk (approved):

PR #6935, opened 2022-03-05
Upgrade to pyo3 0.18

alex:
Follow up action items:
- [ ] Convert `crate::intern!()` to use `pyo3::intern!()` for basically all static strings which are converted to PyObject
- [ ] Use `Py<PyBytes>::as_bytes` instead of copying data to vec.
- [x] Remove `let py = gil.python();` in many places
- [ ] investigate moving some enums into rust

### DIFF
--- a/src/rust/src/x509/sign.rs
+++ b/src/rust/src/x509/sign.rs
@@ -59,15 +59,15 @@ fn identify_key_type(py: pyo3::Python<'_>, private_key: &pyo3::PyAny) -> pyo3::P
         .getattr(crate::intern!(py, "Ed448PrivateKey"))?
         .extract()?;
 
-    if rsa_private_key.is_instance(private_key)? {
+    if private_key.is_instance(rsa_private_key)? {
         Ok(KeyType::Rsa)
-    } else if dsa_key_type.is_instance(private_key)? {
+    } else if private_key.is_instance(dsa_key_type)? {
         Ok(KeyType::Dsa)
-    } else if ec_key_type.is_instance(private_key)? {
+    } else if private_key.is_instance(ec_key_type)? {
         Ok(KeyType::Ec)
-    } else if ed25519_key_type.is_instance(private_key)? {
+    } else if private_key.is_instance(ed25519_key_type)? {
         Ok(KeyType::Ed25519)
-    } else if ed448_key_type.is_instance(private_key)? {
+    } else if private_key.is_instance(ed448_key_type)? {
         Ok(KeyType::Ed448)
     } else {
         Err(pyo3::exceptions::PyTypeError::new_err(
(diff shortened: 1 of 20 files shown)


src/rust/src/x509/sign.rs
```diff
@@ -55,15 +55,15 @@ fn identify_key_type(py: pyo3::Python<'_>, private_key: &pyo3::PyAny) -> pyo3::P
         .getattr("Ed448PrivateKey")?
         .extract()?;
 
-    if rsa_private_key.is_instance(private_key)? {
+    if private_key.is_instance(rsa_private_key)? {
```
reaperhulk, line 58: Flipping the `.is_instance` calls isn't mentioned in the pyo3 0.15 -> 0.16 migration guide. The new format makes way more sense (checking that the instance of the object is an instance of a type rather than calling is_instance on the type), but was it just not properly documented for  https://pyo3.rs/v0.16.2/migration.html?

reaperhulk:
Could you also rebase this just to see if it fixes coverage?

alex:
@reaperhulk it won't -- the coverage issue is caused by removing the RHEL8-fips builder because it's py36. When we get to this point we'll need to decide what to do about that.

tiran:
> @reaperhulk it won't -- the coverage issue is caused by removing the RHEL8-fips builder because it's py36. When we get to this point we'll need to decide what to do about that.

RHEL 8 comes with Python 3.8. 

```
# dnf install -y python38-devel python38-pip
...
=================================================================================================================================================
 Package                               Architecture       Version                                               Repository                  Size
=================================================================================================================================================
Installing:
 python38-devel                        x86_64             3.8.8-4.module+el8.5.0+12205+a865257a                 rhel-AppStream             276 k
 python38-pip                          noarch             19.3.1-4.module+el8.5.0+12205+a865257a                rhel-AppStream             1.9 M
Installing dependencies:
 python38                              x86_64             3.8.8-4.module+el8.5.0+12205+a865257a                 rhel-AppStream              79 k
 python38-libs                         x86_64             3.8.8-4.module+el8.5.0+12205+a865257a                 rhel-AppStream             8.3 M
 python38-pip-wheel                    noarch             19.3.1-4.module+el8.5.0+12205+a865257a                rhel-AppStream             1.2 M
 python38-setuptools                   noarch             41.6.0-5.module+el8.5.0+12205+a865257a                rhel-AppStream             668 k
 python38-setuptools-wheel             noarch             41.6.0-5.module+el8.5.0+12205+a865257a                rhel-AppStream             304 k
Enabling module streams:
 python38                                                 3.8                                                                                   

Transaction Summary
=================================================================================================================================================
Install  7 Packages
...
```

cmhe:
Just FYI: I am currently trying to update the Debian packaging for cryptography to 37, and on Debian pyo3 0.17 packaging is already in the pipeline, while 0.16 is currently available on Debian unstable. Packaging the old pyo3 0.15 version just for cryptography will not happen, so I will likely be ending up picking these patches.

cmhe:
@alex: Do you my chance have a pyo3 0.16 patchset on top of 37.0.4?

alex:
I don't. If it helps, we'll be releasing 38.0 this week.

בתאריך יום ג׳, 6 בספט׳ 2022, 9:50, מאת Claudius Heine ‏<
***@***.***>:

reaperhulk:
@cmhe we've released 38 now, so it may be simpler to package that version, which is compatible with this PR.

xertSuns1 (dismissed):

reaperhulk (approved):

PR #6942, opened 2022-03-09
Begin the changelog for 36.0.2

alex:


reaperhulk:
We're going to have a failing CircleCI check on this branch unless we port the CircleCI config. We probably should just do that?

reaperhulk (approved):

PR #6976, opened 2022-03-18
Extend add_attribute to allow for any ASN.1 type

phughesion:
Previously, add_attribute would always wrap the attribute data as a UTF8String. Now, UTF8String is the default, but passing in another type instead is possible.

### DIFF
--- a/src/cryptography/x509/base.py
+++ b/src/cryptography/x509/base.py
@@ -552,7 +552,7 @@ class CertificateSigningRequestBuilder:
         self,
         subject_name: typing.Optional[Name] = None,
         extensions: typing.List[Extension[ExtensionType]] = [],
-        attributes: typing.List[typing.Tuple[ObjectIdentifier, bytes]] = [],
+        attributes: typing.List[typing.Tuple[ObjectIdentifier, bytes, int]] = [],
     ):
         """
         Creates an empty X.509 certificate request (v1).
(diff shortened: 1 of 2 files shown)


src/cryptography/x509/base.py
```diff
@@ -592,7 +592,7 @@ def add_extension(
         )
 
     def add_attribute(
-        self, oid: ObjectIdentifier, value: bytes
+        self, oid: ObjectIdentifier, value: bytes, tag: int = 12
```
reaperhulk, line 595: We have https://github.com/pyca/cryptography/blob/ccfef6f1076c9581e7641d52664ce90de0fd7e0a/src/cryptography/x509/name.py#L16 so you don't need to have it be an integer.

alex: Are you suggesting the argument have that type, or just that we use it to get the UTF8String constant?

src/rust/src/x509/csr.rs
```diff
@@ -424,15 +424,8 @@ fn create_x509_csr(
     }
 
     for py_attr in builder.getattr("_attributes")?.iter()? {
-        let (py_oid, value): (&pyo3::PyAny, &[u8]) = py_attr?.extract()?;
+        let (py_oid, value, tag): (&pyo3::PyAny, &[u8], u8) = py_attr?.extract()?;
         let oid = py_oid_to_oid(py_oid)?;
-        let tag = if std::str::from_utf8(value).is_ok() {
```
alex, line 429: This removes the validation even when tag _is_ Utf8String, I think we'd still want to keep that?

src/cryptography/x509/base.py
reaperhulk: I think we should use the type itself and probably make it _tag as well

orangepizza: that class misses quite a thing like octetstring

PR #6981, opened 2022-03-19
Switch to codecov official github action

alex:
The bash uploader we use is deprecated. I'm really hoping this helps with the reliability.


alex:
not actually uploading stuff, too tired to care why

PR #7002, opened 2022-03-26
add TLSv1.3 PSKs function bindings

cocobear:
Accord to document of openssl use TLSv1.3 PSKs should use
 [SSL_CTX_set_psk_find_session_callback](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_psk_find_session_callback.html)
[SSL_CTX_set_psk_use_session_callback](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_psk_use_session_callback.html)

also other functions needed.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -21,6 +21,7 @@ static const long Cryptography_HAS_SECURE_RENEGOTIATION;
 static const long Cryptography_HAS_SSL_CTX_CLEAR_OPTIONS;
 static const long Cryptography_HAS_DTLS;
 static const long Cryptography_HAS_PSK;
+static const long Cryptography_HAS_PSK_TLSv1_3;
 static const long Cryptography_HAS_VERIFIED_CHAIN;
 static const long Cryptography_HAS_KEYLOG;
 static const long Cryptography_HAS_GET_PROTO_VERSION;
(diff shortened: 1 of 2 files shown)


src/_cffi_src/openssl/ssl.py
```diff
@@ -255,6 +256,26 @@
...
+                                           ));
+void SSL_CTX_set_psk_use_session_callback(SSL_CTX *,
+                                          int (*)(
+                                              SSL *,
+                                              const EVP_MD *,
+                                              const unsigned char **,
+                                              size_t *,
+                                              SSL_SESSION **
+                                          ));
+const SSL_CIPHER *SSL_CIPHER_find(SSL *, const unsigned char *);
```
reaperhulk, line 274: All these functions need to be removed in the conditional if they're unavailable, much like you're doing for `SSL_CTX_set_psk_use_session_callback`

src/_cffi_src/openssl/ssl.py
```diff
@@ -255,6 +256,27 @@
...
+SSL_SESSION *SSL_SESSION_new(void);
```
reaperhulk, line 275: `build/temp.linux-x86_64-3.10/_openssl.c:30646:10: error: too few arguments to function ‘SSL_SESSION_new’`.

In boring this function is declared as taking `const SSL_CTX *ctx`. You may need to declare a `Cryptography_SSL_SESSION_new()` and implement it in `CUSTOMIZATIONS` to avoid namespace collision here.

It looks like OpenSSL's handling of TLS 1.3 PSK has significantly changed the use case of SSL_SESSION_new, as boring (which forked from 1.0.2) documents it as follows:

> SSL_SESSION_new returns a newly-allocated blank SSL_SESSION or NULL on error. This may be useful when writing tests but should otherwise not be used.

It would be useful to know how boringssl handles PSK in TLS 1.3 as it is definitely a different API.

cocobear: I add `Cryptography_SSL_SESSION_new`, it seems to be ok.
But I didn't know how boringssl handles PKS in TLS 1.3, so only openssl 1.1.1 and above support, is that ok to you?

Maybe I should put the functions in the blocks they should have been in. like:
image

reaperhulk: There's no need to sort them that way.

I'm a bit uncomfortable with the divergence of APIs here but I also don't have any good solution. Boring and Libre will continue to diverge.

Is the intent here to add support for PSK on the pyOpenSSL side or is this just for private use? Be aware that if we don't know about API consumers we sometimes remove functions to try to keep our bindings small (and make compatibility across the forks easier).

cocobear: It's intent to add support for PSK TLS 1.3 on pyOpenSSL.
Accord to openssl s_server.c source code , psk_find_session_cb should call  functions which I diverged.

https://github.com/openssl/openssl/blob/master/apps/s_server.c

reaperhulk (approved):

PR #7011, opened 2022-03-29
support 12-15 byte nonce sizes in OCB3

### LINKED ISSUE #6894: OCB mode (AESOCB3) should support nonce sizes up to 15 bytes
[`cryptography.hazmat.primitives.ciphers.aead.AESOCB3._check_params`](https://github.com/pyca/cryptography/blob/d98e0bd355544a28cc36aba3a05d6e44604ac289/src/cryptography/hazmat/primitives/ciphers/aead.py#L281) requires that the nonce size be fixed at 12 bytes, or 96 bits.

However, [RFC 7253 §4.2](https://datatracker.ietf.org/doc/html/rfc7253#section-4.2) suggests that the nonce can go up to 120 bits.

The OpenSSL backend to cryptography doesn't appear to be the source of the limitation: [OpenSSL's OCB interface](https://github.com/openssl/openssl/blob/master/doc/man3/EVP_EncryptInit.pod#gcm-and-ocb-modes) suggests that `EVP_CTRL_AEAD_SET_IVLEN` permits setting a nonce up to 15 bytes (120 bits), in line with RFC 7253.  (This has to be done between the first `EVP_CipherInit_ex` (where the `iv` parameter should be `NULL`) and the subsequent `EVP_CipherInit_ex` invocation that sets the `iv`, if i'm understanding the OpenSSL interface correctly.)

The [next revision of the OpenPGP standard](https://datatracker.ietf.org/doc/draft-ietf-openpgp-crypto-refresh/) looks likely to encourage the use of OCB, and is in discussion about [how many bits to use for the nonce](https://gitlab.com/openpgp-wg/rfc4880bis/-/issues/83).  If it goes with anything other than 96, any python implementation of OpenPGP (e.g. [pgpy](https://github.com/SecurityInnovation/PGPy/)) would need to write their own OCB wrapper around even lower-level `hazmat` primitives, or use some python module other than `cryptography`.

It would be great if cryptography would be able to accept nonces up to 120 bits (15 bytes) to accomodate this use case.

reaperhulk:
fixes #6894

### DIFF
--- a/tests/hazmat/primitives/test_aead.py
+++ b/tests/hazmat/primitives/test_aead.py
@@ -495,10 +495,20 @@ class TestAESOCB3:
             aesocb3.encrypt(nonce, b"", FakeData())
 
     def test_vectors(self, backend, subtests):
-        vectors = load_vectors_from_file(
-            os.path.join("ciphers", "AES", "OCB3", "rfc7253.txt"),
-            load_nist_vectors,
-        )
+        vectors = []
+        for f in (
+            "rfc7253.txt",
+            "test-vector-1-nonce104.txt",
+            "test-vector-1-nonce112.txt",
+            "test-vector-1-nonce120.txt",
+        ):
+            vectors.extend(
+                load_vectors_from_file(
+                    os.path.join("ciphers", "AES", "OCB3", f),
+                    load_nist_vectors,
+                )
+            )
+
         for vector in vectors:
             with subtests.test():
                 nonce = binascii.unhexlify(vector["nonce"])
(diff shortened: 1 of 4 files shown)


tests/hazmat/primitives/test_aead.py
```diff
@@ -495,10 +495,20 @@ def test_data_too_large(self):
...
-            os.path.join("ciphers", "AES", "OCB3", "rfc7253.txt"),
-            load_nist_vectors,
-        )
+        vectors = []
+        for f in (
+            "rfc7253.txt",
+            "test-vector-1-nonce104.txt",
+            "test-vector-1-nonce112.txt",
+            "test-vector-1-nonce120.txt",
+        ):
```
alex, line 504:
```suggestion
        for f in [
            "rfc7253.txt",
            "test-vector-1-nonce104.txt",
            "test-vector-1-nonce112.txt",
            "test-vector-1-nonce120.txt",
        ]:
```

nit: tuples are heterogenous records, list are homogeneous sequences

alex (approved):

PR #7017, opened 2022-04-01
Move linkcheck back into ci.yml

alex:
The reason we moved it out was because github didn't have a way to retry only failed jobs


.github/workflows/ci.yml
```diff
@@ -544,6 +544,30 @@ jobs:
...
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    runs-on: ubuntu-latest
+    name: "linkcheck"
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v2.4.0
+        with:
+          persist-credentials: false
+      - name: Setup python
+        uses: actions/setup-python@v3.0.0
```
reaperhulk, line 557:
```suggestion
        uses: actions/setup-python@v3.1.0
```

reaperhulk (approved):

PR #7028, opened 2022-04-03
small refactors for upcoming SIV support

reaperhulk:
use f strings (why not) and internally use lists for processing AD (SIV will take a list of bytes rather than bytes for AD)


alex (approved):
all these wasted allocations bother me, but what can you do

PR #7038, opened 2022-04-05
Added a non-public API for setting the tag on an attribute

### LINKED ISSUE #6973: CertificateSigningRequestBuilder.add_attribute doesn't derive attribute ASN.1 type from value bytes
cryptography 37.0.0.dev1
pyOpenSSL 22.0.0

I am trying to make a CSR with a EnrollmentNameValuePair value.
(See https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-wcce/92f07a54-2889-45e3-afd0-94b60daa80ec)

Reproducer:
```
from cryptography import x509
from cryptography.x509.oid import NameOID
from cryptography.hazmat.primitives.serialization import Encoding
from cryptography.x509.oid import NameOID

csr = x509.CertificateSigningRequestBuilder().subject_name(x509.Name([
    x509.NameAttribute(NameOID.COMMON_NAME, "test")
]))

szOID_ENROLLMENT_NAME_VALUE_PAIR = x509.ObjectIdentifier("1.3.6.1.4.1.311.13.2.1")

csr = csr.add_attribute(
    szOID_ENROLLMENT_NAME_VALUE_PAIR,
    bytes.fromhex("302C1E0600530041004E1E2200750070006E003D00410064006D0069006E006900730074007200610074006F0072")
)

key = rsa.generate_private_key(public_exponent=65537, key_size=4096)
cert = csr.sign(key, hashes.SHA256())
bytes = cert.public_bytes(Encoding.PEM)
c = x509.load_pem_x509_csr(bytes)

attribute = c.attributes.get_attribute_for_oid(szOID_ENROLLMENT_NAME_VALUE_PAIR)
print(attribute._type)
```

This prints '12' (decimal), for type UTF8String. However, the first byte 0x30 is for SEQUENCE.

The attribute value bytes are an object with the following structure:
```
EnrollmentNameValuePair ::= SEQUENCE {
         name                BMPSTRING,
         value               BMPSTRING
 }  --#public
```

Where 'name' is 'SAN' and 'value' is 'upn=Administrator'

It looks like the type is instantiated here and not changed afterwards:
https://github.com/pyca/cryptography/blob/b0df70cd2deb8ae8d893a4701f3cb27bda04d5de/src/cryptography/x509/base.py#L81-L87

In the CSR, the bytes I provided in the attribute get wrapped in "0C XX", so the bytes are 0C 2E 30 2C 1E..... instead of just 30 2C 1E.....
where "30" is the start of the SEQUENCE.

I would think that the type should be derived from the first byte of the value provided. Please let me know if I'm wrong.

alex:
fixes #6973

### DIFF
--- a/src/cryptography/x509/base.py
+++ b/src/cryptography/x509/base.py
@@ -56,10 +56,12 @@ def _reject_duplicate_extension(
 
 def _reject_duplicate_attribute(
     oid: ObjectIdentifier,
-    attributes: typing.List[typing.Tuple[ObjectIdentifier, bytes]],
+    attributes: typing.List[
+        typing.Tuple[ObjectIdentifier, bytes, typing.Optional[int]]
+    ],
 ) -> None:
     # This is quadratic in the number of attributes
-    for attr_oid, _ in attributes:
+    for attr_oid, _, _ in attributes:
         if attr_oid == oid:
             raise ValueError("This attribute has already been set.")
 
@@ -552,7 +554,9 @@ class CertificateSigningRequestBuilder:
         self,
         subject_name: typing.Optional[Name] = None,
         extensions: typing.List[Extension[ExtensionType]] = [],
-        attributes: typing.List[typing.Tuple[ObjectIdentifier, bytes]] = [],
+        attributes: typing.List[
+            typing.Tuple[ObjectIdentifier, bytes, typing.Optional[int]]
+        ] = [],
     ):
         """
         Creates an empty X.509 certificate request (v1).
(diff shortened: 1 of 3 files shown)


src/cryptography/x509/base.py
```diff
@@ -552,7 +554,9 @@ def __init__(
         self,
         subject_name: typing.Optional[Name] = None,
         extensions: typing.List[Extension[ExtensionType]] = [],
-        attributes: typing.List[typing.Tuple[ObjectIdentifier, bytes]] = [],
+        attributes: typing.List[
+            typing.Tuple[ObjectIdentifier, bytes, typing.Optional[int]]
```
reaperhulk, line 558: Why an integer and not an element from _ASN1Type like we did in NameAttribute?

alex: Oh, sure.

alex: actually on closer review on `Attribute` it's just an `int` from that enum

alex: Would you still like me to switch to ASN1Type?

reaperhulk: Yeah we extract the int from it but I’d still rather use the enum for the API

reaperhulk (approved):

orangepizza:
even with this it can't support octetstring in attribute, right?

PR #7047, opened 2022-04-06
Added Ubuntu Jammy to CI

alex:


rossburton:
As a distro maintainer who often needs to dig into upstream repositories and find fixes for specific issues, can I just say that "try thing" isn't really explaining the problem or the solution. The commit links to the MR which links to the issue, but that's quite a chain to pick apart if someone is looking at the git history to see if an issue has been fixed.

PR #7052, opened 2022-04-08
Added CentOS Stream 9 to CI

alex:


alex:
@tiran it looks like SHA-1 is disallowed on RSA signatures. How can we test for this to make our tests work reliably.

tiran:
> @tiran it looks like SHA-1 is disallowed on RSA signatures. How can we test for this to make our tests work reliably.

SHA-1 signatures are disabled by a downstream patch, https://gitlab.com/redhat/centos-stream/rpms/openssl/-/blob/c9s/0049-Selectively-disallow-SHA1-signatures.patch

The patch adds ``ossl_ctx_legacy_digest_signatures_allowed`` API. @simo5 should know if the API is part of Red Hat's public, stable ABI.

simo5:
> > @tiran it looks like SHA-1 is disallowed on RSA signatures. How can we test for this to make our tests work reliably.
> 
> SHA-1 signatures are disabled by a downstream patch, https://gitlab.com/redhat/centos-stream/rpms/openssl/-/blob/c9s/0049-Selectively-disallow-SHA1-signatures.patch
> 
> The patch adds `ossl_ctx_legacy_digest_signatures_allowed` API. @simo5 should know if the API is part of Red Hat's public, stable ABI.

It is not and should not be used except in exceptional circumstances, it is an internal OpenSSL function with no guarantees of ABI stability or even existence to be frank.
I am afraid the only sane way to test for this right now is to try to verify a known signature and see if it works.

alex:
Closing for now since this is non-trivial and I'm not going to complete this soon.

PR #7055, opened 2022-04-09
Implement a parser for RFC4514 strings

### LINKED ISSUE #6050: [Feature request] Parse DN string into list of NameAttribute's
There seems to be no way to convert a string (or bytestring) containing a Distinguished Name or X.509 Subject into a list of corresponding NameAttributes, suitable for (example) passing to `x509.CertificateSigningRequestBuilder().subject_name()`.

There's the `rfc4154_string()` which goes in the opposite direction, however.

Using Python 3.8.2 on MacOS 11.3.1, `cryptography` 3.4.7 (installed via `pip3`).

alex:
Based on test cases from #6166 and my own read of the RFC.

Closes #6166
Fixes #6050

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -1225,6 +1225,21 @@ X.509 CSR (Certificate Signing Request) Builder Object
 
         :type: list of :class:`RelativeDistinguishedName`
 
+    .. method:: from_rfc4514_string(data)
+
+        :classmethod:
+
+        :param data: An :rfc:`4514` string.
+
+        :returns: A :class:`Name` parsed from ``data``.
+
+        .. versionadded: 37.0
+
+        .. doctest::
+
+            >>> Name.from_rfc4514_string("CN=cryptography.io")
+            <Name(CN=cryptography.io)>
+
     .. method:: get_attributes_for_oid(oid)
 
         :param oid: An :class:`ObjectIdentifier` instance.
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -1229,7 +1229,7 @@ X.509 CSR (Certificate Signing Request) Builder Object
 
         :classmethod:
 
-        :param data: An :rfc:`4514` string.
+        :param str data: An :rfc:`4514` string.
 
         :returns: A :class:`Name` parsed from ``data``.
 
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -1237,7 +1237,7 @@ X.509 CSR (Certificate Signing Request) Builder Object
 
         .. doctest::
 
-            >>> Name.from_rfc4514_string("CN=cryptography.io")
+            >>> x509.Name.from_rfc4514_string("CN=cryptography.io")
             <Name(CN=cryptography.io)>
 
     .. method:: get_attributes_for_oid(oid)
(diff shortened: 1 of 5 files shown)


docs/x509/reference.rst
```diff
@@ -1225,6 +1225,21 @@ X.509 CSR (Certificate Signing Request) Builder Object
 
         :type: list of :class:`RelativeDistinguishedName`
 
+    .. method:: from_rfc4514_string(data)
+
+        :classmethod:
+
+        :param data: An :rfc:`4514` string.
```
reaperhulk, line 1232:
```suggestion
        :param str data: An :rfc:`4514` string.
```

tests/x509/test_name.py
```diff
@@ -0,0 +1,151 @@
...
+            "C=US,CN=Joe , Smith,DC=example",
+            ",C=US,CN=Joe , Smith,DC=example",
+            "C=US,UNKNOWN=Joe , Smith,DC=example",
+            "C=US,CN,DC=example",
+            "C=US,FOOBAR=example",
+        ]:
+            with pytest.raises(ValueError):
+                Name.from_rfc4514_string(value)
+
+    def test_valid(self):
```
reaperhulk, line 29: this should probably use subtests

tests/x509/test_name.py
```diff
@@ -0,0 +1,151 @@
...
+
+class TestRFC4514:
+    def test_invalid(self):
+        for value in [
+            "C=US,CN=Joe , Smith,DC=example",
+            ",C=US,CN=Joe , Smith,DC=example",
+            "C=US,UNKNOWN=Joe , Smith,DC=example",
+            "C=US,CN,DC=example",
+            "C=US,FOOBAR=example",
+        ]:
```
reaperhulk, line 25: this should also use subtests

reaperhulk (commented):
(not a comprehensive review yet)

src/cryptography/x509/name.py
```diff
@@ -325,3 +359,92 @@ def __len__(self) -> int:
     def __repr__(self) -> str:
         rdns = ",".join(attr.rfc4514_string() for attr in self._attributes)
         return "<Name({})>".format(rdns)
+
+
+class _RFC4514NameParser:
+    _OID_RE = re.compile(r"(\d|[1-9]\d+)(\.\d|[1-9]\d+)+")
```
reaperhulk, line 365: This regex will match invalid OIDs -- do we care?

reaperhulk, line 365: This regex will match invalid OIDs -- do we care?

alex: Invalid how? The results are then immediately passed to the OID constructor which will reject them. Want me to add a test?

reaperhulk: OID constructor rejection is fine.

reaperhulk (approved):

intgr:
Thanks!

Something is not right with processing escape sequences in this implementation though. You missed the test cases I pointed out in https://github.com/pyca/cryptography/pull/6166#discussion_r739105362

Behavior with current merged implementation:

```python
>>> from cryptography.x509 import Name, NameAttribute, NameOID
>>> Name([NameAttribute(NameOID.COMMON_NAME, "\\123")])
<Name(CN=\\123)>
>>> Name.from_rfc4514_string(_.rfc4514_string())
<Name(CN=3)>
>>> Name([NameAttribute(NameOID.COMMON_NAME, "\\#123")])
<Name(CN=\\#123)>
>>> Name.from_rfc4514_string(_.rfc4514_string())
<Name(CN=\#123)>
```

intgr:
I reported as a new issue #7081 just in case to make sure this doesn't slip through the cracks.

PR #7063, opened 2022-04-12
Pin josepy to 1.13.0 in downstream tests

adferrand:
Hello dear maintainers of cryptography. We are currently migrating the Josepy project to Poetry, in order to take advantage of the features it provides over pip (venv management, dependency tree resolution, lock mechanism ...) for developers.

The PR for this is https://github.com/certbot/josepy/pull/129

However before merging this PR, we need to do something about the downstream test in cryptography that uses josepy. Indeed once josepy  master branch is updated, the test `.github/downstream.d/certbot-josepy.sh` will crash because of the lack of support (until recently) of editable installs in pip for PEP-518 compliant projects + the removal of `constraints.txt` file.

This PR fixes this by pinning the current git clone of josepy to its latest release at this time, `1.13.0`.

I propose that once josepy has been updated and starts to use poetry, I create a second PR in cryptography to install and use Poetry, so that it re-establishes the current downstream test behavior (testing master branch with upstream dependencies pinning).

Does it sound good to you? Thanks in advance!

PR #7069, opened 2022-04-14
Use poetry directly to test josepy

adferrand:
Following https://github.com/pyca/cryptography/pull/7063, and since https://github.com/certbot/josepy/pull/129 has been merged on Josepy upstream, this PR adapts the downstream certbot-josepy.sh test to take advantage of Poetry and install the dependencies + dev dependencies before running the tests.

While reviewing the code, I realized that the test was still launching `pytest src`, but the unit tests have been moved out of the src folder, so the actual code is only running the flake8 linting, which is certainly not what we want to do here. I fixed that by running `pytest tests` instead.

### DIFF
--- a/.github/downstream.d/certbot-josepy.sh
+++ b/.github/downstream.d/certbot-josepy.sh
@@ -6,11 +6,12 @@ case "${1}" in
         cd josepy
         git rev-parse HEAD
         curl -sSL https://install.python-poetry.org | python3 -
-        "${HOME}/.local/bin/poetry" install
+        "${HOME}/.local/bin/poetry" export -f requirements.txt --dev --without-hashes > constraints.txt
+        pip install -e . pytest -c constraints.txt
         ;;
     run)
         cd josepy
-        .venv/bin/pytest tests
+        pytest tests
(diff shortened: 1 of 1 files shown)


alex (dismissed):

alex:
is there any way to verify that tests are actually being run with cryptography built from source? I'm concerned that we're not actually installing cryptography from source into the poetry venv.

adferrand:
Indeed we are not! I missed the full picture here, which is to forcibly reinstall the locally build version of cryptography between the install and run phases, which suppose to work on the same python environment.

Poetry is designed to always work in isolation with dedicated environments, so we are testing with the latest released version of cryptography.

Let me propose a different approach.

alex:
Thank you!

adferrand:
You are welcome. So now poetry is used to generate the constraints.txt file only, then pip takes over as before to do the actual installation of josepy + pytest following the constraints provided upstream.

alex (approved):

alex:
Sigh, circleci didn't trigger. Let's try poking.

PR #7071, opened 2022-04-14
version isn't required in wheel builder

alex:


alex:
Sample build: https://github.com/pyca/cryptography/actions/runs/2169859034

reaperhulk (approved):

PR #7082, opened 2022-04-16
Add support for PSS.AUTO and PSS.DIGEST_LENGTH

reaperhulk:
PSS.AUTO allows for verification of signatures where the salt length is not known a priori.

PSS.DIGEST_LENGTH is a convenience constant to make it easier to match the more common PSS conventions in use today (salt length equal to PSS digest length).

### DIFF
--- a/src/_cffi_src/openssl/rsa.py
+++ b/src/_cffi_src/openssl/rsa.py
@@ -45,7 +45,8 @@ int EVP_PKEY_CTX_set_rsa_oaep_md(EVP_PKEY_CTX *, EVP_MD *);
 """
 
 CUSTOMIZATIONS = """
-// BoringSSL doesn't define this constant
+// BoringSSL doesn't define this constant, but the value is used for
+// automatic salt length computation as in OpenSSL and LibreSSL
 #if !defined(RSA_PSS_SALTLEN_AUTO)
 #define RSA_PSS_SALTLEN_AUTO -2
 #endif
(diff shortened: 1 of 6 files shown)


src/cryptography/hazmat/backends/openssl/rsa.py
```diff
@@ -39,14 +41,23 @@
...
     if isinstance(salt, _MaxLength):
         return calculate_max_pss_salt_length(key, hash_algorithm)
+    elif isinstance(salt, _DigestLength):
+        return hash_algorithm.digest_size
+    elif isinstance(salt, _Auto):
+        if isinstance(key, RSAPrivateKey):
+            raise ValueError(
+                "PSS salt length can only be set to AUTO when verifying"
+            )
+        return backend._lib.RSA_PSS_SALTLEN_AUTO  # this is really just -2
```
alex, line 60: this comment is not helpful

src/_cffi_src/openssl/rsa.py
```diff
@@ -44,4 +45,8 @@
 """
 
 CUSTOMIZATIONS = """
+// BoringSSL doesn't define this constant
```
alex, line 48: But I assume passing -2 does work? Please indicate this.

src/cryptography/hazmat/backends/openssl/rsa.py
```diff
@@ -39,14 +41,23 @@
 
 
 def _get_rsa_pss_salt_length(
+    backend: "Backend",
     pss: PSS,
     key: typing.Union[RSAPrivateKey, RSAPublicKey],
     hash_algorithm: hashes.HashAlgorithm,
-) -> typing.Union[int, _MaxLength]:
+) -> typing.Union[int, _MaxLength, _Auto, _DigestLength]:
```
alex, line 48: This return type is wrong, it returns `int` always right?

alex (approved):

PR #7093, opened 2022-04-19
fixes #6681 -- document that we now link against 3.0.2

### LINKED ISSUE #6681: Link against OpenSSL 3
It's probably around time to do this. Sticking in the 37 milestone, although we may choose to bump to 38.

alex:


reaperhulk (approved):

reaperhulk:
And 7 days from now we'll change this line 😂

PR #7108, opened 2022-04-22
Be clear that OpenSSL 1.1.0 is going bye bye

alex:


reaperhulk (approved):

reaperhulk:
You should change the message here too. https://github.com/pyca/cryptography/blob/main/src/cryptography/hazmat/bindings/openssl/binding.py#L197

Issue #7105, opened 2022-04-22
Allow Fernet.decrypt() to accept strings


rcoup:
Currently `Fernet.decrypt()` only accepts `bytes`; we could remove some developer friction by also allowing string tokens to be passed:

1. a valid Fernet token as generated by `Fernet.encrypt()` is url-safe base64-encoded
2. so it's definitely not a valid token if `str_token.encode('ascii')` fails
3. when passing a token as part of any JSON payload, stdlib returns strings
4. when passing a token as part of a HTTP header, python web frameworks return string values

Just removes cycles of "TypeError: token must be bytes" error → "add `.encode('ascii')`" → do it again. Not so much in the few places an app is decoding tokens, but I've found it happens repeatedly writing tests and using via REPL.

Even though it's crypto code 🐉 , given the clear base64 encoding of tokens I feel the "be liberal in what you accept" mantra _can_ be robustly applied here to remove friction from the developer experience.

Happy to prepare a patch/tests if you think it's worthwhile addition.

reaperhulk:
I think this is a reasonable argument for allowing str in the fernet and multi fernet APIs, so please do send a PR!

PR #7116, opened 2022-04-26
Allow Fernet decryption to accept string tokens

rcoup:
Fixes https://github.com/pyca/cryptography/issues/7105

Remove some developer friction by allowing string tokens to be passed to Fernet & MultiFernet decryption methods. Because a valid token as generated by `Fernet.encrypt()` is url-safe base64-encoded, a non-ASCII token is definitely invalid.

- when passing a token as part of any JSON payload, stdlib returns strings
- when passing a token as part of a HTTP header, python web frameworks return string values
- it's common to be using strings in test vectors, REPL environments, etc

### DIFF
--- a/tests/test_fernet.py
+++ b/tests/test_fernet.py
@@ -37,7 +37,7 @@ def json_parametrize(keys, filename):
         return pytest.mark.parametrize(
             keys,
             [tuple([entry[k] for k in keys]) for entry in data],
-            ids=_id_gen(),
+            ids=[f"{filename}[{i}] for i in range(len(data))],
         )
 
 
--- a/tests/test_fernet.py
+++ b/tests/test_fernet.py
@@ -5,7 +5,6 @@
 
 import base64
 import calendar
-import itertools
 import json
 import os
 import time
--- a/tests/test_fernet.py
+++ b/tests/test_fernet.py
@@ -5,6 +5,7 @@
 
 import base64
 import calendar
+import itertools
 import json
 import os
 import time
@@ -25,10 +26,18 @@ def json_parametrize(keys, filename):
     vector_file = cryptography_vectors.open_vector_file(
         os.path.join("fernet", filename), "r"
     )
+
+    def _id_gen():
+        # generate readable test identifiers
+        for idx in itertools.count():
+            yield f"{filename}[{idx}]"
+
     with vector_file:
         data = json.load(vector_file)
         return pytest.mark.parametrize(
-            keys, [tuple([entry[k] for k in keys]) for entry in data]
+            keys,
+            [tuple([entry[k] for k in keys]) for entry in data],
+            ids=_id_gen(),
         )
 
 
--- a/tests/test_fernet.py
+++ b/tests/test_fernet.py
@@ -26,12 +26,6 @@ def json_parametrize(keys, filename):
     vector_file = cryptography_vectors.open_vector_file(
         os.path.join("fernet", filename), "r"
     )
-
-    def _id_gen():
-        # generate readable test identifiers
-        for idx in itertools.count():
-            yield f"{filename}[{idx}]"
-
     with vector_file:
         data = json.load(vector_file)
         return pytest.mark.parametrize(
--- a/tests/test_fernet.py
+++ b/tests/test_fernet.py
@@ -68,16 +68,29 @@ class TestFernet:
     def test_verify(
         self, secret, now, src, ttl_sec, token, backend, monkeypatch
     ):
+        # secret & token are both str
         f = Fernet(secret.encode("ascii"), backend=backend)
         current_time = calendar.timegm(iso8601.parse_date(now).utctimetuple())
         payload = f.decrypt_at_time(
-            token.encode("ascii"),
+            token,  # str
             ttl=ttl_sec,
             current_time=current_time,
         )
         assert payload == src.encode("ascii")
+
+        payload = f.decrypt_at_time(
+            token.encode("ascii"),  # bytes
+            ttl=ttl_sec,
+            current_time=current_time,
+        )
+        assert payload == src.encode("ascii")
+
         monkeypatch.setattr(time, "time", lambda: current_time)
-        payload = f.decrypt(token.encode("ascii"), ttl=ttl_sec)
+
+        payload = f.decrypt(token, ttl=ttl_sec)  # str
+        assert payload == src.encode("ascii")
+
+        payload = f.decrypt(token.encode("ascii"), ttl=ttl_sec)  # bytes
         assert payload == src.encode("ascii")
 
     @json_parametrize(("secret", "token", "now", "ttl_sec"), "invalid.json")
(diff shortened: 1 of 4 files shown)


alex:
Thanks for sending this PR! Unfortunately it won't make it for 37.0. Please take a look at the flake8 failures and we'll review shortly.

alex:
Thanks for sending this PR! Unfortunately it won't make it for 37.0. Please take a look at the flake8 failures and we'll review shortly.

rcoup:
Ah, mindless rebases on changelogs 😁  I wasn't intending it to land anywhere in particular.

I've added the 38.0.0 header so at least it's clear in the changelog, but if you're landing that separately it's obviously a trivial merge/rebase.

reaperhulk:
This can be rebased now. Looks like black needs to be run as well

rcoup:
Ah yes, flake8 complaining about things black doesn't fix 😬 

mypy seems... ["particular"](https://github.com/python/mypy/issues/1032) about where `#ignore` comments go (pre-3.8 at least) so I've opted to make flake8 ignore it

tests/test_fernet.py
```diff
@@ -25,10 +26,18 @@ def json_parametrize(keys, filename):
...
+        while True:
+            yield f"{filename}:{next(idx)}"
+
     with vector_file:
         data = json.load(vector_file)
         return pytest.mark.parametrize(
-            keys, [tuple([entry[k] for k in keys]) for entry in data]
+            keys,
+            [tuple([entry[k] for k in keys]) for entry in data],
+            ids=_id_gen(),
```
alex, line 40: What is this for?

tests/test_fernet.py
```diff
@@ -56,19 +65,24 @@ def test_generate(self, secret, now, iv, src, token, backend):
         ("secret", "now", "src", "ttl_sec", "token"),
         "verify.json",
     )
+    @pytest.mark.parametrize("typ", (str, bytes))
```
alex, line 68: I'd prefer if instead of using `parameterize` for types we just wrote a second test case that covered this variant, e.g. as you did in `test_extract_timestamp`.

alex (commented):
Thanks for working on this. Very close!

tests/test_fernet.py
rcoup: commit message:

> Use the vector filename and array index for the pytest id rather than a
concatenation of the vector content.
>
> eg: `tests/test_fernet.py::TestFernet::test_invalid[invalid.json:2]`

was eg:
```
tests/test_fernet.py::TestFernet::test_invalid[gAAAAAAdwJ6wAAECAwQFBgcICQoLDA0ODy021cpGVWKZ_eEwCGM4BLLF_5CV9dOPmrhuVUPgJobwOz7JcbmrR64jVmpU4IwqDA==-1985-10-26T01:20:01-07:00-60-hello-cw_0x689RpI-jtRR7oE8h_eQsKImvJapLeSbXpwF4e4=]
```

tests/test_fernet.py
rcoup: 🤷 cando.

tests/test_fernet.py
```diff
@@ -25,10 +26,18 @@ def json_parametrize(keys, filename):
     vector_file = cryptography_vectors.open_vector_file(
         os.path.join("fernet", filename), "r"
     )
+
+    def _id_gen():
+        idx = itertools.count()
+        while True:
+            yield f"{filename}:{next(idx)}"
```
alex, line 33:
```suggestion
        for idx in itertools.count():
            yield f"{filename}[{idx}]"
```

tests/test_fernet.py
alex: Can you add a comment explaining that "this improves the output for readability" or something like that.

alex (commented):
Two more small comments. Thanks for working on this!

tests/test_fernet.py
```diff
@@ -25,10 +26,18 @@ def json_parametrize(keys, filename):
     vector_file = cryptography_vectors.open_vector_file(
         os.path.join("fernet", filename), "r"
     )
+
+    def _id_gen():
+        # generate readable test identifiers
+        for idx in itertools.count():
+            yield f"{filename}:{next(idx)}"
```
alex, line 33:
```suggestion
            yield f"{filename}[{idx}]"
```

`next` isn't needed since we're using a `for` loop, and using `[]` makes it clear it's not a line number -- `:` is often used for lines.

alex (dismissed):

alex:
codecov is up to no good again, going to try close/reopen to fix it.

alex:
@rcoup with tremendous apologies, can you push an empty commit: `git commit --allow-empty -m "Kick CI" && git push`? codecov appears unhappy for no good reason.

alex (dismissed):

tests/test_fernet.py
```diff
@@ -25,10 +26,18 @@ def json_parametrize(keys, filename):
...
+        for idx in itertools.count():
+            yield f"{filename}[{idx}]"
+
     with vector_file:
         data = json.load(vector_file)
         return pytest.mark.parametrize(
-            keys, [tuple([entry[k] for k in keys]) for entry in data]
+            keys,
+            [tuple([entry[k] for k in keys]) for entry in data],
+            ids=_id_gen(),
```
alex, line 40:
```suggestion
            ids=[f"{filename}[{i}] for i in range(len(data))],
```

alex (commented):
Sigh. Ok I see what's going on with coverage:

1) codecov is awful, and I'm sorry for the crappy experience it's causing
2) `coverage.py` doesn't like `_id_gen()`. If I run coverage locally I see the following:

Screen Shot 2022-04-28 at 09 42 15

I've added a suggestion that should solve it.

tests/test_fernet.py
```diff
@@ -25,10 +26,18 @@ def json_parametrize(keys, filename):
     vector_file = cryptography_vectors.open_vector_file(
         os.path.join("fernet", filename), "r"
     )
+
+    def _id_gen():
+        # generate readable test identifiers
+        for idx in itertools.count():
+            yield f"{filename}[{idx}]"
+
```
alex, line 34:
```suggestion
```

tests/test_fernet.py
```diff
@@ -5,6 +5,7 @@
 
 import base64
 import calendar
+import itertools
```
alex, line 8:
```suggestion
```

alex (dismissed):

tests/test_fernet.py
```diff
@@ -28,7 +28,9 @@ def json_parametrize(keys, filename):
     with vector_file:
         data = json.load(vector_file)
         return pytest.mark.parametrize(
-            keys, [tuple([entry[k] for k in keys]) for entry in data]
+            keys,
+            [tuple([entry[k] for k in keys]) for entry in data],
+            ids=[f"{filename}[{i}] for i in range(len(data))],
```
alex, line 33:
```suggestion
            ids=[f"{filename}[{i}]" for i in range(len(data))],
```

alex (approved):

PR #7133, opened 2022-04-27
Add typings to default_backend()

Gaasmann:
This commit adds back a return type for default_backend, so mypy would consider this function typed.


reaperhulk (approved):

reaperhulk:
Thanks for sending the backport PR too 😄

PR #7135, opened 2022-04-27
Fix parsing of priv keys via pub key APIs to error properly in ossl3

reaperhulk:
In OpenSSL 3.0.x the PEM_read_bio_PUBKEY function will invoke the default password callback if you pass an encrypted private key. This is very, very, very bad as the default callback can trigger an interactive console prompt, which will hang the Python process. We therefore provide our own callback to catch this and error out properly.

backport of #7132


alex (approved):

alex:
codecov website shows 100% coverage

PR #7148, opened 2022-04-29
Use new ability for OIDs to be consts

alex:


src/rust/src/x509/oid.rs
```diff
@@ -1,75 +1,101 @@
 // This file is dual licensed under the terms of the Apache License, Version
-// 2.0, and the BSD License. See the LICENSE file in the root of this repository
-// for complete details.
+// 2, 0, and the BSD License,  See the LICENSE file in the root of this repository
```
reaperhulk, line 2: lol

reaperhulk (approved):

PR #7161, opened 2022-04-30
add x509 pss certificate and public key

reaperhulk:
the cert is signed using rsa_pss_2048.pem and the public key is the
public components of that private key


docs/development/test-vectors.rst
```diff
@@ -143,6 +143,8 @@ Custom asymmetric vectors
   contain an Ed448 public key.
 * ``asymmetric/PKCS8/rsa_pss_2048.pem`` - A 2048-bit RSA PSS key with no
   explicit parameters set.
+* ``asymmetric/PKCS8/rsa_pss_2048_pub.pem`` - The public key corresponding to
```
alex, line 146:
```suggestion
* ``asymmetric/PKCS8/rsa_pss_2048_pub.der`` - The public key corresponding to
```

alex (approved):

PR #7164, opened 2022-04-30
Resurrect the PoC of OpenSSL from Rust

alex:
Requirements:
- PyPy 7.3.10
  - [x] Release
  - [x] https://github.com/pypa/manylinux/issues/1430
  - [x] https://github.com/actions/runner-images/pull/6757
  - [x] https://github.com/actions/runner-images/pull/6799
  - [x] https://github.com/actions/runner-images/pull/6784
- Rust 1.66
  - [x] Release
  - [x] https://github.com/actions/runner-images/pull/6799
- [x] Update documentation
- [x] `openssl-sys` release (LibreSSL 3.7.0 support)
- [x] Demo use-case PR (https://github.com/pyca/cryptography/pull/7933)
- [x] https://github.com/sfackler/rust-openssl/issues/1768


alex:
Blocked on https://foss.heptapod.net/pypy/pypy/-/issues/3619

alex:
This should be fixed in the next PyPy release. I'll rebase then and we can contemplate how we want to handle it.

alex:
macOS wheel builds are failing due to https://github.com/rust-lang/rust/issues/55235

The PyPy wheel builds are failing because they aren't using PyPy nightly.

alex:
Ok, assuming tests pass, this PR now requires:

- rust 1.66 (December 15th -- incidentally several friends' of mines birthdays)
- PyPy 3.8 and 3.9 releases

alex:
Some of the jobs are failing because of https://github.com/actions/setup-python/issues/518

docs/installation.rst
```diff
@@ -272,14 +271,14 @@ To build cryptography and dynamically link it:
 .. code-block:: console
 
     $ brew install openssl@3 rust
-    $ env LDFLAGS="-L$(brew --prefix openssl@3)/lib" CFLAGS="-I$(brew --prefix openssl@3)/include" pip install cryptography
+    $ env OPENSSL_DIR="$(brew --prefix openssl@3) pip install cryptography
```
reaperhulk, line 274:
```suggestion
    $ env OPENSSL_DIR="$(brew --prefix openssl@3)” pip install cryptography
```

alex:
All these warnings are coming from cffi already -- none of them are new warnings from this branch.

Tarnum-tst:
@alex 
When compiling cffi build I got only two warnings, similar to these:
```
warning: C:\Python37\crypmain\src\rust\target\i686-pc-windows-gnu\release\build\cryptography-rust-74aa9bb878ed6b19\out\_openssl.c: At top level:
warning: C:\Python37\crypmain\src\rust\target\i686-pc-windows-gnu\release\build\cryptography-rust-74aa9bb878ed6b19\out\_openssl.c:784:13: warning: 'ERR_load_Cryptography_OSRandom_strings' declared 'static' but never defined [-Wunused-function]
warning:   784 | static void ERR_load_Cryptography_OSRandom_strings(void);
warning:       |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
warning: C:\Python37\crypmain\src\rust\target\i686-pc-windows-gnu\release\build\cryptography-rust-74aa9bb878ed6b19\out\_openssl.c:785:13: warning: 'ERR_Cryptography_OSRandom_error' declared 'static' but never defined [-Wunused-function]
warning:   785 | static void ERR_Cryptography_OSRandom_error(int function, int reason,
warning:       |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

reaperhulk:
If you believe you have an actual issue please file it in a new issue and describe how to reproduce it. The statements in this PR not only don’t belong here but also appear to be inaccurate. For example, we test against a dynamically linked openssl 3.2b1 in our CI right now and the warnings you’re seeing come directly from cffi generated source (the directory structure, which may have led you to believe it was rust, is not relevant).

Compiling this project (especially with custom openssl) requires a significant amount of knowledge about platform linking behavior, symbol namespaces, etc and we cannot support or help users with every possible configuration.

Tarnum-tst:
@reaperhulk 
No, I don't have an actual issue, I just wanted to give a feedback on this PR according to my experience. :-)
And You are right, I was not fully correct about dynamically linked build. It builds fine but doesn't work for some reason with libcrypto-3.dll/libssl-3.dll loaded by Python's _ssl.pyd because of missing functions ("ImportError: DLL load failed: The specified procedure could not be found."). Dependency walker on (dynamically linked) _rust.pyd shows the absence of some calls related to OpenSSL 3.2 dev. But because static build works just fine on my system, I don't really want to dig into what's wrong with dynamically linked build.
@alex 
Thank You for this PR!

alex:
ok nevermind, solved that problem.

.github/workflows/wheel-builder.yml
```diff
@@ -286,8 +285,10 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
       - name: Configure OpenSSL
         run: |
-            echo "INCLUDE=C:/openssl-${{ matrix.WINDOWS.WINDOWS }}/include;$INCLUDE" >> $GITHUB_ENV
-            echo "LIB=C:/openssl-${{ matrix.WINDOWS.WINDOWS }}/lib;$LIB" >> $GITHUB_ENV
+            echo "OPENSSL_DIR=C:/openssl-${{ matrix.WINDOWS.WINDOWS }}" >> $GITHUB_ENV
+            echo "OPENSSL_STATIC=1" >> $GITHUB_ENV
```
reaperhulk, line 289: Should we be passing this in the windows CI builder as well? I think we only copy over the static libraries on Windows (https://github.com/pyca/infra/blob/main/windows/openssl/build_openssl_win64.bat#L11-L12), so maybe it works without even passing this?

src/_cffi_src/build_openssl.py
```diff
@@ -110,3 +111,17 @@ def _extra_compile_args(platform):
     libraries=_get_openssl_libraries(sys.platform),
     extra_compile_args=_extra_compile_args(sys.platform),
 )
+
+if __name__ == "__main__":
+    out_dir = os.getenv("OUT_DIR")
```
reaperhulk, line 116: Does the setuptools build machinery set this value?

tox.ini
```diff
@@ -73,7 +74,6 @@ commands =
 
 [testenv:rust]
 basepython = python3
-skip_install = True
```
reaperhulk, line 76: Do we need to install here now because `cargo test` exercises things that otherwise can't be built?

src/_cffi_src/build_openssl.py
alex: No, the rust build machinery (from which this is called) does

.github/workflows/wheel-builder.yml
alex: Yeah, if only static is available it'll default to it.

But either way, it doesn't _really_ matter in CI, it's only critical for wheels.

tox.ini
alex: correct -- building the rust package now involves building our cffi code

.github/workflows/wheel-builder.yml
```diff
@@ -286,8 +285,10 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
reaperhulk, line 291: Merge issue?

src/rust/build.rs
```diff
@@ -0,0 +1,145 @@
...
+        .file(openssl_c)
+        .include(python_include)
+        .include(openssl_include)
+        .flag_if_supported("-Wconversion")
+        .flag_if_supported("-Wno-error=sign-conversion");
+
+    // Enable abi3 mode if we're not using PyPy.
+    if python_impl != "PyPy" {
+        // cp36
+        build.define("Py_LIMITED_API", "0x030600f0");
```
reaperhulk, line 79: So we have to keep this synced with our wheel builder flags? Presumably there's no way to have it detect the right value like setuptools-rust or maturin do since there's no guarantee this is called with that machinery?

alex: Honestly not sure, I don't know where this invocation lives in current code.

reaperhulk (approved):

reaperhulk (approved):
💥

PR #7178, opened 2022-05-02
switch to AES-256-CBC by default for encrypted serialization of PKCS12

### LINKED ISSUE #7043: AES256 support for PKCS#12
I came across this piece of code in the openssl backend:
https://github.com/pyca/cryptography/blob/4a4f4d94ce5a641de3020042c70c1734af265d5e/src/cryptography/hazmat/backends/openssl/backend.py#L2194-L2202

It's part of [what gets called when creating a PKXS#12 file](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/?highlight=pkcs12#cryptography.hazmat.primitives.serialization.pkcs12.serialize_key_and_certificates) and It uses 3DES for the encryption.

**Is there any way to switch this to something like AES256?**

When exporting to a PFX in windows 10, you can do this [as mentioned here](https://dirteam.com/sander/2021/02/09/pfx-encryption-security-feature-bypass-vulnerability-cve-2021-1731-important/)

Also, you can make such PFX file though openSSL [as mentioned here](https://www.openssl.org/docs/man3.0/man1/openssl-pkcs12.html) (and [here for v1.1.1](https://www.openssl.org/docs/man1.1.1/man1/openssl-pkcs12.html)):

```
C:\>openssl pkcs12 -export -in cert.pem-inkey private.key -out some.pfx -certpbe AES-256-CBC -keypbe AES-256-CBC
Enter pass phrase for key_private_pem.key:
Enter Export Password:
Verifying - Enter Export Password:

C:\>openssl pkcs12 -noout -info -in some.pfx
Enter Import Password:
MAC: sha1, Iteration 2048
MAC length: 20, salt length: 8
PKCS7 Encrypted data: PBES2, PBKDF2, AES-256-CBC, Iteration 2048, PRF hmacWithSHA256
Certificate bag
PKCS7 Data
Shrouded Keybag: PBES2, PBKDF2, AES-256-CBC, Iteration 2048, PRF hmacWithSHA256
```

reaperhulk:
Add fallback LegacyPKCS12TripleDESEncryption for compatibility with The Past(tm)

fixes #7043

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,12 @@ Changelog
 
 * Final deprecation of OpenSSL 1.1.0. The next release of ``cryptography``
   will drop support.
+* :class:`~cryptography.hazmat.primitives.serialization.BestAvailableEncryption`
+  has been updated to use AES 256 CBC by default on PKCS12 files. Users needing
+  compatibility with older Windows (2016 or earlier) and Java (8) versions may
+  use
+  :class:`~cryptography.hazmat.primitives.serialization.LegacyPKCS12TripleDESEncryption`
+  .
 * We no longer ship ``manylinux2010`` wheels. Users should upgrade to the
   latest ``pip`` to ensure this doesn't cause issues downloading wheels on
   their platform.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,8 +11,8 @@ Changelog
 * Final deprecation of OpenSSL 1.1.0. The next release of ``cryptography``
   will drop support.
 * :class:`~cryptography.hazmat.primitives.serialization.BestAvailableEncryption`
-  has been updated to use AES 256 CBC by default on PKCS12 files. Users needing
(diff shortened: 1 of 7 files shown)


alex (approved):

alex (approved):

reaperhulk:
Looks like boringssl needs to implement the weird NID mapping to PBES2 choice that OpenSSL added if we want to support this. See: https://boringssl.googlesource.com/boringssl/+/refs/heads/master/crypto/pkcs8/pkcs8.c#492

@davidben is there another way to use PBES2 + AES-256-CBC?

davidben:
It's been a while since I'd looked at this, so I don't remember the details (and am currently sick so limited time to dig into it). I think that TODO was about switching the KDF from hmacWithSHA1. Looks like you all are looking to do something about the encryption part?

But anyway, I probably only implemented the subset that anyone (you all, I imagine :-) ) were using at the time, because it was simplest, not so much because we didn't want that mechanism. If you need that feature, happy to take a patch for it.

PR #7180, opened 2022-05-02
Upgrade actions/upload-artifact now that we don't have manylinux2010

alex:


alex:
https://github.com/pyca/cryptography/actions/runs/2260916852

reaperhulk (approved):

PR #7207, opened 2022-05-10
x509/CT: expose more SCT internals

woodruffw:
This PR exposes some more fields within each signed `SignedCertificateTimestamp`:

* `hash_algorithm`: The hash used for the signature
* `signature_algorithm`: The signing algorithm used for the signature
* `signature`: The actual signature bytes

PR #7215, opened 2022-05-12
docs, vectors: add two more x509 SCT testvectors

woodruffw:
Signed-off-by: William Woodruff <william@yossarian.net>


woodruffw:
cc @reaperhulk

alex (commented):
are you sure these encodings are correct? when parsing with the openssl CLI it seems to result in openssl not recognizing the extension at all

woodruffw:
> are you sure these encodings are correct? when parsing with the openssl CLI it seems to result in openssl not recognizing the extension at all

Paul and I tried to do some root-causing of this yesterday. I'm not 100% sure yet, but I think OpenSSL's printing gives up entirely because it doesn't have a NID for any pair of `(none, ...)` or `(..., anonymous)` for the `(hash, sig_algo)` pair. 

(I don't have an easy way to verify that though, since I did the binary patch by hand. But from the unit tests in #7207, the exceptions that I expect to be raised are indeed being raised, so I'm *decently* confident that I twiddled the right bytes.)

woodruffw:
I _think_ what's going on here is this:

https://github.com/openssl/openssl/blob/1c0eede9827b0962f1d752fa4ab5d436fa039da4/crypto/x509/v3_prn.c#L179-L202

Which we get to probably because of this branch:

https://github.com/openssl/openssl/blob/1c0eede9827b0962f1d752fa4ab5d436fa039da4/crypto/x509/v3_prn.c#L93-L94

In other words: we have a valid DER structure for the SCT, but OpenSSL doesn't know what to make of the parameters (since they're invalid).

We get there through an obscene amount of indirection, but it boils down to this check in `o2i_SCT_signature`: https://github.com/openssl/openssl/blob/1c0eede9827b0962f1d752fa4ab5d436fa039da4/crypto/ct/ct_oct.c#L50-L53

Which fails for the aforementioned reason: there's no NIDs for any pairs of `hash=none` or `sig=anonymous`: https://github.com/openssl/openssl/blob/1c0eede9827b0962f1d752fa4ab5d436fa039da4/crypto/ct/ct_sct.c#L216-L231

alex (approved):

PR #7216, opened 2022-05-12
Fix test CRLs and CSRs with invalid versions

davidben:
CRL versions only go up to v2 (numeric value 1) and CSRs only define v1 (numeric value 0). See https://www.rfc-editor.org/rfc/rfc5280.html#section-5.1 and https://www.rfc-editor.org/rfc/rfc2986.html#section-4. Some of the inputs used in tests were misissued.

Unfortunately, the corresponding private keys for these test vectors weren't provided, so I've just edited the inputs with der-ascii and left the signatures invalid. The tests in question don't seem to rely on valid signatures.


reaperhulk:
How did we end up with so many requests with an invalid version. LGTM, but codecov is doing its thing so I'm going to close/reopen this to trigger some CI again.

reaperhulk (approved):

PR #7237, opened 2022-05-19
CT: add `SignedCertificateTimestamp.extensions`

woodruffw:
Exposes yet another field on the SCT.

`extensions` is currently unspecified for v1 SCTs, but not expressly forbidden. So we need it exposed in order to verify digitally-signed element signatures.

Signed-off-by: William Woodruff <william@trailofbits.com>

### DIFF
--- a/docs/x509/certificate-transparency.rst
+++ b/docs/x509/certificate-transparency.rst
@@ -76,6 +76,14 @@ issued.
 
         The raw bytes of the signatures embedded in the SCT.
 
+    .. attribute:: extensions
+
+        .. versionadded:: 38.0
+
+        :type: bytes
+
+        Any raw extension bytes.
+
 
 .. class:: Version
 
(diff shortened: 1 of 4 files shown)


reaperhulk (approved):

reaperhulk:
As a sanity check we verified that Chrome does also pass in the (presumably empty in all cases right now) bytes: https://source.chromium.org/chromium/chromium/src/+/main:net/cert/ct_serialization.cc;l=258-284;drc=8a609a649e36c8099c6801d41aca1ec3b0b3af73

alex:
Am I reading the RFC correctly that there's no structure even defined for
these bytes?

בתאריך יום ה׳, 19 במאי 2022, 13:25, מאת William Woodruff ‏<
***@***.***>:

woodruffw:
That's my reading of the RFC as well. I believe they're just an `opaque<1..2^16-1>`.

alex:
For the record, note the sent date:

Screen Shot 2022-10-11 at 08 35 17

Github lost its mind.

woodruffw:
lol. 

Eventual consistency!

PR #7267, opened 2022-05-25
Add `tbs_precertificate_bytes` property

tetsuo-cpp:
- [ ] Unit tests
- [ ] Docs

This is required for verifying embedded SCTs in `sigstore-python`. According to IETF's RFC 6962, this data is part of the digitally signed payload that forms the SCT's signature. This still needs unit tests.

CC: @woodruffw @di

### DIFF
--- a/src/cryptography/x509/base.py
+++ b/src/cryptography/x509/base.py
@@ -229,6 +229,12 @@ class Certificate(metaclass=abc.ABCMeta):
         Returns the tbsCertificate payload bytes as defined in RFC 5280.
         """
 
+    @abc.abstractproperty
+    def tbs_precertificate_bytes(self) -> bytes:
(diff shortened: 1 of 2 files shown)


alex:
Thanks for contributing! In addition to tests, we'll also need documentation for that.

I haven't done a full review, but a basic question: Should it be an error to try to get the precert bytes for a certificate that doesn't have the PreCert SCT extension?

woodruffw:
> Should it be an error to try to get the precert bytes for a certificate that doesn't have the PreCert SCT extension?

IMO, yes. I think we should probably raise a `ValueError` for that.

woodruffw:
#7279 supersedes this. @alex / @reaperhulk, feel free to close.

PR #7271, opened 2022-05-26
remove NIST AES GCM vectors from test

reaperhulk:
We cover all relevant cases through wycheproof and this removes 50k tests.

One issue: wycheproof tests are run in our CI but *not* by downstreams.

Marked as draft so we can experiment and see how much faster CI is.


reaperhulk:
This is legitimately failing due to lack of coverage on the test_aead utils helpers but the shell scripts need improvement for outputting the failing information in the markdown summary. #7269 should resolve this (and then I can rebase this to see it)

reaperhulk:
This isn't enough faster to be worth gaining full confidence that it's safe.

PR #7279, opened 2022-05-27
X.509/Certificate: Add `tbs_precertificate_bytes` property

woodruffw:
This supersedes #7267, which I can't push to.

This is required for verifying embedded SCTs in sigstore-python. According to IETF's RFC 6962, this data is part of the digitally signed payload that forms the SCT's signature. This still needs unit tests.


tests/x509/test_x509_ext.py
```diff
@@ -5769,6 +5769,15 @@ def test_public_bytes(self):
...
 
 class TestSignedCertificateTimestamps:
+    def test_strip_scts(self, backend):
+        cert = _load_cert(
+            os.path.join("x509", "cryptography-scts.pem"),
+            x509.load_pem_x509_certificate,
+            backend,
+        )
+
+        assert cert.tbs_certificate_bytes != cert.tbs_precertificate_bytes
```
woodruffw, line 5779: Flagging for review: this "test" really only checks that we return something different than the normal `tbsCertificate` body, since we know that this cert has an SCT list extension.

Ideally we'd do something more behavioral like refute the presence of the extension in the `tbsCertificate`, but I don't think `cryptography` has any APIs to load *just* the `tbsCertificate` format outside of its X.509 envelope. Thoughts? Alternatively, I could add a known good hex string here, since we can confirm on `sigstore-python` that this helper is working as expected.

src/rust/src/x509/certificate.rs
```diff
@@ -216,6 +216,27 @@ impl Certificate {
...
 
+    #[getter]
+    fn tbs_precertificate_bytes<'p>(
+        &self,
+        py: pyo3::Python<'p>,
+    ) -> Result<&'p pyo3::types::PyBytes, PyAsn1Error> {
+        let val = self.raw.borrow_value();
+        let mut tbs_precert = val.tbs_cert.clone();
+        // Remove the SCT list extension
+        if let Some(extensions) = tbs_precert.extensions {
```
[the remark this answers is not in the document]
woodruffw: Awesome. In that case the behavioral test for "normal" certs will change as well, which is good 🙂

woodruffw:
> I don't believe this is correct. This OID is for use in EKUs, isn't it?

Whoops, you're absolutely right. I'll move this under the right class.

tests/x509/test_x509_ext.py
reaperhulk: Known good value comparison would be preferable.

woodruffw: Sounds good. @tetsuo-cpp, do you happen to have any of the outputs you produced from experimenting with `certificate-transparency-go`? Those would probably make for a good testvector candidates.

tetsuo-cpp: Yep, I can do this.

src/rust/src/x509/oid.rs
```diff
@@ -10,6 +10,8 @@ pub(crate) const PRECERT_SIGNED_CERTIFICATE_TIMESTAMPS_OID: asn1::ObjectIdentifi
     asn1::oid!(1, 3, 6, 1, 4, 1, 11129, 2, 4, 2);
 pub(crate) const PRECERT_POISON_OID: asn1::ObjectIdentifier =
     asn1::oid!(1, 3, 6, 1, 4, 1, 11129, 2, 4, 3);
+pub(crate) const CERTIFICATE_TRANSPARENCY_OID: asn1::ObjectIdentifier =
+    asn1::oid!(1, 3, 6, 1, 4, 1, 11129, 2, 4, 4);
```
alex, line 14: This isn't used, so it's not needed.

tests/x509/test_x509.py
```diff
@@ -913,6 +913,34 @@ def test_tbs_certificate_bytes(self, backend):
...
+
+        public_key = cert.public_key()
+        assert isinstance(public_key, rsa.RSAPublicKey)
+        assert cert.signature_hash_algorithm is not None
+        public_key.verify(
+            cert.signature,
+            cert.tbs_certificate_bytes,
+            padding.PKCS1v15(),
+            cert.signature_hash_algorithm,
+        )
```
alex, line 942: What is this testing?

tests/x509/test_x509_ext.py
```diff
@@ -5769,6 +5769,22 @@ def test_public_bytes(self):
...
+
+        expected_tbs_precertificate_bytes = load_vectors_from_file(
+            filename=os.path.join("x509", "cryptography-scts-tbs-precert.der"),
+            loader=lambda data: data.read(),
+            mode="rb",
+        )
+        assert (
+            expected_tbs_precertificate_bytes == cert.tbs_precertificate_bytes
+        )
+
```
alex, line 5787: This should be with the certificate tests -- this test class is for testing the SCTs class.

docs/x509/reference.rst
```diff
@@ -462,6 +462,24 @@ X.509 Certificate Object
        An :class:`~cryptography.exceptions.InvalidSignature` exception will be
        raised if the signature fails to verify.
 
+
+    .. attribute:: tbs_precertificate_bytes
+
+        .. versionadded:: 38.0
+
+        :type: bytes
+
```
alex, line 471: The docs should provide more context here, imagine a user who isn't familiar with CT reading this. It doesn't need to explain all of CT, but it does need to provide enough information for them to figure out if this is relevant to them.

src/cryptography/hazmat/_oid.py
```diff
@@ -155,6 +155,7 @@ class ExtendedKeyUsageOID:
     SMARTCARD_LOGON = ObjectIdentifier("1.3.6.1.4.1.311.20.2.2")
     KERBEROS_PKINIT_KDC = ObjectIdentifier("1.3.6.1.5.2.3.5")
     IPSEC_IKE = ObjectIdentifier("1.3.6.1.5.5.7.3.17")
+    CERTIFICATE_TRANSPARENCY = ObjectIdentifier("1.3.6.1.4.1.11129.2.4.4")
```
alex, line 158: Please document this.

alex (commented):
You'll also need to add docs for the test vector.

tests/x509/test_x509.py
woodruffw: Previously, this was testing that we didn't modify certs that weren't precerts. But we're changing the implementation to raise a `ValueError` when the extension is missing, so this will become a `pytest.raises` check instead.

src/rust/src/x509/certificate.rs
```diff
@@ -216,6 +216,39 @@ impl Certificate {
...
+        match tbs_precert.extensions {
+            Some(extensions) => {
+                let readable_extensions = extensions.unwrap_read().clone();
+                let ext_count = readable_extensions.len();
+                let filtered_extensions: Vec<x509::common::Extension<'_>> = readable_extensions
+                    .filter(|x| x.extn_id != oid::PRECERT_SIGNED_CERTIFICATE_TIMESTAMPS_OID)
+                    .collect();
+                if filtered_extensions.len() == ext_count {
+                    return Err(PyAsn1Error::from(pyo3::exceptions::PyValueError::new_err(
+                        "Could not find SCT list extension in TBS certificate",
```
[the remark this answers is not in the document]
alex: Still outstanding

tests/x509/test_x509.py
alex: Pretty sure this can just be deleted, it's testing `tbs_certificate_bytes`, not the precertificate bytes.

src/rust/src/x509/certificate.rs
```diff
@@ -233,7 +233,7 @@ impl Certificate {
                     .collect();
                 if filtered_extensions.len() == ext_count {
                     return Err(PyAsn1Error::from(pyo3::exceptions::PyValueError::new_err(
-                        "Could not find SCT list extension in TBS certificate",
+                        "Could not find SCT list extension in TBS precertificate",
```
alex, line 236:
```suggestion
                        "Could not find pre-certificate SCT list extension",
```

Here's my alternate suggestion

tests/x509/test_x509_ext.py
```diff
@@ -6170,6 +6154,23 @@ def test_public_bytes(self, backend):
             b"\x85\x1f%\xc0#N\\\xdeK\x90d\xe0\x8d<{\xca\xdbdc\xeft\x87g\x10"
         )
 
+    def test_strip_scts(self, backend):
```
alex, line 6157: The name should mention `tbs_precertificate_bytes` and probably should be next to the other test

tests/x509/test_x509_ext.py
woodruffw: Done, I think -- it's now under the certificate extension tests for the extension that's being stripped out in `tbs_precertificate_bytes`.

tests/x509/test_x509_ext.py
woodruffw: Okay, moved.

src/rust/src/x509/certificate.rs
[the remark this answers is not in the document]
woodruffw: Updated with your recommendation (and updated the corresponding test match).

alex:
You'll need to rebase to fix the alpine issue.

tests/x509/test_x509_ext.py
woodruffw: Done, and documented.

woodruffw:
> Spell check doesn't like "Precertificate", see `docs/spelling_list.txt`

I went ahead and reworded to "pre-certificate," since that's used elsewhere in the docs.

alex (approved):

tests/x509/test_x509.py
```diff
@@ -912,6 +912,44 @@ def test_tbs_certificate_bytes(self, backend):
             cert.signature_hash_algorithm,
         )
 
+    @pytest.mark.supported(
+        only_if=lambda backend: backend.signature_hash_supported(
+            hashes.SHA1()
```
reaperhulk, line 917: This can be sha256 if you use a different cert than `post2000utctime` for your no-SCT extension test. `x509/cryptography.io.pem` would work?

reaperhulk (changes requested):

reaperhulk (approved):

src/rust/src/x509/certificate.rs
```diff
@@ -216,6 +216,39 @@ impl Certificate {
...
+                let filtered_extensions: x509::Extensions<'_> = Asn1ReadableOrWritable::new_write(
+                    asn1::SequenceOfWriter::new(filtered_extensions),
+                );
+                tbs_precert.extensions = Some(filtered_extensions);
+                let result = asn1::write_single(&tbs_precert);
+                Ok(pyo3::types::PyBytes::new(py, &result))
+            }
+            None => Err(PyAsn1Error::from(pyo3::exceptions::PyValueError::new_err(
+                "Could not find any extensions in TBS certificate",
+            ))),
```
alex, line 248: this branch is untested, sorry.

woodruffw: Grr. Okay, time to see if there's already a vector with no extensions and reuse it...

alex: There is! I dont' remember which one it is though. Check test-vectors.rst

woodruffw: 🤞

alex (approved):

PR #7282, opened 2022-05-27
add centos stream 9 to ci

reaperhulk:
We have previously not added it due to issues with their OpenSSL patches, but let's see what we get this time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,6 +140,7 @@ jobs:
           - {IMAGE: "ubuntu-rolling", TOXENV: "py310"}
           - {IMAGE: "fedora", TOXENV: "py310"}
           - {IMAGE: "alpine", TOXENV: "py310"}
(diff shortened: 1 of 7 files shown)


alex (approved):

alex:
So the problem here is that some signatures are rejected, but there's not a programatic way to interrogate this. Is that right?

reaperhulk:
Yeah, there are a set of additional constraints added by OS level policies that our tests don't understand. That likely explains almost every test failure, but there were also some test failures (before this latest rebase) that were related to patches CentOS/RHEL carried on that version of OpenSSL. I'm not super motivated to figure this out right now, although perhaps @tiran or someone else at RH is?

alex:
We need a way to know sha1 is rejected for signatures, that we can programmatically interrogate, and then make use of in backend `signature_hash_supported`.

Without centos's openssl somehow exposing that (`#define`, function call, whatever), this won't be supported.

PR #7287, opened 2022-05-29
turn off coverage as an experiment

reaperhulk:


reaperhulk:
This has taught us nothing because CI variance is ludicrously high.

PR #7294, opened 2022-05-31
ci grouping

reaperhulk:
This uses [workflow commands](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions) to do grouping, hiding the verbose tox output. The implementation is a bit inelegant here and it'd be better to try to get tox to support these commands directly long term.

Better ideas probably exist, but this at least serves as a PoC for making it much easier to read our CI output.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,10 +120,12 @@ jobs:
         if: matrix.PYTHON.OPENSSL
       - name: Tests
         run: |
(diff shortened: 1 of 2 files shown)


alex:
As discussed on IRC, we're trending towards unreadable here and a pre-requisuite might be moving to something like `nox`.

reaperhulk:
(Closing until there's ambition to do some nox work)

PR #7311, opened 2022-06-05
remove verify_interface

### LINKED ISSUE #7278: Remove verify_interface
Now that `register_interface` is gone we have no use for `verify_interface`, but https://github.com/aws/aws-encryption-sdk-python/issues/464 is a blocker for removing it.

reaperhulk:
fixes #7278


alex:
I'm a bit nervous about merging this -- it can't be shipped until aws-encryption-sdk is shipped and most users are updated.

reaperhulk:
I’m fine with holding this for 39 or seeing if an sdk release occurs

alex:
I've placed it in the 38th milestone, we can review the state (release + download numbers for aws-encryption-sdk) then.

alex:
for future reference, commit where this was removed: https://github.com/aws/aws-encryption-sdk-python/commit/d3c763821c356e6a64ed681919f189b0340d0865 not yet in a release

alex:
Now in a release, but uptake of that release is slow. <50% of downloads per:

```
SELECT
  FORMAT_DATE("%Y-%m-%d", timestamp) AS yyyymmdd,
  SUM(CASE WHEN file.version = '3.1.1' THEN 1 ELSE 0 END) as updated_downloads,
  COUNT(*) as total_downloads,
  ROUND(100 * SUM(CASE WHEN file.version = '3.1.1' THEN 1 ELSE 0 END) / COUNT(*), 2) as updated_percent
FROM
  `bigquery-public-data.pypi.file_downloads`
WHERE
  file.project = 'aws-encryption-sdk'
  AND DATE(timestamp) > DATE_SUB(CURRENT_DATE(), INTERVAL 30 DAY)
GROUP BY
  yyyymmdd
ORDER BY yyyymmdd DESC
```

reaperhulk:
Latest data shows only around 40% of downloads are the latest version for aws-encryption-sdk.

reaperhulk:
We're now up to 50%, which is still not great.

alex:
If one prefers visual: https://pepy.tech/project/aws-encryption-sdk?versions=3.1.1&versions=* gap between them is the installs that would be broken.

reaperhulk:
closing since we did #7648 instead for now.

PR #7324, opened 2022-06-09
Update setup.py

tsv19su254052:
corrected indents and hyphens


alex:
Our source code is formatted with black.

PR #7346, opened 2022-06-18
Revert "try pinning to older 3.10"

alex:
Reverts pyca/cryptography#7344


alex:
Not to be merged until we know the cause of this!

reaperhulk (approved):

PR #7354, opened 2022-06-20
remove step from downstream tests that isn't required

alex:


alex:
They're already up to date from setup-python as far as I can tell. This has been around since we migrated from travis.

reaperhulk (approved):

PR #7380, opened 2022-06-28
docs: Fix type returned by get_extension_for_oid() and get_extension_for_class()

F30:
The previous wording (to me) sounded as if an instance of the specific `ExtensionType` subclass would be returned.

But the methods actually return the parent `Extension` instance, from which the specific `ExtensionType` instance can be accessed through the `value` property:

```
for ext in self:
    if isinstance(ext.value, extclass):
        return ext
```

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -1568,7 +1568,7 @@ X.509 Extensions
 
         :param oid: An :class:`ObjectIdentifier` instance.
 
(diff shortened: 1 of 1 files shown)


reaperhulk (approved):

reaperhulk:
This is definitely an improvement in clarity, thanks!

alex:
circleci jobs didn't start, going to try poking them

PR #7382, opened 2022-06-28
Add the rustc version to debug assistance output when setup.py fails

### LINKED ISSUE #7381: Include Rust version in DEBUG ASSISTENCE message?
I'm not sure what the best way to do this is but it seems like it would be helpful to include the output of `rustc -V` in the DEBUG ASSISTENCE.

hoodmane:
Resolves #7381.

Example output:

rustc on path:
```
    Python: 3.10.5
    platform: Linux-5.17.5-76051705-generic-x86_64-with-glibc2.31
    pip: 22.0.4
    setuptools: 62.0.0
    setuptools_rust: n/a
    rustc: 1.63.0-dev
```

rustc not on path:
```
    Python: 3.10.5
    platform: Linux-5.17.5-76051705-generic-x86_64-with-glibc2.31
    pip: 22.0.3
    setuptools: 45.2.0
    setuptools_rust: n/a
    rustc: n/a
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,7 +95,7 @@ except:  # noqa: E722
             # If for any reason `rustc --version` fails, silently ignore it
             rustc_output = subprocess.run(["rustc", "--version"], capture_output=True, timeout=0.5, encoding="utf8").stdout
             version = rustc_output.rpartition(" ")[-1]
-        except Exception:
+        except subprocess.SubprocessError:
             pass
     print(f"    rustc: {version}")
 
--- a/setup.py
+++ b/setup.py
@@ -6,6 +6,9 @@
 
 import os
 import platform
+import re
+import shutil
+import subprocess
 import sys
 
 from setuptools import setup
@@ -99,7 +102,7 @@ except:  # noqa: E722
                 timeout=0.5,
                 encoding="utf8",
             ).stdout
-            version = rustc_output.strip().removeprefix("rustc ")
+            version = re.sub("^rustc ", "", rustc_output.strip())
         except subprocess.SubprocessError:
             pass
     print(f"    rustc: {version}")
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -89,6 +89,16 @@
         except pkg_resources.DistributionNotFound:
             version = "n/a"
         print(f"    {dist}: {version}")
+    version = "n/a"
+    if shutil.which("rustc") is not None:
+        try:
+            # If for any reason `rustc --version` fails, silently ignore it
+            rustc_output = subprocess.run(["rustc", "--version"], capture_output=True, timeout=0.5, encoding="utf8").stdout
+            version = rustc_output.rpartition(" ")[-1]
```
alex, line 97:
```suggestion
            version = subprocess.run(["rustc", "--version"], capture_output=True, timeout=0.5, encoding="utf8").stdout.strip()
```

I think this is fine, full detail is more appropriate here.

alex:
I suspect you'll also need to run `black` over this to make the linters happy.

setup.py
hoodmane: How about `.strip().remove_prefix("rustc ")`?

alex: That'd be fine, except `removeprefix` looks like it wasn't added until 3.9 and we support 3.6 😬

hoodmane: `re.sub("^rustc ", "", rustc_output.strip())`?

hoodmane:
Do you squash merge or should I fix up the commits myself?

setup.py
```diff
@@ -89,6 +92,21 @@
...
         print(f"    {dist}: {version}")
+    version = "n/a"
+    if shutil.which("rustc") is not None:
+        try:
+            # If for any reason `rustc --version` fails, silently ignore it
+            rustc_output = subprocess.run(
+                ["rustc", "--version"],
+                capture_output=True,
+                timeout=0.5,
+                encoding="utf8",
```
alex, line 103: Please add `check=True` so an exception will be raised if the command fails.

alex (approved):

PR #7386, opened 2022-07-03
Re-use constant from rust-asn1

alex:


alex:
failing due to an upstream rust bug https://github.com/rust-lang/rust/issues/98833

reaperhulk (approved):

Issue #7358, opened 2022-06-21
Can legacy OpenSSL provider be optional?


kulikjak:
Would it be possible to make legacy provider (when compiled with OpenSSL 3) optional? AFAICT, it's currently not really possible to use cryptography without a legacy provider being available, but I think it should be.

I tried removing the assert that checks for that:
```patch
--- cryptography-37.0.2/src/cryptography/hazmat/bindings/openssl/binding.py
+++ cryptography-37.0.2/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -173,9 +173,6 @@ class Binding:
                     cls._legacy_provider = cls.lib.OSSL_PROVIDER_load(
                         cls.ffi.NULL, b"legacy"
                     )
-                    _openssl_assert(
-                        cls.lib, cls._legacy_provider != cls.ffi.NULL
-                    )
                     cls._default_provider = cls.lib.OSSL_PROVIDER_load(
                         cls.ffi.NULL, b"default"
                     )
```
and everything still seems to work (of course, without those legacy algorithms). The test suite is failing only where expected, and the code using cryptography I tested (Solaris package manager) works as anticipated.

Can something like this be done? Internally, I already made a patch that also handles missing legacy provider in the test suite (so it's fully green again); should I open a PR with it?

reaperhulk:
I’d be interested in supporting building without the legacy provider, but in addition to the test suite there probably needs to be reasonable error messages inside `UnsupportedAlgorithm` so users who run into this understand what has happened.

We’d also want to add a job to the test matrix that exercises this path.

kulikjak:
That's true, currently it fails with a pretty ugly:
```
cryptography.exceptions.InternalError: Unknown OpenSSL error. This error is commonly encountered when another library is not cleaning up the OpenSSL error stack. If you are using cryptography with ano
ther library that uses OpenSSL try disabling it before reporting a bug. Otherwise please file an issue at https://github.com/pyca/cryptography/issues with information on how to reproduce this. ([_OpenSSLErrorWith
Text(code=50856204, lib=6, reason=524556, reason_text=b'error:0308010C:digital envelope routines::unsupported')]
```
so that is not good. I will look into it.

alex:
I think there's an important question here of what the API for this should be. It seems to me either:

- Users must explicitly opt into not loading the legacy provider - how do they specify this?
- We always try to load the legacy provider, but if it's not present we simply ignore it.

kulikjak:
[...]

PR #7391, opened 2022-07-04
WIP: Support for optional legacy provider

kulikjak:
Based on our discussion in #7358 I am opening this PR with the initial changes necessary for this feature.

It doesn't handle usage of legacy ciphers gracefully yet (with `UnsupportedAlgorithm`) and the question about whether it should be enabled explicitly or whenever it's available still needs to be decided, but it's a start.

I am open to any changes or suggestions - this is just what I came up with while having very limited knowledge of cryptography internals.

### DIFF
--- a/src/cryptography/hazmat/bindings/openssl/binding.py
+++ b/src/cryptography/hazmat/bindings/openssl/binding.py
@@ -173,9 +173,6 @@ class Binding:
                     cls._legacy_provider = cls.lib.OSSL_PROVIDER_load(
                         cls.ffi.NULL, b"legacy"
                     )
-                    _openssl_assert(
-                        cls.lib, cls._legacy_provider != cls.ffi.NULL
-                    )
                     cls._default_provider = cls.lib.OSSL_PROVIDER_load(
                         cls.ffi.NULL, b"default"
                     )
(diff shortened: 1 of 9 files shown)


tests/conftest.py
```diff
@@ -33,6 +33,11 @@ def pytest_runtest_setup(item):
     if openssl_backend._fips_enabled:
         for marker in item.iter_markers(name="skip_fips"):
             pytest.skip(marker.kwargs["reason"])
+    if (openssl_backend._lib and
```
alex, line 36: I think `_lib` should always exist -- is there a case where it doesn't?

src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -173,9 +173,6 @@ def _ensure_ffi_initialized(cls):
                     cls._legacy_provider = cls.lib.OSSL_PROVIDER_load(
                         cls.ffi.NULL, b"legacy"
                     )
-                    _openssl_assert(
-                        cls.lib, cls._legacy_provider != cls.ffi.NULL
-                    )
```
alex, line 178: Removing this means that if it fails to load _unintentionally_, the user will never notice. Can this happen? Should we require users to opt into "legacy provider is not required" somehow?

reaperhulk:
Continued in #7650

PR #7414, opened 2022-07-10
Run benchmarks in CI

### LINKED ISSUE #7079: Add benchmarks to the repository
We care about performance at least a little bit, but presently there is no standard way of measuring performance.

I propose we:

- [x] Introduce some performance tests
- [x] Start running them in CI so they don't bit-rot
- [x] Commit to adding new ones / referencing existing ones when we do optimization work
- [ ] Some day maybe figure out how to run performance tests in CI so we can catch regressions

alex:
fixes #7079

### DIFF
--- a/.github/compare_benchmarks.py
+++ b/.github/compare_benchmarks.py
@@ -1,0 +1,42 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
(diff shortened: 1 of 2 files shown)


alex:
Et voila: https://github.com/pyca/cryptography/actions/runs/2645918537

We'll have to actually go out of our way to look at stuff when it's performant relevant, but hey. Numbers.

reaperhulk (approved):

PR #7422, opened 2022-07-15
Remove usage of actions-rs actions

alex:
They appear to be unmaintained at this point: https://github.com/actions-rs/toolchain/issues/216

Also for most of these we can simply use the Rust that comes with the GHA image


.github/workflows/wheel-builder.yml
```diff
@@ -216,12 +213,9 @@ jobs:
         with:
           python-version: ${{ matrix.PYTHON.VERSION }}
           architecture: ${{ matrix.WINDOWS.ARCH }}
-      - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af
+      - uses: dtolnay/rust-toolchain@3d5c338b548e8aff8b55d96390e28fa86945e1b2
```
reaperhulk, line 216: Why do we need this for the windows wheel builder but not the windows CI job now?

PR #7439, opened 2022-07-20
feat: drop max password length of 72 characters from `serialize_ssh_private_key`

### LINKED ISSUE #7436: RFE: remove password length limit on `OpenSSH` encoded private keys
### Summary

The current implementation of `SSH` serialization enforces a [maximum password length](https://github.com/pyca/cryptography/blob/main/src/cryptography/hazmat/primitives/serialization/ssh.py#L56) of 72 characters and [references](https://github.com/pyca/cryptography/blob/28934d16fcad8de97284483aeec5d4c2fffab30d/src/cryptography/hazmat/primitives/serialization/ssh.py#L614-L615) the OpenSSH format as a source.

The [OpenSSH format](https://github.com/openssh/openssh-portable/blob/eb679e2959bdb15454eb94751930eb4c9110da94/PROTOCOL.key) however specifies no such restriction since the passphrase is never stored and is supplied directly to the underlying KDF implementation.

`bcrypt 3.1.5`, required by `cryptography`, does not specify any requirements for the password either (besides [non-zero length](https://github.com/pyca/bcrypt/blob/3.1.5/src/_csrc/bcrypt_pbkdf.c)).

Originally reported [here](https://github.com/ansible-collections/community.crypto/issues/492)

Ajpantuso:
### Summary

_OpenSSH_ encoded private keys using password encryption are currently limited to passwords `<= 72` characters in length when using `cryptography`. This requirement is unnecessary as `bcrypt_pbkdf2`, the key derivation function used by _OpenSSH_, pre-hashes the supplied password avoiding the 72 character input limit that `bcrypt` implementations may impose. Therefore this PR removes the character limit.

Closes #7436.

### DIFF
--- a/tests/hazmat/primitives/test_serialization.py
+++ b/tests/hazmat/primitives/test_serialization.py
@@ -2342,14 +2342,6 @@ class TestOpenSSHSerialization:
 
         private_key = ec.generate_private_key(ec.SECP256R1(), backend)
 
-        # too long password
-        with pytest.raises(ValueError):
-            private_key.private_bytes(
-                Encoding.PEM,
-                PrivateFormat.OpenSSH,
-                BestAvailableEncryption(b"p" * 73),
-            )
-
         # unknown encryption class
         with pytest.raises(ValueError):
             private_key.private_bytes(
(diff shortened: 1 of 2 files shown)


tests/hazmat/primitives/test_serialization.py
```diff
@@ -2358,6 +2350,41 @@ def test_serialize_ssh_private_key_errors(self, backend):
                 DummyKeySerializationEncryption(),
             )
 
+    @pytest.mark.supported(
+        only_if=lambda backend: ssh._bcrypt_supported,
+        skip_message="Requires that bcrypt exists",
+    )
+    @pytest.mark.parametrize(
+        ("password"),
```
alex, line 2358:
```suggestion
        "password",
```

nit

alex (commented):
One small nit.

tests/hazmat/primitives/test_serialization.py
Ajpantuso: Fixed.

alex (approved):

PR #7450, opened 2022-07-26
add SSL_CTX_set_security_level to C binding

lanyizi:
These functions can be used to control the security level of a SSL context, which sometimes can be useful.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -366,6 +366,9 @@ unsigned long SSL_CTX_set_tmp_dh(SSL_CTX *, DH *);
 unsigned long SSL_CTX_set_tmp_ecdh(SSL_CTX *, EC_KEY *);
 unsigned long SSL_CTX_add_extra_chain_cert(SSL_CTX *, X509 *);
 
+void SSL_CTX_set_security_level(SSL_CTX *, int);
+int SSL_CTX_get_security_level(SSL_CTX *);
+
 /*- These aren't macros these functions are all const X on openssl > 1.0.x -*/
 
 /*  methods */


alex:
Can you provide a bit more context on the intended use here? Do you intend to submit a PR to pyOpenSSL using this?

tiran:
FYI, you can already set the security level with ``SSL_CTX_set_cipher_list``, e.g. ``"@SECLEVEL=1:DEFAULT"``.

lanyizi:
> Can you provide a bit more context on the intended use here? Do you intend to submit a PR to pyOpenSSL using this?

We are trying to write a server emulator for a very old game, its server used a very old 512-bit RSA private key for SSL connection.
That key is now considered insecure and [rejected](https://github.com/openssl/openssl/blob/5203a8dfdc209f05c7dbd9c1e5208743fcaa6752/ssl/t1_lib.c#L3050) by OpenSSL (`SSL_R_EE_KEY_TOO_SMALL`), but by setting the security level to 0, it [should allow any key size](https://github.com/openssl/openssl/blob/36c269c3023f5eb626ec79777ed8b285ef939be2/ssl/ssl_cert.c#L988).

Personally I could just obtain the lib module from pyOpenSSL (although it's a bit hackish), but if it's necessary I can also make a relevant pull request to pyOpenSSL to expose a python method for the `Context` object.



> FYI, you can already set the security level with `SSL_CTX_set_cipher_list`, e.g. `"@SECLEVEL=1:DEFAULT"`.
Oh, I didn't know about that, gotta try it first, thanks for the tip!

alex:
Yes if the existing APIs for setting ciphersuties are usable, that'd be enable. I don't love adding new APIs for the sole goal of enabling _weaker_ security!

alex:
No response for a month so I'm going to close this. Please leave a comment if we should re-open.

PR #7458, opened 2022-07-29
Provide a way to export PKCS12 in a way compatible with major OS (closes #7293)

### LINKED ISSUE #7293: Can't import cryptography generated pkcs12 into macos keychain
Versions:

* Python: 3.10.1
* cryptography: 37.0.2
* cffi: 1.15.0

Cryptography installed with poetry 1.2.0b1

I'm trying to create a self-signed certificate with cryptography, which mostly works, but it fails to import to macos (12.1) keychain. If I export the same certificate & private key as a chained PEM file and then use the `openssl` command line tool to convert it into a p12, I can successfully import it.

This is the code I use to try to export the p12 using cryptography:

```python
def export(path: Path, cert: x509.Certificate, key: rsa.RSAPrivateKeyWithSerialization, password: bytes):
    with path.open('wb') as fobj:
        fobj.write(
            pkcs12.serialize_key_and_certificates(
                name=None,
                cas=None,
                key=key,
                cert=cert,
                encryption_algorithm=serialization.BestAvailableEncryption(password),
            )
        )
```

However, if instead I do this, macos is happy to import it:

```python
def export(
    path: Path,
    cert: x509.Certificate,
    key: rsa.RSAPrivateKeyWithSerialization,
    password: bytes,
):
    with path.with_suffix(".pem").open("wb") as fobj:
        fobj.write(
            key.private_bytes(
                encoding=serialization.Encoding.PEM,
                format=serialization.PrivateFormat.TraditionalOpenSSL,
                encryption_algorithm=serialization.BestAvailableEncryption(password),
            )
        )
        fobj.write(cert.public_bytes(encoding=serialization.Encoding.PEM))
    subprocess.check_call(
        [
            "openssl",
            "pkc12",
            "-export",
            "-in",
            str(path.with_suffix(".pem")),
            "-passin",
            f"pass:{password.decode()}",
            "-out",
            str(path),
            "-passout",
            f"pass:{password.decode()}",
        ]
    )
```

Am I using `serialize_key_and_certificates` wrong?

schwabe:
This mainly to get comments if this is a good way for the API to implement this or if another approach should be taken.

### DIFF
--- a/docs/hazmat/primitives/asymmetric/serialization.rst
+++ b/docs/hazmat/primitives/asymmetric/serialization.rst
@@ -1002,6 +1002,15 @@ Serialization Encryption Types
 
     :param bytes password: The password to use for encryption.
 
+.. class:: PKCS12CompatibilityEncryption(password)
+
+    Encrypt using 3DES and SHA1 for a given key.
+    This is an option that should be only used to allow exporting PKCS12
+    to devices/software that does not support stronger encryption of PKCS12
+    like macOS 15.x or Android 12.
+
+    :param bytes password: The password to use for encryption.
+
 .. class:: NoEncryption
 
     Do not encrypt.
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -2247,6 +2247,7 @@ class Backend:
             nid_key = -1
             pkcs12_iter = 0
             mac_iter = 0
+            mac_alg = self._ffi.NULL
         elif isinstance(
             encryption_algorithm, serialization.BestAvailableEncryption
         ):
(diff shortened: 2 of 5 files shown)


reaperhulk:
I think a compatibility class is fine, but there are a lot of questions about what every other serialization method that accepts `KeySerializationEncryption` will do when (incorrectly) passing this new class. Additionally, we should be very explicit about what this supports (e.g. macOS 12.x and below, Windows Server 2016 and below, Android 12 and below). I'd also like to determine what the minimum degradation is to support this. e.g. can we just do SHA1 MAC and leave it AES256? Or is 3DES a hard requirement?

schwabe:
In my tests AES128 and AES256 did not work. Only the 3DES and SHA1 were working. I did not test Windows 2016 but for Android and macOS this seems to be a hard requirement. I renamed the security class `PKCS12CompatibilityEncryption` to make it more clear that it is PKCS12 only. I see your point but I am unsure how a better API would be. Something like doing a special encryption class only for PKCS12?

I have successfully tested generating PKCS12 files using the `PKCS12CompatibilityEncryption` that worked on Android 12 and macos 15.

reaperhulk:
Obsoleted by #7560, but thanks for working on this, it helped significantly in that other PR's evolution.

PR #7481, opened 2022-08-04
provide openssl apis related to SM for python

alittlesir:
provide openssl apis related to SM for python


reaperhulk:
This isn't passing CI because it needs to correctly handle OpenSSL versions where these APIs don't exist. That said, I'd like to understand what your plans on here. We don't merge bindings like this without also talking about how they will be used in our higher level APIs. If the intent is SM2 support then you may want to look at https://github.com/pyca/cryptography/pull/5894 to see if it can be updated/finished.

Additionally, I don't think you'll need `EVP_sm3`. You can obtain an SM3 instance via the `EVP_getdigestbyname` function -- we do so in this project already for SM3 support.

alex:
no action after an extended period of time with failing tests. going to close this. if you're still interested in this, leave a comment and we can reopen.

PR #7511, opened 2022-08-15
remove ext_package so top_level.txt is accurate in wheel metadata

### LINKED ISSUE #7510: cryptography packaging `ext_package` causes `_openssl` to be in `top_level.txt` in the wheel
this is extremely minor and I have an idea on how to work around it -- but it appears some of the distribution metadata is incorrect due to the way the cffi modules are built in `cryptography`.

here's a minimal example:

```console
$ pip download cryptography --no-deps ...
$ unzip -p cryptography-*.whl */top_level.txt 
_openssl
cryptography
```

this file is _supposed_ to list the top level modules afaict and `_openssl` is not one of them.  it appears that comes from this: https://github.com/pyca/cryptography/blob/7a4c612b0280a4a01df636f8d648b9c1a1dad6dc/src/_cffi_src/build_openssl.py#L73

I can probably work around this (I'm using this file to do some rudimentary testing of wheels) and regenerate `top_level.txt` in a different way (perhaps via `RECORD`)

I think this _might_ be a `setuptools` bug with `ext_package` -- it can be worked around like this:

```diff
diff --git a/setup.cfg b/setup.cfg
index f9710764d..d96bb2d4f 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,7 +43,6 @@ zip_safe = False
 package_dir =
     =src
 packages = find:
-ext_package = cryptography.hazmat.bindings
 # `install_requires` must be kept in sync with `pyproject.toml`
 install_requires =
     cffi >=1.12
diff --git a/setup.py b/setup.py
index 946bb0f92..320994e69 100644
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ try:
         ],
         rust_extensions=[
             RustExtension(
-                "_rust",
+                "cryptography.hazmat.bindings._rust",
                 "src/rust/Cargo.toml",
                 py_limited_api=True,
                 # Enable abi3 mode if we're not using PyPy.
diff --git a/src/_cffi_src/build_openssl.py b/src/_cffi_src/build_openssl.py
index c5ab3cb3c..84637f090 100644
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -70,7 +70,7 @@ def _extra_compile_args(platform):
 
 
 ffi = build_ffi_for_binding(
-    module_name="_openssl",
+    module_name="cryptography.hazmat.bindings._openssl",
     module_prefix="_cffi_src.openssl.",
     modules=[
         # This goes first so we can define some cryptography-wide symbols.
```

```console
$ rm -rf build *.whl && pip wheel . --no-deps
...
$ unzip -p *.whl */top_level.txt
cryptography
$ unzip -l *.whl | grep '\.so'
  1974696  2022-08-15 14:04   cryptography/hazmat/bindings/_openssl.abi3.so
  3696800  2022-08-15 14:04   cryptography/hazmat/bindings/_rust.abi3.so
```

let me know if this would be worth a patch (though I'd totally understand if this is a "no improvement" / "upstream problem" issue)

asottile-sentry:
resolves #7510

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,7 +43,6 @@ zip_safe = False
 package_dir =
     =src
 packages = find:
-ext_package = cryptography.hazmat.bindings
 # `install_requires` must be kept in sync with `pyproject.toml`
 install_requires =
     cffi >=1.12
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ try:
         ],
         rust_extensions=[
             RustExtension(
-                "_rust",
+                "cryptography.hazmat.bindings._rust",
                 "src/rust/Cargo.toml",
                 py_limited_api=True,
                 # Enable abi3 mode if we're not using PyPy.
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -70,7 +70,7 @@ def _extra_compile_args(platform):
 
 
 ffi = build_ffi_for_binding(
-    module_name="_openssl",
+    module_name="cryptography.hazmat.bindings._openssl",
     module_prefix="_cffi_src.openssl.",
     modules=[
         # This goes first so we can define some cryptography-wide symbols.


alex (approved):
This looks good to me (assuming CI is green). I'm going to wait for @dstufft to have a chance to weigh in before merging though.

dstufft:
There was a reason for `ext_package`, but for the life of me I can't remember what it was or if it's still even relevant at this point. If things work I don't see a reason off hand not to do it.

asottile-sentry:
it appears present in the initial commit -- not sure if there's history predating that: c62e91f3eca01c7e06974994fc16bce06fbffabf

asottile-sentry:
err actually, the relevant one was introduced in 63195b4a2e472949385d7b752d32604166d04ef1

alex:
I think https://github.com/pyca/cryptography/commit/9a00f0539d5ab9b03b84625791663f11d7bed75c is the right commit. But it suggests only a functional purpose -- this is functional so let's merge.

Issue #7513, opened 2022-08-15
x509.Name.from_rfc4514_string not strictly compliant with RFC 4514


axmachado:
According with section 2.1 of the RFC:

~~~txt
[2.1].  Converting the RDNSequence

   If the RDNSequence is an empty sequence, the result is the empty or
   zero-length string.

   Otherwise, the output consists of the string encodings of each
   RelativeDistinguishedName in the RDNSequence (according to Section 2.2), 
   starting with the last element of the sequence and moving
   backwards toward the first.

   The encodings of adjoining RelativeDistinguishedNames are separated
   by a comma (',' U+002C) character.
~~~

The `rfc4514_string` method is doing it right, reversing the order of the RDNs from the sequence that makes the Name, but the parser is not reversing it back when constructing the sequence. 

The side effect of it is evidenced on the code fragment:

~~~python
from cryptography import x509

components = [
    x509.NameAttribute(x509.NameOID.COUNTRY_NAME, 'BR'),
    x509.NameAttribute(x509.NameOID.ORGANIZATION_NAME, 'Test Organization'),
    x509.NameAttribute(x509.NameOID.ORGANIZATIONAL_UNIT_NAME, 'TEST OU 1'),
    x509.NameAttribute(x509.NameOID.ORGANIZATIONAL_UNIT_NAME, 'Test OU 2'),
    x509.NameAttribute(x509.NameOID.LOCALITY_NAME, 'Test City'),
    x509.NameAttribute(x509.NameOID.COMMON_NAME, 'Test Name')
]

name1 = x509.Name(components)
name1_string = name1.rfc4514_string()
name3 = x509.Name.from_rfc4514_string(name2.rfc4514_string())

if name1 != name2:
    print('Error... should be the same name')
if name1 == name3:
    print('Now we have the same name')
~~~

I just cloned the main repository and will send a pull request later today with the fix.

* Python 3.10.5
* cffi==1.15.1
* pip 22.2.2
* cryptography==37.0.4
*

alex:
Great, thank you for sending a fix.

reaperhulk:
Fixed in #7514

PR #7514, opened 2022-08-15
Fix issue #7513 - parsing of RFC4514 strings should reverse the order of RDNs

axmachado:
This is a quick fix for the issue I've opened earlier today (issue #7513)

### DIFF
--- a/src/cryptography/x509/name.py
+++ b/src/cryptography/x509/name.py
@@ -414,6 +414,14 @@ class _RFC4514NameParser:
         return val
 
     def parse(self) -> Name:
+        """
+        Parses the `data` string and converts it to a Name.
+
+        According to RFC4514 section 2.1 the RDNSequence must be
+        reversed when converting to string representation. So, when
+        we parse it, we need to reverse again to get the RDNs on the
+        correct order.
+        """
         rdns = [self._parse_rdn()]
 
         while self._has_data():
--- a/src/cryptography/x509/name.py
+++ b/src/cryptography/x509/name.py
@@ -420,7 +420,7 @@ class _RFC4514NameParser:
             self._read_char(",")
             rdns.append(self._parse_rdn())
 
-        return Name(rdns)
+        return Name(reversed(rdns))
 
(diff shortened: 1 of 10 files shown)


alex:
Can you please take a look at the flake8 issues in CI? You may be able to resolve them by running `black`.

Thanks for the PR!

axmachado:
Now I think it is correct. Sorry about not following the project standards for code formatting.

src/cryptography/x509/name.py
```diff
@@ -420,7 +420,7 @@ def parse(self) -> Name:
             self._read_char(",")
             rdns.append(self._parse_rdn())
 
-        return Name(rdns)
+        return Name(reversed(rdns))
```
alex, line 423: Can you add a comment here referencing the text from the RFC, so we remember why we're reversing one?

alex (commented):
In additino to that comment, can you please add an entry in `CHANGELOG.rst`?

Thanks!

src/cryptography/x509/name.py
axmachado: Ok. Done :-)

axmachado:
> In additino to that comment, can you please add an entry in `CHANGELOG.rst`?
> 
> Thanks!

Can I associate the comment with the 38.0.0 version?

reaperhulk:
Yes you can associate it with 38.0.0. Tests are currently failing because you used `:method:` when you need `:meth:`, but once you fix that this looks ready 😄

axmachado:
Still failing:
phinx.errors.SphinxWarning: /home/runner/work/cryptography/cryptography/CHANGELOG.rst:49:py:class reference target not found: cryptography.x509.name._RFC4514NameParser

I will suppress the reference to the protected class.

reaperhulk:
Oh, yes, don't link the private class (even though the fix is in there), just the public API surface.

axmachado:
Finally, all check passed :+1: Hope everything is Ok now. Sorry about the multiple fixes. I did not know how to run all the CI tests locally.

reaperhulk (approved):

PR #7520, opened 2022-08-16
Initial APIs for making serialization encryption configurable

alex:
closes #7441

### DIFF
--- a/src/cryptography/hazmat/primitives/serialization/ssh.py
+++ b/src/cryptography/hazmat/primitives/serialization/ssh.py
@@ -15,9 +15,11 @@ from cryptography.hazmat.primitives.asymmetric import dsa, ec, ed25519, rsa
 from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
 from cryptography.hazmat.primitives.serialization import (
     Encoding,
+    KeySerializationEncryption,
     NoEncryption,
     PrivateFormat,
     PublicFormat,
+    _KeySerializationEncryption,
 )
 
 try:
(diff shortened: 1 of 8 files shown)


reaperhulk:
`enc_type = KeySerializationEncryptionBuilder().kdf_rounds(32).build(b"password")` is the usage model then... And a more complex thing might look like:

```python
KeySerializationEncryptionBuilder().mac_algorithm(SOME_ENUM).
    cipher(SOME_OTHER_ENUM).
    mac_rounds(100).
    kdf_rounds(300).
    build(b"password")
```

Do we have separate builders for types (e.g. `PKCS12SerializationEncryptionBuilder`)? Is `SSH` a distinct type? I think this approach is a reasonable one though.

alex:
I was going to just use the same type for each, and then have things check if the method was valid for that `_format`.

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -601,9 +603,10 @@ def load_ssh_private_key(
     return private_key
 
 
-def serialize_ssh_private_key(
+def _serialize_ssh_private_key(
     private_key: _SSH_PRIVATE_KEY_TYPES,
-    password: typing.Optional[bytes] = None,
+    password: typing.Optional[bytes],
```
reaperhulk, line 608: Is it possible to get `None` except by a direct call to an API that's private? `private_bytes` always sets it to an empty byte string or the password so it seems like we could just claim it's always bytes.

reaperhulk (approved):

PR #7542, opened 2022-08-26
add AES128/AES256 classes

reaperhulk:
These let developers be more explicit about the allowable key lengths
for an AES key and make auditing the codebase a bit easier.
    
But that's not really why we're adding them. In some upcoming
serialization features we need to be able to specify AES 128 vs AES 256
and the current class doesn't work for that since it computes key
length from the key you provide it when instantiating the class.
That's incompatible with serialization where the key is derived
later in the process. C'est la vie.


alex:
concept lgtm

alex (approved):

PR #7553, opened 2022-08-31
attempt to fix josepy tests by pinning poetry

alex:


alex:
@adferrand do you remember why we used `-c` with `pip` here? The latest poetry broke this and I'm trying to figure out how to fix it. In that process I've been told that perhaps `-c` was never supported https://github.com/python-poetry/poetry/issues/6275#issuecomment-1232861048

adferrand:
Hello @alex !

We used the `-c` with pip to have reproducible builds.

Indeed poetry maintains a full dependency tree of the project with version locking, and can export this tree in the `requirements.txt` format.

So we can pass this file as a constraint file to pip and ensure that pip will install the exact version of every dependency needed as it has been defined by the upstream project.

alex:
The view of the poetry folks seems to be that using poetry export as a
constraints file isn't supported. Do you think we could switch to using it
as a requirements.txt?

בתאריך יום ד׳, 31 באוג׳ 2022, 11:15, מאת Adrien Ferrand ‏<
***@***.***>:

alex:
@adferrand it looks like not supporting `-c` is somewhat intentional. Any ideas for how we can work around this to get stuff fixed? Pin poetry temporarily to give us time to think?

reaperhulk (approved):

PR #7581, opened 2022-09-08
Add `SSL_peek_ex`, `SSL_read_ex`, and `SSL_write_ex` to bindings

illia-v:
Since support for OpenSSL < 1.1.1 has been dropped, it looks like it is possible to add these functions to bindings and use them in pyOpenSSL to increase the number of bytes the library allows to read and write before raising `OverflowError`.


tiran:
Don't use the functions, they are buggy and won't get fixed until OpenSSL 4.0. I wasn't aware about the bug when I introduced them in Python. See @davidben 's bug https://github.com/openssl/openssl/issues/8208 for more infos.

illia-v:
@tiran interesting, thanks for letting me know. Do you know if `SSL_write_ex` is buggy?

reaperhulk:
Given these don't work on boring/libre I'm reluctant to have us switch to using them in pyOpenSSL. I'm also unclear how much of a performance problem it is to call read/write more often. The current limit is presumably ~2GB, is that a real problem?

illia-v:
@reaperhulk I think users need to call both `SSL_read` and `SSL_read_ex` in a loop because OpenSSL always returns around 16 KiB at a time. Regardless of the peculiarity, the functions are usually called with the total number of needed bytes.

The problem is that, as a user, I would not expect that a program will fail if an integer 2<sup>31</sup> or greater is passed to a Python function.
A program may be tested by downloading/uploading a few kilobytes. Then it may fail in production unexpectedly if it has to deal with a few gigabytes.

alex:
Notwithstanding that using `int` for length types is unfortunate, the changing/wrong semantics around EOF are sufficient for me to say we don't want to expose/use these functions until they are resolved. There's been no movement from the OpenSSL devs, so for now we aren't going to accept this PR.

Thanks for your contribution though.

PR #7601, opened 2022-09-11
use fixed pool to improve perf of aead ChaCha20Poly1305

reaperhulk:
~35-45% speedup on benchmarks when reusing the same key for multiple operations.

Other AEADs will follow, although they're...harder.

```
| tests/bench/test_aead.py::test_chacha20poly1305_decrypt | 7125.08 ns | 4375.12 ns | 39% faster |
| tests/bench/test_aead.py::test_chacha20poly1305_encrypt | 8500.18 ns | 4709.0 ns | 45% faster |
```

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/aead.py
+++ b/src/cryptography/hazmat/backends/openssl/aead.py
@@ -63,6 +63,29 @@ def _evp_cipher(cipher_name: bytes, backend: "Backend"):
     return evp_cipher
 
 
+def _aead_create_ctx(
+    backend: "Backend",
+    cipher: "_AEAD_TYPES",
+    key: bytes,
+):
+    ctx = backend._lib.EVP_CIPHER_CTX_new()
+    backend.openssl_assert(ctx != backend._ffi.NULL)
+    ctx = backend._ffi.gc(ctx, backend._lib.EVP_CIPHER_CTX_free)
+    cipher_name = _aead_cipher_name(cipher)
+    evp_cipher = _evp_cipher(cipher_name, backend)
+    key_ptr = backend._ffi.from_buffer(key)
+    res = backend._lib.EVP_CipherInit_ex(
+        ctx,
+        evp_cipher,
+        backend._ffi.NULL,
+        key_ptr,
+        backend._ffi.NULL,
+        0,
+    )
+    backend.openssl_assert(res != 0)
+    return ctx
+
+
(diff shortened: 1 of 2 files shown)


src/cryptography/hazmat/backends/openssl/aead.py
```diff
@@ -63,6 +63,29 @@ def _evp_cipher(cipher_name: bytes, backend: "Backend"):
     return evp_cipher
 
 
+def _aead_create_ctx(
```
alex, line 66: should aead_setup call this?

alex (approved):

reaperhulk:
@bdraco last year you talked to us about speeding up ChaCha20Poly1305 by ~20% when reusing the same key repeatedly. IIRC the example use case was home-assistant HomeKit integration since HAP uses ChaCha20Poly1305. We've just landed this PR, which (using the exact same API and in a thread safe manner) allows ctx reuse with 35%+ performance improvement and it will be released in version 39. I'll send a PR to home-assistant after release, but I thought you'd likely be interested. 😄

bdraco:
Hi Paul,

Thanks for the ping.  This is fantastic news!

Our solution was to create https://github.com/bdraco/chacha20poly1305-reuseable 

Its currently used in https://github.com/ikalchev/HAP-python https://github.com/postlund/pyatv and https://github.com/Jc2k/aiohomekit

There was one other change since the original PR to squeeze a bit more performance out
https://github.com/bdraco/chacha20poly1305-reuseable/commit/be14cc2dc74d81b49bf761c32ce5536e31be74ce#diff-61eb0fe778bd395a16e1d556d37bc4cc2c1649b88e9d797eeee8f839125dfdf6R114

I'm hoping we can drop that lib and switch everything back over to use your changes.

Cheers
-Nick

reaperhulk:
@bdraco Oh, fascinating. I dropped your code into my test harness and (unsurprisingly) you're faster since you don't have the locking synchronization overhead:

```
| tests/bench/test_aead.py::test_chacha20poly1305_decrypt | 4583.04 ns | 3458.0 ns | 25% faster |
| tests/bench/test_aead.py::test_chacha20poly1305_encrypt | 4749.98 ns | 4374.89 ns | 8% faster |
```

But the original timings were 7100ns and 8500ns, respectively, so we're hopefully performant enough that you can drop this 😄

PR #7604, opened 2022-09-12
Add downstream test for scapy

gpotter2:
Hi all, @alex,

This PR is a follow-up to https://github.com/pyca/cryptography/pull/7234#issuecomment-1241966358. This runs a subset of our tests (<1min) specifically using the cryptography module.

Thanks for the help :P

### DIFF
--- a/.github/downstream.d/scapy.sh
+++ b/.github/downstream.d/scapy.sh
@@ -1,0 +1,18 @@
+#!/bin/bash -ex
+
+case "${1}" in
+    install)
+        git clone --depth=1 https://github.com/secdev/scapy
+        cd scapy
+        git rev-parse HEAD
+        pip install tox
+        ;;
+    run)
+        cd scapy
+        # this tox case uses sitepackages=true to use local cryptography
+        tox -qe cryptography
+        ;;
+    *)
+        exit 1
+        ;;
+esac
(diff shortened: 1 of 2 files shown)


.github/downstream.d/scapy.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -ex
+
+case "${1}" in
+    install)
+        git clone --depth=1 https://github.com/secdev/scapy
+        cd scapy
+        git rev-parse HEAD
+	pip install tox
```
alex, line 8: nit: can you fix the indentation here

.github/downstream.d/scapy.sh
```diff
@@ -0,0 +1,17 @@
...
+        ;;
+    run)
+        cd scapy
+        tox -qe cryptography
```
alex, line 12: Can you add a comment saying that this uses the local cryptography because that tox env has `sitepackages = true` which makes it use externally installed stuff?

Also, is that understanding correct 😂

gpotter2: That's correct ! 😛

.github/downstream.d/scapy.sh
gpotter2: you're right it's a tab

alex (approved):

PR #7641, opened 2022-09-23
Warn users on 32-bit Pythons on 64-bit windows

alex:
They're leaving a ton of performance on the table.


reaperhulk (approved):
What a ridiculous env var.

PR #7649, opened 2022-09-25
mark verify_interface with a deprecation comment

reaperhulk:
this will help us revisit it in the future


alex (approved):
there are now 2 comments that say the same thing. lol

PR #7690, opened 2022-10-11
upgrade some openssls in CI

reaperhulk:


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -2204,9 +2204,10 @@ def load_pkcs12(
             res = self._lib.PKCS12_parse(
                 p12, password_buf, evp_pkey_ptr, x509_ptr, sk_x509_ptr
             )
-
+        # OpenSSL 3.0.6 leaves errors on the stack even in success, so 
+        # we consume all errors unconditionally.
```
alex, line 2208: Can you file a bug with openssl and include a link here.

alex (approved):

PR #7705, opened 2022-10-13
Expand PKCS7 SignerInfo struct

alex:


reaperhulk:
Rust coverage appears to have become confused
image

PR #7710, opened 2022-10-15
Make sure that one of full_name, relative_name or crl_issuer is set i…

mathiasertl:
Noticed that it's possible to create Distribution Points with no full name, relative name or issuer:

```python
>>> x509.DistributionPoint(None, None, None, None)
<DistributionPoint(full_name=None, relative_name=None, reasons=None, crl_issuer=None)>
```

RFC 5280, section 4.2.1.13 says:

```quote
... either distributionPoint or cRLIssuer MUST be present.
```

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -589,6 +589,11 @@ class DistributionPoint:
                 "You cannot provide both full_name and relative_name, at "
                 "least one must be None."
             )
+        if not full_name and not relative_name and not crl_issuer:
+            raise ValueError(
+                "Either full_name, relative_name or crl_issuer must be "
+                "provided."
+            )
 
         if full_name is not None:
             full_name = list(full_name)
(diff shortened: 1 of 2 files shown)


alex (commented):
Looks like there's a coverage issue, presumabely because some other test case is now triggering this check?

mathiasertl:
Just checking this. Discovered that there is a similar check further down in the constructor:

```
        if reasons and not crl_issuer and not (full_name or relative_name):
            raise ValueError(
                "You must supply crl_issuer, full_name, or relative_name when "
                "reasons is not None"
            )
```

... and the `test_reason_only` test would no longer reach these lines.

Now rereading RFC5280 to see if there is any basis for the above (none on my second read...)

mathiasertl:
Ah, hmm, the preceeding line says

```quote
a DistributionPoint MUST NOT consist of only the reasons field;
```

so the check now not covered is a valid check, strictly speaking. But the new check is still valid and covers the "only reasons flags" as well. If you agree, I would remove the check (and leave the test).

alex:
Works for me.

mathiasertl:
done. Thanks for reviewing!

alex (approved):

PR #7718, opened 2022-10-17
disable sparse registry since its not working

alex:


reaperhulk:
Is this expected to be broken for a week or just a few days?

alex:
No timeline known yet.

reaperhulk (approved):
(we'll revert as soon as this is fixed)

PR #7725, opened 2022-10-21
attempt to workaround downstream package testing situation

alex:


alex:
```
E   pkg_resources.DistributionNotFound: The 'cryptography>=2.5.0' distribution was not found and is required by the application
```
progress!

reaperhulk:
We should try to do the renaming to the latest release version rather than fixing it at 38.0.0

reaperhulk (approved):

PR #7729, opened 2022-10-22
Rewrite PKCS7 signature generation in Rust

alex:


src/rust/src/pkcs7.rs
```diff
@@ -81,10 +124,259 @@ fn serialize_certificates<'p>(
...
+            raw_data,
+            options.contains(pkcs7_options.getattr(crate::intern!(py, "Text"))?)?,
+        )
+    };
+
+    let content_type_bytes = asn1::write_single(&PKCS7_DATA_OID)?;
+    let signing_time_bytes = asn1::write_single(&x509::certificate::time_from_chrono(
+        chrono::Utc::now().with_nanosecond(0).unwrap(),
+    )?)?;
+    let smime_cap_bytes = asn1::write_single(&asn1::SequenceOfWriter::new([
```
reaperhulk, line 152: Could we add a comment here about how this is an aggressively curated list based on OpenSSL's implementation with a link to https://github.com/openssl/openssl/blob/667a8501f0b6e5705fd611d5bb3ca24848b07154/crypto/pkcs7/pk7_smime.c#L150?

src/rust/src/pkcs7.rs
```diff
@@ -81,10 +124,259 @@ fn serialize_certificates<'p>(
...
+        AES_256_CBC_OID,
```
alex, line 153:
```suggestion
        // Subset of values OpenSSL provides:
        // https://github.com/openssl/openssl/blob/667a8501f0b6e5705fd611d5bb3ca24848b07154/crypto/pkcs7/pk7_smime.c#L150
        // removing all the ones that are bad cryptography
        AES_256_CBC_OID,
```

reaperhulk (approved):

PR #7734, opened 2022-10-22
Remove SHA1 signatures support from PKCS7

alex:
refs #7729


reaperhulk (approved):

decaz:
Unfortunately SHA1 wasn't removed here:
https://github.com/pyca/cryptography/blob/17da400b651fd940879a14b6d9c2964037de8d7f/src/cryptography/hazmat/primitives/serialization/pkcs7.py#L88-L101

decaz:
Ah, wait.. Why SHA1 support was removed?

alex:
Because signatures over sha1 are not secure.

PR #7739, opened 2022-10-23
In memleak test, track reallocs of startup pointers (#7379)

pkgw:
The conda-forge build of this package was encountering test failures because their Windows build configuration involves a situation where some of the tests cause a realloc of memory in the memleak test's `start_heap`. Here we allow such reallocs and do not count them as leaks if they do not add too much memory usage, where "too much" is currently defined as 1024 bytes.

refs #7379

### DIFF
--- a/tests/hazmat/backends/test_openssl_memleak.py
+++ b/tests/hazmat/backends/test_openssl_memleak.py
@@ -27,6 +27,8 @@ def main(argv):
     from cryptography.hazmat.bindings._openssl import ffi, lib
 
     heap = {}
+    start_heap = {}
+    start_heap_realloc_delta = [0]  # 1-item list so callbacks can mutate it
 
     BACKTRACE_ENABLED = False
     if BACKTRACE_ENABLED:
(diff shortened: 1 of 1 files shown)


tests/hazmat/backends/test_openssl_memleak.py
```diff
@@ -113,18 +129,22 @@ def free(ptr, path, line):
...
         assert result == 1
 
-    remaining = set(heap) - start_heap
+    remaining = set(heap) - set(start_heap)
 
-    if remaining:
-        sys.stdout.write(json.dumps(dict(
+    # The constant here is the number of additional bytes of memory
+    # consumption that are allowed in reallocs of start_heap memory.
+    if remaining or start_heap_realloc_delta[0] > 1024:
```
alex, line 136: Can the threshold be set lower?

pkgw: In my conda-forge test, it turns out that there were a couple that realloc'ed 192 additional bytes. So it could drop to 256, say, but not quite as small as I originally thought.

alex: 256 sounds good to me. thanks.

alex (dismissed):

tests/hazmat/backends/test_openssl_memleak.py
pkgw: Actually it looks like some of the Ubuntu tests actually need an even higher number:

```
AssertionError: {'start_heap_realloc_delta': 768}
```

I think what must be happening here is that on Ubuntu the `realloc()` call is returning the same pointer as before, which is the only way I can think of that a `realloc()` wouldn't have triggered issues before. OK if I raise the limit back to 1024?

pkgw: I've gone ahead and returned the setting to 1024.

alex: Oh fascinating, so it's not that this hits a different code path on linux than Windows, it's that on Windows the allocator patterns are different.

I guess this is fine.

alex (approved):

PR #7742, opened 2022-10-25
Silence a new clippy warning that requires a high MSRV

alex:


reaperhulk (approved):
Do we need to backport this to 38.0.x?

alex:
We can do it with next week's release I guess.

PR #7766, opened 2022-11-02
Build vectors and sdist artifacts in CI as well.

alex:
This is in preperation for ultimately uploading them from GHA once PyPI has OIDC integration.

PR #7767, opened 2022-11-02
Turn download_openssl.py into a composite action

alex:


alex:
This is a bit of a bummer for readability (the program is now in a big yaml string), but it makes https://github.com/pyca/cryptography/pull/7766 much easier since you no longer need a clone of the repo around.

alex:
nevermind, this doesn't even do what I hoped apparently

PR #7806, opened 2022-11-14
Increased minimum macOS version to 10.12

alex:


alex:
@maxbelanger It's my annual checkin on macOS versions :-) It looks like dropbox now has 10.12 as a minimum, is that right?

maxbelanger:
@alex yep, can confirm the minimum is now macOS 10.12.

alex:
thank you!

reaperhulk (approved):

Issue #7818, opened 2022-11-17
SHA3: Signature algorithm OID not recognized


lkubb:
Accessing the `signature_hash_algorithm` attribute of a certificate using a SHA3 message digest for the signature hash algorithm, e.g. `RSA_WITH_SHA3_256`, results in `cryptography.exceptions.UnsupportedAlgorithm: Signature algorithm OID: 2.16.840.1.101.3.4.3.14 not recognized`. The certificate has been generated by `cryptography` itself though. This is caused by the *_SHA3 variants not being included in `cryptography.hazmat._oid._SIG_OIDS_TO_HASH`.

To reproduce:

```python
from cryptography import x509

cert_text = """\
-----BEGIN CERTIFICATE-----
MIICyTCCAbGgAwIBAgIUSvAXOvXZih6hXMjQQNhJtqaCZkUwDQYJYIZIAWUDBAMO
BQAwKzELMAkGA1UEBhMCVVMxDTALBgNVBAMMBFRlc3QxDTALBgNVBAoMBFNhbHQw
HhcNMjIxMTE3MTMwOTU2WhcNMjIxMjE3MTMwOTU2WjASMRAwDgYDVQQDDAdzdWNj
ZXNzMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAzIdEbSkbPIc5F/ae
wNoqWPsF/YP+DByMwvYs+0K+lehc39P82fL8K2GIaGMBBzqlsX6CplAzGcoEQEBL
TouLm+JYC5e1zRjaml4M+Bid8A7qwdjlWd0enCAxVB9BMNnj/mokWzh3hAQMBXfm
ddGRhH0P9KXfjBNh2V09vfHdtS9XMEEQjN6vCxaezXqsOMN3bjRTIcUatH7dVUgU
pt9cye1mdbD5KVBgJ9MArc2tJ3rmB0lxjEbAhTEHrNnIkDOJCKE8TaQOW4RyVWlI
vSEL+Ov0TPeXXaef8HJlkyIpKrCZ+c4i0N7qOlyrJEWTXmKNCj87xgTpY5I7ARIS
eOQD8QIDAQABMA0GCWCGSAFlAwQDDgUAA4IBAQCxtTWbOe8HG8ztrw29CacYMK0H
bbWxh2gkWapnCeEUf7Ih/euyibSBb1nx3ZERXfI0TXRAcZ1kytub28CMv1iDnPPI
[...]

PR #7819, opened 2022-11-17
Add *_SHA3 signature algorithms to OID <-> hash map 

lkubb:
This fix allows accessing the `signature_hash_algorithm` attribute of Certificate objects that were signed using some variant of SHA3 as the hashing algorithm.

Fixes: https://github.com/pyca/cryptography/issues/7818

I hope this PR has the expected format and test coverage (if applicable). Thank you for your hard work on `cryptography`. :)


tests/x509/test_x509.py
```diff
@@ -1934,6 +1934,7 @@ def test_build_cert(self, hashalg, hashalg_oid, backend):
 
         assert cert.version is x509.Version.v3
         assert cert.signature_algorithm_oid == hashalg_oid
+        assert cert.signature_hash_algorithm.__class__ == hashalg
```
alex, line 1937:
```suggestion
        assert type(cert.signature_hash_algorithm) is hashalg
```

tests/x509/test_x509.py
```diff
@@ -2824,6 +2825,7 @@ def test_build_cert_with_ec_private_key(
...
```
alex, line 2828:
```suggestion
        assert type(cert.signature_hash_algorithm) is hashalg
```

alex (commented):
Small comments.

Thanks for the PR!

alex (approved):

PR #7850, opened 2022-11-26
add job to automatically mark issues/PRs stale and close them

reaperhulk:
For issues: 5 days until stale, 7 days after that until close. Nothing gets marked in this fashion unless we apply `waiting-on-reporter`.

### DIFF
--- a/.github/workflows/auto-close-stale.yml
+++ b/.github/workflows/auto-close-stale.yml
@@ -1,0 +1,23 @@
+name: Auto-close stale issues
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'
+
+permissions:
+  issues: "write"
+  pull-requests: "write"
+
+jobs:
+  auto-close:
+    if: github.repository_owner == 'pyca'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v6.0.1
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-close-stale.yml
```diff
@@ -0,0 +1,28 @@
...
+  auto-close:
+    if: github.repository_owner == 'pyca'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v6.0.1
+        with:
+          only-labels: waiting-on-reporter
+          days-before-issue-stale: 5
+          days-before-issue-close: 7
+          days-before-pr-stale: 30
```
alex, line 21: I don't think we should auto-close PRs at all.

alex: Which is to say we should never put the label on PRs :-)

alex (approved):

Bittium1:
This pull request adds a new GitHub Actions workflow to automatically close stale issues that are waiting on a reporter response. The workflow uses the `actions/stale` action to identify and close issues based on specific labels and inactivity periods.

Automation for stale issues:

* Added `.github/workflows/auto-close-stale.yml` workflow to auto-close issues labeled `waiting-on-reporter` if there is no response after 5 days, and to close them after 7 days of inactivity, with custom messages for both stages.

PR #7878, opened 2022-12-05
x509: add `load_pem_x509_certificates`

woodruffw:
This behaves similarly to `load_pem_x509_certificate`, except that it returns a list of loaded certificates instead of just the first X.509 certificate.

It raises if the input contains no PEM-encoded certificates, or if PEM or certificate parsing fail.

Signed-off-by: William Woodruff <william@trailofbits.com>


alex (commented):
So the behavior here is to a) accept garbage between the PEMs, b) ignore other PEM blocks in the file that aren't certs?

This seems basically reasonable to me.

Are there other multiple-PEM-in-one-file use cases we want to support In which case it might be better to have some sort of PEM object or something. Otherwise this is probably reasonable.

alex:
flake8

docs/x509/reference.rst
```diff
@@ -168,6 +168,20 @@ Loading Certificates
...
+.. function:: load_pem_x509_certificates(data)
+
+    .. versionadded:: 39.0
+
+    Deserialize one or more certificates from PEM encoded data.
+
+    This is like :func:`~cryptography.x509.load_pem_x509_certificate`, but
+    allows for loading multiple certificates (as adjacent PEMs) at once.
+
+    :returns: list of :class:`~cryptography.x509.Certificate`
```
alex, line 180:
```suggestion
    :returns: ``list`` of :class:`~cryptography.x509.Certificate`
```

tests/x509/test_x509.py
```diff
@@ -5777,3 +5777,24 @@ def test_no_attributes(self, backend):
...
+    assert len(certs) == 2
+
+    certs = load_vectors_from_file(
+        filename=os.path.join(
+            "x509", "cryptography.io.chain_with_garbage.pem"
+        ),
+        loader=lambda pemfile: x509.load_pem_x509_certificates(pemfile.read()),
+        mode="rb",
+    )
+    assert len(certs) == 2
```
alex, line 5800: Please add some basic assertion about the certs.

woodruffw: Just clarifying: do you want me to assert that these are actually certificates, or do some basic attribute checks on them to make sure we're getting something sensible back in each?

woodruffw:
> So the behavior here is to a) accept garbage between the PEMs, b) ignore other PEM blocks in the file that aren't certs?

Yes and yes! That's consistent with how the single PEM cert loader currently works.

> Are there other multiple-PEM-in-one-file use cases we want to support In which case it might be better to have some sort of PEM object or something. Otherwise this is probably reasonable.

Hmm, I can't think of another major use case. Maybe something like loading public/private keypairs in adjacent sequences, but I have no idea whether that's something anybody actually does?

tests/x509/test_x509.py
alex: yeah, check attrs to make sure we got hte right certs -- serial number or something simple.

woodruffw: Sounds good, will add in a moment.

woodruffw: 7bde98a25

PR #7886, opened 2022-12-08
deny tox4 for now

reaperhulk:


reaperhulk:
#7887 is the right fix here.

PR #7887, opened 2022-12-08
fixes for tox4

reaperhulk:
in tox 4 passenv doesn't allow space delimited separators.


gpotter2:
Another tip I found while doing the [same](https://github.com/secdev/scapy/pull/3814) migration: tox 4 apparently changed the way it handled `changedir`. Previously it used to always be relative to the `tox.ini` folder, but not anymore. e.g. if you do `tox -e rust` in `tests/` it will fail.

You need to change it to something like (I used `toxinidir` instead of `tox_root` to be compatible with <4.0.0)...
```
changedir = {toxinidir}/src/rust/
```

alex (approved):

PR #7904, opened 2022-12-17
document POLICY_MAPPING and SUBJECT_DIRECTORY_ATTRIBUTES

mathiasertl:
Found that these two ExtensionOID classes are missing. Not sure if you want them documented, as the implementation of these extensions seems to be missing. Feel free to close if you think it's pointless :-).

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -3403,6 +3403,14 @@ instances. The following common OIDs are available as constants.
 
         Corresponds to the dotted string ``"2.5.29.28"``.
 
+    .. attribute:: POLICY_MAPPINGS
+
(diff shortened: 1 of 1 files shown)


docs/x509/reference.rst
```diff
@@ -3403,6 +3403,14 @@ instances. The following common OIDs are available as constants.
 
         Corresponds to the dotted string ``"2.5.29.28"``.
 
+    .. attribute:: POLICY_MAPPING
```
alex, line 3406:
```suggestion
    .. attribute:: POLICY_MAPPINGS
```

mathiasertl: fixed embarrassing copy/paste fail

alex (approved):

PR #7918, opened 2022-12-20
Remove binding for X509_STORE_get_get_issuer

davidben:
This was added alongside X509_STORE_set_get_issuer in 21ef4080ec7666299f1268f3bbfb136582744359. It makes sense to want to override issuer lookup, but I've never seen a valid use case for querying the callbacks. If you're trying to get at the function that was already configured, you presumably already know it. (And if you don't, it's not safe to call because an arbitrary callback may have been written to only be usable in particular contexts.)

(I got some mysterious "bad magic number" error in a few tests when I tried to test this locally, so relying on CI to fully test.)

### DIFF
--- a/src/_cffi_src/openssl/x509_vfy.py
+++ b/src/_cffi_src/openssl/x509_vfy.py
@@ -211,7 +211,6 @@ Cryptography_STACK_OF_X509_OBJECT *X509_STORE_get0_objects(X509_STORE *);
 X509 *X509_OBJECT_get0_X509(X509_OBJECT *);
 
 X509 *X509_STORE_CTX_get0_cert(X509_STORE_CTX *);
-X509_STORE_CTX_get_issuer_fn X509_STORE_get_get_issuer(X509_STORE *);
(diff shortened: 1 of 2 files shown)


alex:
I think the motivation here would be wanting to add a callback that only
handles some situations, and falling back to the existing callback.

Are you sending this in anticipation of removing it from Boring?

davidben:
Ah. That was my motivation but that's a fair point. Will look more closely at this API and its users and ponder.

(Someone asked about using a similar-looking API to call into the internals standalone, which did not seem a good idea. And it seemed unused other than this.)

davidben:
Looked at it some more, and I'm not sure `X509_STORE_get_get_issuer` is even usable for this fallback use case. If you haven't configured a callback, the getters just return NULL anyway. The default implementation only figures in when the `X509_STORE` stamps an `X509_STORE_CTX`, if you didn't set the callback.

Wanting to simultaneously support AIA and the default implementation makes some sense there but `X509_STORE_CTX_get1_issuer` is actually public API, so you can just fallback directly.

alex:
If no one is using this, I certainly have no objections to deleting

alex (approved):

PR #7926, opened 2022-12-22
Add SSL_CTX_set_msg_callback functions

nveloso:
This PR adds the following function definitions to openssl/ssl.py file:

- SSL_CTX_set_msg_callback
- SSL_CTX_set_msg_callback_arg

Please let me know if anything is missing.
Thank you!

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -296,7 +296,16 @@ void SSL_CTX_set_client_CA_list(SSL_CTX *, Cryptography_STACK_OF_X509_NAME *);
 void SSL_CTX_set_info_callback(SSL_CTX *, void (*)(const SSL *, int, int));
 void (*SSL_CTX_get_info_callback(SSL_CTX *))(const SSL *, int, int);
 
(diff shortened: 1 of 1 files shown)


src/_cffi_src/openssl/ssl.py
```diff
@@ -296,6 +296,9 @@
 void SSL_CTX_set_info_callback(SSL_CTX *, void (*)(const SSL *, int, int));
 void (*SSL_CTX_get_info_callback(SSL_CTX *))(const SSL *, int, int);
 
+void SSL_CTX_set_msg_callback(SSL_CTX *, void (*)(int, int, int, const void *, size_t, SSL *, void *));
```
alex, line 299: You'll need to wrap this line due to length.

nveloso: Is there any linter/formatter that I should follow? Or is there any rules that I should follow?

alex: `tox -e flake` will run everything, but in short: we use black to auto-format, and ruff to enforce some other rules.

nveloso: I has having problem with building cryptography locally and couldn't run `tox`. Hope everything is good now.

alex (approved):

PR #7933, opened 2022-12-25
Migrate x25519 to use rust-openssl

alex:


alex:
Ok, BoringSSL test failures are quite a yak shave:

- `rust-openssl`'s usual binding approach is hand-written, this includes the values of constants (e.g. `EVP_PKEY_X25519`)
- The values it has are from OpenSSL. BoringSSL doesn't use the same ones.
- This causes `EVP_PKEY_CTX_new_id` to fail since its looking for something by the wrong ID.
- https://github.com/sfackler/rust-openssl/commit/e610c0fe013b60c8bb6c438ba6a655588308a232 describes how linking BoringSSL into `rust-openssl` requires special handling.
- Right now we simply treat BoringSSL the same as OpenSSL, so more thought is required here.

alex:
Self-notes:

- Hang is in memleak `test_x25519_pubkey_from_private_key`
- Stack at the point of the hang: x25519 generate -> calls into OpenSSL -> calls into something else in OpenSSL (probably to allocate) -> calls into an ffi callback (probably our memleak malloc) -> tries to acquire the GIL and hangs
- This obviously all works on CPython
- The proximate change here is that with this PR we're not releasing the GIL on generating a key (or anything else for that matter)
  - Unclear if we should release it for key gen; generating 32 random bytes should be very fast
  - Even if we did release the GIL here, it'd just kick the can down the road
- Unclear if having a cffi callback invoked with the GIL held is expected to work or not (though it clearly does on CPython)

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1860,30 +1860,12 @@ def dh_x942_serialization_supported(self) -> bool:
...
-            raise ValueError("An X25519 private key is 32 bytes long")
-
-        data_ptr = self._ffi.from_buffer(data)
-        evp_pkey = self._lib.EVP_PKEY_new_raw_private_key(
-            self._lib.NID_X25519, self._ffi.NULL, data_ptr, len(data)
-        )
-        self.openssl_assert(evp_pkey != self._ffi.NULL)
-        evp_pkey = self._ffi.gc(evp_pkey, self._lib.EVP_PKEY_free)
-        return _X25519PrivateKey(self, evp_pkey)
+        return rust_openssl.x25519.from_private_bytes(bytes(memoryview(data)))
```
reaperhulk, line 1868: Could we do this without a memory copy of the buffer?

src/rust/build.rs
```diff
@@ -3,6 +3,7 @@ use std::io::Write;
 use std::path::Path;
 use std::process::{Command, Stdio};
 
+#[allow(clippy::unusual_byte_groupings)]
```
reaperhulk, line 6: 😂

reaperhulk (approved):

Issue #7934, opened 2022-12-26
Export EVP_PKEY_set_alias_type in FUNCTIONS to support to use SM2 algorithms


hustliyilin:
In openssl 1.1.1 docs/man3/EVP_PKEY_set1_RSA.pod
(https://github.com/openssl/openssl/blob/OpenSSL_1_1_1/doc/man3/EVP_PKEY_set1_RSA.pod)
The EVP_PKEY_set_alias_type(pkey, EVP_PKEY_SM2) API is possible to convert it to using
SM2 algorithms After loading an ECC key.

So the EVP_PKEY_set_alias_type API is important for people who want to use cryptography
with . Therefore, we need to export EVP_PKEY_set_alias_type in FUNCTIONS.

PR #7935, opened 2022-12-26
src/_cffi_src/openssl/evp.py: export EVP_PKEY_set_alias_type in FUNCTIONS  to support to use SM2 algorithms

hustliyilin:
In openssl 1.1.1 docs/man3/EVP_PKEY_set1_RSA.pod
(https://github.com/openssl/openssl/blob/OpenSSL_1_1_1/doc/man3/EVP_PKEY_set1_RSA.pod) The EVP_PKEY_set_alias_type(pkey, EVP_PKEY_SM2) API is possible to convert it to using SM2 algorithms After loading an ECC key.

So the EVP_PKEY_set_alias_type API is important for people who want to use cryptography with using SM2 algorithms. Therefore, we need to export EVP_PKEY_set_alias_type in FUNCTIONS.

Fixes: #7934 
Signed-off-by: YiLin.Li <YiLin.Li@linux.alibaba.com>

### DIFF
--- a/src/_cffi_src/openssl/evp.py
+++ b/src/_cffi_src/openssl/evp.py
@@ -31,6 +31,7 @@ static const int EVP_CTRL_AEAD_SET_IVLEN;
 static const int EVP_CTRL_AEAD_GET_TAG;
 static const int EVP_CTRL_AEAD_SET_TAG;
 
+static const int Cryptography_HAS_EVP_PKEY_set_alias_type;
 static const int Cryptography_HAS_SCRYPT;
 static const int Cryptography_HAS_EVP_PKEY_DHX;
 static const int Cryptography_HAS_EVP_PKEY_get_set_tls_encodedpoint;
@@ -121,6 +122,7 @@ int EVP_PKEY_keygen(EVP_PKEY_CTX *, EVP_PKEY **);
 int EVP_PKEY_derive_init(EVP_PKEY_CTX *);
 int EVP_PKEY_derive_set_peer(EVP_PKEY_CTX *, EVP_PKEY *);
 int EVP_PKEY_derive(EVP_PKEY_CTX *, unsigned char *, size_t *);
+int EVP_PKEY_set_alias_type(EVP_PKEY *, int);
 int EVP_PKEY_set_type(EVP_PKEY *, int);
 
 int EVP_PKEY_id(const EVP_PKEY *);
@@ -179,6 +181,14 @@ EVP_MD_CTX *Cryptography_EVP_MD_CTX_new(void) {
 void Cryptography_EVP_MD_CTX_free(EVP_MD_CTX *md) {
     EVP_MD_CTX_free(md);
 }
+
+#if CRYPTOGRAPHY_IS_LIBRESSL || CRYPTOGRAPHY_OPENSSL_300_OR_GREATER || \
+        CRYPTOGRAPHY_IS_BORINGSSL
+static const int Cryptography_HAS_EVP_PKEY_set_alias_type = 0;
+int (*EVP_PKEY_set_alias_type)(EVP_PKEY *, int) = NULL;
+#else
+static const int Cryptography_HAS_EVP_PKEY_set_alias_type = 1;
+#endif
 
 #if CRYPTOGRAPHY_IS_LIBRESSL || defined(OPENSSL_NO_SCRYPT)
 static const long Cryptography_HAS_SCRYPT = 0;
--- a/src/cryptography/hazmat/bindings/openssl/_conditional.py
+++ b/src/cryptography/hazmat/bindings/openssl/_conditional.py
@@ -42,6 +42,12 @@ def cryptography_has_tls_st() -> typing.List[str]:
     ]
 
 
+def cryptography_has_evp_pkey_set_alias_type() -> typing.List[str]:
+    return [
+        "EVP_PKEY_set_alias_type",
+    ]
+
+
 def cryptography_has_scrypt() -> typing.List[str]:
     return [
         "EVP_PBE_scrypt",
@@ -303,6 +309,9 @@ CONDITIONAL_NAMES = {
     "Cryptography_HAS_SET_CERT_CB": cryptography_has_set_cert_cb,
     "Cryptography_HAS_SSL_ST": cryptography_has_ssl_st,
     "Cryptography_HAS_TLS_ST": cryptography_has_tls_st,
+    "Cryptography_HAS_EVP_PKEY_set_alias_type": (
+        cryptography_has_evp_pkey_set_alias_type
+    ),
     "Cryptography_HAS_SCRYPT": cryptography_has_scrypt,
     "Cryptography_HAS_EVP_PKEY_DHX": cryptography_has_evp_pkey_dhx,
     "Cryptography_HAS_MEM_FUNCTIONS": cryptography_has_mem_functions,


reaperhulk:
This is failing compilation in openssl 3 as well as boring and libre. However, I don’t really want to add random bindings that we don’t use in cryptography itself. There have been some limited attempts to land explicit SM2 support (https://github.com/pyca/cryptography/pull/5894) in cryptography in the past and it would be better to do that rather than add bindings that we’re likely to remove in a future release if we aren’t aware of any consumers.

reaperhulk:
What’s the correct way to do this in openssl 3? Are you interested in contributing a public SM2 API? Or are you using this within pyopenssl (where we should also have public APIs if needed)?

hustliyilin:
> This is failing compilation in openssl 3 as well as boring and libre. However, I don’t really want to add random bindings that we don’t use in cryptography itself. There have been some limited attempts to land explicit SM2 support (#5894) in cryptography in the past and it would be better to do that rather than add bindings that we’re likely to remove in a future release if we aren’t aware of any consumers.

Hi, @reaperhulk ,

Thanks for your review. Now I fix the failing compilation error in openssl 3 as well as boring and libre.  

The mainstream cryptography libraries `openssl`,  `libgcrypt`, `cryptcpp`, and `the Linux kernel` all support SM2. SM2 is widely used in China and is mandatory in many scenarios. So It may be necessary for cryptography to support it for python users.

After applying this patch to cryptography with OpenSSL 1.1.1, we can add the following patch in pyopenssl to use SM2 sign

```C
diff --git a/src/OpenSSL/crypto.py b/src/OpenSSL/crypto.py
index eda4af6..6dd1273 100644
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -76,6 +76,8 @@ TYPE_DSA = _lib.EVP_PKEY_DSA
 TYPE_DH = _lib.EVP_PKEY_DH
 TYPE_EC = _lib.EVP_PKEY_EC

+NID_sm2 = 1172
+EVP_PKEY_SM2 = NID_sm2

 class Error(Exception):
     """
@@ -3016,6 +3018,9 @@ def sign(pkey, data, digest):
     if digest_obj == _ffi.NULL:
         raise ValueError("No such digest method")

+    if _lib.OPENSSL_VERSION_NUMBER < 0x30000000 and _lib.EVP_PKEY_id(pkey._pkey) == _lib.EVP_PKEY_EC :
+        if _lib.EC_GROUP_get_curve_name(_lib.EC_KEY_get0_group(_lib.EVP_PKEY_get1_EC_KEY(pkey._pkey))) == NID_sm2 :
+            _lib.EVP_PKEY_set_alias_type(pkey._pkey, EVP_PKEY_SM2)
     md_ctx = _lib.Cryptography_EVP_MD_CTX_new()
     md_ctx = _ffi.gc(md_ctx, _lib.Cryptography_EVP_MD_CTX_free)

--
2.31.1
```

hustliyilin:
> What’s the correct way to do this in openssl 3? Are you interested in contributing a public SM2 API? Or are you using this within pyopenssl (where we should also have public APIs if needed)?

Hi @reaperhulk ，

As for `OpenSSL 3.x`, The function EVP_PKEY_set_alias_type()  was removed in OpenSSL 3.0 according to the document ( https://github.com/openssl/openssl/blob/master/doc/man3/EVP_PKEY_set1_RSA.pod#history  ). 

C48CAFBB-9382-497b-BF6C-B410B14E6C70

and `OpenSSL 3.x` can use `EVP_PKEY_assign_EC_KEY()` to look at the curve name id to determine if the passed `EC_KEY` is an [SM2(7)](http://man.he.net/man7/SM2) key, and will set the `EVP_PKEY` type to `EVP_PKEY_SM2` in that case, instead of `EVP_PKEY_EC`. according to the [document](https://github.com/openssl/openssl/blob/master/doc/man3/EVP_PKEY_set1_RSA.pod#notes).


image


So we don't need to  modify `cryptography` and `pyopenssl` to use SM2 with `OpenSSL 3.x` any more.

Thanks again @reaperhulk .

hustliyilin:
> > What’s the correct way to do this in openssl 3? Are you interested in contributing a public SM2 API? Or are you using this within pyopenssl (where we should also have public APIs if needed)?
> 
> Hi @reaperhulk ，
> 
> As for `OpenSSL 3.x`, The function `EVP_PKEY_set_alias_type()` was removed in `OpenSSL 3.0` according to the [document](https://github.com/openssl/openssl/blob/master/doc/man3/EVP_PKEY_set1_RSA.pod#history).
> 
> C48CAFBB-9382-497b-BF6C-B410B14E6C70
> 
> and `OpenSSL 3.x` can use `EVP_PKEY_assign_EC_KEY()` to look at the curve name id to determine if the passed `EC_KEY` is an [SM2(7)](http://man.he.net/man7/SM2) key, and will set the `EVP_PKEY` type to `EVP_PKEY_SM2` in that case, instead of `EVP_PKEY_EC`. according to the [document](https://github.com/openssl/openssl/blob/master/doc/man3/EVP_PKEY_set1_RSA.pod#notes).
> 
> image
> 
> And I'll test it as soon as possible with `OpenSSL 3.x` and give you answer.

Hi @reaperhulk ，

With `OpenSSL 3.x`, we don't need to modify `cryptography` and `pyopenssl` any more to use SM2. 
You can following the steps

1. Generate the SM2 key and cert
```shell
openssl ecparam -genkey -name SM2 -out private.pem
openssl req -new -x509 -days 36500 -key private.pem -out cert.crt -sm3 -subj "/C=CN/ST=Zhejiang/L=Hangzhou/O=Alibaba/OU=OS/CN=CA/emailAddress=ca@foo.com"
```

2. Install `cryptography` and `pyopenssl` 

3. use the testcase to sign and verify with sm2-with-sm3 using your SM2 key and cert(`./test.py [private_key_file] [cert_file]`) 

```Python3
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

import sys

from OpenSSL.crypto import load_certificate, load_privatekey
from OpenSSL.crypto import FILETYPE_PEM
from OpenSSL.crypto import (
    sign,
    verify,
)

if __name__ == "__main__":
    if len(sys.argv) != 3:
        print("Usage: ./test.py [private_key_file] [cert_file]")
        exit()

    private_key_file = open(sys.argv[1], 'rb')
    if not private_key_file:
        print("fail to open private_key_file")
        exit()
    private_key = private_key_file.read()

    cert_file = open(sys.argv[2], 'rb')
    if not cert_file:
        print("fail to open cert_file")
        exit()
    cert = cert_file.read()

    priv_key = load_privatekey(FILETYPE_PEM, private_key)
    cert = load_certificate(FILETYPE_PEM, cert)
    content = (
        b"It was a bright cold day in April, and the clocks were striking "
        b"thirteen. Winston Smith, his chin nuzzled into his breast in an "
        b"effort to escape the vile wind, slipped quickly through the "
        b"glass doors of Victory Mansions, though not quickly enough to "
        b"prevent a swirl of gritty dust from entering along with him."
    )
    sig = sign(priv_key, content, "sm3")
    verify(cert, sig, content, "sm3")
```

so we just need to solve the SM2 problem with openssl 1.1.1x. If you have any questions, please let me know.

src/_cffi_src/openssl/evp.py
```diff
@@ -180,6 +180,11 @@
     EVP_MD_CTX_free(md);
 }
 
+#if !CRYPTOGRAPHY_IS_LIBRESSL && !CRYPTOGRAPHY_IS_BORINGSSL && \
+        !CRYPTOGRAPHY_OPENSSL_300_OR_GREATER
+int EVP_PKEY_set_alias_type(EVP_PKEY *, int);
```
alex, line 185: This needs to be conditionally removed, as you can see right below with scrypt -- you'll also need to edit `_conditional.py`.

reaperhulk:
@hustliyilin Do you plan to submit a PR to pyOpenSSL to call this function and add a test to verify it's working as expected? I saw your example code above, but I don't want to merge this unless you plan to finish the work on the other project.

I'd prefer we add SM2 signing to `cryptography` since we generally tell users to only use `pyOpenSSL` for non-TLS uses if `cryptography` is missing a feature, but that's more work.

src/_cffi_src/openssl/evp.py
hustliyilin: > This needs to be conditionally removed, as you can see right below with scrypt -- you'll also need to edit `_conditional.py`.

Thanks for you review. `EVP_PKEY_set_alias_type` is a function only for OpenSSL 1.1.1, And there are failing compilation in openssl 3 as well as boring and libre . So we need to add some conditions to judge the type of openssl. But after adding the conditional judgment，it can't find the corresponding function symbol(`EVP_PKEY_set_alias_type`) in `_openssl.abi3.so` with OpenSSL 1.1.1. Could you give me some advice? Thanks a lot.

hustliyilin:
> compilation error

@reaperhulk Yes.  From the example I gave you can see that I will submit a patch to `pyopenssl`. However from the patch I gave, you can see that `pyopenssl` relies on `cryptography`, and `cryptography` needs to provide the function symbol `EVP_PKEY_set_alias_type` firstly and only then I can initiate the `pyopenssl` PR.

`EVP_PKEY_set_alias_type` and other sm2 functions have been implemented in `openssl`.  `cryptography` can simply export the functions of `openssl` rather than reimplement them. In other words, all we need to do is export corresponding function symbols in `_openssl.abi3.so`

This is what this PR wants to do. But at present, this PR seems to have some problems, and I am still processing and fixing them

src/_cffi_src/openssl/evp.py
reaperhulk: Our bindings require that you put the declaration in the `FUNCTIONS` section. `CUSTOMIZATIONS` is used to handle cases where the declaration may not always work.

You should look at how we do this with `Cryptography_HAS_SCRYPT`. We add `EVP_PBE_scrypt` to `FUNCTIONS`, then create `Cryptography_HAS_SCRYPT` and set its value based on whether the function is available or not (along with a null func ptr declaration if it is not), and then we have a section in `_conditional.py` that removes the symbol from the Python object if it is unavailable.

hustliyilin: > Our bindings require that you put the declaration in the `FUNCTIONS` section. `CUSTOMIZATIONS` is used to handle cases where the declaration may not always work.
> 
> You should look at how we do this with `Cryptography_HAS_SCRYPT`. We add `EVP_PBE_scrypt` to `FUNCTIONS`, then create `Cryptography_HAS_SCRYPT` and set its value based on whether the function is available or not (along with a null func ptr declaration if it is not), and then we have a section in `_conditional.py` that removes the symbol from the Python object if it is unavailable.

Thank you very much for your help @reaperhulk . Now I have modified the codes according to your suggestion, and I can see the function symbol `EVP_PKEY_set_alias_type` in the `_openssl.abi3.so' with `openssl 1.1.1`. Besides, CI/CD shows that it doesn't affect other versions of `openssl`.

a08236ef69a37cb2362c2384904a58b4

In addition, I will submit the `pyopenssl` PR mentioned above to `pyopenssl` as soon as possible in the past two days, and then I will let you know, Thanks again @reaperhulk .

src/_cffi_src/openssl/evp.py
```diff
@@ -180,6 +182,14 @@
     EVP_MD_CTX_free(md);
 }
 
+#if CRYPTOGRAPHY_IS_LIBRESSL || CRYPTOGRAPHY_OPENSSL_300_OR_GREATER || \
+        CRYPTOGRAPHY_IS_BORINGSSL
+static const int Cryptography_HAS_EVP_PKEY_set_alias_type = 0;
+int (*EVP_PKEY_set_alias_type)(EVP_PKEY *, int);
```
reaperhulk, line 188:
```suggestion
int (*EVP_PKEY_set_alias_type)(EVP_PKEY *, int) = NULL;
```

reaperhulk (commented):
Looks like you need to run black + I have one suggested change.

hustliyilin:
> Looks like you need to run black + I have one suggested change.

Thanks a lot. I used `black` to format the codes and all CI/CDs passed.

reaperhulk (approved):

PR #7937, opened 2022-12-26
Update formatting for new ruff

alex:


reaperhulk (approved):
Not a big fan of immediate instability in our new linter.

PR #7938, opened 2022-12-26
Add canonical names to Sphinx class documentation

mathiasertl:
**NOTE:** This MR is not yet complete, I'm not sure if you even want this change in cryptography. So I made a partial MR (what I needed to get this to work in my project) as a proof of concept. If you agree to the change, I will complete the change for t he remaining classes.

Add "canonical names" (that is: the location of the actual implementation) to the Sphinx documentation. The feature was added as part of sphinx-doc/sphinx#4826. Sphinx documents type hints by default and expects the actual type object to serve as object id. 

When documenting 3rd party projects that use cryptography, typehints of functions that use cryptography objects such as `cryptography.x509.base.Certificate` will not link properly with the `intersphinx` extension, because it's documented as `cryptography.x509.Certificate` (without the "base", as I'm sure you know). 

Take for example this (contrived) function:

```python
from cryptography import x509

def demo(param: str) -> x509.Certificate:
    """A dummy function returning a certificate."""
    pass
```

with this audodocs directive (cgdemo.py is the file containing that function):

```rst
.. automodule:: cgdemo
   :members:
```

For sphinx project documentation, all you need is the `sphinx.ext.autodoc` and `sphinx.ext.intersphinx` extensions as well as this intersphinx mapping in `conf.py`:

```python
intersphinx_mapping = {
    "python": ("https://docs.python.org/3/", None),
    "cg": ("https://cryptography.io/en/38.0.4/", None),
}
```

When you generate this sphinx project, the `str` signature will neatly link to https://docs.python.org/3/library/stdtypes.html#str (not that *that* is particularly useful, of course), but the link to `Certificate` (the return value) will not work. If you run `sphinx-build` with `-W -n`, you get a better error description:

```
$ sphinx-build -M clean "." "_build" -W -n 
...                                                                                                                                                       

Warning, treated as error:
/home/mertl/git/3rd/cgdemo/cgdemo.py:docstring of cgdemo.demo:1:py:class reference target not found: cryptography.x509.base.Certificate
```

With my MR, the signature of my 3rd-party project will link to your documentation for cryptography objects properly.

To test locally, generate the sphinx documentation for cryptography and use an updated intersphinx mapping as documented [here](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) under "Multiple targets for the inventory":

```
intersphinx_mapping = {
    "python": ("https://docs.python.org/3/", None),
    "cg": (
        "https://cryptography.io/en/38.0.4/", 
        ("/path/to/your/git//cryptography/docs/_build/html/objects.inv", None),
    ),
}
```

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -150,6 +150,7 @@ Loading Certificates
 ~~~~~~~~~~~~~~~~~~~~
 
 .. function:: load_pem_x509_certificate(data)
+    :canonical: cryptography.x509.base.load_pem_x509_certificate
 
     .. versionadded:: 0.7
 
@@ -169,6 +170,7 @@ Loading Certificates
         2
 
 .. function:: load_pem_x509_certificates(data)
+    :canonical: cryptography.x509.base.load_pem_x509_certificates
 
     .. versionadded:: 39.0
 
@@ -185,6 +187,7 @@ Loading Certificates
         certificate is malformed.
 
 .. function:: load_der_x509_certificate(data)
+    :canonical: cryptography.x509.base.load_der_x509_certificate
 
     .. versionadded:: 0.7
 
@@ -200,6 +203,7 @@ Loading Certificate Revocation Lists
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. function:: load_pem_x509_crl(data)
+    :canonical: cryptography.x509.base.load_pem_x509_crl
 
     .. versionadded:: 1.1
 
@@ -221,6 +225,7 @@ Loading Certificate Revocation Lists
         True
 
 .. function:: load_der_x509_crl(data)
+    :canonical: cryptography.x509.base.load_der_x509_crl
 
     .. versionadded:: 1.1
 
@@ -236,6 +241,7 @@ Loading Certificate Signing Requests
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. function:: load_pem_x509_csr(data)
+    :canonical: cryptography.x509.base.load_pem_x509_csr
 
     .. versionadded:: 0.9
 
@@ -258,6 +264,7 @@ Loading Certificate Signing Requests
         True
 
 .. function:: load_der_x509_csr(data)
+    :canonical: cryptography.x509.base.load_der_x509_csr
 
     .. versionadded:: 0.9
 
@@ -273,6 +280,7 @@ X.509 Certificate Object
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. class:: Certificate
+    :canonical: cryptography.x509.base.Certificate
 
     .. versionadded:: 0.7
 
@@ -517,6 +525,7 @@ X.509 CRL (Certificate Revocation List) Object
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. class:: CertificateRevocationList
+    :canonical: cryptography.x509.base.CertificateRevocationList
 
     .. versionadded:: 1.0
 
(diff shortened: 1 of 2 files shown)


alex (commented):
Hmm, so I'm not sure I understand what canonical means. Specifically, is it exposed to human users in any way?

We do not guarantee these various names, the only thing we guarantee is the publicly documented names where these things are available. So if a human were to see the canonical name, that'd be a no from us, since it'd simply they're stable. But if this is purely behind the scenes then it may be acceptable.

mathiasertl:
> Hmm, so I'm not sure I understand what canonical means.

I completely agree that the name "canonical" here is confusing. I haven't invented it, it's the term used in the directive and in the linked MR (and by the way: why can't sphinx automatically add that :frowning:). I'm sorry, I could have elaborated further.

What it refers to here is the location where a class is actually implemented.

> We do not guarantee these various names, the only thing we guarantee is the publicly documented names where these things are available.

That's of course a given!

> Specifically, is it exposed to human users in any way?
> 
> ... So if a human were to see the canonical name, that'd be a no from us, since it'd simply they're stable. But if this is purely behind the scenes then it may be acceptable.

The cryptography documentation itself does not show the "canonical" name. Besides a visual confirmation, I did a grep in the HTML build directory (after building, of course):

```
mertl@pcn97:~/git/3rd/cryptography$ grep -r cryptography.x509.base.Certificate docs/_build/
docs/_build/html/_sources/x509/reference.rst.txt:   :canonical: cryptography.x509.base.Certificate
docs/_build/html/_sources/x509/reference.rst.txt:   :canonical: cryptography.x509.base.CertificateRevocationList
docs/_build/html/_sources/x509/reference.rst.txt:    :canonical: cryptography.x509.base.CertificateBuilder
docs/_build/html/_sources/x509/reference.rst.txt:    :canonical: cryptography.x509.base.CertificateSigningRequest
Binary file docs/_build/doctrees/environment.pickle matches
```

This is how the external project shows the signature. Certificate is a link to https://cryptography.io/en/38.0.4/x509/reference.html#cryptography.x509.Certificate :

Screenshot from 2022-12-26 17-11-43

alex:
Then this sounds good to me -- one important question: Does sphinx complain if a canonical directive is _wrong_ (i.e. if we move where something is implemented)?

mathiasertl:
After a quick check: no it doesn't, unfortunately.

I have found one downside to this MR - I want to be transparent about it: If a third-party project sets `autodoc_typehints_format = "fully-qualified"` in Sphinx `conf.py`, the "canonical name" will show up in the text. But of course, only in the third-party project that explicitly sets that directive (and that's kind of what fully qualified means, so I guess that could be okay).

Here's a screenshot, again only looks like this with `autodoc_typehints_format = "fully-qualified"`:

Screenshot from 2022-12-26 17-43-58

FTR, the `:canonical:` option is also documented here and that documentation confirms it's intended for that usecase: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-option-py-class-canonical

mathiasertl:
> After a quick check: no it doesn't, unfortunately.

Not sure if it's worth trying to add tests for this. I could try to hack something together from the MR that implemented `:canonical:`: https://github.com/sphinx-doc/sphinx/pull/9026/files

alex:
I'm ok to land it as-is, with the understanding that its entirely possible we'll break it since sphinx doesn't enforce it.

mathiasertl:
That's good to hear - here you go then!

By the way: 

```
    .. versionadded:: 0.7
```

... isn't that a little excessive? :-)

alex:
We don't remove them. I suppose we could, not sure it matters.

docs/x509/reference.rst
```diff
@@ -1657,6 +1688,7 @@ X.509 Extensions
             A bytes string representing the extension's DER encoded value.
 
 .. class:: KeyUsage(digital_signature, content_commitment, key_encipherment, data_encipherment, key_agreement, key_cert_sign, crl_sign, encipher_only, decipher_only)
+    :canonical: cryptography.x509.extensions.
```
alex, line 1691: Typo.

alex (changes requested):

docs/x509/reference.rst
mathiasertl: fixed

alex (approved):

PR #7944, opened 2022-12-28
Replace more deprecated `abstractproperty`

Viicos:
Follow up on https://github.com/pyca/cryptography/pull/7943 (for some reason, forgot to replace some of them).

As a side note (stated @https://github.com/jpadilla/pyjwt/pull/843#issuecomment-1366190629), I think `HashAlgorithm` implementations should use the `property` decorator, e.g.:

```python
class SHA256(HashAlgorithm):
    @property
    def name() -> str:
        return "sha256"
    ...
```

But imo another pattern should be used, as these classes are only storing data (perhaps dataclasses).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,6 @@ tests =[
 [tool.coverage.report]
 exclude_lines = [
     "@abc.abstractmethod",
-    "@abc.abstractproperty",
     "@typing.overload",
     "if typing.TYPE_CHECKING",
 ]
--- a/src/cryptography/hazmat/primitives/_asymmetric.py
+++ b/src/cryptography/hazmat/primitives/_asymmetric.py
@@ -9,7 +9,8 @@ import abc
 
 
 class AsymmetricPadding(metaclass=abc.ABCMeta):
-    @abc.abstractproperty
+    @property
+    @abc.abstractmethod
     def name(self) -> str:
         """
         A string naming this padding (e.g. "PSS", "PKCS1").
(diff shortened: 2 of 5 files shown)


reaperhulk:
dataclasses were added in 3.7 so we can’t consider them until we drop 3.6 support (which we’re hoping to do as Ubuntu 18.04 drops off in the next 6 months). The mypy issues are unfortunate 😢

reaperhulk (approved):
Incidentally the reason we left this with deprecated annotations originally was due to coverage issues with multiple decorators IIRC. Glad to see they seem to be resolved!

Viicos:
> dataclasses were added in 3.7 so we can’t consider them until we drop 3.6 support (which we’re hoping to do as Ubuntu 18.04 drops off in the next 6 months). The mypy issues are unfortunate 😢

Ah yes forgot about that. The mypy issue can be fixed on the mentioned repo by using an instance of the hash algorithm instead of the class type. Still, is it valid to override properties the way it is done currently (that is overriding as a simple class attribute)? Apart from dataclasses, I was thinking about using `ClassVar` in the abc class:

```python
class HashAlgorithm(metaclass=abc.ABCMeta):
    name: ClassVar[str]
    """
    A string naming this algorithm (e.g. "sha256", "md5").
    """
    digest_size: ClassVar[int]
    """
    The size of the resulting digest in bytes.
    """
    block_size: ClassVar[Optional[int]]
    """
    The internal block size of the hash function, or None if the hash
    function does not use blocks internally (e.g. SHA3).
    """
```

But then this would be breaking for hashes with a `digest_size` that can be initialized (e.g. `SHAKE128`). I guess it's better to wait for dataclasses support, and this is not a big issue anyway.

PR #7960, opened 2023-01-02
support SSH certificate parsing

reaperhulk:


docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -459,6 +459,187 @@ An example ECDSA key in OpenSSH format::
...
+
+    .. versionadded:: 40.0
+
+    .. note::
+
+        This function does not support parsing certificates with DSA public
+        keys or signatures from DSA certificate authorities. DSA is a
+        deprecated algorithm and should not be used.
+
+    Deserialize a public key from OpenSSH (`PROTOCOL.certkeys`_) encoded data
```
alex, line 479: Describing it as deserializing a public key is confusing. It's deserializing a cert... which contains a pub key.

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -459,6 +459,187 @@ An example ECDSA key in OpenSSH format::
...
+        :type: bytes
+
+        The nonce field is a CA-provided random value of arbitrary length
+        (but typically 16 or 32 bytes) included to make attacks that depend on
+        inducing collisions in the signature hash infeasible.
+
+    .. method:: public_key()
+
+        :class:`~cryptography.hazmat.primitives.asymmetric.rsa.RSAPublicKey`,
+        :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPublicKey`,
```
alex, line 510: No DSA!

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -459,6 +459,187 @@ An example ECDSA key in OpenSSH format::
...
+
+        A naïve datetime representing the UTC time before which the certificate
+        is valid. **This time is not inclusive.**
+
+    .. attribute:: valid_after
+
+        :type: :class:`datetime.datetime`
+
+        A naïve datetime representing the UTC time after which the certificate
+        is valid. **This time is inclusive.**
```
alex, line 564: Swap the order of these.

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -459,6 +459,187 @@ An example ECDSA key in OpenSSH format::
...
+
+    .. attribute:: critical_options
+
+        :type: dict[bytes, bytes]
+
+        Critical options is a dict of zero or more options that are
+        considered critical for the certificate to be considered valid. If
```
alex, line 571: "considered critical" is a weird way of saying this.

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -459,6 +459,187 @@ An example ECDSA key in OpenSSH format::
...
+
+        :type: bytes
+
+        The signature of the certificate. SSH signature serialization
+        formats do not always match the signature format used by
+        this library's other signature interfaces so if you need
+        to validate that the SSH certificate signature matches the
+        embedded signature key you should use
+        :meth:`check_cert_signature`.
+
```
alex, line 604: Given we have `check_cert_signature`, do users need this?

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -459,6 +459,187 @@ An example ECDSA key in OpenSSH format::
...
+    .. method:: check_cert_signature()
```
alex, line 605:
```suggestion
    .. method:: verify_cert_signature()
```

?

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -626,6 +620,27 @@ def load_ssh_private_key(
     return private_key
 
 
+def _get_ssh_key_type(
```
alex, line 623: why move

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -703,6 +718,256 @@ def _serialize_ssh_private_key(
...
+
+    @property
+    def key_id(self) -> bytes:
+        return bytes(self._key_id)
+
+    @utils.cached_property
+    def valid_principals(self) -> typing.List[bytes]:
+        rest = self._valid_principals
+        principals = []
+        while len(rest) > 0:
```
alex, line 799:
```suggestion
        while rest:
```

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -703,6 +718,256 @@ def _serialize_ssh_private_key(
...
+
+    @utils.cached_property
+    def extensions(self) -> typing.Dict[bytes, bytes]:
+        return self._parse_exts_opts(self._extensions)
+
+    def _parse_exts_opts(
+        self, exts_opts: memoryview
+    ) -> typing.Dict[bytes, bytes]:
+        result: typing.Dict[bytes, bytes] = {}
+        while len(exts_opts) > 0:
```
alex, line 825:
```suggestion
        while exts_opts:
```

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -703,6 +718,256 @@ def _serialize_ssh_private_key(
...
+            name, exts_opts = _get_sshstr(exts_opts)
+            bname: bytes = bytes(name)
+            if bname in result.keys():
```
alex, line 828:
```suggestion
            if bname in result:
```

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -703,6 +718,256 @@ def _serialize_ssh_private_key(
...
+                raise ValueError("Duplicate extension name")
+            # dicts in pypy & python 3.6+ are insertion ordered so hooray
+            if len(result) > 0 and bname < list(result.keys())[-1]:
+                raise ValueError("Extensions not lexically sorted")
```
alex, line 832: This is way too complicatd. Just store the previous value in a local key.

The way you've written this makes the algorithm O(n^2).

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -703,6 +718,256 @@ def _serialize_ssh_private_key(
...
+            signature_key.verify(
+                computed_sig, bytes(self._tbs_cert_body), ec.ECDSA(hash_alg)
+            )
+        else:
+            assert isinstance(signature_key, rsa.RSAPublicKey)
+            signature_key.verify(
+                bytes(self.signature),
+                bytes(self._tbs_cert_body),
+                padding.PKCS1v15(),
+                hashes.SHA512(),
```
alex, line 878: I don't think this is right, you have to get the hash alg from somewhere.

docs/hazmat/primitives/asymmetric/serialization.rst
reaperhulk: If someone wanted to implement it themselves they'd need access to this and the unsigned body (which isn't currently exposed). We either need to expose both, or neither. I'm fine choosing neither until someone asks for it.

alex: I vote neither

docs/hazmat/primitives/asymmetric/serialization.rst
reaperhulk: I'm vaguely worried about what `verify` means, but we've loudly documented what the function does and it absolutely does verify a signature against a key, so I'm okay with changing it.

mjpieters:
Is there any reason this has to be a separate function?

It feels to me like `load_ssh_public_key()` can can return a plain key or a certificate, because the certificate itself is also a public key. This is how the [go `crypto` library handles parsing](https://pkg.go.dev/golang.org/x/crypto/ssh#ParseAuthorizedKey), it returns a [`PublicKey` _interface_](https://pkg.go.dev/golang.org/x/crypto/ssh#PublicKey) (the Go equivalent of an ABC), and [`Certificate`](https://pkg.go.dev/golang.org/x/crypto/ssh#Certificate) is one implementation of the interface.

If I have a pile of SSH public keys as files, I'd much rather use:

```
keys: Iterable[bytes]
for keydata in keys:
    key = ssh.load_ssh_public_key(keyfile)
    if isinstance(key, SSHCertificate):
        # do something with a certificate, like verifying it against a private key
    else:
        # do something with a non-cert public key, like checking it against a authorized keys store
```

than have to catch the `ValueError` and check that it signalled that the key wasn't a certificate key:

```
keys: Iterable[bytes]
for keydata in keys:
    try:
        key = ssh.load_ssh_public_key(keyfile)
        # do something with a certificate, like verifying it against a private key
    except ValueError as ex:
        if ex.args[0] != "Not an SSH certificate":
            raise
    else:
        # do something with a non-cert public key, like checking it against a authorized keys store
```

If this must be a separate function, at least give us a better way to distinguish certificate public keys from non-certificate public keys.

reaperhulk:
There's no hard requirement that this has to be a separate function, but having it return `SSHCertificate` for certs would represent a breaking API change. We could instead make the new function `load_ssh_public_key_or_cert` (or some better name) though.

mjpieters:
> There's no hard requirement that this has to be a separate function, but having it return `SSHCertificate` for certs would represent a breaking API change. We could instead make the new function `load_ssh_public_key_or_cert` (or some better name) though.

A, yes, code that currently expects the unwrapped public key without certificate metadata is going to be unhappy. However, wouldn't a default-False option on the function not be better option then? E.g., expressed as overloads:

```
@typing.overload
def load_ssh_public_key(
    data: bytes, backend: typing.Any = None, full_certificate: LIteral[False] = ...
) -> _SSH_PUBLIC_KEY_TYPES: ...

@typing.overload
def load_ssh_public_key(
    data: bytes, backend: typing.Any = None, full_certificate: Literal[True] = ...
) -> types.Union[_SSH_PUBLIC_KEY_TYPES, SSHCertificate]: ...

@typing.overload
def load_ssh_public_key(
    data: bytes, backend: typing.Any = None, full_certificate: bool.
) -> types.Union[_SSH_PUBLIC_KEY_TYPES, SSHCertificate]: ...

def load_ssh_public_key(
    data: bytes, backend: typing.Any = None, full_certificate: bool = False
) -> types.Union[_SSH_PUBLIC_KEY_TYPES, SSHCertificate]:
    ...
```

(not married to `full_certificate` as the name of the flag, could very much be improved upon).

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -459,6 +459,174 @@ An example ECDSA key in OpenSSH format::
...
         key is of a type that is not supported.
 
+
+OpenSSH Certificate
+~~~~~~~~~~~~~~~~~~~
+
+The format used by OpenSSH for certificates, as specified in
+`PROTOCOL.certkeys`_.
+
+.. function:: load_ssh_certificate(data)
```
alex, line 469: Where'd be land a unified API?

docs/hazmat/primitives/asymmetric/serialization.rst
```diff
@@ -459,6 +459,174 @@ An example ECDSA key in OpenSSH format::
...
+    .. note::
+
+        This function does not support parsing certificates with DSA public
+        keys or signatures from DSA certificate authorities. DSA is a
+        deprecated algorithm and should not be used.
+
+    Deserialize an OpenSSH certificate (specified in `PROTOCOL.certkeys`_)
+    to an instance of :class:`SSHCertificate`. If you only need the public key
+    you can parse it out of the certificate using :func:`load_ssh_public_key`.
+
```
alex, line 482: We should mention that it doesn't verify anything, it just parses it.

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -703,6 +718,262 @@ def _serialize_ssh_private_key(
...
+    def key_id(self) -> bytes:
+        return bytes(self._key_id)
+
+    @utils.cached_property
+    def valid_principals(self) -> typing.List[bytes]:
+        rest = self._valid_principals
+        principals = []
+        while rest:
+            principal, rest = _get_sshstr(rest)
+            principals.append(bytes(principal))
```
alex, line 803: Lazily parsing this means there could be syntax errors in it that we don't catch early, right? Is there a reason to be lazy?

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -703,6 +718,262 @@ def _serialize_ssh_private_key(
...
+        self, exts_opts: memoryview
+    ) -> typing.Dict[bytes, bytes]:
+        result: typing.Dict[bytes, bytes] = {}
+        last_name = None
+        while exts_opts:
+            name, exts_opts = _get_sshstr(exts_opts)
+            bname: bytes = bytes(name)
+            if bname in result:
+                raise ValueError("Duplicate extension name")
+            # dicts in pypy & python 3.6+ are insertion ordered so hooray
```
alex, line 833: lol

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -703,6 +718,262 @@ def _serialize_ssh_private_key(
...
+
+def load_ssh_certificate(data: bytes) -> SSHCertificate:
+    utils._check_byteslike("data", data)
+
+    m = _SSH_PUBKEY_RC.match(data)
+    if not m:
+        raise ValueError("Invalid line format")
+    key_type = orig_key_type = m.group(1)
+    key_body = m.group(2)
+    if _CERT_SUFFIX != key_type[-len(_CERT_SUFFIX) :]:
```
alex, line 907: Can this just be `if not key_type.endswith(_CERT_SUFFIX)`?

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -703,6 +718,262 @@ def _serialize_ssh_private_key(
...
+        raise ValueError("Not an SSH certificate")
+    else:
+        key_type = key_type[: -len(_CERT_SUFFIX)]
```
alex, line 910: No need for an `else` block

tests/hazmat/primitives/test_ssh.py
```diff
@@ -602,3 +605,218 @@ def test_dsa_private_key_sizes(self, key_path, supported, backend):
...
+                "OpenSSH",
+                "certs",
+                "p256-p256-empty-principals.pub",
+            ),
+            lambda f: f.read(),
+            mode="rb",
+        )
+        data = bytearray(data)
+        # mutate the signature so it's invalid
+        data[-10] = 70
```
alex, line 689: Can we do this with all the cert algorithms?

src/cryptography/hazmat/primitives/serialization/ssh.py
reaperhulk: Yes there could be errors. The only reason to be lazy is to avoid the cost if not accessed. That said, the only expensive operation is loading a key so we could leave that one lazy and change the rest.

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -703,6 +718,269 @@ def _serialize_ssh_private_key(
...
+def _get_ec_hash_alg(curve: ec.EllipticCurve) -> hashes.HashAlgorithm:
+    hash_alg: hashes.HashAlgorithm
+    if isinstance(curve, ec.SECP256R1):
+        hash_alg = hashes.SHA256()
+    elif isinstance(curve, ec.SECP384R1):
+        hash_alg = hashes.SHA384()
+    else:
+        assert isinstance(curve, ec.SECP521R1)
+        hash_alg = hashes.SHA512()
+    return hash_alg
```
alex, line 872:
```suggestion
    if isinstance(curve, ec.SECP256R1):
        return hashes.SHA256()
    elif isinstance(curve, ec.SECP384R1):
        return hashes.SHA384()
    else:
        assert isinstance(curve, ec.SECP521R1)
        return hashes.SHA512()
```

alex (approved):

reaperhulk:
@mjpieters Trying to align our types to allow Go-like interfaces here would be very difficult and would result in odd situations like `verify` on `SSHCertificate` meaning something entirely different from `verify` on a public key. There's also no meaning to encoding formats for an SSH certificate as best I can tell since it only has one?

You've definitely convinced me that we should support both in a single function though, so this PR has been updated to implement that as `load_ssh_public_identity`. A future PR will implement `load_ssh_public_key` in terms of this function to simplify the code, but we'll leave both.

alex (dismissed):

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -703,6 +718,267 @@ def _serialize_ssh_private_key(
...
+                bytes(self._signature),
+                bytes(self._tbs_cert_body),
+                padding.PKCS1v15(),
+                hash_alg,
+            )
+
+
+def _get_ec_hash_alg(curve: ec.EllipticCurve) -> hashes.HashAlgorithm:
+    if isinstance(curve, ec.SECP256R1):
+        return hashes.SHA256()
```
alex, line 865: apparently this is uncovered?

alex (approved):

PR #8009, opened 2023-01-08
deprecate support for DSA in load_ssh_public_key

reaperhulk:


src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -875,6 +875,7 @@ def _get_ec_hash_alg(curve: ec.EllipticCurve) -> hashes.HashAlgorithm:
 
 def load_ssh_public_identity(
     data: bytes,
+    _legacy_dsa_allowed=False,
```
alex, line 878: Instead of a private param on a public API, just make both functions delegate to a private function that has the param. Reduces the odds a user will depend on it.

alex (approved):

PR #8011, opened 2023-01-08
Add Certificate.verify_signed_by

reaperhulk:
Verify that the signature on a certificate was created by the private key belonging to another certificate's public key.

This code does not validate anything else! It is not a path builder, general x509 validator, etc.

### DIFF
--- a/src/rust/src/x509/certificate.rs
+++ b/src/rust/src/x509/certificate.rs
@@ -320,25 +320,28 @@ impl Certificate {
         )
     }
 
-    fn verify_issued_by<'p>(
+    fn verify_directly_issued_by<'p>(
         &self,
         py: pyo3::Python<'p>,
         issuer: pyo3::PyRef<'_, Certificate>,
     ) -> PyAsn1Result<&'p pyo3::PyAny> {
-        let public_key = issuer.public_key(py)?;
-        if self.raw.borrow_value().tbs_cert.issuer != issuer.raw.borrow_value().tbs_cert.subject {
+        let signature_oid = &self.raw.borrow_value().signature_alg.oid;
+        if &self.raw.borrow_value().tbs_cert.signature_alg.oid != signature_oid {
             return Err(PyAsn1Error::from(pyo3::exceptions::PyValueError::new_err(
-                "Issuer certificate subject does not match certificate issuer",
+                "Inner and outer signature algorithms do not match. This is an invalid certificate."
             )));
         };
-        let signature_oid = &self.raw.borrow_value().signature_alg.oid;
+        if self.raw.borrow_value().tbs_cert.issuer != issuer.raw.borrow_value().tbs_cert.subject {
+            return Err(PyAsn1Error::from(pyo3::exceptions::PyValueError::new_err(
+                "Issuer certificate subject does not match certificate issuer.",
+            )));
+        };
+        let public_key = issuer.public_key(py)?;
         let tbs_bytes = asn1::write_single(&self.raw.borrow_value().tbs_cert)?;
-        let signature_hash = self.signature_hash_algorithm(py)?;
-        sign::verify_cert_issuer(
+        sign::verify_signature_with_oid(
             py,
             public_key,
             signature_oid,
-            signature_hash,
             self.raw.borrow_value().signature.as_bytes(),
             &tbs_bytes,
         )
--- a/src/rust/src/x509/certificate.rs
+++ b/src/rust/src/x509/certificate.rs
@@ -339,7 +339,7 @@ impl Certificate {
             public_key,
             signature_oid,
             signature_hash,
-            &self.raw.borrow_value().signature.as_bytes(),
+            self.raw.borrow_value().signature.as_bytes(),
             &tbs_bytes,
         )
     }
--- a/src/rust/src/x509/certificate.rs
+++ b/src/rust/src/x509/certificate.rs
@@ -7,7 +7,7 @@ use crate::asn1::{
     PyAsn1Error, PyAsn1Result,
 };
 use crate::x509;
-use crate::x509::{crl, extensions, oid, sct, Asn1ReadableOrWritable};
+use crate::x509::{crl, extensions, oid, sct, sign, Asn1ReadableOrWritable};
 use chrono::Datelike;
 use pyo3::ToPyObject;
 use std::collections::hash_map::DefaultHasher;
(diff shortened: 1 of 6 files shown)


reaperhulk:
#5116 attempted to implement something similar, but with added issuer/subject name matching. We should decide whether we want to have this function do that type of matching (and thus be more comfortable saying "issued by") or stick to strictly signature verification.

reaperhulk:
There is one bit of coverage jank and one legitimate coverage miss. I'm open to suggestions for how to deal with both 😄

src/rust/src/x509/certificate.rs
```diff
@@ -319,6 +319,98 @@ impl Certificate {
             },
         )
     }
+
+    fn verify_signed_by<'p>(
+        &self,
+        py: pyo3::Python<'p>,
+        issuer: pyo3::PyRef<'_, Certificate>,
+    ) -> pyo3::PyResult<&'p pyo3::PyAny> {
+        let public_key = issuer.public_key(py)?;
```
alex, line 328: I think the core of this logic should go in `sign.rs` for symmetry -- it can be a method that takes the public key, tbs, and signature bytes.

src/rust/src/x509/certificate.rs
```diff
@@ -319,6 +319,98 @@ impl Certificate {
...
+        let key_type = identify_public_key_type(py, public_key)?;
+        let hash_alg = self.signature_hash_algorithm(py)?;
```
alex, line 330: We need to check that the signature algorithm actually corresponds to the key type, don't we? Best to do that by OID I think.

tests/x509/test_x509.py
```diff
@@ -77,6 +81,46 @@ def _load_cert(filename, loader: typing.Callable[..., T], backend=None) -> T:
     return cert
 
 
+def _generate_ca_and_leaf(issuer_private_key, subject_private_key):
```
alex, line 84: type annotations

src/rust/src/x509/certificate.rs
```diff
@@ -319,6 +319,98 @@ impl Certificate {
             },
         )
     }
+
+    fn verify_signed_by<'p>(
```
alex, line 323: Should we be checking that issuer.subject matches our subject?

alex (commented):
Also needs docs and to add the method to the interface of course.

src/rust/src/x509/certificate.rs
```diff
@@ -319,6 +319,98 @@ impl Certificate {
...
+            }
+            sign::KeyType::Dsa => public_key.call_method1(
+                "verify",
+                (
+                    self.signature(py),
+                    self.tbs_certificate_bytes(py)?,
+                    hash_alg,
+                ),
+            ),
+        }?;
```
alex, line 369: Put the `?` on each of the individual `call_method` calls, much easier to read

src/rust/src/x509/certificate.rs
reaperhulk: I asked this question in a comment on the PR too 😄 I'm fine with doing that, although then maybe this should be called `verify_issued_by`?

src/rust/src/x509/certificate.rs
reaperhulk: I originally did that, but ran into rust coverage problems. Of course, this form didn't fix them, just reduced it to one line of incorrect missed coverage.

alex: that coverage issue would indicate missing failed sigs on one of the key types... which we shoudl test

src/rust/src/x509/sign.rs
```diff
@@ -256,3 +258,142 @@ pub(crate) fn sign_data<'p>(
...
+    } else if dsa_key_type.is_instance(public_key)? {
+        Ok(KeyType::Dsa)
+    } else if ec_key_type.is_instance(public_key)? {
+        Ok(KeyType::Ec)
+    } else if ed25519_key_type.is_instance(public_key)? {
+        Ok(KeyType::Ed25519)
+    } else if ed448_key_type.is_instance(public_key)? {
+        Ok(KeyType::Ed448)
+    } else {
+        Err(pyo3::exceptions::PyTypeError::new_err(
```
reaperhulk, line 339: Coverage failing because this is unreachable from Python code (`signature_hash_algorithm` calls catch any error before this can be invoked). I'm structuring this incorrectly clearly 😄

src/rust/src/x509/certificate.rs
```diff
@@ -319,6 +319,33 @@ impl Certificate {
...
         )
     }
+
+    fn verify_directly_issued_by<'p>(
+        &self,
+        py: pyo3::Python<'p>,
+        issuer: pyo3::PyRef<'_, Certificate>,
+    ) -> PyAsn1Result<&'p pyo3::PyAny> {
+        let signature_oid = &self.raw.borrow_value().signature_alg.oid;
+        if &self.raw.borrow_value().tbs_cert.signature_alg.oid != signature_oid {
```
alex, line 329: Should we be comparing the entire algorithm identifier?

src/rust/src/x509/sign.rs
```diff
@@ -256,3 +259,265 @@ pub(crate) fn sign_data<'p>(
     };
     signature.extract()
 }
+
+fn py_hash_from_hash_type(hash_type: HashType) -> Option<&'static str> {
```
alex, line 263:
```suggestion
fn py_hash_name_from_hash_type(hash_type: HashType) -> Option<&'static str> {
```

src/rust/src/x509/sign.rs
```diff
@@ -256,3 +259,265 @@ pub(crate) fn sign_data<'p>(
...
+    data: &[u8],
+) -> PyAsn1Result<&'p pyo3::PyAny> {
+    let key_type = identify_public_key_type(py, issuer_public_key)?;
+    let (sig_key_type, sig_hash_type) = identify_key_hash_type_for_oid(py, signature_oid)?;
+    let sig_hash_str = py_hash_from_hash_type(sig_hash_type);
+    let hashes = py.import("cryptography.hazmat.primitives.hashes")?;
+    let signature_hash = match sig_hash_str {
+        Some(data) => hashes.getattr(data)?.call0()?,
+        None => py.None().into_ref(py),
+    };
```
alex, line 291: Can we put this python logic after the `key_type != sig_key_type` block

src/rust/src/x509/sign.rs
```diff
@@ -256,3 +259,265 @@ pub(crate) fn sign_data<'p>(
...
+    }
+}
+
+pub(crate) fn verify_signature_with_oid<'p>(
+    py: pyo3::Python<'p>,
+    issuer_public_key: &'p pyo3::PyAny,
+    signature_oid: &asn1::ObjectIdentifier,
+    signature: &[u8],
+    data: &[u8],
+) -> PyAsn1Result<&'p pyo3::PyAny> {
```
alex, line 283:
```suggestion
) -> PyAsn1Result<()> {
```

No need for this to return anything

src/rust/src/x509/certificate.rs
```diff
@@ -319,6 +319,33 @@ impl Certificate {
             },
         )
     }
+
+    fn verify_directly_issued_by<'p>(
+        &self,
+        py: pyo3::Python<'p>,
+        issuer: pyo3::PyRef<'_, Certificate>,
+    ) -> PyAsn1Result<&'p pyo3::PyAny> {
```
alex, line 327:
```suggestion
    ) -> PyAsn1Result<()> {
```

src/rust/src/x509/sign.rs
```diff
@@ -256,3 +259,265 @@ pub(crate) fn sign_data<'p>(
...
+    } else if ec_key_type.is_instance(public_key)? {
+        Ok(KeyType::Ec)
+    } else if ed25519_key_type.is_instance(public_key)? {
+        Ok(KeyType::Ed25519)
+    } else if ed448_key_type.is_instance(public_key)? {
+        Ok(KeyType::Ed448)
+    } else {
+        Err(pyo3::PyErr::from_instance(
+            py.import("cryptography.exceptions")?.call_method1(
+                "UnsupportedAlgorithm",
```
alex, line 361: Is this how we usually use this algorithm?

src/rust/src/x509/sign.rs
```diff
@@ -256,3 +259,265 @@ pub(crate) fn sign_data<'p>(
...
+        oid::ECDSA_WITH_SHA3_512_OID => Ok((KeyType::Ec, HashType::Sha3_512)),
+        oid::ED25519_OID => Ok((KeyType::Ed25519, HashType::None)),
+        oid::ED448_OID => Ok((KeyType::Ed448, HashType::None)),
+        oid::DSA_WITH_SHA224_OID => Ok((KeyType::Dsa, HashType::Sha224)),
+        oid::DSA_WITH_SHA256_OID => Ok((KeyType::Dsa, HashType::Sha256)),
+        oid::DSA_WITH_SHA384_OID => Ok((KeyType::Dsa, HashType::Sha384)),
+        oid::DSA_WITH_SHA512_OID => Ok((KeyType::Dsa, HashType::Sha512)),
+        _ => Err(pyo3::PyErr::from_instance(
+            py.import("cryptography.exceptions")?.call_method1(
+                "UnsupportedAlgorithm",
```
alex, line 397: Same Q here.

src/rust/src/x509/sign.rs
```diff
@@ -256,3 +259,265 @@ pub(crate) fn sign_data<'p>(
...
+                identify_key_hash_type_for_oid(py, &oid::DSA_WITH_SHA512_OID).unwrap(),
+                (KeyType::Dsa, HashType::Sha512)
+            );
+            assert!(identify_key_hash_type_for_oid(py, &oid::TLS_FEATURE_OID).is_err());
+        })
+    }
+
+    #[test]
+    fn test_py_hash_from_hash_type() {
+        for item in [
```
alex, line 507:
```suggestion
        for (hash, name) in [
```

src/rust/src/x509/sign.rs
```diff
@@ -256,3 +259,265 @@ pub(crate) fn sign_data<'p>(
...
+            (HashType::Sha512, "SHA512"),
+            (HashType::Sha3_224, "SHA3_224"),
+            (HashType::Sha3_256, "SHA3_256"),
+            (HashType::Sha3_384, "SHA3_384"),
+            (HashType::Sha3_512, "SHA3_512"),
+        ] {
+            let hash = item.0;
+            let name = item.1;
+            let hash_str = py_hash_from_hash_type(hash).unwrap();
+            assert!(hash_str == name);
```
alex, line 520:
```suggestion
            assert_eq!(hash_str, name);
```

alex (dismissed):

alex (approved):

PR #8077, opened 2023-01-15
Add constants for allowed PKCS12 and OpenSSH key types

### LINKED ISSUE #8039: Calling pkcs12.serialize_key_and_certificates() in a type-safe way
This is more of a best practice question, but I think it should be covered in documentation (I can submit a doc patch later).

When I get a key from `load_pem_private_key()`, it can be lots of types such as RSAPrivateKey or X25519PrivateKey.

But the `pkcs12.serialize_key_and_certificates(key=)` argument uses the internal type alias `_ALLOWED_PKCS12_TYPES` and accepts only RSAPrivateKey, DSAPrivateKey, EllipticCurvePrivateKey, Ed25519PrivateKey, Ed448PrivateKey.

What's the "correct" way to check that the key has the right type? Naively I would write:

```python
private_key = load_pem_private_key(...)
try:
    serialize_key_and_certificates(..., key=private_key, ...)
except TypeError:
    # Report to user
```

However this causes an error with type checkers like mypy:

> error: Argument 2 to "serialize_key_and_certificates" has incompatible type "Union[DHPrivateKey, Ed25519PrivateKey, Ed448PrivateKey, RSAPrivateKey, DSAPrivateKey, EllipticCurvePrivateKey, X25519PrivateKey, X448PrivateKey]"; expected "Union[RSAPrivateKey, DSAPrivateKey, EllipticCurvePrivateKey, Ed25519PrivateKey, Ed448PrivateKey, None]"  [arg-type]

I can add an additional `if` to the code to solve this:

```python
private_key = load_pem_private_key(...)
if not isinstance(
    private_key, (RSAPrivateKey, DSAPrivateKey, EllipticCurvePrivateKey, Ed25519PrivateKey, Ed448PrivateKey)
):
    # Report to user
else:
    serialize_key_and_certificates(..., key=private_key, ...)
```

But this seems cumbersome to hard-code and duplicate the list in every user program. Cryptography does not provide a constant containing accepted PKCS 12 key types, the type alias `_ALLOWED_PKCS12_TYPES` is private.

Maybe a public `tuple` like `ALLOWED_PKCS_KEY_TYPES = (RSAPrivateKey, DSAPrivateKey, ...)` should be added, so users can do `if isinstance(private_key, ALLOWED_PKCS_KEY_TYPES)`

intgr:
Added constants that can be used to check whether a key type is valid for PKCS12 or OpenSSH serialization, as discussed in #8039.

Closes #8039.

PS: I think it was a mistake to use UPPER_SNAKE_CASE for the type aliases in cryptography.hazmat.primitives.asymmetric.types. In other projects, type aliases such as Unions still typically use CamelCase like class names. But I guess it's too late to change that.

That's why I appended the `_TUPLE` to these constant names, so a constant `CERTIFICATE_PRIVATE_KEY_TYPES_TUPLE` could be created that doesn't conflict with the already existing `CERTIFICATE_PRIVATE_KEY_TYPES` type alias.

### DIFF
--- a/src/cryptography/hazmat/primitives/serialization/pkcs12.py
+++ b/src/cryptography/hazmat/primitives/serialization/pkcs12.py
@@ -20,6 +20,7 @@ __all__ = [
     "PBES",
     "PKCS12Certificate",
     "PKCS12KeyAndCertificates",
+    "PKCS12_KEY_TYPES_TUPLE",
     "load_key_and_certificates",
     "load_pkcs12",
     "serialize_key_and_certificates",
@@ -32,6 +33,16 @@ _ALLOWED_PKCS12_TYPES = typing.Union[
     ed25519.Ed25519PrivateKey,
     ed448.Ed448PrivateKey,
 ]
+
+PKCS12_KEY_TYPES_TUPLE: typing.Tuple[
+    typing.Type[_ALLOWED_PKCS12_TYPES], ...
+] = (
+    rsa.RSAPrivateKey,
+    dsa.DSAPrivateKey,
+    ec.EllipticCurvePrivateKey,
+    ed25519.Ed25519PrivateKey,
+    ed448.Ed448PrivateKey,
+)
 
 
 class PKCS12Certificate:
@@ -80,16 +91,7 @@ class PKCS12KeyAndCertificates:
         cert: typing.Optional[PKCS12Certificate],
         additional_certs: typing.List[PKCS12Certificate],
     ):
-        if key is not None and not isinstance(
-            key,
-            (
-                rsa.RSAPrivateKey,
-                dsa.DSAPrivateKey,
-                ec.EllipticCurvePrivateKey,
-                ed25519.Ed25519PrivateKey,
-                ed448.Ed448PrivateKey,
-            ),
-        ):
+        if key is not None and not isinstance(key, PKCS12_KEY_TYPES_TUPLE):
             raise TypeError(
                 "Key must be RSA, DSA, EllipticCurve, ED25519, or ED448"
                 " private key, or None."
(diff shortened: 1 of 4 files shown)


alex:
The things in `types.py` aren't a public API (they're not documented anywhere), so if we wanted to change them we could.

src/cryptography/hazmat/primitives/serialization/pkcs12.py
```diff
@@ -33,6 +34,16 @@
     ed448.Ed448PrivateKey,
 ]
 
+PKCS12_KEY_TYPES_TUPLE: typing.Tuple[
```
intgr, line 37: Should this be re-exported in `serialization/__init__.py` as well?

intgr:
> The things in types.py aren't a public API (they're not documented anywhere), so if we wanted to change them we could.

Up to you, if you think it's worth it. I already happened to use these for type hinting in my project, but it's no big deal to change it. :)

Then I could drop the `_TUPLE` suffix from the constant names.

reaperhulk:
So we're just blocked on naming? If the convention for type unions in public interfaces isn't upper case constants then let's just change it. So then we'll have `SSHPublicKeyTypes`, etc.

Since we don't have any public APIs using `serialization.ssh` we should export those values in `__init__`, but for `pkcs12` (and `pkcs7`) we use the module in the docs so we shouldn't re-export.

intgr:
Thanks, makes sense. I'll try to make the suggested changes later.

reaperhulk:
Is there anything left in this PR that needs rebasing to be reviewed?

intgr:
I think adding the constants still makes sense as well.

Although since we kept the old type names (like `CERTIFICATE_PRIVATE_KEY_TYPES` etc) around for deprecation, we need to come up with new names, or wait until they have been removed.

reaperhulk:
It seems like some new names is probably the path forward 😄

alex:
There's also some merge conflicts here now.

reaperhulk:
@alex and I talked a bit about this and we're not sure that there's actually any need for the tuples themselves at this point. If you're on a new enough Python (didn't test 3.9, but 3.10+ for sure) you can do isinstance checks directly with `Union` types, and if you're not you can use `typeguard` on it. Given those two facts, it doesn't seem necessary to expose a public constant for this.

intgr:
> on a new enough Python (didn't test 3.9, but 3.10+ for sure) you can do isinstance checks directly with Union types

Oh! I didn't know that. In that case, yes, let's close this. Thanks :)

PR #8115, opened 2023-01-21
try running downstream tests with 3.11

alex:


alex:
@mattsb42-aws 👋, not sure if you're still the right person for this, but it looks like the test requirements.txt file in aws-encryption-sdk-python (https://github.com/aws/aws-encryption-sdk-python/blob/master/test/upstream-requirements-py37.txt) was originally built for py37, and hasn't been touched in a while. Are y'all able to either a) refresh that file, b) produce one that's intended for py311?

mattsb42-aws:
I think @texastony is the one running point on those projects now. I'll poke the team.

alex:
Thanks

texastony:
@alex Acknowledged. 
I will either pick this up or pass this to the on-call.
If you have time, you could you cut us an issue?
If not, I should be able to keep this in my mind and get it done.

Is there a pressing timeline?

alex:
I see you got to it already. Thanks!

reaperhulk (approved):

PR #8116, opened 2023-01-22
Migrate from setup.cfg to pyproject.toml completely

alex:
One less file?


alex:
minimum setuptools version this requires is py37 only, so alas, this is on hold

alex:
with https://github.com/pyca/infra/pull/449 this will be green (though not mergable until 41)

.github/workflows/ci.yml
```diff
@@ -31,6 +31,7 @@ jobs:
           - {VERSION: "3.11", TOXENV: "docs", OPENSSL: {TYPE: "openssl", VERSION: "3.0.7"}}
           - {VERSION: "pypy-3.8", TOXENV: "pypy3-nocoverage"}
           - {VERSION: "pypy-3.9", TOXENV: "pypy3-nocoverage"}
+          - {VERSION: "3.11", TOXENV: "py311", OPENSSL: {TYPE: "openssl", VERSION: "1.1.1c"}}
```
reaperhulk, line 34: Should we be adding this job now so we can drop focal without coverage loss later? 😄

pyproject.toml
```diff
@@ -1,15 +1,97 @@
...
+[tool.setuptools.packages.find]
+where = ["src"]
+exclude = [
+    "_cffi_src",
+    "_cffi_src.*",
+]
+
+[tool.setuptools.dynamic]
+version = {attr = "cryptography.__version__"}
+readme = {file = "README.rst", content-type = "text/x-rst"}
```
Viicos, line 70: iirc, `readme` doesn't have to be dynamic if an extension (`.rst` in this case) is provided. See https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#readme for reference

reaperhulk (approved):

PR #8122, opened 2023-01-23
use uuidgen instead of a python invocation

reaperhulk:
Save that 60-70ms. Assuming uuidgen is present everywhere.


alex (approved):

reaperhulk:
Not present on Windows or some of our linux distros and not worth effort 😄

PR #8160, opened 2023-01-29
allow CRL validation with Ed448 and Ed25519 keys (fixes #8156)

### LINKED ISSUE #8156: CertificateRevocationList.is_signature_valid does not allow ed448/ed25519 keys
Hi,

In #6276, CRLs where migrated to Rust. The following lines where added (I'm not sure if there was an equivalent to it before, but I couldn't find it) to `CertificateRevocationList.is_signature_valid()` (or rather, it's OpenSSL Backend implementation in `src/cryptography/hazmat/backends/openssl/backend.py`):

```python
    def _crl_is_signature_valid(
        self, crl: x509.CertificateRevocationList, public_key: PUBLIC_KEY_TYPES
    ) -> bool:
        if not isinstance(
            public_key,
            (
                _DSAPublicKey,
                _RSAPublicKey,
                _EllipticCurvePublicKey,
            ),
        ):
            raise TypeError(
                "Expecting one of DSAPublicKey, RSAPublicKey,"
                " or EllipticCurvePublicKey."
            )
```

Is there any reason for this restriction? I stumbled over this while trying out CRLs signed with Ed448 and Ed25519 keys. As an experiment I removed the lines and the validation did still work fine.

If possible, this should be extended properly to include `_Ed448PublicKey` and `_Ed25519PublicKey`, otherwise it should be documented in the method signature, which so far makes no mention of it.

I'm fine to open an MR either way as well, as long as you tell me what to do :-)

kr, Mat

mathiasertl:


alex:
Test + changelog please. Thank you!

mathiasertl:
I would add that `_csr_is_signature_valid` just below the updated function has no such check at all. Should we remove that check after all? Or add it to the csr pendant as well?

alex:
CSRs are slightly different, in that the key comes from the CSR, whereas the CRL comes externally, that's the likely reason for the difference.

alex:
Ok, I believe https://github.com/pyca/cryptography/pull/8162 has supserceded this (sorry about that!). Can you confirm that `main` works for you now?

mathiasertl:
@alex I can confirm itś working in the current main branch - thanks!

mathiasertl:
@alex 

> Ok, I believe https://github.com/pyca/cryptography/pull/8162 has supserceded this (sorry about that!).

No hard feelings of course, thanks for your work!

PR #8196, opened 2023-02-02
try enabling ed25519/448 with fips

alex:
fips 186-5 is final!


alex:
doesn't work on RHEL. ping'd them, but I assume this awaits a FIPS validation

PR #8227, opened 2023-02-07
always specify a minimum bound on tox

alex:
this prevents catastrophic backtracking and favors pip resolution errors over insanely old packages leading to runtime errors

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
           repository: "google/wycheproof"
           path: "wycheproof"
           ref: "master"
-      - run: python -m pip install -c ci-constraints-requirements.txt tox requests coverage[toml]
+      - run: python -m pip install -c ci-constraints-requirements.txt 'tox>3' requests coverage[toml]
       - name: Compute config hash and set config vars
         run: |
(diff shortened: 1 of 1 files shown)


reaperhulk:
Could we not put this as another line in the constraint file?

alex:
Hmm, how so?

alex:
Are you thinking something like `tox==blah; python_version='3.6'`? I'm concerned this will break dependabot.

reaperhulk:
Yeah, although I could imagine that breaking dependabot. One way to find out 😂

alex:
is there an advantage to doing it that way? we already hvae the `tox>3` for debuggability in at least one other place

alex:
I promise it won't proliferate further than this :-)

On Tue, Feb 7, 2023 at 8:55 AM Paul Kehrer ***@***.***> wrote:
>
> I mostly just don’t want it proliferating, but we get to remove this when we drop py36 anyway. Too bad minversion in tox.ini is useless (although it is also set to 2.4 right now I think…)
>
> —
> You are receiving this because you authored the thread.Message ID: ***@***.***>



-- 

reaperhulk (approved):

PR #8230, opened 2023-02-07
Don't allow update_into to mutate immutable objects

alex:


reaperhulk (approved):

alittlesir:
Does this involve interface changes?

reaperhulk:
It behaves exactly the same unless you were passing an immutable type (e.g. a byte string). So the interface has technically changed but in practice it has only changed for the vulnerable path.

alittlesir:
In this case, change the buf class type from `bytes` to `bytearray`. it is OK?

it can't use the `bytes` type anymore.

```python
def update_into(self, data: bytes, buf: bytes) -> int:
```

alittlesir:
I'm not sure what the effect of this vulnerability is.

reaperhulk:
Unfortunately the Python type hint system doesn’t have a way to represent “all objects that implement the buffer protocol but only the mutable types”, so `bytes` is still the closest to “correct” type that we can declare.

alittlesir:
What is the impact of this cve? I don't seem to feel the threat of this vulnerability.

PR #8257, opened 2023-02-10
Dump pkcs12

Kuro66:
Added methods to the PKCS12KeyandCertificates and to the PKCS12Certificate classes to allow easier dump of certificate and key by converting them into bytes.

### DIFF
--- a/src/cryptography/hazmat/primitives/serialization/pkcs12.py
+++ b/src/cryptography/hazmat/primitives/serialization/pkcs12.py
@@ -71,6 +71,9 @@ class PKCS12Certificate:
         return "<PKCS12Certificate({}, friendly_name={!r})>".format(
             self.certificate, self.friendly_name
         )
+    
+    def dump_certificate(self, encoding) -> typing.Optional[bytes]:
+        return self.certificate.public_bytes(encoding)
 
 
 class PKCS12KeyAndCertificates:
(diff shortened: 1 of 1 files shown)


alex:
This is not the approach we take to API design in cryptography, we prefer composability.

Because there is nothing in these APIs that is PKCS12 specific, its more appropriate for the methods to be on the key and certificate classes, as they already are.

Kuro66:
Hello Alex,
I just thought it would make it easier. I came here after using pyOpenSSL and receiving a deprecation warning. pyOpenSSL had those sorts of features and I must admit I was a bit lost when I arrived here and had to untangle how it all worked in order to find out what to call on what. So I thought it could help people in the same situation as I was in.

alex:
pyOpenSSL is deprecated for exactly these reasons :-) The lack of
composability/orthogonality in its design makes it painful to
maintain.

On Fri, Feb 10, 2023 at 8:06 AM Kuro ***@***.***> wrote:
>
> Hello Alex,
> I just thought it would make it easier. I came here after using pyOpenSSL and receiving a deprecation warning. pyOpenSSL had those sorts of features and I must admit I was a bit lost when I arrived here and had to untangle how it all worked in order to find out what to call on what. So I thought it could help people in the same situation as I was in.
>
> —



-- 

Kuro66:
Okay, I didn't know that.
Not reproducing the same mistake would indeed be a good idea. :)

alex:
Thanks for your interest in contributing.

PR #8268, opened 2023-02-12
don't cache cargo/bin, we don't need it

reaperhulk:
cargo/bin is 240MB of stuff that comes with the image -- we shouldn't need to cache it...right?


alex:
We do need it for the coverage jobs.

alex (dismissed):

alex (approved):

PR #8272, opened 2023-02-12
Disallow wildcards in DNSName for name constraints

gbansaghi:
As discussed in #8253, wildcards are unnecessary according to RFC 5280, and cause issues with at least Firefox.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,6 +31,10 @@ Changelog
   :class:`~cryptography.hazmat.primitives.serialization.SSHCertificateBuilder`.
 * Added :meth:`~cryptography.x509.Certificate.verify_directly_issued_by` to
   :class:`~cryptography.x509.Certificate`.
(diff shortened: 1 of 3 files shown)


reaperhulk:
Could you add a changelog entry noting this change? Thanks for the contribution!

gbansaghi:
@reaperhulk I added the note, is there any way to verify that I got the links right before I push?

alex:
You can build docs locally with tox -e docs

reaperhulk:
You can also just push it and see if it builds appropriately in CI (we have a docs job where you can look at the built output).

reaperhulk (approved):

gbansaghi:
The local docs build failed because of one misspelled word, but it didn't say where and persisted even after I rolled back my changes, so probably unrelated.

@reaperhulk I got a notification from github the second I hit push and was briefly but intensely confused.

reaperhulk:
Thanks for the contribution @gbansaghi!

PR #8273, opened 2023-02-12
fix some binding declarations

reaperhulk:
These align more closely with OpenSSL 3.

also adds a comment about why we deliberately incorrectly declare some of them, sigh

### DIFF
--- a/tests/hazmat/bindings/test_openssl.py
+++ b/tests/hazmat/bindings/test_openssl.py
@@ -58,23 +58,6 @@ class TestOpenSSL:
         assert resp == expected_options
         assert b.lib.SSL_get_options(ssl) == expected_options
 
-    def test_ssl_mode(self):
-        # Test that we're properly handling 32-bit unsigned on all platforms.
-        b = Binding()
-        # SSL_OP_ALL is 0 on BoringSSL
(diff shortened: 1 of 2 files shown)


reaperhulk:
Arguably we can do more here. For example, `SSL_CTX_set_mode` and `SSL_set_mode` are only valid for `SSL_MODE_*` constants and there is no `SSL_OP_ALL` type top bit issue.

reaperhulk:
And, given that being true, we can just declare them all as long since the ones where `SSL_OP_ALL` is valid have been switched to uint64_t declarations in OpenSSL itself.

We'll also need to update our tests since we incorrectly pass SSL_OP_ALL to a function where it has no meaning in `test_ssl_mode`

tests/hazmat/bindings/test_openssl.py
```diff
@@ -58,23 +58,6 @@ def test_ssl_options(self):
         assert resp == expected_options
         assert b.lib.SSL_get_options(ssl) == expected_options
 
-    def test_ssl_mode(self):
```
alex, line 61: Why was this removed?

alex (approved):

PR #8287, opened 2023-02-14
make our rust/pip cache a composite action

reaperhulk:
This re-caches `index` on nightly, which is not required, but as soon as 1.68 is released we can remove that cache line and enable sparse registry everywhere.

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -1,0 +1,38 @@
+name: Cache rust and pip
+description: Caches rust and pip data to speed builds
+inputs:
+  additional-paths:
+    description: 'Additional paths to add to the cache'
+    required: false
+    default: ''
+  key:
+    description: 'cache key'
+    required: true
+
+
+runs:
+  using: "composite"
+
+  steps:
+    - name: Get pip cache dir
+      id: pip-cache
+      run: |
+        # Determine the path to our Python. It's in venv for our containers
+        # but just standard $PATH for setup-python pythons.
+        if [[ -f "/venv/bin/python" ]]; then
+          echo "dir=$(venv/bin/python -m pip cache dir)" >> $GITHUB_OUTPUT
+        elif which python >/dev/null; then
+          echo "dir=$(python -m pip cache dir)" >> $GITHUB_OUTPUT
+        fi
+      shell: bash
+    - uses: actions/cache@v3.2.5
+      with:
+        path: |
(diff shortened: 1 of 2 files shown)


.github/actions/cache/action.yml
```diff
@@ -0,0 +1,38 @@
...
+      with:
+        path: |
+          ${{ steps.pip-cache.outputs.dir }}
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/registry/src/
+          ~/.cargo/git/db/
+          src/rust/target/
+          ${{ inputs.additional-paths }}
+        key: ${{ inputs.key }}
```
alex, line 38: Could we be auto-computing the key?

.github/actions/cache/action.yml
```diff
@@ -0,0 +1,38 @@
...
+  using: "composite"
+
+  steps:
+    - name: Get pip cache dir
+      id: pip-cache
+      run: |
+        # Determine the path to our Python. It's in venv for our containers
+        # but just standard $PATH for setup-python pythons.
+        if [[ -f "/venv/bin/python" ]]; then
+          echo "dir=$(venv/bin/python -m pip cache dir)" >> $GITHUB_OUTPUT
```
alex, line 23:
```suggestion
          echo "dir=$(/venv/bin/python -m pip cache dir)" >> $GITHUB_OUTPUT
```

.github/workflows/ci.yml
```diff
@@ -297,19 +276,11 @@ jobs:
...
-      - uses: actions/cache@v3.2.5
-        id: cargo-cache
-        timeout-minutes: 5
+      - name: Cache rust and pip
+        uses: ./.github/actions/cache
+        timeout-minutes: 2
         with:
-          path: |
-            ~/.cache/pip/wheels
-            ~/.cargo/bin/
```
alex, line 306: This seems like it got lost?

.github/actions/cache/action.yml
reaperhulk: ...maybe? We'd have to pass in some extra args and I'm not convinced it will be less complex. If we want to explore that I'd prefer to do it in a future PR (but I probably won't tackle that at the moment)

alex: I think it'd be good, but happy to move to a different PR.

alex (approved):

PR #8289, opened 2023-02-14
add a cache for the benchmark job

reaperhulk:


.github/workflows/benchmark.yml
```diff
@@ -31,15 +31,24 @@ jobs:
...
           python -m venv .venv-main
           .venv-main/bin/pip install -v -c ./cryptography-main/ci-constraints-requirements.txt "./cryptography-main[test]" ./cryptography-main/vectors/
+        env:
+          CARGO_TARGET_DIR: ${{ format('{0}/src/rust/target/', github.workspace) }}
       - name: Create virtualenv (PR)
         run: |
           python -m venv .venv-pr
           .venv-pr/bin/pip install -v -c ./cryptography-pr/ci-constraints-requirements.txt "./cryptography-pr[test]" ./cryptography-pr/vectors/
+        env:
+          CARGO_TARGET_DIR: ${{ format('{0}/src/rust/target/', github.workspace) }}
```
alex, line 51: So this will share the target/ dir between the main and PR, do we want that?

I intentionally did not include any caching in this job because I wanted some assurance of no bleed between jobs.

reaperhulk: Is there any scenario where that would be unsafe? Any changed rust files will have differing intermediate artifact names right?

That said, we could just drop it from the PR and have no caching for that part.

alex: It _shouldn't_ have an impact, but I think I have a soft-preference for not including it out of an abundance of cautioun -- and also since this is far from the slowest job.

reaperhulk:
composite actions make this annoying and the benefit isn't enough for me to spend any more effort 😄

PR #8301, opened 2023-02-16
update to latest pytest-subtests and add the new flag

reaperhulk:


ci-constraints-requirements.txt
```diff
@@ -141,7 +141,7 @@ pytest-randomly==3.12.0
     # via cryptography (setup.cfg)
 pytest-shard==0.1.2
     # via cryptography (setup.cfg)
-pytest-subtests==0.9.0; python_version >= "3.7"
+pytest-subtests==0.10.0; python_version >= "3.7"
```
alex, line 144: update setup.cfg as well, if it's in the default opts we need it as a minimum

alex (approved):

PR #8302, opened 2023-02-16
are 2019 runners faster?

reaperhulk:


reaperhulk:
The variance is so colossal here that I have no idea. We should rerun this during the day with more noisy neighbors, but I doubt we should merge this. Marking as draft for now.

alex:
I don't think there's an action item here, particularly as we've moved to quantifying the actual causes of timouts in #8315

PR #8323, opened 2023-02-19
remove unused ssl bindings

alex:


reaperhulk (approved):
Originally added in https://github.com/pyca/cryptography/pull/1019, but unused by pyOpenSSL and no identified consumers.

PR #8324, opened 2023-02-19
remove unused x509 bindings

alex:


reaperhulk (approved):
Insert added in https://github.com/pyca/cryptography/pull/2569 but no longer used by us. delete added as part of the original en masse binding (https://github.com/pyca/cryptography/pull/126). Neither have users that we're aware of (and no one should use the x509 bindings if they can avoid it. cryptography no longer uses them, this is strictly pyOpenSSL compatibility)

PR #8325, opened 2023-02-19
remove unused x509name bindings

alex:


reaperhulk (approved):
Added in https://github.com/pyca/cryptography/pull/2795 originally

PR #8326, opened 2023-02-19
remove verification error code bindings

alex:


reaperhulk (approved):
Added in a mass binding (https://github.com/pyca/cryptography/pull/1151)

PR #8329, opened 2023-02-19
remove unused ec bindings

alex:


reaperhulk (approved):
We don't know who our consumers are here and we should require a higher level interface if we're going to support arbitrary EC operations, so I'm okay with removing this and seeing if we get any issues.

PR #8330, opened 2023-02-19
remove unused netscape spki binding

alex:


reaperhulk (approved):
Added in https://github.com/pyca/cryptography/pull/1511

PR #8331, opened 2023-02-19
remove unused evp binding

alex:


reaperhulk (approved):
Originally added in https://github.com/pyca/cryptography/pull/3684

PR #8333, opened 2023-02-19
remove unused CRL bindings

alex:


reaperhulk (approved):
Added in #4127, https://github.com/pyca/cryptography/pull/2049, and #1917.

PR #8337, opened 2023-02-19
remove unused DTLS bindings

reaperhulk:
originally added in https://github.com/pyca/cryptography/pull/4089


src/_cffi_src/openssl/ssl.py
```diff
@@ -450,8 +450,6 @@
 /* DTLS support */
 long Cryptography_DTLSv1_get_timeout(SSL *, time_t *, long *);
 long DTLSv1_handle_timeout(SSL *);
-long DTLS_set_link_mtu(SSL *, long);
-long DTLS_get_link_min_mtu(SSL *);
```
alex, line 454: there's conditional crap for these at l570

alex (approved):

PR #8343, opened 2023-02-20
remove unused x509 bindings

alex:


reaperhulk (approved):
Originally added in https://github.com/pyca/cryptography/pull/1955, https://github.com/pyca/cryptography/pull/5315, and https://github.com/pyca/cryptography/pull/1820

PR #8345, opened 2023-02-20
Remove unused rsa binding

alex:


reaperhulk (approved):

alex:
We've been on a gradual effort to remove bindings that are unused, because they add a burden to:

- compatibility (if the API changes or is removed in any of boringssl, libress, or openssl, we need to take action)
- compilation time
- supported surface

PR #8376, opened 2023-02-22
Enforce max number of SSH certificate principals

jake-odoa:
There is an undocumented limit for the maximum number of valid principals accepted by the openssh tooling, as seen at:
* https://github.com/openssh/openssh-portable/blob/27267642699342412964aa785b98afd69d952c88/sshkey.h#L108
* https://github.com/openssh/openssh-portable/blob/25c8a2bbcc10c493d27faea57c42a6bf13fa51f2/sshkey.c#L1801
* https://github.com/openssh/openssh-portable/blob/6180b0fa4f7996687678702806257e661fd5931e/ssh-keygen.c#L1833

This change enforces that same restriction as currently a SSH certificate can be generated that is invalid against the default sshd server. Consideration might be given for any non openssh servers that accept openssh certificates, if they exist and want to allow a greater number of principals.

Of note, the 256 limit is not found in the spec for SSH certificates as defined at https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.certkeys. It instead seems to be arbitrarily chosen by the project as some limit was needed.

### DIFF
--- a/src/cryptography/hazmat/primitives/serialization/ssh.py
+++ b/src/cryptography/hazmat/primitives/serialization/ssh.py
@@ -1052,6 +1052,9 @@ _SSH_CERT_PRIVATE_KEY_TYPES = typing.Union[
 ]
 
 
+_SSHKEY_CERT_MAX_PRINCIPALS = 256
+
+
 class SSHCertificateBuilder:
     def __init__(
         self,
@@ -1181,6 +1184,12 @@ class SSHCertificateBuilder:
             )
         if self._valid_principals:
             raise ValueError("valid_principals already set")
+
+        if len(valid_principals) >= _SSHKEY_CERT_MAX_PRINCIPALS:
+            raise ValueError(
+                    "Reached or exceeded the maximum number of "
+                    "valid_principals"
+            )
 
         return SSHCertificateBuilder(
             _public_key=self._public_key,


jake-odoa:
Failed CI / flake stage ~seems~ to be for code unrelated to this PR

The output of `black ./src/cryptography/hazmat/primitives/serialization/ssh.py` results in a diff of:

```
diff --git a/src/cryptography/hazmat/primitives/serialization/ssh.py b/src/cryptography/hazmat/primitives/serialization/ssh.py
index cc2b38d26..974ee80d0 100644
--- a/src/cryptography/hazmat/primitives/serialization/ssh.py
+++ b/src/cryptography/hazmat/primitives/serialization/ssh.py
@@ -569,9 +569,7 @@ _SSH_PRIVATE_KEY_TYPES = typing.Union[
 
 
 def load_ssh_private_key(
-    data: bytes,
-    password: typing.Optional[bytes],
-    backend: typing.Any = None,
+    data: bytes, password: typing.Optional[bytes], backend: typing.Any = None
 ) -> _SSH_PRIVATE_KEY_TYPES:
     """Load private key from OpenSSH custom encoding."""
     utils._check_byteslike("data", data)
@@ -738,9 +736,7 @@ _SSH_PUBLIC_KEY_TYPES = typing.Union[
 ]
 
 _SSH_CERT_PUBLIC_KEY_TYPES = typing.Union[
-    ec.EllipticCurvePublicKey,
-    rsa.RSAPublicKey,
-    ed25519.Ed25519PublicKey,
+    ec.EllipticCurvePublicKey, rsa.RSAPublicKey, ed25519.Ed25519PublicKey
 ]
 
 
@@ -889,8 +885,7 @@ def _get_ec_hash_alg(curve: ec.EllipticCurve) -> hashes.HashAlgorithm:
 
 
 def _load_ssh_public_identity(
-    data: bytes,
-    _legacy_dsa_allowed=False,
+    data: bytes, _legacy_dsa_allowed=False
 ) -> typing.Union[SSHCertificate, _SSH_PUBLIC_KEY_TYPES]:
     utils._check_byteslike("data", data)
 
@@ -1046,9 +1041,7 @@ def serialize_ssh_public_key(public_key: _SSH_PUBLIC_KEY_TYPES) -> bytes:
 
 
 _SSH_CERT_PRIVATE_KEY_TYPES = typing.Union[
-    ec.EllipticCurvePrivateKey,
-    rsa.RSAPrivateKey,
-    ed25519.Ed25519PrivateKey,
+    ec.EllipticCurvePrivateKey, rsa.RSAPrivateKey, ed25519.Ed25519PrivateKey
 ]

@@ -1111,7 +1104,7 @@ class SSHCertificateBuilder:
     def serial(self, serial: int) -> "SSHCertificateBuilder":
         if not isinstance(serial, int):
             raise TypeError("serial must be an integer")
-        if not 0 <= serial < 2**64:
+        if not 0 <= serial < 2 ** 64:
             raise ValueError("serial must be between 0 and 2**64")
         if self._serial is not None:
             raise ValueError("serial already set")
@@ -1187,8 +1180,7 @@ class SSHCertificateBuilder:
 
         if len(valid_principals) >= _SSHKEY_CERT_MAX_PRINCIPALS:
             raise ValueError(
-                    "Reached or exceeded the maximum number of "
-                    "valid_principals"
+                "Reached or exceeded the maximum number of " "valid_principals"
             )
 
         return SSHCertificateBuilder(
@@ -1232,7 +1224,7 @@ class SSHCertificateBuilder:
         if not isinstance(valid_before, (int, float)):
             raise TypeError("valid_before must be an int or float")
         valid_before = int(valid_before)
-        if valid_before < 0 or valid_before >= 2**64:
+        if valid_before < 0 or valid_before >= 2 ** 64:
             raise ValueError("valid_before must [0, 2**64)")
         if self._valid_before is not None:
             raise ValueError("valid_before already set")
@@ -1256,7 +1248,7 @@ class SSHCertificateBuilder:
         if not isinstance(valid_after, (int, float)):
             raise TypeError("valid_after must be an int or float")
         valid_after = int(valid_after)
-        if valid_after < 0 or valid_after >= 2**64:
+        if valid_after < 0 or valid_after >= 2 ** 64:
             raise ValueError("valid_after must [0, 2**64)")
         if self._valid_after is not None:
             raise ValueError("valid_after already set")

```

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -1052,6 +1052,9 @@ def serialize_ssh_public_key(public_key: _SSH_PUBLIC_KEY_TYPES) -> bytes:
 ]
 
 
+_SSHKEY_CERT_MAX_PRINCIPALS = 256
```
alex, line 1055: Can you please add a comment indicating that while this isn't part of the spec, it is enforced by openssh?

alex:
Please also add a test for this. Thanks!

reaperhulk:
Good catch!

src/cryptography/hazmat/primitives/serialization/ssh.py
jake-odoa: Added in an explainer comment.

jake-odoa:
> Please also add a test for this. Thanks!

Have done so, please let me know if I've missed anything :)

jake-odoa:
> Good catch!

Thanks! Seeing SSH certificate support in the pipeline was an awesome surprise to stumble upon.

alex (approved):

PR #8391, opened 2023-02-26
remove several unused SSL bindings

alex:


reaperhulk (approved):
Added in #857, #266, and #1915

PR #8399, opened 2023-02-27
remove extras we don't need in flake

reaperhulk:


reaperhulk:
Sigh, untrue due to mypy

PR #8401, opened 2023-02-27
remove unused bignum bindings

alex:


reaperhulk (approved):
For future people: If you think you need this then please file an issue and propose an API for how we could expose hazmat functionality for this rather than trying to reach into the bindings (which are not public surface area).

PR #8416, opened 2023-03-01
use pytest-xdist in dynamo downstream

reaperhulk:


alex:
is this even expected to work?

reaperhulk:
They install xdist in their deps, but it doesn't work and they don't use it in their own CI. Not worth pursuing.

PR #8440, opened 2023-03-04
Add an ignore for the newest ruff

alex:
Clearly document why we have each of our ignores


alex:
not positive if that `pyi` change works on py36, and don't have a py36 to test with at this point

alex:
`mypy src/cryptography/ vectors/cryptography_vectors/ tests/ --python-version=3.6` but I don't know if that's truly meaningful

reaperhulk (approved):

PR #8446, opened 2023-03-05
remove iso8601 test dep

reaperhulk:


reaperhulk:
`%z` appears to have been extended to handle the form `±HH:MM` in Python 3.7+. Prior to that it only supported`±HHMM`. Of course, if we're going to wait to drop 3.6 support then we can use `datetime.datetime.fromisoformat` and avoid all of this `strptime` nonsense. I'm leaving this open and tagging to 41, but will rebase to `fromisoformat` when the time comes.

alex (approved):

PR #8458, opened 2023-03-06
double the speed of our rsa tests

reaperhulk:
this both creates a reusable fixture for our most commonly used private keys as well as disables key validation. as always, disabling key validation should not be done unless you never parse untrusted key input. unsurprisingly, our tests are trusted and understood input (and we also continue to have tests where we run check key to verify that it catches corrupt things)

### DIFF
--- a/tests/hazmat/primitives/test_rsa.py
+++ b/tests/hazmat/primitives/test_rsa.py
@@ -60,12 +60,12 @@ from .utils import (
 
 
 @pytest.fixture(scope="session")
-def rsa_key_512():
+def rsa_key_512() -> rsa.RSAPrivateKey:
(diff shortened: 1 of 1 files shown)


tests/hazmat/primitives/test_rsa.py
```diff
@@ -445,7 +458,10 @@ def test_oaep_wrong_label(self, enclabel, declabel, backend):
         skip_message="Does not support PKCS1v1.5.",
     )
     def test_lazy_blinding(self, backend):
-        private_key = RSA_KEY_2048.private_key(backend)
+        # We don't want to reuse the rsa_key_2048 fixture here
```
alex, line 461: Add the _why_ we don't to this comment

alex (approved):

PR #8463, opened 2023-03-06
EC key validation experiment

reaperhulk:
Previously we loaded EC keys in several different ways. This PR semi-unifies that and then enables using unsafe_skip_key_validation for EC as well.

To make this work we  alter _ec_key_set_public_key_affine_coordinates to implement the equivalent of EC_KEY_set_public_key_affine_coordinates without EC_KEY_check_key (which is now called in the __init__ of 
EllipticCurvePrivateKey and also in the other two load paths). However, we don't implement it on `ECPublicKey.__init__` due to massive performance impact in X.509...something to talk about.

Using this unsafe flag can significantly speed up loading numerous keys. However, don't pass this flag unless you're very sure your keys aren't attacker controlled and are always valid.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,8 +26,19 @@ Changelog
   and
   :class:`~cryptography.hazmat.primitives.asymmetric.dsa.DSAPrivateKey`.
 * Deprecated and renamed the keyword argument for disabling RSA key validation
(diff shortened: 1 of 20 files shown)


reaperhulk:
Depends on #8470

reaperhulk:
#8475 replaced this

PR #8470, opened 2023-03-07
rename unsafe_skip_rsa_key_validation to unsafe_skip_key_validation

reaperhulk:
this is in prep for an EC PR that will allow disabling checks for EC keys in a similarly performance sensitive but also footgun-y manner


reaperhulk:
This isn't intended to be merged until after we've had a chance to talk 😄

reaperhulk:
No longer needed

PR #8487, opened 2023-03-10
cache ECDH values in wycheproof too

reaperhulk:
this alters and renames the caching function a bit since it caches **to the group** object but the actual values (in ECDH) come from the testcase itself

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -913,11 +913,11 @@ class WycheproofTest:
     def has_flag(self, flag: str) -> bool:
         return flag in self.testcase["flags"]
 
-    def cache_group_value(self, cache_key: str, func):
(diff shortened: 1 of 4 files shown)


alex:
Ok but this is fundamentally wrong. Reusing a value from the testcase across different testcsaes in a group means using the wrong value.

reaperhulk:
How is it wrong? If the same private scalar is being used for the same curve then the same key is derived in all cases. The test group just serves as a bucket to hold the object. For ECDSA the wycheproof tests hold that key at the "group" level, but for ECDH there are reused keys between testcases (but not every testcase has the same key). This simply caches based on the private scalar value.

alex (approved):

PR #8496, opened 2023-03-11
More fixes for PyPI OIDC publishing (still doesn't work)

reaperhulk:
At least a few issues still remain:

1) It's failing with a 403 ~~but not failing the build~~ (see: https://github.com/pyca/cryptography/actions/runs/4391875928/jobs/7691301517) so we need to both get uploading actually working ~~and have failed builds actually fail~~ (failing builds are fixed with xargs now) 😄 
2) We need to upload to both cryptography and cryptography_vectors so we may need @di to allowlist on one more package.
3) @alex if you want to add me as an owner on the cryptography package I can add you on cryptography_vectors (but you have to verify your email before it will let me)

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -66,4 +66,4 @@ jobs:
               f.write("TWINE_PASSWORD={pypi_token}\n")
         shell: python
 
-      - run: find dist/ -type f -name 'cryptography*' -exec twine upload --repository $TWINE_REPO {} \;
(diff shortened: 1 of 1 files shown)


alex (approved):

di:
> We need to upload to both cryptography and cryptography_vectors so we may need @di to allowlist on one more package.

The beta is per-user, not per project. I can add you to the beta but Alex should also be already able to set up an OIDC publisher for this project as well.

Issue #8450, opened 2023-03-05
Switch to cargo sparse registry


reaperhulk:
https://github.com/rust-lang/cargo/pull/11224 will be in rust 1.68.0 (releasing on March 9). This is a tracking issue for adopting it fully.

- [x] Rebuild container images from infra with latest rustup stable.
- [x] https://github.com/pyca/cryptography/pull/8480
- [x] Upgrade tart base image with latest rustc for arm64 macOS runners.
- [ ] Wait for GH to ship 1.68.0 in base images
  - [x] https://github.com/actions/runner-images/pull/7279
  - [ ] https://github.com/actions/runner-images/pull/7309
  - [x] https://github.com/actions/runner-images/pull/7286
- [ ] Add the same key everywhere else 😄
- [ ] Delete `~/.cargo/registry/index/` from `.github/actions/cache/action.yml` since it is no longer needed.

PR #8497, opened 2023-03-11
add sparse env var to all jobs even though they're not on rust 1.68.0

reaperhulk:
as soon as the weekly builder update rolls out for GHA this will start working

as caches naturally expire we'll get the optimization, although once we see 1.68.0 we can also deliberately delete our caches or rev the cache number.

refs #8450


reaperhulk:
Ah, now I remember why we can't do this 🤣

PR #8516, opened 2023-03-14
Python 3.6 compatibility fix.

akamaryan:
The capture_output argument to subprocess.run() was not introduced until Python 3.7.  Use stdout=subprocess.PIPE and stderr=subprocess.PIPE instead, which is equivalent.


alex:
lol, ruff is upset about this because it considers the "old" form deprecated.

you can get CI passing by adding an ignore to `UP022` in `pyproject.toml` (please include a comment like the others)

reaperhulk (approved):

alex (approved):

PR #8528, opened 2023-03-15
modify the python path in build_openssl.py instead of build.rs

reaperhulk:


src/rust/build.rs
```diff
@@ -23,21 +23,7 @@ fn main() {
...
     println!("cargo:rerun-if-changed=../_cffi_src/");
-    let python_path = match env::var("PYTHONPATH") {
-        Ok(mut val) => {
-            if cfg!(target_os = "windows") {
-                val.push(';');
-            } else {
-                val.push(':');
-            }
-            val.push_str("..");
-            val.push(MAIN_SEPARATOR);
```
alex, line 34: remove `MAIN_SEPARATOR` import

alex (approved):

PR #8531, opened 2023-03-16
enable sparse registry for ubuntu 22.04 runners

reaperhulk:
it looks like rollout is completed for the latest image update


reaperhulk:
Closing in favor of #8532 which does it in more places 😄

PR #8556, opened 2023-03-20
Added support for handling python buffers in Rust code

alex:
This is extra mega cursed, and strictly speaking unsound. It does, however, match the status quo ante, where someone mutating a buffer while its being used in cffi code will basically always be UB.


reaperhulk (approved):

reaperhulk:
:thanks-i-hate-it:

PR #8613, opened 2023-03-25
Switch from pytest-subtests to a mini-version that's faster

alex:


reaperhulk:
I think we should talk to the pytest folks a bit about whether there's a way we can do this without importing `Skipped` from the internals?

alex:
That seems reasonable

reaperhulk (approved):
I'm going to miss my subtest count. 🪦

PR #8635, opened 2023-03-29
define path for type for rust classes (no more builtins)

reaperhulk:


woodruffw:
Looks correct to me! Might be good to have a basic backstop test for these to make sure they remain importable at those module paths.

Also, if you want `from ... import ...` to work with these, you'll need a nasty hack like this:

https://github.com/woodruffw/pyrage/blob/e6e34cb2fbd55daccf686767c6938d5cbeaacef9/src/lib.rs#L191-L200

alex (approved):

PR #8637, opened 2023-03-29
docs: blowfish: add example code

milahu:
add example code

ideally i would remove the prompt string `>>>` as this is better for copy-paste:

```py
from cryptography.utils import CryptographyDeprecationWarning
import warnings
warnings.filterwarnings("ignore", category=CryptographyDeprecationWarning)
import os
key = b"secret"
algorithm = algorithms.Blowfish(key)
iv = os.urandom(8)
cipher = Cipher(algorithm, modes.CBC(iv))
encryptor = cipher.encryptor()
ct = encryptor.update(b"a secret message")
decryptor = cipher.decryptor()
decryptor.update(ct)
```

### DIFF
--- a/docs/hazmat/primitives/symmetric-encryption.rst
+++ b/docs/hazmat/primitives/symmetric-encryption.rst
@@ -254,6 +254,23 @@ Weak ciphers
         :term:`bits` in length in increments of 8 bits.
     :type key: :term:`bytes-like`
 
+    .. doctest::
+
+        >>> from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+        >>> from cryptography.utils import CryptographyDeprecationWarning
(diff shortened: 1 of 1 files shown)


reaperhulk:
Thanks for the contribution, but we don't really want to be providing examples that explicitly suppress the warnings we're generating in advance of removing support for an algorithm. Additionally, this example is unwise (we don't want to provide examples where `key` is anything other than cryptographically random bytes) and breaks later examples since all doctest is in the same namespace.

milahu:
> removing support for an algorithm

blowfish is required for legacy apps like [netgear-telenetenable](https://github.com/insanid/netgear-telenetenable)
using [a separate library](https://pypi.org/project/blowfish/) is kinda stupid

> breaks later examples

just remove `key`

PR #8644, opened 2023-03-30
Stop using Certbot's pip_install_editable

bmw:
I'm planning on removing this weirdly specific script from the Certbot repo soon. `tools/pip_install.py` already exists and will do the same thing as `pip_install_editable.py` after we add `-e` to the command line.


alex (approved):
Thanks!

PR #8646, opened 2023-03-31
Update cargo packages that dependabot couldn't handle

alex:


alex:
Weird versioning here is expected, see https://github.com/microsoft/windows-rs/issues/2410#issuecomment-1490431882

reaperhulk (approved):

PR #8651, opened 2023-04-01
Switch from tox to nox

alex:


docs/development/getting-started.rst
```diff
@@ -6,22 +6,22 @@ Development dependencies
...
     $ # Specify your Python version here.
-    $ tox -e py310
+    $ nox -e tests -p py310
 
 OpenSSL on macOS
 ~~~~~~~~~~~~~~~~
 
 You must have installed `OpenSSL`_ via `Homebrew`_ or `MacPorts`_ and must set
-``CFLAGS`` and ``LDFLAGS`` environment variables before running ``tox``
+``CFLAGS`` and ``LDFLAGS`` environment variables before running ``nox``
```
reaperhulk, line 24: Looks like this is a section we need to update for the new env var?

noxfile.py
```diff
@@ -0,0 +1,147 @@
...
+        "-T",
+        "-W",
+        "-b",
+        "html",
+        "-d",
+        f"{temp_dir}/doctrees",
+        "docs",
+        "docs/_build/html",
+    )
+    session.run(
```
reaperhulk, line 77: Do we ever look at this output or publish it anywhere? If not, should we drop building this in CI?

docs/development/getting-started.rst
alex: Yes, but can we do that in a seperate PR, since it's not part of the cahnges her?

noxfile.py
alex: We broke it once, so we have this to catch regressions.

alex: RTD publishes the PDF AFAIK

reaperhulk (approved):

PR #8665, opened 2023-04-04
remove pytest-shard

reaperhulk:
With recent improvements we've made we're (hopefully) fast enough to entirely drop sharding. Although once chrono is gone that'll help too 😄


reaperhulk:
Not bad 
image

alex (approved):

PR #8671, opened 2023-04-05
support for X509_CRL_get0_nextUpdate and X509_CRL_get0_lastUpdate

fs111:
exposes these above OpenSSL CRL related methods so they can be consumed by other libraries.


fs111:
We are using pyopenssl and cryptography internally and for the work we are doing with CRLs we need the two functions above exposed. (This is for a reporting feature).

Please let me know if you need anything else from me to get this merged.

alex:
Is there a reason you cannot use the CRL APIs exposed by cryptography, which provide a public API for getting these values?

PR #8680, opened 2023-04-06
Don't clone the full history

alex:
32 commits should be enough for mtime-fix


reaperhulk:
This appears to be incorrect. If I clone a full depth and run the mtime fix vs a 32 depth clone the mtime dates change on quite a few files in src/rust?

alex:
Right -- they don't get their _correct_ date, instead they get the date of the 32nd oldest commit.

But we basically never have a cache that's fresh that long, 32nd oldest will basically always be older than the cache date, so it's fine.

reaperhulk:
I'm not convinced that the extra cognitive complexity of remembering that 32 is actually insufficient, but sufficient to exceed the lifetime of our caches, is worth it. It's also not a guarantee that this is faster (although it will always be strictly less total disk IO) since I believe GitHub has fast paths for depth=1 and depth=0.

Can we measure this? Or is any improvement beneath the noise floor?

alex:
On media-01, `--depth=32` was consistently about 25% faster than a full clone.

In CI, it's too noisy to measure. But! My motivation here is to reduce IOPS in the hopes to making the Rust build faster, not just to save time on the clone itself.

PR #8686, opened 2023-04-07
[debugging]

alex:


alex:
`\\fv-az89-458\\PhysicalDisk(_Total)\\Avg. Disk Queue Length` (and `Current`) seem to be the metrics I want (probably that machien name isn't what we want)

Next step: figure out how to run it in the background while `cargo` is doing. I guess I can do a spawn-process in `noxfile.py`.

PR #8700, opened 2023-04-10
support equality checks on all public asymmetric key types

reaperhulk:
I don't think the `__richcmp__` implementation for X25519/X448 is actually equivalent to the Python `__eq__` method in the other objects since it can't return `NotImplementedError`. It is consistent with how we implemented `ObjectIdentifier` though.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/dh.py
+++ b/src/cryptography/hazmat/backends/openssl/dh.py
@@ -261,6 +261,20 @@ class _DHPublicKey(dh.DHPublicKey):
     def key_size(self) -> int:
         return self._key_size_bits
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, dh.DHPublicKey):
+            return NotImplemented
+
+        res = self._backend._lib.EVP_PKEY_cmp(
+            self._evp_pkey, other._evp_pkey  # type: ignore[attr-defined]
+        )
+        if res < 0:
+            # DH public keys have two types (DH, DHX) that OpenSSL
+            # considers different types but we do not. Mismatched types
+            # push an error on the stack, so we need to consume it.
+            self._backend._consume_errors()
+        return res == 1
+
     def public_numbers(self) -> dh.DHPublicNumbers:
         p = self._backend._ffi.new("BIGNUM **")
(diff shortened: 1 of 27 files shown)


src/cryptography/hazmat/backends/openssl/dh.py
```diff
@@ -261,6 +261,20 @@ def __init__(self, backend: Backend, dh_cdata, evp_pkey):
     def key_size(self) -> int:
         return self._key_size_bits
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, dh.DHPublicKey):
+            return NotImplemented
+
+        res = self._backend._lib.EVP_PKEY_cmp(
+            self._evp_pkey, other._evp_pkey  # type: ignore[attr-defined]
+        )
```
alex, line 270: We should either type-check against the concrete class, or we should make the implementation use exclusively the public API.

src/cryptography/hazmat/backends/openssl/ed25519.py
```diff
@@ -78,6 +78,15 @@ def verify(self, signature: bytes, data: bytes) -> None:
             self._backend._consume_errors()
             raise exceptions.InvalidSignature
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Ed25519PublicKey):
+            return NotImplemented
+
+        return (
+            self._raw_public_bytes()
+            == other._raw_public_bytes()  # type:ignore[attr-defined]
```
alex, line 87: There's a public API for this! `public_bytes_raw`

src/rust/src/x509/ocsp_resp.rs
```diff
@@ -738,15 +738,17 @@ fn create_ocsp_response(
...
-            .getattr(pyo3::intern!(py, "backend"))?
-            .call_method1(
-                "_check_keys_correspond",
-                (
-                    responder_cert.call_method0("public_key")?,
-                    private_key.call_method0("public_key")?,
+        if !responder_cert
+            .call_method0("public_key")?
+            .eq(private_key.call_method0("public_key")?)
+            .unwrap()
```
alex, line 744: Don't unwrap, just `?`

alex (approved):

PR #8779, opened 2023-04-21
Do sigstore signatures when uploading to pypi

alex:


.github/workflows/pypi-publish.yml
```diff
@@ -67,3 +67,10 @@ jobs:
         shell: python
 
       - run: twine upload --skip-existing $(find dist/ -type f -name 'cryptography*')
+
+      # Do not perform sigstore signatures for things for PyPI. This is
```
reaperhulk, line 71:
```suggestion
      # Do not perform sigstore signatures for things for TestPyPI. This is
```

reaperhulk (approved):

PR #8788, opened 2023-04-22
Convert Poly1305 to Rust

alex:


alex:
this requires us to capture coverage on more systems

alex:
Sequencing here is specifically:

- @reaperhulk is working on moving our rust coverage stuff into nox
- Then I'll work to make it enabled by default in CI
- Then this should be rebasable and mergable

reaperhulk (approved):

PR #8812, opened 2023-04-24
try measuring with lto all the way off

alex:


alex:
10-25% benchmark regressions. that's a no bueno.

PR #8823, opened 2023-04-25
See if we can always enable abi3

alex:
Previously it wasn't because pypy doesn't support abi3, but maybe the pyo3 feature works.


alex:
Ok, confirmed https://github.com/PyO3/pyo3/commit/73c3911748ee0cd2a2a239498f89aaf8945bd2f5 shows this is intentional and good.

PR #8826, opened 2023-04-25
Stop invoking setup.py in wheel builder

alex:


reaperhulk (approved):
Validated the wheel names by downloading and unpacking all the zips as well.

PR #8795, opened 2023-04-23
add signature_algorithm_parameters to certificate

reaperhulk:
this allows easier verification of cert signatures, but more specifically allows PSS signature verification

PR #8859, opened 2023-05-02
Add support for different rsa signing with padding like rsa pss

bkerler:
This enables rsa pss and several other paddings for crl, csr, pkcs7 and x509.
It was designed to be compatible with previous versions and adds a new function "sign_pad", instead of just "sign", which allows to also specify the padding for the rsa signature.

This solves issue #2850, which was highly requested since 2016 and which is needed for any modern certificate signature using rsa pss.

I've also added examples to the documentation and added some tests. However, I strongly recommend to add more tests later on.

### DIFF
--- a/docs/x509/tutorial.rst
+++ b/docs/x509/tutorial.rst
@@ -56,6 +56,7 @@ a few details:
 
     >>> from cryptography import x509
     >>> from cryptography.x509.oid import NameOID
+    >>> from cryptography.hazmat.primitives.asymmetric import padding
     >>> from cryptography.hazmat.primitives import hashes
     >>> # Generate a CSR
     >>> csr = x509.CertificateSigningRequestBuilder().subject_name(x509.Name([
@@ -74,7 +75,7 @@ a few details:
     ...     ]),
     ...     critical=False,
     ... # Sign the CSR with our private key.
-    ... ).sign(key, hashes.SHA256())
+    ... ).sign_pad(key, padding.PKCS1v15(), hashes.SHA256())
     >>> # Write our CSR out to disk.
     >>> with open("path/to/csr.pem", "wb") as f:
     ...     f.write(csr.public_bytes(serialization.Encoding.PEM))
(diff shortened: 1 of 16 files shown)


reaperhulk:
Thanks for submitting, but at the moment this PR is missing some major pieces. Some notes from a quick read:

* This allows OAEP, which is not a padding that works for signing. Additionally, MGF is an abstract type intended for type checking on values passed to the PSS constructor, and MGF1 is a concrete (and actually the only) implementation of an MGF for PSS. Of these, only PKCS1v15 and PSS are actually possible signing paddings.
* This does not properly encode PSS parameters or the base PSS OID for the signature type, so the resulting signatures are incorrect (being signed over a tbsCertificate that does not contain the proper data).
* X.509 PSS support requires being able to also parse PSS parameters (see: https://github.com/pyca/cryptography/pull/8795 for work that exists on that, although it is incomplete until it is rewritten to support the proper null TLV via default defined_by in the rust asn1)

bkerler:
>     * This allows OAEP, which is not a padding that works for signing. Additionally, MGF is an abstract type intended for type checking on values passed to the PSS constructor, and MGF1 is a concrete (and actually the only) implementation of an MGF for PSS. Of these, only PKCS1v15 and PSS are actually possible signing paddings.

That should be fixed by now, thanks for pointing that out.

> 
>     * This does not properly encode PSS parameters or the base PSS OID for the signature type, so the resulting signatures are incorrect (being signed over a tbsCertificate that does not contain the proper data).

Can you elaborate more on that ? The padding options are used directly by the openssl import as far as I could see.

>     * X.509 PSS support requires being able to also parse PSS parameters (see: [add signature_algorithm_parameters to certificate #8795](https://github.com/pyca/cryptography/pull/8795) for work that exists on that, although it is incomplete until it is rewritten to support the proper null TLV via default defined_by in the rust asn1)

Certificate parsing and validation is out of the scope for this PR for now, as it's only intended to add signing support. Or should I integrate your fixes ?

alex:
> Can you elaborate more on that ? The padding options are used directly by the openssl import as far as I could see.

X.509 signatures contain an AlgorithmIdentifier (OID + parameters) for the signing algorithm used to sign them. Currently we always use the PKCS#1 OID (see `compute_signature_algorithm` in `sign.rs`), PSS signatures require a different OID + parameters, otherwise they will not be able to be verified.

bkerler:
Ok, so I do need help now, as my rust knowledge isn't good enough to implement that:
https://github.com/bkerler/cryptography/blob/padding/src/rust/src/x509/sign.rs#L282
How can I define a hash algorithm to be returned, I see that a TLV Container has to be returned,
but no idea how to construct the TLV Container to reflect the rsa pss parameters.

alex:
https://github.com/pyca/cryptography/commit/b436fafa7cf43c96f66d50162ac495c99ade1f39 is now landed, so this should help address your question.

alex:
#8888 added support for PSS for signing.

Issue #8854, opened 2023-05-02
Missing NameOID attribute for OID 2.5.4.43


harmin-parra:
Missing NameOID attribute for **Initials** (OID 2.5.4.43)

see  [rfc5280](https://www.rfc-editor.org/rfc/rfc5280)

PR #8855, opened 2023-05-02
Fix for #8854

harmin-parra:
Fix for issue #8854

PR #8860, opened 2023-05-02
Fix for issue #8854

harmin-parra:
Fix for issue #8854


alex:
This should be a part of #8855, not its own PR. Thanks.

reaperhulk:
I've pushed this commit into the #8855 branch.

PR #8874, opened 2023-05-05
Use defined_by for RSA signature AlgorithmIdentifiers

alex:
I had hoped the parameters would just be Null (no Option<>), but a review of the RFC (3447, 4055) indicates that both should be allowed, though the WebPKI enforces greater constraints.


alex:
Note to future Alex: Attempting to enforce the presence of NULL params here breaks a few tests. One of them is a DoD OCSP responder cert. Since we added that vector, DoD has fixed their OCSP responder.

reaperhulk (approved):

PR #8883, opened 2023-05-07
Try using the default LTO

alex:


alex:
@reaperhulk given the minimal impact on the benchmarks, and @lqd's measurement of a 10% or so compilation win, I think this makes sense.

If we want to preserve the perf wins for our releases, we could do that (though I'll have to think about how :D)

reaperhulk:
IIRC this was a very low perf impact so I'm fine with this.

reaperhulk (approved):

PR #8884, opened 2023-05-07
invalid visible string support

reaperhulk:
this allows utf8 in visiblestring, which is not valid DER. we raise a warning when this happens, but allow it since belgian eIDs, among others, have encoding errors. Belgium fixed this by 2021 (and possibly earlier), but their eID certificates have 10 year validity.


src/rust/src/x509/certificate.rs
```diff
@@ -412,6 +412,17 @@ fn parse_display_text(
         DisplayText::IA5String(o) => Ok(pyo3::types::PyString::new(py, o.as_str()).to_object(py)),
         DisplayText::Utf8String(o) => Ok(pyo3::types::PyString::new(py, o.as_str()).to_object(py)),
         DisplayText::VisibleString(o) => {
+            if !UnvalidatedVisibleString::verify(o.as_str()) {
```
alex, line 415: Rather than do this, can we do `if asn1::VisibleString::new(o.as_str()).is_none()`?

src/rust/cryptography-x509/src/extensions.rs
```diff
@@ -87,7 +87,7 @@ pub struct NoticeReference<'a> {
 pub enum DisplayText<'a> {
     IA5String(asn1::IA5String<'a>),
     Utf8String(asn1::Utf8String<'a>),
-    VisibleString(asn1::VisibleString<'a>),
+    VisibleString(common::UnvalidatedVisibleString<'a>),
```
alex, line 90: Can you put a comment above this explaining why its unvalidated.

src/rust/src/x509/certificate.rs
```diff
@@ -412,6 +412,17 @@ fn parse_display_text(
...
+                let cryptography_warning = py
+                    .import(pyo3::intern!(py, "cryptography.utils"))?
+                    .getattr(pyo3::intern!(py, "DeprecatedIn41"))?;
+                pyo3::PyErr::warn(
+                    py,
+                    cryptography_warning,
+                    "Invalid ASN.1 (UTF-8 characters in a VisibleString). A future version of cryptography will error on this input.",
```
alex, line 422: Can we provide more context on where the visible string is in the structure? I'd also suggest the second sentence be phrased more like "In a future version of cryptography, an exception will be raised in this circumstance."

alex (approved):

alex (approved):

PR #8888, opened 2023-05-07
support X.509 certificate PSS signing

reaperhulk:
This enables X.509 certificate signing using PSS with SHA2/SHA3 (for both hash algorithm and MGF1 hash). No support for CSR or others at this time -- I'd like to hear use cases there.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,9 @@ Changelog
 * Added support for obtaining X.509 certificate signature algorithm parameters
   (including PSS) via
   :meth:`~cryptography.x509.Certificate.signature_algorithm_parameters`.
+* Support signing :class:`~cryptography.hazmat.primitives.asymmetric.padding.PSS`
+  X.509 certificates via a new keyword-only argument on
+  :meth:`~cryptography.x509.CertificateBuilder.sign`.
 
 .. _v40-0-2:
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,7 +24,7 @@ Changelog
   (including PSS) via
   :meth:`~cryptography.x509.Certificate.signature_algorithm_parameters`.
 * Support signing :class:`~cryptography.hazmat.primitives.asymmetric.padding.PSS`
-  X.509 certificates via a new keyword-only argument on
+  X.509 certificates via the new keyword-only argument ``rsa_padding`` on
   :meth:`~cryptography.x509.CertificateBuilder.sign`.
 
 .. _v40-0-2:
--- a/src/cryptography/x509/base.py
+++ b/src/cryptography/x509/base.py
@@ -925,7 +925,9 @@ class CertificateBuilder:
         algorithm: typing.Optional[_AllowedHashTypes],
         backend: typing.Any = None,
         *,
-        rsa_padding: typing.Optional[padding.AsymmetricPadding] = None,
+        rsa_padding: typing.Optional[
+            typing.Union[padding.PSS, padding.PKCS1v15]
+        ] = None,
     ) -> Certificate:
         """
         Signs the certificate using the CA's private key.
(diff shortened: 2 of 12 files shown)


src/cryptography/x509/base.py
```diff
@@ -924,6 +924,8 @@ def sign(
         private_key: CertificateIssuerPrivateKeyTypes,
         algorithm: typing.Optional[_AllowedHashTypes],
         backend: typing.Any = None,
+        *,
+        rsa_padding: typing.Optional[padding.AsymmetricPadding] = None,
```
alex, line 928: Is there a reason not to use `Union[PSS, PKCS1v15]`? `OAEP` == guaranted type error.

(That interface seems like a mistake tbh)

src/cryptography/x509/base.py
```diff
@@ -946,7 +948,20 @@ def sign(
...
+        if rsa_padding is not None:
+            if isinstance(private_key, rsa.RSAPrivateKey):
+                if not isinstance(
+                    rsa_padding, (padding.PSS, padding.PKCS1v15)
+                ):
+                    raise TypeError(
+                        "Padding must be PSS or PKCS1v15 for RSA keys"
+                    )
+            else:
+                raise ValueError("Padding must be None for non-RSA keys")
```
alex, line 960: Suggest re-writing this to be:

```
if not isinstance(rsa_padding, (PSS, PKCS1v15):
    raise TypeError
if not isinstance(private_key, RSAPrivateKey):
    raise TypeError
````

File: src/rust/src/x509/sign.rs [hunk 3]
```diff
@@ -123,10 +123,53 @@ pub(crate) fn compute_signature_algorithm<'p>(
...
+        .extract()?;
+    // If this is RSA-PSS we need to compute the signature algorithm from the
+    // parameters provided in rsa_padding.
+    if !rsa_padding.is_none() && rsa_padding.is_instance(pss_type)? {
+        let hash_alg_params = identify_alg_params_for_hash_type(hash_type)?;
+        let hash_algorithm = common::AlgorithmIdentifier {
+            oid: asn1::DefinedByMarker::marker(),
+            params: hash_alg_params,
+        };
+        let salt_length = rsa_padding.getattr("_salt_length")?.extract::<u16>()?;
```
alex, line 146: what about the special constants?

alex: also intern!

src/rust/src/x509/sign.rs
```diff
@@ -257,14 +301,22 @@ pub(crate) fn sign_data<'p>(
...
+                    .getattr(pyo3::intern!(py, "PKCS1v15"))?
+                    .call0()?;
+                private_key
+                    .call_method1(pyo3::intern!(py, "sign"), (data, pkcs1v15, hash_algorithm))?
+            } else {
+                private_key.call_method1(
+                    pyo3::intern!(py, "sign"),
+                    (data, rsa_padding, hash_algorithm),
+                )?
+            }
```
alex, line 319: Maybe rewrite to:

```
if rsa_padding.is_one() {
    rsa_padding = ...
}
call sign
```

CHANGELOG.rst
```diff
@@ -23,6 +23,9 @@ Changelog
 * Added support for obtaining X.509 certificate signature algorithm parameters
   (including PSS) via
   :meth:`~cryptography.x509.Certificate.signature_algorithm_parameters`.
+* Support signing :class:`~cryptography.hazmat.primitives.asymmetric.padding.PSS`
+  X.509 certificates via a new keyword-only argument on
```
alex, line 27: Say what the kwarg is!

src/rust/src/x509/sign.rs
```diff
@@ -119,14 +119,88 @@ fn identify_hash_type(
...
+    } else if py_saltlen.is_instance(digestlen)? {
+        hash_algorithm
+            .getattr(pyo3::intern!(py, "digest_size"))?
+            .extract::<u16>()
+    } else if py_saltlen.is_instance(py.get_type::<pyo3::types::PyLong>())? {
+        py_saltlen.extract::<u16>()
+    } else {
+        Err(pyo3::exceptions::PyTypeError::new_err(
+            "salt_length must be an int, MaxLength, or DigestLength.",
+        ))
```
alex, line 149: I feel like this validation should be in `PSS.__init__`, not here. I say just make it an `else` and do the `extract`. If you want this validation, seperate PR to init IMO.

reaperhulk: `PSS.__init__` does enforce: https://github.com/pyca/cryptography/blob/3b8cb2b7337a7dc3a004555fd4bfa032e35a34af/src/cryptography/hazmat/primitives/asymmetric/padding.py#L47-L53
but it allows `AUTO`, which has no meaning here (it's only valid in verification contexts for recovering salt length without specifying it).

So we can do the else extract but then if they've passed auto this will fail.

alex: 😭 ok, I forgot about AUTO

alex (approved):

PR #8892, opened 2023-05-08
certificate: add a `get_extension` helper

woodruffw:
Breakout from #8873.

Needs tests, but just to get something up for review.

### DIFF
--- a/src/rust/cryptography-x509/src/certificate.rs
+++ b/src/rust/cryptography-x509/src/certificate.rs
@@ -13,6 +13,21 @@ pub struct Certificate<'a> {
     pub signature: asn1::BitString<'a>,
 }
 
+impl Certificate<'_> {
+    /// Retrieves the extension identified by the given OID,
+    /// or None if the extension is not present (or no extensions are present).
+    pub fn get_extension(&self, oid: asn1::ObjectIdentifier) -> Option<extensions::Extension> {
+        match &self.tbs_cert.extensions {
+            None => None,
+            Some(extensions) => {
+                let mut extensions = extensions.unwrap_read().clone();
+
+                extensions.find(|ext| ext.extn_id == oid)
+            }
+        }
+    }
+}
+
 #[derive(asn1::Asn1Read, asn1::Asn1Write, Hash, PartialEq, Clone)]
 pub struct TbsCertificate<'a> {
     #[explicit(0)]


src/rust/cryptography-x509/src/certificate.rs
```diff
@@ -13,6 +13,21 @@ pub struct Certificate<'a> {
...
 }
 
+impl Certificate<'_> {
+    /// Retrieves the extension identified by the given OID,
+    /// or None if the extension is not present (or no extensions are present).
+    pub fn get_extension(&self, oid: asn1::ObjectIdentifier) -> Option<extensions::Extension> {
+        match &self.tbs_cert.extensions {
+            None => None,
+            Some(extensions) => {
+                let mut extensions = extensions.unwrap_read().clone();
```
woodruffw, line 23: NB: The `clone` here is because `unwrap_read` returns a `&SequenceOf`, which we can't then mutably borrow for the `find(...)` call.

I messed around with this a bit, but couldn't find a clean way to avoid this clone. But I might have missed something 🙂

woodruffw: (I'd also like to avoid the `unwrap_read` if possible, but that appears to be a design invariant with `common::Asn1ReadableOrWritable`)

src/rust/cryptography-x509/src/certificate.rs
```diff
@@ -13,6 +13,21 @@ pub struct Certificate<'a> {
     pub signature: asn1::BitString<'a>,
 }
 
+impl Certificate<'_> {
+    /// Retrieves the extension identified by the given OID,
+    /// or None if the extension is not present (or no extensions are present).
+    pub fn get_extension(&self, oid: asn1::ObjectIdentifier) -> Option<extensions::Extension> {
```
alex, line 19: One thing to note: currently the requirement that OIDs not be repeated is enforced in the python bindings layer, but we want to be sure we're enforcing that, else you can get clowny behavior (if there's two extensions with the same OID, which wins?)

alex, line 19: Any reason not to take by ref?

src/rust/cryptography-x509/src/certificate.rs
alex: `unwrap_read` can be avoidable once we get to an MSRV with GATs (1.65.0)

src/rust/cryptography-x509/src/certificate.rs
woodruffw: Yeah, I was thinking about solutions to that -- it'd be nice if we could dedupe and do the extension uniqueness check exactly once (in the Rust layer).

Two random ideas:

1. Lazy: add a `Certificate::check_invariants()` (better name needed) or similar API, which users *must* invoke to correctly iterate over extensions, etc.
2. More involved: make `Certificate` correct by construction by having it be a newtype of `Certificate(ParsedCertificate)`, where `ParsedCertificate` is the current `CertificateType`. From there, the only allowed construction of `Certificate` would be via a `TryFrom`.

Thoughts?

woodruffw: Nope! Fixing.

src/rust/cryptography-x509/src/certificate.rs
woodruffw: Cool, resolving this thread then 🙂

src/rust/cryptography-x509/src/certificate.rs
alex: Because reasons we currently only check for duplicate extensions when accessing `.extensions`, not on initial load.

I'd probably start with a `check_duplicate_extensions()` method that can be called from both places.

woodruffw: Sounds good; that also gives me transitive coverage 🙂

src/rust/cryptography-x509/src/certificate.rs
```diff
@@ -13,6 +15,46 @@ pub struct Certificate<'a> {
     pub signature: asn1::BitString<'a>,
 }
 
+impl Certificate<'_> {
+    /// Checks whether this certificate contains any duplicated extensions;
+    /// returns the first duplicate's OID.
+    ///
+    /// If no duplicates are found, returns `None`.
+    pub fn check_duplicate_extensions(&self) -> Option<asn1::ObjectIdentifier> {
```
woodruffw, line 23: Here's the new helper; I removed the corresponding code in the Python bindings and inserted a call to it.

(I haven't inserted the same call inside of `get_extension`, since I figure it should happen in the callsite(s) to `get_extension`).

woodruffw: (I also broke it, debugging now.)

woodruffw: Oh, this is my own fault -- I thought `extensions` was the only place this gets built up, but there are a few others. So this duplication logic needs to be reusable from contexts that don't have a full `Certificate`.

woodruffw:
Should be good to go now, modulo coverage -- I've rewritten existing `Extensions` usage into `RawExtensions` (and added a `raw_` prefix as appropriate), with the new `Extensions` type being a duplicate-checking wrapper.

The existing exception behavior is preserved by having the `Err(...)` variant carry the duplicate OID.

src/rust/cryptography-x509/src/certificate.rs
```diff
@@ -13,6 +14,12 @@ pub struct Certificate<'a> {
     pub signature: asn1::BitString<'a>,
 }
 
+impl<'a> Certificate<'a> {
+    pub fn extensions(&'a self) -> Result<Option<Extensions<'a>>, asn1::ObjectIdentifier> {
+        self.tbs_cert.extensions()
```
alex, line 19: Not sure it's worth it to have the method on TBS, is there a use case?

src/rust/cryptography-x509/src/certificate.rs
```diff
@@ -31,7 +38,16 @@ pub struct TbsCertificate<'a> {
...
-    pub extensions: Option<extensions::Extensions<'a>>,
+    pub raw_extensions: Option<extensions::RawExtensions<'a>>,
+}
+
+impl<'a> TbsCertificate<'a> {
+    pub fn extensions(&'a self) -> Result<Option<Extensions<'a>>, asn1::ObjectIdentifier> {
+        match &self.raw_extensions {
+            None => Ok(None),
+            Some(extensions) => Some(extensions.try_into()).transpose(),
+        }
```
alex, line 49: This code block is repeated everywhere, instead of this `try_into` thing can we just do `Extensions::from_raw_extensions(raw_extensions)` or something and then it can have the match?

src/rust/src/x509/certificate.rs
```diff
@@ -193,9 +193,17 @@ impl Certificate {
...
-                let readable_extensions = extensions.unwrap_read().clone();
+        match val.tbs_cert.extensions() {
+            Err(oid) => {
+                let oid_obj = oid_to_py_oid(py, &oid)?;
+                Err(exceptions::DuplicateExtension::new_err((
+                    format!("Duplicate {} extension found", oid),
+                    oid_obj.into_py(py),
+                ))
+                .into())
+            }
```
alex, line 204: Err goes last in the match statement please

src/rust/src/x509/certificate.rs
```diff
@@ -193,9 +193,17 @@ impl Certificate {
...
+            Ok(Some(extensions)) => {
+                let readable_extensions = extensions.as_ref().unwrap_read().clone();
```
alex, line 206: If this is the use case for `as_ref()`, I wouldn't use the trait, just add a regular method

src/rust/src/x509/crl.rs
```diff
@@ -260,11 +260,24 @@ impl CertificateRevocationList {
...
+            Ok(exts) => exts,
+            Err(oid) => {
+                let oid_obj = oid_to_py_oid(py, &oid)?;
+                return Err(exceptions::DuplicateExtension::new_err((
+                    format!("Duplicate {} extension found", oid),
+                    oid_obj.into_py(py),
+                )));
+            }
+        };
+
```
alex, line 275: This shouldn't be repeated in every `extensions` method, it should be in `parse_and_cache_extensions` which can take the result of calling `extensions()` on the cert.

src/rust/cryptography-x509/src/certificate.rs
woodruffw: That was mostly for `tbs_precertificate_bytes`, since we need `TbcCertificate.extensions()` in that context to perform the poison extension filtering. But maybe there's another way I should be getting the extensions there?

src/rust/src/x509/crl.rs
woodruffw: Making sure I understand: in that case, `parse_and_cache_extensions` should take a `Result<Option<Extensions>>`? Because that'll be the return type from `extensions()`.

(Or are you thinking it should take the `RawExtensions`, and then convert it to an `Extensions` internally? That would avoid passing a potential `Err` variant through as a parameter.)

src/rust/cryptography-x509/src/certificate.rs
woodruffw: Ah yep, I can get rid of this entirely.

src/rust/src/x509/certificate.rs
woodruffw: Done; made it `as_raw`.

src/rust/src/x509/crl.rs
woodruffw: I've done this with `RawExtensions`; the error handling is now pushed fully into `parse_and_cache_extensions`.

alex:
clippy

woodruffw:
GitHub is struggling today, but I've added some coverage for the new `get_extension` helper.

woodruffw:
Remaining coverage items:

* `tbs_precertificate_bytes` needs a testcase for duplicate extensions (these weren't previously handled)
* `Certificate::extensions` needs a testcase

woodruffw:
This should be fully good to go: I've removed `Certificate::extensions` in favor of an extensional trait in that path validation library being scaffolded in #8873, which means we can punt on coverage.

src/rust/cryptography-x509/src/extensions.rs
```diff
@@ -2,16 +2,58 @@
...
 >;
 
+pub struct Extensions<'a> {
+    inner: RawExtensions<'a>,
+}
+
+impl<'a> Extensions<'a> {
+    pub fn from_raw_extensions(
+        raw: Option<&RawExtensions<'a>>,
+    ) -> Result<Option<Self>, asn1::ObjectIdentifier> {
```
woodruffw, line 24: I'm not super happy with this signature -- we end up pushing an `Option` through as a no-op to avoid having to destructure the optional extensions in multiple other places, but that results in more invalid states than are strictly needed here (and the smelly `Result<Option<...>>` type).

alex: I think a reasonable follow up is, instead of `as_raw()`, we make `Extensions` into an iterator, and just handle the `Option<>` internally. But let's do that in a seperate PR, this one is already complicated enough.

src/rust/src/x509/certificate.rs
```diff
@@ -207,18 +207,26 @@ impl Certificate {
...
                 ),
             )),
+            Err(oid) => {
+                let oid_obj = oid_to_py_oid(py, &oid)?;
+                Err(exceptions::DuplicateExtension::new_err((
+                    format!("Duplicate {} extension found", oid),
+                    oid_obj.into_py(py),
+                ))
+                .into())
+            }
```
alex, line 229: To avoid needing to repeat this stanza anywhere we touch extensions, can we make `extensions` return a more-specific error type and then we can do a `From` impl on that?

src/rust/src/x509/common.rs
```diff
@@ -391,27 +390,30 @@ pub(crate) fn parse_and_cache_extensions<
 >(
     py: pyo3::Python<'p>,
     cached_extensions: &mut Option<pyo3::PyObject>,
-    raw_exts: &Option<Extensions<'_>>,
+    raw_extensions: &Option<RawExtensions<'_>>,
```
alex, line 393: I think we actually want this method to take an `Option<Extensions>` and then the callers can obtain the extensions, via an `extensions()` method on the various `cryptography-x509` types. That way those types have the right methods that can be used from the validator crate.

src/rust/src/x509/certificate.rs
woodruffw: Yeah, good idea -- I'll do that now.

src/rust/src/x509/common.rs
woodruffw: That SGTM, although if I understand correctly doing that will require us to make the `Iterator` change mentioned in  https://github.com/pyca/cryptography/pull/8892#discussion_r1189196107: without that change, we'd need to perform the fallible conversion from `RawExtensions -> Extensions` before `parse_and_cache_extensions`, defeating the point of having it inside of that function. Or did I miss something?

alex: Ok, let's leave this change for a seperate PR.

src/rust/src/x509/certificate.rs
woodruffw: https://github.com/pyca/cryptography/pull/8892/commits/212b75ff2f69a3b3cfc9d6a55949f23877f8f618 is my attempt at this -- it saves us the repetition of this error clause in exchange for a new error enum + `From` implementation.

alex (approved):

PR #8896, opened 2023-05-09
Attempt to use sccache

alex:


alex:
awaiting a resolution to https://github.com/mozilla/sccache/issues/1250

reaperhulk:
Do we have a run that shows the ideal caching? In the last two successful runs it's still spending 30-60s compiling the rust despite nothing changing, but that may be just because the cache isn't hot? I'd like to understand best case here.

alex:
Take a look at the post-cache output in the job. What you'll see is that for most of them, there's close to a 100% cache rate... for cacheable items. Unfortunately not every compilation is cacheable, see https://github.com/mozilla/sccache#rust

reaperhulk:
I don't love that, but I guess we'll merge and see how this feels...

reaperhulk (approved):

reaperhulk (approved):

Issue #8907, opened 2023-05-11
AlgorithmIdentifier null parameters


reaperhulk:
While researching some PSS behaviors I discovered that LibreSSL omits the `NULL` parameters on the hash function algorithm identifiers. However, [RFC 4055 section 2.1](https://www.rfc-editor.org/rfc/rfc4055#section-2.1) states `All implementations MUST accept both NULL and absent parameters as legal and equivalent encodings.`, but makes another apparently conflicting statement right after that.

Confusing statements in the RFC notwithstanding, should we allow parsing `Option<asn1::Null>` on these `AlgorithmIdentifier` parameters?

PR #8908, opened 2023-05-11
support PSS signatures in verify_directly_issued_by

reaperhulk:
This performs a significant refactoring in moving the impl of `signature_algorithm_parameters` into a standalone func `identify_signature_algorithm_parameters` and then using that in `verify_signature_with_signature_parameters` (which is a refactored `verify_signature_with_oid`).

Outstanding question: this refactor likely means CSR and CRL could verify a PSS signature, but we don't permit creating them and I don't have any test vectors so we should decide one of the following:
* Don't care, they're unsupported for now.
* Do care, we need test vectors (either implementing support so we can create them or digging them up somewhere)

This work triggered filing #8907 as well.

### DIFF
--- a/src/rust/src/x509/sign.rs
+++ b/src/rust/src/x509/sign.rs
@@ -4,6 +4,7 @@
 
 use crate::error::{CryptographyError, CryptographyResult};
 use crate::exceptions;
+use crate::x509::certificate;
 use cryptography_x509::{common, oid};
 
 #[derive(Debug, PartialEq)]
@@ -15,7 +16,6 @@ pub(crate) enum KeyType {
     Ed448,
 }
 
-#[derive(Debug, PartialEq)]
 enum HashType {
     None,
     Sha224,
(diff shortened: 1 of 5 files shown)


reaperhulk:
(Need to look at coverage)

alex:
Can you seperate out the refactoring parts of this into their own PR?

src/rust/src/x509/sign.rs
```diff
@@ -351,74 +351,46 @@ pub(crate) fn sign_data<'p>(
...
-    let hashes = py.import(pyo3::intern!(py, "cryptography.hazmat.primitives.hashes"))?;
-    let signature_hash = match sig_hash_name {
-        Some(data) => hashes.getattr(data)?.call0()?,
-        None => py.None().into_ref(py),
-    };
-
+    let py_signature_algorithm_parameters =
+        certificate::identify_signature_algorithm_parameters(py, signature_algorithm)?;
+    let py_signature_hash_algorithm =
+        certificate::identify_signature_hash_algorithm(py, signature_algorithm)?;
```
alex, line 373: I'm going to merge these as-is, but can you please do a follow-up PR to move them from `certificate.rs` to `sign.rs`.

alex (approved):

PR #8921, opened 2023-05-13
refactor signature algorithm parameters into a separate function

reaperhulk:
this will be used in the verify_directly_issued_by PR

### DIFF
--- a/src/rust/src/x509/certificate.rs
+++ b/src/rust/src/x509/certificate.rs
@@ -304,67 +304,11 @@ impl Certificate {
         &'p self,
         py: pyo3::Python<'p>,
     ) -> CryptographyResult<&'p pyo3::PyAny> {
-        match &self.raw.borrow_value().signature_alg.params {
-            common::AlgorithmParameters::RsaPss(opt_pss) => {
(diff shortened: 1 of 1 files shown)


alex (dismissed):

src/rust/src/x509/certificate.rs
```diff
@@ -1076,6 +1020,74 @@ pub(crate) fn set_bit(vals: &mut [u8], n: usize, set: bool) {
...
 
+pub(crate) fn identify_signature_algorithm_parameters<'p>(
+    py: pyo3::Python<'p>,
+    params: &common::AlgorithmParameters<'p>,
+    signature_hash_algorithm: &pyo3::PyAny,
+) -> CryptographyResult<&'p pyo3::PyAny> {
+    match params {
+        common::AlgorithmParameters::RsaPss(opt_pss) => {
+            let pss = opt_pss.as_ref().ok_or_else(|| {
+                pyo3::exceptions::PyValueError::new_err("Invalid RSA PSS parameters")
```
alex, line 1031: This is not reachable because this error will already be raised by the `signature_hash_algorithm` call, which is now happens first.

alex (approved):

PR #8931, opened 2023-05-15
don't use a set

reaperhulk:
We don't need one here and it creates ordering instability when iterating over an RDN


alex:
I think so, yes.

alex (approved):

PR #8978, opened 2023-05-28
Port DSA to Rust

alex:
Needs a rust-openssl release


src/rust/src/backend/utils.rs
```diff
@@ -45,6 +46,7 @@ pub(crate) fn pkey_private_bytes<'p>(
     format: &pyo3::PyAny,
     encryption_algorithm: &pyo3::PyAny,
     openssh_allowed: bool,
+    raw_allowed: bool,
```
reaperhulk, line 49: This is an ugly function signature, although I don't have a lot of ideas on how to make it better.

alex: Dislike the various `_allowed`? We could add a `SerializationConfig` with fields for each of them. Seems like only a marginal improvement though.

reaperhulk: Yeah I don't like the alloweds, but I agree that a struct probably isn't much better here.

src/rust/src/backend/dsa.rs
```diff
@@ -0,0 +1,333 @@
...
+    let parameter_numbers = public_numbers.getattr(pyo3::intern!(py, "parameter_numbers"))?;
+
+    let dsa = openssl::dsa::Dsa::from_private_components(
+        utils::py_int_to_bn(py, parameter_numbers.getattr(pyo3::intern!(py, "p"))?)?,
+        utils::py_int_to_bn(py, parameter_numbers.getattr(pyo3::intern!(py, "q"))?)?,
+        utils::py_int_to_bn(py, parameter_numbers.getattr(pyo3::intern!(py, "g"))?)?,
+        utils::py_int_to_bn(py, numbers.getattr(pyo3::intern!(py, "x"))?)?,
+        utils::py_int_to_bn(py, public_numbers.getattr(pyo3::intern!(py, "y"))?)?,
+    )
+    .unwrap();
```
reaperhulk, line 71: Why unwrap? We handle `ErrorStack` conversion to `CryptographyError` already so should we be doing `?`?

alex: Two parter:

1) The `?` is uncovered if I use it, because...
2) These _cannot fail_, I reviewed both rust-openssl and OpenSSL itself. There is no way for it to fail.

reaperhulk (approved):

PR #8981, opened 2023-05-29
bump for 40.0.3

reaperhulk:
testing CI for the upcoming release we'll need to do


reaperhulk:
Ran out of time for the moment, but it appears there's a bindgen issue with boring as well as a pip problem for py312.

alex:
https://github.com/pyca/cryptography/commit/ec5d5d77f3eb6f49dbd4d6c8b210d2c46481076b has the fix for boring, it required a rust-openssl bump

PR #8989, opened 2023-05-30
openssl backend.py: do not allow RSA public exponent < F4

paulwouters:
Public Exponents really need to be at least the size of F4 (65537).

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -396,7 +396,7 @@ class Backend:
         self, public_exponent: int, key_size: int
     ) -> bool:
         return (
-            public_exponent >= 3
+            public_exponent >= 65537
             and public_exponent & 1 != 0
             and key_size >= 512
         )
--- a/src/cryptography/hazmat/primitives/asymmetric/rsa.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/rsa.py
@@ -140,11 +140,8 @@ def generate_private_key(
 
 
 def _verify_rsa_parameters(public_exponent: int, key_size: int) -> None:
-    if public_exponent not in (3, 65537):
-        raise ValueError(
-            "public_exponent must be either 3 (for legacy compatibility) or "
-            "65537. Almost everyone should choose 65537 here!"
-        )
+    if public_exponent != 65537:
+        raise ValueError("public_exponent must be 65537")
 
     if key_size < 512:
         raise ValueError("key_size must be at least 512-bits.")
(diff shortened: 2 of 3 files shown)


gmfull2014:
Hello, I have several questions. 1. I have a slight doubt, can the public factor be variable? Why should it be within that range and I don't know a random prime number? 2. I am trying to create a storage for the private keys but the class where the private key is stored is an object and it does not let me manipulate it or store it in a file, I want to create a place where I can store them and when receiving an encrypted file I can identify the private key. Is there something where I can investigate about this?

reaperhulk:
@paulwouters The change you've made only affects a method that we use for testing (`generate_rsa_parameters_supported`) and doesn't actually restrict the generate method. If you look at our generation method you can see we actually only accept two exponents:
https://github.com/pyca/cryptography/blob/dc7330e9f071d510227e625dfe134297276ef095/src/cryptography/hazmat/primitives/asymmetric/rsa.py#L142-L147

gmfull2014:
> @paulwouters The change you've made only affects a method that we use for testing (`generate_rsa_parameters_supported`) and doesn't actually restrict the generate method. If you look at our generation method you can see we actually only accept two exponents:
> 
> https://github.com/pyca/cryptography/blob/dc7330e9f071d510227e625dfe134297276ef095/src/cryptography/hazmat/primitives/asymmetric/rsa.py#L142-L147

Hello, how are you? I tried to change the public exponent, disabling the verification. I try to understand if I can generate a search for a prime number with urandom but I have problems with the data types. Another query can I handle the private and public key classes directly or should I serialize them. Another query, once the text is encrypted, how can I serialize it? I can't do it because serialization only leaves me with classes.objects

reaperhulk:
@gmfull2014 this is a PR to this project that is unrelated to the topic you are trying to discuss. Please open a separate issue if you have a problem, but remember that this issue tracker is not a general support forum and we can't help with basic questions about cryptography.

gmfull2014:
> this is a PR to this project that is unrelated to the topic you are trying to discuss. Please open a separate issue if you have a problem, but remember that this issue tracker is not a general support forum and we can't help with basic questions about cryptography.

Ok I understand thank you very much. Somewhere where I can document myself a little more, I have already reviewed the documents. I appreciate your help

paulwouters:
> @paulwouters The change you've made only affects a method that we use for testing (`generate_rsa_parameters_supported`) and doesn't actually restrict the generate method. If you look at our generation method you can see we actually only accept two exponents:
> 
> https://github.com/pyca/cryptography/blob/dc7330e9f071d510227e625dfe134297276ef095/src/cryptography/hazmat/primitives/asymmetric/rsa.py#L142-L147

thanks. I fixed it there as well.

reaperhulk:
As noted in the exception in the code, there are specific compatibility cases where we unfortunately want to allow 3. Since we only accept 3 and 65537 (and our [documentation](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.rsa.generate_private_key) is quite loud about 65537 being the right choice) I don't think there's really anything to be done here.

PR #8994, opened 2023-05-31
AWS Lambda layer - Update installation.rst

johndpope:


reaperhulk:
This PR adds a list item to say that we test on lambda, but we do not. It also adds instructions, but we currently link to https://docs.aws.amazon.com/lambda/latest/dg/python-package.html and https://docs.aws.amazon.com/lambda/latest/dg/python-image.html in a [FAQ entry](https://cryptography.io/en/latest/faq/#i-m-getting-errors-installing-or-importing-cryptography-on-aws-lambda) about lambda so that we don't have to maintain instructions for a platform that shifts over time. If you think we should link those in the installation section we'd be happy to review a PR doing that though!

johndpope:
those instructions are not fruitful - I waste days following instructions. Try it.

alex:
Perhaps you should direct that complaints to _Amazon_, who are both the creators of that documentation and the creators of Lambda.

Paul has said what needs to happen for this PR to be mergable.

PR #9002, opened 2023-05-31
tolerate NULL params in ECDSA SHA2 AlgorithmIdentifier

reaperhulk:
Java 11 does this incorrectly. It was fixed in Java16+ and they are planning to do a backport, but we'll need to tolerate this invalid encoding for a while.


src/rust/src/x509/certificate.rs
```diff
@@ -391,6 +391,9 @@ fn load_der_x509_certificate(
     // determine if the serial is negative and raise a warning if it is. We want to drop support
     // for this sort of invalid encoding eventually.
     warn_if_negative_serial(py, raw.borrow_value().tbs_cert.serial.as_bytes())?;
+    // determine if the signature algorithm has incorrect parameters and raise a warning if it
+    // does. this is a bug in JDK11 and we want to drop support for it eventually.
+    warn_if_invalid_ecdsa_params(py, raw.borrow_value().signature_alg.clone())?;
```
alex, line 396: Should we check both the inner and outer sig alg?

src/rust/src/x509/certificate.rs
```diff
@@ -418,24 +418,38 @@ fn warn_if_negative_serial(py: pyo3::Python<'_>, bytes: &'_ [u8]) -> pyo3::PyRes
...
+            .signature_alg
+            .params
+            .clone(),
+        owned_certificate
+            .borrow_value()
+            .tbs_cert
+            .signature_alg
+            .params
+            .clone(),
+    ] {
```
alex, line 435: This is way too convoluted. Just use the function from the previous commit and call it 2x, no need for this for loop.

alex (approved):

PR #9024, opened 2023-06-04
Migrate EC support to Rust

alex:


alex:
Note to self: Do we have a test that verifies that we always serialize in named curve form?

src/rust/src/backend/ec.rs
```diff
@@ -0,0 +1,570 @@
...
+                exceptions::UnsupportedAlgorithm::new_err((
+                    "Unsupported EC exchange algorithm",
+                    exceptions::Reasons::UNSUPPORTED_EXCHANGE_ALGORITHM,
+                )),
+            ));
+        }
+
+        let mut deriver = openssl::derive::Deriver::new(&self.pkey)?;
+        #[cfg(CRYPTOGRAPHY_OPENSSL_300_OR_GREATER)]
+        deriver
```
reaperhulk, line 332: Worth a comment here to explain why we don't re-validate?

src/rust/src/backend/ec.rs
```diff
@@ -0,0 +1,570 @@
...
+                ),
+            )?
+            .extract()?;
+
+        let mut verifier = openssl::pkey_ctx::PkeyCtx::new(&self.pkey)?;
+        verifier.verify_init()?;
+        let valid = verifier.verify(data, signature).unwrap_or(false);
+        // TODO: Empty the error stack. BoringSSL leaves one in the event of
+        // signature validation failure. Upstream to rust-openssl?
+        openssl::error::ErrorStack::get();
```
reaperhulk, line 502: Should we annotate this for just boring?

alex: Done, but also, what do you think about upstreaming?

reaperhulk: I think we should upstream. Without doing so every caller needs to know that they need this conditional or else their error stack gets polluted and who knows what their error handling does.

reaperhulk (approved):

PR #9043, opened 2023-06-08
Raise MSRV to 1.63.0

alex:


reaperhulk (approved):

reaperhulk:
The numbers support this and MSRV has thus far proven to be a bit less of an obstacle than previously expected.

PR #9049, opened 2023-06-10
Added several tests and cleanups for serialization and EC

alex:


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -990,6 +993,11 @@ def _handle_key_loading_error(self) -> typing.NoReturn:
             )
 
     def elliptic_curve_supported(self, curve: ec.EllipticCurve) -> bool:
+        if self._fips_enabled and not isinstance(
```
reaperhulk, line 996: This changes the logic such that curves unsupported for ECDH are also considered generally unsupported. Is that correct? (It might be! It probably is?)

alex: My understanding is that proper adherence to FIPS means that you're only supposed to do ECDSA with the correct curves, and it's basically a bug in OpenSSL's FIPS that it happens to work sometimes.

reaperhulk (approved):

PR #9084, opened 2023-06-17
Remove a bunch of pointless indirections

alex:


reaperhulk (approved):
Backend isn't a public object any more and hasn't been for a few years, but let's see if this causes us some heartache after release 😄

PR #9103, opened 2023-06-20
Don't build dh code on BoringSSL

alex:
It doesn't support DH via EVP


alex:
Relevant context: `from_dh` is removed on BoringSSL in teh latest `openssl` crate. It used to simply always fail. This change hides all methods that use that function on boringssl. They are never invoked because `dh_supported()` returns `False` on boring.

alex:
It'll actually work. BoringSSL has DH, it just doesn't have EVP support for DH.

alex:
That was true before this PR as well, to be clear.

reaperhulk (approved):

PR #9113, opened 2023-06-22
CI: Update GitHub owned actions to be referenced by SHA

diogoteles08:
Implementation of first part of the issue #9073

### DIFF
--- a/.github/workflows/auto-close-stale.yml
+++ b/.github/workflows/auto-close-stale.yml
@@ -13,7 +13,7 @@ jobs:
     if: github.repository_owner == 'pyca'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/stale@v8.0.0
+      - uses: actions/stale@1160a2240286f5da8ec72b1c0816ce2481aabf84 # v8.0.0
         with:
           only-labels: waiting-on-reporter
           days-before-stale: 3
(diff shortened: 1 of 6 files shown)


.github/workflows/auto-close-stale.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     if: github.repository_owner == 'pyca'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/stale@v8.0.0
+      - uses: actions/stale@1160a2240286f5da8ec72b1c0816ce2481aabf84 # v8.0.0
```
reaperhulk, line 16: do these comments with versions get updated as part of the dependabot bumps? If not we should probably just drop them since they'll swiftly become inaccurate.

diogoteles08: Yes they do! Dependabot also takes care of updating the comments

reaperhulk: Hey, it has improved significantly since the last time I looked. Thanks @diogoteles08

alex: Neat! We should go add comments for all of other hashes at some point.

reaperhulk (approved):

PR #9128, opened 2023-06-23
Stop caching cargo and pip dirs

alex:
They're of very limited value (I think), and they're some of the most unreliable parts of our infra.


woodruffw:
FWIW, you may be able to use `setup-python`'s own caching layer: https://github.com/actions/setup-python#caching-packages-dependencies

YMMV on how well it works for you -- it's very reliable for me on other (less complicated) projects.

reaperhulk (approved):

PR #9129, opened 2023-06-23
Fix include directory when cross compiling

### LINKED ISSUE #9119: fatal error: pyconfig.h: No such file or directory when cross compiling
I'm seeing this build failure starting with cryptography version `40.0.0`(`39.0.2` builds fine) when cross compiling with [buildroot](https://buildroot.org/).

```rust
     Running `rustc --crate-name build_script_build --edition=2018 /home/buildroot/buildroot/output/build/python-cryptography-41.0.1/VENDOR/pyo3/build.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --diagnostic-width=209 --crate-type bin --emit=dep-info,link -C embed-bitcode=no -C debug-assertions=off -C overflow-checks=on --cfg 'feature="abi3"' --cfg 'feature="abi3-py310"' --cfg 'feature="abi3-py311"' --cfg 'feature="abi3-py37"' --cfg 'feature="abi3-py38"' --cfg 'feature="abi3-py39"' --cfg 'feature="default"' --cfg 'feature="extension-module"' --cfg 'feature="indoc"' --cfg 'feature="macros"' --cfg 'feature="pyo3-macros"' --cfg 'feature="unindent"' -C metadata=11bdde305c3ecfe9 -C extra-filename=-11bdde305c3ecfe9 --out-dir /home/buildroot/buildroot/output/build/python-cryptography-41.0.1/src/rust/target/release/build/pyo3-11bdde305c3ecfe9 -L dependency=/home/buildroot/buildroot/output/build/python-cryptography-41.0.1/src/rust/target/release/deps --extern pyo3_build_config=/home/buildroot/buildroot/output/build/python-cryptography-41.0.1/src/rust/target/release/deps/libpyo3_build_config-36a51f7ec3cba2f9.rlib --cap-lints allow -C link-args=-L/home/buildroot/buildroot/output/per-package/python-cryptography/host/lib -C link-args=-Wl,-rpath,/home/buildroot/buildroot/output/per-package/python-cryptography/host/lib`
The following warnings were emitted during compilation:

warning: In file included from /home/buildroot/buildroot/output/build/python-cryptography-41.0.1/src/rust/target/aarch64-unknown-linux-gnu/release/build/cryptography-cffi-4ab5309155355ef6/out/_openssl.c:57:
warning: /home/buildroot/buildroot/output/build/python3-3.11.4/Include/Python.h:12:10: fatal error: pyconfig.h: No such file or directory
warning:    12 | #include "pyconfig.h"
warning:       |          ^~~~~~~~~~~~
warning: compilation terminated.

error: failed to run custom build command for `cryptography-cffi v0.1.0 (/home/buildroot/buildroot/output/build/python-cryptography-41.0.1/src/rust/cryptography-cffi)`

Caused by:
  process didn't exit successfully: `/home/buildroot/buildroot/output/build/python-cryptography-41.0.1/src/rust/target/release/build/cryptography-cffi-0f846530999218fc/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-env-changed=PYO3_PYTHON
  cargo:rerun-if-changed=../../_cffi_src/
  cargo:rustc-cfg=python_implementation="CPython"
  TARGET = Some("aarch64-unknown-linux-gnu")
  OPT_LEVEL = Some("3")
  HOST = Some("x86_64-unknown-linux-gnu")
  cargo:rerun-if-env-changed=CC_aarch64-unknown-linux-gnu
  CC_aarch64-unknown-linux-gnu = None
  cargo:rerun-if-env-changed=CC_aarch64_unknown_linux_gnu
  CC_aarch64_unknown_linux_gnu = None
  cargo:rerun-if-env-changed=TARGET_CC
  TARGET_CC = None
  cargo:rerun-if-env-changed=CC
  CC = Some("/home/buildroot/buildroot/output/per-package/python-cryptography/host/bin/aarch64-linux-gcc")
  cargo:rerun-if-env-changed=CFLAGS_aarch64-unknown-linux-gnu
  CFLAGS_aarch64-unknown-linux-gnu = None
  cargo:rerun-if-env-changed=CFLAGS_aarch64_unknown_linux_gnu
  CFLAGS_aarch64_unknown_linux_gnu = None
  cargo:rerun-if-env-changed=TARGET_CFLAGS
  TARGET_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS
  CFLAGS = Some("-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -Os -g0 -D_FORTIFY_SOURCE=1")
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
  DEBUG = Some("false")
  CARGO_CFG_TARGET_FEATURE = Some("neon")
  cargo:rerun-if-env-changed=CC_aarch64-unknown-linux-gnu
  CC_aarch64-unknown-linux-gnu = None
  cargo:rerun-if-env-changed=CC_aarch64_unknown_linux_gnu
  CC_aarch64_unknown_linux_gnu = None
  cargo:rerun-if-env-changed=TARGET_CC
  TARGET_CC = None
  cargo:rerun-if-env-changed=CC
  CC = Some("/home/buildroot/buildroot/output/per-package/python-cryptography/host/bin/aarch64-linux-gcc")
  cargo:rerun-if-env-changed=CFLAGS_aarch64-unknown-linux-gnu
  CFLAGS_aarch64-unknown-linux-gnu = None
  cargo:rerun-if-env-changed=CFLAGS_aarch64_unknown_linux_gnu
  CFLAGS_aarch64_unknown_linux_gnu = None
  cargo:rerun-if-env-changed=TARGET_CFLAGS
  TARGET_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS
  CFLAGS = Some("-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -Os -g0 -D_FORTIFY_SOURCE=1")
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
  CARGO_CFG_TARGET_FEATURE = Some("neon")
  cargo:rerun-if-env-changed=CC_aarch64-unknown-linux-gnu
  CC_aarch64-unknown-linux-gnu = None
  cargo:rerun-if-env-changed=CC_aarch64_unknown_linux_gnu
  CC_aarch64_unknown_linux_gnu = None
  cargo:rerun-if-env-changed=TARGET_CC
  TARGET_CC = None
  cargo:rerun-if-env-changed=CC
  CC = Some("/home/buildroot/buildroot/output/per-package/python-cryptography/host/bin/aarch64-linux-gcc")
  cargo:rerun-if-env-changed=CFLAGS_aarch64-unknown-linux-gnu
  CFLAGS_aarch64-unknown-linux-gnu = None
  cargo:rerun-if-env-changed=CFLAGS_aarch64_unknown_linux_gnu
  CFLAGS_aarch64_unknown_linux_gnu = None
  cargo:rerun-if-env-changed=TARGET_CFLAGS
  TARGET_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS
  CFLAGS = Some("-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -Os -g0 -D_FORTIFY_SOURCE=1")
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
  CARGO_CFG_TARGET_FEATURE = Some("neon")
  cargo:rerun-if-env-changed=CC_aarch64-unknown-linux-gnu
  CC_aarch64-unknown-linux-gnu = None
  cargo:rerun-if-env-changed=CC_aarch64_unknown_linux_gnu
  CC_aarch64_unknown_linux_gnu = None
  cargo:rerun-if-env-changed=TARGET_CC
  TARGET_CC = None
  cargo:rerun-if-env-changed=CC
  CC = Some("/home/buildroot/buildroot/output/per-package/python-cryptography/host/bin/aarch64-linux-gcc")
  cargo:rerun-if-env-changed=CFLAGS_aarch64-unknown-linux-gnu
  CFLAGS_aarch64-unknown-linux-gnu = None
  cargo:rerun-if-env-changed=CFLAGS_aarch64_unknown_linux_gnu
  CFLAGS_aarch64_unknown_linux_gnu = None
  cargo:rerun-if-env-changed=TARGET_CFLAGS
  TARGET_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS
  CFLAGS = Some("-D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -Os -g0 -D_FORTIFY_SOURCE=1")
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
  CARGO_CFG_TARGET_FEATURE = Some("neon")
  running: "/home/buildroot/buildroot/output/per-package/python-cryptography/host/bin/aarch64-linux-gcc" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-D_LARGEFILE_SOURCE" "-D_LARGEFILE64_SOURCE" "-D_FILE_OFFSET_BITS=64" "-Os" "-g0" "-D_FORTIFY_SOURCE=1" "-I" "/home/buildroot/buildroot/output/build/python3-3.11.4/Include" "-I" "/home/buildroot/buildroot/output/per-package/python-cryptography/host/bin/../aarch64-buildroot-linux-gnu/sysroot/usr/include" "-Wconversion" "-Wno-error=sign-conversion" "-Wno-unused-parameter" "-DPy_LIMITED_API=0x030700f0" "-o" "/home/buildroot/buildroot/output/build/python-cryptography-41.0.1/src/rust/target/aarch64-unknown-linux-gnu/release/build/cryptography-cffi-4ab5309155355ef6/out/2f1fe1929bb62df3-_openssl.o" "-c" "/home/buildroot/buildroot/output/build/python-cryptography-41.0.1/src/rust/target/aarch64-unknown-linux-gnu/release/build/cryptography-cffi-4ab5309155355ef6/out/_openssl.c"
  cargo:warning=In file included from /home/buildroot/buildroot/output/build/python-cryptography-41.0.1/src/rust/target/aarch64-unknown-linux-gnu/release/build/cryptography-cffi-4ab5309155355ef6/out/_openssl.c:57:
  cargo:warning=/home/buildroot/buildroot/output/build/python3-3.11.4/Include/Python.h:12:10: fatal error: pyconfig.h: No such file or directory
  cargo:warning=   12 | #include "pyconfig.h"
  cargo:warning=      |          ^~~~~~~~~~~~
  cargo:warning=compilation terminated.
  exit status: 1

  --- stderr


  error occurred: Command "/home/buildroot/buildroot/output/per-package/python-cryptography/host/bin/aarch64-linux-gcc" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-D_LARGEFILE_SOURCE" "-D_LARGEFILE64_SOURCE" "-D_FILE_OFFSET_BITS=64" "-Os" "-g0" "-D_FORTIFY_SOURCE=1" "-I" "/home/buildroot/buildroot/output/build/python3-3.11.4/Include" "-I" "/home/buildroot/buildroot/output/per-package/python-cryptography/host/bin/../aarch64-buildroot-linux-gnu/sysroot/usr/include" "-Wconversion" "-Wno-error=sign-conversion" "-Wno-unused-parameter" "-DPy_LIMITED_API=0x030700f0" "-o" "/home/buildroot/buildroot/output/build/python-cryptography-41.0.1/src/rust/target/aarch64-unknown-linux-gnu/release/build/cryptography-cffi-4ab5309155355ef6/out/2f1fe1929bb62df3-_openssl.o" "-c" "/home/buildroot/buildroot/output/build/python-cryptography-41.0.1/src/rust/target/aarch64-unknown-linux-gnu/release/build/cryptography-cffi-4ab5309155355ef6/out/_openssl.c" with args "aarch64-linux-gcc" did not execute successfully (status code exit status: 1).


warning: build failed, waiting for other jobs to finish...

    =============================DEBUG ASSISTANCE=============================
    If you are seeing a compilation error please try the following steps to
    successfully install cryptography:
    1) Upgrade to the latest pip and try again. This will fix errors for most
       users. See: https://pip.pypa.io/en/stable/installing/#upgrading-pip
    2) Read https://cryptography.io/en/latest/installation/ for specific
       instructions for your platform.
    3) Check our frequently asked questions for more information:
       https://cryptography.io/en/latest/faq/
    4) Ensure you have a recent Rust toolchain installed:
       https://cryptography.io/en/latest/installation/#rust
    
    Python: 3.11.4
    platform: Linux-5.19.0-45-generic-x86_64-with-glibc2.35
    pip: n/a
    setuptools: 67.8.0
    setuptools_rust: 1.6.0
    rustc: 1.70.0 (90c541806 2023-05-31)
    =============================DEBUG ASSISTANCE=============================
    
error: `cargo rustc --lib --message-format=json-render-diagnostics --manifest-path src/rust/Cargo.toml --target aarch64-unknown-linux-gnu --release -v --features pyo3/extension-module --crate-type cdylib --` failed with code 101
```

jameshilliard:
Fixes #9119.

### DIFF
--- a/src/rust/cryptography-cffi/build.rs
+++ b/src/rust/cryptography-cffi/build.rs
@@ -47,9 +47,14 @@ fn main() {
     )
     .unwrap();
     println!("cargo:rustc-cfg=python_implementation=\"{}\"", python_impl);
-    let python_include = run_python_script(
+    let python_includes = run_python_script(
         &python,
-        "import sysconfig; print(sysconfig.get_path('include'), end='')",
+        "import os; \
+         import setuptools.dist; \
+         import setuptools.command.build_ext; \
+         b = setuptools.command.build_ext.build_ext(setuptools.dist.Distribution()); \
+         b.finalize_options(); \
+         print(os.pathsep.join(b.include_dirs), end='')",
     )
     .unwrap();
     let openssl_include =
@@ -59,11 +64,14 @@ fn main() {
     let mut build = cc::Build::new();
     build
         .file(openssl_c)
-        .include(python_include)
         .include(openssl_include)
         .flag_if_supported("-Wconversion")
         .flag_if_supported("-Wno-error=sign-conversion")
         .flag_if_supported("-Wno-unused-parameter");
+
+    for python_include in env::split_paths(&python_includes) {
+        build.include(python_include);
+    }
 
     // Enable abi3 mode if we're not using PyPy.
     if python_impl != "PyPy" {


alex:
I'm concerned this may have the practical effect of reverting https://github.com/pyca/cryptography/commit/aa8dbd81f0e4b11fb0028fe29c476e721e496483, particularly since we have no ability to test _any_ of these cross-compilation patterns in CI.

Is there any reasonable way to validate this works correctly?

jameshilliard:
> I'm concerned this may have the practical effect of reverting https://github.com/pyca/cryptography/commit/aa8dbd81f0e4b11fb0028fe29c476e721e496483, particularly since we have no ability to test any of these cross-compilation patterns in CI.

I think it's probably not going to cause issues as `headers` is from my understanding only set for a subset of python cross compilation configurations(it doesn't look like yocto sets it).

> Is there any reasonable way to validate this works correctly?

I'm not all that familiar with how yocto handles python cross compilation here.

alex:
Perhaps a dumb question, but I don't see `headers` mentioned in https://docs.python.org/3/library/sysconfig.html, is this an official thing?

jameshilliard:
> is this an official thing

It's set [here](https://github.com/python/cpython/blob/v3.11.4/Lib/sysconfig.py#L238) I think.

alex:
Ah, so it's merely under-documented.

On Fri, Jun 23, 2023 at 6:44 PM James Hilliard ***@***.***> wrote:
>
> is this an official thing
>
> It's set here I think.
>
> —



-- 

alex:
@kanavin can you confirm that this PR doesn't have a negative impact on your cross-compilation?

kanavin:
On our setup 'headers' is not present at all:
```
>>> import sysconfig; print(sysconfig.get_paths())              
{'stdlib': '/usr/lib/python3.11', 'platstdlib': '/usr/lib/python3.11', 'purelib': '/usr/lib/python3.11/site-packages', 'platlib': '/usr/lib/python3.11/site-packages', 'include': '/usr/include/python3.11', 'platinclude': '/usr/include/python3.11', 'scripts': '/usr/bin', 'data': '/usr'}
```

Reading the code for sysconfig, 'headers' entry is only present if python is called from its own source/build directory, and so shouldn't be relied on for building other items, such as this one :) It's not documented for the same reason.

kanavin:
I'd also like to ask @jameshilliard , how do you get the sysroot prefix from the build host into these paths reported by sysconfig? Yocto doesn't do it, assuming sysconfig should be reporting target paths. Python cross-compile is a mess.

kanavin:
I found it:
https://github.com/buildroot/buildroot/blob/master/package/python3/0004-Adjust-library-header-paths-for-cross-compilation.patch
and
https://github.com/buildroot/buildroot/blob/master/package/pkg-python.mk#L45

I don't have a clear idea yet how we can meet in the middle, need to research further.

kanavin:
Okay, sorry for the flood. 

@alex The more I look at it, the more I'm convinced that "sysconfig.get_path ('include')" was correct all along. I'd suggest at this point that my change is reverted as well - both yocto and buildroot need to figure out how to make that work without having to patch components.

jameshilliard:
> Reading the code for sysconfig, 'headers' entry is only present if python is called from its own source/build directory, and so shouldn't be relied on for building other items, such as this one :) It's not documented for the same reason.

Yeah...I think to some degree we're treating our 3rd party python package builds somewhat like python stdlib module builds. I recall that approach was the least broken and that [this patch](https://github.com/buildroot/buildroot/commit/7c1bb5c2bfdc9d726c41a66b30b7aa8e759c6596) was related.

> Python cross-compile is a mess.

Very true, and it's largely undocumented as well with everyone handling things slightly differently, rust is pretty bad at cross compilation as well considering setting env variables with names like [`__CARGO_TEST_CHANNEL_OVERRIDE_DO_NOT_USE_THIS`](https://github.com/buildroot/buildroot/blob/2023.05/package/pkg-cargo.mk#L28-L35) are effectively mandatory to work around cargo linker bugs when cross compiling with stable rust toolchains.

> I don't have a clear idea yet how we can meet in the middle, need to research further.

Well if `headers` isn't normally defined...then this shouldn't cause side effects right?

> @alex The more I look at it, the more I'm convinced that "sysconfig.get_path ('include')" was correct all along. I'd suggest at this point that my change is reverted as well - both yocto and buildroot need to figure out how to make that work without having to patch components such as this one.

Yeah, I agree `sysconfig.get_config_var('INCLUDEPY')` does not really look right to me, I think using the `headers` variable when present is probably harmless as it's not normally defined at all. I'm still somewhat confused exactly why this issue only started popping up with https://github.com/pyca/cryptography/commit/6c39999b40d8794945a8cd47106725334d6d3ae3 is that the same commit that introduced the build issue for yocto as well?

reaperhulk:
It popped up in that commit because we fundamentally changed how we build and link openssl with that work. We needed to do this so we could link rust-openssl against the same copy of openssl that we use for the cffi layer.

kanavin:
I submitted a revert PR to my change. Yocto will carry the fix locally and mark it as unsuitable for upstream submission. (and we need to seriously look into cross-building python in a better way).

I cannot agree that using undocumented API is harmless, 'headers' or otherwise. It's problematic for two reasons:

- anyone reading the code will have a hard time understanding what the undocumented api call does and why is it there. You'd need to research commit history, PR comments, and python code which implements undocumented stuff, and hope you can grasp all of it and make things clear for yourself.

- python upstream can change the behavior behind undocumented items at any point, and then there's suddenly a bug in components using it with no clear resolution, as adapting to new behavior will break installations of python with old behavior.

I'd suggest that this change stays local to buildroot as well, as it is very much specific to that environment (running python out of its build tree) and is unlikely to be useful elsewhere.

jameshilliard:
> It popped up in that commit because we fundamentally changed how we build and link openssl with that work.

Was this include directory not being used anywhere in the build previously? I'm wondering if there's a different way of pulling this directory that wouldn't trigger this issue that is being used in other places.

> anyone reading the code will have a hard time understanding what the undocumented api call does and why is it there. 

Well anything related to python cross compilation tends to be poorly or entirely undocumented in general, I updated the PR to add a comment to indicate how `headers` is used.

> python upstream can change the behavior behind undocumented items at any point, and then there's suddenly a bug in components using it with no clear resolution, as adapting to new behavior will break installations of python with old behavior.

That's not all that clear, it seems to be used by distutils/setuptools based on what I'm seeing [here](https://github.com/pypa/distutils/issues/88) in some situations which would indicate that the behavior should be semi-stable as long as it's used similar to distutils/setuptools.

> I'd suggest that this change stays local to buildroot as well, as it is very much specific to that environment (running python out of its build tree) and is unlikely to be useful elsewhere.

To clarify we're not actually running python out of its build tree for these 3rd party module builds, we're setting the sysconfig environment up similar to how module builds from within the python build tree work from my understanding but still running the host python binary itself from its out of tree location.

alex:
> Was this include directory not being used anywhere in the build previously? I'm wondering if there's a different way of pulling this directory that wouldn't trigger this issue that is being used in other places.

Previously the include directory was being found _somewhere_ inside of setuptools logic. My preference would be to just reproduce whatever it's doing (that's what we thought we were doing with the sysconfig usage, but perhaps there's a subtly we're missing).

jameshilliard:
> Previously the include directory was being found _somewhere_ inside of setuptools logic.

It looks like it might be [here](https://github.com/pypa/setuptools/blob/v68.0.0/setuptools/_distutils/command/build_ext.py#L152-L159), not entirely sure I've traced this properly however.

alex:
Note that the sysconfig there is _not_ the stdlib sysconfig module,
its distutils.sysconfig. This looks plausible though, I think we'd be
happy to take a patch that emulated this.

On Sat, Jun 24, 2023 at 10:10 PM James Hilliard
***@***.***> wrote:
>
> Previously the include directory was being found somewhere inside of setuptools logic.
>
> It looks like it might be here, not entirely sure I've traced this properly however.
>
> —



-- 

jameshilliard:
> I think we'd be happy to take a patch that emulated this.

Seems easier and probably more reliable to just pass the include_dirs via an env variable from setuptools-rust like [this](9132).

alex:
@jameshilliard @kanavin can you both confirm that this latest version works for your setups? Seems like it should, which will be great, and we can put this cursed issue to bed :-)

alex (approved):

jameshilliard:
> can you both confirm that this latest version works for your setups?

Yeah, seems to work for me.

kanavin:
Works for me as well, thanks!

PR #9105, opened 2023-06-21
cryptography-cffi: substitute include path from target sysroot in cross builds

kanavin:
Existing code gets include path on the target device (e.g. where the code will run). For the cross build environments such as yocto, INCLUDEPY contains the correct value.

Fixes #8867.

PR #9131, opened 2023-06-24
Revert "cryptography-cffi: substitute include path from target sysroot in cross builds (#9105)"

kanavin:
The original code was right all along: it uses the official API for obtaining header locations, and it is on the build environment to ensure python supplies that in cross-build scenarios (another option is to apply a custom patch, however any such patch is not eligible for upstream submission due to its specificity).

Further info: https://github.com/pyca/cryptography/pull/9129


alex (approved):
Thanks for continuing to investigate this issue!

PR #9132, opened 2023-06-25
Try to fetch include_dirs from env

jameshilliard:
This depends on [#343](https://github.com/PyO3/setuptools-rust/pull/343).

Fixes: #9119
Alternative to: #9129

### DIFF
--- a/src/rust/cryptography-cffi/build.rs
+++ b/src/rust/cryptography-cffi/build.rs
@@ -47,11 +47,6 @@ fn main() {
     )
     .unwrap();
     println!("cargo:rustc-cfg=python_implementation=\"{}\"", python_impl);
-    let python_include = run_python_script(
-        &python,
-        "import sysconfig; print(sysconfig.get_path('include'), end='')",
-    )
-    .unwrap();
     let openssl_include =
         std::env::var_os("DEP_OPENSSL_INCLUDE").expect("unable to find openssl include path");
     let openssl_c = Path::new(&out_dir).join("_openssl.c");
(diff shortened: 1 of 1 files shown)


kanavin:
Can you give an example of how this is intended to be used?

jameshilliard:
> Can you give an example of how this is intended to be used?

Not sure what you mean, it's designed to work automatically in combination with [this](https://github.com/PyO3/setuptools-rust/pull/343) change to setuptools-rust. Since setuptools-rust already generates the correct `include_paths` we just need to pass them to cryptography using the env essentially.

kanavin:
> > Can you give an example of how this is intended to be used?
> 
> Not sure what you mean, it's designed to work automatically in combination with [this](https://github.com/PyO3/setuptools-rust/pull/343) change to setuptools-rust. Since setuptools-rust already generates the correct `include_paths` we just need to pass them to cryptography using the env essentially.

Thanks, this wasn't clear. Does setuptools-rust obtain them from setuptools?

jameshilliard:
> Does setuptools-rust obtain them from setuptools?

Yeah, which in turn obtains them from distutils from my understanding.

kanavin:
distutils story is confusing. It's been removed from upcoming python 3.12 as something that shouldn't be used, but a copy survives in setuptools. And it has its own sysconfig, separate from one in python core. But I'm fine with that, if it's not being used directly - setuptools can do what it deems right.

jameshilliard:
> But I'm fine with that, if it's not being used directly - setuptools can do what it deems right.

Yeah, the main idea here is that rather than trying to recreate the methods used by setuptools/distutils for getting include paths...we can instead just pass the include paths that are known to work. We should be much less likely to hit edge cases caused by behavior divergences and allow us to not have to do any downstream patching.

alex:
At some point we're going to switch to maturin (#8815), at which point this approach won't work.

jameshilliard:
> At some point we're going to switch to maturin (#8815), at which point this approach won't work.

Yeah, not really sure how maturin handles include paths, [we](https://buildroot.org/) do have some packages using it and haven't seen problems with the include paths so I assume it also has a method for obtaining them correctly as well similar to setuptools-rust.

In any case I kept the existing method as a fallback to prevent this introducing regressions.

alex:
I think we can just use setuptool's public API to extract these things:

```
import setuptools.dist
import setuptools.command.build_ext
b = setuptools.command.build_ext.build_ext(setuptools.dist.Distribution())
b.finalize_options()
print(b.include_dirs)
```

Does this work correctly for your environment? If so, I think this is much simpler than having it wired via setuptools-rust.

jameshilliard:
> Does this work correctly for your environment?

It seems to, although when migrating to maturin I suppose we would still need to depend on setuptools.

alex:
Closing in favor of #9129

PR #9143, opened 2023-06-27
Stop explicitly enabling RSA blinding

alex:
It is on by default in OpenSSL, going back at least as far 1.1.1d, and probably much farther.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/rsa.py
+++ b/src/cryptography/hazmat/backends/openssl/rsa.py
@@ -4,7 +4,6 @@
 
 from __future__ import annotations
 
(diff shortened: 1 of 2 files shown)


reaperhulk:
And this is also true for libre and boring? (Just making sure 😄)

alex:
Fuck. I didn't check those.

alex:
Confirmed that BoringSSL its always on.

alex:
libressl 3.6.3 appears to have the same code structure as openssl, and therefore is on by default.

reaperhulk (approved):
💯

PR #9145, opened 2023-06-27
Try using the rust cache action

alex:


reaperhulk:
Well it looks like it works, but what are we getting/losing here? 😄

alex:
This is in a theoretical sense worse than sccache, but it won't blow the rate limit, so it will actually work.

Notably it doesn't cache local package, so we'll be rebuilding _our_ packages every build, but effectively never rebuilding our dependencies.

Downside is we're back to manual management of cache keys.

reaperhulk:
Love to churn our caching techniques.

alex:
Indeed.

If Github 10x'd their GHA cache rate limit quota, this would not be necessary.

reaperhulk (approved):

PR #9152, opened 2023-06-29
Port RSA to rust

alex:
Has a number of TODOs that need to be dealt with


src/rust/src/backend/ec.rs
```diff
@@ -503,10 +503,6 @@ impl ECPublicKey {
         let mut verifier = openssl::pkey_ctx::PkeyCtx::new(&self.pkey)?;
         verifier.verify_init()?;
         let valid = verifier.verify(data, signature).unwrap_or(false);
-        // TODO: Empty the error stack. BoringSSL leaves one in the event of
-        // signature validation failure. Upstream to rust-openssl?
-        #[cfg(CRYPTOGRAPHY_IS_BORINGSSL)]
```
reaperhulk, line 508: Should this be a separate PR?

alex: Yeah

reaperhulk (approved):

PR #9181, opened 2023-07-05
Convert AESGCM AEAD to Rust

alex:


alex:
Ok, LibreSSL is failing because:

- rust-openssl has an assert, when you set an IV, that the `iv.len() <= EVP_CIPHER_CTX_iv_length()`
- On LibreSSL `EVP_CIPHER_CTX_iv_length` always returns the cipher's length, it does not respect `EVP_CTRL_AEAD_SET_IVLEN`
- The cipher's length is 12.
- Therefore using an IV shorter than 12 bytes will always hit this assert on LibreSSL

Not immediately sure what the right fix it. This feels like a LibreSSL bug, but would take a while to roll out.

alex:
Looks like BoringSSL will have this same problem.

alex:
Ok this is now green, minus LibreSSL 3.7.3, meaning that we're waiting for May 1 when the relevant OpenBSD version is dropped.

reaperhulk (approved):

reaperhulk:
Presented without comment:
image

PR #9193, opened 2023-07-07
update linkcheck

reaperhulk:
skip github links with anchor tags due to their new page rendering and switch to running linkcheck automatically when we touch conf.py or the linkcheck yaml.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -202,6 +202,9 @@ linkcheck_ignore = [
     r"https://www.oscca.gov.cn",
     # Cloudflare returns 403s for all non-browser requests
     r"https://speakerdeck.com",
+    # GitHub changed how they do page renders so anchor detection
+    # no longer works in source view
+    r"https://github.com/.*/blob/.*#L\d+",
 ]
 
 autosectionlabel_prefix_document = True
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -202,6 +202,11 @@
     r"https://www.oscca.gov.cn",
     # Cloudflare returns 403s for all non-browser requests
     r"https://speakerdeck.com",
+    # GitHub changed how they do page renders so anchor detection
+    # no longer works
+    r"https://github.com/openssl/openssl/blob/97cf1f6c2854a3a955fd7dd3a1f113deba00c9ef/crypto/evp/evptests.txt#L232",
+    r"https://github.com/openssl/openssl/blob/5a7bc0be97dee9ac715897fe8180a08e211bc6ea/test/evpciph.txt#L2362",
+    r"https://github.com/ruby/ruby/blob/4ccb387f3bc436a08fc6d72c4931994f5de95110/test/openssl/test_pkey_dsa.rb#L53",
```
alex, line 209: This can be a regex, should we just do `github.com/.*/blob/.*#L\d+` or something?

reaperhulk: We could -- do we want to constrain to blob or just `github.com/.*#L\d+`?

alex: I think just blob? I don't see how other pages would have this pattern

alex (approved):

PR #9194, opened 2023-07-07
Add support for ChaCha20 with LibreSSL

facutuesca:
This PR adds support for ChaCha20 with LibreSSL. Since cryptography's ChaCha20 uses OpenSSL's implementation (which uses a 64 bit counter + 64 bit nonce), and LibreSSL does the same, it is straightforward to use LibreSSL's API.

The only complication is that the current `_CipherContext` implementation assumes that the underlying cipher can be accessed through the `EVP_CIPHER` API. This is not the case for LibreSSL's ChaCha20, which has a separate [CRYPTO_chacha_20()](https://man.openbsd.org/ChaCha.3) API for it.

In order to solve that, this PR makes `_CipherContext` an abstract class with two implementations, `_CipherContextEVP` (which is the previous `_CipherContext`) and `_CipherContextChaCha`, which is a simple context that is used only for the LibreSSL+ChaCha20 combination.

It also leaves the option open in the future if we want to add BoringSSL's ChaCha20, since we can reuse the same `_CipherContextChaCha` for it (the API is almost the same, and we can deal with the differences in a C wrapper)

### DIFF
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -26,6 +26,7 @@ ffi = build_ffi_for_binding(
         "asn1",
         "bignum",
         "bio",
+        "chacha",
         "cmac",
         "crypto",
         "dh",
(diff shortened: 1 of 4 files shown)


facutuesca:
Needs further work

PR #9209, opened 2023-07-10
Add support for ChaCha20 with LibreSSL

facutuesca:
This PR adds support for ChaCha20 with LibreSSL. Since cryptography's ChaCha20 uses OpenSSL's implementation (which uses a 64 bit counter + 64 bit nonce), and LibreSSL does the same, it is straightforward to use LibreSSL's API.

The only complication is that the current `_CipherContext` implementation assumes that the underlying cipher can be accessed through the `EVP_CIPHER` API. This is not the case for LibreSSL's ChaCha20, which has a separate [CRYPTO_chacha_20()](https://man.openbsd.org/ChaCha.3) API for it.

In order to solve that, this PR makes `_CipherContext` an abstract class with two implementations, `_CipherContextEVP` (which is the previous `_CipherContext`) and `_CipherContextChaCha`, which is a simple context that is used only for the LibreSSL+ChaCha20 combination.

It also leaves the option open in the future if we want to add BoringSSL's ChaCha20, since we can reuse the same `_CipherContextChaCha` for it (the API is almost the same, and we can deal with the differences in a C wrapper)

### DIFF
--- a/src/_cffi_src/openssl/chacha.py
+++ b/src/_cffi_src/openssl/chacha.py
@@ -1,0 +1,42 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+from __future__ import annotations
+
+INCLUDES = """
+#if CRYPTOGRAPHY_IS_LIBRESSL
+#include <openssl/chacha.h>
+#endif"""
+
+TYPES = """
+static const long Cryptography_HAS_CHACHA20_API;
+"""
+
+FUNCTIONS = """
+/* Signature is different between LibreSSL and BoringSSL, so expose via
+   different symbol name */
+void Cryptography_CRYPTO_chacha_20(uint8_t *, const uint8_t *, size_t,
+                                   const uint8_t[32], const uint8_t[8],
+                                   uint64_t);
+"""
+
+CUSTOMIZATIONS = """
+#if CRYPTOGRAPHY_IS_LIBRESSL
+static const long Cryptography_HAS_CHACHA20_API = 1;
+#else
+static const long Cryptography_HAS_CHACHA20_API = 0;
+#endif
+
+#if CRYPTOGRAPHY_IS_LIBRESSL
+void Cryptography_CRYPTO_chacha_20(uint8_t *out, const uint8_t *in,
+                                   size_t in_len, const uint8_t key[32],
+                                   const uint8_t nonce[8], uint64_t counter) {
+    CRYPTO_chacha_20(out, in, in_len, key, nonce, counter);
+}
+#else
+void (*Cryptography_CRYPTO_chacha_20)(uint8_t *, const uint8_t *, size_t,
+                                      const uint8_t[32], const uint8_t[8],
+                                      uint64_t) = NULL;
+#endif
+"""
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -12,7 +12,10 @@ import typing
 from cryptography import utils, x509
 from cryptography.exceptions import UnsupportedAlgorithm, _Reasons
 from cryptography.hazmat.backends.openssl import aead
-from cryptography.hazmat.backends.openssl.ciphers import _CipherContext
+from cryptography.hazmat.backends.openssl.ciphers import (
+    _CipherContext,
+    create_cipher_context,
+)
 from cryptography.hazmat.backends.openssl.cmac import _CMACContext
 from cryptography.hazmat.bindings._rust import openssl as rust_openssl
 from cryptography.hazmat.bindings.openssl import binding
@@ -221,6 +224,10 @@ class Backend:
         return self.hash_supported(algorithm)
 
     def cipher_supported(self, cipher: CipherAlgorithm, mode: Mode) -> bool:
+        # ChaCha20 is supported in LibreSSL by a different API than OpenSSL,
+        # so checking for the corresponding EVP_CIPHER is not useful
+        if self._lib.CRYPTOGRAPHY_IS_LIBRESSL and isinstance(cipher, ChaCha20):
+            return True
         if self._fips_enabled:
             # FIPS mode requires AES. TripleDES is disallowed/deprecated in
             # FIPS 140-3.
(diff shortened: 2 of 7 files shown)


src/_cffi_src/openssl/chacha.py
```diff
@@ -0,0 +1,40 @@
...
+#else
+static const long Cryptography_HAS_CHACHA20_API = 0;
+#endif
+
+#if CRYPTOGRAPHY_IS_LIBRESSL
+void Cryptography_CRYPTO_chacha_20(uint8_t *out, const uint8_t *in,
+                                   size_t in_len, const uint8_t key[32],
+                                   const uint8_t nonce[8], uint64_t counter) {
+    CRYPTO_chacha_20(out, in, in_len, key, nonce, counter);
+}
```
alex, line 34: Is the Cryptography prefix name required because OpenSSL has the same symbol with a different signature? If yes, please add a comment.

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -226,6 +229,12 @@ def hmac_supported(self, algorithm: hashes.HashAlgorithm) -> bool:
         return self.hash_supported(algorithm)
 
     def cipher_supported(self, cipher: CipherAlgorithm, mode: Mode) -> bool:
+        # ChaCha20 is supported in LibreSSL by a different API than OpenSSL,
+        # so checking for the corresponding EVP_CIPHER is not useful
+        if (self._lib.CRYPTOGRAPHY_IS_LIBRESSL) and isinstance(
```
alex, line 234:
```suggestion
        if self._lib.CRYPTOGRAPHY_IS_LIBRESSL and isinstance(
```

src/cryptography/hazmat/backends/openssl/ciphers.py
```diff
@@ -14,7 +15,114 @@
     from cryptography.hazmat.backends.openssl.backend import Backend
 
 
-class _CipherContext:
+class _CipherContext(metaclass=abc.ABCMeta):
```
alex, line 18: Why is a new interface required? This seems to match our existing public cipher context ABC.

src/cryptography/hazmat/backends/openssl/ciphers.py
```diff
@@ -14,7 +15,114 @@
...
+        baseoutbuf = self._backend._ffi.from_buffer(buf, require_writable=True)
+        baseinbuf = self._backend._ffi.from_buffer(data)
+
+        self._backend._lib.Cryptography_CRYPTO_chacha_20(
+            baseoutbuf,
+            baseinbuf,
+            total_data_len,
+            self._backend._ffi.from_buffer(self._cipher.key),
+            self._backend._ffi.from_buffer(self._iv_nonce),
+            self._counter,
```
alex, line 107: Hmm, shouldn't counter be advancing across multiple calls?

src/cryptography/hazmat/backends/openssl/ciphers.py
facutuesca: It's an interface for the backend cipher context, which needs a couple of methods not present in the existing public `CipherContext` ABC:

```python
    @abc.abstractmethod
    def authenticate_additional_data(self, data: bytes) -> None:
        ...

    @abc.abstractmethod
    def finalize_with_tag(self, tag: bytes) -> bytes:
        ...
```

I think it also makes sense to have separate interfaces for public vs backend ctx, since the public context wraps the backend one:
```python
    def _wrap_ctx(
        self, ctx: _BackendCipherContext, encrypt: bool
    ) -> typing.Union[
        AEADEncryptionContext, AEADDecryptionContext, CipherContext
    ]:
        if isinstance(self.mode, modes.ModeWithAuthenticationTag):
            if encrypt:
                return _AEADEncryptionContext(ctx)
            else:
                return _AEADDecryptionContext(ctx)
        else:
            return _CipherContext(ctx)
  ```

File: src/cryptography/hazmat/backends/openssl/ciphers.py (hunk 4 above)
Comment (facutuesca) [reply src/cryptography/hazmat/backends/openssl/ciphers.py:133]: Yes, I missed that completely. I have rewritten the `update_into()` function as per our discussion

File: src/cryptography/hazmat/backends/openssl/backend.py (hunk 2 above)
Comment (facutuesca) [reply src/cryptography/hazmat/backends/openssl/backend.py]: Fixed

**Comment by facutuesca** [comment]
The current CI failure is due to not having coverage of https://github.com/pyca/cryptography/pull/9209/files#diff-b81ebda89021aab11aceb5d95cf253b81f9dcdf630468c0483af411cc5005662R166-R167 (the counter rollover during 64-bit overflow). Once the test vectors from [this PR](https://github.com/pyca/cryptography/pull/9221) are merged, the coverage should go to 100%

**Comment by alex** [comment]
Ok. @reaperhulk has the TODO to review that one, once it merges I'll look here.

**Comment by alex** [comment]
I now realize I'm doing this out of order, but did we ever talk to the LibreSSL maintainers about if they'd be interested in an EVP API for this?

**Comment by facutuesca** [comment]
> I now realize I'm doing this out of order, but did we ever talk to the LibreSSL maintainers about if they'd be interested in an EVP API for this?

We haven't, I believe, because the main goal for these PRs was to [improve BoringSSL feature parity](https://github.com/pyca/cryptography/issues/8956), and this LibreSSL implementation was just a bonus due to the LibreSSL API being the same as BoringSSL's.

My idea was to follow up this PR with the BoringSSL implementation (which is trivial once this LibreSSL PR is merged).

**Comment by woodruffw** [comment]
>  I don't know if we're currently interested in implementing the EVP API for ChaCha20 in LibreSSL

I believe that _doing_ that implementation in LibreSSL would be out of scope for us. That being said, we can certainly raise it upstream and see if they'd be interested in exposing ChaCha20 via `EVP_CIPHER`.

Would you prefer we do that before moving forwards here @alex?

**Comment by botovq** [comment]
I must be missing something. EVP_chacha20 has existed in LibreSSL since forever. We adjusted it to be compatible with OpenSSL for OpenSSH in early 2020, even if that is wrong per spec.

**Comment by facutuesca** [comment]
> I must be missing something. [EVP_chacha20](https://man.openbsd.org/EVP_chacha20) has existed in LibreSSL since forever (it was committed on May 1, 2014, not even a month into the fork's existence).

Thanks for the clarification, I now see why I thought the EVP API was not available in LibreSSL: Turns out that `cryptography` uses `EVP_get_cipherbyname()` to get the EVP cipher using the corresponding name. When I tried this with LibreSSL, it seemed as if the cipher was not available.

However, the error was because the name for ChaCha20 in OpenSSL is different from LibreSSL. OpenSSL [uses](https://github.com/libressl/openbsd/blob/c58d283fbdca7a7ec041efd5bd41531edab6610c/src/lib/libcrypto/objects/objects.txt#L1428-L1430
) `chacha20` and `ChaCha20`:
```c
#define SN_chacha20             "ChaCha20"
#define LN_chacha20             "chacha20"
#define NID_chacha20            1019
```

Whereas LibresSSL [uses](https://github.com/openssl/openssl/blob/master/include/openssl/obj_mac.h#L5289-L5291
) `chacha` and `ChaCha`:
```
# ChaCha Stream Cipher
!Cname chacha20
			: ChaCha		: chacha
```

I'll close this PR and open a new one that conditionally changes the name according to the backend used.
@botovq Do you know why the names don't match?

cc @woodruffw

botovq:
I see.

> @botovq Do you know why the names don't match?

There will be no deeper reason than that it was added one way to
LibreSSL and the other way to OpenSSL and no one noticed in the
following 9 years...

I guess there is no harm in adding an alias on our side to make
OpenSSL's naming choice work, but that won't be available before 3.9.0

facutuesca:
Closing in favor of https://github.com/pyca/cryptography/pull/9758

PR #9211, opened 2023-07-10
Backport ssh cert fix

reaperhulk:
for 41.0.x we want to allow the invalid encoding to be parsed


tests/hazmat/primitives/test_ssh.py
```diff
@@ -1159,6 +1159,33 @@ def test_loads_ssh_cert(self, backend):
...
+            )
+        assert isinstance(cert, SSHCertificate)
+        cert.verify_cert_signature()
+        assert list(cert.extensions.items()) == [
+            (b"permit-pty", b""),
+        ]
+        assert list(cert.critical_options.items()) == [
+            (b"force-command", b"echo aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"),
+            (b"verify-required", b""),
+        ]
```
alex, line 1187: Is there a reason these aren't just assertions about the dict, and instead convert ot a list?

CHANGELOG.rst
```diff
@@ -1,6 +1,16 @@
...
 
+.. _v41-0-2:
+
+41.0.2 - 2023-07-10
+~~~~~~~~~~~~~~~~~~~
+
+* Fix SSH certificate encoding for critical options with values to be
+  compatible with OpenSSH. Previously generated invalid encodings will parse
+  successfully with a warning, but this invalid encoding support will be
+  removed in the next release.
```
alex, line 12: I think this reads confusingly, suggested alternative:

> * Fixed bugs in creating and parsing SSH certificates, where critical options with values were handled incorrectly. With this release, certificates are now created correctly and parsing accepts *both* correct values, but also allows the previously generated invalid forms with a warning. In the next release, support for parsing these invalid forms will be removed.

tests/hazmat/primitives/test_ssh.py
reaperhulk: Tests order preservation, but we don't need to test that here obviously 😄

alex (approved):

PR #9212, opened 2023-07-10
oid: add more extension, EKU OIDs

woodruffw:
This adds the `subjectDirectoryAttributes` extension OID, as well as the RFC 5280 EKU OIDs.

Breakout from #8873 (slightly expounded with OIDs I expect we'll need).


reaperhulk:
So this is the beginning of the "break #8873 apart so it's reviewable" PRs? But you've added a few you speculatively expect to need?

woodruffw:
Yep and yep -- it includes two OIDs that I know I need, and then a few more of the EKUs since I expect I'll need them (for configurable verification policies).

reaperhulk (approved):

PR #9213, opened 2023-07-10
x509: more extension APIs

woodruffw:
More breakout from #8873:

* Brings in extension types/aliases from that PR
* Saves `N` clones in the `RawExtensions` to `Extensions` APIs
* Adds `Extension::value` for extracting inner extension types

### DIFF
--- a/src/rust/cryptography-x509/src/extensions.rs
+++ b/src/rust/cryptography-x509/src/extensions.rs
@@ -236,7 +236,6 @@ pub struct BasicConstraints {
 pub type SubjectAlternativeName<'a> = asn1::SequenceOf<'a, name::GeneralName<'a>>;
 pub type ExtendedKeyUsage<'a> = asn1::SequenceOf<'a, asn1::ObjectIdentifier>;
 
-#[derive(Hash, PartialEq, Clone)]
 pub struct KeyUsage<'a>(asn1::BitString<'a>);
 
 impl<'a> asn1::SimpleAsn1Readable<'a> for KeyUsage<'a> {
--- a/src/rust/cryptography-x509/src/extensions.rs
+++ b/src/rust/cryptography-x509/src/extensions.rs
@@ -73,6 +73,12 @@ pub struct Extension<'a> {
     pub extn_value: &'a [u8],
 }
 
+impl<'a> Extension<'a> {
+    pub fn value<T: asn1::Asn1Readable<'a>>(&'a self) -> asn1::ParseResult<T> {
+        Ok(asn1::parse_single(self.extn_value)?)
+    }
+}
+
 #[derive(asn1::Asn1Read, asn1::Asn1Write)]
 pub struct PolicyConstraints {
     #[implicit(0)]
--- a/src/rust/cryptography-x509/src/extensions.rs
+++ b/src/rust/cryptography-x509/src/extensions.rs
@@ -335,9 +335,9 @@ mod tests {
         let extensions = SequenceOfWriter::new(vec![extension]);
 
         let der = asn1::write_single(&extensions).unwrap();
+        let parsed = asn1::parse_single(&der).unwrap();
 
-        let extensions: Extensions =
-            Extensions::from_raw_extensions(Some(&asn1::parse_single(&der).unwrap())).unwrap();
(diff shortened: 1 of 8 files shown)


woodruffw:
CI flake: https://github.com/pyca/cryptography/actions/runs/5512929463/jobs/10050408655?pr=9213

src/rust/cryptography-x509/src/extensions.rs
```diff
@@ -18,15 +18,15 @@ pub type RawExtensions<'a> = common::Asn1ReadableOrWritable<
 ///
 /// In particular, an `Extensions` cannot be constructed from a `RawExtensions`
 /// that contains duplicated extensions (by OID).
-pub struct Extensions<'a>(Option<RawExtensions<'a>>);
+pub struct Extensions<'a, 'b>(Option<&'b RawExtensions<'a>>);
```
alex, line 21: This is adding an extra indirection. The `clones` are basically free, really just moving a few pointers (they are notably _not_ memory copies anywhere). If the motivation for this is just to remove `clone`, we should remove unless there's a benchmark showing value.

src/rust/cryptography-x509/src/extensions.rs
```diff
@@ -73,6 +72,12 @@ pub struct Extension<'a> {
     pub extn_value: &'a [u8],
 }
 
+impl<'a> Extension<'a> {
+    pub fn value<T: asn1::Asn1Readable<'a>>(&'a self) -> asn1::ParseResult<T> {
+        asn1::parse_single(self.extn_value)
+    }
+}
```
alex, line 79: This probably makes this PR a fair bit larger, but I think we have a bunch of use cases for this in cryptography already, notably: `src/rust/src/x509/common.rs`, line 412. we should probably change that to pass an Extension value to the callback, and then use `value()` there.

src/rust/cryptography-x509/src/extensions.rs
```diff
@@ -228,31 +233,84 @@ pub struct BasicConstraints {
     pub path_length: Option<u64>,
 }
 
+pub type SubjectAlternativeName<'a> = asn1::SequenceOf<'a, name::GeneralName<'a>>;
+pub type ExtendedKeyUsage<'a> = asn1::SequenceOf<'a, asn1::ObjectIdentifier>;
+
+pub struct KeyUsage<'a>(asn1::BitString<'a>);
```
alex, line 239: Similarly, there are certainly places in cryptography that should use this.

src/rust/cryptography-x509/src/extensions.rs
```diff
@@ -228,31 +233,84 @@ pub struct BasicConstraints {
...
+impl<'a> asn1::SimpleAsn1Readable<'a> for KeyUsage<'a> {
+    const TAG: asn1::Tag = asn1::BitString::TAG;
+
+    fn parse_data(data: &'a [u8]) -> asn1::ParseResult<Self> {
+        asn1::BitString::parse_data(data).map(Self)
+    }
+}
+
+impl KeyUsage<'_> {
+    pub fn zeroed(&self) -> bool {
```
alex, line 250: `is_zeroed`?

src/rust/cryptography-x509/src/extensions.rs
woodruffw: Yeah, I have no numbers showing value here. I did this thinking that `RawExtensions` was a `Vec` under the hood and so avoiding a clone on it would be nice, but now I see that it's a `SequenceOf` so very little is gained by saving a clone here :slightly_smiling_face:

src/rust/cryptography-x509/src/extensions.rs
woodruffw: [f4a5664](https://github.com/pyca/cryptography/pull/9213/commits/f4a56648da2318db9f85564df05179bff0816ab7) should cover this.

src/rust/cryptography-x509/src/extensions.rs
woodruffw: I only found one deserialization site in Rust for this; included in https://github.com/pyca/cryptography/pull/9213/commits/f4a56648da2318db9f85564df05179bff0816ab7.

alex (approved):

PR #9254, opened 2023-07-17
Migrate more types

woodruffw:
This moves additional types from `cryptography_x509_validation`. In particular:

1. `IA5String` is an invariant-preserving `String` wrapper (the inner string must be ASCII only).
2. `DNSName` is an invariant preserving `IA5String` wrapper (the inner value must look like a domain name, and is normalized as lowercase before being stored).
3. `DNSPattern` is an invariant preserving `DNSName` wrapper (giving us the ability to both exact and RFC 6125-style wildcard matches).

These types roughly mirror some of the Python APIs (service IDs, `CertificatePattern`) in functionality; my thinking here is that those APIs could be replaced with bindings to these once they're rounded out more (e.g. IP address and other types) if desired.

See #8873.


src/rust/cryptography-x509/src/common.rs
```diff
@@ -301,9 +301,72 @@ impl<'a> asn1::SimpleAsn1Writable for UnvalidatedVisibleString<'a> {
...
+
+        match (hostname.split_once('.'), pattern.split_once('.')) {
+            // If both hostname and pattern contain multiple labels, then
+            // we attempt to match using a subset of RFC 6125 6.4.3.
+            // In particular, we don't attempt to support anything
+            // except left-most wildcards.
+            (Some((subdomain, parent)), Some((pat_subdomain, pat_parent))) => {
+                // TODO: This is almost certainly insufficient: we also need
+                // to check for nonsense patterns like `*.`.
+                (pat_subdomain == "*" || pat_subdomain == subdomain) && pat_parent == parent
```
woodruffw, line 356: IMO it makes sense to round this out before merging, but I opened this PR with the TODO still in place in case @alex or @reaperhulk would prefer to merge an MVP first.

src/rust/cryptography-x509/src/common.rs
```diff
@@ -301,9 +301,72 @@ impl<'a> asn1::SimpleAsn1Writable for UnvalidatedVisibleString<'a> {
...
+pub struct DNSString(IA5String);
+
+impl DNSString {
+    pub fn new(value: &str) -> Option<Self> {
+        if value.is_empty() || value == " " {
+            None
+        } else {
+            IA5String::new(value.to_lowercase()).map(Self)
+        }
+    }
```
woodruffw, line 338: Question for @reaperhulk and @alex: how strong would you like invariant preservation to be here? Right now we're just checking for ASCII-ness and a few things that 5280 explicitly disallows, but we could make this into a full DNS label/domain validation check instead (which would be similar to what OpenSSL does.)

src/rust/cryptography-x509/src/name.rs
```diff
@@ -4,9 +4,11 @@
 
 use crate::common;
 
+pub type NameReadable<'a> = asn1::SequenceOf<'a, asn1::SetOf<'a, common::AttributeTypeValue<'a>>>;
```
woodruffw, line 7: NB: I broke this out of the below because it's convenient to have as an alias in some of the validation APIs (e.g., APIs for retrieving subjects and issuers).

src/rust/cryptography-x509/src/common.rs
```diff
@@ -301,9 +301,120 @@ impl<'a> asn1::SimpleAsn1Writable for UnvalidatedVisibleString<'a> {
...
+            Some(Self(value))
+        } else {
+            None
+        }
+    }
+
+    pub(crate) fn as_str(&self) -> &str {
+        &self.0
+    }
+}
```
alex, line 321: This is really the same as `asn1::IA5String`, except that one is not owned. Is there a reason we need an owned type here? I don't see where in teh API we've discussed it'd be required.

src/rust/cryptography-x509/src/common.rs
```diff
@@ -301,9 +301,120 @@ impl<'a> asn1::SimpleAsn1Writable for UnvalidatedVisibleString<'a> {
...
+
+impl DNSPattern {
+    pub fn new(pat: &str) -> Option<Self> {
+        if let Some(pat) = pat.strip_prefix("*.") {
+            DNSName::new(pat).map(Self::Wildcard)
+        } else {
+            DNSName::new(pat).map(Self::Exact)
+        }
+    }
```
alex, line 399: It should probably be an error to have a wildcard in any other position, right?

src/rust/cryptography-x509/src/name.rs
alex: FWIW, this would be trivial to merge if it was in its own PR.

src/rust/cryptography-x509/src/common.rs
woodruffw: Yep, and it is -- `DNSName::new(...)` will return `None` if there are any wildcards present (since they aren't valid domain characters).

e.g., in tests below:

```
        assert_eq!(DNSPattern::new("f*o.example.com"), None);
        assert_eq!(DNSPattern::new("*oo.example.com"), None);
        assert_eq!(DNSPattern::new("fo*.example.com"), None);
        assert_eq!(DNSPattern::new("foo.*.example.com"), None);
        assert_eq!(DNSPattern::new("*.foo.*.example.com"), None);
```

src/rust/cryptography-x509/src/name.rs
woodruffw: #9282

src/rust/cryptography-x509/src/common.rs
woodruffw: Actually, thinking about it more, it really should be case-preserving and the comparison itself should be case insensitive. Doing that now.

woodruffw: Done.

alex:
CI failing here, FYI

src/rust/cryptography-x509/src/common.rs
```diff
@@ -301,9 +301,117 @@ impl<'a> asn1::SimpleAsn1Writable for UnvalidatedVisibleString<'a> {
...
+/// assert_eq!(DNSName::new("foo.com").unwrap(), DNSName::new("FOO.com").unwrap());
+/// ```
+#[derive(Debug)]
+pub struct DNSName<'a>(asn1::IA5String<'a>);
+
+impl<'a> DNSName<'a> {
+    pub fn new(value: &'a str) -> Option<Self> {
+        // Domains cannot be empty; cannot contain whitespace; must (practically)
+        // be less than 253 characters (255 in RFC 1034's octet encoding).
+        if value.is_empty() || value.chars().any(char::is_whitespace) || value.len() > 253 {
```
alex, line 326: Is the whitespace check not covered by the loop below? (Trying to avoid iterating the string 2x)

src/rust/cryptography-x509/src/common.rs
```diff
@@ -330,4 +438,135 @@ mod tests {
...
+    fn test_dnsname_debug_trait() {
+        // Just to get coverage on the `Debug` derive.
+        assert_eq!(
+            "DNSName(IA5String(\"example.com\"))",
+            format!("{:?}", DNSName::new("example.com").unwrap())
+        );
+    }
+
+    #[test]
+    fn test_dnsname_constructs() {
```
alex, line 452:
```suggestion
    fn test_dnsname_new() {
```

alex:
Oh. I think maybe this code should live in the new x509-validation crate?

woodruffw:
> Oh. I think maybe this code should live in the new x509-validation crate?

Yeah, I think you're right. I can turn this PR into a breakout of a skeleton for the crate + this changeset. Or I can do the skeleton separately, if you'd like to keep things small.

src/rust/cryptography-x509/src/common.rs
woodruffw: Yep, it is -- thanks for catching that!

woodruffw, line 326:
```suggestion
        if value.is_empty() || value.len() > 253 {
```

woodruffw:
[aa78be8](https://github.com/pyca/cryptography/pull/9254/commits/aa78be815fdafd67bd1d2b894ec9876ffc83c018) adds the `cryptography_x509_validation` crate skeleton and moves these back to `types.rs`.

alex (approved):

PR #9299, opened 2023-07-26
validation: profile trait, error types

woodruffw:
More breakout from #8873.


alex:
I'm not sure it makes much sense to review this in isolation, its pretty intertwined with the rest of the validator design.

woodruffw:
Yeah, I think you're right. Closing; I'll open a separate PR for some other things that I think can be broken out (specifically, the `CertificateExt` functionality that can be brought upstream).

PR #9300, opened 2023-07-26
Certificate: useful APIs

woodruffw:
More breakout from #8873.

### DIFF
--- a/src/rust/cryptography-x509/src/certificate.rs
+++ b/src/rust/cryptography-x509/src/certificate.rs
@@ -27,12 +27,6 @@ impl Certificate<'_> {
         self.tbs_cert.subject.unwrap_read()
     }
 
-    /// Returns whether the certificate is "self-issued", whether its
-    /// issuer and subject are the same.
-    pub fn is_self_issued(&self) -> bool {
-        self.issuer() == self.subject()
-    }
-
     /// Returns an iterable container over the certificate's extension, or
     /// an error if the extension set contains a duplicate extension.
     pub fn extensions(&self) -> Result<Extensions<'_>, DuplicateExtensionsError> {
--- a/src/rust/cryptography-x509/src/certificate.rs
+++ b/src/rust/cryptography-x509/src/certificate.rs
@@ -9,18 +9,6 @@ use crate::extensions::ExtensionsError;
 use crate::name;
 use crate::name::NameReadable;
 
-#[derive(Debug, PartialEq)]
-pub enum CertificateError {
-    DuplicateExtension(asn1::ObjectIdentifier),
-    Malformed(asn1::ParseError),
-}
-
-impl From<asn1::ParseError> for CertificateError {
-    fn from(value: asn1::ParseError) -> Self {
-        CertificateError::Malformed(value)
-    }
-}
-
 #[derive(asn1::Asn1Read, asn1::Asn1Write, Hash, PartialEq, Eq, Clone)]
 pub struct Certificate<'a> {
     pub tbs_cert: TbsCertificate<'a>,
--- a/src/rust/cryptography-x509/src/certificate.rs
+++ b/src/rust/cryptography-x509/src/certificate.rs
@@ -5,6 +5,7 @@
 use crate::common;
 use crate::extensions;
 use crate::extensions::Extensions;
+use crate::extensions::ExtensionsError;
 use crate::name;
 use crate::name::NameReadable;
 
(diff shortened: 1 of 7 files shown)


src/rust/cryptography-x509/src/certificate.rs
```diff
@@ -14,6 +27,52 @@ pub struct Certificate<'a> {
...
+    /// an error if the extension set contains a duplicate extension.
+    pub fn extensions(&self) -> Result<Extensions<'_>, CertificateError> {
+        self.tbs_cert
+            .extensions()
+            .map_err(CertificateError::DuplicateExtension)
+    }
+
+    /// Returns whether the given extension (by OID) is critical, or
+    /// false if the extension is not present.
+    pub fn extension_is_critical(&self, oid: &asn1::ObjectIdentifier) -> bool {
```
woodruffw, line 57: This feels pretty ugly to me, but it avoids a bit of combinator juggling in the validator APIs, e.g.:

```rust
        if cert.extension_is_critical(&SUBJECT_DIRECTORY_ATTRIBUTES_OID) {
            return Err("SubjectDirectoryAttributes must not be marked critical".into());
        }
```

...rather than having to poke through `extension()?` and then handling the `None` case.

alex: To me this seems too specific, and will lead to way too much iteration over extensions. The right way to do this API is to get the extension, check its criticality,then parse it.

src/rust/cryptography-x509/src/certificate.rs
```diff
@@ -14,6 +27,52 @@ pub struct Certificate<'a> {
...
+        }
+    }
+
+    /// Returns a specific extension by OID.
+    pub fn extension(
+        &self,
+        oid: &asn1::ObjectIdentifier,
+    ) -> Result<Option<Extension<'_>>, CertificateError> {
+        Ok(self.extensions()?.get_extension(oid))
+    }
```
alex, line 73: I don't think we want this -- it incurs re-validation of extensions every time.The point of the extensions object is to avoid revalidating things.

src/rust/cryptography-x509/src/certificate.rs
```diff
@@ -14,6 +27,52 @@ pub struct Certificate<'a> {
...
+    pub fn is_self_issued(&self) -> bool {
+        self.issuer() == self.subject()
+    }
+
+    /// Returns an iterable container over the certificate's extension, or
+    /// an error if the extension set contains a duplicate extension.
+    pub fn extensions(&self) -> Result<Extensions<'_>, CertificateError> {
+        self.tbs_cert
+            .extensions()
+            .map_err(CertificateError::DuplicateExtension)
```
alex, line 52: If we want a real error type, it should be propagated into the TBS extensions method.

src/rust/cryptography-x509/src/certificate.rs
woodruffw: Makes sense -- I can't think of a nice way to provide this API without either re-iterating or doing something else we probably don't want (like separately caching the extensions here), so I'll drop it.

src/rust/src/x509/certificate.rs
```diff
@@ -185,10 +185,10 @@ impl Certificate {
                 let result = asn1::write_single(&tbs_precert)?;
                 Ok(pyo3::types::PyBytes::new(py, &result))
             }
-            Err(oid) => {
-                let oid_obj = oid_to_py_oid(py, &oid)?;
+            Err(err) => {
```
alex, line 188: Suggest unpacking the type, rather than `.0`

src/rust/src/x509/common.rs
```diff
@@ -395,10 +395,10 @@ pub(crate) fn parse_and_cache_extensions<
 
     let extensions = match Extensions::from_raw_extensions(raw_extensions.as_ref()) {
         Ok(extensions) => extensions,
-        Err(oid) => {
-            let oid_obj = oid_to_py_oid(py, &oid)?;
+        Err(err) => {
```
alex, line 398: same

woodruffw:
Got rid of the `CertificateError` entirely and replaced it with an `ExtensionsError` everywhere -- that completely unifies the error types here and keeps us from breaking the Python bindings by introducing another variant that needs to be handled.

alex:
Should probably be named `DuplicateExtensionError`, right÷

src/rust/src/x509/certificate.rs
alex, line 188:
```suggestion
            Err(DuplicateExtensionsError(err)) => {
```

alex (approved):

woodruffw:
(Still needs coverage, doing that in a moment.)

alex (dismissed):

alex (approved):

alex (approved):

alex (approved):

PR #9303, opened 2023-07-27
Add `SSL_OP_LEGACY_SERVER_CONNECT` binding

mhils:
This is useful to expose in pyOpenSSL so that it can be referenced downstream for `Context.set_options`. (https://github.com/mitmproxy/mitmproxy/pull/6281)


mhils:
See previous discussion in https://github.com/pyca/pyopenssl/pull/1193 which I completely forgot about, sorry! :)

alex (approved):

PR #9346, opened 2023-08-02
x509: construct `IPAddress` and `IPRange` types

tnytown:
This contains two more types for `cryptography_x509_validation`:

1. `IPAddress` is a thin wrapper around `std::net::IpAddr` with helpers for masking and conversion to prefixes.
2. `IPRange` represents a CIDR range. Its `from_bytes` can parse IPs from [RFC 5280's `iPAddress` fields].

[RFC 5280's `iPAddress` fields]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.10

### DIFF
--- a/src/rust/cryptography-x509-validation/src/types.rs
+++ b/src/rust/cryptography-x509-validation/src/types.rs
@@ -453,12 +453,14 @@ mod tests {
             \x00\x00\x00\x00\x00\x00\x00\x01\
             \xf0\x00\x00\x00\x00\x00\x00\x00\
             \x00\x00\x00\x00\x00\x00\x00\x00";
+        let bad = b"\xff\xff\xff";
 
         assert_eq!(IPRange::from_bytes(ipv4_bad), None);
         assert_eq!(IPRange::from_bytes(ipv4_bad_many_bits), None);
         assert_eq!(IPRange::from_bytes(ipv4_bad_octet), None);
         assert_eq!(IPRange::from_bytes(ipv6_bad), None);
         assert_ne!(IPRange::from_bytes(ipv6_good), None);
+        assert_eq!(IPRange::from_bytes(bad), None);
 
         // 192.168.1.1/16
         let ipv4_with_extra = b"\xc0\xa8\x01\x01\xff\xff\x00\x00";
--- a/src/rust/cryptography-x509-validation/src/types.rs
+++ b/src/rust/cryptography-x509-validation/src/types.rs
@@ -2,6 +2,8 @@
 // 2.0, and the BSD License. See the LICENSE file in the root of this repository
 // for complete details.
 
+use std::str::FromStr;
+
 /// A `DNSName` is an `asn1::IA5String` with additional invariant preservations
 /// per [RFC 5280 4.2.1.6], which in turn uses the preferred name syntax defined
 /// in [RFC 1034 3.5] and amended in [RFC 1123 2.1].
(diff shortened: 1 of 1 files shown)


src/rust/cryptography-x509-validation/src/types.rs
```diff
@@ -110,9 +112,136 @@ impl<'a> DNSPattern<'a> {
...
+    pub fn from_std(addr: std::net::IpAddr) -> Self {
+        match addr {
+            std::net::IpAddr::V4(a) => Self::V4(a),
+            std::net::IpAddr::V6(a) => Self::V6(a),
+        }
+    }
+
+    pub fn from_str(s: &str) -> Option<Self> {
+        std::net::IpAddr::from_str(s).ok().map(Self::from_std)
+    }
```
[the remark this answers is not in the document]
woodruffw: I'm okay with keeping `from_bytes` for now, per your point about `FromStr` probably being wrong + no corresponding core trait anyays.

src/rust/cryptography-x509-validation/src/types.rs
```diff
@@ -110,9 +112,144 @@ impl<'a> DNSPattern<'a> {
...
+impl IPAddress {
+    pub fn from_str(s: &str) -> Option<Self> {
+        std::net::IpAddr::from_str(s).ok().map(Self::from)
+    }
+
+    /// Constructs an `IPAddress` from a slice. The provided data must be
+    /// 4 (IPv4) or 16 (IPv6) bytes in "network byte order", as specified by
+    /// [RFC 5280].
+    ///
+    /// [https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6]
```
woodruffw, line 133: Link syntax here:

```suggestion
    /// [RFC 5280]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6
```

src/rust/cryptography-x509-validation/src/types.rs
```diff
@@ -110,9 +112,144 @@ impl<'a> DNSPattern<'a> {
...
+    pub fn mask(&self, prefix: u8) -> Self {
+        match self {
+            Self::V4(a) => {
+                let prefix = 32u8.checked_sub(prefix).unwrap_or(0).into();
+                let masked = u32::from_be_bytes(a.octets())
+                    & u32::MAX
+                        .checked_shr(prefix)
+                        .unwrap_or(0)
+                        .checked_shl(prefix)
+                        .unwrap_or(0);
```
alex, line 186: This is the same as `u32::MAX - (1 << prefix - 1)`, more or less, right? Just trying to make sure I understand the logic.

src/rust/cryptography-x509-validation/src/types.rs
```diff
@@ -110,9 +112,144 @@ impl<'a> DNSPattern<'a> {
     }
 }
 
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum IPAddress {
+    V4(std::net::Ipv4Addr),
+    V6(std::net::Ipv6Addr),
```
alex, line 118: Is there an advantage or disadvantage to being an enum, rather than wrapping `IpAddr` which is already an enum?

src/rust/cryptography-x509-validation/src/types.rs
```diff
@@ -110,9 +112,144 @@ impl<'a> DNSPattern<'a> {
...
+    address: IPAddress,
+    prefix: u8,
+}
+
+/// An `IPRange` represents a CIDR-style address range used in a name constraints
+/// extension, as defined by [RFC 5280 4.2.1.10].
+///
+/// [RFC 5280 4.2.1.10]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.10
+impl IPRange {
+    pub fn from_bytes(b: &[u8]) -> Option<Self> {
```
alex, line 223: I think this could use a comment explaining what format the bytes are in.

alex (commented):
Going to need to muse on this, I can't tell if it's elegant or too clever, or maybe it's just too late for IP address math.

src/rust/cryptography-x509-validation/src/types.rs
woodruffw: None that I can think of.

woodruffw: I probably suggested it, thinking that `FromStr` wasn't implemented on `IpAddr`. But I was wrong :slightly_smiling_face:

src/rust/cryptography-x509-validation/src/types.rs
alex: And the checks are basically handling the case where prefix is 32, right?

alex (approved):

PR #9392, opened 2023-08-10
Add `poly1305` implementation for BoringSSL and LibreSSL

facutuesca:
Currently, we support `poly1305` only with OpenSSL. This PR adds support for Boring and Libre.

The current Rust implementation uses the OpenSSL API `EVP_DigestUpdate`. Since Boring and Libre implement a [different API](https://github.com/google/boringssl/blob/master/include/openssl/poly1305.h) for `poly1305`, this PR changes the implementation so that `Poly1305` is now a struct that has two possible backends:

```rust
#[pyo3::prelude::pyclass(module = "cryptography.hazmat.bindings._rust.openssl.poly1305")]
struct Poly1305 {
    #[cfg(any(CRYPTOGRAPHY_IS_BORINGSSL, CRYPTOGRAPHY_IS_LIBRESSL))]
    backend: Poly1305Boring,
    #[cfg(not(any(CRYPTOGRAPHY_IS_LIBRESSL, CRYPTOGRAPHY_IS_BORINGSSL)))]
    backend: Poly1305Open,
}
```

The `Poly1305Open` backend is the same as the current `Poly1305` implementation, whereas the `Poly1305Boring` backend is the new implementation that uses the Libre/Boring Rust bindings exposed through [`openssl-sys 0.9.91`](https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/CHANGELOG.md#v0991---2023-08-06)

This fixes https://github.com/pyca/cryptography/issues/8959

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -1256,7 +1256,13 @@ class Backend:
     def poly1305_supported(self) -> bool:
         if self._fips_enabled:
             return False
-        return self._lib.Cryptography_HAS_POLY1305 == 1
+        elif (
+            self._lib.CRYPTOGRAPHY_IS_BORINGSSL
+            or self._lib.CRYPTOGRAPHY_IS_LIBRESSL
+        ):
+            return True
+        else:
+            return self._lib.Cryptography_HAS_POLY1305 == 1
 
     def pkcs7_supported(self) -> bool:
         return not self._lib.CRYPTOGRAPHY_IS_BORINGSSL
--- a/src/rust/cryptography-openssl/src/poly1305.rs
+++ b/src/rust/cryptography-openssl/src/poly1305.rs
@@ -1,0 +1,40 @@
+// This file is dual licensed under the terms of the Apache License, Version
+// 2.0, and the BSD License. See the LICENSE file in the root of this repository
+// for complete details.
+
+use std::mem::MaybeUninit;
+
+pub struct Poly1305State {
+    // The state data must be allocated in the heap so that its address does not change. This is
+    // because BoringSSL APIs that take a `poly1305_state*` ignore all the data before an aligned
+    // address. Since a stack-allocated struct would change address on every copy, BoringSSL would
+    // interpret each copy differently, causing unexpected behavior.
+    context: Box<ffi::poly1305_state>,
+}
+
+impl Poly1305State {
+    pub fn new(key: &[u8]) -> Poly1305State {
+        #[cfg(CRYPTOGRAPHY_IS_BORINGSSL)]
+        let mut ctx: Box<ffi::poly1305_state> = Box::new([0; 512usize]);
+        #[cfg(CRYPTOGRAPHY_IS_LIBRESSL)]
+        let mut ctx: Box<ffi::poly1305_state> = Box::new(ffi::poly1305_state {
+            aligner: 0,
+            opaque: [0; 136usize],
+        });
+
+        unsafe {
+            ffi::CRYPTO_poly1305_init(ctx.as_mut(), key.as_ptr());
+        }
+        Poly1305State { context: ctx }
+    }
+
+    pub fn update(&mut self, data: &[u8]) -> () {
+        unsafe {
+            ffi::CRYPTO_poly1305_update(self.context.as_mut(), data.as_ptr(), data.len());
+        };
+    }
+
+    pub fn finalize(&mut self, output: &mut [u8]) -> () {
+        unsafe { ffi::CRYPTO_poly1305_finish(self.context.as_mut(), output.as_mut_ptr()) };
+    }
+}
--- a/src/rust/cryptography-openssl/src/poly1305.rs
+++ b/src/rust/cryptography-openssl/src/poly1305.rs
@@ -14,18 +14,22 @@ pub struct Poly1305State {
 
 impl Poly1305State {
     pub fn new(key: &[u8]) -> Poly1305State {
-        #[cfg(CRYPTOGRAPHY_IS_BORINGSSL)]
-        let mut ctx: Box<ffi::poly1305_state> = Box::new([0; 512usize]);
-        #[cfg(CRYPTOGRAPHY_IS_LIBRESSL)]
-        let mut ctx: Box<ffi::poly1305_state> = Box::new(ffi::poly1305_state {
-            aligner: 0,
-            opaque: [0; 136usize],
-        });
+        assert_eq!(key.len(), 32);
+        let mut ctx: Box<MaybeUninit<ffi::poly1305_state>> =
+            Box::new(MaybeUninit::<ffi::poly1305_state>::uninit());
 
-        unsafe {
-            ffi::CRYPTO_poly1305_init(ctx.as_mut(), key.as_ptr());
+        // After initializing the context, unwrap the Box<MaybeUninit<poly1305_state>> into
+        // a Box<poly1305_state> while keeping the same memory address. See the docstring of the
+        // Poly1305State struct above for the rationale.
+        let initialized_ctx: Box<ffi::poly1305_state> = unsafe {
+            ffi::CRYPTO_poly1305_init(ctx.as_mut().as_mut_ptr(), key.as_ptr());
+            let raw_ctx_ptr = (*Box::into_raw(ctx)).as_mut_ptr();
+            Box::from_raw(raw_ctx_ptr)
+        };
+
+        Poly1305State {
+            context: initialized_ctx,
         }
-        Poly1305State { context: ctx }
     }
 
     pub fn update(&mut self, data: &[u8]) -> () {
@@ -35,6 +39,7 @@ impl Poly1305State {
     }
 
     pub fn finalize(&mut self, output: &mut [u8]) -> () {
+        assert_eq!(output.len(), 16);
         unsafe { ffi::CRYPTO_poly1305_finish(self.context.as_mut(), output.as_mut_ptr()) };
     }
 }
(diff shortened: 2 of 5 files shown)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -1496,7 +1496,13 @@ def serialize_key_and_certificates_to_pkcs12(
...
         if self._fips_enabled:
             return False
-        return self._lib.Cryptography_HAS_POLY1305 == 1
+        elif (
+            self._lib.CRYPTOGRAPHY_IS_BORINGSSL
+            or self._lib.CRYPTOGRAPHY_IS_LIBRESSL
+        ):
+            return True
+        else:
+            return self._lib.Cryptography_HAS_POLY1305 == 1
```
woodruffw, line 1505: I wonder if it makes sense to compute `Cryptography_HAS_POLY1305` based on `CRYPTOGRAPHY_IS_BORINGSSL` and `CRYPTOGRAPHY_IS_LIBRESSL`, rather than special-casing them here? Not sure.

facutuesca: Maybe. I didn't do it because `Cryptography_HAS_POLY1305` is computed in [nid.py](https://github.com/pyca/cryptography/blob/34c2f024141e40232bf1de238bd6a4b8480cefa7/src/_cffi_src/openssl/nid.py#L40-L44) based on `NID_poly1305`, which is OpenSSL-specific (Boring and Libre do not have a NID for Poly1305).
So it was mostly me not wanting to put not-NID-related code inside `nid.py`. But if that's OK then we can do it there.

woodruffw: That makes sense to me. We'll see what @alex thinks 🙂

alex: I think this is probably fine as-is -- if we changed `Cryptography_HAS_POLY1305` to be true on libre/boring, we'd need a new name for its existing purpose.

src/rust/src/backend/poly1305.rs
```diff
@@ -7,26 +7,78 @@ use crate::buf::CffiBuf;
 use crate::error::{CryptographyError, CryptographyResult};
 use crate::exceptions;
 
-#[pyo3::prelude::pyclass(module = "cryptography.hazmat.bindings._rust.openssl.poly1305")]
-struct Poly1305 {
+#[cfg(any(CRYPTOGRAPHY_IS_BORINGSSL, CRYPTOGRAPHY_IS_LIBRESSL))]
+struct Poly1305Boring {
+    context: Option<Box<openssl_sys::poly1305_state>>,
```
alex, line 12: We need to seperate out the safe-wrapper part from the Python API part. Basically, the wrapper part should live in the cryptography-openssl crate -- see the hmac there as an example.

facutuesca: Done! Now the safe wrapper lives in `cryptography-openssl`. I also added a small comment explaining the use of `Box` for the C struct data

src/rust/cryptography-openssl/src/poly1305.rs
```diff
@@ -0,0 +1,40 @@
...
+        #[cfg(CRYPTOGRAPHY_IS_LIBRESSL)]
+        let mut ctx: Box<ffi::poly1305_state> = Box::new(ffi::poly1305_state {
+            aligner: 0,
+            opaque: [0; 136usize],
+        });
+
+        unsafe {
+            ffi::CRYPTO_poly1305_init(ctx.as_mut(), key.as_ptr());
+        }
+        Poly1305State { context: ctx }
```
alex, line 28: Hmm, should this be using `MaybeUninit`? Meaning we'd create a `Box<MaybeUninit<poly1305_state>>` pass the ptr to `init`, then convert that to `Box<poly1305_state>`?

src/rust/cryptography-openssl/src/poly1305.rs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    pub fn update(&mut self, data: &[u8]) -> () {
+        unsafe {
+            ffi::CRYPTO_poly1305_update(self.context.as_mut(), data.as_ptr(), data.len());
+        };
+    }
+
+    pub fn finalize(&mut self, output: &mut [u8]) -> () {
+        unsafe { ffi::CRYPTO_poly1305_finish(self.context.as_mut(), output.as_mut_ptr()) };
```
alex, line 38: 1. What happens if you call finalize 2x? As long as it's a soundness issue, that's fine.
2. We need to either length check output, or make it an array ptr. (Or we could have finalize return a `[u8; 16]`).

src/rust/cryptography-openssl/src/poly1305.rs
```diff
@@ -0,0 +1,40 @@
...
+        #[cfg(CRYPTOGRAPHY_IS_BORINGSSL)]
+        let mut ctx: Box<ffi::poly1305_state> = Box::new([0; 512usize]);
+        #[cfg(CRYPTOGRAPHY_IS_LIBRESSL)]
+        let mut ctx: Box<ffi::poly1305_state> = Box::new(ffi::poly1305_state {
+            aligner: 0,
+            opaque: [0; 136usize],
+        });
+
+        unsafe {
+            ffi::CRYPTO_poly1305_init(ctx.as_mut(), key.as_ptr());
```
alex, line 26: Need to have a length check for key, otherwise it'll buffer overflow.

src/rust/src/backend/poly1305.rs
```diff
@@ -35,33 +76,60 @@ impl Poly1305 {
...
+        Ok(result)
+    }
+}
+
+#[pyo3::prelude::pyclass(module = "cryptography.hazmat.bindings._rust.openssl.poly1305")]
+struct Poly1305 {
+    #[cfg(any(CRYPTOGRAPHY_IS_BORINGSSL, CRYPTOGRAPHY_IS_LIBRESSL))]
+    backend: Poly1305Boring,
+    #[cfg(not(any(CRYPTOGRAPHY_IS_LIBRESSL, CRYPTOGRAPHY_IS_BORINGSSL)))]
+    backend: Poly1305Open,
```
alex, line 118: I think you want to move the optionality up to this level, to reduce the duplication.

src/rust/cryptography-openssl/src/poly1305.rs
facutuesca: 1. Yeah, there are no crashes (if that's what you meant)
2. Done, I added an assert

src/rust/cryptography-openssl/src/poly1305.rs
facutuesca: Done!

src/rust/cryptography-openssl/src/poly1305.rs
```diff
@@ -0,0 +1,48 @@
+// This file is dual licensed under the terms of the Apache License, Version
+// 2.0, and the BSD License. See the LICENSE file in the root of this repository
+// for complete details.
+
+#[cfg(any(CRYPTOGRAPHY_IS_BORINGSSL, CRYPTOGRAPHY_IS_LIBRESSL))]
```
alex, line 5: Rather than putting this `cfg` on everything, does it make more sense to just `cfg` out the `pub mod` in `lib.rs`?

src/rust/cryptography-openssl/src/poly1305.rs
```diff
@@ -0,0 +1,48 @@
...
+        assert_eq!(key.len(), 32);
+        let mut ctx: Box<MaybeUninit<ffi::poly1305_state>> =
+            Box::new(MaybeUninit::<ffi::poly1305_state>::uninit());
+
+        // After initializing the context, unwrap the Box<MaybeUninit<poly1305_state>> into
+        // a Box<poly1305_state> while keeping the same memory address. See the docstring of the
+        // Poly1305State struct above for the rationale.
+        let initialized_ctx: Box<ffi::poly1305_state> = unsafe {
+            ffi::CRYPTO_poly1305_init(ctx.as_mut().as_mut_ptr(), key.as_ptr());
+            let raw_ctx_ptr = (*Box::into_raw(ctx)).as_mut_ptr();
```
alex, line 29: What is the deref + `as_mut_ptr` for?

src/rust/src/backend/poly1305.rs
```diff
@@ -35,33 +57,56 @@ impl Poly1305 {
...
+        Ok(result)
+    }
+}
+
+#[pyo3::prelude::pyclass(module = "cryptography.hazmat.bindings._rust.openssl.poly1305")]
+struct Poly1305 {
+    #[cfg(any(CRYPTOGRAPHY_IS_BORINGSSL, CRYPTOGRAPHY_IS_LIBRESSL))]
+    backend: Option<Poly1305Boring>,
+    #[cfg(not(any(CRYPTOGRAPHY_IS_LIBRESSL, CRYPTOGRAPHY_IS_BORINGSSL)))]
+    backend: Option<Poly1305Open>,
```
alex, line 95: nit; instead of backend can we call this `inner`?

src/rust/cryptography-openssl/src/poly1305.rs
facutuesca: `into_raw()` consumes the `Box` and returns a `*mut MaybeUninit<poly1305_state>`. In order to go from `*mut MaybeUninit<poly1305_state>` to `*mut poly1305_state`, I need to call `MaybeUninit::as_mut_ptr` (which I can only do after I dereference). Unrolling the code above and adding type annotations, you get:
```rust
let raw_maybe_ptr: *mut MaybeUninit<ffi::poly1305_state> = Box::into_raw(ctx);
let raw_ctx_ptr: *mut ffi::poly1305_state = (*raw_maybe_ptr).as_mut_ptr();
Box::from_raw(raw_ctx_ptr)
```

Is there any other way of achieving this? There's [an API](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.assume_init) to convert from `Box<MaybeUninit<T>>` to `Box<T>`, but it's still experimental.

alex: Got it

src/rust/cryptography-openssl/src/poly1305.rs
facutuesca: Makes sense, fixed!

src/rust/src/backend/poly1305.rs
facutuesca: yup, fixed

alex (approved):

PR #9399, opened 2023-08-10
Migrate ChaCha20Poly1305 AEAD to Rust

alex:
- [x] BoringSSL support
- [x] Re-use ctx
- [x] https://github.com/sfackler/rust-openssl/pull/2011


alex:
Grumble, not working due to https://github.com/openssl/openssl/issues/21853

alex:
This disable ctx reuse on OpenSSL 3.0-3.2. There's a bunch of code we can delete once we support only versions with workign copies.

Even with the ctx reuse disabled on 3.0.2, which is used in the benchmarks, it's 20% faster than main.

reaperhulk:
Do we use fixedpool anywhere now that this PR removes it from chacha?

alex:
Nope. Will remove it.

reaperhulk (approved):

reaperhulk:
I do not envy future us looking at those conditionals.

PR #9402, opened 2023-08-11
Try running an extra ruff rule

alex:
I think `from __future__ import annotations` means this is fine, even on older Python


reaperhulk (approved):

reaperhulk:
Actually, what happens if you run mypy on python 3.7 with this?

alex:
Should work (though I didn't test): mypy handles parsing these, not python.

reaperhulk:
(Safe in 3.7+ which is what we support so 🎉 )

PR #9405, opened 2023-08-11
Path validation: builder/verifier API skeletons

woodruffw:
This breaks out #8873, including everything *except* the top-level `verify` APIs (both Python and Rust).


alex:
I think we're going to want to break this down even further: basically take each type in the public API, and put that into its own PR.

woodruffw:
> I think we're going to want to break this down even further: basically take each type in the public API, and put that into its own PR.

Opened #9411 for just the `Store` bits; I'll do the same for others.

woodruffw:
Just to keep things synchronized here: this is blocked on #9548; once that's merged, we can deconflict here and determine what (if anything else) needs to be split out to get this to a reviewable size.

woodruffw:
This now contains just the `Policy` API (both Python and Rust), as well as the `PolicyBuilder` API and related types. I've deferred the various `permits_` methods on the Rust side to another PR, since they aren't surfaced directly in these APIs and would have made reviewing much more difficult.

woodruffw:
This is now ready for review. To summarize:

1. The following Python APIs are added: `Policy`, `PolicyBuilder`, `Profile`, and `Store`.
2. The above are mapped to pyO3 and pure-Rust types, where appropriate; `Store` is not yet mapped to its equivalent pure-Rust type since it wouldn't be easy to add coverage for.
3. The `PyCryptoOps` implementation of `CryptoOps` is added, and is covered by reimplementing `verify_directly_issued_by` in terms of it.

alex:
Given the size of this PR, I'm going to request you each of these types into their own PR, probably starting with just the `PolicyBuilder`, then we can review them 1 by 1.

alex:
As we've discussed, Profile should not be part of the design going forward, so it'll also need to be removed at some point.

woodruffw:
> As we've discussed, Profile should not be part of the design going forward, so it'll also need to be removed at some point.

Yeah, this might have been a confusion on my part: I removed the Rust-side profile abstraction, but kept the Python-side enum because it no longer reflects any sort of Rust abstraction (just which Rust-side constructor we call).

Are you thinking instead that this would be `PolicyBuilder().whatever(...).webpki().build()`? If so that makes sense, just want to avoid more confusion 🙂

alex:
Yes, that's right (though it's probably `PolicyBuilder.webpki().whatever().build()`, so the method is in a reasonable place). In general, unless said otherwise when I've been giving high level API feedback, its been applicable to both Rust and Python.

src/rust/cryptography-x509-validation/Cargo.toml
```diff
@@ -11,3 +11,6 @@ rust-version = "1.63.0"
 asn1 = { version = "0.15.5", default-features = false }
 cryptography-x509 = { path = "../cryptography-x509" }
 once_cell = "1"
+
+[dev-dependencies]
+pem = "1.1"
```
alex, line 16: Old version

src/rust/cryptography-x509-validation/src/policy/mod.rs
```diff
@@ -97,6 +100,39 @@ pub static WEBPKI_PERMITTED_ALGORITHMS: Lazy<HashSet<&AlgorithmIdentifier<'_>>>
...
+    /// the CA/B Forum's Basic Requirements.
+    pub fn webpki(ops: B, subject: Option<Subject<'a>>, time: asn1::DateTime) -> Self {
+        Self {
+            _ops: ops,
+            subject,
+            validation_time: time,
+        }
+    }
+}
+
```
alex, line 135: Remove

src/rust/src/x509/verify.rs
```diff
@@ -43,6 +47,134 @@ impl CryptoOps for PyCryptoOps {
...
+}
+
+fn build_subject_owner(
+    py: pyo3::Python<'_>,
+    subject: pyo3::Py<pyo3::PyAny>,
+) -> pyo3::PyResult<SubjectOwner> {
+    let subject = subject.as_ref(py);
+
+    let x509_general_name_module =
+        py.import(pyo3::intern!(py, "cryptography.x509.general_name"))?;
```
alex, line 107: We've replaced this pattern with the one seen in `types.rs`, basically never use `py.import` :-)

src/rust/src/x509/verify.rs
```diff
@@ -43,6 +47,134 @@ impl CryptoOps for PyCryptoOps {
...
+
+#[pyo3::prelude::pyfunction]
+fn create_server_verifier(
+    py: pyo3::Python<'_>,
+    subject: pyo3::Py<pyo3::PyAny>,
+    time: Option<&pyo3::PyAny>,
+) -> pyo3::PyResult<PyServerVerifier> {
+    let time = match time {
+        Some(time) => py_to_datetime(py, time)?,
+        None => datetime_now(py)?,
```
alex, line 162: Add to docs that if not specifies, uses `now` at the time the verifier is constructed

src/rust/cryptography-x509-validation/Cargo.toml
alex: Actually unused?

woodruffw: Yeah, unused now that `NullOps` is removed. I'll remove entirely.

src/rust/src/x509/verify.rs
woodruffw: [de2c7b0](https://github.com/pyca/cryptography/pull/9405/commits/de2c7b0bb2933e435b0bc07e2c8cb013981e1aa0)

src/rust/src/x509/verify.rs
woodruffw: [a5a88e6](https://github.com/pyca/cryptography/pull/9405/commits/a5a88e6291c7f85e8a1915d70c618ba407a5a56a)

src/rust/cryptography-x509-validation/Cargo.toml
woodruffw: [a5a88e6](https://github.com/pyca/cryptography/pull/9405/commits/a5a88e6291c7f85e8a1915d70c618ba407a5a56a)

src/rust/cryptography-x509-validation/src/policy/mod.rs
woodruffw: (Renamed to `new` per convo.)

docs/x509/verification.rst
```diff
@@ -30,6 +30,33 @@ chain building, etc.
...
+
+        The verifier's subject.
+
+    .. attribute:: validation_time
+
+        :type: :class:`datetime.datetime`
+
+        The verifier's validation time. If not specified during construction
+        (see :class:`PolicyBuilder`), this is set to
+        :meth:`datetime.datetime.now`.
```
alex, line 59: The `if not specified` should go in the policy builder docs, and should state that it's now at the time you call build

src/rust/src/types.rs
```diff
@@ -201,7 +201,6 @@ pub static UNRECOGNIZED_EXTENSION: LazyPyImport =
     LazyPyImport::new("cryptography.x509", &["UnrecognizedExtension"]);
 pub static EXTENSION: LazyPyImport = LazyPyImport::new("cryptography.x509", &["Extension"]);
 pub static EXTENSIONS: LazyPyImport = LazyPyImport::new("cryptography.x509", &["Extensions"]);
-pub static IPADDRESS: LazyPyImport = LazyPyImport::new("cryptography.x509", &["IPAddress"]);
```
woodruffw, line 204: NB: I moved this down and renamed it to make it consistent with `DNS_NAME` + co-locate it with the other exported name types (`DNS_NAME` and `RFC822_NAME`).

src/rust/src/x509/verify.rs
```diff
@@ -43,6 +47,125 @@ impl CryptoOps for PyCryptoOps {
     }
 }
 
+struct FixedPolicy<'a>(Policy<'a, PyCryptoOps>);
```
alex, line 50: Is there a better name than `FixedPolicy`?

src/rust/src/x509/verify.rs
```diff
@@ -43,6 +47,125 @@ impl CryptoOps for PyCryptoOps {
...
+
+/// This enum exists solely to provide heterogeneously typed ownership for `OwnedPolicy`.
+enum SubjectOwner {
+    // TODO: Switch this to `Py<PyString>` once Pyo3's `to_str()` preserves a
+    // lifetime relationship between an a `PyString` and its borrowed `&str`
+    // reference in all limited API builds. PyO3 can't currently do that in
+    // older limited API builds because it needs `PyUnicode_AsUTF8AndSize` to do
+    // so, which was only stabilized with 3.10.
```
alex, line 58: Doesn't need to block this, but would be good to send a PR that implements the method for Python >= 3.10.

src/rust/src/x509/verify.rs
```diff
@@ -43,6 +47,125 @@ impl CryptoOps for PyCryptoOps {
...
+    DNSName((pyo3::Py<pyo3::PyAny>, String)),
+    IPAddress((pyo3::Py<pyo3::PyAny>, pyo3::Py<pyo3::types::PyBytes>)),
```
alex, line 60: I think it'd be marginally cleaner if there was a seperate `py_subject` field on the Verifier, and then `SubjectOwner`'s enum elements only needed to be a single field.

src/rust/src/x509/verify.rs
```diff
@@ -43,6 +47,125 @@ impl CryptoOps for PyCryptoOps {
...
+) -> pyo3::PyResult<PyServerVerifier> {
+    let time = match time {
+        Some(time) => py_to_datetime(py, time)?,
+        None => datetime_now(py)?,
+    };
+
+    let subject_owner = build_subject_owner(py, subject)?;
+    let policy = OwnedPolicy::try_new(subject_owner, |subject_owner| {
+        let subject = build_subject(py, subject_owner)?;
+        Ok::<FixedPolicy<'_>, pyo3::PyErr>(FixedPolicy(Policy::new(PyCryptoOps {}, subject, time)))
```
alex, line 163: Is the turbofish required here?

woodruffw: Yes, unfortunately:

```
error[E0282]: type annotations needed
   --> src/x509/verify.rs:163:9
    |
163 |         Ok(FixedPolicy(Policy::new(PyCryptoOps {},...
    |         ^^ cannot infer type of the type parameter `E` declared on the enum `Result`
    |
help: consider specifying the generic arguments
    |
163 |         Ok::<x509::verify::FixedPolicy<'_>, E>(FixedPolicy(Policy::new(PyCryptoOps {}, subject, time)))
    |           ++++++++++++++++++++++++++++++++++++

For more information about this error, try `rustc --explain E0282`.
error: could not compile `cryptography-rust` (lib) due to previous error
```

src/rust/src/x509/verify.rs
woodruffw: Hmm, maybe `BoundPolicy` (to emphasize that we're binding a `Policy` to its `PyCryptoOps`? Or maybe I could just import `Policy` as `_Policy` or similar and rename this to `Policy`.

alex: How about `PyCryptoPolicy` :-)

docs/x509/verification.rst
woodruffw: [2e7bd70](https://github.com/pyca/cryptography/pull/9405/commits/2e7bd70c01465a4983e1e9a3e8ce0d4a930055d2)

src/rust/src/x509/verify.rs
woodruffw: https://github.com/pyca/cryptography/pull/9405/commits/35f5e6f8a91c012fc59d3e5d7b3f965f9b8c393b

src/rust/src/x509/verify.rs
woodruffw: https://github.com/pyca/cryptography/pull/9405/commits/3c036bcd157c0ccb92d50e1a8e44d6099e970fe7

src/rust/src/x509/verify.rs
```diff
@@ -43,6 +46,129 @@ impl CryptoOps for PyCryptoOps {
...
+#[pyo3::pyclass(
+    name = "ServerVerifier",
+    module = "cryptography.hazmat.bindings._rust.x509"
+)]
+struct PyServerVerifier {
+    // TODO: Switch this to `Py<PyString>` once Pyo3's `to_str()` preserves a
+    // lifetime relationship between an a `PyString` and its borrowed `&str`
+    // reference in all limited API builds. PyO3 can't currently do that in
+    // older limited API builds because it needs `PyUnicode_AsUTF8AndSize` to do
+    // so, which was only stabilized with 3.10.
```
alex, line 75: Pretty sure this goes in `SubjectOwner`?

src/rust/src/x509/verify.rs
```diff
@@ -43,6 +46,129 @@ impl CryptoOps for PyCryptoOps {
...
+    py_subject: pyo3::Py<pyo3::PyAny>,
```
alex, line 76: You can use `pyo3(get, name = "subject")` here (https://pyo3.rs/v0.19.2/class#object-properties-using-pyo3get-set)

woodruffw: [fbb2e6a](https://github.com/pyca/cryptography/pull/9405/commits/fbb2e6a63d521795acd4b38d285dac8a71b50626)

src/rust/src/x509/verify.rs
woodruffw: Yep, bad copy. Thanks.

alex (approved):

src/rust/src/x509/verify.rs
woodruffw: https://github.com/pyca/cryptography/pull/9405/commits/0f9126570baf427166ac211a883eef6134a5e67b

PR #9412, opened 2023-08-14
common: add more RSA-PSS algorithm id definitions

woodruffw:
This adds hash and mask generation algorithm ID definitions for the three RSA-PSS configurations allowed under the CABF BRs, mirroring the existing definitions for RSA-PSS with SHA-1.

Breakout from #9405.

See `7.1.3.2.1 RSA` in the CABF BRs for this: https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-v2.0.0.pdf (page 97).

### DIFF
--- a/src/rust/cryptography-x509/src/common.rs
+++ b/src/rust/cryptography-x509/src/common.rs
@@ -234,6 +234,22 @@ pub const PSS_SHA1_HASH_ALG: AlgorithmIdentifier<'_> = AlgorithmIdentifier {
     params: AlgorithmParameters::Sha1(Some(())),
 };
 
+// RSA-PSS ASN.1 hash algorithm definitions specified under the CA/B Forum BRs.
+pub const PSS_SHA256_HASH_ALG: AlgorithmIdentifier<'_> = AlgorithmIdentifier {
+    oid: asn1::DefinedByMarker::marker(),
+    params: AlgorithmParameters::Sha256(Some(())),
+};
+
+pub const PSS_SHA384_HASH_ALG: AlgorithmIdentifier<'_> = AlgorithmIdentifier {
+    oid: asn1::DefinedByMarker::marker(),
+    params: AlgorithmParameters::Sha384(Some(())),
+};
+
+pub const PSS_SHA512_HASH_ALG: AlgorithmIdentifier<'_> = AlgorithmIdentifier {
+    oid: asn1::DefinedByMarker::marker(),
+    params: AlgorithmParameters::Sha512(Some(())),
+};
+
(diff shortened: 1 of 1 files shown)


src/rust/cryptography-x509/src/common.rs
```diff
@@ -249,6 +265,22 @@ pub const PSS_SHA1_MASK_GEN_ALG: MaskGenAlgorithm<'_> = MaskGenAlgorithm {
...
+
+pub const PSS_SHA384_MASK_GEN_ALG: MaskGenAlgorithm<'_> = MaskGenAlgorithm {
+    oid: oid::MGF1_OID,
+    params: PSS_SHA384_HASH_ALG,
+};
+
+pub const PSS_SHA512_MASK_GEN_ALG: MaskGenAlgorithm<'_> = MaskGenAlgorithm {
+    oid: oid::MGF1_OID,
+    params: PSS_SHA512_HASH_ALG,
+};
```
woodruffw, line 282: NB: No coverage needed since these are just data, but for maximum correctness I could also test these against the exact expected DER encodings supplied by the CABF rules.

src/rust/cryptography-x509/src/common.rs
```diff
@@ -234,6 +234,22 @@ pub const PSS_SHA1_HASH_ALG: AlgorithmIdentifier<'_> = AlgorithmIdentifier {
     params: AlgorithmParameters::Sha1(Some(())),
 };
 
+// RSA-PSS ASN.1 hash algorithm definitions specified under the CA/B Forum BRs.
```
alex, line 237: Is this comment really correct? It's not really defined by CABF, these come from the RFCs, right?

woodruffw: I think that's right, although CABF doesn't make reference to the RFCs for their definitions of these. Want me to dig them up regardless?

woodruffw: https://www.rfc-editor.org/rfc/rfc8017 looks like the right RFC, although it doesn't explicitly parametrize these the way CABF does.

alex: Hmm, really? I just assumed that these were all natural consequences of how it worked, and didn't need to be explicitly specified (same as SHA1)

woodruffw: Yeah, CABF specifies the precise set that can be used (and provides their exact DER encodings as well) -- both are on page 97 of the PDF I linked in the top.

alex: Ok

alex (approved):

PR #9439, opened 2023-08-17
Build wheels with python 3.11, targeting 3.7

alex:


reaperhulk:
This appears to be using 3.9 rather than 3.10? Also, why not just use 3.11?

alex:
Because I'd confused myself, fixed.

reaperhulk (approved):

PR #9456, opened 2023-08-18
LibreSSL 3.8.1 and later is OPENSSL_NO_ENGINE

botovq:
Unfortunately, some projects are not prepared to build without ENGINE symbols, so just like BoringSSL we needed to keep some stubs.


alex (approved):

alex:
Thanks!

PR #9490, opened 2023-08-23
Remove setup.py

alex:
All of our configuration is now declarative


reaperhulk:
I worry about the loss of troubleshooting instructions when install fails, but I guess we'll look at enhancing pyproject.toml's spec to allow that if needed.

reaperhulk (approved):

PR #9506, opened 2023-08-26
Add support for Python 3.12

hugovk:
The [Python 3.12 release candidate is out!](https://discuss.python.org/t/python-3-12-0-release-candidate-1-released/31137?u=hugovk) :rocket:

> ## Call to action
> 
> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.12 compatibilities during this phase, and where necessary publish Python 3.12 wheels on PyPI to be ready for the final release of 3.12.0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
           - {VERSION: "3.11", NOXSESSION: "tests", OPENSSL: {TYPE: "libressl", VERSION: "3.7.3"}}
           - {VERSION: "3.11", NOXSESSION: "tests", OPENSSL: {TYPE: "libressl", VERSION: "3.8.0"}}
           - {VERSION: "3.11", NOXSESSION: "tests-randomorder"}
-          - {VERSION: "3.12-dev", NOXSESSION: "tests"}
+          - {VERSION: "3.12", NOXSESSION: "tests"}
           # Latest commit on the BoringSSL master branch, as of Aug 26, 2023.
           - {VERSION: "3.11", NOXSESSION: "tests", OPENSSL: {TYPE: "boringssl", VERSION: "792e77c52b5a85bee15a6f644494c10d8db5f7a0"}}
           # Latest commit on the OpenSSL master branch, as of Aug 26, 2023.
@@ -66,6 +66,7 @@ jobs:
         uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1 # v4.7.0
         with:
           python-version: ${{ matrix.PYTHON.VERSION }}
+          allow-prereleases: true
           cache: pip
           cache-dependency-path: ci-constraints-requirements.txt
       - name: Setup rust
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           - {VERSION: "3.11", NOXSESSION: "tests", OPENSSL: {TYPE: "libressl", VERSION: "3.7.3"}}
           - {VERSION: "3.11", NOXSESSION: "tests", OPENSSL: {TYPE: "libressl", VERSION: "3.8.0"}}
           - {VERSION: "3.11", NOXSESSION: "tests-randomorder"}
-          - {VERSION: "3.12-dev", NOXSESSION: "tests"}
+          - {VERSION: "3.12", NOXSESSION: "tests"}
```
alex, line 45: Why is this preferable?

.github/workflows/ci.yml
```diff
@@ -223,6 +224,7 @@ jobs:
         PYTHON:
           - {VERSION: "3.7", NOXSESSION: "tests-nocoverage"}
           - {VERSION: "3.11", NOXSESSION: "tests"}
+          - {VERSION: "3.12", NOXSESSION: "tests"}
```
alex, line 227: On macOS and Windows, we test on only two versions of Python: our minimum, and our maximum.

alex (commented):
As you can see, we already test on Python 3.12. As such, I don't think this PR makes much sense.

hugovk:
> As you can see, we already test on Python 3.12. As such, I don't think this PR makes much sense.

Sure, if you think there's no value in also testing macOS and Windows on 3.12, let's close this. (Personally I'd want to make sure they also pass to avoid any surprises later.)

How about the Trove classifier? Do you want to wait for closer to full release for that?

.github/workflows/ci.yml
hugovk: Some people prefer to include `allow-prereleases: true` so that they can omit the `-dev`, meaning there's no need to tidy it up later. But not much difference otherwise.

alex:
PyPI already supports the classifier?

I'm ambivalent about adding now vs. release day. @reaperhulk ?

hugovk:
> PyPI already supports the classifier?

Yup! 33 of the top 360 have set it, and 1,785 in total:

* https://pyreadiness.org/3.12/
* https://pypi.org/search/?c=Programming+Language+%3A%3A+Python+%3A%3A+3.12


(Also [3.13](https://pypi.org/search/?c=Programming+Language+%3A%3A+Python+%3A%3A+3.13).)

alex:
I guess I'd be fine with a PR that did the classifier and nothing else. Thanks.

reaperhulk:
A classifier PR is fine by me.

Regarding why we don't test on 3.12 dev on macOS/windows: Over the ten years we've been working on this project we've repeatedly rebuilt the CI to make it faster and more efficient while retaining as much coverage as possible. Through painful experience we've made the decision that we want to test minimum and maximum (released) versions on those platforms while using linux resources to test more broadly. Notably, in that time we've never had a Python bug which only manifested on one platform. That's not to say it's impossible, but we consider the ROI on those additional CI jobs to be too low.

Regardless, thanks 😄

hugovk:
That sounds more than reasonable. It's good one platform is testing 3.12, that will catch things like deprecations and syntax changes. I'll open the other PR, thanks!

PR #9548, opened 2023-09-06
x509/policy: add WebPKI permitted algorithms

woodruffw:
This is another breakout from #8873 and #9405.

~~WIP while we work out some unexpected encodings.~~

CC @facutuesca to take over.


woodruffw:
xref https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-v2.0.0.pdf pages 97 ff. for the canonical encodings here.

src/rust/cryptography-x509-validation/src/policy/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+
+use once_cell::sync::Lazy;
+
+use cryptography_x509::common::{
+    AlgorithmIdentifier, AlgorithmParameters, RsaPssParameters, PSS_SHA256_HASH_ALG,
+    PSS_SHA256_MASK_GEN_ALG, PSS_SHA384_HASH_ALG, PSS_SHA384_MASK_GEN_ALG, PSS_SHA512_HASH_ALG,
+    PSS_SHA512_MASK_GEN_ALG,
+};
+
+pub static WEBPKI_PERMITTED_ALGORITHMS: Lazy<HashSet<AlgorithmIdentifier<'_>>> = Lazy::new(|| {
```
woodruffw, line 15: @facutuesca Might be worth adding a comment linking to the CA/B Forum BRs doc here (with a reference to the pages these are from)

(Sorry, I forgot to do this myself 😅)

facutuesca: Done!

src/rust/cryptography-x509-validation/src/policy/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        let mut actual_encodings: Vec<_> = WEBPKI_PERMITTED_ALGORITHMS
+            .iter()
+            .map(|ai| asn1::write_single(ai).unwrap())
+            .collect();
+        actual_encodings.sort();
+
+        for (exp, act) in expected_encodings.iter().zip(actual_encodings.iter()) {
+            assert_eq!(act, exp);
+        }
```
alex, line 128: I think it'd make a lot more sense if each of the allowed algortihms was its own const, which we asserted had the correct encoding.

src/rust/cryptography-x509-validation/src/policy/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+            .unwrap()
+    }
+
+    #[test]
+    fn test_webpki_permitted_algorithms_canonical_encodings() {
+        // Expected encodings, from CA/B Forum's Baseline Requirements, section 7.1.3.2 (pages 96-98)
+        // https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-v2.0.0.pdf
+        let mut expected_encodings: Vec<_> = vec![
+            // RSASSA‐PKCS1‐v1_5 with SHA‐256
+            decode_hex("300d06092a864886f70d01010b0500"),
```
alex, line 100: Can we just write the hex-decoded bytes here as a `b""`.

src/rust/cryptography-x509-validation/src/policy/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+
+use cryptography_x509::common::{
+    AlgorithmIdentifier, AlgorithmParameters, RsaPssParameters, PSS_SHA256_HASH_ALG,
+    PSS_SHA256_MASK_GEN_ALG, PSS_SHA384_HASH_ALG, PSS_SHA384_MASK_GEN_ALG, PSS_SHA512_HASH_ALG,
+    PSS_SHA512_MASK_GEN_ALG,
+};
+
+/// Permitted algorithms, from CA/B Forum's Baseline Requirements, section 7.1.3.2 (pages 96-98)
+/// https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-v2.0.0.pdf
+pub static WEBPKI_PERMITTED_ALGORITHMS: Lazy<HashSet<AlgorithmIdentifier<'_>>> = Lazy::new(|| {
```
alex, line 17: How do you envision this being used?

src/rust/Cargo.toml
```diff
@@ -14,6 +14,7 @@ asn1 = { version = "0.15.5", default-features = false }
 cryptography-cffi = { path = "cryptography-cffi" }
 cryptography-x509 = { path = "cryptography-x509" }
 cryptography-openssl = { path = "cryptography-openssl" }
+cryptography-x509-validation = { path = "cryptography-x509-validation" }
```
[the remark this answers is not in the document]
woodruffw: Yep, removing.

src/rust/cryptography-x509-validation/src/policy/mod.rs
woodruffw: I envisoned it as part of the algorithms restriction functionality added with #9405: https://github.com/pyca/cryptography/pull/9405/files#diff-b15c4ba048c4bbef3e3e40afef6eaf31c7edcdd7d46627cb470c3c8db684a040R223-R254

alex: Ok we might need to take a step back here, because when we spoke I think we agreed to move away from the profiles idea, so I'm struggling a bit to understand how the pieces fit together here.

woodruffw: Yeah, there are no more profiles in the current design -- there's only one "policy," and in our last discussion we came to the conclusion (IIRC) that the right way to offer configurability here was to allow "helper" instantiations like `Policy::webpki()` that pre-fill the things that would otherwise go under a full "profile."

alex (approved):

Issue #9581, opened 2023-09-11
Add public algorithm/MGF properties to paddings


mccoyp:
Hello! I work on the Azure SDK for Python, and we recently got a feature request to support `cryptography`'s `RSAPrivateKey` interface: https://github.com/Azure/azure-sdk-for-python/issues/31877.

The only trouble with implementation has come from the opacity of `AsymmetricPadding` types. The aim is to use Azure Key Vault-managed keys to perform private key operations, but we can't determine the algorithm or MGF to use in some cases from operation inputs.

For example, the `RSAPrivateKey` interface's `decrypt` method accepts a ciphertext and padding. To decrypt with Key Vault, we need a ciphertext and encryption algorithm. This is fine for `PKCS1v15`, but `OAEP`'s algorithm only seems to be stored in a private `_algorithm` attribute. We'd have to either use a default encryption algorithm or access a private attribute -- we'd like to avoid either option.

Similarly, for `sign`, we accept data, padding, and an algorithm. Even though `PSS` only accepts `MGF1`, it would be ideal if we could inspect the padding's MGF with a public property.

Is there a reason these attributes are private? Also, is there a recommended method of implementing interfaces that have this dilemma?

Thank you!

alex:
I think we'd be happy to take PRs that make these properties public (via an `@property` so they're not mutable), I don't think there's a good reason for them to be private, other than that no one has ever asked.

reaperhulk:
As @alex mentioned we generally default to private until we have a reason to make them public. This seems like a fine reason since we'd like to have key management services supporting these interfaces.

mccoyp:
@alex @reaperhulk Awesome, thanks for the quick responses! I'll get started on a PR and update when it's available.

PR #9582, opened 2023-09-11
Add algorithm/MGF properties to asymmetric paddings

mccoyp:
Resolves https://github.com/pyca/cryptography/issues/9581.


alex (commented):
Can you add tests for these -- they'll be trivial, but we require 100% test coverage.

Thanks!

mccoyp:
@alex Things should be in a reviewable state now, thanks! I exposed the `MGF` class in docs now that something's pointing to it, but please let me know if you would prefer a different approach.

alex (approved):

PR #9589, opened 2023-09-12
x509/sct: replace another utcfromtimestamp call

### LINKED ISSUE #9580: Call to `utcfromtimestamp` deprecated in Python 3.12
`src/rust/src/x509/sct.rs` still calls `datetime.utcfromtimestamp()`, which is deprecated in Python 3.12. The Python 3.12. release notes say:

> call [now()](https://docs.python.org/3.12/library/datetime.html#datetime.datetime.now) and [fromtimestamp()](https://docs.python.org/3.12/library/datetime.html#datetime.datetime.fromtimestamp) with the tz parameter set to [datetime.UTC](https://docs.python.org/3.12/library/datetime.html#datetime.UTC).

Unfortunately my rust-skills are non-existent, so I cannot contribute without a huge effort on what I suspect will be a very trivial change. I hope that's okay.

This is related to #9260.

woodruffw:
Fixes #9580.

### DIFF
--- a/src/rust/src/x509/sct.rs
+++ b/src/rust/src/x509/sct.rs
@@ -4,7 +4,6 @@
 
 use crate::error::CryptographyError;
 use crate::types;
-use pyo3::types::IntoPyDict;
 use pyo3::ToPyObject;
 use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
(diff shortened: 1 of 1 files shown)


src/rust/src/x509/sct.rs
```diff
@@ -175,17 +174,19 @@ impl Sct {
...
-                (self.timestamp / 1000,),
+                pyo3::intern!(py, "fromtimestamp"),
+                (self.timestamp / 1000, utc),
             )?
-            .call_method(
-                "replace",
-                (),
-                Some(vec![("microsecond", self.timestamp % 1000 * 1000)].into_py_dict(py)),
-            )
+            .call_method("replace", (), Some(kwargs))
```
woodruffw, line 189: NB: We have to do this song-and-dance because `fromtimestamp` creates an aware `datetime` when given a timezone, which breaks the current API promise (a naive `datetime`).

If we don't pass it a timezone it "cleverly" adjusts the given timestamp to the local timezone and returns it as a naive `datetime`, which is also incorrect. Instead, we create it as an aware `datetime` and update the pre-existing call to `replace` to make it naive after the fact.

This is the aware-to-naive conversion mechanism documented (encouraged?) by Python itself: https://docs.python.org/3/library/datetime.html#datetime.datetime.replace

reaperhulk (approved):

PR #9606, opened 2023-09-14
rust: add PyCryptoOps

woodruffw:
Reimplements `verify_directly_issued_by` in terms of `PyCryptoOps`, for free coverage.

Breakout from #9405.

### DIFF
--- a/src/rust/src/x509/verify.rs
+++ b/src/rust/src/x509/verify.rs
@@ -2,7 +2,46 @@
 // 2.0, and the BSD License. See the LICENSE file in the root of this repository
 // for complete details.
 
-use crate::x509::certificate::Certificate as PyCertificate;
+use cryptography_x509::certificate::Certificate;
+use cryptography_x509_validation::ops::CryptoOps;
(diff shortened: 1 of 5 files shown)


src/rust/src/x509/verify.rs
```diff
@@ -2,7 +2,46 @@
...
+            let spki_der = pyo3::types::PyBytes::new(py, &asn1::write_single(&cert.tbs_cert.spki)?);
+
+            Ok(types::LOAD_DER_PUBLIC_KEY
+                .get(py)?
+                .call1((spki_der,))?
+                .into())
+        })
+    }
+
+    fn is_signed_by(&self, cert: &Certificate<'_>, key: Self::Key) -> Result<(), Self::Err> {
```
alex, line 33: This name needs to be `verify_signed_by` now that it's not returning a bool

woodruffw: baff77803

alex (approved):

PR #9608, opened 2023-09-15
validation/ops: add test-only NullOps

woodruffw:
Another breakout from #9405.

This adds a `cfg(test)`-only `NullOps` implementation of `CryptoOps`, which #9405 requires for coverage of some of its API signatures.


alex:
I think it would be more helpful to try to break out some more pieces starting at the Python API: for example, making the `build_server_verifier_` function do something :-)

It's very difficult to evaluate the merits of a PR like this -- it's hard to tell how it's useful or if it's the best design direction.

woodruffw:
Closing as premature.

PR #9642, opened 2023-09-21
verification: fill in policy API internals

woodruffw:
This follows #9405 by rounding out the internals of the Rust-side `Policy` API.

Break-out from #8873.


src/rust/cryptography-x509-validation/src/policy/mod.rs
```diff
@@ -100,17 +106,70 @@ pub static WEBPKI_PERMITTED_ALGORITHMS: Lazy<HashSet<&AlgorithmIdentifier<'_>>>
...
+    &[BASIC_CONSTRAINTS_OID, SUBJECT_ALTERNATIVE_NAME_OID];
+
 /// Represents a logical certificate "subject," i.e. a principal matching
 /// one of the names listed in a certificate's `subjectAltNames` extension.
 pub enum Subject<'a> {
     DNS(DNSName<'a>),
     IP(IPAddress),
 }
 
+impl Subject<'_> {
```
alex, line 121: General name matching is likely something that could be extracted and tested in its own PR.

src/rust/cryptography-x509-validation/src/policy/mod.rs
```diff
@@ -119,28 +178,103 @@ pub struct Policy<'a, B: CryptoOps> {
...
+    pub(crate) critical_ee_extensions: HashSet<ObjectIdentifier>,
 }
 
 impl<'a, B: CryptoOps> Policy<'a, B> {
-    /// Creates a new policy with the given `CryptoOps`, an optional subject,
-    /// and a validation time.
-    pub fn new(ops: B, subject: Option<Subject<'a>>, time: asn1::DateTime) -> Self {
+    /// Create a new policy with defaults for the certificate profile defined in
+    /// the CA/B Forum's Basic Requirements.
+    pub fn webpki(ops: B, subject: Option<Subject<'a>>, time: asn1::DateTime) -> Self {
```
alex, line 201: Why did WebPKI come back?

src/rust/cryptography-x509-validation/src/policy/mod.rs
```diff
@@ -119,28 +178,103 @@ pub struct Policy<'a, B: CryptoOps> {
...
+    /// algorithm identifiers.
+    ///
+    /// If not `None`, all certificates validated by this policy MUST
+    /// have a signature algorithm in this set.
+    ///
+    /// If `None`, all signature algorithms are permitted.
+    pub permitted_algorithms: Option<HashSet<AlgorithmIdentifier<'a>>>,
+
+    pub(crate) critical_ca_extensions: HashSet<ObjectIdentifier>,
+    pub(crate) critical_ee_extensions: HashSet<ObjectIdentifier>,
```
alex, line 195: Do I understand right that the point of these to enforce that some extensions must be marked critical?

@reaperhulk let me know if you disagree, but I don't think we care about enforcing that: it's a lot of compatibility cost, for effectively no gain. I don't believe any other validators enforce "this extension must be marked critical", just the inverse (rejecting unknown critical extensions).

src/rust/cryptography-x509-validation/src/policy/mod.rs
woodruffw: Bad merge, will fix.

src/rust/cryptography-x509-validation/src/policy/mod.rs
woodruffw: https://github.com/pyca/cryptography/pull/9659

src/rust/cryptography-x509-validation/src/policy/mod.rs
woodruffw: Fixed in [90e502b](https://github.com/pyca/cryptography/pull/9642/commits/90e502b9e240b3c7aa4751d3b57f2c94f38297da) -- I kept the doc change since the defaults are changing internally, but I can also revert that if you'd prefer.

src/rust/cryptography-x509-validation/src/policy/mod.rs
```diff
@@ -119,28 +178,103 @@ pub struct Policy<'a, B: CryptoOps> {
...
+
+    /// Inform this policy of an expected critical extension in EE certificates.
+    ///
+    /// This allows the policy to accept critical extensions that the baseline
+    /// profile does not cover. The user is responsible for separately validating
+    /// these extensions.
+    pub fn assert_critical_ee_extension(mut self, oid: ObjectIdentifier) -> Self {
+        self.critical_ee_extensions.insert(oid);
+        self
+    }
```
woodruffw, line 240: NB: These were included as intended APIs for the Rust side only; I can remove them to keep the diff small for the MVP if you'd prefer.

src/rust/cryptography-x509-validation/src/policy/mod.rs
```diff
@@ -153,20 +170,82 @@ pub struct Policy<'a, B: CryptoOps> {
...
+
+    pub(crate) critical_ca_extensions: HashSet<ObjectIdentifier>,
+    pub(crate) critical_ee_extensions: HashSet<ObjectIdentifier>,
 }
 
 impl<'a, B: CryptoOps> Policy<'a, B> {
-    /// Creates a new policy with the given `CryptoOps`, an optional subject,
-    /// and a validation time.
+    /// Create a new policy with defaults for the certificate profile defined in
+    /// the CA/B Forum's Basic Requirements.
```
alex, line 192: The only default from CAB here is the permitted algorithms and the EKU right?

alex (commented):
I'm going to apologize, because I think this is a recurring theme in my reviews:

This PR brings in the APIs around customizing the verification logic in various ways.

I think at this point it'd be better to get a minimal path verifier, and then layer these additional validations on top of that, rather than build in the extension points and then try to do a complete verifier.

Happy to hear arguments the other way, but I think it'll be much easier to review if we can do the functional pieces and then generalize.

src/rust/cryptography-x509-validation/src/policy/mod.rs
woodruffw: Yep, the actual policy logic itself isn't filled in yet.

src/rust/cryptography-x509-validation/src/policy/mod.rs
woodruffw: Removed.

alex (approved):

PR #9659, opened 2023-09-27
validation/policy: general name matching

woodruffw:
Breakout of #9642.


src/rust/cryptography-x509-validation/src/policy/mod.rs
```diff
@@ -107,6 +109,38 @@ pub enum Subject<'a> {
     IP(IPAddress),
 }
 
+impl Subject<'_> {
+    fn general_name_matches(&self, general_name: &GeneralName<'_>) -> bool {
```
alex, line 113: Can we name this `subject_alt_name_matches` or something like that -- I want to distinguish between the name constraints matching use case, which is subtly different.

woodruffw: [8ecaafb](https://github.com/pyca/cryptography/pull/9659/commits/8ecaafbc6d1332c3afce7ce3939faa87a0ab4af7)

alex (approved):

PR #9681, opened 2023-10-02
Backport fix for CVE-2023-23931 - Don't allow update_into mutate immutable objects

decko:
This is a Backport from 39.0.1 for a fix of the CVE-2023-23931

- Don't allow update_into to mutate immutable objects

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/ciphers.py
+++ b/src/cryptography/hazmat/backends/openssl/ciphers.py
@@ -157,7 +157,7 @@ class _CipherContext:
         data_processed = 0
         total_out = 0
         outlen = self._backend._ffi.new("int *")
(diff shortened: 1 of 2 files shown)


reaperhulk:
You can't "backport" from a branch to main, the fix is already present.

https://github.com/pyca/cryptography/blob/4166517b739558ce5efb8d8067e0b681abd36654/src/cryptography/hazmat/backends/openssl/ciphers.py#L159

reaperhulk:
To correct myself a bit, I see this actually backports to 38.0.x (in a branch correction shortly before I closed it). We do not support older versions of cryptography in this fashion -- it is our policy that users upgrade to a supported version to get fixes of this form. Others (including OS distributions) may choose to create their own backports, which is of course their prerogative.

PR #9682, opened 2023-10-02
Use 3.12 final in CI

alex:


reaperhulk:
This won't work until https://github.com/actions/python-versions/releases has a 3.12 final I think?

reaperhulk (approved):

PR #9700, opened 2023-10-06
validation/types: add DNSConstraint, rename IPConstraint

woodruffw:
This further fleshes out the helper types for name constraint checking, as a breakout from #8873.

It also renames `IPRange` to `IPConstraint`, to emphasize that the type is relevant to name constraint checking, not SAN checking (which does exact IP matching).

### DIFF
--- a/src/rust/cryptography-x509-validation/src/types.rs
+++ b/src/rust/cryptography-x509-validation/src/types.rs
@@ -69,6 +69,16 @@ impl<'a> DNSName<'a> {
             None => None,
         }
     }
+
+    /// Returns this DNS name's labels, in reversed order
+    /// (from top-level domain to most-specific subdomain).
+    fn rlabels(&self) -> impl Iterator<Item = &'_ str> {
+        self.as_str()
+            .split('.')
+            .collect::<Vec<_>>()
+            .into_iter()
+            .rev()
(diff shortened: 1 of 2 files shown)


src/rust/cryptography-x509-validation/src/types.rs
```diff
@@ -69,6 +69,16 @@ impl<'a> DNSName<'a> {
...
     }
+
+    /// Returns this DNS name's labels, in reversed order
+    /// (from top-level domain to most-specific subdomain).
+    fn rlabels(&self) -> impl Iterator<Item = &'_ str> {
+        self.as_str()
+            .split('.')
+            .collect::<Vec<_>>()
+            .into_iter()
+            .rev()
```
alex, line 80: Surely this can just be an rsplit?

woodruffw: Yeah, thanks for catching that -- I'm surprised clippy didn't. Will fix.

woodruffw: Fixed, and opened a Clippy request: https://github.com/rust-lang/rust-clippy/issues/11632

alex (approved):

PR #9712, opened 2023-10-11
Added missing type annotation

alex:


reaperhulk (approved):
Feels like it could infer that 😂

PR #9744, opened 2023-10-20
validation: add Rust-side trust store APIs

woodruffw:
Breakout from #8873.

### DIFF
--- a/src/rust/cryptography-x509-validation/src/ops.rs
+++ b/src/rust/cryptography-x509-validation/src/ops.rs
@@ -45,10 +45,18 @@ pub(crate) mod tests {
         }
     }
 
-    #[test]
-    fn test_nullops() {
-        // Arbitrary relatively small cert (v1_cert.pem from cryptography_vectors).
-        let v1_cert = "
+    #[macro_export]
+    macro_rules! cert {
+        ($pem:literal) => {{
+            let parsed = Box::leak(Box::new(pem::parse($pem).unwrap()));
+            asn1::parse_single::<Certificate<'static>>(&parsed.contents()).unwrap()
+        }};
+    }
+
+    // Arbitrary relatively small cert (v1_cert.pem from cryptography_vectors).
+    pub(crate) fn v1_cert() -> Certificate<'static> {
+        cert!(
+            "
 -----BEGIN CERTIFICATE-----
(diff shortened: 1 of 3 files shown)


woodruffw:
Documenting here since the context is probably lost elsewhere: the value of a separate `Store` type here is twofold:

* Misuse resistance: a newtype here means that users can't accidentally confuse the intermediate and store chain sets
* Optimizability: the current implementation does a linear search through the trusted set, but having a newtype gives us flexibility needed to optimize lookups in the future (e.g. keying off of subjects)

src/rust/cryptography-x509-validation/src/ops.rs
```diff
@@ -45,10 +45,18 @@ pub(crate) mod tests {
...
     }
 
-    #[test]
-    fn test_nullops() {
-        // Arbitrary relatively small cert (v1_cert.pem from cryptography_vectors).
-        let v1_cert = "
+    #[macro_export]
+    macro_rules! cert {
+        ($pem:literal) => {{
+            let parsed = Box::leak(Box::new(pem::parse($pem).unwrap()));
```
alex, line 51: I have questions. Namely: Why?

woodruffw: Ah, this was a convenience macro while I was writing path validation tests. It's pretty ugly though and there's no repetition here, so I can just drop it.

alex: I don't mind a convenience macro, but there's no need to be leaking things :-)

woodruffw: I'll see if I can work around it, but I _think_ the leak is necessary to get the lifetimes to work -- either a macro or a full function here will have local ownership of the `Pem` that's being borrowed from, so there's no way to get a lifetime shorter than `'static`.

woodruffw: Okay, I've rewritten this as a fixture that returns a `pem::Pem` and a `cert(...)` helper -- that avoids the need to a leak, at the cost of a more explicit lifetime relationship between the two + binding for the `pem::Pem` at each callsite 🙂

alex (approved):

PR #9758, opened 2023-10-24
Add support for ChaCha20 in LibreSSL

facutuesca:
Replaces https://github.com/pyca/cryptography/pull/9209, since LibreSSL does have support for the EVP API for ChaCha20.

### DIFF
--- a/src/cryptography/hazmat/backends/openssl/backend.py
+++ b/src/cryptography/hazmat/backends/openssl/backend.py
@@ -264,8 +264,14 @@ class Backend:
         self.register_cipher_adapter(
             TripleDES, ECB, GetCipherByName("des-ede3")
         )
+        # ChaCha20 uses the Short Name "chacha20" in OpenSSL, but in LibreSSL
+        # it uses "chacha"
         self.register_cipher_adapter(
-            ChaCha20, type(None), GetCipherByName("chacha20")
+            ChaCha20,
+            type(None),
+            GetCipherByName(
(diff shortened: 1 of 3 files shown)


botovq:
Thank you. That looks a lot better :)

I am still going to add a way of allowing the use of chacha20 instead of chacha, but its not entirely trivial, unfortunately.

alex (approved):

botovq:
Managed to figure out the [right incantation][1]. So using 'chacha20' will work starting with LibreSSL 3.9.0.

[1]: https://github.com/libressl/openbsd/commit/fb35c4b4f6b87177cd121d1f1811a2557cc872f8

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -264,8 +264,14 @@ def _register_default_ciphers(self) -> None:
         self.register_cipher_adapter(
             TripleDES, ECB, GetCipherByName("des-ede3")
         )
+        # ChaCha20 uses the Short Name "chacha20" in OpenSSL, but in LibreSSL
+        # it uses "chacha"
```
botovq, line 268: Nit: this is the "Long Name", actually :)

The [Short Name][1] is the capitalized "ChaCha20", or "ChaCha" respectively.

facutuesca: Ah good catch, fixed here: https://github.com/pyca/cryptography/pull/9764

Issue #8956, opened 2023-05-24
BoringSSL parity: ChaCha20


woodruffw:
Tracking breakout from #7747, for ChaCha20.

Expected BoringSSL API: `CRYPTO_chacha_20`

CC @facutuesca 🙂

facutuesca:
In progress

woodruffw:
From email discussions: integrating BoringSSL's ChaCha20 isn't straightforward, since BoringSSL strictly complies with RFC 7539 while LibreSSL and OpenSSL both use the original paper's 64/64 nonce/counter split.

Two possible routes forward:

1. Make a breaking change to the current `ChaCha20` API: turn the current `bytes nonce[16]` into a `bytes nonce[12]`, and internally initialize the counter word to either 0 or 1 (as suggested by the RFC)
  * Pro: This will make the OpenSSL and LibreSSL behave as if they're RFC 7539 compliant, which is how the Cryptography docs state the public API behaves.
  * Pro: This makes it easier to track the counter and fail a block operation if the counter overflows (i.e., after the 2^32 block limit specified in the RFC). Doing so might be overkill and would be a divergence from the current API behavior, so this is strictly optional.
  * Con: This would reduce the internal block count of the OpenSSL and LibreSSL implementations to 2^32, rather than 2^64. This is arguably not a con at all, since it's what the RFC specifies.

2. Retain the current API, but wrap Boring's implementation so that the counter's overflow behavior matches that of LibreSSL and OpenSSL.
  * Pro: API stability?
  * Con: Turns an RFC compliant implementation into a non-complying one; also requires that Cryptography's documentation be updated to emphasize that the `ChaCha20` API doesn't provide an RFC compliant implementation.
  * Con: Adds a lot of complexity for a probably very rare edge case (expecting to share ChaCha20 outputs across implementations).

cc @alex and @reaperhulk for opinions/preferences here.

PR #9762, opened 2023-10-24
Add support for ChaCha20 with BoringSSL

facutuesca:
This PR adds support for ChaCha20 with BoringSSL. Since cryptography's ChaCha20 uses OpenSSL and LibreSSL's implementations (which use a 64 bit counter + 64 bit nonce), and BoringSSL uses a 32 bit counter and 96 bit nonce, we need to adapt Boring's implementation to behave as if it used a 64 bit counter. This is done in the C wrapper.

On the Python side, the only complication is that the current `_CipherContext` implementation assumes that the underlying cipher can be accessed through the `EVP_CIPHER` API. This is not the case for BoringSSL's ChaCha20, which has a separate [CRYPTO_chacha_20()](https://github.com/google/boringssl/blob/master/include/openssl/chacha.h#L29-L40) API for it.

In order to solve that, this PR makes `_CipherContext` an abstract class with two implementations, `_CipherContextEVP` (which is the previous `_CipherContext`) and `_CipherContextChaCha`, which is a simple context that is used only for the BoringSSL+ChaCha20 combination.

The only other thing of note is that BoringSSL's API is stateless, as opposed to OpenSSL's EVP API. This means we need to emulate the statefulness (subsequent calls to the API have memory of previous inputs) in order to get the same results. This is described in detail in the comments of the `_CipherContextChaCha` class.

Fixes https://github.com/pyca/cryptography/issues/8956

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -40,7 +40,7 @@ Changelog
   in favor of the new timezone-aware variants mentioned above.
 * Added support for
   :class:`~cryptography.hazmat.primitives.ciphers.algorithms.ChaCha20`
-  on LibreSSL.
+  on LibreSSL and BoringSSL.
 
 .. _v41-0-4:
 
--- a/src/_cffi_src/build_openssl.py
+++ b/src/_cffi_src/build_openssl.py
@@ -26,6 +26,7 @@ ffi = build_ffi_for_binding(
         "asn1",
         "bignum",
         "bio",
+        "chacha",
         "cmac",
         "crypto",
         "dh",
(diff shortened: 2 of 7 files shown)


alex:
@reaperhulk Would be interested in your take -- I'm concerned that the implementation here is (necessarily) very complex/subtle. When we originalyl made our compatibility list, we didn't support ChaCha20Poly1305 on Boring either, but now that we do, how significant is it to have vanilla-ChaCha20?

reaperhulk:
If we want to switch to boring by default it's an unfortunate gap to not support this -- that said, I'd be comfortable saying we won't fill this gap at the moment and we can always change our minds in the future.

alex:
Thanks for working on this and driving it (and figuring out the LibreSSL issue).

After having reviewed this, I think the complexity here is too high relative to the actual value. Needing to manually buffer everything is very painful and complex, particularly when this is the only cipher that requires it.

AFAIK, most users of ChaCha20 go through the ChaCha20Poly1305 AEAD, which works on BoringSSL.

PR #9800, opened 2023-10-30
X.509: Add WebPKI SPKI AlgorithmIdentifiers

woodruffw:
Similar to the AlgorithmIdentifiers listed for signatures (but with different contents).

I've added canonical encoding tests as well, to ensure that the encodings match the ones specified by CABF.

Refs: CABF 7.1.3.1 (Page 96): <https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-v2.0.0.pdf>

### DIFF
--- a/src/rust/cryptography-x509/src/common.rs
+++ b/src/rust/cryptography-x509/src/common.rs
@@ -47,7 +47,7 @@ pub enum AlgorithmParameters<'a> {
 
     // These encodings are only used in SPKI AlgorithmIdentifiers.
     #[defined_by(oid::EC_OID)]
-    EcNamedCurve(asn1::ObjectIdentifier),
+    Ec(EcParameters<'a>),
 
     #[defined_by(oid::RSA_OID)]
     Rsa(Option<asn1::Null>),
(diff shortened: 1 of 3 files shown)


src/rust/cryptography-x509/src/common.rs
```diff
@@ -45,6 +45,13 @@ pub enum AlgorithmParameters<'a> {
     #[defined_by(oid::ED448_OID)]
     Ed448,
 
+    // These encodings are only used in SPKI AlgorithmIdentifiers.
+    #[defined_by(oid::EC_OID)]
+    EcNamedCurve(asn1::ObjectIdentifier),
+
+    #[defined_by(oid::RSA_OID)]
+    Rsa(Option<asn1::Null>),
```
woodruffw, line 53: I wasn't sure if these were the best names for these, but I couldn't think of better ones in the moment. Happy to change to whatever.

woodruffw:
Ah, looks like adding an explicit parameter form for `id-ecPublicKey` broke some tests (since they were probably using the fallback decoding before). I'll look into this further.

src/rust/cryptography-x509/src/common.rs
```diff
@@ -281,6 +288,21 @@ pub const PSS_SHA512_MASK_GEN_ALG: MaskGenAlgorithm<'_> = MaskGenAlgorithm {
...
+//     -- implicitCurve   NULL
+//     -- specifiedCurve  SpecifiedECDomain }
+//
+// Only the namedCurve form may appear in PKIX. Other forms may be found in
+// other PKIs.
+#[derive(asn1::Asn1Read, asn1::Asn1Write, Hash, Clone, PartialEq, Eq, Debug)]
+pub enum EcParameters<'a> {
+    NamedCurve(asn1::ObjectIdentifier),
+    ImplicitCurve(asn1::Null),
+    SpecifiedCurve(asn1::Tlv<'a>),
```
woodruffw, line 303: NB: I didn't bother to include the `SpecifiedCurve` definition here, since the original fallback logic didn't do it either. If we wanted to, it's somewhere in X9.62.

src/rust/cryptography-x509/src/common.rs
woodruffw: (Renamed `EcNamedCurve` to just `Ec`, due to https://github.com/pyca/cryptography/pull/9800#issuecomment-1785814705.)

src/rust/cryptography-x509/src/common.rs
alex: This should probably be an `asn1::Sequence<'a>` at least, yeah?

woodruffw: Yep, that looks like it should be safe per X9.62 6.3:

```asn1
ECParameters ::= SEQUENCE {
version INTEGER { ecpVer1(1) } (ecpVer1),
fieldID FieldID {{FieldTypes}},
curve Curve,
base ECPoint,
order INTEGER,
cofactor INTEGER OPTIONAL,
...
}
```

(Page 20 of <https://safecurves.cr.yp.to/grouper.ieee.org/groups/1363/private/x9-62-09-20-98.pdf>)

woodruffw: 1a6952dfa..18ad3af8c

alex (approved):

PR #9805, opened 2023-10-31
Add top-level ServerVerifier.verify API

woodruffw:
This is a breakout from #8873, with just the interface/types and a `NotImplementedError` stub.


src/cryptography/hazmat/bindings/_rust/x509.pyi
```diff
@@ -53,6 +53,12 @@ class ServerVerifier:
     def subject(self) -> x509.verification.Subject: ...
     @property
     def validation_time(self) -> datetime.datetime: ...
+    def verify(
+        self,
+        leaf: x509.Certificate,
+        intermediates: list[x509.Certificate],
+        store: Store,
```
alex, line 60: Shouldn't store be on the verifier? Seems weird for it not to be.

But maybe I'm missing a use case?

woodruffw: I think it could go either way -- I can't remember why we didn't do that earlier, if there was a reason. Happy to move it 🙂 

(I'm guessing the idea here would be `PolicyBuilder().store(...).build_server_verifier(...)`?)

woodruffw:
(Twisted downstream failure appears unrelated.)

woodruffw:
GHA is having a bad day today: https://github.com/pyca/cryptography/actions/runs/6711028783/job/18237478253?pr=9805 appears to have run with the not-most-recent commit...

alex (approved):

PR #9844, opened 2023-11-09
src, tests: add max_chain_depth to validation API

woodruffw:
This adds `max_chain_depth` to the Python API, and plumbs it through to the equivalent (previously invariant) setting on the Rust API. It's needed to get the last bit of coverage on #8873, since we'd otherwise be unable to exercise the corresponding Limbo testcases from Python.

Breakout from #8873.

### DIFF
--- a/src/rust/cryptography-x509-validation/src/policy/mod.rs
+++ b/src/rust/cryptography-x509-validation/src/policy/mod.rs
@@ -148,6 +148,17 @@ const RFC5280_CRITICAL_CA_EXTENSIONS: &[asn1::ObjectIdentifier] =
 const RFC5280_CRITICAL_EE_EXTENSIONS: &[asn1::ObjectIdentifier] =
     &[BASIC_CONSTRAINTS_OID, SUBJECT_ALTERNATIVE_NAME_OID];
 
+/// A default reasonable maximum chain depth.
+///
+/// This depth was chosen to balance between common validation lengths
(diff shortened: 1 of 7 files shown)


src/rust/src/x509/verify.rs
```diff
@@ -155,11 +161,16 @@ fn create_server_verifier(
     subject: pyo3::Py<pyo3::PyAny>,
     store: pyo3::Py<PyStore>,
     time: Option<&pyo3::PyAny>,
+    max_chain_depth: Option<&pyo3::types::PyInt>,
```
alex, line 164: You can just use `Option<u8>` here :-)

src/rust/cryptography-x509-validation/src/policy/mod.rs
```diff
@@ -230,10 +230,15 @@ pub struct Policy<'a, B: CryptoOps> {
...
+    pub fn new(
+        ops: B,
+        subject: Option<Subject<'a>>,
+        time: asn1::DateTime,
+        max_chain_depth: Option<u8>,
+    ) -> Self {
         Self {
             _ops: ops,
-            max_chain_depth: 8,
+            max_chain_depth: max_chain_depth.unwrap_or(8),
```
alex, line 241: 8 should be a constant with an explanation of why

woodruffw: 41058a787..61ace70bd

alex (approved):

PR #9886, opened 2023-11-16
Write down another reason we care about MSRV of 1.64

alex:


alex:
Details: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table

reaperhulk (approved):

PR #9890, opened 2023-11-17
validation: flatten error types

woodruffw:
Breakout from #8873: this removes `PolicyError` in favor of the top-level `ValidationError`, which in turn allows us to get rid of a layer of error types.

(These changes don't look like much here, but they end up saving a few dozen lines in #8873.)

### DIFF
--- a/src/rust/cryptography-x509-validation/src/lib.rs
+++ b/src/rust/cryptography-x509-validation/src/lib.rs
@@ -10,3 +10,14 @@ pub mod ops;
 pub mod policy;
 pub mod trust_store;
 pub mod types;
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum ValidationError {
+    Other(String),
+}
+
+impl From<&str> for ValidationError {
+    fn from(value: &str) -> Self {
+        Self::Other(value.into())
+    }
+}
--- a/src/rust/cryptography-x509-validation/src/lib.rs
+++ b/src/rust/cryptography-x509-validation/src/lib.rs
@@ -11,7 +11,6 @@ pub mod policy;
 pub mod trust_store;
 pub mod types;
 
-#[derive(Debug, PartialEq, Eq)]
 pub enum ValidationError {
     Other(&'static str),
 }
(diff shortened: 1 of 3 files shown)


src/rust/cryptography-x509-validation/src/lib.rs
```diff
@@ -10,3 +10,14 @@ pub mod ops;
 pub mod policy;
 pub mod trust_store;
 pub mod types;
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum ValidationError {
+    Other(String),
```
woodruffw, line 16: NB: Owned rather than `&'static` here because some errors in #8873 will ultimately be "rendered," e.g. including OIDs for critical extensions that aren't handled.

src/rust/cryptography-x509-validation/src/lib.rs
```diff
@@ -10,3 +10,14 @@ pub mod ops;
...
+}
+
+impl From<&str> for ValidationError {
```
alex, line 19: Not a fan of `From<&str>` it's way too non-specific and makes the code hard to read

src/rust/cryptography-x509-validation/src/lib.rs
alex, line 16: Also looks like `&'static str` should be sufficient here?

src/rust/cryptography-x509-validation/src/lib.rs
woodruffw: Ok, removing.

src/rust/cryptography-x509-validation/src/lib.rs
woodruffw: Left a note about that here: https://github.com/pyca/cryptography/pull/9890#discussion_r1397981880

(I could do two variants, not sure how much better that is? Whatever your preference is 🙂)

alex: Let'd do `&'static str` for now, so this PR is just about changing the hierarchy, and we can do a `String` in future.

alex (approved):

PR #9947, opened 2023-12-01
raise an exception instead of returning an empty list for pkcs7 cert loading

reaperhulk:
as davidben points out in #9926 we are calling a load certificates function and an empty value doesn't necessarily mean empty because PKCS7 contains multitudes. erroring is more correct.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,7 @@ Changelog
   :func:`~cryptography.hazmat.primitives.serialization.pkcs7.load_pem_pkcs7_certificates`
   or
   :func:`~cryptography.hazmat.primitives.serialization.pkcs7.load_der_pkcs7_certificates`
-  will now raise an exception rather than return an empty list.
+  will now raise a ``ValueError`` rather than return an empty list.
 * Parsing SSH certificates no longer permits malformed critical options with
   values, as documented in the 41.0.2 release notes.
 * Updated the minimum supported Rust version (MSRV) to 1.63.0, from 1.56.0.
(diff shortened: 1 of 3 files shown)


alex (commented):
changelog

CHANGELOG.rst
```diff
@@ -9,6 +9,11 @@ Changelog
 .. note:: This version is not yet released and is under active development.
 
 * **BACKWARDS INCOMPATIBLE:** Dropped support for LibreSSL < 3.7.
+* **BACKWARDS INCOMPATIBLE:** Loading a PKCS7 with no certificates using
+  :func:`~cryptography.hazmat.primitives.serialization.pkcs7.load_pem_pkcs7_certificates`
+  or
+  :func:`~cryptography.hazmat.primitives.serialization.pkcs7.load_der_pkcs7_certificates`
+  will now raise an exception rather than return an empty list.
```
alex, line 16: It's not _no_ certificates, it's specifically if something is null, not if it's an empty list.

CHANGELOG.rst
```diff
@@ -9,6 +9,11 @@ Changelog
 .. note:: This version is not yet released and is under active development.
 
 * **BACKWARDS INCOMPATIBLE:** Dropped support for LibreSSL < 3.7.
+* **BACKWARDS INCOMPATIBLE:** Loading a PKCS7 with no content field using
+  :func:`~cryptography.hazmat.primitives.serialization.pkcs7.load_pem_pkcs7_certificates`
+  or
+  :func:`~cryptography.hazmat.primitives.serialization.pkcs7.load_der_pkcs7_certificates`
+  will now raise an exception rather than return an empty list.
```
alex, line 16:
```suggestion
  will now raise a ``ValueError`` rather than return an empty list.
```

alex (approved):

PR #9951, opened 2023-12-02
we'll ship 3.2.0 for 42

reaperhulk:
but we won't be merging the infra PR until we know we don't need to do another 41.0.x


alex (dismissed):
Approving, but not merging until we do the infra merge

alex (approved):

Issue #8675, opened 2023-04-06
[feature request] take back ssl binding "SSL_CTX_set_cert_store"


SeaHOH:
I am using `SSL_CTX_set_cert_store` which allow me could maintain only one strore. This is very convenient and functional, but removed from v40. Please, could you take it back to new releases?

reaperhulk:
Do you use this in conjunction with pyopenssl? If so, it would be desirable to add it as a feature there and use the binding from pyopenssl. That way we understand how it’s being used, can test for it, etc.

SeaHOH:
@reaperhulk Yes, I use it with pyOpenSSL. And, you suggestion me make a feature/pull request to pyOpenSSL project (not its dependency). Do I understand correct? If so, could you transport this issue to pyOpenSSL repository? I will change the title then.

SeaHOH:
> This is a multi-step process, so we'll need work on both sides:
> ...

So, we donot need to transport this issue. OK, let's go on.

I think a context configurable cert store will be most useful. There is already `SSL.Context.get_cert_store` here, why not a `SSL.Context.set_cert_store`?

mhils:
FWIW for @mitmproxy we're reusing one SSL.Context for all client connections. `SSL.Connection` nowadays has `use_certificate` and `use_privatekey` methods which make that possible. Not sure if that would be an alternative in your case. :)

SeaHOH:
@mhils Thanks for your suggestion. But I think session reuse will not be work between different servers in your case.

reaperhulk:
Having an SSL.Context.set_cert_store seems reasonable to me. Let's assume that's what we're going to do. Do you have time to implement that on the pyOpenSSL side? As Alex mentioned we need to add the binding here and then add the API in pyOpenSSL. Complicating matters a bit is that we also need to fix a pyOpenSSL test issue, but we can work around that during dev. We'll also have to have an extra branch in `set_cert_store` to handle the versions that don't bind this function until such time as we require >=40.0.3 (or whatever version we release the re-added binding in). This is being done in https://github.com/pyca/pyopenssl/pull/1202 as well.

SeaHOH:
> Do you have time to implement that on the pyOpenSSL side?

Me? I can write the funticon code, but I am not good at English as you see, I cannot use accurate/seemly wording into the code comment. And I am not a programer, write unit tests is hard to me.

reaperhulk:
[...]

PR #9978, opened 2023-12-09
#9969 Expose SSL_session_reused.

### LINKED ISSUE #9969: SSL_session_reused bindings not available
Hi,

I might be missing something here.

I see that `SSL_session_reused` is defined here, but I can't acess it via the bindings

https://github.com/pyca/cryptography/blob/f69f27b1dd20ad2d24f48053a72545527e808104/src/_cffi_src/openssl/ssl.py#L410-L412

I am using Ubuntu 23.10 , with default py3 and latest cryptography installed in a virtualenv

```
$ virtualenv venv
$ . venv/bin/activate
$ pip install cryptography
$ python
Python 3.11.6 (main, Oct  8 2023, 05:06:43) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from cryptography.hazmat.bindings.openssl.binding import Binding
>>> binding = Binding()
>>> binding.init_static_locks()

>>> binding.lib.SSL_session_reused
AttributeError: module 'lib' has no attribute 'SSL_session_reused'

>>> binding.lib.SSL_select_next_proto
AttributeError: module 'lib' has no attribute 'SSL_select_next_proto'

>>> binding.lib.SSL_CTX_set_alpn_protos
<built-in method SSL_CTX_set_alpn_protos of _cffi_backend.Lib object at 0x7fe38561bab0>
>>> 
```

```
$ python -m pip freeze
cffi==1.16.0
cryptography==41.0.7
pycparser==2.21
```

--------

I don't know why `SSL_CTX_set_alpn_protos` is available but `SSL_session_reused` and other functions are not available.

This was working in `cryptography==39.0.2`

I guess that this might be part of 40.0.0 

> Removed many unused CFFI OpenSSL bindings. This will not impact you unless you are using cryptography to directly invoke OpenSSL's C API. Note that these have never been considered a stable, supported, public API by cryptography, this note is included as a courtesy.

--------

I think that `SSL_session_reused` is very important as it is critical for implemeting FTPS server-side session reusal validation.

The `SSL_set_session` that is already avaiable, is used for the FTPS client-side part.

----------

Thanks

adiroiban:
Fixes #9969 

This (re)adds the `SSL_session_reused` API.

I think that this API is criticial.

First, this complements the already supported `int SSL_set_session(SSL *, SSL_SESSION *);
` to allow you to verify that after the handshake the session was reused.

Second, when implementing an FTPS server, this API is the only way to make sure you don't end up with hijacked FTP data connection.

I only added a simple tests to check that the API is there.

Let me know if more tests are required.

Thanks

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -142,6 +142,7 @@ FUNCTIONS = """
 const char *SSL_state_string_long(const SSL *);
 SSL_SESSION *SSL_get1_session(SSL *);
 int SSL_set_session(SSL *, SSL_SESSION *);
+int SSL_session_reused(const SSL *ssl);
 SSL *SSL_new(SSL_CTX *);
 void SSL_free(SSL *);
 int SSL_set_fd(SSL *, int);
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -142,7 +142,7 @@ FUNCTIONS = """
 const char *SSL_state_string_long(const SSL *);
 SSL_SESSION *SSL_get1_session(SSL *);
 int SSL_set_session(SSL *, SSL_SESSION *);
-int SSL_session_reused(const SSL *ssl);
+int SSL_session_reused(const SSL *);
 SSL *SSL_new(SSL_CTX *);
 void SSL_free(SSL *);
 int SSL_set_fd(SSL *, int);
--- a/tests/hazmat/bindings/test_openssl.py
+++ b/tests/hazmat/bindings/test_openssl.py
@@ -112,16 +112,3 @@ class TestOpenSSL:
         assert error.reason == b.lib.EVP_R_DATA_NOT_MULTIPLE_OF_BLOCK_LENGTH
         if not b.lib.CRYPTOGRAPHY_IS_BORINGSSL:
             assert b"data not multiple of block length" in error.reason_text
-
-    def test_ssl_session_reused_new(self):
-        """
-        The SSL_session_reused API is available
-        and return 0 for a new connection.
-        """
-        b = Binding()
-        ctx = b.lib.SSL_CTX_new(b.lib.TLS_method())
-        ctx = b.ffi.gc(ctx, b.lib.SSL_CTX_free)
-        ssl = b.lib.SSL_new(ctx)
-        ssl = b.ffi.gc(ssl, b.lib.SSL_free)
-
-        assert b.lib.SSL_session_reused(ssl) == 0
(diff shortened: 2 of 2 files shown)


adiroiban:
This is read for review. Thanks!

adiroiban:
I now see https://github.com/pyca/cryptography/issues/8675 

There are 2 things here: session cache vs session reuse.

The server/client can have session cache disabled and still reuse the session.

For example a session is initiated between server and client via connection A.

The client can initiate a new connection B to the server and setup the connection B via `session_a = SSL_get1_session(connection_a) -> SSL_set_session(connection_b, session_a)`

tests/hazmat/bindings/test_openssl.py
```diff
@@ -112,3 +112,16 @@ def test_rust_internal_error(self):
         assert error.reason == b.lib.EVP_R_DATA_NOT_MULTIPLE_OF_BLOCK_LENGTH
         if not b.lib.CRYPTOGRAPHY_IS_BORINGSSL:
             assert b"data not multiple of block length" in error.reason_text
+
+    def test_ssl_session_reused_new(self):
```
alex, line 116: No need for a test, we don't add them for all bindings.

src/_cffi_src/openssl/ssl.py
```diff
@@ -142,6 +142,7 @@
 const char *SSL_state_string_long(const SSL *);
 SSL_SESSION *SSL_get1_session(SSL *);
 int SSL_set_session(SSL *, SSL_SESSION *);
+int SSL_session_reused(const SSL *ssl);
```
alex, line 145:
```suggestion
int SSL_session_reused(const SSL *);
```

tests/hazmat/bindings/test_openssl.py
adiroiban: Is this a request to remove the test?

I think that a test can help with any accedental removal of the API.

It can also serve as an example on how to use the API.

alex: Yes, please remove the test.

adiroiban:
I have removed the test. I was not expecting a request to submit a PR without tests.

I don't understand why is bad to have that test.

The test was kind a reminder that the binding in important and it should not be removed at the next "great purge".

No hard feelings. I am happy to have the new API available with or without tests.
I have automated tests for this API in my project, so that's fine be me.

Thanks for maintaining `cryptography`. Much appreciated.

alex (approved):
The reason we don't add tests for bindings is that they're not actually testing anything about this project, except that we bound the function correctly -- but that's handled in other ways.

We'll know not to remove this because a) it'll be used in pyOpenSSL (which we test in CI), b) we check the commit history before removing something.

PR #10044, opened 2023-12-23
add automatic PRs for new commits on x509-limbo and wycheproof

reaperhulk:
will this work? let's find out.

we could also combine this with the boring/openssl bump but ¯\_(ツ)_/¯


reaperhulk:
The refs here are both one out of date so we can see it work immediately...post merge.

alex (approved):

PR #10054, opened 2023-12-23
Make extension handling in x.509 verifier less meta-programmed

alex:
We now iterate over the extensions only once.

### DIFF
--- a/src/rust/cryptography-x509-verification/src/policy/mod.rs
+++ b/src/rust/cryptography-x509-verification/src/policy/mod.rs
@@ -17,15 +17,12 @@ use cryptography_x509::common::{
 use cryptography_x509::extensions::{BasicConstraints, Extensions, SubjectAlternativeName};
 use cryptography_x509::name::GeneralName;
 use cryptography_x509::oid::{
-    AUTHORITY_INFORMATION_ACCESS_OID, AUTHORITY_KEY_IDENTIFIER_OID, BASIC_CONSTRAINTS_OID,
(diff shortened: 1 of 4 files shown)


src/rust/cryptography-x509-verification/src/policy/mod.rs
```diff
@@ -247,103 +243,96 @@ impl<'a, B: CryptoOps> Policy<'a, B> {
...
                 // NOTE: MUST be critical in 5280, but CABF relaxes to MAY.
-                ExtensionPolicy::maybe_present(
-                    NAME_CONSTRAINTS_OID,
+                name_constraints: ExtensionValidator::maybe_present(
                     Criticality::Agnostic,
                     Some(ca::name_constraints),
                 ),
                 // 5280 4.2.1.11: Policy Constraints
-                ExtensionPolicy::maybe_present(POLICY_CONSTRAINTS_OID, Criticality::Critical, None),
+                policy_constraints: ExtensionValidator::maybe_present(Criticality::Critical, None),
```
woodruffw, line 282: Thinking about it a bit, this should also probably have a validator function that rejects this extension outright -- 5280 says that it's always critical (which we check) which means that we must handle it, which we intentionally don't.

woodruffw: (I'll build a testcase that'll work for that.)

reaperhulk (approved):

PR #10058, opened 2023-12-25
Keep the original DER for an SPKI around

alex:
This lets us parse it without needing to re-serialize.

Eventually we can extend this to TBS data itself.


alex:
I don't like the name raw content, it's what Go uses for this https://pkg.go.dev/encoding/asn1#RawContent but to me it's a bit vague.

alex:
Solved the naming problem.

reaperhulk (approved):

PR #10064, opened 2023-12-25
Migrate private key parsing to Rust

alex:
It's still OpenSSL, but now there's more Rust

### DIFF
--- a/src/cryptography/hazmat/primitives/serialization/base.py
+++ b/src/cryptography/hazmat/primitives/serialization/base.py
@@ -21,10 +21,10 @@ def load_pem_private_key(
     *,
     unsafe_skip_rsa_key_validation: bool = False,
 ) -> PrivateKeyTypes:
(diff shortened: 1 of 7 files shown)


src/cryptography/hazmat/primitives/serialization/base.py
```diff
@@ -21,10 +21,10 @@ def load_pem_private_key(
     *,
     unsafe_skip_rsa_key_validation: bool = False,
 ) -> PrivateKeyTypes:
```
reaperhulk, line 23: Is there any reason to keep the wrapper functions here rather than simply doing `load_pem_private_key = rust_openssl.keys.load_pem_private_key`?

alex:
It probably could, there's a lot of places that's true I think

reaperhulk:
Okay, I'll file an issue to follow up.

reaperhulk (approved):

Issue #10034, opened 2023-12-22
X.509 path building follow-ups


alex:
```[tasklist]
### For 42.0
42. [ ] https://github.com/pyca/cryptography/pull/10035
43. [ ] https://github.com/pyca/cryptography/pull/10036
44. [ ] https://github.com/pyca/cryptography/pull/10039
45. [ ] https://github.com/pyca/cryptography/pull/10229
```

```[tasklist]
### Functionality
- [ ] https://github.com/pyca/cryptography/pull/10044
- [ ] https://github.com/pyca/cryptography/pull/10042
- [ ] https://github.com/pyca/cryptography/issues/10276
- [ ] Relaxing places where we're too strict relative to other implementations
- [ ] Revocation checking (CRL, OCSP)
```

```[tasklist]
### Performance
- [ ] https://github.com/pyca/cryptography/pull/10050
- [ ] https://github.com/pyca/cryptography/pull/10047
- [ ] https://github.com/pyca/cryptography/pull/10051
- [ ] https://github.com/pyca/cryptography/pull/10052
- [ ] https://github.com/pyca/cryptography/pull/10056
- [ ] https://github.com/pyca/cryptography/pull/10054
- [ ] https://github.com/pyca/cryptography/pull/10100
- [ ] https://github.com/pyca/cryptography/pull/10069
- [ ] https://github.com/pyca/cryptography/pull/10092
- [ ] https://github.com/pyca/cryptography/pull/10121
- [ ] Use `WithTlv` on `tbs_cert` to avoid allocations
- [ ] https://github.com/pyca/cryptography/pull/10217
- [ ] https://github.com/pyca/cryptography/pull/10218
```

woodruffw:
Another functionality follow-up: CRLs and CRL checking. This will likely require its own non-trivial design and planning period.

reaperhulk:
Added as Revocation (CRL, OCSP) 👍

alex:
Closing in favor of individual issues

PR #10069, opened 2023-12-27
Migrate PolicyBuilder to Rust

alex:


src/rust/src/x509/verify.rs
```diff
@@ -46,6 +43,121 @@ impl CryptoOps for PyCryptoOps {
     }
 }
 
+pyo3::create_exception!(
+    cryptography.hazmat.bindings._rust.x509,
+    VerificationError,
+    pyo3::exceptions::PyException
+);
+
+#[pyo3::pyclass(frozen, module = "cryptography.hazmat.bindings._rust.x509")]
```
reaperhulk, line 52: module path

alex:
FWIW with respect to the naive datetime question: I don't believe this PR changes our behavior, and so it shouldn't block this PR. I have added that to the TODO list in https://github.com/pyca/cryptography/issues/10034 for us to ensure we're happy with the behavior before we release.

reaperhulk (approved):

PR #10071, opened 2023-12-27
Format all the imports with

alex:
https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#group_imports using a value of `StdExternalCrate`

This option isn't stable, so we can't enforce it in CI, which is annoying.


alex:
This probably conflicts with everything, so it should be merged last :-)

reaperhulk (approved):

PR #7503, opened 2022-08-13
try `EVP_CIPHER_fetch` due to `EVP_get_cipherbyname` is deprecated

boyxuper:
see: https://www.openssl.org/docs/manmaster/man3/EVP_EncryptInit.html#EVP_get_cipherbyname-EVP_get_cipherbynid-and-EVP_get_cipherbyobj
see: https://github.com/openssl/openssl/issues/13667#issuecomment-1034505341

use case:
obtain sm4-gcm / sm4-ccm with OpenSSL 3.1.0 (master branch)

PR #10072, opened 2023-12-27
Add support for GCM mode of SM4 cipher

mnixry:
ref: #7503
ref: https://github.com/openssl/openssl/issues/13667

[RFC8998](https://datatracker.ietf.org/doc/html/rfc8998), SM4-GCM support, has been added into OpenSSL for a year. 
It will be nice if we can add support to this package.


reaperhulk:
Thanks for the contribution. A few questions/requests:

* Could you format the test vectors as we do for our other SM4 test vectors? (see: https://github.com/pyca/cryptography/blob/main/vectors/cryptography_vectors/ciphers/SM4/draft-ribose-cfrg-sm4-10-cbc.txt). You can name the file `rfc8998.txt` in that directory. Please make sure you also add the information about the vectors to `test-vectors.rst` under the Symmetric Ciphers section.
* I'd like to see some tests for the negative case (`InvalidTag`) as well.
* Is SM4-GCM used outside of this TLS cipher suite? Are Chinese browsers implementing this suite? Understanding the use case is always helpful 😄

mnixry:
Thank you for your suggestions! Tests has been update according to request.

> Is SM4-GCM used outside of this TLS cipher suite? Are Chinese browsers implementing this suite?

Yes. AFAIK SM2/3/4 supported browser (and certificate for hosting website) has been a part of compliance requirement of some highly confidential scenario. 
Some VPN, for accessing internal network, are also adapting SM4-GCM as their encryption approaches.

reaperhulk:
This mostly looks good, but two final requests 😄 

* Could you add a positive and negative test case using `finalize_with_tag`?
* Please add an entry to `CHANGELOG.rst` (something along the lines of "added support for SM4 GCM when using OpenSSL 3.0 or greater")

reaperhulk (approved):

reaperhulk:
Thanks again!

PR #10079, opened 2023-12-28
Migrate EC Numbers to Rust

alex:


src/rust/src/backend/ec.rs
```diff
@@ -501,17 +422,226 @@ impl ECPublicKey {
...
+            .hash(&mut hasher);
+        Ok(hasher.finish())
+    }
+
+    fn __repr__(&self, py: pyo3::Python<'_>) -> pyo3::PyResult<String> {
+        let x = self.x.as_ref(py);
+        let y = self.y.as_ref(py);
+        let curve_name = self.curve.as_ref(py).getattr(pyo3::intern!(py, "name"))?;
+        Ok(format!(
+            "<EllipticCurvePublicNumbers(curve={curve_name}, x={x}, y={y}>"
```
alex, line 629: TODO: Fix typo, misssing `)`

src/rust/src/backend/ec.rs
```diff
@@ -501,17 +422,226 @@ impl ECPublicKey {
     }
 }
 
+#[pyo3::prelude::pyclass(frozen, module = "cryptography.hazmat.bindings._rust.openssl.ec")]
```
reaperhulk, line 425: We've been generally doing this for rust objects, but should we actually claim the module is `cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePrivateNumbers` since that's the only public place we expose this class now?

reaperhulk (approved):

PR #10087, opened 2023-12-30
Small refactor of cipher registry

alex:
Makes it easy to add support for owned ciphers


reaperhulk (approved):
I just realized the builder is named “Registery” and not Registry 🤔

PR #10094, opened 2023-12-31
Parse PEM public keys in Rust

alex:


alex:
It occurs to me that this allows arbitrary PEM headers. What are the correct ones?

reaperhulk (approved):

Issue #10118, opened 2024-01-03
Remove unneeded indirection


reaperhulk:
As we've been pulling things into rust we have sometimes left unneeded wrapper functions. Let's go clean some of those up. (Example: https://github.com/pyca/cryptography/pull/10064)

PR #10119, opened 2024-01-03
remove indirection in pem/der loaders, use the rust funcs directly

reaperhulk:


reaperhulk:
refs #10118 but there may still be more 😄

alex (approved):

PR #10121, opened 2024-01-04
Migrate SPKI parsing from OpenSSL to Rust

alex:
TODO:

- [x] Handle unsupported key types
- [x] Handle un-named EC keys
- [x] Figure out how to handle DHX stuff (`DH_set_flags(dh, DH_FLAG_TYPE_DHX)`)
- [x] More EC curves?
- [x] rust-openssl release
- [x] Figure out the RHEL issue
- [x] Missing EC coverage
  - [x] Private key with explicit curve parameters
  - [x] Private key with an unsupported curve
  - [x] Loading keys with obscure curves:
    - [x] SECT163K1
    - [x] SECT233K1
    - [x] SECT163R2
    - [x] SECT233R1

### DIFF
--- a/src/rust/cryptography-key-parsing/src/spki.rs
+++ b/src/rust/cryptography-key-parsing/src/spki.rs
@@ -1,0 +1,142 @@
+// This file is dual licensed under the terms of the Apache License, Version
+// 2.0, and the BSD License. See the LICENSE file in the root of this repository
+// for complete details.
+
+use cryptography_x509::common::{AlgorithmParameters, EcParameters, SubjectPublicKeyInfo};
+
+use crate::{KeyParsingError, KeyParsingResult};
+
+pub fn parse_public_key(
+    data: &[u8],
+) -> KeyParsingResult<openssl::pkey::PKey<openssl::pkey::Public>> {
+    let k = asn1::parse_single::<SubjectPublicKeyInfo>(data)?;
+
+    match k.algorithm.params {
+        AlgorithmParameters::Ec(ec_params) => match ec_params {
+            EcParameters::NamedCurve(curve_oid) => {
+                let curve_nid = match curve_oid {
+                    cryptography_x509::oid::EC_SECP192R1 => openssl::nid::Nid::X9_62_PRIME192V1,
+                    cryptography_x509::oid::EC_SECP224R1 => openssl::nid::Nid::SECP224R1,
+                    cryptography_x509::oid::EC_SECP256R1 => openssl::nid::Nid::X9_62_PRIME256V1,
+                    cryptography_x509::oid::EC_SECP384R1 => openssl::nid::Nid::SECP384R1,
+                    cryptography_x509::oid::EC_SECP521R1 => openssl::nid::Nid::SECP521R1,
+
+                    cryptography_x509::oid::EC_SECP256K1 => openssl::nid::Nid::SECP256K1,
+
+                    cryptography_x509::oid::EC_SECT233R1 => openssl::nid::Nid::SECT233R1,
+                    cryptography_x509::oid::EC_SECT283R1 => openssl::nid::Nid::SECT283R1,
+                    cryptography_x509::oid::EC_SECT409R1 => openssl::nid::Nid::SECT409R1,
+                    cryptography_x509::oid::EC_SECT571R1 => openssl::nid::Nid::SECT571R1,
+
+                    cryptography_x509::oid::EC_SECT163R2 => openssl::nid::Nid::SECT163R2,
+
+                    cryptography_x509::oid::EC_SECT163K1 => openssl::nid::Nid::SECT163K1,
+                    cryptography_x509::oid::EC_SECT233K1 => openssl::nid::Nid::SECT233K1,
+                    cryptography_x509::oid::EC_SECT283K1 => openssl::nid::Nid::SECT283K1,
+                    cryptography_x509::oid::EC_SECT409K1 => openssl::nid::Nid::SECT409K1,
+                    cryptography_x509::oid::EC_SECT571K1 => openssl::nid::Nid::SECT571K1,
+
+                    #[cfg(not(CRYPTOGRAPHY_IS_BORINGSSL))]
+                    cryptography_x509::oid::EC_BRAINPOOLP256R1 => {
+                        openssl::nid::Nid::BRAINPOOL_P256R1
+                    }
+                    #[cfg(not(CRYPTOGRAPHY_IS_BORINGSSL))]
+                    cryptography_x509::oid::EC_BRAINPOOLP384R1 => {
+                        openssl::nid::Nid::BRAINPOOL_P384R1
+                    }
+                    #[cfg(not(CRYPTOGRAPHY_IS_BORINGSSL))]
+                    cryptography_x509::oid::EC_BRAINPOOLP512R1 => {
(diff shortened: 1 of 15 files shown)


src/rust/cryptography-x509/src/common.rs
```diff
@@ -112,6 +120,11 @@ pub enum AlgorithmParameters<'a> {
     #[defined_by(oid::DSA_WITH_SHA512_OID)]
     DsaWithSha512(Option<asn1::Null>),
 
+    #[defined_by(oid::DH_OID)]
+    Dh(DHParams<'a>),
+    #[defined_by(oid::DH_KEY_AGREEMENT_OID)]
+    DhKeyAgreement(DHParams<'a>),
```
alex, line 126: I am not confident either of these are right, I'm not sure how to verify further.

tests/x509/test_x509_ext.py
```diff
@@ -1758,22 +1758,6 @@ def test_invalid_bit_string_padding_from_public_key(self, backend):
         with pytest.raises(ValueError, match="Invalid public key encoding"):
             _key_identifier_from_public_key(pretend_key)
 
-    def test_no_optional_params_allowed_from_public_key(self, backend):
```
alex, line 1761: I didn't understand what this was trying to test _at all_, and now loading a DSA key with broken parameters is an error.

reaperhulk: Looks like we added it in 2019 to fix coverage of a conditional branch that no longer exists.

alex: Oh that would have been back with the old pure python DER parser.

src/rust/cryptography-key-parsing/src/spki.rs
```diff
@@ -0,0 +1,137 @@
...
+
+                    cryptography_x509::oid::EC_SECT163R2 => openssl::nid::Nid::SECT163R2,
+
+                    cryptography_x509::oid::EC_SECT163K1 => openssl::nid::Nid::SECT163K1,
+                    cryptography_x509::oid::EC_SECT233K1 => openssl::nid::Nid::SECT233K1,
+                    cryptography_x509::oid::EC_SECT283K1 => openssl::nid::Nid::SECT283K1,
+                    cryptography_x509::oid::EC_SECT409K1 => openssl::nid::Nid::SECT409K1,
+                    cryptography_x509::oid::EC_SECT571K1 => openssl::nid::Nid::SECT571K1,
+
+                    #[cfg(not(any(CRYPTOGRAPHY_IS_LIBRESSL, CRYPTOGRAPHY_IS_BORINGSSL)))]
```
reaperhulk, line 39: Does libressl really not support brainpool?

alex: Hmm, now I'm not sure. The same branch exists for them in ec.rs. Let me do a separate PR removing that.

alex: Ugh, it's a rust-openssl issue. Let me fix.

alex: https://github.com/sfackler/rust-openssl/pull/2150

src/rust/cryptography-x509/src/common.rs
```diff
@@ -327,6 +340,13 @@ pub struct RsaPssParameters<'a> {
     pub _trailer_field: u8,
 }
 
+#[derive(asn1::Asn1Read, asn1::Asn1Write, Hash, Clone, PartialEq, Eq, Debug)]
+pub struct DssParms<'a> {
```
reaperhulk, line 344: A part of me is tempted to leave this typo for the future parmesan joke, but we should probably make it say `DssParams`

reaperhulk: 🤯 We should probably match the RFC so I guess let's not change it. Maybe a comment that links to that so I don't have this conversation with you every 6 months 😆

alex: Already changed the name and added a comment.

src/rust/cryptography-x509/src/common.rs
```diff
@@ -235,6 +248,24 @@ pub struct DHParams<'a> {
     pub g: asn1::BigUint<'a>,
     pub q: Option<asn1::BigUint<'a>>,
 }
+
+// Based on OpenSSL's `DHparams`
+#[derive(asn1::Asn1Read, asn1::Asn1Write, Clone, PartialEq, Eq, Debug, Hash)]
+pub struct BasicDHParams<'a> {
+    pub p: asn1::BigUint<'a>,
+    pub g: asn1::BigUint<'a>,
+    // No idea what this is for/what it means.
```
botovq, line 257: This is the bit length of the private value, see PKCS#3, section 6 and 7.1. The central authority could optionally specify that `x` be selected of this length, in addition to `0 < x < p-1`

alex: Oh perfect, finally have a citation for the underlying sequence structure. Perfect.

alex: Thank you!

src/rust/cryptography-x509/src/common.rs
```diff
@@ -235,6 +248,28 @@ pub struct DHParams<'a> {
...
+    pub g: asn1::BigUint<'a>,
+    pub private_value_length: Option<u32>,
+}
+
+// Based on OpenSSL's `DHxparams`
+#[derive(asn1::Asn1Read, asn1::Asn1Write, Clone, PartialEq, Eq, Debug, Hash)]
+pub struct DHXParams<'a> {
+    pub p: asn1::BigUint<'a>,
+    pub g: asn1::BigUint<'a>,
+    pub q: asn1::BigUint<'a>,
```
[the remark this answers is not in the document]
alex: First, thank you! I guess I'll add j and `validation_params`, even though there's no evidence of their use.

How are you finding the sources for these ASN.1 definitions? It looked to me that this was defined in an ANSI spec and I decided I didn't want to spend 100 francs.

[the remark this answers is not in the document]
alex: I haven't seen a single test vector using either of the extension fields, so I'm going to ignore them for now :-)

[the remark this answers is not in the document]
alex: Your comments have been super helpful here! Can I ask if you have a particular interest in this PR, or just being helpful :-)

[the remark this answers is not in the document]
botovq: Pure curiosity what you needed the Brainpool NIDs for. Then I saw the comment indicating a potential gap, and since I happened to know the answer I pointed that out, then I saw that you wanted to document where the ASN.1 came from so I thought I could figure that out as well. That's all.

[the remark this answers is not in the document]
alex: Much appreciated!

reaperhulk (approved):

tamird:
This seems to be stricter than OpenSSL. In particular, some keys with `BEGIN RSA PUBLIC KEY` which previously decoded without error no longer do until changed to `BEGIN PUBLIC KEY`.

tamird:
It looks like the key was indeed not correct, passing it through `openssl rsa -pubin -pubout` strips "RSA" from the header and footer.

Just flagging that this is a behavior change.

PR #10169, opened 2024-01-12
docs/x509: fix verification example

woodruffw:
Fixes a few of my own mistakes in this example, and makes it complete (with a peer and intermediates).

(Manually ran it locally and confirmed that it works, as well.)

### DIFF
--- a/docs/x509/verification.rst
+++ b/docs/x509/verification.rst
@@ -15,12 +15,9 @@ or chain building.
 Example usage, with `certifi <https://pypi.org/project/certifi/>`_ providing
 the root of trust:
 
-.. code-block:: python
+.. testsetup::
 
-    from cryptography.x509 import Certificate, DNSName, load_pem_x509_certificate, load_pem_x509_certificates
-    from cryptography.x509.verification import PolicyBuilder, Store
-    import certifi
-
+    from cryptography.x509 import load_pem_x509_certificate, load_pem_x509_certificates
     from datetime import datetime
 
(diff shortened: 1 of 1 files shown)


docs/x509/verification.rst
```diff
@@ -17,15 +17,63 @@ the root of trust:
 
 .. code-block:: python
```
woodruffw, line 18: I think I could make this into `.. doctest::` instead, if desired.

docs/x509/verification.rst
```diff
@@ -17,15 +17,63 @@ the root of trust:
...
 
-    from cryptography.x509 import Certificate, DNSName, load_pem_x509_certificates
+    from cryptography.x509 import Certificate, DNSName, load_pem_x509_certificate, load_pem_x509_certificates
     from cryptography.x509.verification import PolicyBuilder, Store
     import certifi
 
+    from datetime import datetime
```
alex, line 24: Hmm, this makes it super verbose. Is there some way for us to make clear what perr and untrusted are without making the example 200 lines long

woodruffw: If you're okay with it going back to being non-functional, I could probably do something like `peer, intermediates = load_pem_x509_certificates(open("whatever.pem").read())`. WDYT?

alex: Hmm, so using `doctest` + `testsetup` and then adding a comment explaining what they are would be my suggestion

alex (approved):

PR #10174, opened 2024-01-13
Use flit instead of setuptools for vectors

alex:
Its much faster


reaperhulk:
Do we need to modify the build requirements pins for this for when we build and publish?

alex:
Hmm, it seems like we _should_ need to, but nothing is failing?

alex:
Yeah, looks like it's not required.

alex:
`build-requirements.txt` isn't used when we build the sdists, and that's the only place we do anything with vectors.

reaperhulk (approved):

PR #10188, opened 2024-01-15
fixes #10187 -- handle passing a curve class when generating an EC key

### LINKED ISSUE #10187: Breaking change in type of a private elliptic key's "curve" from instance-of-class to class
While trying to test cryptography 42 changes related to UTC times going from "naive" to "timezone aware" in aioquic, I encountered an unrelated failure from our tests.

The problem was that for a SECP256R1 private key, we had code testing `isinstance(key, ec.SECP256R1)` and this is `True` in 41.0.7 and `False` in the development mainline.  I built from 00a94b00 this morning for testing with python 3.12, and I built with "python -m build".  I then installed the wheel into my aioquic test environment.

I did not see this change listed in the list of breaking changes and figure it might have snuck in unintentionally.

Here is a comparison between 41.0.7 and main:

```
# cryptography 41.0.7

>>> from cryptography.hazmat.primitives.asymmetric import ec, ed448, ed25519, rsa
>>> key = ec.generate_private_key(ec.SECP256R1)
>>> key.curve 
<cryptography.hazmat.primitives.asymmetric.ec.SECP256R1 object at 0x102da2840>
>>> isinstance(key.curve, ec.SECP256R1)
True
>>> key.curve is ec.SECP256R1  
False


# cryptography main

>>> from cryptography.hazmat.primitives.asymmetric import ec, ed448, ed25519, rsa
>>> key = ec.generate_private_key(ec.SECP256R1)
>>> key.curve
<class 'cryptography.hazmat.primitives.asymmetric.ec.SECP256R1'>
>>> isinstance(key.curve, ec.SECP256R1)
False
>>> key.curve is ec.SECP256R1
True

alex:


alex:
FWIW I initially thought derive and from encoded point would have the same issues, but on review they both already error in this case.

reaperhulk (approved):

PR #10196, opened 2024-01-17
Migrate AESGCM to Rust

jlaine:
This is my first attempt at writing Rust code, so my apologies for any noobiisms in my code.

I had to use `LazyEvpCipherAead` otherwise I encountered a `"not able to copy ctx"` error with my OpenSSL version. This seems related to https://github.com/openssl/openssl/issues/21955 which may be fixed in more recent versions.

I'm struggling to express the correct `cfg` condition here, I've got it working for OpenSSL 3.x but not LibreSSL / BoringSSL / OpenSSL 1.1.x


jlaine:
OK, and now I discovered #9181, sorry for the noise.

I'm kind of happy my code looks very close to what @alex did though :)

PR #10203, opened 2024-01-18
Revert "Silence new clippy false-positive"

alex:
Reverts pyca/cryptography#10168

In theory this has been properly fixed upstream


alex:
Not fixed. Will try again in a few days.

PR #10204, opened 2024-01-18
Added warning about SHA1 being used for response signing in ocsp.rst

Hacksawfred3232:
While SHA1 can't be used for signing requests/responses, it can still be used for adding certificates to the request/response. I am guessing this is allowable due to the fact that we're signing with a stronger algorithm (SHA256), preventing modification of digest made with a weaker algorithm (SHA1)

Despite this, there is not enough clarity to indicate this, as someone may see the SHA1 algorithm being used in the "add_response" function and thinking "Oh! We can use SHA1 here", despite that not being the case. This commits adds the needed clarity.

### DIFF
--- a/docs/x509/ocsp.rst
+++ b/docs/x509/ocsp.rst
@@ -343,7 +343,7 @@ Creating Responses
             otherwise. Please note that
             :class:`~cryptography.hazmat.primitives.hashes.SHA1`
             can not be used here, regardless of if it was used for
-            :meth: `~cryptography.x509.ocsp.OCSPResponseBuilder.add_response`
+            :meth:`~cryptography.x509.ocsp.OCSPResponseBuilder.add_response`
             or not.
 
         :returns: A new :class:`~cryptography.x509.ocsp.OCSPResponse`.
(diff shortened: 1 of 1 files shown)


Hacksawfred3232:
Apparently dictionary didn't catch "algorithim" somehow... Should be all good now.

alex:
FYI the rust nightly failure is unrelated to this PR and we're identifying a fix.

docs/x509/ocsp.rst
```diff
@@ -340,7 +340,11 @@ Creating Responses
             :class:`~cryptography.hazmat.primitives.asymmetric.ed448.Ed448PrivateKey`
             and an instance of a
             :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm`
-            otherwise.
+            otherwise. This should not be the same algorithm used to add
+            a certificate to the response, as
+            :class:`~cryptography.hazmat.primitives.hashes.SHA1` can't be used
+            for signing, despite its compatibility requirement for certain 
+            applications.
```
alex, line 347: I'd suggest rephrasing to state that this does not need to be the same algorithm used in `add_response` (i.e., link to it) and separately that SHA1 is disallowed here even though it's allowed here.

We don't want people to think that if they pass `SHA256` to `add_response` that they shouldn't use it here.

Hacksawfred3232: I think I rephrased it well enough?

alex (approved):
Looks great. Will merge once we fix our CI issues. Thanks!

Issue #8770, opened 2023-04-21
Migrate backend to rust


alex:
- [x] https://github.com/pyca/cryptography/pull/8697
- [x] https://github.com/pyca/cryptography/pull/8705
- [x] https://github.com/pyca/cryptography/pull/8263
- [x] https://github.com/pyca/cryptography/pull/8691
- [x] https://github.com/pyca/cryptography/pull/8768
- [x] `aead`
  - [x] https://github.com/pyca/cryptography/pull/9181
  - [x] https://github.com/pyca/cryptography/pull/9399
  - [x] https://github.com/pyca/cryptography/pull/10279
  - [x] https://github.com/pyca/cryptography/pull/9569
  - [x] https://github.com/pyca/cryptography/pull/9359
- [x] https://github.com/pyca/cryptography/pull/9859
- [x] https://github.com/pyca/cryptography/pull/9639
- [x] https://github.com/pyca/cryptography/pull/8978
- [x] https://github.com/pyca/cryptography/pull/9024
- [x] https://github.com/pyca/cryptography/pull/8775
- [x] https://github.com/pyca/cryptography/pull/8781
- [x] https://github.com/pyca/cryptography/pull/8788
- [x] https://github.com/pyca/cryptography/pull/9152
- [x] https://github.com/pyca/cryptography/pull/8787
- [x] https://github.com/pyca/cryptography/pull/11200
- [x] https://github.com/pyca/cryptography/pull/10228

rozhuk-im:
How to use python staff on platforms without rust support?

alex:
Cryptography has required rust for more than 2 years. This issue does not change that, it merely uses rust for an additional area of functionality.

To use cryptography on a platform without rust support, you'll need to add rust support for that platform.

rozhuk-im:
Who will pay for adding rust to new platforms?

alex:
All the developers of PyCA/cryptography are volunteers, we're not being
paid for our work here. Perhaps you'd like to ask the commercial vendors
for these platforms to sponsor Rust/LLVM platform support.

rozhuk-im:
Ok, I see.
Project without moderation, responsibility and plans.
Since this is one of base python packet that widely used, all python products may be broken in any time on any platform.
This is bad for business and for users.

I will avoid python based projects and will recommend to avoid for other peoples.

alex:
I'm not going to engage further, except to note that our adoption of rust
was extensively documented and communicated more than 2 years ago. It was
[...]

PR #10228, opened 2024-01-22
Migrate PKCS7 backend to Rust

facutuesca:
Part of https://github.com/pyca/cryptography/issues/8770. Now that `rust-openssl` has the [necessary bindings](https://github.com/sfackler/rust-openssl/pull/2111), this PR replaces the Python implementation with calls to the Rust OpenSSL bindings.


src/rust/src/pkcs7.rs
```diff
@@ -290,11 +292,67 @@ fn smime_canonicalize(data: &[u8], text_mode: bool) -> (Cow<'_, [u8]>, Cow<'_, [
     }
 }
 
+fn load_pkcs7_certificates(
+    py: pyo3::Python<'_>,
+    pkcs7: Pkcs7,
+) -> CryptographyResult<Vec<x509::certificate::Certificate>> {
+    let nid = pkcs7.type_().map_or(openssl::nid::Nid::UNDEF, |x| x.nid());
+    assert_ne!(nid, openssl::nid::Nid::UNDEF);
+    if nid != openssl::nid::Nid::PKCS7_SIGNED {
```
alex, line 301: What's the reason for this assertion? Can we not just do `nid = pkcs7.type_().map(|t| t.nid()); if nid != Some(PKCS7_SIGNED) {}`?

src/rust/src/pkcs7.rs
```diff
@@ -290,11 +292,67 @@ fn smime_canonicalize(data: &[u8], text_mode: bool) -> (Cow<'_, [u8]>, Cow<'_, [
...
+    let signed_certificates = pkcs7.signed().and_then(|x| x.certificates());
+    match signed_certificates {
+        None => Err(CryptographyError::from(
+            pyo3::exceptions::PyValueError::new_err(
+                "The provided PKCS7 has no certificate data, but a cert loading method was called.",
+            ),
+        )),
+        Some(c) => c
+            .iter()
+            .map(|c| load_pem_x509_certificate(py, c.to_pem()?.as_slice(), None))
```
alex, line 318: Round-trip via DER, it'll be faster.

src/rust/src/pkcs7.rs
facutuesca: I replicated the behavior from the Python implementation: https://github.com/pyca/cryptography/pull/10228/files#diff-f5134bf8f3cf0a5cc8601df55e50697acc866c603a38caff98802bd8e17976c5L896. Should I change it?

alex: Yes, I think my formulation here is better -- 

it's important to note that the `openssl_assert()` would produce a python execption while this code would produce a rust panic, so they're not quite teh same

facutuesca: fixed

facutuesca:
Leaving as a draft for now since `rust-openssl` doesn't have PKCS#7 bindings enabled for BoringSSL

alex:
I believe that's because BoringSSL is missing PKCS#7 functionality -- we should just disable this there.

src/rust/src/pkcs7.rs
```diff
@@ -290,11 +296,97 @@ fn smime_canonicalize(data: &[u8], text_mode: bool) -> (Cow<'_, [u8]>, Cow<'_, [
...
+                    "Unable to parse PKCS7 data",
+                ))
+            })?;
+            load_pkcs7_certificates(py, pkcs7_decoded)
+        } else {
+            return Err(CryptographyError::from(
+                exceptions::UnsupportedAlgorithm::new_err((
+                    "PKCS#7 is not supported by this backend.",
+                    exceptions::Reasons::BACKEND_MISSING_INTERFACE,
+                )),
```
alex, line 377: Use `UNSUPPORTED_SERIALIZATION` here I think

src/rust/src/pkcs7.rs
```diff
@@ -290,11 +296,97 @@ fn smime_canonicalize(data: &[u8], text_mode: bool) -> (Cow<'_, [u8]>, Cow<'_, [
     }
 }
 
+#[cfg(not(CRYPTOGRAPHY_IS_BORINGSSL))]
+fn load_pkcs7_certificates(
+    py: pyo3::Python<'_>,
+    pkcs7: Pkcs7,
+) -> CryptographyResult<Vec<x509::certificate::Certificate>> {
```
alex, line 303: Small nit: Instead of returning a Vec, I'd suggest the `&pyo3::types::PyList`. This avoids a copy+malloc round trip.

See verify.rs l239 for an example of this

alex (dismissed):

src/rust/src/pkcs7.rs
```diff
@@ -290,11 +296,101 @@ fn smime_canonicalize(data: &[u8], text_mode: bool) -> (Cow<'_, [u8]>, Cow<'_, [
...
+            let result = pyo3::types::PyList::empty(py);
+            for c in certificates {
+                result.append(
+                    load_der_x509_certificate(
+                        py,
+                        pyo3::types::PyBytes::new(py, c.to_der()?.as_slice()).into_py(py),
+                        None,
+                    )?
+                    .into_py(py),
+                )?;
```
alex, line 332: You're getting coverage issues here. Best resolution is to rewrite as:

```
let cert_der = pyo3::types::PyBytes::new(py, c.to_der()?.as_slice()).into_py(py);
let cert = load_der_x509_certificate(py, cert_der, None)?;
result.append(cert.into_py(py))?;
```

src/rust/src/pkcs7.rs
facutuesca: fixed

src/rust/src/pkcs7.rs
facutuesca: fixed

src/rust/src/pkcs7.rs
facutuesca: thx! fixed

alex (approved):

PR #10248, opened 2024-01-24
allow SPKI RSA keys to be parsed even if they have an incorrect delimiter

reaperhulk:
This allows RSA SPKI keys (typically delimited with PUBLIC KEY) to be parsed even if they are using the RSA PUBLIC KEY delimiter.

refs #10121


src/rust/src/backend/keys.rs
```diff
@@ -165,7 +165,15 @@ fn load_pem_public_key(
...
     let p = pem::parse(data.as_bytes())?;
     let pkey = match p.tag() {
-        "RSA PUBLIC KEY" => cryptography_key_parsing::rsa::parse_pkcs1_public_key(p.contents())?,
+        "RSA PUBLIC KEY" => {
+            // We try to parse it as a PKCS1 first since that's the PEM delimiter, and if
+            // that fails we try to parse it as an SPKI. This is to match the permissiveness
+            // of OpenSSL, which doesn't care about the delimiter.
+            match cryptography_key_parsing::rsa::parse_pkcs1_public_key(p.contents()) {
+                Ok(pkey) => pkey,
+                Err(_) => cryptography_key_parsing::spki::parse_public_key(p.contents())?,
```
alex, line 174: I think using the err from parsing a pkcs1 key here makes sense.

alex, line 174: Also this means that an EC key would be allowed if it was an SPKI.

Does OpenSSL allow this? Should we?

reaperhulk: A quick check shows that wrapping an EC public key with `RSA PUBLIC KEY` as a delimiter fails with OpenSSL.

alex (approved):

PR #10266, opened 2024-01-26
improve the performance of cffibuf

reaperhulk:


alex:
Please put this ina function so I can use it in the cipher pr when I add cffibufmut

reaperhulk:
While this is a performance uplift for everything that uses CffiBuf, here are some specific numbers for ChaCha on my M1 Max laptop:

```python
import random
import time

from cryptography.hazmat.primitives.ciphers.aead import ChaCha20Poly1305

ITERATIONS = 1000000
KEY = bytes(32)
DATA = bytes(64)

start = time.time()
c = ChaCha20Poly1305(KEY)
for i in range(ITERATIONS):
    nonce = random.randbytes(12) # non-CSPRNG but just used to ensure there's always a new nonce
    c.encrypt(nonce, DATA, None)
elapsed = time.time() - start
print(f"{ITERATIONS} iterations with teardown in {elapsed:.1f} s")
```

| Version  | Time |
| -------- | ----- |
| 42.0.1    | 2.2s   |
| [ChaCha20Reuseable](https://github.com/bdraco/chacha20poly1305-reuseable) | 2.1s |
| #10266 | 1.6s   |

src/rust/src/buf.rs
```diff
@@ -12,6 +12,18 @@ pub(crate) struct CffiBuf<'p> {
...
 }
 
+fn _extract_buffer_length(pyobj: &pyo3::PyAny) -> pyo3::PyResult<(&pyo3::PyAny, usize)> {
+    let py = pyobj.py();
+    let bufobj = types::FFI_FROM_BUFFER.get(py)?.call1((pyobj,))?;
+    let ptrval = types::FFI_CAST
+        .get(py)?
+        .call1((pyo3::intern!(py, "uintptr_t"), bufobj))?
+        .getattr(pyo3::intern!(py, "__int__"))?
+        .call0()?
```
alex, line 22:
```suggestion
        .call_method0(pyo3::intern!(py, "__int__"))?
```

alex (approved):

PR #10278, opened 2024-01-28
Don't generate RSA keys <1024 bits

alex:


alex:
Needs https://github.com/twisted/twisted/pull/12095

reaperhulk:
We should probably document this change

reaperhulk (approved):

terryjreedy:
Thank you.  https://arstechnica.com/security/2024/08/home-energy-system-gives-researcher-control-of-virtual-power-plant/ is an example of why libraries should not allow easily cracked (for $70) keys.  Naive users will assume that they are OK.  (I am partly responding to https://github.com/pyca/cryptography/commit/83dcbc190165ad5c1f86bddaee76e0b288803c43#commitcomment-144797986 from @craigjbass.)  The referenced article links to this commit.

craigjbass:
Yeah I understand the wider implications of pushing for secure key lengths but some way to override (opt in and partly hidden) would be welcomed by advanced users.

chaen:
We run into the same problem and would be very happy to have a way of keeping the key length at 512

reaperhulk:
It is increasingly likely that OpenSSL and other libraries are going to disable generation of 512-bit keys in the relatively near future. We still allow the parsing of this key length (although that may also be disabled by underlying libraries like boringssl soon!), but generally we'd strongly recommend moving to larger key sizes even in tests.

PR #10284, opened 2024-01-28
add decrepit namespace and put SEED, IDEA, Blowfish, and CAST5 in it

reaperhulk:
diff is a big large but most of it is moving the `test_{seed,idea,cast5,blowfish}.py` files into `test_decrepit.py` along with lots of docs.


src/cryptography/hazmat/decrepit/ciphers/algorithms.py
```diff
@@ -0,0 +1,25 @@
...
+)
+from cryptography.hazmat.primitives.ciphers.algorithms import (
+    _CAST5Internal as CAST5,
+)
+from cryptography.hazmat.primitives.ciphers.algorithms import (
+    _IDEAInternal as IDEA,
+)
+from cryptography.hazmat.primitives.ciphers.algorithms import (
+    _SEEDInternal as SEED,
+)
```
alex, line 18: Surely we should just do the reverse -- move the classes here, and re-export them in the other place but deprecated.

tests/hazmat/primitives/test_decrepit.py
```diff
@@ -0,0 +1,340 @@
+# This file is dual licensed under the terms of the Apache License, Version
```
alex, line 1: This filename doesn't line up with the name of the module with teh code?

reaperhulk: should we do `hazmat/primitives/decrepit/test_algorithms.py`?

alex: yeah

src/cryptography/hazmat/decrepit/ciphers/algorithms.py
reaperhulk: Circular dependencies make this annoying, but we can do it (albeit with a significantly larger diff probably). We could also do that as a followup PR.

reaperhulk: (done)

alex (approved):

misterzed88:
That is a nice solution. Thank you for doing the change!

PR #10286, opened 2024-01-29
migrate ARC4 and TripleDES to decrepit

reaperhulk:


reaperhulk:
I marked these as being removed 3 releases later than the others, but that's open for discussion.

I also didn't combine the test files since we have a `tests/hazmat/decrepit` now. Assuming we merge this I'll actually un-merge the test_algorithms file so we have `test_{seed,idea,cast5,blowfish}` again in a follow-up PR.

alex (approved):

PR #10290, opened 2024-01-29
rebuild if the OPENSSL_DIR has changed

reaperhulk:
This probably needs to be done in other crates too? And what if `OPENSSL_DIR` isn't set?

### DIFF
--- a/src/rust/cryptography-cffi/build.rs
+++ b/src/rust/cryptography-cffi/build.rs
@@ -25,6 +25,9 @@ fn main() {
     let out_dir = env::var("OUT_DIR").unwrap();
     // FIXME: maybe pyo3-build-config should provide a way to do this?
     let python = env::var("PYO3_PYTHON").unwrap_or_else(|_| "python3".to_string());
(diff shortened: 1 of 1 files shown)


src/rust/cryptography-cffi/build.rs
```diff
@@ -25,6 +25,9 @@ fn main() {
     let out_dir = env::var("OUT_DIR").unwrap();
     // FIXME: maybe pyo3-build-config should provide a way to do this?
     let python = env::var("PYO3_PYTHON").unwrap_or_else(|_| "python3".to_string());
+    if let Ok(dir) = env::var("OPENSSL_DIR") {
```
alex, line 28: This should probably just be in rust-openssl right? Is there a reason it shouldn't?

reaperhulk: Nope! Up there would make the most sense, especially since it finds `OPENSSL_DIR` if the env isn't set anyway.

reaperhulk:
https://github.com/sfackler/rust-openssl/pull/2157

PR #10302, opened 2024-01-30
parsing, verification: check RSA key size against WebPKI minimum

woodruffw:
Depends on https://github.com/C2SP/x509-limbo/pull/184 for coverage.

### DIFF
--- a/.github/actions/fetch-vectors/action.yml
+++ b/.github/actions/fetch-vectors/action.yml
@@ -16,5 +16,5 @@ runs:
       with:
         repository: "C2SP/x509-limbo"
         path: "x509-limbo"
-        # Latest commit on the x509-limbo main branch, as of Jan 30, 2024.
+        # Latest commit on the x509-limbo main branch, as of Jan 31, 2024.
         ref: "481b5d595b00ce55824607e1e8c2f1174539f3f8" # x509-limbo-ref
--- a/.github/actions/fetch-vectors/action.yml
+++ b/.github/actions/fetch-vectors/action.yml
@@ -17,4 +17,4 @@ runs:
         repository: "C2SP/x509-limbo"
         path: "x509-limbo"
         # Latest commit on the x509-limbo main branch, as of Jan 30, 2024.
-        ref: "dd7541dac329f03756f6358ad0c01d32e5677619" # x509-limbo-ref
+        ref: "481b5d595b00ce55824607e1e8c2f1174539f3f8" # x509-limbo-ref
--- a/src/rust/cryptography-x509-verification/src/policy/mod.rs
+++ b/src/rust/cryptography-x509-verification/src/policy/mod.rs
@@ -436,19 +436,6 @@ impl<'a, B: CryptoOps> Policy<'a, B> {
 
         self.ca_extension_policy.permits(self, cert, extensions)?;
 
-        // CA/B 6.1.5: Key sizes
-        // NOTE: We don't currently enforce that RSA moduli are divisible by 8,
-        // since other implementations don't bother.
-        let cert_spki = &cert.tbs_cert.spki;
-        if cert_spki.algorithm.is_rsa_key() {
-            let rsa_key: Pksc1RsaPublicKey<'_> =
-                asn1::parse_single(cert_spki.subject_public_key.as_bytes())?;
-
-            if rsa_key.n.as_bytes().len() * 8 < self.minimum_rsa_modulus {
-                return Err(ValidationError::Other("RSA key is too weak".into()));
-            }
-        }
-
         Ok(())
     }
 
@@ -508,6 +495,22 @@ impl<'a, B: CryptoOps> Policy<'a, B> {
                 "Forbidden signature algorithm: {:?}",
                 &child.signature_alg
             )));
+        }
+
+        // CA/B 6.1.5: Key sizes
+        // NOTE: We don't currently enforce that RSA moduli are divisible by 8,
+        // since other implementations don't bother.
+        let issuer_spki = &issuer.certificate().tbs_cert.spki;
+        if matches!(
+            issuer_spki.algorithm.params,
+            AlgorithmParameters::Rsa(_) | AlgorithmParameters::RsaPss(_)
+        ) {
+            let rsa_key: Pksc1RsaPublicKey<'_> =
+                asn1::parse_single(issuer_spki.subject_public_key.as_bytes())?;
+
+            if rsa_key.n.as_bytes().len() * 8 < self.minimum_rsa_modulus {
+                return Err(ValidationError::Other("RSA key is too weak".into()));
+            }
         }
 
         let pk = issuer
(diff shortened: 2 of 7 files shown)


src/rust/cryptography-x509-verification/src/policy/mod.rs
```diff
@@ -329,6 +338,19 @@ impl<'a, B: CryptoOps> Policy<'a, B> {
...
+        // since other implementations don't bother.
+        let cert_spki = &cert.tbs_cert.spki;
+        if cert_spki.algorithm.is_rsa() {
+            let rsa_key: Pksc1RsaPublicKey<'_> =
+                asn1::parse_single(cert_spki.subject_public_key.as_bytes())?;
+
+            if rsa_key.n.as_bytes().len() * 8 < self.minimum_rsa_modulus {
+                return Err(ValidationError::Other("RSA key is too weak".into()));
+            }
+        }
```
alex, line 352: This should probably be below the other checks.

Also is `permits_basic` the right place for this, or should it be a part of the signature algorithm checks. That is to say: Do we enforce a minimum key size for the EE key, even though it's never used in signature validation?

src/rust/cryptography-x509/src/common.rs
```diff
@@ -17,6 +17,24 @@ impl AlgorithmIdentifier<'_> {
...
+                | AlgorithmParameters::RsaWithSha256(_)
+                | AlgorithmParameters::RsaWithSha384(_)
+                | AlgorithmParameters::RsaWithSha512(_)
+                | AlgorithmParameters::RsaWithSha3_224(_)
+                | AlgorithmParameters::RsaWithSha3_256(_)
+                | AlgorithmParameters::RsaWithSha3_384(_)
+                | AlgorithmParameters::RsaWithSha3_512(_)
+                | AlgorithmParameters::RsaPss(_)
+        )
+    }
```
alex, line 37: This combines key algorithms and signature algorithms, which I think is pretty confusing

src/rust/cryptography-x509-verification/src/policy/mod.rs
woodruffw, line 352: NB: This made me realize that the other key/algorithm checks also probably belong in `permits_*`, not in `valid_issuer` (since CABF doesn't declare keys/algos valid on an issuance basis, but in absolute terms).

I didn't move the other checks in this same PR to keep the diff small, but I can do so (or open a follow-up for them.)

woodruffw: Whoops, raced you on this; https://github.com/pyca/cryptography/pull/10302#discussion_r1472130273 has my rationale for this being in `permits_basic`. I'll move this below the other checks, though.

woodruffw: I'll also check to see what other implementations do re: EE key sizes.

src/rust/cryptography-x509/src/common.rs
woodruffw: Nixed the signature algorithms and renamed to `is_rsa_key` (but maybe it should be `is_rsa_key_algo`?)

src/rust/cryptography-x509-verification/src/policy/mod.rs
woodruffw: Moved this to `permits_ca` since other implementations (besides OpenSSL, which decides to be special) don't enforce the key strength check on EEs.

src/rust/cryptography-x509-verification/src/policy/mod.rs
```diff
@@ -427,6 +436,19 @@ impl<'a, B: CryptoOps> Policy<'a, B> {
...
+        let cert_spki = &cert.tbs_cert.spki;
+        if cert_spki.algorithm.is_rsa_key() {
+            let rsa_key: Pksc1RsaPublicKey<'_> =
+                asn1::parse_single(cert_spki.subject_public_key.as_bytes())?;
+
+            if rsa_key.n.as_bytes().len() * 8 < self.minimum_rsa_modulus {
+                return Err(ValidationError::Other("RSA key is too weak".into()));
+            }
+        }
+
```
alex, line 451: Why not valid issuer with the other key checks?

src/rust/cryptography-x509/src/common.rs
```diff
@@ -17,6 +17,10 @@ impl AlgorithmIdentifier<'_> {
     pub fn oid(&self) -> &asn1::ObjectIdentifier {
         self.params.item()
     }
+
+    pub fn is_rsa_key(&self) -> bool {
+        matches!(self.params, AlgorithmParameters::Rsa(_))
+    }
```
alex, line 23: Wellll, now this excludes RsaPss which is a distinct key type.

I think maybe you just want an if statement in the validator :-)

woodruffw: Yeah, that'll be cleaner!

src/rust/cryptography-x509-verification/src/policy/mod.rs
woodruffw: https://github.com/pyca/cryptography/pull/10302#discussion_r1472130273 is my underlying reason, but I think consistency is an equally good reason. Moving!

alex (approved):

src/rust/cryptography-x509-verification/src/policy/mod.rs
```diff
@@ -488,6 +497,22 @@ impl<'a, B: CryptoOps> Policy<'a, B> {
             )));
         }
 
+        // CA/B 6.1.5: Key sizes
+        // NOTE: We don't currently enforce that RSA moduli are divisible by 8,
+        // since other implementations don't bother.
+        let issuer_spki = &issuer.certificate().tbs_cert.spki;
+        if matches!(
+            issuer_spki.algorithm.params,
+            AlgorithmParameters::Rsa(_) | AlgorithmParameters::RsaPss(_)
```
woodruffw, line 506: Sanity checking myself: I can remove this `RsaPss` variant, right? I know it's a key type but AFAICT it never appears in SPKI AlgorithmIdentifiers.

alex: It can absolutely occur in an SPKI AlgorithmIdentifier.

But it'll never be true _here_ because it's not listed as a valid algorithm identifier which we check a few lines up.

.github/actions/fetch-vectors/action.yml
```diff
@@ -17,4 +17,4 @@ runs:
         repository: "C2SP/x509-limbo"
         path: "x509-limbo"
         # Latest commit on the x509-limbo main branch, as of Jan 30, 2024.
```
alex, line 19:
```suggestion
        # Latest commit on the x509-limbo main branch, as of Jan 31, 2024.
```

alex (approved):

PR #10311, opened 2024-01-31
verification/policy: tweak key checks

woodruffw:
This changes the `valid_issuer` logic slightly, to the following:

1. The issuer's SPKI is checked instead of the child's, since all issuing certificates must confirm to CABF (previously, this would allow a root with a non-CABF-allowed key when used in `root -> EE` configuration)
2. The subject's signature is still checked, and I've added a note explaining why we check the subject's signature rather than the issuer's

With these changes, I've confirmed that the existing tests pass as expected (and additionally, the new tests in the limbo PR below also newly pass).

Needs https://github.com/C2SP/x509-limbo/pull/185.

### DIFF
--- a/tests/x509/verification/test_limbo.py
+++ b/tests/x509/verification/test_limbo.py
@@ -27,7 +27,10 @@ LIMBO_UNSUPPORTED_FEATURES = {
     # Our support for custom EKUs is limited, and we (like most impls.) don't
     # handle all EKU conditions under CABF.
     "pedantic-webpki-eku",
-    # Similarly: contains tests that fail based on a strict reading of RFC 5280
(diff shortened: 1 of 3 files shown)


tests/x509/verification/test_limbo.py
```diff
@@ -64,7 +67,7 @@
     "webpki::aki::root-with-aki-ski-mismatch",
     # We allow RSA keys that aren't divisible by 8, which is technically
     # forbidden under CABF. No other implementation checks this either.
-    "webpki::forbidden-rsa-key-not-divisable-by-8",
```
woodruffw, line 67: NB: This test was just renamed; no behavioral change.

alex (approved):

PR #10323, opened 2024-02-02
initialize openssl's legacy provider in rust

reaperhulk:
as we oxidize we need to do this here to ensure it actually happens


src/cryptography/hazmat/bindings/openssl/binding.py
```diff
@@ -106,25 +105,6 @@ def _ensure_ffi_initialized(cls) -> None:
...
-                # any time soon.
-                if cls.lib.CRYPTOGRAPHY_OPENSSL_300_OR_GREATER:
-                    if not os.environ.get("CRYPTOGRAPHY_OPENSSL_NO_LEGACY"):
-                        cls._legacy_provider = cls.lib.OSSL_PROVIDER_load(
-                            cls.ffi.NULL, b"legacy"
-                        )
-                        cls._legacy_provider_loaded = (
-                            cls._legacy_provider != cls.ffi.NULL
-                        )
-                        _legacy_provider_error(cls._legacy_provider_loaded)
```
alex, line 122: This function can be removed now.

src/rust/src/lib.rs
```diff
@@ -25,6 +47,35 @@ fn is_fips_enabled() -> bool {
...
+        .map(|v| v.is_empty() || v == "0")
+        .unwrap_or(true);
+    if load_legacy {
+        let legacy = provider::Provider::load(None, "legacy");
+        _legacy_provider_error(legacy.is_ok())?;
+        let default = provider::Provider::load(None, "default")?;
+        Ok(LoadedProviders::new(Some(default), Some(legacy?)))
+    } else {
+        Ok(LoadedProviders::new(None, None))
+    }
```
alex, line 67: This does not match the previous behavior, which was to always load the the default provider.

src/rust/src/lib.rs
```diff
@@ -52,6 +103,21 @@ fn _rust(py: pyo3::Python<'_>, m: &pyo3::types::PyModule) -> pyo3::PyResult<()>
     m.add_submodule(cryptography_cffi::create_module(py)?)?;
 
     let openssl_mod = pyo3::prelude::PyModule::new(py, "openssl")?;
+    cfg_if::cfg_if! {
+        if #[cfg(CRYPTOGRAPHY_OPENSSL_300_OR_GREATER)] {
+            let providers = _initialize_legacy_provider()?;
+            m.add_class::<LoadedProviders>()?;
```
alex, line 109: You don't actually need to add class, no one is using the class.

src/rust/src/lib.rs
```diff
@@ -15,6 +20,23 @@ mod pkcs7;
 pub(crate) mod types;
 mod x509;
 
+#[cfg(CRYPTOGRAPHY_OPENSSL_300_OR_GREATER)]
+#[pyo3::prelude::pyclass(frozen, module = "cryptography.hazmat.bindings._rust")]
+struct LoadedProviders {
+    _default: Option<provider::Provider>,
```
alex, line 26: No need this to be `Option<>`

src/rust/src/lib.rs
```diff
@@ -25,6 +36,35 @@ fn is_fips_enabled() -> bool {
     cryptography_openssl::fips::is_enabled()
 }
 
+#[cfg(CRYPTOGRAPHY_OPENSSL_300_OR_GREATER)]
+fn _initialize_legacy_provider() -> CryptographyResult<LoadedProviders> {
+    /* As of OpenSSL 3.0.0 we must register a legacy cipher provider
+    to get RC2 (needed for junk asymmetric private key
+    serialization), RC4, Blowfish, IDEA, SEED, etc. These things
+    are ugly legacy, but we aren't going to get rid of them
+    any time soon. */
```
alex, line 45: Use `//` comments, not this weirdo crap

alex (approved):

PR #10333, opened 2024-02-02
backport: initialize openssl's legacy provider in rust (#10323)

reaperhulk:
* initialize openssl's legacy provider in rust

as we oxidize we need to do this here to ensure it actually happens

* alex is a comment format pedant


alex:
Tests are failing because this doesn't include the memleak backport.

alex (approved):

Issue #10276, opened 2024-01-28
Allow verifying an x509 cert chain without making assertions about the subject name


kislyuk:
Thanks to all who worked on the X.509 verification support in version 42.

I am trying to use this API for verifying a signing certificate, and realizing that the API requires me to assert a subject name (DNS name or IP address) to get the validation output. The subject name is not defined/not relevant in this application.

How can I verify that a certificate is in the chain of trust without asserting on the subject name?

alex:
We don't yet support this use case. We intend to in the future.

woodruffw:
I've been thinking a bit about how to do this (since I'll also likely need it at some point for `sigstore-python`. Some ideas:

1. Make the current `build_server_verifier(subject: Subject)` take an optional subject instead (maybe explicitly `None`?), which would then disable any SAN checks during chain building.
2. Add a new `build_generic_verifier(subject: Subject | None)`, which would then be an explicit path for chain building operations that are more "generic" than server authentication (presumed by `build_server_identifier`).

alex:
Well, we should not be changing the build_server_verifier function. You should never verify a server cert without checking the subject is valid.

This needs to be a new type of verifier.

woodruffw:
That makes sense. How does `build_generic_verifier`/`GenericVerifier` sound? 

(Other meh names: `AnonymousVerifier`/`SubjectlessVerifier` for explicit non-subject checking.)

alex:
What's teh difference between that and a client cert verifier?

woodruffw:
Hmm, not sure. My idea of a `ClientVerifier` was something that takes a nonempty `list[Subject]` to match against so it wouldn't work for the "explicitly not checking the subject" case. But maybe I'm thinking about that the wrong way.

alex:
I don't think that's how a clientverifier would work -- imagine a webserver using client certs for auth.

woodruffw:
So in that case `ClientVerifier.verify()` would return the `Subject` for subsequent AuthZ/handling, right? In that case yeah, I think this is indistinguishable from the client cert verifier case and can be handled through it.

alex:
Yes, I think that's what you'd want, indeed.

woodruffw:
Just to give a public update: I've offered to do the work here. I'll try and get a PR up for this tonight.

woodruffw:
[...]

PR #10335, opened 2024-02-03
verification/policy: make subject optional internally

woodruffw:
This is not surfaced in a public API yet; it's purely an internal change to enable a `ClientVerifier` API (which won't take a subject).

See #10276.

### DIFF
--- a/src/rust/cryptography-x509-verification/src/policy/extension.rs
+++ b/src/rust/cryptography-x509-verification/src/policy/extension.rs
@@ -304,7 +304,11 @@ pub(crate) mod ee {
         };
 
         let san: SubjectAlternativeName<'_> = extn.value()?;
-        if !policy.subject.matches(&san) {
+        if !policy
+            .subject
+            .as_ref()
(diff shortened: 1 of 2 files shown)


src/rust/cryptography-x509-verification/src/policy/extension.rs
```diff
@@ -304,7 +304,11 @@ pub(crate) mod ee {
         };
 
         let san: SubjectAlternativeName<'_> = extn.value()?;
-        if !policy.subject.matches(&san) {
+        if !policy
+            .subject
+            .as_ref()
+            .map_or_else(|| false, |sub| sub.matches(&san))
+        {
```
woodruffw, line 311: NB: This fails closed if the surrounding `Policy` doesn't have a configured subject. AFAICT this is the only reasonable behavior, since the SAN extension verifier here only makes sense when a subject is supplied by the user.

(In the context of a new `ClientVerifier` API, we'll need a different SAN extension verifier.)

alex (approved):

PR #10344, opened 2024-02-04
openssl: Support OpenSSL v3

Sonic-Amiga:
Allow build with modern OpenSSL. Some projects may require old cryptography version

### DIFF
--- a/src/_cffi_src/openssl/err.py
+++ b/src/_cffi_src/openssl/err.py
@@ -45,4 +45,11 @@ int ERR_GET_REASON(unsigned long);
 """
 
 CUSTOMIZATIONS = """
+#if OPENSSL_VERSION_MAJOR >= 3
+// OpenSSL v3 has dropped this function
+int ERR_GET_FUNC(unsigned long errcode)
+{
+    return 0;
+}
+#endif
 """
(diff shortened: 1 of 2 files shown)


Sonic-Amiga:
Hello there!

I know 3.4.x is ve-e-e-ry old; but this environment https://docs.espressif.com/projects/esp8266-rtos-sdk/en/latest/ requires no later than this version. The goal of this patch is to allow having such old python lib version on modern systems.

Sonic-Amiga:
And could you kindly give me some help on CI failures ? The patch itself is quite primitive; it does not change anything concerning for instance this:
```
flake: failed with pass_env values cannot contain whitespace, use comma to have multiple values in a single line, invalid values found 'ARCHFLAGS LDFLAGS CFLAGS INCLUDE LIB LD_LIBRARY_PATH USERNAME PYTHONIOENCODING OPENSSL_FORCE_FIPS_MODE'
```
Sorry for stupid question, what is pass_env and where does it come from ? Yes, i am basically a C coder, not a python one.

alex:
Sorry, but we do not support versions this old. You'll need to either upgrade the version of crpytography you depend on, or maintain your own fork.

PR #10345, opened 2024-02-04
verification: client verification APIs

### LINKED ISSUE #10276: Allow verifying an x509 cert chain without making assertions about the subject name

woodruffw:
~~Still WIP -- the public APIs are not fully filled in yet. Outstanding tasks:~~

- [x] Fill in `build_client_verifier`
- [x] Add and implement `ClientVerifier.verify`
- [x] Unit tests (including limbo tests?)

Closes #10276.


src/cryptography/hazmat/bindings/_rust/x509.pyi
```diff
@@ -62,10 +62,26 @@ class PolicyBuilder:
...
+    def store(self) -> Store: ...
+    @property
+    def max_chain_depth(self) -> int: ...
+    def verify(
+        self,
+        leaf: x509.Certificate,
+        intermediates: list[x509.Certificate],
+    ) -> tuple[
+        x509.Name, x509.SubjectAlternativeName | None, list[x509.Certificate]
+    ]: ...
```
woodruffw, line 83: Before I go further here: does this return type make sense? I figure it makes sense to return the subject explicitly (to make usage clear + make misuse harder, since the types won't align with server verification), but it feels a little funky.

OTOH, I can't think of a cleaner signature -- I think this needs to return both the subject and the SAN, since client verification isn't super well-defined.

src/rust/cryptography-x509-verification/src/policy/mod.rs
```diff
@@ -234,18 +234,17 @@ pub struct Policy<'a, B: CryptoOps> {
...
-    /// the CA/B Forum's Basic Requirements.
-    pub fn new(
+    fn new(
         ops: B,
-        subject: Subject<'a>,
+        subject: Option<Subject<'a>>,
         time: asn1::DateTime,
         max_chain_depth: Option<u8>,
     ) -> Self {
+        let has_subject = subject.is_some();
```
woodruffw, line 243: NB: We need this binding because `subject` is stolen by the `Self { ... }` ctor, meaning that we can't bind on it in the `match` below.

src/rust/src/x509/verify.rs
```diff
@@ -287,6 +330,7 @@ fn build_subject<'a>(
 
             Ok(Subject::IP(ip_addr))
         }
+        SubjectOwner::None => Err(pyo3::exceptions::PyValueError::new_err("missing subject")),
```
woodruffw, line 333: Flagging: this is kind of nasty -- we need it for exhaustion reasons, but it should be unreachable in practice (since this variant is never passed into `build_subject`).

src/rust/src/x509/verify.rs
```diff
@@ -169,6 +200,7 @@ enum SubjectOwner {
     // so, which was only stabilized with 3.10.
     DNSName(String),
     IPAddress(pyo3::Py<pyo3::types::PyBytes>),
+    None,
```
woodruffw, line 203: Flagging: This is a bit of a hack; it's the simplest way I could think of to get the "no subject provided" state into the `SubjectOwner` type, which is then required to erase the lifetime on `Policy<'_>` (via `OwnedPolicy`).

src/cryptography/hazmat/bindings/_rust/x509.pyi
woodruffw: Named type makes sense! 

Re: subject: in that case, perhaps client verification should enforce the presence of a SAN? That should be trivial to do.

alex:
That makes sense to me

src/cryptography/x509/verification.py
```diff
@@ -7,18 +7,24 @@
...
+    "VerifiedClient",
+    "ClientVerifier",
     "ServerVerifier",
     "PolicyBuilder",
     "VerificationError",
 ]
 
 Store = rust_x509.Store
 Subject = typing.Union[DNSName, IPAddress]
+VerifiedClient = tuple[Extension[SubjectAlternativeName], list[Certificate]]
```
woodruffw, line 26: NB: I did this as a type alias for a tuple, but I could make this stronger as a `NewType` if wanted.

alex: Please make it a full class -- tuples are huge backwards compat footguns.

src/rust/src/x509/verify.rs
```diff
@@ -180,6 +215,92 @@ self_cell::self_cell!(
...
+
+        // NOTE: These `unwrap()` cannot fail, since the underlying policy
+        // enforces the presence of a SAN and the well-formedness of the
+        // extension set.
+        let leaf_san = &chain[0]
+            .certificate()
+            .extensions()
+            .unwrap()
+            .get_extension(&SUBJECT_ALTERNATIVE_NAME_OID)
+            .unwrap();
```
woodruffw, line 285: Flagging: these unwraps are pretty gross, but we need them because we don't have an easy way to propagate the fact that the SAN/extensions were already enforced up here.

src/rust/cryptography-x509-verification/src/policy/extension.rs
```diff
@@ -704,7 +704,7 @@ mod tests {
         let cert_pem = v1_cert_pem();
         let cert = cert(&cert_pem);
         let ops = PublicKeyErrorOps {};
-        let policy = Policy::new(
+        let policy = Policy::server(
```
alex, line 707: Let's put this rename into it's own PR please.

src/rust/cryptography-x509-verification/src/policy/mod.rs
```diff
@@ -315,11 +314,17 @@ impl<'a, B: CryptoOps> Policy<'a, B> {
                     Criticality::Agnostic,
                     Some(ee::key_usage),
                 ),
-                // CA/B 7.1.2.7.12 Subscriber Certificate Subject Alternative Name
-                subject_alternative_name: ExtensionValidator::present(
-                    Criticality::Agnostic,
-                    Some(ee::subject_alternative_name),
-                ),
+                subject_alternative_name: match has_subject {
```
alex, line 317: You can just use `if has_subject { blah } else { blah }`.

src/rust/src/x509/verify.rs
```diff
@@ -118,6 +124,34 @@ impl PolicyBuilder {
...
+            }
+        };
+
+        let time = match self.time.as_ref() {
+            Some(t) => t.clone(),
+            None => datetime_now(py)?,
+        };
+
+        let policy = OwnedPolicy::try_new(SubjectOwner::None, |_| {
+            Ok::<PyCryptoPolicy<'_>, pyo3::PyErr>(PyCryptoPolicy(Policy::client(
```
alex, line 145: Can't `Policy::client` have a `'static` lifetime, and so this dance isn't required?

src/rust/src/x509/verify.rs
```diff
@@ -180,6 +215,107 @@ self_cell::self_cell!(
...
+        // NOTE: These `unwrap()` cannot fail, since the underlying policy
+        // enforces the presence of a SAN and the well-formedness of the
+        // extension set.
+        let leaf_san = &chain[0]
+            .certificate()
+            .extensions()
+            .unwrap()
+            .get_extension(&SUBJECT_ALTERNATIVE_NAME_OID)
+            .unwrap();
+
```
alex, line 298: You need to handle what if the SAN has GNs that are of unsupported types here.

docs/x509/verification.rst
```diff
@@ -104,6 +104,72 @@ the root of trust:
...
+
+
+.. class:: ClientVerifier
+
+    .. versionadded:: 43.0.0
+
+    A ClientVerifier verifies client certificates.
+
+    It contains and describes various pieces of configurable path
+    validation logic, such as which subject to expect, how deep prospective
```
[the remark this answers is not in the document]
woodruffw: Bad copy paste job, will fix.

docs/x509/verification.rst
```diff
@@ -104,6 +104,72 @@ the root of trust:
     :class:`cryptography.x509.general_name.DNSName`,
     :class:`cryptography.x509.general_name.IPAddress`.
 
+.. class:: VerifiedClient
+
+    .. versionadded:: 43.0.0
+
+    .. attribute:: subject
+
+        :type: :class:`~cryptography.x509.Extension`
```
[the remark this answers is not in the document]
woodruffw: No strong reason I can think of -- I was speculating about whether a client would ever care about the critical bit, but I don't think so.

src/rust/src/x509/verify.rs
woodruffw: Yes, thanks!

alex:
PS: Yes, there should be limbo tests for this. In principle the schema was designed in a way to incorporate them.

src/rust/src/x509/verify.rs
woodruffw: I did this by propagating the error out of `x509::parse_general_names` -- is that what you were thinking? I think the GN parsing here can't directly fail, since the Rust-only `GeneralName` type has variants for all possible GN forms, whereas the Python side that's produced with `x509::parse_general_names` is what doesn't.

woodruffw:
I'll work on the limbo tests for this shortly (✈️)

src/rust/src/x509/verify.rs
alex: Sorry, when I said unsupported types I specifically meant: anything besides DNSName + IPAddress (the thing we know how to do nameconstraint checks for).

If you have a SAN with both an EDIPartyName and an IPAddress, either the valid names should be [IP] only, or it should be an error -- not sure which.

woodruffw: Makes sense, thank you for clarifying!

One question: when there are no name constraints but the SAN contains GNs other than DNS and IPs, should we return those? My mental model for how client certs typically convey subjects is a little murky, but I can imagine that both email addresses and X.500 names are somewhat common (and wouldn't necessarily have NCs).

alex: Seems like a big compatibility hazard to do that: Just adding _any_ NC to any CA would break that.

Seems like really we should implement NC checking for emails at a minimum, in a seperate PR, before merging this PR.

woodruffw:
For tracking: https://github.com/C2SP/x509-limbo/pull/196 has the initial client cases. I'm going to poke at email NC handling in a separate PR first, and then align those tests here.

woodruffw:
Email NC bits are merged, so I'll take another poke at this tonight or tomorrow.

MOUBA96:
Good

woodruffw:
This will be at 100% coverage once https://github.com/C2SP/x509-limbo/pull/221 lands. After that, I'll also look at filtering the set of subjects returned by the API to just ones that we currently have NC support for.

src/rust/src/x509/verify.rs
woodruffw: Okay, we now handle NCs for unsupported types more correctly (https://github.com/pyca/cryptography/pull/10570) and surface `UnsupportedGeneralNameType` if any of the leaf's SANs don't have a corresponding Python type.

src/rust/cryptography-x509-verification/src/policy/mod.rs
```diff
@@ -315,11 +316,18 @@ impl<'a, B: CryptoOps> Policy<'a, B> {
...
-                // CA/B 7.1.2.7.12 Subscriber Certificate Subject Alternative Name
-                subject_alternative_name: ExtensionValidator::present(
-                    Criticality::Agnostic,
-                    Some(ee::subject_alternative_name),
-                ),
+                subject_alternative_name: if has_subject {
+                    // CA/B 7.1.2.7.12 Subscriber Certificate Subject Alternative Name
+                    ExtensionValidator::present(
+                        Criticality::Agnostic,
+                        Some(ee::subject_alternative_name),
```
alex, line 323: In addition to checking that subject matches, this function also checks some stuff about criticality and the cert's subject. Do we not need that logic for client certs?

tests/x509/verification/test_limbo.py
```diff
@@ -131,12 +127,33 @@ def _limbo_testcase(id_, testcase):
...
+            )
+        else:
+            verified_client = verifier.verify(
+                peer_certificate, untrusted_intermediates
+            )
+
+            expected_subjects = {
+                _get_limbo_peer(p) for p in testcase["expected_peer_names"]
+            }
+            assert expected_subjects.issubset(verified_client.subjects)
```
alex, line 154: Why is this a subset check and not an equal check?

src/rust/cryptography-x509-verification/src/policy/mod.rs
woodruffw: Yep, thanks for catching that: CABF has nothing to say about client certs, but 5280 says that the empty subject/SAN criticality rule applies. 

I think the right approach here is going to be to tweak `ee::subject_alternative_name` to pass when the `Policy` doesn't contain a subject, with the expectation that `Policy::server` is always constructed such that a subject is present. Doing that now.

tests/x509/verification/test_limbo.py
woodruffw: My original thinking for that was to enable simpler tests on the limbo side: if we know that just a single particular subject must appear and don't care about the rest, all we need is a subset check. Thinking about it more though, having it be exact probably makes the tests more descriptive.

alex (approved):

PR #10360, opened 2024-02-07
raise an exception instead of returning an empty list for pkcs7 cert …

frenzymadness:
…loading (#9947)

* raise an exception instead of returning an empty list

as davidben points out in #9926 we are calling a specific load certificates function and an empty value doesn't necessarily mean empty because PKCS7 contains multitudes. erroring is more correct.

---------

I know this change is backward incompatible but it synchronizes the behavior of the fix for CVE-2023-49083. Versions 41.0.x are not vulnerable anymore but version 42 has better implementation that raises ValueError.

The concern of backward compatibility is just between the latest 41.0.x releases because:
* 41.0.5 - is vulnerable, segfaults
* 41.0.6 and 41.0.7 - is fixed, returns empty list
* 41.0.8 - is fixed, raises ValueError (if this is merged and released)

What do you think about it?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,17 @@
 Changelog
 =========
 
+.. _v41-0-8:
+
+41.0.8 - 2024-02-07
+~~~~~~~~~~~~~~~~~~~
+
+* **BACKWARDS INCOMPATIBLE:** Loading a PKCS7 with no content field using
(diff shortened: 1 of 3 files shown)


alex:
We intentionally did not backport this change to the 41.x release, as it is not backwards compatible.

frenzymadness:
Ok, thanks for the quick reply. I'm closing the PR then.

PR #10365, opened 2024-02-08
smaller mmap in tests to fit in a 32-bit ssize_t

reaperhulk:
this still triggers the overflows we expect in the tests and should also work on 32-bit systems

fixes #10364

Issue #10364, opened 2024-02-08
test_data_too_large failing on i686-linux since 42.0.0


mweinelt:
We're seeing various test failures in nixpkgs due to the following error on i686-linux.

```
    def large_mmap():
>       return mmap.mmap(-1, 2**32, prot=mmap.PROT_READ)
E       OverflowError: Python int too large to convert to C ssize_t
```

This was likely introduced in https://github.com/pyca/cryptography/commit/596e8fbb480f5278dbbccc221fb18bb8b03b802c.

Is 32 Bit support something you are still looking into? This affects e.g. Lutris for us as a distro.

### Environment
- Python 3.11.7
- cryptography 42.0.2
- OpenSSL 3.0.13
- cffi 1.16.0
- setuptools 69.0.3
- pip 23.3.1

### Test log

```
cryptography-i686-linux> ________________________ TestAESCCM.test_data_too_large ________________________
cryptography-i686-linux> [gw12] linux -- Python 3.11.7 /nix/store/ah5382l41gaffmx8yj8w3j34n18mh1al-python3-3.11.7/bin/python3.11
cryptography-i686-linux> 
cryptography-i686-linux> self = <tests.hazmat.primitives.test_aead.TestAESCCM object at 0xf55df8f0>
cryptography-i686-linux> 
cryptography-i686-linux>     @pytest.mark.skipif(
cryptography-i686-linux>         sys.platform not in {"linux", "darwin"}, reason="mmap required"
cryptography-i686-linux>     )
cryptography-i686-linux>     def test_data_too_large(self):
cryptography-i686-linux>         key = AESCCM.generate_key(128)
cryptography-i686-linux>         aesccm = AESCCM(key)
cryptography-i686-linux>         nonce = b"0" * 12
cryptography-i686-linux>     
cryptography-i686-linux> >       large_data = large_mmap()
cryptography-i686-linux> 
cryptography-i686-linux> tests/hazmat/primitives/test_aead.py:207: 
cryptography-i686-linux> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
cryptography-i686-linux> 
cryptography-i686-linux>     def large_mmap():
cryptography-i686-linux> >       return mmap.mmap(-1, 2**32, prot=mmap.PROT_READ)
[...]

PR #10366, opened 2024-02-08
skip overflow aead tests on 32-bit systems

reaperhulk:
and revert the incorrect fix. refs #10364 and #10365


alex (approved):

mweinelt (approved):
Checks out. Thanks!

Issue #9795, opened 2023-10-28
OpenSSL 3.2 features to expose


reaperhulk:
This is a tracking issue for the features in OpenSSL 3.2.0 we want to expose or investigate further:

- [x] [AES-GCM-SIV](https://github.com/pyca/cryptography/pull/9843)
- [x] RFC6979 (needs rust-openssl work for `OSSL_PARAM`)
- [x] [argon2id](https://github.com/pyca/cryptography/pull/11524)
- [x] [Investigate whether we need to compile with `no-sm2-precomp` to prevent linking in a half megabyte of precomputed table.](https://github.com/pyca/infra/pull/526)
- [x] [Build with `no-apps` and `no-docs`](https://github.com/pyca/infra/pull/526)
- [ ] ed25519/ed448 [ctx/ph]
- [ ] HPKE
- [x] default implicit rejection is now enabled for RSA (#9785)

alex:
https://github.com/pyca/cryptography/pull/9914

adiroiban:
Thanks for including `OpenSSL 3.2.1` in `42.0.2`

I am looking into argon2 support, so that I can read Putty v3 SSH key files.

I see that `argon2` support in `cryptography` is blocked by [rust-openssl](https://github.com/sfackler/rust-openssl)
I don't see any issue created for argon2 in the `rust-openssl` repo.

I guess that this needs to be defined here https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/src/evp.rs

I am new to rust, but I am happy to help with the work required for argon2

Thanks!

reaperhulk:
Adding argon2 in rust-OpenSSL is the blocker, yes. However that requires implementing EVP_KDF support, which in turn requires OSSL_PARAM. The latter should not be exposed as a public API in the rust library. We intend to do this work but don’t have an ETA so if you want to contribute feel free!

adiroiban:
I can see that `OSSL_PARAM` is available since version 0.9.100
https://github.com/sfackler/rust-openssl/pull/2144
 I guess that is just the low-level binding API.

So it also needs https://www.openssl.org/docs/manmaster/man3/EVP_KDF.html this bindings

----------

Why not implement this via cffi python bindings, similar to what is already implemented here  https://github.com/pyca/cryptography/blob/main/src/_cffi_src/openssl/evp.py  ?

reaperhulk:
We are actively moving away from cffi and intend to eliminate it entirely at some point in the future.

[...]

PR #10369, opened 2024-02-08
Support for ECDSA deterministic signing (RFC 6979)

facutuesca:
Part of https://github.com/pyca/cryptography/issues/9795. Now that the Rust bindings needed are in place (https://github.com/sfackler/rust-openssl/pull/2144), this PR adds support for ECDSA deterministic signing (RFC 6979), a new feature in `OpenSSL >= 3.2.0`.

The way to enable it is by passing a new boolean parameter `deterministic_signing` to `ec.ECDSA()`:
``` python
signature = private_key.sign(
    data,
    ec.ECDSA(hashes.SHA256(), deterministic_signing=True)
)
```
- When using anything other than OpenSSL >= 3.2.0, this will raise an `UnsupportedAlgorithm` exception. It will also raise if [FIPS is enabled](https://github.com/openssl/openssl/blob/da1c088f599af3755aaeed1c447a39621ef12e1f/crypto/ec/ecdsa_ossl.c#L183-L190).
- The new `deterministic_signing` parameter defaults to `False`, so that old code that does not specify it maintains the same behavior.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,7 @@ Changelog
   and :class:`~cryptography.hazmat.primitives.ciphers.algorithms.ARC4` into
   :doc:`/hazmat/decrepit/index` and deprecated them in the ``cipher`` module.
   They will be removed from the ``cipher`` module in 48.0.0.
+* Added support for deterministic ECDSA (:rfc:`6979`)
 
 .. _v42-0-5:
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,7 +27,8 @@ Changelog
   and :class:`~cryptography.hazmat.primitives.ciphers.algorithms.ARC4` into
   :doc:`/hazmat/decrepit/index` and deprecated them in the ``cipher`` module.
   They will be removed from the ``cipher`` module in 48.0.0.
-* Added support for deterministic ECDSA (:rfc:`6979`)
+* Added support for deterministic
+  :class:`~cryptography.hazmat.primitives.asymmetric.ec.ECDSA` (:rfc:`6979`)
 
 .. _v42-0-5:
 
--- a/docs/hazmat/primitives/asymmetric/ec.rst
+++ b/docs/hazmat/primitives/asymmetric/ec.rst
@@ -49,7 +49,10 @@ Elliptic Curve Signature Algorithms
 
     :param bool deterministic_signing: A boolean flag defaulting to ``False``
         that specifies whether the signing procedure should be deterministic
-        or not, as defined in :rfc:`6979`.
+        or not, as defined in :rfc:`6979`. This only impacts the signing
+        process, verification is not affected (the verification process
+        is the same for both deterministic and non-deterministic signed
+        messages).
 
         .. versionadded:: 43.0.0
 
--- a/docs/hazmat/primitives/asymmetric/ec.rst
+++ b/docs/hazmat/primitives/asymmetric/ec.rst
@@ -47,6 +47,16 @@ Elliptic Curve Signature Algorithms
     :param algorithm: An instance of
         :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm`.
 
+    :param bool deterministic_signing: A boolean flag defaulting to ``False``
+        that specifies whether the signing procedure should be deterministic
+        or not, as defined in :rfc:`6979`.
+
+        .. versionadded:: 43.0.0
+
+    :raises cryptography.exceptions.UnsupportedAlgorithm: If
+        ``deterministic_signing`` is set to ``True`` and the version of
+        OpenSSL does not support ECDSA with deterministic signing.
+
     .. doctest::
 
         >>> from cryptography.hazmat.primitives import hashes
--- a/src/rust/src/backend/ec.rs
+++ b/src/rust/src/backend/ec.rs
@@ -273,7 +273,7 @@ impl ECPrivateKey {
                 )),
             ));
         }
-        let (data, _algo) = utils::calculate_digest_and_algorithm(
+        let (data, algo) = utils::calculate_digest_and_algorithm(
             py,
             data.as_bytes(),
             signature_algorithm.getattr(pyo3::intern!(py, "algorithm"))?,
@@ -287,7 +287,7 @@ impl ECPrivateKey {
         cfg_if::cfg_if! {
             if #[cfg(CRYPTOGRAPHY_OPENSSL_320_OR_GREATER)]{
                 if deterministic {
-                    let hash_function_name = _algo
+                    let hash_function_name = algo
                         .getattr(pyo3::intern!(py, "name"))?
                         .extract::<&str>()?;
                     let hash_function = openssl::md::Md::fetch(None, hash_function_name, None)?;
@@ -299,6 +299,7 @@ impl ECPrivateKey {
                     signer.set_nonce_type(openssl::pkey_ctx::NonceType::RANDOM_K)?;
                 }
             } else {
+                let _ = algo;
                 assert!(!deterministic);
             }
         }
--- a/src/rust/src/backend/ec.rs
+++ b/src/rust/src/backend/ec.rs
@@ -300,7 +300,12 @@ impl ECPrivateKey {
                 }
             } else {
                 let _ = algo;
-                assert!(!deterministic);
+                return Err(CryptographyError::from(
+                    exceptions::UnsupportedAlgorithm::new_err((
+                        "ECDSA with deterministic signature (RFC 6979) is not supported by this version of OpenSSL.",
+                        exceptions::Reasons::UNSUPPORTED_PUBLIC_KEY_ALGORITHM,
+                    )),
+                ));
             }
         }
 
--- a/src/rust/src/backend/ec.rs
+++ b/src/rust/src/backend/ec.rs
@@ -273,7 +273,6 @@ impl ECPrivateKey {
                 )),
             ));
         }
-
         let (data, _algo) = utils::calculate_digest_and_algorithm(
             py,
             data.as_bytes(),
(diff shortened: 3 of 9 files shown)


alex:
rust-openssl has had a release.

Also can you please make a seperate PR adding the test vectors?

facutuesca:
> Also can you please make a seperate PR adding the test vectors?

@alex : https://github.com/pyca/cryptography/pull/10438

facutuesca:
Rebased now that test vectors are merged

src/rust/src/backend/ec.rs
```diff
@@ -274,14 +274,35 @@ impl ECPrivateKey {
             ));
         }
 
-        let (data, _) = utils::calculate_digest_and_algorithm(
+        let (data, _algo) = utils::calculate_digest_and_algorithm(
```
alex, line 277: if this is used, the name shouldn't have a leading prefix

src/rust/src/backend/ec.rs
```diff
@@ -274,14 +274,35 @@ impl ECPrivateKey {
...
         )?;
 
         let mut signer = openssl::pkey_ctx::PkeyCtx::new(&self.pkey)?;
         signer.sign_init()?;
+        let deterministic: bool = signature_algorithm
+            .getattr(pyo3::intern!(py, "deterministic_signing"))?
+            .extract()?;
+        cfg_if::cfg_if! {
+            if #[cfg(CRYPTOGRAPHY_OPENSSL_320_OR_GREATER)]{
+                match deterministic {
```
alex, line 290: This can be an `if`, no need for a `match`

src/rust/src/backend/ec.rs
```diff
@@ -274,14 +274,35 @@ impl ECPrivateKey {
...
+                    true => {
+                        let hash_function_name = _algo.getattr(pyo3::intern!(py, "name"))?.extract::<&str>()?;
+                        let hash_function = openssl::md::Md::fetch(None, hash_function_name, None)?;
```
alex, line 293: There's a function in the hashes module that you can use for this.

tests/utils.py
```diff
@@ -701,6 +701,57 @@ def load_kasvs_ecdh_vectors(vector_data):
...
+            data["input"] = line.split("=")[1].strip(' "')
+        elif line.startswith("Output = "):
+            data["output"] = line.split("=")[1].strip()
+        elif line.startswith("Result = "):
+            data["verify_error"] = line.split("=")[1].strip() == "VERIFY_ERROR"
+
+        elif not line:
+            if data:
+                vectors.append(data)
+                data = dict()
```
alex, line 750:
```suggestion
                data = {}
```

docs/hazmat/primitives/asymmetric/ec.rst
```diff
@@ -47,6 +47,16 @@ Elliptic Curve Signature Algorithms
     :param algorithm: An instance of
         :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm`.
 
+    :param bool deterministic_signing: A boolean flag defaulting to ``False``
+        that specifies whether the signing procedure should be deterministic
+        or not, as defined in :rfc:`6979`.
```
alex, line 52: Docs should say that this has no impact on verification.

src/rust/src/backend/ec.rs
facutuesca: it's only used when OpenSSL >=3.2.0, so without it it complains when compiling with older OpenSSLs (or libre/boring). I could do something like this instead?:
```rust
cfg_if::cfg_if! {
    if #[cfg(CRYPTOGRAPHY_OPENSSL_320_OR_GREATER)]{
        let (data, _algo) = utils::calculate_digest_and_algorithm(
            py,
            data.as_bytes(),
            signature_algorithm.getattr(pyo3::intern!(py, "algorithm"))?,
        )?;
    } else {
        let (data, _) = utils::calculate_digest_and_algorithm(
            py,
            data.as_bytes(),
            signature_algorithm.getattr(pyo3::intern!(py, "algorithm"))?,
        )?;
    }    
````

File: src/rust/src/backend/ec.rs (hunk 3 above)
Comment (facutuesca) [reply src/rust/src/backend/ec.rs:293]: I tried, but I'm having trouble constructing the appropriate type. The `hashes` function I tried to use returns a `openssl::hash::MessageDigest`:
```rust
pub(crate) fn message_digest_from_algorithm(
    py: pyo3::Python<'_>,
    algorithm: &pyo3::PyAny,
) -> CryptographyResult<openssl::hash::MessageDigest> {
```
The function I'm calling that uses the hash function takes an `openssl::md::MdRef`:
```rust
pub fn set_signature_md(&self, md: &MdRef) -> Result<(), ErrorStack> 
```

I can convert one to the other by adding unsafe code:
```rust
let hash_function = hashes::message_digest_from_algorithm(py, _algo)?;
unsafe {
    signer.set_signature_md(&openssl::md::Md::from_ptr(hash_function.as_ptr().cast_mut()))?;
}
 ```
 But not sure if that's what we want (as opposed to directly using openssl::md::Md::fetch(), which already encapsulates the unsafe use)

File: docs/hazmat/primitives/asymmetric/ec.rst (hunk 5 above)
Comment (facutuesca) [reply docs/hazmat/primitives/asymmetric/ec.rst]: Fixed. The new docstring now says:
```rst
    :param bool deterministic_signing: A boolean flag defaulting to ``False``
        that specifies whether the signing procedure should be deterministic
        or not, as defined in :rfc:`6979`. This only impacts the signing
        process, verification is not affected (the verification process
        is the same for both deterministic and non-deterministic signed
        messages).
```

src/rust/src/backend/ec.rs
facutuesca: fixed

Tarnum-tst:
Merge, please!

CHANGELOG.rst
```diff
@@ -26,6 +26,7 @@ Changelog
   and :class:`~cryptography.hazmat.primitives.ciphers.algorithms.ARC4` into
   :doc:`/hazmat/decrepit/index` and deprecated them in the ``cipher`` module.
   They will be removed from the ``cipher`` module in 48.0.0.
+* Added support for deterministic ECDSA (:rfc:`6979`)
```
alex, line 29: Please add a link to the relevant code -- making ECDSA a link to the class should be enough.

src/rust/src/backend/ec.rs
alex: You can do `let _ = algo;` in the other branch.

tests/hazmat/primitives/test_ec.py
```diff
@@ -508,6 +513,70 @@ def test_signature_failures(self, backend, subtests):
                         signature, vector["message"], ec.ECDSA(hash_type())
                     )
 
+    def test_unsupported_deterministic_nonce(self, backend, subtests):
```
alex, line 516:
```suggestion
    def test_unsupported_deterministic_nonce(self, backend):
```

tests/hazmat/primitives/test_ec.py
```diff
@@ -508,6 +513,70 @@ def test_signature_failures(self, backend, subtests):
...
+            load_rfc6979_vectors,
+        )
+
+        for vector in vectors:
+            input = bytes(vector["input"], "utf-8")
+            output = bytes.fromhex(vector["output"])
+            key = bytes("\n".join(vector["key"]), "utf-8")
+            if "digest_sign" in vector:
+                algorithm = vector["digest_sign"]
+                assert algorithm in supported_hash_algorithms
```
alex, line 554:
```suggestion
```

Not needed, in tests it's fine for the next line to raise `KeyError`

tests/hazmat/primitives/test_ec.py
```diff
@@ -508,6 +513,70 @@ def test_signature_failures(self, backend, subtests):
...
+            ec.ECDSA(hashes.SHA256(), deterministic_signing=True)
+
+    def test_deterministic_nonce(self, backend, subtests):
+        if not backend.ecdsa_deterministic_supported():
+            pytest.skip(
+                f"ECDSA deterministic signing is not supported by this"
+                f" backend {backend}"
+            )
+
+        supported_hash_algorithms: typing.Dict[
```
alex, line 532: It's simpler if you just make the values here instance of the hash algorithm, rather than instantiating them below.

CHANGELOG.rst
facutuesca: fixed

src/rust/src/backend/ec.rs
facutuesca: oh, that is a lot better! fixed

tests/hazmat/primitives/test_ec.py
facutuesca: fixed

tests/hazmat/primitives/test_ec.py
facutuesca: removed

tests/hazmat/primitives/test_ec.py
facutuesca: fixed

facutuesca:
Rebased now that https://github.com/pyca/cryptography/pull/10486 is merged

Tarnum-tst:
@facutuesca 
openssl-sys is 0.9.101 in cryptography-openssl  currently.

src/rust/src/backend/ec.rs
```diff
@@ -273,15 +273,37 @@ impl ECPrivateKey {
...
+                    // Setting a deterministic nonce type requires to explicitly set the hash function.
+                    // See https://github.com/openssl/openssl/issues/23205
+                    signer.set_signature_md(&hash_function)?;
+                    signer.set_nonce_type(openssl::pkey_ctx::NonceType::DETERMINISTIC_K)?;
+                } else {
+                    signer.set_nonce_type(openssl::pkey_ctx::NonceType::RANDOM_K)?;
+                }
+            } else {
+                let _ = algo;
+                assert!(!deterministic);
```
reaperhulk, line 303: So this assert won't trigger unless you do something dumb (instantiate ECDSA then set ._deterministic = True yourself on non-OpenSSL 3.2.0) but I'd rather it wasn't possible to cause program termination even if you do something truly stupid.

tests/hazmat/primitives/test_ec.py
```diff
@@ -508,6 +513,67 @@ def test_signature_failures(self, backend, subtests):
...
+            "SHA512": hashes.SHA512(),
+        }
+        vectors = load_vectors_from_file(
+            os.path.join(
+                "asymmetric", "ECDSA", "RFC6979", "evppkey_ecdsa_rfc6979.txt"
+            ),
+            load_rfc6979_vectors,
+        )
+
+        for vector in vectors:
```
reaperhulk, line 546: It looks like you added the subtests fixture to this test but forgot to add the `with subtests.test():` inside this for loop.

facutuesca: yup, fixed!

reaperhulk (approved):

PR #10373, opened 2024-02-09
Fix ASN.1 issues in PKCS#7 and S/MIME signing

facutuesca:
This PR fixes two issues with the ASN.1 generated when signing using PKCS#7:

## First issue
The [current implementation](https://github.com/pyca/cryptography/blob/bfcdfbefb32c5a9786ef66d4eb0777f70ae5943b/src/rust/src/pkcs7.rs#L103-L109) defines the SMIMECapabilities attribute so that its value is a SEQUENCE of all the algorithm OIDs that are supported. This is what the ASN.1 currently looks like for a signed message using `cryptography`:
```asn.1
  U.P.SEQUENCE {
     U.P.OBJECTIDENTIFIER 1.2.840.113549.1.9.15 (S/MIME capabilities)
     U.P.SET {
        U.P.SEQUENCE {
           U.P.OBJECTIDENTIFIER 2.16.840.1.101.3.4.1.42 (aes256-CBC-PAD)
           U.P.OBJECTIDENTIFIER 2.16.840.1.101.3.4.1.22 (aes192-CBC-PAD)
           U.P.OBJECTIDENTIFIER 2.16.840.1.101.3.4.1.2 (aes128-CBC-PAD)
        }
     }
  }
```

However, the S/MIME v3 spec (RFC 2633) specifies that each algorithm should be specified in its own SEQUENCE:

```asn.1
SMIMECapabilities ::= SEQUENCE OF SMIMECapability

SMIMECapability ::= SEQUENCE {
   capabilityID OBJECT IDENTIFIER,
   parameters ANY DEFINED BY capabilityID OPTIONAL }
```

([RFC 2633, Appendix A](https://datatracker.ietf.org/doc/html/rfc2633#appendix-A))

The spec matches what OpenSSL outputs (using `openssl smime -sign`):
```asn.1
  U.P.SEQUENCE {
     U.P.OBJECTIDENTIFIER 1.2.840.113549.1.9.15 (S/MIME capabilities)
     U.P.SET {
        U.P.SEQUENCE {
           U.P.SEQUENCE {
              U.P.OBJECTIDENTIFIER 2.16.840.1.101.3.4.1.42 (aes256-CBC-PAD)
           }
           U.P.SEQUENCE {
              U.P.OBJECTIDENTIFIER 2.16.840.1.101.3.4.1.22 (aes192-CBC-PAD)
           }
           U.P.SEQUENCE {
              U.P.OBJECTIDENTIFIER 2.16.840.1.101.3.4.1.2 (aes128-CBC-PAD)
           }
           U.P.SEQUENCE {
              U.P.OBJECTIDENTIFIER 1.2.840.113549.3.7 (id_des_EDE3_CBC)
           }
          .....
```         

This PR changes the implementation so that each algorithm is inside its own SEQUENCE. After the changes in this PR, the ASN.1 output looks like:
```asn.1
  U.P.SEQUENCE {
     U.P.OBJECTIDENTIFIER 1.2.840.113549.1.9.15 (S/MIME capabilities)
     U.P.SET {
        U.P.SEQUENCE {
           U.P.SEQUENCE {
              U.P.OBJECTIDENTIFIER 2.16.840.1.101.3.4.1.42 (aes256-CBC-PAD)
           }
           U.P.SEQUENCE {
              U.P.OBJECTIDENTIFIER 2.16.840.1.101.3.4.1.22 (aes192-CBC-PAD)
           }
           U.P.SEQUENCE {
              U.P.OBJECTIDENTIFIER 2.16.840.1.101.3.4.1.2 (aes128-CBC-PAD)
           }
        }
     }
  }
```

## Second issue
There is an issue with the RSA OID used for signing PKCS#7/SMIME, in the `SignatureAlgorithmIdentifier` field.
    
The current implementation computes the algorithm identifier used in the `digest_encryption_algorithm` PKCS#7 field
(or `SignatureAlgorithmIdentifier` in S/MIME) based on both the algorithm used to sign (e.g. RSA) and the digest algorithm (e.g. SHA512).
    
This is correct for ECDSA signatures, where the OIDs used include the digest algorithm (e.g: `ecdsa-with-SHA512`). However, due to historical reasons, when signing with RSA the OID specified should be the one
corresponding to just RSA (`"1.2.840.113549.1.1.1" rsaEncryption`), rather than OIDs which also include the digest algorithm (such as `"1.2.840.113549.1.1.13" sha512WithRSAEncryption`).

This means that the logic to compute the algorithm identifier is the same except when signing with RSA, in which case the OID will always be `rsaEncryption`. This is consistent with the OpenSSL implementation, and the RFCs that define PKCS#7 and S/MIME.

See [RFC 3851 (section 2.2)](https://www.rfc-editor.org/rfc/rfc3851#section-2.2), and [RFC 3370 (section 3.2)](https://www.rfc-editor.org/rfc/rfc3370#section-3.2) for more details.

### DIFF
--- a/src/rust/src/pkcs7.rs
+++ b/src/rust/src/pkcs7.rs
@@ -269,13 +269,13 @@ fn compute_pkcs7_signature_algorithm<'p>(
     rsa_padding: &'p pyo3::PyAny,
 ) -> pyo3::PyResult<common::AlgorithmIdentifier<'static>> {
     let key_type = x509::sign::identify_key_type(py, private_key)?;
-    let has_pss_padding = !rsa_padding.is_none() && rsa_padding.is_instance(types::PSS.get(py)?)?;
+    let has_pss_padding = rsa_padding.is_instance(types::PSS.get(py)?)?;
     // For RSA signatures (with no PSS padding), the OID is always the same no matter the
     // digest algorithm. See RFC 3370 (section 3.2).
     if key_type == x509::sign::KeyType::Rsa && !has_pss_padding {
         Ok(common::AlgorithmIdentifier {
             oid: asn1::DefinedByMarker::marker(),
-            params: common::AlgorithmParameters::Rsa(None),
+            params: common::AlgorithmParameters::Rsa(Some(())),
         })
     } else {
         x509::sign::compute_signature_algorithm(py, private_key, hash_algorithm, rsa_padding)
--- a/src/rust/src/pkcs7.rs
+++ b/src/rust/src/pkcs7.rs
@@ -205,7 +205,7 @@ fn sign_and_serialize<'p>(
             },
             digest_algorithm: digest_alg,
             authenticated_attributes: authenticated_attrs,
-            digest_encryption_algorithm: x509::sign::compute_signature_algorithm(
+            digest_encryption_algorithm: compute_pkcs7_signature_algorithm(
                 py,
                 py_private_key,
                 py_hash_alg,
@@ -259,6 +259,26 @@ fn sign_and_serialize<'p>(
     } else {
         // Handles the DER, PEM, and error cases
         encode_der_data(py, "PKCS7".to_string(), ci_bytes, encoding)
+    }
+}
+
+fn compute_pkcs7_signature_algorithm<'p>(
+    py: pyo3::Python<'p>,
+    private_key: &'p pyo3::PyAny,
+    hash_algorithm: &'p pyo3::PyAny,
+    rsa_padding: &'p pyo3::PyAny,
+) -> pyo3::PyResult<common::AlgorithmIdentifier<'static>> {
+    let key_type = x509::sign::identify_key_type(py, private_key)?;
+    let has_pss_padding = !rsa_padding.is_none() && rsa_padding.is_instance(types::PSS.get(py)?)?;
+    // For RSA signatures (with no PSS padding), the OID is always the same no matter the
+    // digest algorithm. See RFC 3370 (section 3.2).
+    if key_type == x509::sign::KeyType::Rsa && !has_pss_padding {
+        Ok(common::AlgorithmIdentifier {
+            oid: asn1::DefinedByMarker::marker(),
+            params: common::AlgorithmParameters::Rsa(None),
+        })
+    } else {
+        x509::sign::compute_signature_algorithm(py, private_key, hash_algorithm, rsa_padding)
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/rust/src/pkcs7.rs
```diff
@@ -262,6 +262,26 @@ fn sign_and_serialize<'p>(
...
+    rsa_padding: &'p pyo3::PyAny,
+) -> pyo3::PyResult<common::AlgorithmIdentifier<'static>> {
+    let key_type = x509::sign::identify_key_type(py, private_key)?;
+    let has_pss_padding = !rsa_padding.is_none() && rsa_padding.is_instance(types::PSS.get(py)?)?;
+    // For RSA signatures (with no PSS padding), the OID is always the same no matter the
+    // digest algorithm. See RFC 3370 (section 3.2).
+    if key_type == x509::sign::KeyType::Rsa && !has_pss_padding {
+        Ok(common::AlgorithmIdentifier {
+            oid: asn1::DefinedByMarker::marker(),
+            params: common::AlgorithmParameters::Rsa(None),
```
alex, line 278:
```suggestion
            params: common::AlgorithmParameters::Rsa(Some(())),
```

src/rust/src/pkcs7.rs
```diff
@@ -262,6 +262,26 @@ fn sign_and_serialize<'p>(
...
 }
 
+fn compute_pkcs7_signature_algorithm<'p>(
+    py: pyo3::Python<'p>,
+    private_key: &'p pyo3::PyAny,
+    hash_algorithm: &'p pyo3::PyAny,
+    rsa_padding: &'p pyo3::PyAny,
+) -> pyo3::PyResult<common::AlgorithmIdentifier<'static>> {
+    let key_type = x509::sign::identify_key_type(py, private_key)?;
+    let has_pss_padding = !rsa_padding.is_none() && rsa_padding.is_instance(types::PSS.get(py)?)?;
```
alex, line 272:
```suggestion
    let has_pss_padding = rsa_padding.is_instance(types::PSS.get(py)?)?;
```

I don't think the `is_none()` check is required? None isn't an instance of PSS :-)

facutuesca: Ah, I got that code from https://github.com/pyca/cryptography/blob/50ea0faab70d2830e7d89756731fecf9ca64528e/src/rust/src/x509/sign.rs#L135-L137. Should I remove it from there as well?

alex: Probably that should be fixed, but no need to do it in this PR.

src/rust/src/pkcs7.rs
facutuesca: fixed

src/rust/src/pkcs7.rs
facutuesca: done

facutuesca:
@alex I also added an item to the CHANGELOG

alex (approved):

PR #10407, opened 2024-02-17
support RC2-CBC

reaperhulk:
This PR supports a bad old algorithm to support a scapy use case (and unblock full oxidization of the cipher layer), but does not expose support for effective key bits or any key length other than 128-bit. CBC support only -- no other modes.

I've got mixed feelings about adding documentation for this, so there isn't any right now.

### DIFF
--- a/src/cryptography/hazmat/decrepit/ciphers/algorithms.py
+++ b/src/cryptography/hazmat/decrepit/ciphers/algorithms.py
@@ -90,3 +90,18 @@ class IDEA(BlockCipherAlgorithm):
     @property
     def key_size(self) -> int:
         return len(self.key) * 8
+
+
+# This class only allows RC2 with a 128-bit key. No support for
+# effective key bits or other key sizes is provided.
+class RC2(BlockCipherAlgorithm):
+    name = "RC2"
+    block_size = 64
+    key_sizes = frozenset([128])
+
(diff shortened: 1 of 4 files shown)


src/cryptography/hazmat/decrepit/ciphers/algorithms.py
```diff
@@ -90,3 +90,18 @@ def __init__(self, key: bytes):
     @property
     def key_size(self) -> int:
         return len(self.key) * 8
+
+
+# This class only allows RC2 with a 128-bit key. No support for
+# effective key bits or other key sizes is provided.
+class RC2(BlockCipherAlgorithm):
```
alex, line 97: Should it be named `RC2_128` or something? Or is the idea that if we ever _did_ want to support more key sizes, we'd just add more args to the constructor?

alex (approved):

src/cryptography/hazmat/decrepit/ciphers/algorithms.py
reaperhulk: Yeah it's actually not very hard to support the effective key bits stuff so I'd rather just have the "generic" class. I had support and ripped it out since I want to see someone come with a use case.

alex (approved):

PR #10408, opened 2024-02-17
Fix rust warnings when building with BoringSSL

alex:


noxfile.py
```diff
@@ -292,7 +292,7 @@ def local(session):
         "noxfile.py",
     )
 
-    install(session, "cryptography @ .")
```
reaperhulk, line 295: Did you mean to revert this?

alex: Yeah. Turns out this syntax doesn't work with pip, so we can't actually switch until uv is usable.

reaperhulk (approved):

PR #10458, opened 2024-02-23
Stop running linkcheck on every push - its pointless

alex:


reaperhulk (approved):
It feels fitting that a change to reduce flakiness has a PyPy flake failure as well. RIP reliable CI

PR #10487, opened 2024-02-26
verification: add RFC822Name

woodruffw:
This is a backing piece for #10345. Once merged, I'll do a follow up PR for `RFC822Constraint` for the Name Constraint side.

Flagging one thing: this implementation doesn't currently bother with the "quoted" form of mailbox names. These are treated as invalid addresses for the time being, and I've added a testcase as a backstop for that behavior so that it doesn't change unexpectedly.

(I can try with the quoted form, but it'll probably make a PR a decent bit bigger and may just be worth a crate that gets it right. They're also uncommon and the relevant RFCs advise against using them for compatibility reasons.)

### DIFF
--- a/src/rust/cryptography-x509-verification/src/types.rs
+++ b/src/rust/cryptography-x509-verification/src/types.rs
@@ -312,7 +312,10 @@ impl IPConstraint {
 ///
 /// [RFC 822 6.1]: https://datatracker.ietf.org/doc/html/rfc822#section-6.1
 /// [RFC 2821 4.1.2]: https://datatracker.ietf.org/doc/html/rfc2821#section-4.1.2
-pub struct RFC822Name<'a>((IA5String<'a>, DNSName<'a>));
+pub struct RFC822Name<'a> {
+    pub mailbox: IA5String<'a>,
+    pub domain: DNSName<'a>,
+}
 
 impl<'a> RFC822Name<'a> {
     pub fn new(value: &'a str) -> Option<Self> {
@@ -338,15 +341,10 @@ impl<'a> RFC822Name<'a> {
             }
         }
 
-        DNSName::new(domain).map(|domain| Self((local_part, domain)))
-    }
-
-    pub fn mailbox(&self) -> &str {
-        (self.0).0.as_str()
-    }
-
-    pub fn domain(&self) -> &DNSName<'_> {
-        &(self.0).1
+        Some(Self {
+            mailbox: local_part,
+            domain: DNSName::new(domain)?,
+        })
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/rust/cryptography-x509-verification/src/types.rs
```diff
@@ -5,6 +5,11 @@
 use std::net::IpAddr;
 use std::str::FromStr;
 
+use asn1::IA5String;
+
+// RFC 2822 3.2.4
+static ATEXT_CHARS: &str = "!#$%&'*+-/=?^_`{|}~";
```
woodruffw, line 11: This might make more sense/be more efficient as `&[u8]` instead.

src/rust/cryptography-x509-verification/src/types.rs
```diff
@@ -298,9 +303,56 @@ impl IPConstraint {
...
+            if component.is_empty()
+                || !component
+                    .chars()
+                    .all(|c| c.is_ascii_alphanumeric() || ATEXT_CHARS.contains(c))
+            {
+                return None;
+            }
+        }
+
+        DNSName::new(domain).map(|domain| Self((local_part, domain)))
```
alex, line 341: I think it'd much easier to read `Self((local_part, DNSName::new(domain)?))`

src/rust/cryptography-x509-verification/src/types.rs
```diff
@@ -298,9 +303,56 @@ impl IPConstraint {
...
+/// An `RFC822Name` represents an email address, as defined in [RFC 822 6.1]
+/// and as amended by [RFC 2821 4.1.2]. In particular, it represents the `Mailbox`
+/// rule from RFC 2821's grammar.
+///
+/// This type does not currently support the quoted local-part form; email
+/// addresses that use this form will be rejected.
+///
+/// [RFC 822 6.1]: https://datatracker.ietf.org/doc/html/rfc822#section-6.1
+/// [RFC 2821 4.1.2]: https://datatracker.ietf.org/doc/html/rfc2821#section-4.1.2
+pub struct RFC822Name<'a>((IA5String<'a>, DNSName<'a>));
```
alex, line 315: Why a tuple instead of two fields?

src/rust/cryptography-x509-verification/src/types.rs
```diff
@@ -587,4 +639,65 @@ mod tests {
...
+            "@@example.com",
+            // Invalid local part (invalid characters).
+            "lol\"lol@example.com",
+            "lol;lol@example.com",
+            "🙈@example.com",
+            // Intentionally unsupported quoted local parts.
+            "\"validbutunsupported\"@example.com",
+        ];
+
+        for case in bad_cases {
```
alex, line 679: You can just loop directly over the array, no need for a variable

src/rust/cryptography-x509-verification/src/types.rs
woodruffw: No good reason! Changing.

alex (approved):

PR #10497, opened 2024-02-28
verification: add RFC822Constraint

woodruffw:
Subcomponent of #10345, follow-on from #10487.

Adds `RFC822Constraint`, which expresses the three kinds of email address constraints defined in RFC 5280: exact mailbox, exact domain ("on domain"), and strict subdomain ("in domain").

### DIFF
--- a/src/rust/cryptography-x509-verification/src/types.rs
+++ b/src/rust/cryptography-x509-verification/src/types.rs
@@ -323,17 +323,10 @@ impl IPConstraint {
 ///
 /// [RFC 822 6.1]: https://datatracker.ietf.org/doc/html/rfc822#section-6.1
 /// [RFC 2821 4.1.2]: https://datatracker.ietf.org/doc/html/rfc2821#section-4.1.2
+#[derive(PartialEq)]
 pub struct RFC822Name<'a> {
     pub mailbox: IA5String<'a>,
     pub domain: DNSName<'a>,
-}
-
-impl<'a> PartialEq for RFC822Name<'a> {
-    fn eq(&self, other: &Self) -> bool {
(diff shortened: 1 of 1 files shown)


src/rust/cryptography-x509-verification/src/types.rs
```diff
@@ -317,6 +328,14 @@ pub struct RFC822Name<'a> {
     pub domain: DNSName<'a>,
 }
 
+impl<'a> PartialEq for RFC822Name<'a> {
```
alex, line 331: Can't this just be `derive(PartialEq)`?

src/rust/cryptography-x509-verification/src/types.rs
```diff
@@ -348,10 +367,45 @@ impl<'a> RFC822Name<'a> {
...
+    /// For example, `InDomain("example.com")` will match `foo@bar.example.com`
+    /// but not `foo@example.com`, since `bar.example.com` is in `example.com`
+    /// but `example.com` is not within itself.
+    InDomain(DNSName<'a>),
+}
+
+impl<'a> RFC822Constraint<'a> {
+    pub fn new(constraint: &'a str) -> Option<Self> {
+        if let Some(constraint) = constraint.strip_prefix('.') {
+            DNSName::new(constraint).map(Self::InDomain)
```
alex, line 386: Please don't use map for this stuff, it's just easier to read: `Self::InDomain(DNSName::new(constraint)?)`

src/rust/cryptography-x509-verification/src/types.rs
woodruffw: Ah yep, I think so.

alex (approved):

PR #10544, opened 2024-03-06
Add Zero padding method

GreyElaina:
I'm new here so if there are problems, pointing out is welcome. :)

Unit tests & Docs will come soon. It's too late in my region and I need time to explore workflow.
miss maturin :p


alex:
a) can you describe where this padding method is used
b) prima facie this looks ambiguous if the content can contain a nul byte. Am I wrong

GreyElaina:
> a) can you describe where this padding method is used b) prima facie this looks ambiguous if the content can contain a nul byte. Am I wrong

hmmm, this pr looks like a consequence of drunkenness: it only took me 15 minutes to implement this thing, mainly so that it can be used in the same way as other implementations such as `PKCS7`
i only remember i see `ZeroPadding` in crypto-js but in cryptography. i donot find it and i was new to use cryptography.
as a "consequence of drunkenness", haha, i'll close it, this algorithm is so annoying (it donot tell us the padding size).

PR #10570, opened 2024-03-11
verification: forbid unsupported NCs

woodruffw:
...rather than silently ignoring them.

Needs coverage, which I'll add with https://github.com/C2SP/x509-limbo/pull/228 once I merge that.

### DIFF
--- a/.github/actions/fetch-vectors/action.yml
+++ b/.github/actions/fetch-vectors/action.yml
@@ -17,4 +17,4 @@ runs:
         repository: "C2SP/x509-limbo"
         path: "x509-limbo"
         # Latest commit on the x509-limbo main branch, as of Mar 11, 2024.
-        ref: "1bf45ad24ef5d2c0d7bc5082a9c690998cb595da" # x509-limbo-ref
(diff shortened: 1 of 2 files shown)


alex:
Do we only want to reject matching pairs, or reject all validations with an other name constraint?

woodruffw:
> Do we only want to reject matching pairs, or reject all validations with an other name constraint?

I _think_ just matching pairs: RFC 5280 says that NCs only need to be evaluated if a subject of the name form appears, so rejecting any cert that has an OtherName NC could (in practice, unlikely) prematurely exclude certs/chains that don't actually have any OtherName SANs.

woodruffw:
I suppose that means I should add a testcase for "validator doesn't reject chain with an OtherName constraint when there's no OtherName SANs" 🙂

alex:
Ok, makes sense, and yeah always test cases are good.

woodruffw:
Merged both new testcases and the CI is green, so this should be good to go 🙂

alex (approved):

PR #10627, opened 2024-03-22
Pass -fmacro-prefix-map to cc to not leak paths into the binary

EchterAgo:
Without this compiling the CFFI generated `_openssl.c` file embeds the build path into the binary. When installed using PyPi this path is random, which makes the resulting binary not reproducible.

### DIFF
--- a/src/rust/cryptography-cffi/build.rs
+++ b/src/rust/cryptography-cffi/build.rs
@@ -69,6 +69,13 @@ fn main() {
         .flag_if_supported("-Wno-error=sign-conversion")
         .flag_if_supported("-Wno-unused-parameter");
 
+    // We use the `-fmacro-prefix-map` option to replace the output directory in macros with a dot.
+    // This is because we don't want a potentially random build path to end up in the binary because
+    // CFFI generated code uses the __FILE__ macro in its debug messages.
+    if let Some(out_dir_str) = Path::new(&out_dir).to_str() {
+        build.flag_if_supported(format!("-fmacro-prefix-map={}=.", out_dir_str).as_str());
+    }
+
     for python_include in env::split_paths(&python_includes) {
         build.include(python_include);
     }


EchterAgo:
Note that for a completely reproducible build I also need to enable `cargo`s unstable `trim-path` feature, which I currently do in the `cargo` `config.toml`:

```
mkdir -p "$HOME/.cargo"
cat > "$HOME/.cargo/config.toml" << EOF
[unstable]
trim-paths = true
EOF
```

EchterAgo:
The reason why the path ends up in the binary in the first place is that CFFI uses assert for the `_cffi_type` macro in the generated `_openssl.c`, which in turn embeds `__FILE__` macro in the debug message:

```c
  if ((DAT_003df240 & 1) != 0) {
                    /* WARNING: Subroutine does not return */
    __assert_fail("(((uintptr_t)_cffi_types[84]) & 1) == 0",
                  "/tmp/pip-install-1f6ut9qw/cryptography_aae62eb4378e4c688a9d878cde9052fa/src/rust/ target/release/build/cryptography-cffi-eab8ad0c7ada57ad/out/_openssl.c"
                  ,0x6130,"_cffi_f_i2d_X509_CRL_bio");
  }
```

src/rust/cryptography-cffi/build.rs
```diff
@@ -69,6 +69,10 @@ fn main() {
         .flag_if_supported("-Wno-error=sign-conversion")
         .flag_if_supported("-Wno-unused-parameter");
 
+    if let Some(out_dir_str) = Path::new(&out_dir).to_str() {
+        build.flag_if_supported(format!("-fmacro-prefix-map={}=.", out_dir_str).as_str());
```
alex, line 73: Can you please add a comment explaining what this is doing.

EchterAgo: Added

alex (approved):

alex:
Thanks!

Issue #10676, opened 2024-04-01
Remove gil-refs feature for pyo3


alex:
This requires fixing hundreds of warnings. Ideally we will do this in many small PRs, perhaps file at a time.

If you're interested in contributing:

- Edit `src/rust/Cargo.toml` to remove the `gil-refs` feature
- Try compiling (`nox -e local`)
- Fix an error
- Send a PR!

### Progress:

- 4/3: 5 previous errors; 531 warnings emitted
- 4/4: 5 previous errors; 359 warnings emitted
- 4/5: 3 previous errors; 105 warnings emitted
- 4/6: 25 previous errors; 1 warning emitted

```[tasklist]
### Phase 1
- [ ] https://github.com/pyca/cryptography/pull/10711
- [ ] https://github.com/pyca/cryptography/pull/10696
- [ ] https://github.com/pyca/cryptography/pull/10700
- [ ] https://github.com/pyca/cryptography/pull/10703
- [ ] https://github.com/pyca/cryptography/pull/10702
- [ ] https://github.com/pyca/cryptography/pull/10714
- [ ] https://github.com/pyca/cryptography/pull/10715
- [ ] https://github.com/pyca/cryptography/pull/10721
- [ ] https://github.com/pyca/cryptography/pull/10722
- [ ] https://github.com/pyca/cryptography/pull/10725
- [x] `src/backend/hashes.rs`
- [ ] https://github.com/pyca/cryptography/pull/10726
- [ ] https://github.com/pyca/cryptography/pull/10717
- [ ] https://github.com/pyca/cryptography/pull/10719
- [ ] https://github.com/pyca/cryptography/pull/10728
- [ ] https://github.com/pyca/cryptography/pull/10729
- [ ] https://github.com/pyca/cryptography/pull/10678
- [ ] https://github.com/pyca/cryptography/pull/10730
- [ ] https://github.com/pyca/cryptography/pull/10723
- [ ] https://github.com/pyca/cryptography/pull/10684
- [ ] https://github.com/pyca/cryptography/pull/10686
- [ ] https://github.com/pyca/cryptography/pull/10712
- [ ] https://github.com/pyca/cryptography/pull/10745
- [ ] https://github.com/pyca/cryptography/pull/10688
- [ ] https://github.com/pyca/cryptography/pull/10687
- [x] `src/pkcs7.rs`
- [ ] https://github.com/pyca/cryptography/pull/10754
- [x] `src/x509/certificate.rs`
[...]

PR #10679, opened 2024-04-01
sign: bound-ify sig_alg APIs

woodruffw:
See #10676.

This unfortunately taints a few certificate, CRL, etc. APIs in the process. 

However, each is a singular top-level API, so the diff isn't too bad.

### DIFF
--- a/src/rust/src/types.rs
+++ b/src/rust/src/types.rs
@@ -28,6 +28,21 @@ impl LazyPyImport {
             })
             .map(|p| p.as_ref(py))
     }
+
+    pub fn get_bound<'p>(
+        &'p self,
+        py: pyo3::Python<'p>,
+    ) -> pyo3::PyResult<pyo3::Bound<'p, pyo3::PyAny>> {
+        let p = self.value.get_or_try_init(py, || {
+            let mut obj = py.import(self.module)?.as_ref();
+            for name in self.names {
+                obj = obj.getattr(*name)?;
+            }
+            obj.extract()
+        })?;
+
+        Ok(p.clone().into_bound(py))
+    }
 }
 
 pub static DATETIME_DATETIME: LazyPyImport = LazyPyImport::new("datetime", &["datetime"]);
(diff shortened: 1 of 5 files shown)


src/rust/src/types.rs
```diff
@@ -28,6 +28,21 @@ impl LazyPyImport {
             })
             .map(|p| p.as_ref(py))
     }
+
+    pub fn get_bound<'p>(
```
alex, line 32: `self.get()` should just call this I think.

src/rust/src/x509/sign.rs
```diff
@@ -442,20 +444,22 @@ fn hash_oid_py_hash(
...
+) -> CryptographyResult<pyo3::Bound<'p, pyo3::PyAny>> {
     let sig_oids_to_hash = types::SIG_OIDS_TO_HASH.get(py)?;
     match &signature_algorithm.params {
         common::AlgorithmParameters::RsaPss(opt_pss) => {
             let pss = opt_pss.as_ref().ok_or_else(|| {
                 pyo3::exceptions::PyValueError::new_err("Invalid RSA PSS parameters")
             })?;
-            hash_oid_py_hash(py, pss.hash_algorithm.oid().clone())
+            Ok(hash_oid_py_hash(py, pss.hash_algorithm.oid().clone())?
+                .to_object(py)
```
alex, line 455: I think you want `as_borrowed().to_owned()`, not `to_object().into_bound()`

src/rust/src/x509/sign.rs
```diff
@@ -442,20 +444,22 @@ fn hash_oid_py_hash(
...
+                .into_bound(py))
         }
         _ => {
             let py_sig_alg_oid = oid_to_py_oid(py, signature_algorithm.oid())?;
             let hash_alg = sig_oids_to_hash.get_item(py_sig_alg_oid);
             match hash_alg {
-                Ok(data) => Ok(data),
+                Ok(data) => Ok(data.to_object(py).into_bound(py)),
```
alex, line 462: Same

src/rust/src/types.rs
woodruffw: I did this via `Ok(self.get_bound(py)?.into_gil_ref())`, not sure if that's optimal :slightly_smiling_face:

alex (approved):

PR #10732, opened 2024-04-05
Convert `src/x509/common.rs` to new pyo3 APIs

facutuesca:

This finishes with all the migration warnings for `src/x509/common.rs`

One thing to pay attention to, in `common.rs:455`:
```rust
// before
extn_value: py_data.as_bytes(),

// after
extn_value: py_data.extract()?,
```
which I'm not 100% sure is correct.

cc @alex @reaperhulk


alex (dismissed):

alex:
Looks like there's coverage issues

facutuesca:
> Looks like there's coverage issues

The issue due to a `?` now being on a new line on its own, due to `cargo fmt`:
image

I added a intermediate variable to make the line shorter again. LMK if we want to actually add a test case that triggers that `?`.

alex (approved):

facutuesca:
I missed another coverage failure in the same file, same as the previous one

alex (approved):

PR #10733, opened 2024-04-05
Start converting `src/x509/csr.rs` to new pyo3 APIs

facutuesca:

This fixes all **except one** of the migration warnings for `src/x509/csr.rs`

This is the one warning I wasn't sure how to fix:
```
error[E0277]: the trait bound `(pyo3::Bound<'_, pyo3::PyAny>, &[u8], std::option::Option<u8>): pyo3::FromPyObject<'_>` is not satisfied
    --> src/x509/csr.rs:336:22
     |
336  |             py_attr?.extract()?;
     |                      ^^^^^^^ the trait `pyo3::FromPyObject<'_>` is not implemented for `(pyo3::Bound<'_, pyo3::PyAny>, &[u8], std::option::Option<u8>)`, which is required by `(pyo3::Bound<'_, pyo3::PyAny>, &[u8], std::option::Option<u8>): pyo3::conversion::FromPyObjectBound<'_, '_>`
     |
```

cc @alex @reaperhulk

### DIFF
--- a/src/rust/src/x509/csr.rs
+++ b/src/rust/src/x509/csr.rs
@@ -8,7 +8,8 @@ use std::hash::{Hash, Hasher};
 use asn1::SimpleAsn1Readable;
 use cryptography_x509::csr::{check_attribute_length, Attribute, CertificationRequestInfo, Csr};
 use cryptography_x509::{common, oid};
-use pyo3::{IntoPy, PyNativeType};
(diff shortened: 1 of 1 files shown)


alex (commented):
Yes, that's one of the examples that will be hard to fix, and taht we should defer.

alex (approved):

PR #10736, opened 2024-04-05
Start converting `src/x509/verify.rs` to new pyo3 APIs

facutuesca:

This fixes all **except one** of the migration warnings for `src/x509/verify.rs`. The remaining warning should be fixed by the changes on `verify.rs` on this PR: [link](https://github.com/pyca/cryptography/pull/10734/files#diff-6207df575875d8f1ffd6b2a7999a84061967780e6936ec7a25ad7386ea4d6e83)

cc @alex @reaperhulk

### DIFF
--- a/src/rust/src/x509/verify.rs
+++ b/src/rust/src/x509/verify.rs
@@ -379,7 +379,7 @@ fn build_subject_owner(
             .call0()?
             .downcast::<pyo3::types::PyBytes>()?
             .clone();
-        Ok(SubjectOwner::IPAddress(value.clone().unbind()))
+        Ok(SubjectOwner::IPAddress(value.unbind()))
     } else {
         Err(pyo3::exceptions::PyTypeError::new_err(
             "unsupported subject type",
--- a/src/rust/src/x509/verify.rs
+++ b/src/rust/src/x509/verify.rs
@@ -11,7 +11,7 @@ use cryptography_x509_verification::{
     trust_store::Store,
     types::{DNSName, IPAddress},
 };
-use pyo3::IntoPy;
+use pyo3::prelude::{PyAnyMethods, PyListMethods, PyStringMethods};
 
 use crate::backend::keys;
 use crate::error::{CryptographyError, CryptographyResult};
@@ -75,7 +75,7 @@ impl PolicyBuilder {
     fn time(
         &self,
         py: pyo3::Python<'_>,
-        new_time: &pyo3::PyAny,
+        new_time: &pyo3::Bound<'_, pyo3::PyAny>,
     ) -> CryptographyResult<PolicyBuilder> {
         if self.time.is_some() {
             return Err(CryptographyError::from(
@@ -85,7 +85,7 @@ impl PolicyBuilder {
             ));
         }
         Ok(PolicyBuilder {
-            time: Some(py_to_datetime(py, new_time)?),
+            time: Some(py_to_datetime(py, new_time.clone().into_gil_ref())?),
             store: self.store.as_ref().map(|s| s.clone_ref(py)),
             max_chain_depth: self.max_chain_depth,
         })
@@ -273,7 +273,7 @@ impl PyClientVerifier {
         )
         .map_err(|e| VerificationError::new_err(format!("validation failed: {e:?}")))?;
 
-        let py_chain = pyo3::types::PyList::empty(py);
+        let py_chain = pyo3::types::PyList::empty_bound(py);
         for c in &chain {
             py_chain.append(c.extra())?;
         }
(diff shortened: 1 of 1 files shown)


alex (approved):

facutuesca:
Ah hold off the review, I pushed too early, still need to fix a couple of errors

src/rust/src/x509/verify.rs
```diff
@@ -366,21 +366,19 @@ fn build_subject_owner(
     py: pyo3::Python<'_>,
     subject: &pyo3::Py<pyo3::PyAny>,
 ) -> pyo3::PyResult<SubjectOwner> {
-    let subject = subject.as_ref(py);
+    let subject = subject.bind(py);
```
alex, line 369: If you add a `clone()` here, I don't believe all the temporary variables are required below.

facutuesca: I added:
```rust
    let subject = subject.bind(py).clone();
```
But I still get:

```
    |
372 |           let value = subject
    |  _____________________^
373 | |             .getattr(pyo3::intern!(py, "value"))?
    | |_________________________________________________^ creates a temporary value which is freed while still in use
374 |               .downcast::<pyo3::types::PyString>()?;
    |                                                    - temporary value is freed at the end of this statement
375 |
376 |           Ok(SubjectOwner::DNSName(value.to_str()?.to_owned()))
    |                                    ----- borrow later used here
    |
    = note: consider using a `let` binding to create a longer lived value
```

alex: Oh. Put a clone at the end of that expression. It's more readable than the intermediate bounds.

facutuesca:
@alex Done, please check the last commit: I had to remove a call to `Bound::to_str` because it's [not supported in Python < 3.10](https://docs.rs/pyo3/0.21.1/pyo3/prelude/struct.Bound.html#method.to_str).
The change is:

```rust
// before
        let value = subject
            .getattr(pyo3::intern!(py, "value"))?
            .downcast::<pyo3::types::PyString>()?
            .clone();
        Ok(SubjectOwner::DNSName(value.to_str()?.to_owned()))

// after
        let value = subject
            .getattr(pyo3::intern!(py, "value"))?
            .extract::<String>()?;
        Ok(SubjectOwner::DNSName(value))
```        

So I changed the `downcast + clone` to a `extract::<String>::` call. Do you think that makes sense?

src/rust/src/x509/verify.rs
```diff
@@ -366,21 +366,20 @@ fn build_subject_owner(
...
+    } else if subject.is_instance(&types::IP_ADDRESS.get_bound(py)?)? {
         let value = subject
             .getattr(pyo3::intern!(py, "_packed"))?
             .call0()?
-            .downcast::<pyo3::types::PyBytes>()?;
-
-        Ok(SubjectOwner::IPAddress(value.into()))
+            .downcast::<pyo3::types::PyBytes>()?
+            .clone();
+        Ok(SubjectOwner::IPAddress(value.clone().unbind()))
```
alex, line 382: I don't think this clone is required.

src/rust/src/x509/verify.rs
```diff
@@ -366,21 +366,20 @@ fn build_subject_owner(
...
-    if subject.is_instance(types::DNS_NAME.get(py)?)? {
+    if subject.is_instance(&types::DNS_NAME.get_bound(py)?)? {
         let value = subject
             .getattr(pyo3::intern!(py, "value"))?
-            .downcast::<pyo3::types::PyString>()?;
-
-        Ok(SubjectOwner::DNSName(value.to_str()?.to_owned()))
-    } else if subject.is_instance(types::IP_ADDRESS.get(py)?)? {
+            .extract::<String>()?;
+        Ok(SubjectOwner::DNSName(value))
```
alex, line 375: Can you add a comment reminding us that when our minimum python is 3.10, we can switch to borrowing the str.

src/rust/src/x509/verify.rs
facutuesca: good catch, fixed!

src/rust/src/x509/verify.rs
facutuesca: done

alex (approved):

facutuesca:
Looks like a GHA error on one of the runners:
```
INTERNALERROR>   File "/home/runner/work/cryptography/cryptography/.nox/tests-nocoverage/lib/pypy3.10/site-packages/xdist/dsession.py", line 137, in loop_once
INTERNALERROR>     raise RuntimeError("Unexpectedly no active workers available")
INTERNALERROR> RuntimeError: Unexpectedly no active workers available
```

PR #10748, opened 2024-04-05
Misc oscp pyo3 migrations

facutuesca:

This contains misc migrations for the `ocsp` files and `pkcs7`. One of the changes triggers a pretty big `cargo fmt` change in `pkcs7.rs` making the diff very difficult to read, so I split that change into a second commit.
The only thing that actually changes in the second commit is:

```rust
// pkcs7.rs:165
// before
let digest = asn1::write_single(&x509::ocsp::hash_data(py, py_hash_alg, &data_with_header)?)?;

// after
let digest = asn1::write_single(&x509::ocsp::hash_data(py, &py_hash_alg.as_borrowed(), &data_with_header)?)?;
```
the rest is all formatting.

I suggest reviewing the first commit normally, and then the second to check the above change.

cc @alex @reaperhulk

### DIFF
--- a/src/rust/src/pkcs7.rs
+++ b/src/rust/src/pkcs7.rs
@@ -134,71 +134,73 @@ fn sign_and_serialize<'p>(
         .map(|p| p.raw.borrow_dependent())
         .collect::<Vec<_>>();
     for (cert, py_private_key, py_hash_alg, rsa_padding) in &py_signers {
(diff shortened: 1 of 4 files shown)


alex (approved):

facutuesca:
Ah this also removes the last of the `ocsp_resp.rs` warnings

PR #10749, opened 2024-04-05
Migrate more `x509/extensions.rs` APIs to new pyo3 APIs (and other migrations)

facutuesca:

This one does not fix any warnings, but it does remove multiple `into_gil_ref()` calls and migrates several functions to use `Bound` parameters.

cc @alex @reaperhulk

### DIFF
--- a/src/rust/src/x509/certificate.rs
+++ b/src/rust/src/x509/certificate.rs
@@ -960,10 +960,7 @@ fn create_x509_certificate(
         subject_unique_id: None,
         raw_extensions: x509::common::encode_extensions(
             py,
-            builder
-                .getattr(pyo3::intern!(py, "_extensions"))?
-                .clone()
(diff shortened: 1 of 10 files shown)


src/rust/src/x509/certificate.rs
```diff
@@ -103,14 +103,11 @@ impl Certificate {
...
-            result,
-            encoding.clone().into_gil_ref(),
-        )?
-        .as_borrowed()
-        .to_owned())
+        Ok(
+            encode_der_data(py, "CERTIFICATE".to_string(), result, encoding)?
+                .as_borrowed()
+                .to_owned(),
+        )
```
alex, line 110: Pretty sure you can drop these.

facutuesca: yup, fixed!

alex (approved):

PR #10763, opened 2024-04-06
Simplify OCSP response generating code

alex:
Wrapping nearly the entire function body with an `if` made it less readable


reaperhulk (approved):
An unpleasant diff to read 😬

PR #10764, opened 2024-04-07
Introduce a keepalive abstraction

alex:
This effectively emulates what pyo3's old GIL pool was doing, but we'll use it in a far more targetted manner.


reaperhulk (approved):
🤞

PR #10778, opened 2024-04-09
Fix lifetime errors in `asn1.rs` with `gil-refs` disabled

facutuesca:

This should be merged **after** the [keepalive PR](https://github.com/pyca/cryptography/pull/10764) is merged

Fixes all of the lifetime errors in `asn1.rs`, changing a return type of `py_uint_to_big_endian_bytes` from `&'p [u8]` to `PyBackedBytes`.
It uses the new `keepalive` in the places where the `PyBackedBytes` object is borrowed for longer than its lifetime.

cc @alex @reaperhulk

### DIFF
--- a/src/rust/cryptography-keepalive/src/lib.rs
+++ b/src/rust/cryptography-keepalive/src/lib.rs
@@ -20,8 +20,8 @@ pub unsafe trait StableDeref: Deref {}
 // slice returned by `deref` remains valid.
 unsafe impl<T> StableDeref for Vec<T> {}
 
-// SAFETY: `PyBackedBytes`'s data is on the heap, so as long as it's not mutated, the
-// slice returned by `deref` remains valid.
+// SAFETY: `PyBackedBytes`'s data is on the heap and `bytes` objects in
+// Python are immutable.
 unsafe impl StableDeref for PyBackedBytes {}
 
 #[allow(clippy::new_without_default)]
--- a/src/rust/cryptography-keepalive/src/lib.rs
+++ b/src/rust/cryptography-keepalive/src/lib.rs
@@ -4,6 +4,7 @@
 
 #![deny(rust_2018_idioms, clippy::undocumented_unsafe_blocks)]
 
+use pyo3::pybacked::PyBackedBytes;
 use std::cell::UnsafeCell;
 use std::ops::Deref;
 
@@ -18,6 +19,10 @@ pub unsafe trait StableDeref: Deref {}
 // SAFETY: `Vec`'s data is on the heap, so as long as it's not mutated, the
 // slice returned by `deref` remains valid.
 unsafe impl<T> StableDeref for Vec<T> {}
+
+// SAFETY: `PyBackedBytes`'s data is on the heap, so as long as it's not mutated, the
+// slice returned by `deref` remains valid.
+unsafe impl StableDeref for PyBackedBytes {}
 
 #[allow(clippy::new_without_default)]
 impl<T: StableDeref> KeepAlive<T> {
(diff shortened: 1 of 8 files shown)


src/rust/cryptography-keepalive/src/lib.rs
```diff
@@ -19,6 +20,10 @@ pub unsafe trait StableDeref: Deref {}
 // slice returned by `deref` remains valid.
 unsafe impl<T> StableDeref for Vec<T> {}
 
+// SAFETY: `PyBackedBytes`'s data is on the heap, so as long as it's not mutated, the
+// slice returned by `deref` remains valid.
```
alex, line 24: `bytes` are immutable in python, so the comment here should say that, not that it's fine as long as it's not mutated.

src/rust/src/x509/extensions.rs
```diff
@@ -52,9 +52,13 @@ pub(crate) fn encode_authority_key_identifier<'a>(
     } else {
         None
     };
+    let ka = cryptography_keepalive::KeepAlive::new();
```
alex, line 55: I don't think this is needed, you just need to change the scope of the `serial_bytes` local to be function, rather than block.

src/rust/cryptography-keepalive/src/lib.rs
facutuesca: fixed

src/rust/src/x509/extensions.rs
facutuesca: yes, fixed!

alex:
Should be good to rebase

src/rust/src/x509/extensions.rs
```diff
@@ -52,10 +53,11 @@ pub(crate) fn encode_authority_key_identifier<'a>(
     } else {
         None
     };
+    let serial_bytes: PyBackedBytes;
```
alex, line 56:
```suggestion
    let serial_bytes;
```

src/rust/src/x509/ocsp_req.rs
```diff
@@ -188,7 +189,8 @@ fn create_ocsp_request(
         (issuer_name_hash, issuer_key_hash, py_serial, py_hash) = builder
             .getattr(pyo3::intern!(py, "_request_hash"))?
             .extract()?;
-        let serial_number = asn1::BigInt::new(py_uint_to_big_endian_bytes(py, py_serial)?).unwrap();
+        let serial_number_bytes = ka.add(py_uint_to_big_endian_bytes(py, py_serial)?);
```
alex, line 192: I don't think a `ka` is needed here. Can just do the scoping trick.

src/rust/src/x509/extensions.rs
facutuesca: fixed

src/rust/src/x509/ocsp_req.rs
facutuesca: good catch, fixed

src/rust/src/x509/ocsp_req.rs
```diff
@@ -171,6 +171,8 @@ fn create_ocsp_request(
     builder: &pyo3::Bound<'_, pyo3::PyAny>,
 ) -> CryptographyResult<OCSPRequest> {
     let builder_request = builder.getattr(pyo3::intern!(py, "_request"))?;
+    let serial_number_bytes;
+    //let ka = cryptography_keepalive::KeepAlive::new();
```
alex, line 175:
```suggestion
```

facutuesca: 🤦 thanks!

alex (approved):

PR #10812, opened 2024-04-14
Finish removal of `gil-refs` feature from pyo3

alex:


alex:
@davidhewitt We're seeing something interesting here where, with the removal of the gil-refs feature, about half of CI jobs are failing, and half are fine.

Looking at the code in question, I'd kind of expect it to _never_ compile, do you have any idea what sort of non-determinism maybe living here? And which direction it's going... is it a bug that it's an error, or a bug that it ever compiles!

alex:
I thought maybe this would be connected to CI caching, but it doesn't appear to be. I'm so confused.

alex:
Ok, well, I just fixed those compilation issues, so I guess that's a fine solution.

I still don't understand why it only showed up on some CI runs.

davidhewitt:
The CI runs which failed are running < 3.10 so the `FromPyObjectBound` implementation doesn't exist (we can only implement it on `abi3` for 3.10+, as we need `PyUnicode_AsUtf8AndSize`.

alex:
AHHHHH. Of course. Thanks for reminding me, and sorry for the noise!

reaperhulk (approved):

Issue #10818, opened 2024-04-15
No `_utc` variant of `InvalidityDate.invalidity_date`


felixfontein:
While most timestamps now have a `_utc` variant, `InvalidityDate.invalidity_date` does not.

PR #10848, opened 2024-04-18
Add timezone-aware API variant for `x509.InvalidityDate.invalidity_date`

facutuesca:
Fixes https://github.com/pyca/cryptography/issues/10818.

This PR adds `x509.InvalidityDate.invalidity_date_utc`, a timezone-aware variant of `x509.InvalidityDate.invalidity_date`.

Currently, `InvalidityDate(invalidity_date: datetime.datetime)` is constructed with an arbitrary `datetime` object (which can be naive or time-aware), and then when the extension is encoded it gets normalized to UTC ([src1](https://github.com/pyca/cryptography/blob/194570150d1d83c8b3e30dff4f2bf38c7fbecff8/src/rust/src/x509/extensions.rs#L534), [src2](https://github.com/pyca/cryptography/blob/194570150d1d83c8b3e30dff4f2bf38c7fbecff8/src/rust/src/x509/common.rs#L513-L536)). If the `datetime` object was naive, it's treated as if it was UTC. Otherwise it's converted to UTC using `datetime.astimezone()`.

The new `invalidity_date_utc` attribute follows the same behavior, assuming naive `datetime` objects are UTC and adding the timezone, or converting aware `datetime` objects to UTC.

I haven't changed most of the current usages of `invalidity_date` in the codebase to `invalidity_date_utc`, like in the `repr`, `eq` and `hash` methods of `InvalidityDate`, since this is a breaking change we might want to make only once we deprecate `invalidity_date`. I did change it in the `encode_extension` Rust function, since it should behave exactly the same as before.

cc @alex @reaperhulk @woodruffw

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -41,6 +41,9 @@ Changelog
   :attr:`~cryptography.x509.CertificateSigningRequest.public_key_algorithm_oid`
   to determine the :class:`~cryptography.hazmat._oid.PublicKeyAlgorithmOID`
   Object Identifier of the public key found inside the certificate.
+* Added :attr:`~cryptography.x509.InvalidityDate.invalidity_date_utc`, a
+  timezone-aware alternative to the naïve ``datetime`` attribute
+  :attr:`~cryptography.x509.InvalidityDate.invalidity_date`.
 
 .. _v42-0-5:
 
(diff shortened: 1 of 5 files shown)


alex (approved):

felixfontein:
@facutuesca thanks a lot for implementing this!

Issue #10790, opened 2024-04-11
DH exchange fails in 42.0.0


daveboutcher:
The following code works with cryptography 41.0.7 and earlier, and fails with 42.0.0 and later:
```
from cryptography.hazmat.primitives.serialization import load_pem_public_key

key = b"""-----BEGIN PUBLIC KEY-----
MIICJTCCARcGCSqGSIb3DQEDATCCAQgCggEBAP//////////yQ/aoiFowjTExmKL
gNwc0SkCTgiKZ8x0Agu+pjsTmyJRSgh5jjQE3e+VGbPNOkMbMCsKbfJfFDdP4TVt
bVHCReSFtXZiXn7G9ExC6aY37WsL/1y29Aa37e44a/taiZ+lrp8kEXxLH+ZJKGZR
7ORbPcIAfLihY78FmNpINhxV05ppFj+o/STPX4NlXSPco62WHGLzViCFUrue1SkH
cJaWbWcMNU5KvJgE8XRsCMoYIXwykF5GLjbOO+OedywYDoYDmyeDouwHoo+1xV3w
b0xSyd4ry/aVWBcYOZVJfOqVauUV0iYYmPoFEBVyjlqKrKpo//////////8CAQID
ggEGAAKCAQEAoely6vSHw+/Q3zGYLaJj7eeQkfd25K8SvtC+FMY9D7jwS4g71pyr
U3FJ98Fi45Wdksh+d4u7U089trF5Xbgui29bZ0HcQZtfHEEz0Mh69tkipCm2/QIj
6eDlo6sPk9hhhvgg4MMGiWKhCtHrub3x1FHdmf7KjOhrGeb5apiudo7blGFzGhZ3
NFnbff+ArVNd+rdVmSoZn0aMhXRConlDu/44IYe5/24VLl7G+BzZlIZO4P2M83fd
mBOvR13cmYssQjEFTbaZVQvQHa3t0+aywfdCgsXGmTTK6QDCBP8D+vf1bmhEswzs
oYn1GLtJ3VyYyMBPDBomd2ctchZgTzsX1w==
-----END PUBLIC KEY-----"""

peer_key = load_pem_public_key(key)

params = peer_key.parameters()
private_key = params.generate_private_key()

# Create a shared secret
shared_secret = private_key.exchange(peer_key)
```

The exchange call fails with a cryptic in 42.0.0 and later
```
    shared_secret = private_key.exchange(peer_key)
[...]

PR #10864, opened 2024-04-21
Fix DH Key exchange incompatibility

daveboutcher:
A change in key types for DH keys was introduced in 42.0.0 such that DH key exchange with older versions (e.g. 41.0.7) no longer works.  This affects key exchange with other OpenSSL servers as well.

This commit reverts the code that attempts to differentiate between DH and DHX key exchanges.

Fixes issue #10790

### DIFF
--- a/src/rust/src/backend/dh.rs
+++ b/src/rust/src/backend/dh.rs
@@ -77,11 +77,7 @@ fn pkey_from_dh<T: openssl::pkey::HasParams>(
         if #[cfg(CRYPTOGRAPHY_IS_LIBRESSL)] {
             Ok(openssl::pkey::PKey::from_dh(dh)?)
         } else {
-            if dh.prime_q().is_some() {
-                Ok(openssl::pkey::PKey::from_dhx(dh)?)
-            } else {
-                Ok(openssl::pkey::PKey::from_dh(dh)?)
-            }
+            Ok(openssl::pkey::PKey::from_dh(dh)?)
         }
     }
 }
--- a/src/rust/src/backend/dh.rs
+++ b/src/rust/src/backend/dh.rs
@@ -12,6 +12,7 @@ use pyo3::prelude::{PyAnyMethods, PyModuleMethods};
 
 const MIN_MODULUS_SIZE: u32 = 512;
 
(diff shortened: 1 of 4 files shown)


src/rust/src/backend/dh.rs
```diff
@@ -77,11 +77,7 @@ fn pkey_from_dh<T: openssl::pkey::HasParams>(
         if #[cfg(CRYPTOGRAPHY_IS_LIBRESSL)] {
             Ok(openssl::pkey::PKey::from_dh(dh)?)
         } else {
-            if dh.prime_q().is_some() {
-                Ok(openssl::pkey::PKey::from_dhx(dh)?)
-            } else {
-                Ok(openssl::pkey::PKey::from_dh(dh)?)
-            }
+            Ok(openssl::pkey::PKey::from_dh(dh)?)
```
alex, line 80: So with this change it will never create a `dhx` variant, which I think is more aggressive than we want.

I'm honestly unsure how it is that no tests fail because of that.

I think you only want to apply this change to the `generate` codepaths. My suggest would be to add an `allow_dhx` param (maybe a better name), pass false in generate, but true everywhere else.

tests/hazmat/primitives/test_dh.py
```diff
@@ -473,6 +474,39 @@ def test_public_key_copy(self):
 
         assert key1 == key2
 
+    def test_old_public_key_exchange(self, backend):
+        # This is a public key generated on cryptograpy version 41.0.7
+        key_data = textwrap.dedent(
```
alex, line 479: Generally we use keys from the `vectors/` directory in our tests -- there should be an existing one with a matching format.

daveboutcher:
@alex sorry for the extreme lag, its taken me a while to get back to this.

I don't love what I did here, but see if its along the lines of what you were looking for, both in the testing side and keygen side.  None of the existing test keys in `vectors/` exhibited the issue I am fighting with.

All comments and suggestions welcome...

alex:
The twisted test failures here appear to be genuine DH failures, suggesting that there's an issue in this PR. It also likely indicates missing test cases in our own test suite.

PR #10889, opened 2024-04-25
Add support for encrypting S/MIME messages

facutuesca:
I'm opening this PR with an initial implementation of S/MIME encryption, in order to better discuss the API design, the algorithms we want to support, and how we want to approach testing.

The target is a subset of S/MIME v3.2 ([RFC5751](https://datatracker.ietf.org/doc/html/rfc5751)):
- Content encryption is done using AES-128-CBC
- Key management is done with key transport: the symmetric encryption key used for the message is included and encrypted using the recipients' public keys.
- The other two key management methods [(previously-distributed keys and key agreement)](https://datatracker.ietf.org/doc/html/rfc2630#section-12.3) are not supported.
- The symmetric encryption key is encrypted using RSA (PKCS1 v1.5). That is, we only support recipients with RSA public keys, and we use `PKCS1v15` padding.

Here are the openssl commands that can be used for testing, to compare our encrypted output against, or to decrypt our encrypted output.
```sh
# encrypt
openssl smime -encrypt  -aes-128-cbc -in msg.txt -out out.txt -outform PEM vectors/cryptography_vectors/x509/custom/ca/rsa_ca.pem
# decrypt
openssl smime -decrypt -recip vectors/cryptography_vectors/x509/custom/ca/rsa_ca.pem -inkey vectors/cryptography_vectors/x509/custom/ca/rsa_key.pem -in out.txt -inform pem
```

~I added some tests for the unencrypted parts of the message, but complete testing would require that we parse and decrypt the messages. We could follow a similar approach as with testing S/MIME signing, where we call OpenSSL directly to parse and check our output during the tests~

The tests now use OpenSSL's [PKCS7_decrypt](https://www.openssl.org/docs/man3.3/man3/PKCS7_decrypt.html) in order to see if our encrypted output can be correctly decrypted by OpenSSL.

cc @alex @reaperhulk @woodruffw 

(the issue tracking this feature is https://github.com/pyca/cryptography/issues/5488)

### TODO:
- [x] Write the docs
- [x] Add to CHANGELOG

### DIFF
--- a/src/cryptography/hazmat/primitives/serialization/pkcs7.py
+++ b/src/cryptography/hazmat/primitives/serialization/pkcs7.py
@@ -12,6 +12,7 @@ import io
 import typing
 
 from cryptography import utils, x509
+from cryptography.exceptions import UnsupportedAlgorithm, _Reasons
 from cryptography.hazmat.bindings._rust import pkcs7 as rust_pkcs7
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import ec, padding, rsa
@@ -177,7 +178,93 @@ class PKCS7SignatureBuilder:
         return rust_pkcs7.sign_and_serialize(self, encoding, options)
 
 
-def _smime_encode(
+class PKCS7EnvelopeBuilder:
+    def __init__(
+        self,
+        *,
+        _data: bytes | None = None,
+        _recipients: list[x509.Certificate] | None = None,
+    ):
+        from cryptography.hazmat.backends.openssl.backend import (
+            backend as ossl,
+        )
+
+        if not ossl.rsa_encryption_supported(padding=padding.PKCS1v15()):
+            raise UnsupportedAlgorithm(
+                "RSA with PKCS1 v1.5 padding is not supported by this version"
+                " of OpenSSL.",
+                _Reasons.UNSUPPORTED_PADDING,
+            )
+        self._data = _data
+        self._recipients = _recipients if _recipients is not None else []
+
+    def set_data(self, data: bytes) -> PKCS7EnvelopeBuilder:
+        _check_byteslike("data", data)
+        if self._data is not None:
+            raise ValueError("data may only be set once")
+
+        return PKCS7EnvelopeBuilder(_data=data, _recipients=self._recipients)
+
+    def add_recipient(
+        self,
+        certificate: x509.Certificate,
+    ) -> PKCS7EnvelopeBuilder:
+        if not isinstance(certificate, x509.Certificate):
+            raise TypeError("certificate must be a x509.Certificate")
+
+        if not isinstance(certificate.public_key(), rsa.RSAPublicKey):
+            raise TypeError("Only RSA keys are supported at this time.")
+
+        return PKCS7EnvelopeBuilder(
+            _data=self._data,
+            _recipients=[
+                *self._recipients,
+                certificate,
+            ],
+        )
+
+    def encrypt(
+        self,
+        encoding: serialization.Encoding,
+        options: typing.Iterable[PKCS7Options],
+        backend: typing.Any = None,
+    ) -> bytes:
+        if len(self._recipients) == 0:
+            raise ValueError("Must have at least one recipient")
+        if self._data is None:
+            raise ValueError("You must add data to encrypt")
+        options = list(options)
+        if not all(isinstance(x, PKCS7Options) for x in options):
+            raise ValueError("options must be from the PKCS7Options enum")
+        if encoding not in (
+            serialization.Encoding.PEM,
+            serialization.Encoding.DER,
+            serialization.Encoding.SMIME,
+        ):
+            raise ValueError(
+                "Must be PEM, DER, or SMIME from the Encoding enum"
+            )
+
+        # Only allow options that make sense for encryption
+        if any(
+            opt not in [PKCS7Options.Text, PKCS7Options.Binary]
+            for opt in options
+        ):
+            raise ValueError(
+                "Only the following options are supported for encryption: "
+                "Text, Binary"
+            )
+        elif PKCS7Options.Text in options and PKCS7Options.Binary in options:
+            # OpenSSL accepts both options at the same time, but ignores Text.
+            # We fail defensively to avoid unexpected outputs.
+            raise ValueError(
+                "Cannot use Binary and Text options at the same time"
+            )
+
+        return rust_pkcs7.encrypt_and_serialize(self, encoding, options)
+
+
+def _smime_signed_encode(
     data: bytes, signature: bytes, micalg: str, text_mode: bool
 ) -> bytes:
     # This function works pretty hard to replicate what OpenSSL does
@@ -223,6 +310,23 @@ def _smime_encode(
     )
     g.flatten(m)
     return fp.getvalue()
+
+
+def _smime_enveloped_encode(data: bytes) -> bytes:
+    m = email.message.Message()
+    m.add_header("MIME-Version", "1.0")
+    m.add_header("Content-Disposition", "attachment", filename="smime.p7m")
+    m.add_header(
+        "Content-Type",
+        "application/pkcs7-mime",
+        smime_type="enveloped-data",
+        name="smime.p7m",
+    )
+    m.add_header("Content-Transfer-Encoding", "base64")
+
+    m.set_payload(email.base64mime.body_encode(data, maxlinelen=65))
+
+    return m.as_bytes(policy=m.policy.clone(linesep="\n", max_line_length=0))
 
 
 class OpenSSLMimePart(email.message.MIMEPart):
--- a/src/cryptography/hazmat/primitives/serialization/pkcs7.py
+++ b/src/cryptography/hazmat/primitives/serialization/pkcs7.py
@@ -227,7 +227,6 @@ class PKCS7EnvelopeBuilder:
         self,
         encoding: serialization.Encoding,
         options: typing.Iterable[PKCS7Options],
-        backend: typing.Any = None,
     ) -> bytes:
         if len(self._recipients) == 0:
             raise ValueError("Must have at least one recipient")
--- a/src/rust/cryptography-x509/src/common.rs
+++ b/src/rust/cryptography-x509/src/common.rs
@@ -136,6 +136,16 @@ pub enum AlgorithmParameters<'a> {
     #[defined_by(oid::HMAC_WITH_SHA256_OID)]
     HmacWithSha256(asn1::Null),
 
+    // Used only in PKCS#7 AlgorithmIdentifiers
+    // https://datatracker.ietf.org/doc/html/rfc3565#section-4.1
+    //
+    // From RFC 3565 section 4.1:
+    // The AlgorithmIdentifier parameters field MUST be present, and the
+    // parameters field MUST contain a AES-IV:
+    //
+    // AES-IV ::= OCTET STRING (SIZE(16))
+    #[defined_by(oid::AES_128_CBC_OID)]
+    Aes128Cbc([u8; 16]),
     #[defined_by(oid::AES_256_CBC_OID)]
     Aes256Cbc([u8; 16]),
 
--- a/src/rust/cryptography-x509/src/pkcs7.rs
+++ b/src/rust/cryptography-x509/src/pkcs7.rs
@@ -6,6 +6,7 @@ use crate::{certificate, common, csr, name};
 
 pub const PKCS7_DATA_OID: asn1::ObjectIdentifier = asn1::oid!(1, 2, 840, 113549, 1, 7, 1);
 pub const PKCS7_SIGNED_DATA_OID: asn1::ObjectIdentifier = asn1::oid!(1, 2, 840, 113549, 1, 7, 2);
+pub const PKCS7_ENVELOPED_DATA_OID: asn1::ObjectIdentifier = asn1::oid!(1, 2, 840, 113549, 1, 7, 3);
 pub const PKCS7_ENCRYPTED_DATA_OID: asn1::ObjectIdentifier = asn1::oid!(1, 2, 840, 113549, 1, 7, 6);
 
 #[derive(asn1::Asn1Write)]
@@ -18,6 +19,8 @@ pub struct ContentInfo<'a> {
 
 #[derive(asn1::Asn1DefinedByWrite)]
 pub enum Content<'a> {
+    #[defined_by(PKCS7_ENVELOPED_DATA_OID)]
+    EnvelopedData(asn1::Explicit<Box<EnvelopedData<'a>>, 0>),
     #[defined_by(PKCS7_SIGNED_DATA_OID)]
     SignedData(asn1::Explicit<Box<SignedData<'a>>, 0>),
     #[defined_by(PKCS7_DATA_OID)]
@@ -54,6 +57,21 @@ pub struct SignerInfo<'a> {
 
     #[implicit(1)]
     pub unauthenticated_attributes: Option<csr::Attributes<'a>>,
+}
+
+#[derive(asn1::Asn1Write)]
+pub struct EnvelopedData<'a> {
+    pub version: u8,
+    pub recipient_infos: asn1::SetOfWriter<'a, RecipientInfo<'a>>,
+    pub encrypted_content_info: EncryptedContentInfo<'a>,
+}
+
+#[derive(asn1::Asn1Write)]
+pub struct RecipientInfo<'a> {
+    pub version: u8,
+    pub issuer_and_serial_number: IssuerAndSerialNumber<'a>,
+    pub key_encryption_algorithm: common::AlgorithmIdentifier<'a>,
+    pub encrypted_key: &'a [u8],
 }
 
 #[derive(asn1::Asn1Write)]
--- a/src/rust/src/pkcs7.rs
+++ b/src/rust/src/pkcs7.rs
@@ -13,14 +13,14 @@ use cryptography_x509::{common, oid, pkcs7};
 use once_cell::sync::Lazy;
 #[cfg(not(CRYPTOGRAPHY_IS_BORINGSSL))]
 use openssl::pkcs7::Pkcs7;
-use pyo3::types::{PyAnyMethods, PyBytes, PyBytesMethods, PyListMethods};
+use pyo3::types::{PyAnyMethods, PyBytesMethods, PyListMethods};
 #[cfg(not(CRYPTOGRAPHY_IS_BORINGSSL))]
 use pyo3::IntoPy;
 
 use crate::asn1::encode_der_data;
 use crate::buf::CffiBuf;
 use crate::error::{CryptographyError, CryptographyResult};
-use crate::padding::PKCS7PaddingContext;
+use crate::pkcs12::symmetric_encrypt;
 #[cfg(not(CRYPTOGRAPHY_IS_BORINGSSL))]
 use crate::x509::certificate::load_der_x509_certificate;
 use crate::{exceptions, types, x509};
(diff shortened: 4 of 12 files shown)


src/rust/cryptography-x509/src/common.rs
```diff
@@ -57,6 +57,17 @@ pub enum AlgorithmParameters<'a> {
...
+    // Used only in PKCS#7 AlgorithmIdentifiers
+    // https://datatracker.ietf.org/doc/html/rfc3565#section-4.1
+    //
+    // From RFC 3565 section 4.1:
+    // The AlgorithmIdentifier parameters field MUST be present, and the
+    // parameters field MUST contain a AES-IV:
+    //
+    // AES-IV ::= OCTET STRING (SIZE(16))
+    #[defined_by(oid::AES_128_CBC_OID)]
+    AesCbc(&'a [u8]),
```
woodruffw, line 69: Out of curiosity: does `rust-asn1` know to turn this `&[u8]` into the inner value of the `OCTET STRING`, or does this end up containing the raw TLV for the `OCTET STRING`? I suspect it's the former, but we should confirm.

alex: `&[u8]` in rust-asn1 are OCTET STRINGs, not raw TLVs.

src/rust/cryptography-x509/src/oid.rs
```diff
@@ -74,6 +74,9 @@ pub const EC_BRAINPOOLP384R1: asn1::ObjectIdentifier = asn1::oid!(1, 3, 36, 3, 3
 pub const EC_BRAINPOOLP512R1: asn1::ObjectIdentifier = asn1::oid!(1, 3, 36, 3, 3, 2, 8, 1, 1, 13);
 
 pub const RSA_OID: asn1::ObjectIdentifier = asn1::oid!(1, 2, 840, 113549, 1, 1, 1);
+pub const AES_256_CBC_OID: asn1::ObjectIdentifier = asn1::oid!(2, 16, 840, 1, 101, 3, 4, 1, 42);
+pub const AES_192_CBC_OID: asn1::ObjectIdentifier = asn1::oid!(2, 16, 840, 1, 101, 3, 4, 1, 22);
+pub const AES_128_CBC_OID: asn1::ObjectIdentifier = asn1::oid!(2, 16, 840, 1, 101, 3, 4, 1, 2);
```
woodruffw, line 79: Just for cross-checking: confirmed these against their parent arc: https://oidref.com/2.16.840.1.101.3.4.1

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -225,6 +296,34 @@ def _smime_encode(
...
+
+    fp = io.BytesIO()
+    g = email.generator.BytesGenerator(
+        fp,
+        maxheaderlen=0,
+        mangle_from_=False,
+        policy=m.policy.clone(linesep="\n"),
+    )
+    g.flatten(m)
+    return fp.getvalue()
```
woodruffw, line 324: Nitpick: you might be able to get away with just [`m.as_bytes(...)](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage.as_bytes)` rather than jumping through a `BytesIO` + `BytesGenerator` here.

facutuesca: fixed!

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -225,6 +307,26 @@ def _smime_encode(
     return fp.getvalue()
 
 
+def _smime_enveloped_encode(data: bytes) -> bytes:
+    # This function works pretty hard to replicate what OpenSSL does
+    # precisely. For good and for ill.
```
woodruffw, line 312: Nitpick: might be good to include a URL or code reference for OpenSSL's construction of the encoding here, just in case this ever needs to be re-evaluated 🙂

facutuesca: This is actually a comment from the other encoder function that I used as a base for this one (`_smime_signed_encode`). 

 I think it doesn't make much sense here since we just add the headers, so I removed it

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -177,7 +178,88 @@ def sign(
         return rust_pkcs7.sign_and_serialize(self, encoding, options)
 
 
-def _smime_encode(
+class PKCS7EnvelopeBuilder:
+    def __init__(
+        self,
+        data: bytes | None = None,
+        recipients: list[x509.Certificate] | None = None,
```
alex, line 185:
```suggestion
        *,
        _data: bytes | None = None,
        _recipients: list[x509.Certificate] | None = None,
```

People keep assuming these are public, let's make that harder for them.

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -177,7 +178,88 @@ def sign(
...
+                "Must be PEM, DER, or SMIME from the Encoding enum"
+            )
+
+        # These options only make sense when signing, not encrypting
+        if (
+            PKCS7Options.NoAttributes in options
+            or PKCS7Options.NoCapabilities in options
+            or PKCS7Options.NoCerts in options
+            or PKCS7Options.DetachedSignature in options
+        ):
```
alex, line 253: I think we probably want to do an allow-list of sensible options, not the inverse.

src/rust/cryptography-x509/src/common.rs
alex, line 69: Pretty sure if you rebase you'll find this was added in the meantime :-)

src/rust/cryptography-x509/src/pkcs7.rs
```diff
@@ -54,6 +56,29 @@ pub struct SignerInfo<'a> {
     pub unauthenticated_attributes: Option<csr::Attributes<'a>>,
 }
 
+#[derive(asn1::Asn1Write)]
+pub struct EncryptedContentInfo<'a> {
+    pub _content_type: asn1::ObjectIdentifier,
+    pub content_encryption_algorithm: common::AlgorithmIdentifier<'a>,
+    #[implicit(0)]
+    pub content: Option<&'a [u8]>,
+}
```
alex, line 65: Note this is also defined in #11059

src/rust/src/pkcs7.rs
```diff
@@ -75,6 +77,105 @@ fn serialize_certificates<'p>(
...
+    };
+
+    let padder = types::SYMMETRIC_PADDING_PKCS7
+        .get(py)?
+        .call1((128,))?
+        .call_method0(pyo3::intern!(py, "padder"))?;
+    let padded_content_start =
+        padder.call_method1(pyo3::intern!(py, "update"), (data_with_header,))?;
+    let padded_content_end = padder.call_method0(pyo3::intern!(py, "finalize"))?;
+    let padded_content = padded_content_start.add(padded_content_end)?;
```
alex, line 102: There's now a rust pkcs7 padding API

src/rust/src/pkcs7.rs
```diff
@@ -75,6 +77,105 @@ fn serialize_certificates<'p>(
...
+
+    // The message is encrypted with AES-128-CBC, which the S/MIME v3.2 RFC
+    // specifies as MUST support (https://datatracker.ietf.org/doc/html/rfc5751#section-2.7)
```
alex, line 105: Someone noted that there's an S/MIME 4.0 which supports AEADs. Is there a reason not to use them?

alex (commented):
Left a few comments. Deciding on a testing strategy for the encrypted components is probably the most important next step.

src/cryptography/hazmat/primitives/serialization/pkcs7.py
facutuesca: Fixed!

src/cryptography/hazmat/primitives/serialization/pkcs7.py
facutuesca: Fixed

src/rust/cryptography-x509/src/pkcs7.rs
facutuesca: Fixed so that the field names match

src/rust/src/pkcs7.rs
facutuesca: Fixed. Since the new `PKCS7PaddingContext::update` method takes a `CffiBuf`, I had to use `PyBytes::new_bound` to go from `Cow[u8]` to `PyBound<PyBytes>` to `CffiBuf`, which means there is now an extra copy operation. Is there a better way of doing that?

src/rust/src/pkcs7.rs
facutuesca: Not really, mostly scope (I limited this PR to a subset of S/MIME 3.2)

facutuesca:
@alex @woodruffw 

> Deciding on a testing strategy for the encrypted components is probably the most important next step

I added a binding for [PKCS7_decrypt](https://www.openssl.org/docs/man1.0.2/man3/PKCS7_decrypt.html) and changed the tests to check that the encrypted output can be correctly decrypted by OpenSSL.

There are 4 new commits since the last review round:
- `Use correct PKCS7/SMIME OID for enveloped data`
  - This was a mistake I did while rebasing. We use `envelopedData` rather than `encryptedData` for S/MIME encryption.
- `Fail when passing Text and Binary options at the same time`
  - OpenSSL allows passing both options at the same time, but ignores `Text`. We defensively fail here so that users know they're doing something weird/wrong.
- `Use PKCS7 tag for PEM-encoding encrypted messages`
  - Originally I used the `CMS` tag for PEM encoded messages (`-----BEGIN CMS-----`), matching the output of `openssl cms`. However, this doesn't match our other APIs which use `-----BEGIN PKCS7-----` (which corresponds to `openssl smime`). Furthermore, if we wanted to test decryption of the former, we would need to add new bindings for `cms.h` functions and structures, whereas with `PKCS7` we can reuse the existing bindings and be consistent with our other PCKS7 APIs.
- `Test S/MIME encryption by decrypting and comparing with plaintext`
  -  As mentioned before, it adds a binding for `PKCS7_decrypt` and uses it to test our encrypted output.

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -177,7 +178,95 @@ def sign(
...
+            raise ValueError(
+                "Must be PEM, DER, or SMIME from the Encoding enum"
+            )
+
+        # Only allow options that make sense for encryption
+        if any(
+            [
+                opt not in [PKCS7Options.Text, PKCS7Options.Binary]
+                for opt in options
+            ]
```
alex, line 253: No need for list comprehension, generator is fine

src/rust/cryptography-x509/src/common.rs
alex: Oops, missed that. I think you'll want to follow the style on that one -- using fixed size array and name including 128

src/rust/cryptography-x509/src/pkcs7.rs
```diff
@@ -54,6 +57,29 @@ pub struct SignerInfo<'a> {
     pub unauthenticated_attributes: Option<csr::Attributes<'a>>,
 }
 
+#[derive(asn1::Asn1Write)]
```
alex, line 60: In main now

alex (commented):
Started to review but realized there's going to be a few things that are no longer needed now that pkcs12 is merged.

src/cryptography/hazmat/primitives/serialization/pkcs7.py
facutuesca: Fixed

src/rust/cryptography-x509/src/pkcs7.rs
facutuesca: Removed!

src/rust/cryptography-x509/src/common.rs
facutuesca: Fixed!

src/rust/src/pkcs7.rs
```diff
@@ -75,6 +78,100 @@ fn serialize_certificates<'p>(
...
+    let key = types::OS_URANDOM.get(py)?.call1((16,))?;
+    let aes128_algorithm = types::AES128.get(py)?.call1((&key,))?;
+    let iv = types::OS_URANDOM.get(py)?.call1((16,))?;
+    let cbc_mode = types::CBC.get(py)?.call1((&iv,))?;
+    let cipher = types::CIPHER.get(py)?.call1((aes128_algorithm, cbc_mode))?;
+    let encryptor = cipher.call_method0(pyo3::intern!(py, "encryptor"))?;
+    let encrypted_content_start =
+        encryptor.call_method1(pyo3::intern!(py, "update"), (padded_content,))?;
+    let encrypted_content_end = encryptor.call_method0(pyo3::intern!(py, "finalize"))?;
+    let encrypted_content = encrypted_content_start.add(encrypted_content_end)?;
```
alex, line 113: We've now got symmetric encryption repeated a few times. I'm going to refactor the duplication into a fucntion, then it can be used here.

src/rust/src/pkcs7.rs
```diff
@@ -416,6 +513,10 @@ pub(crate) fn create_submodule(
         serialize_certificates,
         &submod
     )?)?;
+    submod.add_function(pyo3::wrap_pyfunction_bound!(
+        encrypt_and_serialize,
+        &submod
+    )?)?;
```
alex, line 519: Merge conflict here, sorry!

tests/hazmat/primitives/test_pkcs7.py
```diff
@@ -155,6 +155,63 @@ def _pkcs7_verify(encoding, sig, msg, certs, options, backend):
...
+
+    ossl_cert = backend._cert2ossl(cert_recipient)
+    ossl_pkey = _key2ossl(pkey, backend)
+    out_bio = backend._create_mem_bio_gc()
+    res = backend._lib.PKCS7_decrypt(p7, ossl_pkey, ossl_cert, out_bio, flags)
+    backend.openssl_assert(res == 1)
+
+    return _read_mem_bio(out_bio, backend)
+
+
```
alex, line 214: Does rust-openssl have teh functions we need? It might make mroe sense to expose a test API from there (see the asn1 testcertificate API for an example of this).

What do you think @reaperhulk ?

src/rust/src/pkcs7.rs
alex: https://github.com/pyca/cryptography/pull/11273

facutuesca: Fixed, now this uses `pkcs12::symmetric_encrypt` too

src/rust/src/pkcs7.rs
facutuesca: fixed!

tests/hazmat/primitives/test_pkcs7.py
facutuesca: Makes sense, I removed the decrypt (and helper) Python functions from here, and added a Rust `pkcs7_decrypt` to `test_support.rs`, similar to the new `pkcs7_verify`.

facutuesca:
@alex I squashed the old commits into a single one, and then added two new commits with the changes since your last review

tests/hazmat/primitives/test_pkcs7.py
```diff
@@ -834,6 +834,245 @@ def test_add_multiple_additional_certs(self, backend):
...
+    )
+    def test_smime_encrypt_pem_encoding(self, backend, options):
+        data = b"hello world\n"
+        cert, private_key = _load_rsa_cert_key()
+        builder = (
+            pkcs7.PKCS7EnvelopeBuilder().set_data(data).add_recipient(cert)
+        )
+        enveloped = builder.encrypt(serialization.Encoding.PEM, options)
+        with open("msg.p7m", "wb") as f:
+            f.write(enveloped)
```
alex, line 1041: I assume this was debugging leftover?

alex (commented):
One comment, and this needs docs and a changelog entry, but otherwise looks great. Thanks for your work here!

tests/hazmat/primitives/test_pkcs7.py
facutuesca: yes, good catch! Fixed

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -177,7 +178,93 @@ def sign(
...
+                *self._recipients,
+                certificate,
+            ],
+        )
+
+    def encrypt(
+        self,
+        encoding: serialization.Encoding,
+        options: typing.Iterable[PKCS7Options],
+        backend: typing.Any = None,
```
reaperhulk, line 230: Why are we including this on new APIs?

src/rust/src/pkcs7.rs
reaperhulk: I’m fine with that unless 4.0 is broadly supported, at which point we should just do 4.0 only. Is there any reasonable way to survey current support?

facutuesca: I'm not sure actually. @woodruffw, do you know if there's a way?

reaperhulk: A rough way to determine this would be just to see if the major S/MIME clients all support it. e.g. does Mail.app? Does Outlook? Gmail has S/MIME, what about that? Thunderbird?

woodruffw: I did a bit of searching for this:

* I can't find references to AEAD cipher support (or other S/MIME 4.0 features) in the public docs for Mail.app, Outlook, or Gmail
* Thunderbird *appears* to default to S/MIME 3.x, e.g. using AES-128-CBC by default. It's unclear whether they support AEADs for verifying.
* Other parts of RFC 8551 appear to be unsupported by Thunderbird: https://bugzilla.mozilla.org/show_bug.cgi?id=1600776

woodruffw: With a bit more searching: as best I can tell, all of the major mail clients handle S/MIME encryption by parroting whatever cipher selection they receive (with some enforcing a floor, e.g. no 3DES unless explicitly enabled). Thunderbird *may* support the `SMIMECapabilities` extension for improved negotiation, but not unless explicitly enabled through an `about:config` setting. 

Support among other MUAs is similarly mixed or not present: https://security.stackexchange.com/questions/271353/any-client-still-supporting-the-s-mime-capabilities-extension-in-2023

TL;DR: It seems like MUAs are still pretty firmly on S/MIME 3.1 and 3.2.

woodruffw: This comment has a good summary as well: https://bugzilla.mozilla.org/show_bug.cgi?id=1847703#c7

reaperhulk (commented):
This generally looks good now, so just one final question: have we verified that S/MIME encrypted messages can be successfully decrypted by one or more of the major clients? I see we test decryption in OpenSSL and we follow the spec, but it'd be good to have that final bit of validation.

facutuesca:
> and this needs docs and a changelog entry,

I pushed a commit with the doc and changelog entry

> have we verified that S/MIME encrypted messages can be successfully decrypted by one or more of the major clients?

I'll do this next

alex:
Let us know once you've confirmed and we'll do a final review. Thanks for your work on this!

facutuesca:
@alex  @reaperhulk I verified that the output can be read by both Mail.app  and Thunderbird. The way I did it was by saving the output of `encrypt()` to an `.eml` file, and importing that file in both applications. Before adding the corresponding certificate and private key to the settings/keychain, this is what you see:

### Mail.app (no private key available)
macos-encrypted

### Thunderbird (no private key available)
thunderbird-encrypted

After importing the private key:

### Mail.app (private key in keychain)
macos-decrypted

### Thunderbird (private key in keychain)
thunderbird-decrypted

reaperhulk:
@facutuesca Thanks! Apologies for one final question here: Users are likely to want to do both signing and encryption now that we have them -- are our current APIs composable to achieve that or is there some gross PKCS7 data structure we need to implement still?

facutuesca:
> @facutuesca Thanks! Apologies for one final question here: Users are likely to want to do both signing and encryption now that we have them -- are our current APIs composable to achieve that or is there some gross PKCS7 data structure we need to implement still?

They are already composable: the S/MIME RFC specifies that creating signed+encrypted messages is just a matter of doing one operation on the output of the other:

>    The signed-only, enveloped-only, and compressed-only MIME formats can
   be nested.  This works because these formats are all MIME entities
   that encapsulate other MIME entities.

and

> It is possible to apply any of the signing, encrypting, and
   compressing operations in any order.  It is up to the implementer and
   the user to choose.

([src](https://datatracker.ietf.org/doc/html/rfc5751.html#section-3.6))

reaperhulk:
That's what I wanted to hear, thanks! 😄

reaperhulk (approved):

woodruffw:
> It is possible to apply any of the signing, encrypting, and
> compressing operations in any order. It is up to the implementer and
> the user to choose.

lol, very good cryptography design.

PR #10932, opened 2024-05-05
Apply ruff/flake8-implicit-str-concat rule ISC001

DimitriPapadopoulos:
ISC001 Implicitly concatenated string literals on one line

This rule is currently disabled because it conflicts with the formatter:
	https://github.com/astral-sh/ruff/issues/8272

These are typically left over by black when unfolding lines (#5324).


alex:
CI is broken and needs https://github.com/pyca/cryptography/pull/10936 to fix it. Once that merges this can be rebased.

PR #10934, opened 2024-05-05
Use raw string for literal backslashes

DimitriPapadopoulos:
I might be missing something, but I believe these backslashes require a raw string:
https://github.com/pyca/cryptography/blob/95131abed8f3cf0a45ed8b8ff948a36926c6c6b3/src/_cffi_src/openssl/cryptography.py#L8-L9
https://github.com/pyca/cryptography/blob/95131abed8f3cf0a45ed8b8ff948a36926c6c6b3/src/_cffi_src/openssl/cryptography.py#L23-L24
https://github.com/pyca/cryptography/blob/95131abed8f3cf0a45ed8b8ff948a36926c6c6b3/src/_cffi_src/openssl/cryptography.py#L49-L52

### DIFF
--- a/src/_cffi_src/openssl/cryptography.py
+++ b/src/_cffi_src/openssl/cryptography.py
@@ -4,7 +4,7 @@
 
 from __future__ import annotations
 
-INCLUDES = """
+INCLUDES = r"""
 /* define our OpenSSL API compatibility level to 1.1.0. Any symbols older than
    that will raise an error during compilation. */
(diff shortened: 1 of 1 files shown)


alex (approved):

alex:
CI is broken and needs https://github.com/pyca/cryptography/pull/10936 to fix it. Once that merges this can be rebased.

reaperhulk (approved):

PR #10935, opened 2024-05-05
Apply a refurb suggestion

DimitriPapadopoulos:
FURB110 Replace ternary `if` expression with `or` operator


alex (changes requested):
This doesn't seem like an improvement to me, in general explicit ternary is more readable than `or`.

PR #11009, opened 2024-05-22
Add monthly downloads badge to README

ayushjain01:
Added a monthly downloads badge from pip Trends to the README.
View more at - https://piptrends.com/widgets/cryptography


ixmatus:
@reaperhulk a colleague of this person tried to open this exact PR against a project I'm the maintainer for as well (https://github.com/ixmatus/inflector/pull/16). This is a promotional campaign and they (or colleagues) appear to be doing it on reddit too, I think this is suspicious activity. They are creating an external dependency on a web property that (as far as I can tell) is not officially blessed by pypi.

Users can get package statistics from `libraries.io` which is linked to directly by pypi itself.

PR #11022, opened 2024-05-25
add missing verification_time

milahu:


docs/x509/verification.rst
```diff
@@ -76,7 +76,8 @@ the root of trust:
     >>> with open(certifi.where(), "rb") as pems:
     ...    store = Store(load_pem_x509_certificates(pems.read()))
     >>> builder = PolicyBuilder().store(store)
-    >>> builder = builder.time(verification_time)
+    >>> # verification_time = datetime.now() # default
+    >>> # builder = builder.time(verification_time)
```
alex, line 80: You can't just comment this out, because then the doctest fails.

If you want to add a comment over the existing code, that's fine.

milahu: lets just remove `builder = builder.time(verification_time)`

in 99% of all use cases, `verification_time` is the default `datetime.now()`

docs/x509/verification.rst
```diff
@@ -76,7 +76,6 @@ the root of trust:
     >>> with open(certifi.where(), "rb") as pems:
     ...    store = Store(load_pem_x509_certificates(pems.read()))
     >>> builder = PolicyBuilder().store(store)
-    >>> builder = builder.time(verification_time)
```
alex, line 79: As I said, you cannot simply remove this, because it means the tests won't work.

If you want to add a comment above this saying that the default is now and it's not required in most cases, that's fine, but you can't remove it.

milahu: aah, because `verification_time` was never missing in the first place

```py
verification_time = datetime.fromisoformat("2024-01-12T00:00:00Z")
```

sorry for the noise

PR #11030, opened 2024-05-27
re-add branch we dropped in the past

reaperhulk:
let's see if this fails since we shouldn't have a test covering this branch...

### DIFF
--- a/src/rust/src/backend/utils.rs
+++ b/src/rust/src/backend/utils.rs
@@ -150,6 +150,13 @@ pub(crate) fn pkey_private_bytes<'p>(
                 let pem_bytes = if password.is_empty() {
                     rsa.private_key_to_pem()?
                 } else {
+                    if cryptography_openssl::fips::is_enabled() {
+                        return Err(CryptographyError::from(
+                            pyo3::exceptions::PyValueError::new_err(
+                                "Encrypted traditional OpenSSL format is not supported in FIPS mode",
+                            ),
+                        ));
+                    }
                     rsa.private_key_to_pem_passphrase(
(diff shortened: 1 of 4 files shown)


src/rust/src/backend/utils.rs
```diff
@@ -150,6 +150,13 @@ pub(crate) fn pkey_private_bytes<'p>(
                 let pem_bytes = if password.is_empty() {
                     rsa.private_key_to_pem()?
                 } else {
+                    if cryptography_openssl::fips::is_enabled() {
```
alex, line 153: This same issue is applicable to DSA and EC right? I think this should be higher up, at line 148 or so.

alex (approved):

alex (changes requested):

alex (approved):

tests/hazmat/primitives/test_dsa.py
```diff
@@ -726,6 +726,24 @@ def test_private_bytes_encrypted_pem(self, backend, fmt, password):
...
+        )
+        key = serialization.load_pem_private_key(key_bytes, None, backend)
+        assert isinstance(key, dsa.DSAPrivateKey)
+        with pytest.raises(ValueError):
+            key.private_bytes(
+                serialization.Encoding.PEM,
+                serialization.PrivateFormat.TraditionalOpenSSL,
+                serialization.BestAvailableEncryption(b"password"),
+            )
+
```
alex, line 746: This is never reachable, we don't support DSA with FIPS

alex (approved):

PR #11122, opened 2024-06-19
See if wheel dep is required

alex:
The setuptools changelog sort of implies its not anymore

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,6 @@
 requires = [
     # First version of setuptools to support pyproject.toml configuration
     "setuptools>=61.0.0",
-    "wheel",
     # Must be kept in sync with `project.dependencies`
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -4,7 +4,6 @@
 requires = [
     # First version of setuptools to support pyproject.toml configuration
     "setuptools>=61.0.0",
```
reaperhulk, line 6: Do we need to bump minimum setuptools dep?

alex: Unclear? Latest setuptools is 3.8+, but obviously our 3.7 tests are passing with this, which suggests it was never required?

reaperhulk (approved):
I don’t understand when wheel became superfluous but here we are.

PR #11124, opened 2024-06-19
Revert "pin python version for downstream tests due to twisted/python bug"

alex:
Reverts pyca/cryptography#11121


reaperhulk (approved):
Churn baby churn

PR #11137, opened 2024-06-22
Update reference.rst

ndragon798:
This code snippet works when importing the extension oid
Without it you get 
```
ext = cert.extensions.get_extension_for_oid(ExtensionOID.SUBJECT_ALTERNATIVE_NAME)
                                                ^^^^^^^^^^^^
NameError: name 'ExtensionOID' is not defined

```

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -2404,6 +2404,7 @@ X.509 Extensions
         .. doctest::
 
             >>> from cryptography import x509
+            >>> from cryptography.x509.oid import ExtensionOID
             >>> from cryptography.hazmat.primitives import hashes
             >>> cert = x509.load_pem_x509_certificate(cryptography_cert_pem)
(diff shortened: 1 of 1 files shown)


docs/x509/reference.rst
```diff
@@ -2404,6 +2404,7 @@ X.509 Extensions
         .. doctest::
 
             >>> from cryptography import x509
+            >>> from cryptography.x509.oid import ExtensionOID
             >>> from cryptography.hazmat.primitives import hashes
```
alex, line 2408:
```suggestion
            >>> from cryptography.hazmat.primitives import hashes
            >>> from cryptography.x509.oid import ExtensionOID
```

Small nit to keep teh imports ordered correctly

ndragon798: Applied.

alex (approved):

alex:
thanks

PR #11149, opened 2024-06-24
Remove the requirement for VerificationCertificate to be Clone

alex:
This is done by passing around references, rather than owned copies. Necessary for the pyo3 0.22 upgrade.


woodruffw (approved):
Can't think of a cleaner way to do this at the moment 😞

reaperhulk (approved):

reaperhulk:
This feels substantially less readable to me but c’est la vie.

Issue #11161, opened 2024-06-25
Extension errors should indicate the extension that is at issue


nbastin:
VerificationError can report "Certificate extension has incorrect criticality", but it doesn't tell you which extension it is checking.

PR #11162, opened 2024-06-25
policy/extension: improve extension policy errors

woodruffw:
This improves some (but not all, see comment) of the errors rendered on various extension policy check failures.

This can be further refined by mixing in an identifying feature of the certificate being verified, but I've left that out for this PR to keep the diff small.

Fixes https://github.com/pyca/cryptography/issues/11161.


src/rust/cryptography-x509-verification/src/policy/extension.rs
```diff
@@ -205,9 +205,10 @@ impl<B: CryptoOps> ExtensionValidator<B> {
             // Extension MUST NOT be present and isn't; OK.
             (ExtensionValidator::NotPresent, None) => Ok(()),
             // Extension MUST NOT be present but is; NOT OK.
-            (ExtensionValidator::NotPresent, Some(_)) => Err(ValidationError::Other(
-                "Certificate contains prohibited extension".to_string(),
-            )),
+            (ExtensionValidator::NotPresent, Some(extn)) => Err(ValidationError::Other(format!(
```
alex, line 208: IDK if this will make the PR way more complicated, but I think we shoudl do `ValidationError::ExtensionError` or something, and it can have the OID as structured data, and thus we can defer the string formatting until we're ready to show an error to the user.

WDYT?

woodruffw:
NB: This change doesn't improve the error in this case:

```rust
            // Extension MUST be present but is not; NOT OK.
            (ExtensionValidator::Present { .. }, None) => Err(ValidationError::Other(
                "Certificate is missing required extension".to_string(),
            )),
```

...since the error in this case happens when the extension validator is expecting an extension, but none is given. We previously had the OID from the validator itself, but that was removed as part of the performance refactor on `ExtensionValidator` to make it less magical. That'll probably require a large refactor.

src/rust/cryptography-x509-verification/src/policy/extension.rs
woodruffw: That wouldn't be too bad IMO -- I can either do that as part of this PR, or defer it to another as part of fixing https://github.com/pyca/cryptography/pull/11162#issuecomment-2190213329 as well, your preference :slightly_smiling_face:

alex: Let's do it here/now.

woodruffw:
Re-drafting this while I iterate on it.

src/rust/src/x509/verify.rs
```diff
@@ -282,7 +282,7 @@ impl PyClientVerifier {
             policy,
             store.raw.borrow_dependent(),
         )
-        .map_err(|e| VerificationError::new_err(format!("validation failed: {e:?}")))?;
+        .map_err(|e| VerificationError::new_err(format!("validation failed: {e}")))?;
```
woodruffw, line 285: NB: This will make all of the Python-surfaced errors much nicer, now that we have `Display for VerificationError`.

src/rust/cryptography-x509-verification/src/lib.rs
```diff
@@ -35,10 +37,45 @@ pub enum ValidationError {
...
+                write!(f, "candidates exhausted: {inner}")
+            }
+            ValidationError::Malformed(err) => write!(f, "malformed ASN.1: {err}"),
+            ValidationError::DuplicateExtension(DuplicateExtensionsError(oid)) => {
+                write!(f, "malformed certificate: duplicate extension: {oid}")
+            }
+            ValidationError::ExtensionError { oid, reason } => {
+                write!(f, "invalid extension: {oid}: {reason}")
+            }
+            ValidationError::FatalError(err) => write!(f, "fatal error: {err}"),
```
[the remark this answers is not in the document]
woodruffw: Yeah, those are going to be annoying to reach from Python without some really deep new testcases. I'll write the Rust-side tests....

alex (approved):

PR #11193, opened 2024-07-03
Add rsa_recover_private_exponent function

dlenskiSB:
Given the RSA public exponent (`e`), and the RSA primes (`p`, `q`), it is possible to calculate the corresponding private exponent `d = e⁻¹ mod λ(n)` where `λ(n) = lcm(p-1, q-1)`.

With this function added, it becomes possible to use the library to reconstruct an RSA private key given *only* `p`, `q`, and `e`:

```py

n = p * q
d = rsa.rsa_recover_private_exponent(e, p, q)  # newly-added piece
iqmp = rsa.rsa_crt_iqmp(p, q)                  # preexisting
dmp1 = rsa.rsa_crt_dmp1(d, p)                  # preexisting
dmq1 = rsa.rsa_crt_dmq1(d, q)                  # preexisting

assert rsa.rsa_recover_prime_factors(n, e, d) in ((p, q), (q, p))  # verify consistency

privk = rsa.RSAPrivateNumbers(p, q, d, dmp1, dmq1, iqmp, rsa.RSAPublicNumbers(e, n)).private_key()
```

Older RSA implementations, including the original RSA paper, often used the Euler totient function `ɸ(n) = (p-1) * (q-1)` instead of `λ(n)`.  The private exponents generated by that method work equally well, but may be larger than strictly necessary (`λ(n)` always divides `ɸ(n)`).  This commit additionally implements `_rsa_recover_euler_private_exponent`, so that tests of the internal structure of RSA private keys can allow for either the Euler or the Carmichael versions of the private exponents.

It makes sense to expose only the more modern version (using the Carmichael totient function) for public usage, given that it is slightly more computationally efficient to use the keys in this form, and that some standards like FIPS 186-4 require this form.  (See https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf#page=63)


tests/hazmat/primitives/utils.py
```diff
@@ -528,7 +528,10 @@ def _check_rsa_private_numbers(skey):
     assert pkey
     assert pkey.e
     assert pkey.n
-    assert skey.d
+    assert skey.d in (
+        rsa.rsa_recover_private_exponent(pkey.e, skey.p, skey.q),
+        rsa._rsa_recover_euler_private_exponent(pkey.e, skey.p, skey.q),
+    )
```
dlenskiSB, line 534: There aretwo different ways to calculate valid/working `d` given `e`, `p`, `q`.

In my opinion, it makes sense to expose _only_ the more modern one using the Carmichael totient function for public usage, given that it's slightly more computationally efficient to use the keys in this form, and that some standards require it:

https://en.wikipedia.org/wiki/RSA_(cryptosystem)#:~:text=In%20the%20original,smaller%20equivalent%20exponent.

docs/hazmat/primitives/asymmetric/rsa.rst
```diff
@@ -554,6 +554,11 @@ this without having to do the math themselves.
     Computes the ``dmq1`` parameter from the RSA private exponent (``d``) and
     prime ``q``.
 
+.. function:: rsa_recover_private_exponent(n, e, d)
```
alex, line 557: Signature doesn't look right

src/cryptography/hazmat/primitives/asymmetric/rsa.py
```diff
@@ -190,6 +190,38 @@ def rsa_crt_dmq1(private_exponent: int, q: int) -> int:
...
+    # The original RSA paper uses the Euler totient function
+    # here: phi_n = (p - 1) * (q - 1)
+    # Either version of the private exponent will work, but the
+    # one generated by the older formulation may be larger
+    # than necessary. (lambda_n always divides phi_n)
+    lambda_n = (p - 1) * (q - 1) // gcd(p - 1, q - 1)
+    return _modinv(e, lambda_n)
+
+
+def _rsa_recover_euler_private_exponent(e: int, p: int, q: int) -> int:
```
alex, line 211: Looks unused?

alex (commented):
Two observations. Will defer the meat of the review to @reaperhulk

src/cryptography/hazmat/primitives/asymmetric/rsa.py
dlenskiSB: See the test code, and my comment there for why it's needed.

docs/hazmat/primitives/asymmetric/rsa.rst
dlenskiSB: Fixed in 7ffb0b0b0, thank you.

src/cryptography/hazmat/primitives/asymmetric/rsa.py
```diff
@@ -190,6 +190,38 @@ def rsa_crt_dmq1(private_exponent: int, q: int) -> int:
...
+    This uses the Carmichael totient function to generate the
+    smallest possible working value of the private exponent.
+    """
+    # This lambda_n is the Carmichael totient function.
+    # The original RSA paper uses the Euler totient function
+    # here: phi_n = (p - 1) * (q - 1)
+    # Either version of the private exponent will work, but the
+    # one generated by the older formulation may be larger
+    # than necessary. (lambda_n always divides phi_n)
+    lambda_n = (p - 1) * (q - 1) // gcd(p - 1, q - 1)
```
reaperhulk, line 207: It's a shame `math.lcm` wasn't added until Python 3.9, since this function could just be `_modinv(e, lcm(p - 1, q -1 ))` otherwise.

reaperhulk (commented):
generally lgtm, just needs changelog and doc updates.

docs/hazmat/primitives/asymmetric/rsa.rst
```diff
@@ -554,6 +554,11 @@ this without having to do the math themselves.
     Computes the ``dmq1`` parameter from the RSA private exponent (``d``) and
     prime ``q``.
 
+.. function:: rsa_recover_private_exponent(e, p, q)
+
```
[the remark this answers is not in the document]
dlenskiSB: Got it. Both done in 678e758ea

reaperhulk:
Looks like you'll need to add `totient` to `docs/spelling_wordlist.txt`

src/cryptography/hazmat/primitives/asymmetric/rsa.py
alex: This should probably just be in the tests file if it's not used in production.

src/cryptography/hazmat/primitives/asymmetric/rsa.py
alex: Maybe add a comment reminding us to do that when we bump our minimum Python

dlenskiSB:
> Looks like you'll need to add totient to docs/spelling_wordlist.txt

@reaperhulk Yep, just fixed that. (And added `Euler` and `Carmichael` while we're at it.)

src/cryptography/hazmat/primitives/asymmetric/rsa.py
reaperhulk, line 211: Let's move this implementation into the test package since it has no use outside of the tests.

src/cryptography/hazmat/primitives/asymmetric/rsa.py
reaperhulk: Yeah, in a year or so we can simplify this then 😄

reaperhulk (commented):
Two final small requests, then this is ready. Thanks!

src/cryptography/hazmat/primitives/asymmetric/rsa.py
dlenski: Added a TODO note in 5a58cb2cca2f302e28195b37b01a1f6014611af6

src/cryptography/hazmat/primitives/asymmetric/rsa.py
dlenski: Done in 5a58cb2cca2f302e28195b37b01a1f6014611af6

dlenskiSB: Done in https://github.com/pyca/cryptography/commit/5a58cb2cca2f302e28195b37b01a1f6014611af6

src/cryptography/hazmat/primitives/asymmetric/rsa.py
dlenskiSB: Added a TODO note in https://github.com/pyca/cryptography/commit/5a58cb2cca2f302e28195b37b01a1f6014611af6, @alex.

reaperhulk (approved):

PR #11214, opened 2024-07-07
verification: remove an error variant

woodruffw:
`DuplicateExtension` is well-covered under the `ExtensionError` variant, so this removes it. NFC otherwise.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -265,6 +265,7 @@ def local(session):
         *test_dependencies,
         *pyproject_data["project"]["optional-dependencies"]["ssh"],
         *pyproject_data["project"]["optional-dependencies"]["nox"],
+        "-e",
         "./vectors/",
         verbose=False,
     )
--- a/src/rust/cryptography-x509-verification/src/lib.rs
+++ b/src/rust/cryptography-x509-verification/src/lib.rs
@@ -36,7 +36,6 @@ use crate::ApplyNameConstraintStatus::{Applied, Skipped};
 pub enum ValidationError {
     CandidatesExhausted(Box<ValidationError>),
     Malformed(asn1::ParseError),
-    DuplicateExtension(DuplicateExtensionsError),
     ExtensionError {
         oid: ObjectIdentifier,
         reason: &'static str,
@@ -53,7 +52,10 @@ impl From<asn1::ParseError> for ValidationError {
 
 impl From<DuplicateExtensionsError> for ValidationError {
     fn from(value: DuplicateExtensionsError) -> Self {
-        Self::DuplicateExtension(value)
+        Self::ExtensionError {
+            oid: value.0,
+            reason: "duplicate extension",
+        }
     }
 }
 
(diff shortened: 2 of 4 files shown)


src/rust/cryptography-x509-verification/src/lib.rs
```diff
@@ -53,7 +52,10 @@ impl From<asn1::ParseError> for ValidationError {
 
 impl From<DuplicateExtensionsError> for ValidationError {
     fn from(value: DuplicateExtensionsError) -> Self {
-        Self::DuplicateExtension(value)
+        Self::ExtensionError {
+            oid: value.0,
+            reason: "duplicate extension",
+        }
     }
 }
```
woodruffw, line 60: NB: Could remove this entire `From<...>` impl as well.

alex (approved):

src/rust/cryptography-x509-verification/src/lib.rs
```diff
@@ -452,12 +449,13 @@ mod tests {
         let err = ValidationError::Malformed(ParseError::new(asn1::ParseErrorKind::InvalidLength));
         assert_eq!(err.to_string(), "ASN.1 parsing error: invalid length");
 
-        let err = ValidationError::DuplicateExtension(DuplicateExtensionsError(
-            SUBJECT_ALTERNATIVE_NAME_OID,
-        ));
+        let err = ValidationError::ExtensionError {
```
alex, line 452: This isn't meaningfully testing anything any more.

Should either go via the from() impl or be deleted

noxfile.py
```diff
@@ -265,6 +265,7 @@ def local(session):
         *test_dependencies,
         *pyproject_data["project"]["optional-dependencies"]["ssh"],
         *pyproject_data["project"]["optional-dependencies"]["nox"],
+        "-e",
```
alex, line 268: This shouldn't be required?

woodruffw, line 268: NB: Noticed during local testing that changes to vectors weren't being propagated between runs here; adding `-e` fixes the problem (and is similarly used in other environments in the noxfile).

alex (approved):

noxfile.py
woodruffw: I didn't think so either, but adding it resolved some surprising test failures locally: without it, I think `pip install ./vectors/` was performing a no-op due to the venv's installed version matching the local copy, even though the local copy had new changes since the last time the local venv was created.

alex:
The `debug` on duplicate extension is no longer covered. can just remove that I guess

alex (approved):

noxfile.py
reaperhulk: I have seen this behavior in `nox -e local` as well, although in my case it persisted even when deleting the .nox dir and a uv upgrade appeared to resolve it? Our aggressive pursuit of speed seems to have resulted in some occasionally caching issues…

woodruffw: Yeah, I saw the same -- persisted even when `.nox` was blown away...

PR #11249, opened 2024-07-10
Support loading Ed448 public keys in OpenSSH format

dlenskiSB:
The 'ssh-ed448' key type is documented along with 'ssh-ed25519' in [1], but has never been supported by any as-yet-released version of OpenSSH.

However, LANcom router devices (which appear to be primarily used in Germany, see [2] for examples on the public Internet) appear to support these keys, so this library can and should support loading them.

Ed448 private keys are not yet implemented here, because OpenSSH itself does not yet support them, and it is the de facto authority for private key formats.  However, PuTTY has already implemented support for generating and
using Ed448 keys, and the PuTTY developers note in [3] that the OpenSSH developers are in agreement with them as to the correct Ed448 private key format:

> I checked with them [OpenSSH developers], and they agreed that there's an obviously right format for Ed448 keys, which is to do them exactly like Ed25519 except that you have a 57-byte string everywhere Ed25519 had a 32-byte string.  So I've done that.

See also [4] in which I extended `ssh-audit` to allow it to scan and discover host keys of type 'ssh-ed488'.

[1] https://datatracker.ietf.org/doc/html/rfc8709#name-public-key-format
[2] https://www.shodan.io/search?query=ssh+%22ed448%22
[3] https://github.com/github/putty/commit/a085acbadf829ac5b426323ca98058d6aa4048ba
[4] https://github.com/jtesta/ssh-audit/pull/277


src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -582,6 +588,59 @@ def encode_private(
...
+        )
+        f_pub.put_sshstr(raw_public_key)
+
+    def encode_private(
+        self, private_key: ed448.Ed448PrivateKey, f_priv: _FragList
+    ) -> None:
+        """Write Ed448 private key"""
+        raise UnsupportedAlgorithm(
+            "Serializing Ed448 SSH private keys is unsupported"
+        )
```
dlenskiSB, line 641: I think it's *quite likely* that Ed448 private keys for OpenSSH would (will?) have a format very similar to Ed25519 private keys (e.g. https://github.com/pyca/cryptography/blob/d3eda718f1fc9fc5510bfb80127f7f2649cbfea2/tests/hazmat/primitives/test_ssh.py#L520-L530), but since I haven't seen any real-world examples of private keys *and* since the Ed25519 format is weirdly redundant in terms of its repetition of private and public key parts… I'm reluctant to make any assumptions about it here.

alex:
Are ed448 keys documented by OpenSSH? `ssh-keygen -ted448` doesn't appear to work.

dlenskiSB:
> Are ed448 keys documented by OpenSSH? ssh-keygen -ted448 doesn't appear to work.

@alex, they are not supported by OpenSSH, as I write in the commit message: `The 'ssh-ed448' key type … has never been supported by any as-yet-released version of OpenSSH.`

But their public key format is [_documented_ in RFC8709](https://datatracker.ietf.org/doc/html/rfc8709#name-public-key-format), alongside the format for Ed25519 keys which OpenSSH _does_ support.

Although OpenSSH doesn't support them, other SSHv2-compatible software apparently does. (Including the widely-used [Putty](https://www.chiark.greenend.org.uk/~sgtatham/putty/wishlist/ed448.html) client, v0.75+)

src/cryptography/hazmat/primitives/serialization/ssh.py
dlenskiSB: On the other hand, from https://github.com/github/putty/commit/a085acbadf829ac5b426323ca98058d6aa4048ba …

> since the OpenSSH agent protocol is the de facto
standard … OpenSSH themselves get to
say what the format for a key should or shouldn't be. So if they don't
support a particular key method, what do you do?
>
> I checked with them, and they agreed that there's an obviously right
format for Ed448 keys, which is to ***do them exactly like Ed25519 except
that you have a 57-byte string everywhere Ed25519 had a 32-byte
string***. So I've done that.

So we could follow suit and support an OpenSSH-like Ed448 _private_ key format just as Putty does.

dlenskiSB:
The [flake8 failure here](https://github.com/pyca/cryptography/actions/runs/9951346163/job/27490679803?pr=11249#step:14:65) is due to it objecting to the preexisting sort order of an `import` block. 🤷🏻‍♂️

dlenskiSB:
> Are you sure you don’t just need to put ed448 before ed25519 in the imports?

Apparently so. I had no idea that numeric substrings in module names get sorted like this, woah 😵
https://github.com/PyCQA/isort/issues/1732#issuecomment-846350228

vectors/cryptography_vectors/asymmetric/OpenSSH/ed448-nopsw.key.pub
```diff
@@ -0,0 +1 @@
+ssh-ed448 AAAACXNzaC1lZDQ0OAAAADnVY+2PC4Oj9MSsYZORD7xivKK3zyyHFKYj3eMCMPsAwNVk6fqGHeSIRDN39ld5Jto8S5Y1lemtJHA=
```
alex, line 1: Vectors should be documented in `test-vectors.rst`, thanks

dlenskiSB: Thanks, done in 0ac431c.

reaperhulk (dismissed):

reaperhulk:
I'm comfortable with merging public key parsing support (let's wait on private until we see user demand or it lands in a prominent server implementation). At the moment there are some missing lines of coverage though. You can see a visual report by downloading the html-report from the artifacts here: https://github.com/pyca/cryptography/actions/runs/9981453975?pr=11249

fochoao1986:
I am not sure this is safe at all, I tried doing so manually via openssl and ssh-keygen, it requires of OpenSC.

alex:
Were you interested in continuing to work on this? We're happy to review once tests are passing.

reaperhulk:
Closing due to lack of activity, but please reopen if you decide to rebase and finish this!

PR #11288, opened 2024-07-17
Bump vectors

woodruffw:
Allows `build_server_verifier` to fail when failures are expected.

Replaces #11287.

### DIFF
--- a/tests/x509/verification/test_limbo.py
+++ b/tests/x509/verification/test_limbo.py
@@ -139,6 +139,7 @@ def _limbo_testcase(id_, testcase):
             verifier = builder.build_server_verifier(peer_name)
         except ValueError:
             assert not should_pass
+            return
     else:
         assert testcase["extended_key_usage"] == ["clientAuth"]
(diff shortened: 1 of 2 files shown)


tests/x509/verification/test_limbo.py
```diff
@@ -133,7 +133,12 @@ def _limbo_testcase(id_, testcase):
             "extended_key_usage"
         ] == ["serverAuth"]
         peer_name = _get_limbo_peer(testcase["expected_peer_name"])
-        verifier = builder.build_server_verifier(peer_name)
+        # Some tests exercise invalid leaf SANs, which get caught before
+        # validation even begins.
+        try:
+            verifier = builder.build_server_verifier(peer_name)
+        except ValueError:
+            assert not should_pass
```
alex, line 141: This will just name error below, won't it?

woodruffw: Ah, yep. Needs a return.

alex (approved):

PR #11328, opened 2024-07-22
Don't include engine.h when OPENSSL_NO_ENGINE is defined

tiran:
Fedora 41 and RHEL 10 are deprecating and phasing out OpenSSL ENGINE support. Downstream has moved `openssl/engine.h` into a separate RPM package and is recompiling packages with `-DOPENSSL_NO_ENGINE=1`. The compiler flag disables PyCA cryptography's ENGINE support successfully. We also like to build the downstream package without the `engine.h` header file present.

This commit makes the include conditional. The `ENGINE` type is defined in `openssl/types.h`.

See: https://src.fedoraproject.org/rpms/openssl/c/e67e9d9c40cd2cb9547e539c658e2b63f2736762?branch=rawhide
See: https://issues.redhat.com/browse/RHEL-33747


alex:
This looks fine. FWIW we already test and build with openssl built with `no-engine`, which leaves this header but simply defines nothing. It's strange to me that `no-engine` and `OPENSSL_NO_ENGINE` have different effects.

reaperhulk:
Yes, why is RHEL removing the header entirely rather than the approach `no-engine` takes?

h-vetinari:
From a packager's perspective (though not speaking for Fedora/RedHat/etc.), this makes a lot of sense, because it saves the analysis of whether a given package still relies on engine bits _despite_ `no-engine` (whether through an intentional hack or unintentionally through a bug or build system leak).

It's a bit like double book-keeping; one side is what a given package says it depends on, and the other is what the packager actually provides for building. The two should agree, and having both sides makes it easier/faster to catch bugs, especially at scale.

reaperhulk:
I'm fine with this, but @tiran is it ready? Did you leave it in draft intentionally?

alex:
No response for 10 days, so I'm going to close this out. Leave a comment if you'd like to reopen.

Issue #11352, opened 2024-07-28
Certificate chain validation always fails with "required EKU not found" error


kislyuk:
Hello,

I am trying to validate certificate chains with the `build_client_verifier()` functionality introduced in Cryptography 43. I am finding that verification never succeeds on any certificate chains in use by my counterparty applications, always failing with a "required EKU not found" error.

Upon some digging, EKU (extended key usage) is never set by any applications in my application domain (XML signature/SAML). Key usage can be reasonably expected to be set to "Digital signature" in this case, but EKU should not be expected to be set (and I can't find an appropriate value for it).

Reading through the code (https://github.com/pyca/cryptography/blob/edc43b1d69c7606fd2c7e7e1ace1b6312d8b9565/src/rust/cryptography-x509-verification/src/policy/mod.rs#L333-L338, https://github.com/pyca/cryptography/blob/edc43b1d69c7606fd2c7e7e1ace1b6312d8b9565/src/rust/cryptography-x509-verification/src/policy/extension.rs#L334-L344) it looks to me like there are two issues:

* There is no way to control the EKU assertion behavior in the client verifier policy builder (unlike validation time, max chain depth, etc.)
* The intended behavior appears to be to permit the absence of an EKU, but the actual behavior is to forbid it.

Does this sound right? If so, what's the best way to fix this?

alex:
cc: @woodruffw

woodruffw:
Thanks for the ping! @kislyuk could you provide an example cert you're seeing this with? The "maybe present" logic is pretty well trodden on other extensions so I'd be surprised if it was causing issues just on EKU (and we should have testcases for EKU not present), but an example would help me better isolate this 🙂

PR #11364, opened 2024-07-29
extensions: forbid empty EKU set

woodruffw:
This was previously caught indirectly in path validation, via an `any(...)` check over the listed EKUs which would then fail when none were present. Moving it into the type's construction will hopefully make it clearer that the EKU is outright malformed.

xref https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12

See: https://github.com/pyca/cryptography/issues/11352

### DIFF
--- a/src/rust/cryptography-x509/src/extensions.rs
+++ b/src/rust/cryptography-x509/src/extensions.rs
@@ -231,7 +231,32 @@ pub struct BasicConstraints {
 
 pub type SubjectAlternativeName<'a> = asn1::SequenceOf<'a, name::GeneralName<'a>>;
 pub type IssuerAlternativeName<'a> = asn1::SequenceOf<'a, name::GeneralName<'a>>;
-pub type ExtendedKeyUsage<'a> = asn1::SequenceOf<'a, asn1::ObjectIdentifier>;
+
+pub struct ExtendedKeyUsage<'a>(asn1::SequenceOf<'a, asn1::ObjectIdentifier>);
+
+impl<'a> Iterator for ExtendedKeyUsage<'a> {
+    type Item = asn1::ObjectIdentifier;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.next()
+    }
+}
+
+impl<'a> asn1::SimpleAsn1Readable<'a> for ExtendedKeyUsage<'a> {
+    const TAG: asn1::Tag = asn1::Sequence::TAG;
+
+    fn parse_data(data: &'a [u8]) -> asn1::ParseResult<Self> {
+        let seq = asn1::SequenceOf::parse_data(data)?;
+
+        // RFC 5280 4.2.1.12: EKU contains one or more purposes, and therefore
+        // must not be empty.
+        if seq.is_empty() {
+            return Err(asn1::ParseError::new(asn1::ParseErrorKind::InvalidValue));
+        }
+
+        Ok(Self(seq))
+    }
+}
 
 pub struct KeyUsage<'a>(asn1::BitString<'a>);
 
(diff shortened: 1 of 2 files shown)


src/rust/cryptography-x509/src/extensions.rs
```diff
@@ -231,7 +231,32 @@ pub struct BasicConstraints {
...
+
+pub struct ExtendedKeyUsage<'a>(asn1::SequenceOf<'a, asn1::ObjectIdentifier>);
+
+impl<'a> Iterator for ExtendedKeyUsage<'a> {
+    type Item = asn1::ObjectIdentifier;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.next()
+    }
+}
```
woodruffw, line 243: NB: Added this to avoid making the newtype's inner member public just for iteration, but I could save these ~5 lines by doing that instead 🙂

src/rust/cryptography-x509/src/extensions.rs
```diff
@@ -231,7 +231,32 @@ pub struct BasicConstraints {
...
+
+impl<'a> asn1::SimpleAsn1Readable<'a> for ExtendedKeyUsage<'a> {
```
alex, line 245: Do we need a rust-asn1 builtin way to enforce size constraints? Needing to write this for everything feels painful

woodruffw: Yeah, that would be fantastic. Do you want me to hold off on this and look at generalizing?

woodruffw: (One idea: there could be a `NonemptySequenceOf` or similar.)

alex: I think it'd be good to at least figure out what our options are for doing this in rust-asn1. Maybe it'll be more trouble than it's worth, but this is far from the only place for this.

woodruffw: Gotcha -- for this kind of pattern in particular, some options I can think of are:

1. Retain the `pub type = ...` pattern, but add a new set of constraining `rust-asn1` types (e.g. `NonemptySequenceOf`, `NonemptySetOf`). 
2. Prefer the newtype pattern, and offer additional proc-macro annotation support, e.g.:

```rust
struct Whatever<'a>(#[nonempty] asn1::SequenceOf<'a, asn1::ObjectIdentifier>);
```

I personally find (2) a little bit cleaner, but it's probably slightly less trivial to implement 🙂

alex: Another idea: make it `struct SequenceOf<T, const MINIMUM_SIZE: usize = 0, const MAXIMUM_SIZE: usize = usize::MAX>`

woodruffw: That seems a lot nicer!

alex: Yeah, I kind of like it, my concerns are:

1. Will it break type inferences or otherwise cause pain for users who don't override min/max size?
2. This means you'll no longer be able to do anything with an instance with different bounds, even if it's already parsed.
3. Does it have incorrect generality? (size bounds are really applicable to a whole bunch of types, do we have to repeat these everywhere?)

woodruffw: Yeah, hmm. I can play around with it a bit and see if I can answer those.

alex: Dope. If it looks viable, feel free to go directly to rust-asn1 PR

woodruffw: Opened https://github.com/alex/rust-asn1/pull/470 as a WIP for the above; I'll try it against `cryptography-x509` tonight if I have the time 🙂

woodruffw:
Closing since https://github.com/alex/rust-asn1/pull/470 will make this approach suboptimal once a new release of `rust-asn1` is made. Once that's public I'll make a PR here to use those const generics 🙂

PR #11376, opened 2024-07-31
Add information about PolicyBuilder internal clock.

adiroiban:
This is just a minor update for an error that I have observed while reading the docs.

### DIFF
--- a/docs/x509/verification.rst
+++ b/docs/x509/verification.rst
@@ -76,10 +76,13 @@ the root of trust:
     >>> with open(certifi.where(), "rb") as pems:
     ...    store = Store(load_pem_x509_certificates(pems.read()))
     >>> builder = PolicyBuilder().store(store)
-    >>> # If significant time passes between the PolicyBuilder is initialized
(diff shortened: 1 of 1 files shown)


docs/x509/verification.rst
```diff
@@ -76,7 +76,10 @@ the root of trust:
     >>> with open(certifi.where(), "rb") as pems:
     ...    store = Store(load_pem_x509_certificates(pems.read()))
     >>> builder = PolicyBuilder().store(store)
-    >>> builder = builder.time(verification_time)
+    >>> # If significant time passes between the PolicyBuilder is initialized
+    >>> # and the time the validation is perform, you will need to update
+    >>> # the internal clock.
+    >>> builder = builder.time(datetime.now())
```
alex, line 82: We run these as doc tests and this will fail when the cert expires, so we can'd do that.

I'd suggest just doing a comment explaining this, without changing the code.

adiroiban:
I think that this is ready for review.

I have added a comment to explain what is going on there.

PR #11407, opened 2024-08-08
chore: improve deprecation messages

JohnVillalovos:
There has been confusion regarding the current deprecation messages as some are reading them as the algorithms will be removed from the cryptography library. When in reality they are just being removed from the module. Make it more explicit about it being removed.

An example of the confusion:
https://github.com/paramiko/paramiko/pull/2421#issuecomment-2276253111


alex (approved):
Thank you! This is a great improvement

PR #11422, opened 2024-08-12
Do not use mutable argument defaults

DimitriPapadopoulos:
[Python Mutable Defaults Are The Source of All Evil](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil)

### DIFF
--- a/src/cryptography/hazmat/primitives/serialization/pkcs7.py
+++ b/src/cryptography/hazmat/primitives/serialization/pkcs7.py
@@ -56,12 +56,12 @@ class PKCS7SignatureBuilder:
                 PKCS7HashTypes,
                 padding.PSS | padding.PKCS1v15 | None,
             ]
-        ] = [],
-        additional_certs: list[x509.Certificate] = [],
+        ] | None = None,
+        additional_certs: list[x509.Certificate] | None = None,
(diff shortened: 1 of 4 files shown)


alex:
a) Mono-causality is, in fact, the root of all evil.
b) All the objects in question here are immutable, these various list attributes are never mutated. Therefore this PR does not improve safety.

As a result, thank you for the contribution, but we're not going to merge this.

DimitriPapadopoulos:
If they're immutable, how about using tuples instead of lists?

alex:
tuples and lists are not semantically the same: tuples are fixed-length heterogeneous containers, lists are arbitrary length homogeneous containers. using tuples would be semantically wrong.

DimitriPapadopoulos:
immutable ≠ not mutated

DimitriPapadopoulos:
> [...] these various list attributes are never mutated.

If they are not mutated, they could changed into tuples couldn't they?

DimitriPapadopoulos:
Then reopen a new PR? I think you misunderstand what this was about.

alex:
No. We are not accepting this PR. Please stop bothering us about it.

PR #11423, opened 2024-08-12
Fix misspelling

DimitriPapadopoulos:


alex:
This file is directly imported from NIST's CAVP (see `test-vectors.rst` for full details), and this typo (still) exists in the original. We want these to precisely match the original.

Again, thanks for the PR, but as a result of ^ I'm going to close this.

PR #11428, opened 2024-08-13
add .DS_Store to .gitignore

sagarishere:
adding **/*.DS_Store to prevent Mac users from accidently committing .DS_Store files.

Just making life easier for developers using Macs.


reaperhulk:
Thanks for the contribution, but generally we suggest that people on macOS (which include the two primary developers of this project!) add `.DS_Store` in their global gitignore rather than putting it in every project individually.

PR #11443, opened 2024-08-17
Bump RSA-512 test keys to RSA-2048

davidben:
RSA-512 was broken in 1999. cryptography.io should not be requesting its backend library support it in 2024.


alex:
`test-vectors.rst` needs to be updated, these are referenced at 

```
* OpenSSL PEM RSA serialization vectors from the `OpenSSL example key`_ and
  `GnuTLS key parsing tests`_.
```

alex:
The docs for those should be moved down to the custom section now

davidben:
Ah, didn't see the custom section. One moment...

davidben:
Done.

alex (approved):

PR #11444, opened 2024-08-17
Remove unnecessary test dependencies on RSA-512

davidben:
`test_unsupported_hash` and `test_prehashed_digest_mismatch` work just fine with realistic RSA key sizes. (They also, as written, silently test nothing when the backend rejects RSA-512. As a reminder, RSA-512 was broken since 1999.)

`test_rsa_fips_small_key` wants a small key, but I assume RSA-1024 is fine.


alex (dismissed):

alex:
... the RHEL-8 in FIPS mode is apparently ok with a 1024-bit RSA key? I quit.

davidben:
Amazing. Restored that one.

alex (approved):

PR #11462, opened 2024-08-20
Relax root CA AKI field checks

### LINKED ISSUE #11461: Verification may be falsely rejected when root cert has authorityCertIssuer field
Cryptography's x509 verification appears to reject verification on a certificate when the authorityCertIssuer field is present in the authority key identifier extension of the root certificate. This may be correct per CABF baseline, but is suspiciously different from all other implementations I tested.

I have [written a small script that isolates the issue](https://gist.github.com/mxsasha/9ca38da42c08f7ec9d1a120f6c82f9d8). This fails on `cryptography.hazmat.bindings._rust.x509.VerificationError: validation failed: CandidatesExhausted(Other("authorityKeyIdentifier must not contain authorityCertIssuer"))`. The leaf and intermediate do not contain authorityCertIssuer - the root does. 

The error message [originates in the code here](https://github.com/pyca/cryptography/blob/7fda121e69f40b16cc8bf46f9f7ea8cf217e88cb/src/rust/cryptography-x509-verification/src/policy/extension.rs#L418). This code seems to have an exception that allows the authority key identifier extension to be absent, but if it is found, the presence of authorityCertIssuer will lead to an error.

This is a widely trusted root certificate, and no browsers and other testing tools don't seem to have an issue with it. I do see the [CABF baseline](https://cabforum.org/working-groups/server/baseline-requirements/requirements/#71213-root-ca-authority-key-identifier) does not allow authorityCertIssuer. But this was [introduced in 2020](https://github.com/cabforum/servercert/pull/195), the root certificate is from 2006. Perhaps there are other details I don't know - I'm not too familiar with these policies and their implementations. It definitely seems strange and undesirable that cryptography rejects a certificate where other implementations find it valid.

It would be great if someone with more experience in this area could look at this and determine if cryptography is too strict in this case, and what the correct behavior would be.

For completeness, my versions, with cryptography installed in a venv through pip install:
```
cffi==1.15.1
cryptography==43.0.0
pip-tools==6.14.0
setuptools==73.0.0

Python 3.9.6 (default, Nov 10 2023, 13:38:27) 
[Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
```

woodruffw:
This addresses #11461 by dropping our forbidden AKI field checks on CAs. It also marks three x509-limbo tests as explicit skips, since these three tests enforced failure on those AKI fields.

CC @mxsasha

Closes #11461.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,9 @@ Changelog
   not be empty.
 * Added support for timestamp extraction to the
   :class:`~cryptography.fernet.MultiFernet` class.
+* Relax the Authority Key Identifier requirements on root CA certificates
+  during X.509 verification to allow fields permitted by :rfc:`5280` but
+  forbidden by the CA/Browser BRs.
 
 .. _v43-0-0:
 
--- a/src/rust/cryptography-x509-verification/src/policy/extension.rs
+++ b/src/rust/cryptography-x509-verification/src/policy/extension.rs
@@ -412,18 +412,13 @@ pub(crate) mod ca {
                 ));
             }
 
-            // authorityCertIssuer and authorityCertSerialNumber MUST NOT be present.
-            if aki.authority_cert_issuer.is_some() {
-                return Err(ValidationError::Other(
-                    "authorityKeyIdentifier must not contain authorityCertIssuer".to_string(),
-                ));
-            }
-
-            if aki.authority_cert_serial_number.is_some() {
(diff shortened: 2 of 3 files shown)


reaperhulk:
Let’s add a changelog entry for this, otherwise 👍

woodruffw:
Done!

reaperhulk (approved):

PR #11467, opened 2024-08-21
Inconsistent IDP extension constraint check

hwooley:
Hello, I found an inconsistent IDP extension constraint check per RFC5280 Section 5.2.5. I do not think the indirectCRL needs to be subject to that. Excerpt from the RFC:

-- at most one of onlyContainsUserCerts, onlyContainsCACerts,
-- and onlyContainsAttributeCerts may be set to TRUE.

Thank you

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -1988,10 +1988,12 @@ class IssuingDistributionPoint(ExtensionType):
                 "must all be boolean."
             )
 
+        # Per RFC5280 Section 5.2.5, the Issuing Distribution Point extension in a CRL
+        # can have only one of onlyContainsUserCerts, onlyContainsCACerts,
+        # onlyContainsAttributeCerts set to TRUE.
         crl_constraints = [
             only_contains_user_certs,
             only_contains_ca_certs,
-            indirect_crl,
             only_contains_attribute_certs,
         ]
 
@@ -1999,7 +2001,7 @@ class IssuingDistributionPoint(ExtensionType):
             raise ValueError(
                 "Only one of the following can be set to True: "
                 "only_contains_user_certs, only_contains_ca_certs, "
-                "indirect_crl, only_contains_attribute_certs"
+                "only_contains_attribute_certs"
             )
 
         if not any(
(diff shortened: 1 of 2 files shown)


alex:
Cc: @woodruffw this may be a CABF requirement

woodruffw:
> Cc: @woodruffw this may be a CABF requirement

Here's the relevant CABF section: <https://cabforum.org/working-groups/server/baseline-requirements/requirements/#7221-crl-issuing-distribution-point>

```
The indirectCRL and onlyContainsAttributeCerts fields MUST be set to FALSE (i.e., not asserted).

The CA MAY set either of the onlyContainsUserCerts and onlyContainsCACerts fields to TRUE, depending on the scope of the CRL.

The CA MUST NOT assert both of the onlyContainsUserCerts and onlyContainsCACerts fields.
```

So CABF is consistent with 5280 on this, it seems, insofar as `onlyContainsAttributeCerts` is always `FALSE` and the other two are always exclusive of each other.

(I think this has no effect on the validator, though, since this extension is for CRLs only and not CAs/leafs?)

hwooley:
Ah, I see... I think your comment about the having no effect on the validator may be true though I am not 100% sure. But, wouldn't this check mean that indirectCRL can never be set to TRUE in the CRL for it to be valid? I basically discovered this when loading the CRL that contains the IDP with indirectCRL, onlyContainsUserCerts set to TRUE.

reaperhulk:
Yes this appears to be a mistake, so if the tests get fixed here then this fix looks correct.

hwooley:
Hi @alex I can work on fixing the tests. Would you let me know how I can invoke the tests locally?

alex:
https://cryptography.io/en/latest/development/getting-started/

On Tue, Sep 24, 2024 at 12:49 AM Han Yu ***@***.***> wrote:
>
> Hi @alex I can work on fixing the tests. Would you let me know how I can invoke the tests locally?
>
> —



-- 

hwooley:
Thank you. I shorted the comment line and also removed the invalid test case. I used nox -e local to check the all test cases passed. Could you take a look? First time doing this, so let me know if I missed anything. Thank you.

alex:
Thanks for contributing!

hwooley:
You bet!

bh3000:
@alex Can you confirm if this change will be included in the 44.0.0 release and when that (or the release it will be in) is expected to be live?
Thanks!

reaperhulk:
This change will be in 44. We do not have a date, but it's likely it will be before the end of the year.

PR #11499, opened 2024-08-27
Mark that check-sdist is a Python 3.8+ only dependency

alex:
It has no versions that support Python 3.7. This is necessary to support using `uv` to manage our ci-constraints file


alex:
Blocked on https://github.com/python-cffi/cffi/issues/117

reaperhulk (approved):

PR #11500, opened 2024-08-27
Ensure build dependencies are pinned in wheel builder

alex:


alex:
https://github.com/pypa/pip/issues/12942

alex:
We're going to do this with uv. Will be a new PR, tracked in https://github.com/pyca/cryptography/issues/11548

PR #11524, opened 2024-09-02
argon2id support

reaperhulk:
Needs a rust-openssl release

### DIFF
--- a/docs/hazmat/primitives/key-derivation-functions.rst
+++ b/docs/hazmat/primitives/key-derivation-functions.rst
@@ -30,6 +30,105 @@ Different KDFs are suitable for different tasks such as:
 Variable cost algorithms
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
+Argon2id
+--------
+
+.. currentmodule:: cryptography.hazmat.primitives.kdf.argon2
+
+.. class:: Argon2id(*, salt, length, iterations, lanes, memory_cost, ad=None, secret=None)
+
+    .. versionadded:: 44.0.0
+
+    Argon2id is a KDF designed for password storage. It is designed to be
+    resistant to hardware attacks and is described in :rfc:`9106`.
+
+    This class conforms to the
+    :class:`~cryptography.hazmat.primitives.kdf.KeyDerivationFunction`
+    interface.
+
+    .. doctest::
+
+        >>> import os
+        >>> from cryptography.hazmat.primitives.kdf.argon2 import Argon2id
+        >>> salt = os.urandom(16)
+        >>> # derive
+        >>> kdf = Argon2id(
+        ...     salt=salt,
+        ...     length=32,
+        ...     iterations=1,
+        ...     lanes=4,
+        ...     memory_cost=64 * 1024,
+        ...     ad=None,
+        ...     secret=None,
+        ... )
+        >>> key = kdf.derive(b"my great password")
+        >>> # verify
+        >>> kdf = Argon2id(
+        ...     salt=salt,
+        ...     length=32,
+        ...     iterations=1,
+        ...     lanes=4,
+        ...     memory_cost=64 * 1024,
+        ...     ad=None,
+        ...     secret=None,
+        ... )
+        >>> kdf.verify(b"my great password", key)
+
+    **All arguments to the constructor are keyword-only.**
+
+    :param bytes salt: A salt.
+    :param int length: The desired length of the derived key in bytes.
+    :param int iterations: Also known as passes, this is used to tune
+        the running time independently of the memory size.
+    :param int lanes: The number of lanes (parallel threads) to use. Also
+        known as parallelism.
+    :param int memory_cost: The amount of memory to use in kibibytes.
+        1 kibibyte (KiB) is 1024 bytes.
+    :param bytes ad: Optional associated data.
+    :param bytes secret: Optional secret data.
+
+    :rfc:`9106` has recommendations for `parameter choice`_.
+
+    :raises cryptography.exceptions.UnsupportedAlgorithm: If Argon2id is not
+        supported by the OpenSSL version ``cryptography`` is using.
+
+    .. method:: derive(key_material)
+
+        :param key_material: The input key material.
+        :type key_material: :term:`bytes-like`
+        :return bytes: the derived key.
+        :raises TypeError: This exception is raised if ``key_material`` is not
+                           ``bytes``.
+        :raises cryptography.exceptions.AlreadyFinalized: This is raised when
+                                                          :meth:`derive` or
+                                                          :meth:`verify` is
+                                                          called more than
+                                                          once.
+
+        This generates and returns a new key from the supplied password.
+
+    .. method:: verify(key_material, expected_key)
+
+        :param bytes key_material: The input key material. This is the same as
+                                   ``key_material`` in :meth:`derive`.
+        :param bytes expected_key: The expected result of deriving a new key,
+                                   this is the same as the return value of
+                                   :meth:`derive`.
+        :raises cryptography.exceptions.InvalidKey: This is raised when the
+                                                    derived key does not match
+                                                    the expected key.
+        :raises cryptography.exceptions.AlreadyFinalized: This is raised when
+                                                          :meth:`derive` or
+                                                          :meth:`verify` is
+                                                          called more than
+                                                          once.
+
+        This checks whether deriving a new key from the supplied
+        ``key_material`` generates the same key as the ``expected_key``, and
+        raises an exception if they do not match. This can be used for
+        checking whether the password a user provides matches the stored derived
+        key.
+
 
 PBKDF2
 ------
@@ -1039,3 +1138,4 @@ Interface
 .. _`recommends`: https://datatracker.ietf.org/doc/html/rfc7914#section-2
 .. _`The scrypt paper`: https://www.tarsnap.com/scrypt/scrypt.pdf
 .. _`understanding HKDF`: https://soatok.blog/2021/11/17/understanding-hkdf/
+.. _`parameter choice`: https://datatracker.ietf.org/doc/html/rfc9106#section-4
--- a/docs/hazmat/primitives/key-derivation-functions.rst
+++ b/docs/hazmat/primitives/key-derivation-functions.rst
@@ -84,7 +84,7 @@ Argon2id
     :param int lanes: The number of lanes (parallel threads) to use. Also
         known as parallelism.
     :param int memory_cost: The amount of memory to use in kibibytes.
-        1 kibibyte (KiB) is 1024 bytes.
+        1 kibibyte (KiB) is 1024 bytes. This must be at minimum 8 * ``lanes``.
     :param bytes ad: Optional associated data.
     :param bytes secret: Optional secret data; used for keyed hashing.
 
--- a/docs/hazmat/primitives/key-derivation-functions.rst
+++ b/docs/hazmat/primitives/key-derivation-functions.rst
@@ -84,7 +84,7 @@ Argon2id
     :param int lanes: The number of lanes (parallel threads) to use. Also
         known as parallelism.
     :param int memory_cost: The amount of memory to use in kibibytes.
-        1 kibibyte (KiB) is 1024 bytes. This must be at minimum 8 * ``lanes``.
+        1 kibibyte (KiB) is 1024 bytes. This must be at minimum ``8 * lanes``.
     :param bytes ad: Optional associated data.
     :param bytes secret: Optional secret data; used for keyed hashing.
 
--- a/docs/hazmat/primitives/key-derivation-functions.rst
+++ b/docs/hazmat/primitives/key-derivation-functions.rst
@@ -76,7 +76,8 @@ Argon2id
 
     **All arguments to the constructor are keyword-only.**
 
-    :param bytes salt: A salt.
+    :param bytes salt: A salt should be unique (and randomly generated) per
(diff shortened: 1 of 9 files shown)


src/cryptography/hazmat/primitives/kdf/argon2.py
```diff
@@ -0,0 +1,86 @@
...
+    AlreadyFinalized,
+    InvalidKey,
+    UnsupportedAlgorithm,
+)
+from cryptography.hazmat.bindings._rust import openssl as rust_openssl
+from cryptography.hazmat.primitives import constant_time
+from cryptography.hazmat.primitives.kdf import KeyDerivationFunction
+
+
+class Argon2id(KeyDerivationFunction):
```
alex, line 18: Is there a reason we shouldn't just make this whole class be rust?

src/cryptography/hazmat/primitives/kdf/argon2.py
```diff
@@ -0,0 +1,86 @@
...
+        lanes: int,
+        memory_cost: int,
+        ad: bytes | None,
+        secret: bytes | None,
+    ):
+        from cryptography.hazmat.backends.openssl.backend import (
+            backend as ossl,
+        )
+
+        if ossl.openssl_version_number() < 0x30200000:
```
alex, line 33: Please use one of hte named properties so we can grep to find it. They all live on rust_openssl now. (Add a new one if needed, obviously)

tests/hazmat/primitives/test_argon2.py
```diff
@@ -0,0 +1,137 @@
...
+    only_if=lambda backend: not backend.openssl_version_number() >= 0x30200000,
+    skip_message="Supports argon2 so can't test unsupported path",
+)
+def test_unsupported_backend(backend):
+    with raises_unsupported_algorithm(None):
+        Argon2id(b"salt", None, None, 32, 1, 1, 32)
+
+
+@pytest.mark.supported(
+    only_if=lambda backend: backend.openssl_version_number() >= 0x30200000,
```
alex, line 34: same

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -30,6 +30,105 @@ Different KDFs are suitable for different tasks such as:
...
+        ...     lanes=4,
+        ...     memory_cost=64 * 1024,
+        ...     ad=None,
+        ...     secret=None,
+        ... )
+        >>> kdf.verify(b"my great password", key)
+
+    **All arguments to the constructor are keyword-only.**
+
+    :param bytes salt: A salt.
```
alex, line 79: Surely there's another few words that belong.

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -30,6 +30,105 @@ Different KDFs are suitable for different tasks such as:
...
+    :param int length: The desired length of the derived key in bytes.
+    :param int iterations: Also known as passes, this is used to tune
+        the running time independently of the memory size.
+    :param int lanes: The number of lanes (parallel threads) to use. Also
+        known as parallelism.
+    :param int memory_cost: The amount of memory to use in kibibytes.
+        1 kibibyte (KiB) is 1024 bytes.
+    :param bytes ad: Optional associated data.
+    :param bytes secret: Optional secret data.
```
alex, line 88: Same

src/cryptography/hazmat/primitives/kdf/argon2.py
alex: See https://github.com/pyca/cryptography/pull/11818

src/cryptography/hazmat/primitives/kdf/argon2.py
```diff
@@ -0,0 +1,87 @@
...
+        # OpenSSL requires a salt of at least 8 bytes
+        if len(salt) < 8:
+            raise ValueError("salt must be at least 8 bytes")
+        # Minimum length is 4 bytes as specified in RFC 9106
+        if not isinstance(length, int) or length < 4:
+            raise ValueError("length must be an integer greater >= 4")
+        if not isinstance(iterations, int) or iterations < 1:
+            raise ValueError("iterations must be an integer greater than 0")
+        if not isinstance(lanes, int) or lanes < 1:
+            raise ValueError("lanes must be an integer greater than 0")
```
alex, line 49: If they're not `int`, it shoudl really be a `TypeError`, if for no other reason than that's what would happen when we wrote it in Rust.

tests/hazmat/primitives/test_argon2.py
```diff
@@ -0,0 +1,163 @@
...
+        digest = argon2id.derive(b"password")
+        with pytest.raises(AlreadyFinalized):
+            argon2id.verify(b"password", digest)
+
+    def test_invalid_verify(self, backend):
+        argon2id = Argon2id(
+            salt=b"salt" * 2, length=32, iterations=1, lanes=1, memory_cost=32
+        )
+        with pytest.raises(InvalidKey):
+            argon2id.verify(b"password", b"invalidkey")
```
alex, line 148: There should probably be a testcase where the key is the right length, but wrong value

alex (commented):
Where'd we land on setting max threads for OpenSSL? Ignore for now?

src/rust/src/lib.rs
```diff
@@ -221,6 +221,17 @@ mod _rust {
...
             }
+            cfg_if::cfg_if! {
+                if #[cfg(CRYPTOGRAPHY_OPENSSL_320_OR_GREATER)] {
+                    use std::ptr;
+                    if let Ok(available) = std::thread::available_parallelism() {
+                        // SAFETY: This sets a libctx provider limit, but we always use the same libctx by passing NULL.
+                        unsafe {
+                            openssl_sys::OSSL_set_max_threads(ptr::null_mut(), available.get() as u64);
+                        }
+                    };
```
alex, line 232: This line is somehow uncovered. I think you don't need a semicolon

src/rust/src/lib.rs
```diff
@@ -221,6 +221,17 @@ mod _rust {
...
             }
+            cfg_if::cfg_if! {
+                if #[cfg(CRYPTOGRAPHY_OPENSSL_320_OR_GREATER)] {
+                    use std::ptr;
+                    if let Ok(available) = std::thread::available_parallelism() {
+                        // SAFETY: This sets a libctx provider limit, but we always use the same libctx by passing NULL.
+                        unsafe {
+                            openssl_sys::OSSL_set_max_threads(ptr::null_mut(), available.get() as u64);
+                        }
+                    }
```
alex, line 232: Sigh, this is the line that's uncovered, the else branch. But of course available_parallelism never fails in CI... not sure what the right thing to do is.

Rewrite the code somehow?

Also, what happens if someone else has called `OSSL_set_max_threads` (e.g., in theory, CPython?)

alex:
@reaperhulk I see that this is marked draft. Is it ready for review, or are there still TODOs?

reaperhulk:
This is now reviewable

reaperhulk:
That said, if there are no comments let’s not merge, I want a final pass myself since this has gone so long. I already see some doc improvements I could make

src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -122,6 +122,12 @@ def scrypt_supported(self) -> bool:
         else:
             return hasattr(rust_openssl.kdf.Scrypt, "derive")
 
+    def argon2_supported(self) -> bool:
+        if self._fips_enabled:
+            return False
```
alex, line 127: Any decade now...

src/rust/src/backend/kdf.rs
```diff
@@ -164,10 +164,178 @@ impl Scrypt {
...
+            } else {
+                if cryptography_openssl::fips::is_enabled() {
+                    return Err(CryptographyError::from(
+                        exceptions::UnsupportedAlgorithm::new_err(
+                            "This version of OpenSSL does not support argon2id"
+                        ),
+                    ));
+                }
+
+                if salt.extract::<&[u8]>(py)?.len() < 8 {
```
alex, line 227: Instead of `extract()`, use `as_bytes(py)` (big perf difference)

src/rust/src/backend/kdf.rs
```diff
@@ -164,10 +164,178 @@ impl Scrypt {
...
+                }
+                if lanes < 1 {
+                    return Err(CryptographyError::from(
+                        pyo3::exceptions::PyValueError::new_err(
+                            "lanes must be greater than or equal to 1."
+                        ),
+                    ));
+                }
+
+                if memory_cost < 8 * lanes {
```
alex, line 256: Ughhh, maybe rewrite to use `/ 8` instead of `8 *`, otherwise Very Large (tm) `lanes` would lead to integer overflow.

src/rust/src/lib.rs
```diff
@@ -221,6 +221,20 @@ mod _rust {
...
+                if #[cfg(CRYPTOGRAPHY_OPENSSL_320_OR_GREATER)] {
+                    use std::ptr;
+                    use std::cmp::max;
+
+                    let available = std::thread::available_parallelism().map_or(0, |v| v.get() as u64);
+                    // SAFETY: This sets a libctx provider limit, but we always use the same libctx by passing NULL.
+                    unsafe {
+                        let current = openssl_sys::OSSL_get_max_threads(ptr::null_mut());
+                        // Set the thread limit to the max of available parallelism or current limit.
+                        openssl_sys::OSSL_set_max_threads(ptr::null_mut(), max(available, current));
```
alex, line 234: Also, what happens if someone else has called `OSSL_set_max_threads` before us (e.g. CPython?)

reaperhulk: We'll call it again with the max of either the value they set or the max parallelism we detected (whatever is larger).

If the question is what happens if via a race condition we end up setting a max thread limit smaller than the current limit then `OSSL_set_max_threads` just updates a value in the `OSSL_LIB_CTX_THREADS` structure defined in `include/internal/thread.h`. The actual logic for the threadpool lives in `crypto/thread/internal.c`. If there's no active threads then changing it to a lower value is trivial, but if there are OpenSSL may have an issue? On initial review (without having tried to write a test to see if it actually does this):

* Set max threads to 10 at program start
* Spawn a thread (not in the OpenSSL threadpool) which calls argon2id with 40 lanes and enough iterations to take a bit of time.
* In main thread call `OSSL_set_max_threads` and set to 2
* Until the number of available threads drops to 2 or less any subsequent call to [_ossl_get_available_threads](https://github.com/openssl/openssl/blob/012353bdf21b98def920ac317b94c4a9ed501b79/crypto/thread/internal.c#L22) would return an underflowed value.

alex: Ok. I guess this is as good as can be done.

src/cryptography/hazmat/primitives/kdf/argon2.py
```diff
@@ -0,0 +1,11 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+from __future__ import annotations
+
+from cryptography.hazmat.bindings._rust import openssl as rust_openssl
+from cryptography.hazmat.primitives.kdf import KeyDerivationFunction
+
```
alex, line 9: Oops, we should add an `__all__` with just `"Argon2id"`

src/cryptography/hazmat/primitives/kdf/argon2.py
```diff
@@ -0,0 +1,11 @@
...
+Argon2id = rust_openssl.kdf.Argon2id
```
alex, line 10: Is this the righ capitalization?

alex (commented):
Almost done

src/cryptography/hazmat/primitives/kdf/argon2.py
reaperhulk: Yes, RFC 9106 also uses "Argon2id" in its capitalization.

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -30,6 +30,106 @@ Different KDFs are suitable for different tasks such as:
...
+
+    :param bytes salt: A salt should be unique (and randomly generated) per
+        password and is recommended to be 16 bytes or longer
+    :param int length: The desired length of the derived key in bytes.
+    :param int iterations: Also known as passes, this is used to tune
+        the running time independently of the memory size.
+    :param int lanes: The number of lanes (parallel threads) to use. Also
+        known as parallelism.
+    :param int memory_cost: The amount of memory to use in kibibytes.
+        1 kibibyte (KiB) is 1024 bytes. This must be at minimum 8 * ``lanes``.
```
alex, line 87:
```suggestion
        1 kibibyte (KiB) is 1024 bytes. This must be at minimum ``8 * lanes``.
```

alex (approved):

src/cryptography/hazmat/backends/openssl/backend.py
ofek: What is the issue with FIPS support? I'm just trying to understand what this comment means.

reaperhulk: NIST has been profoundly slow at adopting modern things. This includes any PBKDF that is better than PBKDF2. They asked for comments with a general intent to likely allow argon2id multiple years ago now, but there's still been no public forward progress.

Issue #6263, opened 2021-09-06
Add S/MIME decryption


BastianZim:
_Feature request_

It would be nice if `cryptography` could be used to decrypt S/MIME encrypted emails.

This could be low-level right now, similar to https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/?highlight=s%2Fmime#cryptography.hazmat.primitives.serialization.pkcs7.PKCS7SignatureBuilder or something more high level later. 
But since there is currently no stable python tool to decrypt S/MIME (That I know of), anything is already very much appreciated. :)

BastianZim:
I formulated this quite vague as I'm not sure how much info is needed but please let me know if I should add more info. Also, I saw some attempts at creating an issue to track all S/MIME-related features from a more high-level approach but from what I can see, that hasn't happened yet. Happy to create something, if required.

reaperhulk:
CMS is an enormous (and terrible) specification so you'll need to very precisely define what it is you need. This really means spelunking through RFCs and determining the minimal set of supported structures.

No one has stepped up to try to even define the scope of this work, much less determine how we would actually accomplish it. As for SMIME, no API has been proposed for encryption/decryption and no API has been proposed for verification so those remain items for future work. They're definitely more tractable than expanding to full CMS support, but carry a variety of challenges of their own.

None of this is impossible, but making progress will require consistent commitment since we need to agree on a design, discuss implementation, and then it needs to get written and go through what will likely be several cycles of significant feedback.

mij:
I agree with @reaperhulk . And an incremental approach would be:

1. Implement S/MIME signature generation
2. Implement S/MIME signature verification
3. Implement S/MIME encryption
4. Implement S/MIME decryption

I would speculate a few million emails are sent daily with S/MIME signature for each one sent with S/MIME encryption.
And I would also speculate that S/MIME signature generation limits the effort to <20% of the spec. All of the variability in crypto algorithms, is made irrelevant.

That's also what e-mail clients focus on.

alex:
As a status update, we now do signing and encryption, we don't have verification or decryption yet.

bluetech:
A use case for PKCS7 / CMS verify is [Apple Pay Payment Token](https://developer.apple.com/documentation/passkit_apple_pay_and_wallet/apple_pay/payment_token_format_reference).

[...]

Issue #10501, opened 2024-02-28
Signed S/MIME message has `CRLF` termination, whereas OpenSSL uses `LF`


facutuesca:
## Description

Signed S/MIME message using `cryptography`:
```python
b'MIME-Version: 1.0\r\nContent-Type: multipart/signed; protocol="application/x-pkcs7-signature"; micalg="sha-512"; boundary="===============3031221955182774630=="\r\n\r\nThis is an S/MIME signed message\r\n\r\n....'
```

Signed S/MIME message using `OpenSSL 3.2.0`:
```python
b'MIME-Version: 1.0\nContent-Type: multipart/signed; protocol="application/x-pkcs7-signature"; micalg="sha-256"; boundary="----E9313AF891A4913F5C5CCBBD0E0E0B80"\n\nThis is an S/MIME signed message\n\n...'
```

The `cryptography` output has `CRLF` line endings, while the OpenSSL output has `LF` line endings. 

## How to reproduce

Here is code to generate the outputs:

For `OpenSSL`:
```console
$ echo "message" > message.txt
$ openssl req -x509 -newkey rsa:4096 -keyout signer_key.pem -out signer.pem -sha256 -days 365 -nodes -subj "/C=US/ST=AAA/L=BBB/O=CCC/CN=www.example.com"

$ openssl smime -sign -in message.txt -text  -signer signer.pem -inkey signer_key.pem -out out.txt
```

For `cryptography`:
```python
from cryptography.hazmat.primitives.serialization import load_pem_private_key, pkcs7, Encoding
from cryptography.hazmat.primitives.asymmetric import padding
from cryptography.hazmat.primitives import hashes
from cryptography.x509 import load_pem_x509_certificate

with open('signer_key.pem', 'rb') as key_data:
    key = load_pem_private_key(key_data.read(), password=None)
with open('signer.pem', 'rb') as cert_data:
    cert = load_pem_x509_certificate(cert_data.read())

with open('message.txt', 'rb') as data:
    output = pkcs7.PKCS7SignatureBuilder().set_data(
      b'message'
    ).add_signer(
        cert, key, hashes.SHA512(), rsa_padding=padding.PKCS1v15()
    ).sign(
        Encoding.SMIME, [pkcs7.PKCS7Options.DetachedSignature, pkcs7.PKCS7Options.Text]
    )
print(output)
```

## Extra context
[...]

PR #11555, opened 2024-09-06
Add support for decrypting S/MIME messages

nitneuqr:
As discussed on #6263, I'm opening this PR with an initial implementation of S/MIME decryption, in order to better discuss the API design, the algorithms we want to support, and how we want to approach testing.

My essential thoughts for testing were to do the round-trip: encryption using the `PKCS7EnvelopeBuilder` and decryption using the `PKCS7EnvelopeDecryptor` (or whatever its name will be). No tests were made against openssl, even though it might interesting.

For the Python API, I've tried to stick as much as possible with the current API design (`PKCS7SignatureBuilder`, `PKCS7EnvelopeBuilder`).

I'm new to rust, so please let me know if you see some issues in variable lifetime, or some unnecessary copying between Python & Rust.

During development, I've realized that we could migrate PKCS7 unpadding to rust instead of using types, so I did migrate it. I'll probably open another smaller PR for that matter (if that makes sense?). I still have some code using PKCS7 Unpadding on Python if needed.

cc @alex


alex:
Thanks for working on this! I haven't had a chance to look in depth, but re: PKCS7Unpadder, it'd be great to have that in a stand-alone PR.

nitneuqr:
> Thanks for working on this! I haven't had a chance to look in depth, but re: PKCS7Unpadder, it'd be great to have that in a stand-alone PR.

Sure! Opened #11556 for that matter.

nitneuqr:
Rebased to main after integration of #11556. Should I split it again in smaller PRs? Maybe around `symmetric_decrypt` or other subfunctions in the code.

alex:
We're always happy to have smaller PRs split out (though it can sometimes be complicated due to coverage). I'm hoping to have time to review this today, though it might be tomorrow.

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -263,6 +263,116 @@ def encrypt(
         return rust_pkcs7.encrypt_and_serialize(self, encoding, options)
 
 
+class PKCS7EnvelopeDecryptor:
```
alex, line 266: Hmm, I wonder if we need a builder for this case, is `pkcs7_decrypt(data, recipient_cert, recipient_key, options)` sufficient, or might we want other things?

cc: @reaperhulk

alex:
@facutuesca FYI, if you're interested

src/cryptography/hazmat/primitives/serialization/pkcs7.py
nitneuqr: TBH on my side I was more focused on developing code that worked, so any API design is fine to me. I'd probably go for the unique function as it's the simplest one.

nitneuqr: Two ideas I'm thinking about right now:
- Should we allow multiple certs / pkeys as arguments and let `pkcs7_decrypt` choose the (or a) certificate that works? Outlook does that for instance (I've got multiple certificates associated with different mailboxes).
- In your case, you're missing the `encoding` parameter (DER, PEM, SMIME). Are we expecting `pkcs7_decrypt` to figure it out by itself?

reaperhulk: I'd bias to the function as well. I would not want the function to guess though -- we should either have a param or 3 separate functions.

alex: Ah yes, everywhere else in our API we have seperate functions for DER vs. PEM.

nitneuqr: Just pushed first implementation of `pkcs7_decrypt_{encoding}` 😄

nitneuqr:
@alex @reaperhulk took into account the first comments and adapted the tests / coverage accordingly.

Do you have any time for another review? Do you need anything else from me before reviewing? 😊

alex:
Sorry I got behind, will have a look now. Thanks for your patience!

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -263,6 +263,87 @@ def encrypt(
...
+
+def pkcs7_decrypt_pem(
+    data: bytes,
+    certificate: x509.Certificate,
+    private_key: rsa.RSAPrivateKey,
+    options: typing.Iterable[PKCS7Options],
+) -> bytes:
+    return _pkcs7_decrypt(
+        data, certificate, private_key, options, serialization.Encoding.PEM
+    )
```
alex, line 285: I don't love needing to pass an encoding down, ideally we'd be able to PEM decode the data and then simply call `pkcs7_decrypt_der` with the results. That's probably annoying to do in Python though.

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -328,6 +409,15 @@ def _smime_enveloped_encode(data: bytes) -> bytes:
     return m.as_bytes(policy=m.policy.clone(linesep="\n", max_line_length=0))
 
 
+def _smime_enveloped_decode(data: bytes) -> bytes:
+    m = email.message_from_bytes(data)
+    # Content type can be application/x-pkcs7-mime or application/pkcs7-mime
+    # Could also be using the filename parameter, or no checks at all.
+    if "pkcs7-mime" not in m.get_content_type():
```
alex, line 416: Let's explicitly check for those two, not allow anything that happens to have this substring.

src/rust/src/pkcs12.rs
```diff
@@ -107,6 +107,40 @@ pub(crate) fn symmetric_encrypt(
     Ok(ciphertext)
 }
 
+pub(crate) fn symmetric_decrypt(
```
alex, line 110: Probably this belongs in a utilities file, since its not actually used in pkcs12.

src/rust/src/pkcs7.rs
```diff
@@ -162,6 +166,116 @@ fn encrypt_and_serialize<'p>(
...
+) -> CryptographyResult<pyo3::Bound<'p, pyo3::types::PyBytes>> {
+    // Extract the data on the right format (PEM or DER, SMIME handled by Python)
+    let extracted_data = if encoding.is(&types::ENCODING_DER.get(py)?) {
+        data.as_bytes().to_vec()
+    } else {
+        let pem_str = std::str::from_utf8(data.as_bytes())
+            .map_err(|_| pyo3::exceptions::PyValueError::new_err("Invalid PEM data"))?;
+        let pem = pem::parse(pem_str)
+            .map_err(|_| pyo3::exceptions::PyValueError::new_err("Failed to parse PEM data"))?;
+        pem.contents().to_vec()
```
alex, line 186:
```suggestion
        pem.into_contents()
```

this avoids an allocation/copy

src/rust/src/pkcs7.rs
```diff
@@ -162,6 +166,116 @@ fn encrypt_and_serialize<'p>(
...
+    };
+
+    // Deserialize the content info
+    let content_info = asn1::parse_single::<pkcs7::ContentInfo<'_>>(&extracted_data).unwrap();
```
alex, line 190: @reaperhulk This would be our first time using our DER parser for PKCS7. On the one hand: new API, so there's no backwards-compat concerns. On the other hand, _surely_ someone is emitting awful BER PKCS#7. Any concerns?

src/rust/src/pkcs7.rs
```diff
@@ -162,6 +166,116 @@ fn encrypt_and_serialize<'p>(
...
+        pkcs7::Content::EnvelopedData(data) => {
+            // Extract enveloped data
+            let enveloped_data = data.into_inner();
+
+            // Get recipients, and the one matching with the given certificate (if any)
+            let mut recipient_infos = enveloped_data.recipient_infos.unwrap_read().clone();
+            let recipient_serial_number = certificate.get().raw.borrow_dependent().tbs_cert.serial;
+            let found_recipient_info = recipient_infos.find(|info| {
+                info.issuer_and_serial_number.serial_number.as_bytes()
+                    == recipient_serial_number.as_bytes()
```
alex, line 201: You should just be able to do `==` without the `as_bytes`

src/rust/src/test_support.rs
```diff
@@ -103,6 +103,57 @@ fn pkcs7_verify(
     Ok(())
 }
 
+#[cfg(not(CRYPTOGRAPHY_IS_BORINGSSL))]
```
alex, line 106: What is this needed for? We have public APIs for PKCS#7 encryption

alex (commented):
Did an initial review of the code -- I haven't really sat down and reviewed the core PKCS#7 logic against hte RFC yet though.

src/cryptography/hazmat/primitives/serialization/pkcs7.py
nitneuqr: A few solutions:
- Separating the PEM decode logic in another rust function, and calling it separately
- Finding a library that does PEM to DER in Python
- Coding it in-house (but probably annoying and / or dirty)

I'm for the 1st one, I think.

src/rust/src/pkcs7.rs
nitneuqr: Done!

src/rust/src/test_support.rs
nitneuqr: I coded this for testing purposes: to encode using another algorithm, so that we can check the algorithm is unsupported (for coverage, mostly).

For most of the tests, I can replace the `test_support.pkcs7_encrypt` with `PKCS7EnvelopeBuilder`, but not for `test_smime_decrypt_algorithm_not_supported`.

Another solution can be to store an AES-256-CBC encoded PKCS7, and read it during the tests, if you prefer 😄

src/cryptography/hazmat/primitives/serialization/pkcs7.py
nitneuqr: Done!

src/rust/src/pkcs12.rs
nitneuqr: OK, will see where I can put this and let you know.

src/cryptography/hazmat/primitives/serialization/pkcs7.py
nitneuqr: I've implemented a quick function in python using rust to decode pem to der. Please let me know what you think!

src/rust/src/pkcs7.rs
```diff
@@ -162,6 +166,116 @@ fn encrypt_and_serialize<'p>(
...
+        pkcs7::Content::EnvelopedData(data) => {
+            // Extract enveloped data
+            let enveloped_data = data.into_inner();
+
+            // Get recipients, and the one matching with the given certificate (if any)
+            let mut recipient_infos = enveloped_data.recipient_infos.unwrap_read().clone();
+            let recipient_serial_number = certificate.get().raw.borrow_dependent().tbs_cert.serial;
+            let found_recipient_info = recipient_infos.find(|info| {
+                info.issuer_and_serial_number.serial_number == recipient_serial_number
+            });
```
alex, line 202: We probalby need to check that hte issuers match as well, right?

src/rust/src/pkcs7.rs
```diff
@@ -162,6 +166,116 @@ fn encrypt_and_serialize<'p>(
...
+    };
+
+    // Return content in different form, based on options
+    let text_mode = options.contains(types::PKCS7_TEXT.get(py)?)?;
+    let plain_data = if options.contains(types::PKCS7_BINARY.get(py)?)? {
+        plain_content
+    } else {
+        let decanonicalized = smime_decanonicalize(plain_content.as_bytes(), text_mode);
+        pyo3::types::PyBytes::new_bound(py, decanonicalized.into_owned().as_slice())
+    };
```
alex, line 274: Note to self: Think about this more.

src/rust/src/test_support.rs
alex: For test cases, is there a reason not to pre-generate them and then just check them into the vectors folder?

vectors/cryptography_vectors/pkcs7/rsa_ca.pem
```diff
@@ -0,0 +1,19 @@
+-----BEGIN CERTIFICATE-----
```
alex, line 1: Anything in the vectors/ dir needs to be documented in `test-vectors.rst`. (You can also split that out into a seperate PR so we can merge it easily).

alex (commented):
@reaperhulk would still like your views on deserialization for PKCS#7 in Rust.

src/rust/src/test_support.rs
nitneuqr: No, there's none, i can definitely do it! 

Also, does that mean I should get rid of the `test_support.pkcs7_decrypt` function and use the newly developed one?

vectors/cryptography_vectors/pkcs7/rsa_ca.pem
nitneuqr: OK, will do!

src/rust/src/pkcs7.rs
nitneuqr: I'm not familiar with x509 certificates; what are the necessary attributes to check to verify certificate unicity?

src/rust/src/test_support.rs
alex: If we pre-generate the test cases, we actually don't need the generator code checked in at all.

src/rust/src/pkcs7.rs
alex: `issuer_and_serial_number` and has an issuer attribute, and a serial attribute. We need to check both here.

src/rust/src/test_support.rs
nitneuqr: The new commits should remove the `test_support.pkcs_encyrpt` function and replace its usage with new vectors (for which I'll do a separate PR today).

I'm talking about the previous tests for the `PKCS7EnvelopeBuilder`, that are using `test_support.pkcs7_decrypt`. I assume I should replace this function with the new public API we're developing, `pkcs7_decrypt_{encoding}`?

alex: Ahhhh, sorry I misunderstood. Yes, I think replacing those with the new public API would be good.

nitneuqr: Will do today too :)

nitneuqr: Done!

vectors/cryptography_vectors/pkcs7/rsa_ca.pem
nitneuqr: Opened #11843 for the matter 😄

src/rust/src/pkcs7.rs
nitneuqr: Added code to check issuer :)

alex:
(Thanks for updating, I'm going to pause on reviewing until https://github.com/pyca/cryptography/pull/11843 is merged, hopefully that'll be quick.)

nitneuqr:
#11843 is merged, I've rebased the branch onto main to take the new changes into account. It should b ready to review again!

alex:
Great, will try to get to it this evening. Thanks!

On Wed, Nov 6, 2024 at 3:37 AM Quentin Retourne
***@***.***> wrote:
>
> #11843 is merged, I've rebased the branch onto main to take the new changes into account. It should b ready to review again!
>
> —



-- 

src/rust/cryptography-x509/src/pkcs7.rs
```diff
@@ -29,22 +29,38 @@ pub enum Content<'a> {
     EncryptedData(asn1::Explicit<EncryptedData<'a>, 0>),
 }
 
-#[derive(asn1::Asn1Write)]
+#[derive(asn1::Asn1Write, asn1::Asn1Read)]
 pub struct SignedData<'a> {
     pub version: u8,
-    pub digest_algorithms: asn1::SetOfWriter<'a, common::AlgorithmIdentifier<'a>>,
+    pub digest_algorithms: common::Asn1ReadableOrWritable<
```
alex, line 35: Would you be up for pulling the `Asn1ReadableOrWritable` pieces out into a separate PR? I think this is very very close, and I'm looking for ways to keep it moving as I review the last pieces.

src/rust/src/pkcs7.rs
alex: @reaperhulk just going to keep pinging you on this :-) (Enjoy your vacation)

reaperhulk: Let's assume we release with DER only parsing. In that scenario we're attempting to empirically determine whether or not there's enough BER out there to matter. If no one comes to complain then great, job done. However, if we discover we're wrong and BER is relevant, what is our plan? Are we going to add BER support in rust?

alex: If we do determine that this API is basically low utility for many users because their S/MIME messages aren't valid DER, it seems we have a few options:

1) Declare that's how it goes. This is a net new API, so no users are worse off than they were before. May result in greater timelines for adoption of this, and migration away from other libraries.

2) Re-implement in terms of OpenSSL S/MIME APIs. Probably straightforward, although as-ever, assessing the compatibility surface will be a PITA.

3) Add BER support to rust-asn1. No fucking chance, not worth discussing.

4) We force ourselves to do a survey of the ecosystem before landing this so we can contact producers of non-DER S/MIMEs.


I have a soft spot for 1, I could be persuaded of 4, and if we really need to do it, I guess 2 is the way to go.

WDYT?

alex: Go stdlib doesn't support PKCS#7 I don't think. I did find:

- https://github.com/mozilla-services/pkcs7/blob/v0.9.0/pkcs7.go#L155-L160
- https://github.com/InfiniteLoopSpace/go_S-MIME/blob/master/cms/protocol/contentinfo.go#L25


Sooooo

src/rust/cryptography-x509/src/pkcs7.rs
nitneuqr: Don, #11922 is opened!

src/rust/src/pkcs12.rs
nitneuqr: Just checked this conversation was resolved, but we didn't agree on anything. Moved the file to `pkcs7.rs` for now, it makes more sense :)

alex: I resolved it because it didn't feel like a blocker, but that works!

nitneuqr:
Hey, any news on this? Do you need anything else from me? 😊

src/rust/src/pkcs7.rs
reaperhulk: Okay, I'm comfortable with doing this DER only and finding out how it goes. It's net-new API surface so we're not regressing and I'd love to find out BER doesn't matter.

alex:
Ooops, sorry this fell off my radar. I will get to it today!

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -263,6 +263,81 @@ def encrypt(
         return rust_pkcs7.encrypt_and_serialize(self, encoding, options)
 
 
+def pkcs7_decrypt_der(
```
reaperhulk, line 266: I think we should move all these functions to rust and export them here.

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -263,6 +263,81 @@ def encrypt(
...
+        )
+
+    options = list(options)
+    if not all(isinstance(x, PKCS7Options) for x in options):
+        raise ValueError("options must be from the PKCS7Options enum")
+    if any(
+        opt not in [PKCS7Options.Text, PKCS7Options.Binary] for opt in options
+    ):
+        raise ValueError(
+            "Only the following options are supported for encryption: "
```
alex, line 319:
```suggestion
            "Only the following options are supported for decryption: "
```

src/rust/src/pkcs7.rs
```diff
@@ -162,6 +168,153 @@ fn encrypt_and_serialize<'p>(
...
+                .unwrap();
+            let decrypted_content = symmetric_decrypt(py, algorithm, mode, encrypted_content)?;
+            pyo3::types::PyBytes::new_bound(py, decrypted_content.as_slice())
+        }
+        _ => {
+            return Err(CryptographyError::from(
+                exceptions::UnsupportedAlgorithm::new_err((
+                    "Only EnvelopedData structures are currently supported.",
+                    exceptions::Reasons::UNSUPPORTED_SERIALIZATION,
+                )),
```
alex, line 267: Is this unsupported, or just a ValueError, because the data is in the wrong format?

src/rust/src/pkcs7.rs
```diff
@@ -162,6 +168,153 @@ fn encrypt_and_serialize<'p>(
...
+                    && info.issuer_and_serial_number.issuer == recipient_issuer
+            });
+
+            // Raise error when no recipient is found
+            // Unsure if this is the right exception to raise
+            let recipient_info = match found_recipient_info {
+                Some(info) => info,
+                None => {
+                    return Err(CryptographyError::from(
+                        exceptions::AttributeNotFound::new_err((
```
alex, line 215: This isn't really the correct exception type here, I think a regular value error is fine.

src/rust/src/pkcs7.rs
```diff
@@ -162,6 +168,153 @@ fn encrypt_and_serialize<'p>(
...
+                            "No recipient found that matches the given certificate.",
+                            exceptions::Reasons::UNSUPPORTED_X509,
+                        )),
+                    ));
+                }
+            };
+
+            // Decrypt the key using the private key
```
alex, line 223: Before this, you need to check that `key_encryption_algorithm` is the expected type.

src/rust/src/pkcs7.rs
```diff
@@ -407,6 +566,39 @@ fn smime_canonicalize(data: &[u8], text_mode: bool) -> (Cow<'_, [u8]>, Cow<'_, [
     }
 }
 
+fn smime_decanonicalize(data: &[u8], text_mode: bool) -> Cow<'_, [u8]> {
```
alex, line 569: @reaperhulk can I bug you to review this, I feel like I could use a second pair of eyes.

src/rust/src/pkcs7.rs
```diff
@@ -407,6 +566,39 @@ fn smime_canonicalize(data: &[u8], text_mode: bool) -> (Cow<'_, [u8]>, Cow<'_, [
...
+    let mut new_data = vec![];
+    let mut last_idx = 0;
+
+    // Remove the header if text_mode is true
+    let data = if text_mode {
+        let header = b"Content-Type: text/plain\r\n\r\n";
+        if data.starts_with(header) {
+            &data[header.len()..]
+        } else {
```
alex, line 578: If the header isn't present in text mode, should that be an error?

src/cryptography/hazmat/primitives/serialization/pkcs7.py
nitneuqr: Which also means, passing the smime-related function in Python as private, and calling it on Rust, I guess?

src/rust/src/pkcs7.rs
nitneuqr: As far as I understand from S/MIME 3.2 [RFC](https://datatracker.ietf.org/doc/html/rfc5751), it seems that only enveloped data can be decrypted. So I'm passing this as a ValueError then. I guess I initially thought other formats could be decrypted.

src/rust/src/pkcs7.rs
nitneuqr: I'm fine with it, but is it really possible that it is something else?

alex: The only way it could happen is a bug, or an attempt to exploit something. We just want to be conservative in making sure we strictly enforce things.

src/rust/src/pkcs7.rs
nitneuqr: I have no idea how OpenSSL handles it for now. I'll check it out when I have time and let you know.

src/cryptography/hazmat/primitives/serialization/pkcs7.py
alex: If you look at `pkcs7.rs` and search for `SMIME_ENVELOPED_ENCODE` you can see how we call into the S/MIME Python from Rust.

vectors/cryptography_vectors/pkcs7/enveloped-rsa-oaep.pem
```diff
@@ -0,0 +1,16 @@
+-----BEGIN PKCS7-----
```
alex, line 1: If you wouldn't mind a separate PR with this vector, we can get it merged ASAP.

alex (commented):
I think this is super super close. Thanks for all your hard work on it!

vectors/cryptography_vectors/pkcs7/enveloped-rsa-oaep.pem
nitneuqr: #11998

src/rust/src/pkcs7.rs
nitneuqr: After some checking with my version of OpenSSL (on Windows!), it seems the behavior of the `-text` and `-binary` options are much weirder than what was expected on `decrypt.` They are not symmetrical to their usage in `encrypt` :
- No options passed : replaces the \n into \r\n
- `-binary` : leaves the \n as such, do not replace them by \r\n
- `-text` :  Removes the header, raises an error if there is none (your comment is correct)
- Both options can be combined, where in `encrypt`, Binary was taking precedence.

This implies that if you encrypt then decrypt data using `openssl smime` and the `text` option, you can start with `Hello, world!\n` and end up with `Hello, world!\r\r\n` 😢 

I was mostly using `-binary` optons in my use cases, so I never quite found out. Adding this right now.

facutuesca: @nitneuqr this looks similar to a bug I encountered while implementing s/mime signing: https://github.com/openssl/openssl/issues/23886. There is also a comment on that issue from someone finding the same bug during encryption

nitneuqr: Thanks for that. To me, the documentation for the `-binary` option makes sense, at least during the `-encrypt` part. But for decryption, it doesn't. The comment exactly describes my issue 😄

> -binary
> 
> Normally the input message is converted to "canonical" format which is effectively using CR and LF as end of line: as required by the S/MIME specification. When this option is present no translation occurs. This is useful when handling binary data which may not be in MIME format.

To be honest, for this PR I'd stick with the current behavior of OpenSSL (even if seems incorrect). What's your opinion on this?

facutuesca: > To be honest, for this PR I'd stick with the current behavior of OpenSSL (even if seems incorrect). What's your opinion on this?

I think it's up to Alex and Paul. For context, the decision for signing [was to follow the RFC](https://github.com/pyca/cryptography/issues/10501#issuecomment-2212032777), rather than imitate what OpenSSL does.

nitneuqr: I think we're confusing the format of the output file (which is what is supposed to control `-crlfeol` in OpenSSL) with the format of the ciphered message inside it (which is controlled by `-text` or `-binary`), at least for encryption/decryption.

Right now, the `-crlfeol` parameter does not seem to have any effect in decryption, so I'm ignoring it for now.

nitneuqr:
Hey, we're getting closer now, but this conversation just raised new issue:

https://github.com/pyca/cryptography/pull/11555#discussion_r1845593139

Using the `openssl` executable, a round-trip (encryption + decryption) seems to transform `Hello, world!\n` into `Hello, world!\r\r\n`. This issue did not seem to happen in the old `test_support.pkcs7_decrypt` function (using openssl on Rust): the round-trip seemed to work fine. Is it correct that we should stick to this behavior instead of the other one?

In any cases, I'm going to add more tests cases when some data is encrypted with some options, and decrypted with some others.

alex:
Are you sure the old version round tripped in that way? If you look at the tests, it looks to me like this PR changes them -- they used to have things like:

```py
        # New lines are canonicalized to '\r\n' when not using Binary
        expected_data = (
            data
            if pkcs7.PKCS7Options.Binary in options
            else data.replace(b"\n", b"\r\n")
        )
        assert decrypted_bytes == expected_data
```

and now they have

```py
        assert decrypted_bytes == data
```

nitneuqr:
> That's correct. I'll check this out. It might be yet another way, as it seems only one "\r" is added.

Should be good now. `decanonicalize` function was way too complicated, `-binary` option does not seem to be handled by `openssl smime -decrypt`. I should have compared with `test_support` function way sooner, and should not have changed the previous tests assertions 😞 I'll move what's possible to `pkcs7.rs` and let you know when you can do a (hopefully!) final review.

src/cryptography/hazmat/primitives/serialization/pkcs7.py
nitneuqr: Done! One last question: should I also move the logic of the new _check_pkcs7_decrypt_arguments function to rust, or just leave it as such and call it the same way I'm calling SMIME_ENVELOPED_DECODE?

nitneuqr:
@alex @reaperhulk took most comments into account, still have some minor questions but would love a (hopefully!) final review.

src/cryptography/hazmat/primitives/serialization/pkcs7.py
alex: Yes, let's move that as well -- it actually will probably become a fair bit easier, as many things are automatically handled by giving parameters types.

nitneuqr: Should be done. I'd love a quick review on the `check_decrypt_parameters` function, it is probably more intricate than it should be!

src/rust/src/pkcs7.rs
```diff
@@ -164,6 +166,257 @@ fn encrypt_and_serialize<'p>(
...
 
+#[pyo3::pyfunction]
+fn decrypt_der<'p>(
+    py: pyo3::Python<'p>,
+    data: CffiBuf<'p>,
+    certificate: pyo3::Bound<'p, x509::certificate::Certificate>,
+    private_key: pyo3::Bound<'p, backend::rsa::RsaPrivateKey>,
+    options: &pyo3::Bound<'p, pyo3::types::PyList>,
+) -> CryptographyResult<pyo3::Bound<'p, pyo3::types::PyBytes>> {
+    decrypt(py, data.as_bytes(), certificate, private_key, options)
```
alex, line 177: Nit, isntead of having a decrypt function that everything calls, just have the others call `decrypt_der`.

I'm not sure there's a reason for `data` to be `CffiBuf`, you can just make it a `&[u8]` directly.

src/rust/src/pkcs7.rs
```diff
@@ -164,6 +166,257 @@ fn encrypt_and_serialize<'p>(
...
+    py: pyo3::Python<'p>,
+    data: CffiBuf<'p>,
+    certificate: pyo3::Bound<'p, x509::certificate::Certificate>,
+    private_key: pyo3::Bound<'p, backend::rsa::RsaPrivateKey>,
+    options: &pyo3::Bound<'p, pyo3::types::PyList>,
+) -> CryptographyResult<pyo3::Bound<'p, pyo3::types::PyBytes>> {
+    let pem_str = std::str::from_utf8(data.as_bytes())
+        .map_err(|_| pyo3::exceptions::PyValueError::new_err("Invalid PEM data"))?;
+    let pem = pem::parse(pem_str)
+        .map_err(|_| pyo3::exceptions::PyValueError::new_err("Failed to parse PEM data"))?;
```
alex, line 191: Doh, should have caught this earlier: we need to make sure the PEM has the right tag! We don't want to allow `----- BEGIN CERTIFICATE -----` :-)

src/rust/src/pkcs7.rs
```diff
@@ -164,6 +166,257 @@ fn encrypt_and_serialize<'p>(
     }
 }
 
+#[pyo3::pyfunction]
+fn decrypt_der<'p>(
+    py: pyo3::Python<'p>,
+    data: CffiBuf<'p>,
+    certificate: pyo3::Bound<'p, x509::certificate::Certificate>,
+    private_key: pyo3::Bound<'p, backend::rsa::RsaPrivateKey>,
```
alex, line 174: Sadly this can't take `backend::rsa::RsaPrivateKey`, we need to make sure we allow anything that implements the interface (including other Python classes), so this needs to be a `Bound<'p, PyAny>`

src/rust/src/pkcs7.rs
```diff
@@ -164,6 +166,257 @@ fn encrypt_and_serialize<'p>(
...
+            ));
+        }
+    };
+
+    // If text_mode, strip the header if possible, else return an error
+    let plain_data = plain_content.as_bytes();
+    let plain_data = if options.contains(types::PKCS7_TEXT.get(py)?)? {
+        let header = b"Content-Type: text/plain\r\n\r\n";
+        if plain_data.starts_with(header) {
+            &plain_data[header.len()..]
```
alex, line 316: You can use https://doc.rust-lang.org/stable/std/primitive.slice.html#method.strip_prefix to combine these into one step

src/rust/src/pkcs7.rs
```diff
@@ -164,6 +166,257 @@ fn encrypt_and_serialize<'p>(
...
+        _ => {
+            return Err(CryptographyError::from(
+                pyo3::exceptions::PyValueError::new_err(
+                    "The PKCS7 data is not an EnvelopedData structure.",
+                ),
+            ));
+        }
+    };
+
+    // If text_mode, strip the header if possible, else return an error
```
alex, line 311: @reaperhulk can I get your opinion on the right behavior for text vs. binary?

src/rust/src/pkcs7.rs
```diff
@@ -164,6 +166,257 @@ fn encrypt_and_serialize<'p>(
...
+                exceptions::Reasons::UNSUPPORTED_PADDING,
+            )),
+        ));
+    }
+
+    // Check if all options are from the PKCS7Options enum
+    let pkcs7_option_type = types::PKCS7_TEXT.get(py)?.get_type();
+    if !options
+        .iter()
+        .all(|opt| opt.get_type().eq(pkcs7_option_type.clone()).unwrap())
```
alex, line 351: You can use https://docs.rs/pyo3/latest/pyo3/prelude/trait.PyAnyMethods.html#tymethod.is_instance which is a bit nicer than `get_type().eq()`

src/rust/src/pkcs7.rs
```diff
@@ -164,6 +166,257 @@ fn encrypt_and_serialize<'p>(
...
+            pyo3::exceptions::PyValueError::new_err(
+                "Only the following options are supported for decryption: Text",
+            ),
+        ));
+    }
+
+    // Check if certificate's public key is an RSA public key
+    if !certificate
+        .call_method0(pyo3::intern!(py, "public_key"))?
+        .is_instance_of::<backend::rsa::RsaPublicKey>()
```
alex, line 374: Similar note to above, you need to use `types::RSA_PUBLIC_KEY` here.

alex (commented):
A few small nits, but this has come together great!

I want @reaperhulk's opinion on the text vs. binary stuff, since IIRC he did the original impl for the encryption side.

src/rust/src/pkcs7.rs
nitneuqr: OK! Does this mean the same for `pyo3::Bound<'p, x509::certificate::Certificate>` or not?

src/rust/src/pkcs7.rs
nitneuqr: Done! Thought `CffiBuf` was important ^^

src/rust/src/pkcs7.rs
nitneuqr: Done! Added test accordingly.

src/rust/src/pkcs7.rs
nitneuqr: Done! Indeed, seems more "rustic" 😄

src/rust/src/pkcs7.rs
nitneuqr: Done!

src/rust/src/pkcs7.rs
nitneuqr: Done!

src/rust/src/pkcs7.rs
alex: Certificate is ok as-is.

nitneuqr: Perfect! I've added checks for RSA keys then. Thanks a lot!

src/rust/src/pkcs7.rs
reaperhulk: We added these flags to behave like OpenSSL (sigh).

OpenSSL calls `SMIME_text` in its decryption path when you pass `PKCS7_TEXT`. That function errors for a few conditions, but most notably if `parsing the MIME headers fails, if there is no “content-type” header, or if the content type is not “text/plain”.`

This code appears to make two assumptions that may not match OpenSSL's though:
* Case sensitivity on `Content-Type`
* That the Content-Type header is guaranteed to be the last header before the payload.

The behavior I believe we want is to parse the MIME headers, check that content-type is text/plain, and then discard them entirely and only keep the payload.

src/rust/src/pkcs7.rs
```diff
@@ -164,6 +166,267 @@ fn encrypt_and_serialize<'p>(
...
+    data: &[u8],
+    certificate: pyo3::Bound<'p, x509::certificate::Certificate>,
+    private_key: pyo3::Bound<'p, pyo3::types::PyAny>,
+    options: &pyo3::Bound<'p, pyo3::types::PyList>,
+) -> CryptographyResult<pyo3::Bound<'p, pyo3::types::PyBytes>> {
+    // Check the decrypt parameters
+    check_decrypt_parameters(py, &certificate, &private_key, options)?;
+
+    // Decrypt the data
+    let content_info = asn1::parse_single::<pkcs7::ContentInfo<'_>>(data).unwrap();
```
alex, line 221:
```suggestion
    let content_info = asn1::parse_single::<pkcs7::ContentInfo<'_>>(data)?;
```

src/rust/src/pkcs7.rs
```diff
@@ -164,6 +166,267 @@ fn encrypt_and_serialize<'p>(
...
+        return Err(CryptographyError::from(
+            pyo3::exceptions::PyValueError::new_err("options must be from the PKCS7Options enum"),
+        ));
+    }
+
+    // Check if any option is not PKCS7Options::Text
+    let text_option = types::PKCS7_TEXT.get(py)?;
+    if options
+        .iter()
+        .any(|opt| !opt.eq(text_option.clone()).unwrap())
```
alex, line 361: These unwraps need to propagate the error up (which annoyingly probably means you can't use any)

src/rust/src/pkcs7.rs
```diff
@@ -164,6 +166,267 @@ fn encrypt_and_serialize<'p>(
...
+                        )),
+                    ));
+                }
+            };
+
+            // Decrypt the content using the key and proper algorithm
+            let encrypted_content = enveloped_data
+                .encrypted_content_info
+                .encrypted_content
+                .unwrap();
```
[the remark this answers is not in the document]
nitneuqr: Will do! Having coverage for these tests cases is not an easy task.

src/rust/src/pkcs7.rs
nitneuqr: Will do! Didn't know the issues with `unwrap()`. Just Copilot-ed the difference between `?` and `unwrap()`.

nitneuqr: Is an `unwrap_or(false)` acceptable? Or, more precisely, why isn't it a good idea?

alex: No, we don't want `unwrap_or(false)` here -- if an `Err(_)` is returned, that means an exception was raised on `eq()` somehow, so we want to propagate that up the stack, using `unwrap_or()` would just treat any exception as being `false`.

src/rust/src/pkcs7.rs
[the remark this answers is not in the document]
alex: Yes, it can be frustrating, we definitely end up with way more test cases than most folks as a result.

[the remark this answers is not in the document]
nitneuqr: Done!

src/rust/src/pkcs7.rs
nitneuqr: So, I've been trying to raise an error on __eq__, which is actually not that simple; but manage to do so, will push something in the evening (Europe time). Can you give me more details as to why an exception of this type is not acceptable? Because this is what is raised right now.

`pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: PyErr { type: <class 'ValueError'>, value: ValueError("I'm broken"), traceback: Some(<traceback object at 0x000002C4B44BE8C0>) }`

alex: `PanicExceptions` are always considered bugs.

FWIW in this case I believe you should be able to use `options.contains(&text_option)?` to get what you want.

src/rust/src/pkcs7.rs
```diff
@@ -164,6 +166,273 @@ fn encrypt_and_serialize<'p>(
...
+    // Check if all options are from the PKCS7Options enum
+    let pkcs7_option_type = types::PKCS7_TEXT.get(py)?.get_type();
+    if !options
+        .iter()
+        .all(|opt| opt.is_instance(&pkcs7_option_type).unwrap())
+    {
+        return Err(CryptographyError::from(
+            pyo3::exceptions::PyValueError::new_err("options must be from the PKCS7Options enum"),
+        ));
+    }
```
alex, line 361: And in this case you can unroll the loop:

```
for opt in options.iter() {
    if !opt.is_instance(&pkcs7_option_type)? {
        return Err(...);
    }
}

File: src/rust/src/pkcs7.rs (hunk 29 above)
Comment (nitneuqr) [reply src/rust/src/pkcs7.rs:367]: That's correct, but doing so, we are not defensively failing if someone passes another PKCS #7
Option to `pkcs7_decrypt`, even if it has no impact. Are we fine with this?

File: src/rust/src/pkcs7.rs (hunk 31 above)
Comment (nitneuqr) [reply src/rust/src/pkcs7.rs:361]: Thanks! I'll go with this

File: src/rust/src/pkcs7.rs (hunk 29 above)
Comment (alex) [reply src/rust/src/pkcs7.rs:367]: Ah, whoops, I inverted the logic. The same approach to writing a full loop, instead of the `any()`, that I suggested above should work here.

Comment (nitneuqr) [reply src/rust/src/pkcs7.rs:367]: Yup! Just saw your other comment, works perfectly!

File: src/rust/src/pkcs7.rs (hunk 25 above)
Comment (nitneuqr) [reply src/rust/src/pkcs7.rs:315]: Do you have any suggestion of a lib to handle MIME efficiently in rust? For now, I'm thinking about
using some Python function. Seems simpler to me.

Comment (alex) [reply src/rust/src/pkcs7.rs:315]: I would just use Python for this.

Comment (nitneuqr) [reply src/rust/src/pkcs7.rs]: First version pushed, will compare with `openssl` and let you know!

And yeah, it should be up to the user to sanitize the decrypted output after `pkcs7_decrypt`, especially with the libraries available on Python that are perfect for this ...

Comment (nitneuqr) [reply src/rust/src/pkcs7.rs]: I've tested against some custom headers examples, behavior is the same as `openssl` (using `test_support.pkcs7_decrypt` from `main`).

**Comment by nitneuqr** [comment]
> Besides these unwrap comments and @reaperhulk's on text, I think this is ready. Thanks so much for your continued work on it.

Thanks for your reviews! Should be good now, header removal is now tested in `test_pkcs7_decrypt_der_text_handmade_header`, and `unwrap`s have been removed.

File: src/cryptography/hazmat/primitives/serialization/pkcs7.py [hunk 32]
```diff
@@ -328,6 +333,33 @@ def _smime_enveloped_encode(data: bytes) -> bytes:
...
+        "application/pkcs7-mime",
+    }:
+        raise ValueError("Not an S/MIME enveloped message")
+    return bytes(m.get_payload(decode=True))
+
+
+def _smime_remove_text_headers(data: bytes) -> bytes:
+    m = email.message_from_bytes(data)
+    headers = {key.title(): value for key, value in m.items()}
+    if "Content-Type" not in headers:
```
reaperhulk, line 349: Could you add a comment explaining that `message_from_byes` handles case insensitivity for headers? I didn't realize it did that so reminding me in the future would be good 😂

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -328,6 +333,33 @@ def _smime_enveloped_encode(data: bytes) -> bytes:
...
+        raise ValueError(
+            "Decrypted MIME data has no 'Content-Type' header. "
+            "Please remove the 'Text' option to visualize it."
```
reaperhulk, line 352: I'm not sure what visualize means here. Would just `parse` be better?

src/cryptography/hazmat/primitives/serialization/pkcs7.py
```diff
@@ -328,6 +333,33 @@ def _smime_enveloped_encode(data: bytes) -> bytes:
...
+        )
+    content_type = headers["Content-Type"]
+    if "text/plain" not in content_type:
+        raise ValueError(
+            f"Decrypted MIME data content is of type '{content_type}', not "
+            "'text/plain'. Please remove the 'Text' option to visualize it."
```
reaperhulk, line 358: Same question re: visualize

reaperhulk (commented):
Other than my small nits LGTM on the text path now. Thanks!

src/cryptography/hazmat/primitives/serialization/pkcs7.py
alex: Eh? The case insensitivty handling appears to come from the line above.

src/rust/src/pkcs7.rs
```diff
@@ -164,6 +166,265 @@ fn encrypt_and_serialize<'p>(
...
+        return Err(CryptographyError::from(
+            exceptions::UnsupportedAlgorithm::new_err((
+                "RSA with PKCS1 v1.5 padding is not supported by this version of OpenSSL.",
+                exceptions::Reasons::UNSUPPORTED_PADDING,
+            )),
+        ));
+    }
+
+    // Check if all options are from the PKCS7Options enum
+    let pkcs7_option_type = types::PKCS7_TEXT.get(py)?.get_type();
```
alex, line 345: Instead of `get_type()`, it'd be better to add `PKCS7Options` to `types.rs` and use that.

alex (commented):
One final nit + Paul's comments and we can merge!

src/cryptography/hazmat/primitives/serialization/pkcs7.py
reaperhulk: Oh yes, title is doing the casing. I think I’d prefer just lowering

nitneuqr: I'm fine with both, I tried to stick with the standard case for these headers. `.lower()` might just be easier to catch though. Done!

src/cryptography/hazmat/primitives/serialization/pkcs7.py
nitneuqr: done!

src/cryptography/hazmat/primitives/serialization/pkcs7.py
nitneuqr: done!

src/rust/src/pkcs7.rs
nitneuqr: Done!

nitneuqr:
> One final nit + Paul's comments and we can merge!

Should be done now! Merci beaucoup!

src/cryptography/hazmat/primitives/serialization/pkcs7.py
nitneuqr: Actually, `get()` is straight-out a better way to handle this (case-insensitive, built-in). Modified accordingly.

alex (approved):
Thank you for all your hard work and persistence on this!

reaperhulk:
Yes, thank you for sticking with this and being so responsive. It's a pleasure having contributors like you!

nitneuqr:
My pleasure! I might come back in a few days / weeks to tackle some more problems (namely, pkcs verify and encrypt/decrypt with other content encryption algorithms)

PR #11556, opened 2024-09-06
Rustify PKCS7 unpadding

nitneuqr:
Opening this as a smaller PR needed for #11555. 

I'm new to rust, so please let me know if you see some issues in variable lifetime, or some unnecessary copying between Python & Rust.

cc @alex


src/rust/src/padding.rs
```diff
@@ -111,6 +111,65 @@ impl PKCS7PaddingContext {
...
+        py: pyo3::Python<'a>,
+        buf: CffiBuf<'a>,
+    ) -> CryptographyResult<pyo3::Bound<'a, pyo3::types::PyBytes>> {
+        match self.buffer.as_mut() {
+            Some(v) => {
+                v.extend_from_slice(buf.as_bytes());
+                let finished_blocks = (v.len() / self.block_size).saturating_sub(1);
+                let result_size = finished_blocks * self.block_size;
+                let result: Vec<u8> = v.drain(..result_size).collect();
+                Ok(pyo3::types::PyBytes::new_bound(py, &result))
```
alex, line 141: Instead of doing `collect()`, I think you can use `as_slice()`, which will save an allocation/copy.

src/rust/src/padding.rs
```diff
@@ -111,6 +111,65 @@ impl PKCS7PaddingContext {
...
+                    );
+                }
+                if !check_pkcs7_padding(&v) {
+                    return Err(
+                        pyo3::exceptions::PyValueError::new_err("Invalid padding bytes.").into(),
+                    );
+                }
+
+                let pad_size = *v.last().unwrap();
+                let result = v[..v.len() - pad_size as usize].to_vec();
```
alex, line 165:
```suggestion
                let result = &v[..v.len() - pad_size as usize];
```

reaperhulk:
You can remove the pyfunction annotation from `check_pkcs7_padding` (which should also fix your coverage issue).

nitneuqr:
> You can remove the pyfunction annotation from `check_pkcs7_padding` (which should also fix your coverage issue).

Good catch! Actually, since we're making check_pkcs7_padding private, should we simplify the i/o of the function?

alex:
Given the way check_pkcs7_padding is sensitive to being constant time,
we should be conservative in making changes. That said, if you have an
idea for how to improve it, happy to discuss!

On Fri, Sep 6, 2024 at 3:29 PM Quentin Retourne
***@***.***> wrote:
>
> You can remove the pyfunction annotation from check_pkcs7_padding (which should also fix your coverage issue).
>
> Good catch! Actually, since we're making check_pkcs7_padding private, should we simplify the i/o of the function?
>
> —



-- 

nitneuqr:
I'll check it out but saw the constant time part, and not planning on changing anything important on that side

src/rust/src/padding.rs
nitneuqr: thanks! done

src/rust/src/padding.rs
nitneuqr: Done!

nitneuqr:
Should be better!

> > You can remove the pyfunction annotation from `check_pkcs7_padding` (which should also fix your coverage issue).
> 
> Good catch! Actually, since we're making check_pkcs7_padding private, should we simplify the i/o of the function?

Just realized this didn't make sense since the function input and output types are not pyo3 already

PR #11570, opened 2024-09-09
Silencing mmap mypy warning on windows

nitneuqr:
As per #11560. Even though the lib doesn't exist on Windows

Just realized coverage does not pass as such; the simplest solution would probably be `# type: ignore` 😞

### DIFF
--- a/tests/hazmat/primitives/test_ciphers.py
+++ b/tests/hazmat/primitives/test_ciphers.py
@@ -255,7 +255,12 @@ class TestCipherUpdateInto:
     sys.platform not in {"linux", "darwin"}, reason="mmap required"
 )
 def test_update_auto_chunking():
-    large_data = mmap.mmap(-1, 2**29 + 2**20, prot=mmap.PROT_READ)
+    # Silencing mypy warning on Windows, even though mmap doesn't exist. See:
+    # https://mypy.readthedocs.io/en/stable/common_issues.html#version-and-platform-checks
+    if sys.platform == "win32":
+        large_data = mmap.mmap(-1, 2**29 + 2**20)
+    else:
+        large_data = mmap.mmap(-1, 2**29 + 2**20, prot=mmap.PROT_READ)
 
     key = b"\x00" * 16
     c = ciphers.Cipher(AES(key), modes.ECB())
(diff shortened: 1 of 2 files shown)


nitneuqr:
Just realized coverage does not pass as such; I think the simplest solution might be `# type: ignore` 😞

tests/hazmat/primitives/test_ciphers.py
```diff
@@ -255,7 +255,12 @@ def test_update_into_buffer_too_small_gcm(self, backend):
     sys.platform not in {"linux", "darwin"}, reason="mmap required"
 )
 def test_update_auto_chunking():
-    large_data = mmap.mmap(-1, 2**29 + 2**20, prot=mmap.PROT_READ)
+    # Silencing mypy warning on Windows, even though mmap doesn't exist. See:
+    # https://mypy.readthedocs.io/en/stable/common_issues.html#version-and-platform-checks
+    if sys.platform == "win32":
```
alex, line 260: Unfortunately this will produce coverage issues, since it's never reached.

I wonder if an `assert sys.platform in {"linux", "darwin"}` would work?

nitneuqr: I've seen it in the documentation, and tried it, as well as `assert sys.platform != "win32"`, without any success. it seems it only works to silence the whole file 😞

nitneuqr:
Pushed new version, probably the dirtiest we can do without going for `# type: ignore` (which might be the best option).

tests/hazmat/primitives/test_ciphers.py
```diff
@@ -255,7 +255,7 @@ def test_update_into_buffer_too_small_gcm(self, backend):
     sys.platform not in {"linux", "darwin"}, reason="mmap required"
 )
 def test_update_auto_chunking():
-    large_data = mmap.mmap(-1, 2**29 + 2**20, prot=mmap.PROT_READ)
+    large_data = large_mmap(length=2**29 + 2**20)
```
alex, line 258: Regardless of the coverage bit, a PR that simply moved to a single utility function for these would be good.

nitneuqr:
Should be good, finally :)

alex (approved):
This is dumb, I wish mypy had a better way to handle this, but thanks for chasing it down!

PR #11678, opened 2024-10-01
Remove check for OpenSSL 3.0 that's not needed any longer

alex:


alex:
it's not needed because PKCS#12 generation is now implemented by us, we don't use openssl for it.

reaperhulk (approved):

PR #11682, opened 2024-10-02
nonces

gonatienza:
Is there any interest in this nonces module?  I wrote it recently.  I thought of making it pip installable and made a repo for it but it would make more sense for it to have a home here.  If so, I would get rid of the separate repo and the pypi package.  Let me know!  Hope it helps.


gonatienza:
Will look into those errors and squash a new commit.

gonatienza:
Not sure why all-green is failing.  Will close this request and open a new one.

PR #11683, opened 2024-10-02
nonces

gonatienza:
Is there any interest in this nonces module? I wrote it recently. I thought of making it pip installable and made a repo for it but it would make more sense for it to have a home here. If so, I would get rid of the separate repo and the pypi package. Let me know! Hope it helps.

### DIFF
--- a/docs/hazmat/primitives/index.rst
+++ b/docs/hazmat/primitives/index.rst
@@ -16,3 +16,4 @@ Primitives
     symmetric-encryption
     padding
     twofactor
+    nonces
--- a/docs/hazmat/primitives/nonces.rst
+++ b/docs/hazmat/primitives/nonces.rst
@@ -1,0 +1,266 @@
+.. hazmat::
+
+Nonces
+======
+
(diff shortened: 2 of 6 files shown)


alex:
Thanks for submitting this to start a discussion.

Personally, I don't think this adds much value, because the overwhelming majority of use cases need something far simpler than `Nonces`.

Had we started with an API like this from the beginning (ala https://docs.rs/ring/latest/ring/aead/struct.Nonce.html) it might have made sense, but I think retrofitting this doesn't make much sense.

Curious for your thoughts @reaperhulk

reaperhulk:
I agree with Alex here. This is amazing interesting idea, but the standard nonce use in this library doesn’t match the complexity and retrofitting our APIs isn’t something I have any appetite for. I think it makes sense for this to continue to be a separate package that users can use in conjunction with cryptography if they so desire.

gonatienza:
Thanks for your thoughts, folks!  I will go ahead and close this PR.

PR #11820, opened 2024-10-23
Support 128-bit OID arcs

robby-cornelissen:
Addresses #11338, but can only be merged after the release of version `0.18.0` of [alex/rust-asn1](https://github.com/alex/rust-asn1) which will include the (already merged) alex/rust-asn1#484 PR.

In essence, this PR is just a unit test fix to allow OID arcs up to 128 bits. I updated the `Cargo.toml` files to point to the (as of yet unreleased) `asn1 = { version = "0.18.0", default-features = false }`, but the `Cargo.lock` file will still need to be updated after the release.


alex:
If you send a PR bumping the version in rust-asn1, I can cut a release this evening. Thanks for your work here!

alex:
If you rebase this + update Cargo.lock, we should be good to merge. Thanks for your work here!

alex (approved):

alex:
Thanks for your work here!

robby-cornelissen:
> Thanks for your work here!

Thanks for the support! Any idea when you are planning to bump the version?  No rush, just managing expectations :)

reaperhulk:
We’re not sure, but likely within the next month.

PR #11843, opened 2024-10-27
added new vectors for PKCS7 tests

nitneuqr:
As discussed on #11555, opening a quick PR to store the custom PKCS7 vectors needed for testing.


docs/development/test-vectors.rst
```diff
@@ -860,6 +863,8 @@ Custom PKCS7 Test Vectors
 * ``pkcs7/amazon-roots.der`` - A DER encoded PCKS7 file containing Amazon Root
   CA 2 and 3 generated by OpenSSL.
 * ``pkcs7/enveloped.pem`` - A PEM encoded PKCS7 file with enveloped data.
+* ``pkcs7/enveloped-aes-256-cbc.pem`` - A PEM encoded PKCS7 file with
```
reaperhulk, line 866: Is this encrypted under the public key of `/pkcs7/ca/ca.pem`? If so let's say that.

nitneuqr: It is actually encrypted under the public key of `x509/custom/ca/rsa_ca.pem`, which is used in most tests for the PKCS7 module. Adding this right now to the documentation.

docs/development/test-vectors.rst
```diff
@@ -546,6 +546,9 @@ Custom X.509 Vectors
   This is an invalid certificate per :rfc:`5280` 4.2.1.12.
 * ``malformed-san.pem`` - A certificate with a malformed SAN.
 * ``malformed-ian.pem`` - A certificate with a malformed IAN.
+* ``pkcs7/ca/rsa_ca.pem`` - A self-signed RSA certificate with ``basicConstraints``
```
reaperhulk, line 549: I don't understand what this is going to be used for. I can't really parse "This key is encoded is used to attempt decoding the PKCS7 custom vectors", but if it's just being used to test failed decrypts is there any reason we need a new vector here?

nitneuqr: That's a very good point actually! Two thoughts there:
- Is there any reason to use a RSA certificate from x509/custom/ca? Maybe a separate PR to use a dedicated CA chain for PKCS7 tests can be interesting?
- For now, if you point me towards an existing RSA cert chain, I can definitely use it for testing errors in PKCS7 decryption. I haven't found an appropriate one (probably haven't searched enough).

reaperhulk: I'm fine with a dedicated CA chain, although please make sure it's got enough lifetime in it that either quantum computers come into being or 20 years from now we upgrade it out of guilt 😄

If you need to build a valid chain of length > 1 then we typically just dynamically build those off a root (see `test_x509.py`).

nitneuqr: For now, for this type of testing (no_recipient_match) I'd rather dynamically create a chain. I'll look into `test_x509.py` and let you know. If I can manage to do without a dedicated chain, I'll update #11555 and this pull request accordingly :)

nitneuqr: I've managed to do something similar to `test_x509.py`, so I'm now only adding one new vector.

reaperhulk (approved):

Issue #11160, opened 2024-06-25
VerificationError should indicate which cert is being checked


nbastin:
VerificationError exceptions don't report which cert causes the problem, which can make debugging fairly time consuming (or impossible, if you're not in the situation of issuing your own certs).

PR #11844, opened 2024-10-28
start refactoring `ValidationError` in prep for tracking which cert had the error

alex:
The end goal is that `ValidationError` will include a `cert` field, which optionally contains a `VerificationCertificate` where relevant

refs #11160


src/rust/cryptography-x509-verification/src/policy/mod.rs
```diff
@@ -552,17 +552,21 @@ impl<'a, B: CryptoOps> Policy<'a, B> {
...
         }
 
-        let pk = issuer
-            .public_key(&self.ops)
-            .map_err(|_| ValidationError::Other("issuer has malformed public key".to_string()))?;
+        let pk = issuer.public_key(&self.ops).map_err(|_| {
+            ValidationError::new(ValidationErrorKind::Other(
+                "issuer has malformed public key".to_string(),
+            ))
+        })?;
```
alex, line 565: Sigh, this is not covered apparently.

reaperhulk (approved):

Issue #11875, opened 2024-11-01
Add support for the `Admissions` extension


hoefling:
Originally asked about whether it is possible to add support for this extension in the mailing list: [[Cryptography-dev] Adding support for Admissions extension](https://mail.python.org/pipermail/cryptography-dev/2024-October/001245.html). The extension is part of the [Common PKI v2.0](https://www.elektronische-vertrauensdienste.de/EVD/SharedDocuments/Downloads/QES/Common_PKI_v2.0_02.pdf) specification; document attached here for reference as well: [Common_PKI_v2.0_02.pdf](https://github.com/user-attachments/files/17604749/Common_PKI_v2.0_02.pdf). The relevant part of the ASN.1 syntax:
```
AdmissionSyntax ::= SEQUENCE {
  admissionAuthority GeneralName OPTIONAL,
  contentsOfAdmissions SEQUENCE OF Admissions
}

Admissions ::= SEQUENCE {
  admissionAuthority [0] EXPLICIT GeneralName OPTIONAL,
  namingAuthority [1] EXPLICIT NamingAuthority OPTIONAL,
  professionInfos SEQUENCE OF ProfessionInfo
}

NamingAuthority ::= SEQUENCE {
  namingAuthorityId OBJECT IDENTIFIER OPTIONAL,
  namingAuthorityUrl IA5String OPTIONAL,
  namingAuthorityText DirectoryString(SIZE(1..128)) OPTIONAL
}

ProfessionInfo ::= SEQUENCE {
  namingAuthority [0] EXPLICIT NamingAuthority OPTIONAL,
  professionItems SEQUENCE OF DirectoryString (SIZE(1..128)),
  professionOIDs SEQUENCE OF OBJECT IDENTIFIER OPTIONAL,
  registrationNumber PrintableString(SIZE(1..128)) OPTIONAL,
  addProfessionInfo OCTET STRING OPTIONAL
}
```

hoefling:
As the single PR is too large to submit, I will split it into a series of small PRs that will all reference this issue, so it is easier to track all of them down. I'm also in the process of getting real world X509 certificates with the extension for the testing purposes.

hoefling:
[...]

PR #11876, opened 2024-11-01
feat(admissions): add naming authority type for the admissions extension

hoefling:
This is the first PR for #11875. The relevant part of the ASN.1 syntax, as specified in [Common PKI v2.0](https://www.elektronische-vertrauensdienste.de/EVD/SharedDocuments/Downloads/QES/Common_PKI_v2.0_02.pdf), Table 29b:
```
NamingAuthority ::= SEQUENCE {
  namingAuthorityId OBJECT IDENTIFIER OPTIONAL,
  namingAuthorityUrl IA5String OPTIONAL,
  namingAuthorityText DirectoryString(SIZE(1..128)) OPTIONAL
}
```
I tried to follow the guideline of the implementation of the MS template done in #8676, especially in the field naming (not using full names like `namingAuthorityId` for fields) and unit tests.


src/rust/cryptography-x509/src/extensions.rs
```diff
@@ -372,4 +379,29 @@ mod tests {
         assert!(ku.encipher_only());
         assert!(ku.decipher_only());
     }
+
+    #[test]
```
alex, line 383: We don't usually write these like this, because they're not _really_ testing anything. They just ensure that the structure we wrote works with rust-asn1.

My suggestion would be to remove the `#[derive(asn1::Asn1Read, asn1::Asn1Write)]` from the type, and remove this test. And then you can add the derive back in a follow up PR.

src/rust/cryptography-x509/src/extensions.rs
```diff
@@ -285,9 +285,16 @@ impl KeyUsage<'_> {
     }
 }
 
+#[derive(asn1::Asn1Read, asn1::Asn1Write)]
+pub struct NamingAuthority<'a> {
+    pub id: Option<asn1::ObjectIdentifier>,
+    pub url: Option<asn1::IA5String<'a>>,
+    pub text: Option<asn1::Utf8String<'a>>,
```
alex, line 292: I see this field being a `DirectoryString` in the spec. Why UTF8 only?

alex (approved):

src/rust/cryptography-x509/src/extensions.rs
hoefling: Because I didn't pay enough attention, as usual :see_no_evil:  Thank you for the review - this is extremely helpful! Won't repeat the same mistakes in the subsequent PRs.

PR #11879, opened 2024-11-02
fixes #11878 -- check for keys too large when deriving an EC key from…

### LINKED ISSUE #11878: Serializing ECDSA key with too large private value causes to "Unknown OpenSSL error" (cryptography.exceptions.InternalError)
I noticed an odd behavior when accidentally creating an ECDSA key with a too large private key value.

Example code:
```
#!/usr/bin/python3
ecval = 0xE2563328DFABF68188606B91324281C1D58A4456431B09D510B35FECC9F307CA1822846FA2671371A9A81BAC0E35749D
key = ec.derive_private_key(ecval, ec.SECP256R1())
pem = key.private_bytes(encoding=serialization.Encoding.PEM, format=serialization.PrivateFormat.TraditionalOpenSSL, encryption_algorithm=serialization.NoEncryption())
print(pem)
```

This creates an ECDSA key with curve secp256r1, but the private key value is not 32 byte (as secp256r1 would expect), but 48 bytes (it comes from a secp384r1 key). It is expected that this code will fail, but there are 2 things unusual:
* It does not fail on derive_private_key() (maybe it should), but only after trying to export the key
* It fails with an error message indicating an internal OpenSSL error, indicating I should report a bug (which I hereby do). It should probably fail with a more helpful error message and a different exception type.

The error I get is:
```
                        when another library is not cleaning up the OpenSSL error
                        stack. If you are using cryptography with another library
                        that uses OpenSSL try disabling it before reporting a bug.
                        Otherwise please file an issue at
                        https://github.com/pyca/cryptography/issues with
                        information on how to reproduce this. ([<OpenSSLError(code=134217828, lib=16, reason=100, reason_text=buffer too small)>, <OpenSSLError(code=134742032, lib=16, reason=524304, reason_text=EC lib)>, <OpenSSLError(code=76021773, lib=9, reason=524301, reason_text=ASN1 lib)>])
```

alex:
… a private value

### DIFF
--- a/src/rust/src/backend/ec.rs
+++ b/src/rust/src/backend/ec.rs
@@ -186,7 +186,9 @@ fn derive_private_key(
     point.mul_generator(&curve, &private_value, &bn_ctx)?;
     let ec = openssl::ec::EcKey::from_private_components(&curve, &private_value, &point)
         .map_err(|_| pyo3::exceptions::PyValueError::new_err("Invalid EC key"))?;
-    check_key_infinity(&ec)?;
+    ec.check_key().map_err(|_| {
+        pyo3::exceptions::PyValueError::new_err("Invalid EC key (key out of range, infinity, etc.)")
+    })?;
     let pkey = openssl::pkey::PKey::from_ec_key(ec)?;
 
     Ok(ECPrivateKey {
--- a/tests/hazmat/primitives/test_ec.py
+++ b/tests/hazmat/primitives/test_ec.py
@@ -144,6 +144,16 @@ def test_derive_point_at_infinity(backend):
         ec.derive_private_key(q, ec.SECP256R1())
 
 
+def test_derive_point_invalid_key(backend):
+    curve = ec.SECP256R1()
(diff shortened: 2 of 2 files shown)


src/rust/src/backend/ec.rs
```diff
@@ -187,6 +187,8 @@ fn derive_private_key(
     let ec = openssl::ec::EcKey::from_private_components(&curve, &private_value, &point)
         .map_err(|_| pyo3::exceptions::PyValueError::new_err("Invalid EC key"))?;
     check_key_infinity(&ec)?;
+    ec.check_key()
```
reaperhulk, line 190: Does this also do the infinity check?

alex: It does, yes.

reaperhulk: So then should we remove `check_key_infinity` on line 189?

alex: done

reaperhulk (approved):

PR #11881, opened 2024-11-03
feat(admissions): add profession info type for the admissions extension

hoefling:
This is the second PR for https://github.com/pyca/cryptography/issues/11875. The relevant part of the ASN.1 syntax, as specified in [Common PKI v2.0](https://www.elektronische-vertrauensdienste.de/EVD/SharedDocuments/Downloads/QES/Common_PKI_v2.0_02.pdf), Table 29b:

```
ProfessionInfo ::= SEQUENCE {
  namingAuthority [0] EXPLICIT NamingAuthority OPTIONAL,
  professionItems SEQUENCE OF DirectoryString(SIZE(1..128)),
  professionOIDs SEQUENCE OF OBJECT IDENTIFIER OPTIONAL,
  registrationNumber PrintableString(SIZE(1..128)) OPTIONAL,
  addProfessionInfo OCTET STRING OPTIONAL
}
```

I am esp. unsure whether I chose the `asn1::Tlv` type for the `add_profession_info` correctly - please advise.


src/rust/cryptography-x509/src/extensions.rs
```diff
@@ -291,6 +291,24 @@ pub struct NamingAuthority<'a> {
...
+    asn1::SequenceOfWriter<'a, DisplayText<'a>, Vec<DisplayText<'a>>>,
+>;
+
+type SequenceOfObjectIdentifiers<'a> = common::Asn1ReadableOrWritable<
+    asn1::SequenceOf<'a, asn1::ObjectIdentifier>,
+    asn1::SequenceOfWriter<'a, asn1::ObjectIdentifier, Vec<asn1::ObjectIdentifier>>,
+>;
+
+pub struct ProfessionInfo<'a> {
+    pub naming_authority: Option<NamingAuthority<'a>>,
```
alex, line 305: Please add a commented out `#[explicit(0)]` since that will be needed when we add the derives.

src/rust/cryptography-x509/src/extensions.rs
```diff
@@ -291,6 +291,24 @@ pub struct NamingAuthority<'a> {
...
+    pub profession_items: SequenceOfDisplayTexts<'a>,
+    pub profession_oids: Option<SequenceOfObjectIdentifiers<'a>>,
+    pub registration_number: Option<asn1::PrintableString<'a>>,
+    pub add_profession_info: Option<asn1::Tlv<'a>>,
```
alex, line 309: This should be an octet string (`&'a [u8]`), not a TLV.

src/cryptography/x509/extensions.py
```diff
@@ -2222,6 +2222,104 @@ def __hash__(self) -> int:
...
+            and self.registration_number == other.registration_number
+            and self.add_profession_info == other.add_profession_info
+        )
+
+    def __hash__(self) -> int:
+        return hash(
+            (
+                self.naming_authority,
+                *self.profession_items,
+                *self.profession_oids,
```
alex, line 2316: Traditionally we do `tuple(self.profession_items)`, which I think is modestly more readable.

src/rust/cryptography-x509/src/extensions.rs
hoefling: I thought at first this should be it, [even tried that in the POC](https://github.com/pyca/cryptography/commit/d6a7b1563b827ff30eb655d14ae4cfff32153abc#diff-44c30c40fd401c71ad99b759debcbc52001793695a65fe349f93a20dd37e9baf)), but found no other examples for using bytes types in fields, so refrained from using it. Great that my initial thought wasn't that bad :-)

src/rust/cryptography-x509/src/extensions.rs
hoefling: @alex done, please verify. Do you think it makes sense to also add the `#[derive(asn1::Asn1Read, asn1::Asn1Write)]` headers as well, commented out for now?

alex: sure

hoefling: @alex done!

alex (approved):

PR #11883, opened 2024-11-03
feat(admissions): add admission type for the admissions extension

hoefling:
This is the third PR for https://github.com/pyca/cryptography/issues/11875. The relevant part of the ASN.1 syntax, as specified in [Common PKI v2.0](https://www.elektronische-vertrauensdienste.de/EVD/SharedDocuments/Downloads/QES/Common_PKI_v2.0_02.pdf), Table 29b:

```
Admissions ::= SEQUENCE {
  admissionAuthority [0] EXPLICIT GeneralName OPTIONAL,
  namingAuthority [1] EXPLICIT NamingAuthority OPTIONAL,
  professionInfos SEQUENCE OF ProfessionInfo
}
```

### DIFF
--- a/src/rust/cryptography-x509/src/extensions.rs
+++ b/src/rust/cryptography-x509/src/extensions.rs
@@ -312,6 +312,15 @@ pub struct ProfessionInfo<'a> {
     pub add_profession_info: Option<&'a [u8]>,
 }
 
+pub struct Admission<'a> {
+    pub admission_authority: Option<name::GeneralName<'a>>,
+    pub naming_authority: Option<NamingAuthority<'a>>,
+    pub profession_infos: common::Asn1ReadableOrWritable<
+        asn1::SequenceOf<'a, ProfessionInfo<'a>>,
+        asn1::SequenceOfWriter<'a, ProfessionInfo<'a>, Vec<ProfessionInfo<'a>>>,
+    >,
+}
+
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


src/rust/cryptography-x509/src/extensions.rs
```diff
@@ -312,6 +312,20 @@ pub struct ProfessionInfo<'a> {
...
+pub struct Admission<'a> {
+    // #[explicit(0)]
+    pub admission_authority: Option<name::GeneralName<'a>>,
+    // #[explicit(1)]
+    pub naming_authority: Option<NamingAuthority<'a>>,
+    /*
+    pub profession_infos: common::Asn1ReadableOrWritable<
+        asn1::SequenceOf<'a, ProfessionInfo<'a>>,
+        asn1::SequenceOfWriter<'a, ProfessionInfo<'a>, Vec<ProfessionInfo<'a>>>,
+    >,
```
alex, line 325: I assume this needs to be commented out because the traits are not yet available?

hoefling: @alex yes exactly, the `SequenceOfWriter` expects the Asn1Readable/Writable traits, orherwise the code will not compile.

alex: I guess this is fine for now.

alex (approved):

PR #11892, opened 2024-11-04
feat(admissions): implement encoding of admissions extension

hoefling:
This is the fifth PR for https://github.com/pyca/cryptography/issues/11875 which adds encoding of the `Admissions` extension object. Please review with caution - this is the part I am least confident in, and will be grateful for all improvement suggestions. Please also bear with my Rust coding skills - for this PR, I mostly learned from the existing codebase and PyO3 docs.

### DIFF
--- a/src/rust/src/x509/extensions.rs
+++ b/src/rust/src/x509/extensions.rs
@@ -416,6 +416,34 @@ fn encode_scts(ext: &pyo3::Bound<'_, pyo3::PyAny>) -> CryptographyResult<Vec<u8>
     Ok(asn1::write_single(&result.as_slice())?)
 }
 
+fn encode_naming_authority<'a>(
+    py: pyo3::Python<'_>,
+    ka_str: &'a cryptography_keepalive::KeepAlive<pyo3::pybacked::PyBackedStr>,
+    py_naming_authority: &pyo3::Bound<'a, pyo3::PyAny>,
+) -> Result<extensions::NamingAuthority<'a>, CryptographyError> {
+    let py_oid = py_naming_authority.getattr(pyo3::intern!(py, "id"))?;
+    let id = if py_oid.is_truthy()? {
+        Some(py_oid_to_oid(py_oid)?)
+    } else {
+        None
+    };
+    let py_url = py_naming_authority.getattr(pyo3::intern!(py, "url"))?;
+    let url = if py_url.is_truthy()? {
+        let py_url_str = ka_str.add(py_url.extract::<PyBackedStr>()?);
+        asn1::IA5String::new(py_url_str)
+    } else {
+        None
+    };
+    let py_text = py_naming_authority.getattr(pyo3::intern!(py, "text"))?;
+    let text = if py_text.is_truthy()? {
+        let py_text_str = ka_str.add(py_text.extract::<PyBackedStr>()?);
+        Some(extensions::DisplayText::Utf8String(asn1::Utf8String::new(py_text_str)))
+    } else {
+        None
+    };
+    Ok(extensions::NamingAuthority { id, url, text })
+}
+
 pub(crate) fn encode_extension(
     py: pyo3::Python<'_>,
     oid: &asn1::ObjectIdentifier,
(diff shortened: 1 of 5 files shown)


src/rust/src/x509/extensions.rs
```diff
@@ -416,6 +416,133 @@ fn encode_scts(ext: &pyo3::Bound<'_, pyo3::PyAny>) -> CryptographyResult<Vec<u8>
     Ok(asn1::write_single(&result.as_slice())?)
 }
 
+fn encode_naming_authority<'a>(
+    py: pyo3::Python<'_>,
+    ka_str: &'a cryptography_keepalive::KeepAlive<pyo3::pybacked::PyBackedStr>,
+    py_naming_authority: &pyo3::Bound<'a, pyo3::PyAny>,
+) -> Result<extensions::NamingAuthority<'a>, CryptographyError> {
```
alex, line 423: You can use `CryptographyResult<extensions::NamingAuthority<'a>>` which is marginally shorter

src/rust/src/x509/extensions.rs
```diff
@@ -416,6 +416,133 @@ fn encode_scts(ext: &pyo3::Bound<'_, pyo3::PyAny>) -> CryptographyResult<Vec<u8>
...
+    let py_oid = py_naming_authority.getattr(pyo3::intern!(py, "id"))?;
+    let id = if py_oid.is_truthy()? {
```
alex, line 425: Prefer `is_none()`, it better expresses what we want, and is also a smidge faster.

src/rust/src/x509/extensions.rs
```diff
@@ -416,6 +416,133 @@ fn encode_scts(ext: &pyo3::Bound<'_, pyo3::PyAny>) -> CryptographyResult<Vec<u8>
...
+        Some(py_oid_to_oid(py_oid)?)
+    } else {
+        None
+    };
+    let py_url = py_naming_authority.getattr(pyo3::intern!(py, "url"))?;
+    let url = if py_url.is_truthy()? {
```
alex, line 431: Here its particularly important to use `is_none()`: If you have an empty string, right now it'd get turned into `None`.

src/rust/src/x509/extensions.rs
```diff
@@ -416,6 +416,133 @@ fn encode_scts(ext: &pyo3::Bound<'_, pyo3::PyAny>) -> CryptographyResult<Vec<u8>
...
+        let py_url_str = ka_str.add(py_url.extract::<PyBackedStr>()?);
+        asn1::IA5String::new(py_url_str)
```
alex, line 433: If this returns `None` we want to propagate an exception; that can happen if the user passes a non-ascii string.

src/rust/src/x509/extensions.rs
hoefling: Duly noted :-) also switched to `CryptographyResult` for consistency in other introduced functions as well.

src/rust/src/x509/extensions.rs
hoefling: This is a really great hint, thank you! Indeed using it makes sense in all occurences, and by doing that I even uncovered an error I slipped into the codebase in #11881 - the `professionOIDs` field is optional, but I missed that. Will amend this in a separate PR.

hoefling: @alex what I forgot to ask: what do you prefer more?
```rs
let val = if !py_val.is_none() {
    Some(...)
} else {
    None
};
```
or
```rs
let val = if py_val.is_none() {
    None
} else {
    Some(...)
};
```

alex: Whichever you prefer, I'm sure we're not consistent

src/rust/src/x509/extensions.rs
hoefling: @alex please check whether the suggested implementation is readable and the error message is descriptive. I added a simple test to guard against non-ascii values. Do I understand it correctly that the same applies to `registration_number` which is a `PrintableString`?

alex: Yup!

hoefling: @alex done! Thank you yet again for a helpful review :handshake:

alex (approved):

PR #11898, opened 2024-11-05
fixes #11160 -- include the cert in the error message for verification error

### LINKED ISSUE #11160: VerificationError should indicate which cert is being checked

alex:


reaperhulk (approved):
This does make me think it might be useful to print a cert fingerprint in the repr as well so that there’s a completely unambiguous identifier in this error.

Issue #11585, opened 2024-09-11
Releasing GIL in cryptographic operations


JakubKahovec:
I'd like to ask if GIL gets released when cryptographic operations are being performed as I can't find it in the code, i.e https://github.com/pyca/cryptography/blob/60913069bb27d788c57687840a8b1b54904e9139/src/rust/src/backend/kdf.rs#L10

alex:
Not really. I thought we did for things like scrypt, but I don't see it, so
I guess not. I'd be open to taking PRs for these though

PR #11900, opened 2024-11-05
Release GIL during cipher update

kcking:
Related to discussion in https://github.com/pyca/cryptography/issues/11585.

### DIFF
--- a/src/rust/src/backend/ciphers.rs
+++ b/src/rust/src/backend/ciphers.rs
@@ -179,22 +179,26 @@ impl CipherContext {
         }
 
         let mut total_written = 0;
-        for chunk in data.chunks(1 << 29) {
-            // SAFETY: We ensure that outbuf is sufficiently large above.
-            unsafe {
-                let n = if self.py_mode.bind(py).is_instance(&types::XTS.get(py)?)? {
-                    self.ctx.cipher_update_unchecked(chunk, Some(&mut buf[total_written..])).map_err(|_| {
-                    pyo3::exceptions::PyValueError::new_err(
-                        "In XTS mode you must supply at least a full block in the first update call. For AES this is 16 bytes."
-                    )
-                })?
-                } else {
-                    self.ctx
-                        .cipher_update_unchecked(chunk, Some(&mut buf[total_written..]))?
-                };
-                total_written += n;
+        let is_xts = self.py_mode.bind(py).is_instance(&types::XTS.get(py)?)?;
+        py.allow_threads::<CryptographyResult<()>, _>(|| {
+            for chunk in data.chunks(1 << 29) {
+                // SAFETY: We ensure that outbuf is sufficiently large above.
+                unsafe {
+                    let n = if is_xts {
+                        self.ctx.cipher_update_unchecked(chunk, Some(&mut buf[total_written..])).map_err(|_| {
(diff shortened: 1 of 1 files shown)


alex:
Can you share anything about the performance impact of this?

I'm modestly nervous about performance overhead for very small chunks, but what can you do 🙃

src/rust/src/backend/ciphers.rs
```diff
@@ -183,14 +183,18 @@ impl CipherContext {
             // SAFETY: We ensure that outbuf is sufficiently large above.
             unsafe {
                 let n = if self.py_mode.bind(py).is_instance(&types::XTS.get(py)?)? {
```
alex, line 185: Actually, we should hoist this out of the loop, and then you can wrpa the whole loop in `py.allow_threads()`. This will improve readability, and also means we won't re-acquire the GIL multiple times for a giant ciphertext.

kcking: Great point, just updated this.

kcking:
> modestly nervous about performance overhead
Releasing the GIL should be order of microseconds. I could put together a benchmark for small blocks to see if the overhead is meaningful

alex (approved):
If you can share some performance numbers, we can get this merged. Thanks!

alex:
Ooops, crossed streams :-)

Do you have any numbers for the impact of this on your application?

kcking:
I can get some for you hopefully tonight, thanks for the quick review!

alex:
Sounds great, thanks for the contribution

reaperhulk:
Just a quick ping re: perf numbers 😄

reaperhulk:
I ran some tests on an M1 Max (10 cores, 10 threads) and it appears GIL release is slower than single threading up to ~23kB `update` calls. I'd be curious if the same is true on other CPUs.

Some numbers:
| Payload Size                | Main  | PR    |
|-----------------------------|-------|-------|
| 64 bytes (100MB per thread) | 16.6s | 91s   |
| 23 kB (1024MB per thread)   | 4.7s  | 4.5s  |
| 1 MB (1024MB per thread)     | 4.3s  | 0.55s |


These were derived using the following script and varying the size of `DATA` and `TOTAL` (100 for the first test, 1024 for the remaining).
```python
import time
import concurrent.futures
from cryptography.hazmat.primitives.ciphers import Cipher, algorithms

TOTAL = 1024 * 1024 * 100
DATA = bytes(64)
ITERATIONS = TOTAL // len(DATA)
KEY = bytes(32)

start = time.time()

def encrypt(num):
    buf = bytearray(len(DATA))
    c = Cipher(algorithms.ChaCha20(KEY, b"0" * 16), None)
    encryptor = c.encryptor()
    for _ in range(ITERATIONS):
        encryptor.update_into(DATA, buf)
    print(f"{num} thread complete. {ITERATIONS * len(DATA) / 1024 / 1024} MiB encrypted")


with concurrent.futures.ThreadPoolExecutor() as executor:
    executor.map(encrypt, range(0, 10))

print(time.time() - start)
```

alex:
There's been no action here for a couple months. Given there's a need to find an approach that has consistent performance on smaller chunks, I'm going to close this. If you're interested in continuing to work on it, just let us know, happy to reopen.

PR #11908, opened 2024-11-06
fix(admissions): allow profession_oids field being none

hoefling:
This is the fix that I mentioned [in the review of #11892](https://github.com/pyca/cryptography/pull/11892#discussion_r1828456536). The ASN.1 syntax for `ProfessionInfo` is
```
ProfessionInfo ::= SEQUENCE {
  namingAuthority [0] EXPLICIT NamingAuthority OPTIONAL,
  professionItems SEQUENCE OF DirectoryString (SIZE(1..128)),
  professionOIDs SEQUENCE OF OBJECT IDENTIFIER OPTIONAL,
  registrationNumber PrintableString(SIZE(1..128)) OPTIONAL,
  addProfessionInfo OCTET STRING OPTIONAL
}
```
but I missed the optional part for `professionOIDs` in #11881. This PR makes sure that `professionOIDs` can be initialized with/encoded and parsed as `None`.

Note that this PR contains changes of both #11892 and #11903, its actual diff is a lot smaller. Will make this PR to a draft for now, first the both PRs above should get a good shape for merging.

### DIFF
--- a/src/cryptography/x509/extensions.py
+++ b/src/cryptography/x509/extensions.py
@@ -2227,7 +2227,7 @@ class ProfessionInfo:
         self,
         naming_authority: NamingAuthority | None,
         profession_items: typing.Iterable[str],
-        profession_oids: typing.Iterable[ObjectIdentifier],
+        profession_oids: typing.Iterable[ObjectIdentifier] | None,
         registration_number: str | None,
         add_profession_info: bytes | None,
     ) -> None:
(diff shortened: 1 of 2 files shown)


hoefling:
This PR is also part of resolution for #11875, linking it to the issue.

alex:
I think it'd probably make more sense to land this PR _first_. Is it possible to extract the parts that apply on what's in main now?

hoefling:
@alex sure, I can leave only the changes for the Python side in this one and land the additional changes for parsing/encoding in the related PRs. Will do!

alex:
Great, thanks.

src/cryptography/x509/extensions.py
```diff
@@ -2309,11 +2310,15 @@ def __eq__(self, other: object) -> bool:
...
+        if self.profession_oids is None:
+            profession_oids: tuple[ObjectIdentifier | None, ...] = (None,)
+        else:
+            profession_oids = tuple(self.profession_oids)
         return hash(
             (
                 self.naming_authority,
                 *tuple(self.profession_items),
-                *tuple(self.profession_oids),
+                *tuple(profession_oids),
```
alex, line 2321: Instead of doing this, I'd suggest:

```
if self.professional_oids is None:
    profession_oids = None
else:
    profession_oids = tuple(self.profession_oids)
```

and then remove the `*` unpacking

alex: It's probably not needed, no. If you want to do taht cleanup, let's do it in a separate PR.

alex:
Can you mark this PR as ready, unless there's something you're still working on for it.

PR #11922, opened 2024-11-08
passing PKCS7 Content Info to readable

nitneuqr:
Linked to #11555 😄


alex (approved):
We might need to remove the `derive()` due to coverage, but substance-wise it LGTM. Thank you!

PR #11924, opened 2024-11-09
Specify minimum versions for more deps

alex:
Right now our deps are basically wrong, and impossible to use with lowest version resolution. Let's start trying to specify minimums so our deps are properly accurate.


alex:
(https://github.com/astral-sh/uv/issues/8962 is a blocker to really properly debugging and resolving all of these)

reaperhulk (approved):

PR #11925, opened 2024-11-09
Move asn1 to be a workspace dep

alex:
This makes it easier to change, you only need to touch one thing


reaperhulk:
Do we have any other global deps we should hoist up?

alex:
Almost certainly, pyo3 and openssl are the other obvious ones. But this was the one I was futzing with :-)

reaperhulk (approved):

PR #11942, opened 2024-11-13
chore(CI): Update AWS Encryption SDK downstream test invocation

lucasmcdonald3:


alex:
Thanks for sending this so quickly!

Just to confirm: `tests/mpl/` isn't something that make sense to run in our CI?

alex (approved):

lucasmcdonald3:
> Thanks for sending this so quickly!
> 
> Just to confirm: `tests/mpl/` isn't something that make sense to run in our CI?

I would say "not immediately".
`/tests/mpl` requires Python 3.11+ and an additional (optional) dependency: https://pypi.org/project/aws-cryptographic-material-providers/
I'd be happy to open a new PR soon that augments your ESDK downstream tests to include this directory's tests.
But this PR as-is will unblock your builds immediately.

alex:
I confess I don't totally understand what that library does, but if it's tests exercise the way the library interacts with cryptography, they're probably worth running.

Will merge this for now.

Thanks again!

PR #11964, opened 2024-11-16
docs(admissions): add documentation for the admissions extension

hoefling:
This should be my last PR for https://github.com/pyca/cryptography/issues/11875, adding the documentation for the Admissions extension. It was suprisingly hard to describe the types and fields, since the original specification is rather vague and my particular use case of the extension is often too specific to add it (e.g. the extension should have exactly one `ProfessionInfo`, with exactly one pair of text and OID in the `profession_items` and `profession_oids` etc). Therefore, I will gladly rewrite the passages that are less readable or are unclear.

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -3000,25 +3000,22 @@ X.509 Extensions
 
     .. versionadded:: 44.0.0
 
-    The admissions extension
+    The admissions extension contains information on registration and professional admission,
+    as specified by `Common PKI v2`_.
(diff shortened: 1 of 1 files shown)


hoefling:
Looks like the CI failed not due to the changes in this PR; added #11965 to address this. Will rebase onto it and retry.

alex:
Merged that, thanks!

alex (approved):
If you can send a PR for the changelog, I think that'll be the last piece!

DorisJeanParish:
GitHub.com/ID:049921093

PR #11966, opened 2024-11-16
refactor: replace returning pyobject with bound<'p, pyany> in asn1 module

hoefling:
This is a small proof of concept for the idea of amending two things:
1. Replacement of `Result<..., CryptographyError>` usages with `CryptographyResult<...>`, originally suggested by @alex in https://github.com/pyca/cryptography/pull/11892#discussion_r1828409224
2. Replacement of `CryptographyResult<pyo3::PyObject>` return types by `CryptographyResult<pyo3::Bound<'p, pyo3::PyAny>>`, originally suggested by @alex in https://github.com/pyca/cryptography/pull/11903#discussion_r1833019409

The changes are done for the `asn1` module only as a showcase. I didn't want to change it eveywhere at once because a) not sure if the change is worth it and b) the PR will become too large. @alex do you think it is a good idea to do this for consistency throughout the whole Rust codebase?


alex (approved):

alex:
I think this is a reasonable cleanup. The current code works fine, but we're always happy to improve.

PR #11971, opened 2024-11-16
chore(admissions): add changelog entry for the admissions extension addition

hoefling:
This is the last PR for adding the `Admissions` X509 extension, closing #11875. This only adds an entry to the changelog.


alex (approved):
Thanks for all your work here!

Issue #12043, opened 2024-11-26
ImportError: DLL load failed while importing _rust: The specified procedure could not be found.


bigb-me:
**Hi i'm getting the error:**

Traceback (most recent call last):
  File "C:\Users\--\OneDrive\Desktop\OneDrive Dump\Messenger v7.py", line 10, in <module>
    from cryptography.fernet import Fernet
  File "C:\Users\--\AppData\Roaming\Python\Python39\site-packages\cryptography\fernet.py", line 14, in <module>
    from cryptography.exceptions import InvalidSignature
  File "C:\Users\--\AppData\Roaming\Python\Python39\site-packages\cryptography\exceptions.py", line 9, in <module>
    from cryptography.hazmat.bindings._rust import exceptions as rust_exceptions
ImportError: DLL load failed while importing _rust: The specified procedure could not be found.
>>> 

I have no clue how to fix this, I have tried installing cryptography in a virtual environment:
python -m venv myenv
myenv\Scripts\activate
pip install cryptography

I tried coping the - cryptography and cryptography-43.0.0.dist-info folder (from another laptop where cryptography works) onto this device, wich gives the same error on this device

I have tried going to the Rust website and installing it - I installed and verified with: rustc --version.

When compiling the code into exe (with auto-py-to-exe) it works on this devise. 

I got this error when adding encryption to my code, the error also happens on smaller pieces of code so, to keep this short I will just show a small example program wich still gives the error. It is below:

from cryptography.fernet import Fernet

key = Fernet.generate_key()
cipher = Fernet(key)
message = "Test message."
encrypted_message = cipher.encrypt(message.encode())
decrypted_message = cipher.decrypt(encrypted_message).decode()

print(f"Original message: {message}")
print(f"Encrypted message: {encrypted_message}")
print(f"Decrypted message: {decrypted_message}")

**OS:**
Windows 10

**Installed with:**
pip install cryptography  
**I also tried:**
pip install cryptography --no-binary cryptography

Help would be appreciated!

**Versions:**
Cryptography Version: 43.0.0
Python 3.9.0
pip 24.3.1
Setuptools Version: 75.6.0

bigb-me:
> Try upgrading to a later Python version (3.9.x or otherwise). See #11325

Upgrade to a higher VS then my current 3.9.0?

reaperhulk:
Yes.

bigb-me:
> Yes.

[...]

Issue #11325, opened 2024-07-21
ImportError: DLL load failed while importing _rust: The specified procedure could not be found.


byteherder:
I am trying to install Scrapy using PyCharm as my IDE. I have followed the install procedure on their website. I am getting the following error.

Traceback (most recent call last):
  File "C:\Users\chris\AppData\Local\Programs\Python\Python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\chris\AppData\Local\Programs\Python\Python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\chris\PycharmProjects\WebCrawler\venv\Scripts\scrapy.exe\__main__.py", line 4, in <module>
  File "c:\users\chris\pycharmprojects\webcrawler\venv\lib\site-packages\scrapy\cmdline.py", line 9, in <module>
    from scrapy.commands import BaseRunSpiderCommand, ScrapyCommand, ScrapyHelpFormatter
  File "c:\users\chris\pycharmprojects\webcrawler\venv\lib\site-packages\scrapy\commands\__init__.py", line 11, in <module>
    from scrapy.crawler import CrawlerProcess
  File "c:\users\chris\pycharmprojects\webcrawler\venv\lib\site-packages\scrapy\crawler.py", line 27, in <module>
    from scrapy.core.engine import ExecutionEngine
  File "c:\users\chris\pycharmprojects\webcrawler\venv\lib\site-packages\scrapy\core\engine.py", line 28, in <module>
    from scrapy.core.downloader import Downloader
  File "c:\users\chris\pycharmprojects\webcrawler\venv\lib\site-packages\scrapy\core\downloader\__init__.py", line 16, in <module>
    from scrapy.signalmanager import SignalManager
  File "c:\users\chris\pycharmprojects\webcrawler\venv\lib\site-packages\scrapy\signalmanager.py", line 6, in <module>
    from scrapy.utils import signal as _signal
  File "c:\users\chris\pycharmprojects\webcrawler\venv\lib\site-packages\scrapy\utils\signal.py", line 20, in <module>
    from scrapy.utils.log import failure_to_exc_info
  File "c:\users\chris\pycharmprojects\webcrawler\venv\lib\site-packages\scrapy\utils\log.py", line 26, in <module>
    from scrapy.utils.versions import scrapy_components_versions
[...]

PR #12045, opened 2024-11-26
Prohibit Python 3.9.0, 3.9.1 -- they have a bug that causes errors

alex:


reaperhulk:
See https://github.com/pyca/cryptography/issues/12043 and #11325

reaperhulk (approved):

PR #12077, opened 2024-11-29
Allow sub-class certificate

alex:
This is to enable folks who previously subclassed teh ABC to persist

### DIFF
--- a/src/rust/src/x509/certificate.rs
+++ b/src/rust/src/x509/certificate.rs
@@ -39,7 +39,7 @@ self_cell::self_cell!(
     }
 );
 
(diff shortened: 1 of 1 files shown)


reaperhulk:
Should we allow subclassing on all of them or wait for users to request?

If we consider subclassing officially part of our public surface we should add a test that simply subclasses without an exception with a small comment explaining why.

alex:
I suppose it's a begrudgingly supported part of our API surface. I'll add a test. No reason to extend it further though.

alex:
It's a good thing you told me to add a test. While subclassing now works, it's not actually possible to _instantiate_ an instance of this AFAICT.

alex:
I'm not sure what the correct way to proceed is, as we don't want these to be instantiable normally.

PR #12082, opened 2024-12-01
Fix the return type from get_revoked_certificate_by_serial_number

### LINKED ISSUE #12081: Wrong typing on get_revoked_certificate_by_serial_number
After https://github.com/pyca/cryptography/pull/11991 the typed return value from `get_revoked_certificate_by_serial_number()`on `CertificateRevocationList` is `cryptography.hazmat.bindings._rust.x509.RevokedCertificate` instead of `cryptography.x509.base.RevokedCertificate` (which is what the iterator returns). They should probably match.

```
from typing import reveal_type

from cryptography import x509


def load_crl_and_print_revocation_date(crl_bytes: bytes) -> None:
    crl = x509.load_der_x509_crl(crl_bytes)
    revoked_cert = crl.get_revoked_certificate_by_serial_number(123)

    assert revoked_cert is not None

    reveal_type(revoked_cert)

    print(revoked_cert.revocation_date_utc)
```

With cryptography 43.0.3:
```
❯ mypy testulf.py
testulf.py:12: note: Revealed type is "cryptography.x509.base.RevokedCertificate"
Success: no issues found in 1 source file
```

With cryptography 44.0.0:
```
❯ mypy testulf.py
testulf.py:12: note: Revealed type is "cryptography.hazmat.bindings._rust.x509.RevokedCertificate"
testulf.py:14: error: "RevokedCertificate" has no attribute "revocation_date_utc"  [attr-defined]
Found 1 error in 1 file (checked 1 source file)
```

magnuswatn:
Fixes #12081


alex (approved):
Good catch. (Annoyingly, I caught one of these mistakes in the original PR, but missed this one).

Thanks for the PR!

PR #12088, opened 2024-12-02
PKCS7 Decryption: more content decryption algorithms

nitneuqr:
Opening this PR to handle more decryption cases: emails in Microsoft Outlook are generally content-encrypted with AES-256-CBC.

In terms of code, I have one question: should I leave the code as-is or add a `AES192` subclass in `algorithms.py`, to be used in Rust  by pyO3?


alex:
Does anything in the real world use AES-192? We don't have a type for it precisely because its basically a historical releic.

nitneuqr:
I've never seen it in Outlook: I've seen AES 128, 256, and RC2 (!!). I don't mind removing it from decryption & encryption. Up to you!

nitneuqr:
Removed support then. Should be good now!

alex (approved):

PR #12101, opened 2024-12-04
handle case where a "valid" pkey does not contain a valid EC key

reaperhulk:
This probably needs the test case from #12100


alex:
Yes, tests needed. Patch looks good though

reaperhulk:
OpenSSL 1.1.1 and LibreSSL do not fail on this vector, while boring and OpenSSL do.

alex (approved):

Issue #11859, opened 2024-10-30
copy.copy for private keys


joakimnordling:
Private keys (at least RSAPrivateKey) can not be copied in cryptography 43.0.3:
```
>>> import copy
>>> from cryptography.hazmat.primitives.asymmetric import rsa
>>> private_key = rsa.generate_private_key(
...     public_exponent=65537,
...     key_size=2048,
... )
>>> copied_key = copy.copy(private_key)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/jocke/.pyenv/versions/3.11.7/lib/python3.11/copy.py", line 92, in copy
    rv = reductor(4)
         ^^^^^^^^^^^
TypeError: cannot pickle 'cryptography.hazmat.bindings._rust.openssl.rsa.RSAPrivateKey' object
```

There was a similar issue for public keys reported in https://github.com/pyca/cryptography/issues/9403 and it seems like the fix to that issue only made it possible to copy all the public keys, not any of the private ones.

joakimnordling:
The actual use case I had was to make my own `RsaPrivateKey` (note the lower case `sa`) subclass of the `SecretStr` from [pydantic](https://docs.pydantic.dev/) so that I could have an environment variable with a private key in PEM format and get it parsed to this "RsaPrivateKey" subclass when loading the settings for my app. My thought was that I would in the `__init__` use the `load_pem_private_key` to parse the key to the `RSAPrivateKey` (the one from cryptography with capital `SA`) and store it as an instance variable for later use, similar to how for example the `last4` is done in the [PaymentCardNumber](https://github.com/pydantic/pydantic/blob/main/pydantic/types.py#L1863). The problem is that pydantic decides that it wants to copy the object for some reason when initing the settings (I didn't investigate deeply why it wants to do that). And the reason I want to do it in the init and save it is to avoid doing the `load_pem_private_key` multiple times, as it's somewhat slow on larger keys. I hope this was not too confusing and I can try to explain in more detail if needed. But in short it's somewhat similar to wrapping it into a dataclass.

joakimnordling:
I agree it's weird that it does that deepcopy, but still I don't see that as a reason why cryptography shouldn't support copying the key. There's likely sooner or later some other use case for it and shouldn't be many lines of change to support it, if it's immutable. So I'd rather ask the question as why should the RSAPrivateKey not support copying?

[...]

PR #12110, opened 2024-12-06
Support for copying private keys

joakimnordling:
This should solve the issue https://github.com/pyca/cryptography/issues/11859, that private keys, like RSAPrivateKey can not be copied.

### DIFF
--- a/src/rust/src/backend/dh.rs
+++ b/src/rust/src/backend/dh.rs
@@ -235,6 +235,10 @@ impl DHPrivateKey {
             false,
         )
     }
+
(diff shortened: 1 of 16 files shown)


alex (commented):
In addition to adding the implementations to the concrete classes, you'll want to add them to the ABCs for each of these key types as well.

alex:
I think it's fine to load a weak key and skip on FIPS and be marginally faster.

joakimnordling:
Thanks, I reverted back to loading the key from a file and added a `skip_fips` to that particular test.

I also added abstract methods to the abstract base classes. The ABCs for the public keys were also missing the abstract version of the `__copy__` methods, so I added it to those as well.

alex (approved):

PR #12113, opened 2024-12-06
Add ec.EllipticCurve group_order constants

DavidBuchanan314:
Follow-up to https://github.com/pyca/cryptography/pull/10600

This PR hard-codes `group_order` constants for all the curves specified in `ec.py`

For what it's worth, I sourced them from https://neuromancer.sk/std/search/ for ease of copy-pasting - please do cross-check!

Some of the int literals are very long and had to be #noqa'd (I don't think there's any good way to avoid it)

### DIFF
--- a/docs/hazmat/primitives/asymmetric/ec.rst
+++ b/docs/hazmat/primitives/asymmetric/ec.rst
@@ -500,6 +500,14 @@ Key Interfaces
         Size (in :term:`bits`) of a secret scalar for the curve (as generated
         by :func:`generate_private_key`).
 
+    .. attribute:: group_order
+
+        .. versionadded:: 45
+
+        :type: int
+
+        The order of the curve's group.
+
 
 .. class:: EllipticCurveSignatureAlgorithm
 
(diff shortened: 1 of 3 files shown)


DavidBuchanan314:
I came up with a fairly simple way to test the correctness:

```py
from cryptography.hazmat.primitives.asymmetric import utils

MSG = b"hello"
ECDSA_SHA256 = ec.ECDSA(hashes.SHA256())

for curve_name, curve in ec._CURVE_TYPES.items():
	try:
		signing_key = ec.generate_private_key(curve)
	except:
		print(curve_name, "not supported by this backend")
		continue

	verifying_key = signing_key.public_key()

	# generate a regular signature
	sig1 = signing_key.sign(MSG, ECDSA_SHA256)
	verifying_key.verify(sig1, MSG, ECDSA_SHA256) # sanity check

	# demostrate signature malleability
	r, s = utils.decode_dss_signature(sig1)
	sig2 = utils.encode_dss_signature(r, curve.group_order - s)
	
	assert(sig1 != sig2)
	verifying_key.verify(sig2, MSG, ECDSA_SHA256) # still verifies

	print(curve_name, "passed")
```

If the value of `group_order` was incorrect, `sig2` would not verify.

(let me know if you want an equivalent test put into the actual test suite)

alex:
I'm going to defer to @reaperhulk on reviewing this. He's currently traveling, so a review may be slightly delayed.

reaperhulk (approved):
Validated, thanks for doing this!

PR #12114, opened 2024-12-06
Adopt sphinx-inline-tabs

alex:
Use it in one demo place to show how to install us with multiple installers


alex:
https://cryptography--12114.org.readthedocs.build/en/12114/#installation

reaperhulk (approved):

PR #12116, opened 2024-12-06
support for PKCS7_NO_VERIFY in test_support.pkcs7_verify

zworkb:
This PR adds support for Pkcs7Flags::NOVERIFY so that we can verify pkcs7 signatures without verifying the certificates. 

M2Crypto also exposes this option but I would prefer to use cryptography ;)

It is needed when the verifying a signed apple wallet pass. Now I can verify it with the call

```python
    test_support.pkcs7_verify(
        serialization.Encoding.DER,
        signature,
        manifest_json.encode("utf-8"),
        [],
        [pkcs7.PKCS7Options.NoVerify],
    )
```

without the NoVerify option I get the error:

```
...
... (error:10800075:PKCS7 routines:PKCS7_verify:certificate verify error:../crypto/pkcs7/pk7_smime.c:295:Verify error: unable to get local issuer certificate)
```

The unittests as in https://github.com/pyca/cryptography/blob/7a20576a420983e5ca04aa70a9bfb47fd285f097/tests/hazmat/primitives/test_pkcs7.py#L412

use ECPrivateKey but when using RSAPrivateKey I get the above error.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ target/
 .rust-cov/
 *.lcov
 *.profdata
+venv
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+    "python.testing.pytestArgs": [
+        "tests"
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true
+}
--- a/src/cryptography/hazmat/primitives/serialization/pkcs7.py
+++ b/src/cryptography/hazmat/primitives/serialization/pkcs7.py
@@ -43,6 +43,7 @@ class PKCS7Options(utils.Enum):
     NoCapabilities = "Don't embed SMIME capabilities"
     NoAttributes = "Don't embed authenticatedAttributes"
     NoCerts = "Don't embed signer certificate"
+    NoVerify = "Don't verify the signers certificate of a signed message"
 
 
 class PKCS7SignatureBuilder:
--- a/src/rust/src/test_support.rs
+++ b/src/rust/src/test_support.rs
@@ -81,6 +81,9 @@ fn pkcs7_verify(
     if options.contains(types::PKCS7_TEXT.get(py)?)? {
         flags |= openssl::pkcs7::Pkcs7Flags::TEXT;
     }
+    if options.contains(types::PKCS7_NO_VERIFY.get(py)?)? {
+        flags |= openssl::pkcs7::Pkcs7Flags::NOVERIFY;
+    }
 
     let store = {
         let mut b = openssl::x509::store::X509StoreBuilder::new()?;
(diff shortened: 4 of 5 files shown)


alex:
`test_support.pkcs7_verify` is not a public API, so we're not going to take PRs to expand its functionality.

If you're interested in helping us design a public API for PKCS#7 verification, that's something we'd be interested in!

zworkb:
Currently I am migrating a project from M2Crypto to cryptography. There is an API for verification of pkcs7: https://github.com/eventbrite/m2crypto/blob/0f61819722c07ecffd7cf47c25095b516bf2d85c/M2Crypto/SMIME.py#L207, which is quite similar to test_support.pkcs7_verify.

I am willing to help, but since I am new to the cryptography project I would need some hints what you are expecting from such a public API that exceeds the complexity of M2Crypto's variant

alex:
That's great to hear re:migration. It's definitely a use case we want to support.

The big thing we'd be looking for in a public API is: Something that fits in place with the rest of our PKCS#7 APIs, something that is general enough to support many different verification use cases, and something that pushes users towards a secure result.

Internal APIs like `test_support` are really just the bare minimum to write tests, so they're not a good starting point for figuring out what our public APIs should look like. 

https://github.com/pyca/cryptography/pull/11555 is a recent example of working through the API design to add PKCS#7 decrpytion APIs.

zworkb:
Do you have in mind encapsuling the api in configurable objects as in x509 like 
[PyServerVerifier/rust](https://github.com/pyca/cryptography/blob/738598d1cf3d4f8cabc7730279b105846e4c5fd6/src/rust/src/x509/verify.rs#L347) and [ServerVerifyer/python](https://github.com/pyca/cryptography/blob/738598d1cf3d4f8cabc7730279b105846e4c5fd6/src/cryptography/hazmat/bindings/_rust/x509.pyi#L227) ?

Additionally I would appreciate for pointers which PKCS#7 verification usecases we should cover for starter.

nitneuqr:
> Do you have in mind encapsuling the api in configurable objects as in x509 like [PyServerVerifier/rust](https://github.com/pyca/cryptography/blob/738598d1cf3d4f8cabc7730279b105846e4c5fd6/src/rust/src/x509/verify.rs#L347) and [ServerVerifyer/python](https://github.com/pyca/cryptography/blob/738598d1cf3d4f8cabc7730279b105846e4c5fd6/src/cryptography/hazmat/bindings/_rust/x509.pyi#L227) ?
> 
> Additionally I would appreciate some pointers which PKCS#7 verification usecases we should cover for starter.

Hey, I've worked on the new `pkcs7_decrypt` features the past months. 

On my side, my approach was simple: develop something that mimics `test_support.pkcs7_decrypt` to remove the function test support function completely.

Since decryption is to encryption what verifying is to signing, we should probably adopt the same approach, then improve the function with new features afterwards (maybe `PKCS7_NO_VERIFY` ?)

alex:
Closing this, as https://github.com/pyca/cryptography/pull/12267 is the next steps here.

PR #12180, opened 2024-12-20
fix typing in ciphers.pyi

alex:


reaperhulk:
Can we add a test that mypy will fail if we screw this up again in the future?

alex:
mypy 1.14 catches this, see https://github.com/pyca/cryptography/pull/12179

reaperhulk (approved):

PR #12241, opened 2025-01-06
poetry 2.0 no longer has export, install it for certbot-josepy

reaperhulk:


.github/downstream.d/certbot-josepy.sh
```diff
@@ -6,6 +6,7 @@ case "${1}" in
         cd josepy
         git rev-parse HEAD
         curl -sSL https://install.python-poetry.org | python3 -
+        ~/.local/share/pypoetry/venv/bin/pip install poetry-plugin-export
```
alex, line 9:
```suggestion
        "${HOME}/.local/bin/poetry" self add poetry-plugin-export
```

seems to be the prefered way to write this?

alex (approved):

PR #12243, opened 2025-01-07
Update recommended PBKDF2HMAC iteration counts

MajorTanya:
This uses the current (as of 2025-01-07) default of 1,000,000 iterations used by Django in the [referenced code location](https://github.com/django/django/blob/40d5516385448a73426aad396778f369a363eda9/django/contrib/auth/hashers.py#L321).


alex (approved):
Thank you!

PR #12288, opened 2025-01-15
Move scripts used in GHA to a more sensible place

alex:


alex:
sob, didn't even notice. this is just a stock zed :-/

reaperhulk:
Close enough, most of the reformatting was removed 😄

PR #12289, opened 2025-01-15
Attempt to drop our usage of lcov and do things with coverage.py

alex:


alex:
> Can't combine statement coverage data with branch data

Hmm, not sure if/how we can address this.

alex:
https://github.com/nedbat/coveragepy/pull/1919 would allow cleaning this up some, but I'm not sure that should be a blocker for this

.github/bin/merge_rust_coverage.py
```diff
@@ -0,0 +1,100 @@
...
+def coverage_init(
+    reg: coverage.plugin_support.Plugins,
+    options: coverage.types.TConfigSectionOut,
+) -> None:
+    reg.add_file_tracer(RustCoveragePlugin())
+
+
+def main(*lcov_paths: str):
+    cov = coverage.Coverage()
+    # XXX: Nasty mucking in semi-public APIs
```
reaperhulk, line 34: We don't need to do this if https://github.com/nedbat/coveragepy/pull/1919 or something like it merges right?

alex: Correct

.github/bin/merge_rust_coverage.py
```diff
@@ -0,0 +1,100 @@
...
+            file_name: "coverage_rust_plugin.RustCoveragePlugin"
+            for file_name in covered_lines
+        }
+    )
+    coverage_data.write()
+
+    cov.combine()
+    coverage_percent = cov.report(show_missing=True)
+    if coverage_percent < 100:
+        print("+++ Python coverage under 100% +++")
```
reaperhulk, line 94:
```suggestion
        print("+++ Combined coverage under 100% +++")
```

reaperhulk (approved):

PR #12296, opened 2025-01-16
Migrate parsing of private keys to Rust

alex:


alex:
paramiko failures are due to https://github.com/paramiko/paramiko/pull/2490 (probably)

src/rust/src/backend/keys.rs
```diff
@@ -50,13 +65,88 @@ fn load_pem_private_key<'p>(
...
+    let tag = p.tag().to_string();
+    let data = match p.headers().get("Proc-Type") {
+        Some("4,ENCRYPTED") => {
+            password_used = true;
+            let Some(dek_info) = p.headers().get("DEK-Info") else {
+                todo!()
+            };
+            let Some((cipher_algorithm, iv)) = dek_info.split_once(',') else {
+                todo!()
+            };
```
Tarnum-tst, line 86: @alex , could You please make this code (81-86) compatible with Rust 1.63! let...else is unstable in 1.63... :-(

alex: We've had an MSRV of 1.65 since cryptography 43.0.0 in July of last year, we're not going to revert that absent exceptional circumstances

alex:
Test cases we need:

- [ ] Encrypted PKCS#8, PBES2 with 3DES-CBC
- [ ] Encrypted PKCS#8, PBES2 with unknown encryption algorithm
- [ ] Encrypted PKCS#8, PBES2 PBKDF2 with unknown PRF algorithm
- [ ] Encrypted PKCS#8, PBES2 with unknown KDF
- [ ] Encrypted PEM with no DEK-Info
- [ ] Encrypted PEM with DEK-Info that doesn't have two, comma separated components
- [ ] Encrypted PEM with DEK-Info where the IV isn't valid hex
- [ ] Unit tests for `hex_decode`

Things to verify:
- [ ] The `EXPLICIT_CURVE` check in `py_curve_from_curve` is unreachable
- [ ] The `UNSUPPORTED_ELLIPTIC_CURVE` check in `py_curve_from_curve` is unreachable

src/rust/src/backend/keys.rs
Tarnum-tst: @alex , I still maintain my custom build based on Rust 1.63 w/o GATs, maturin & workspace deps. It works fast on my retro systems (Win'7 and earlier) and very stable, too. :-)
Repacement:
```
            let dek_info = match p.headers().get("DEK-Info") {
                Some(dek_info) => dek_info,
                None => {
                    todo!()
                },
            };
            let (cipher_algorithm, iv) = match dek_info.split_once(',') {
                Some((cipher_algorithm, iv)) => (cipher_algorithm, iv),
                None => {
                    todo!()
                },
            };
```

alex: We're not going out of our way to to support things that we don't support.

src/rust/src/backend/keys.rs
```diff
@@ -34,7 +34,12 @@ pub(crate) fn load_der_private_key_bytes<'p>(
     password: Option<&[u8]>,
     unsafe_skip_rsa_key_validation: bool,
 ) -> CryptographyResult<pyo3::Bound<'p, pyo3::PyAny>> {
-    if let Ok(pkey) = openssl::pkey::PKey::private_key_from_der(data) {
+    let pkey = cryptography_key_parsing::pkcs8::parse_private_key(data)
```
reaperhulk, line 37: This nit doesn't really matter, but it seems probable that the least likely key type is DSA so having it be the first fallback is likely inefficient.

src/rust/src/backend/keys.rs
```diff
@@ -64,13 +65,110 @@ fn load_pem_private_key<'p>(
...
+                    return Err(CryptographyError::from(
+                        pyo3::exceptions::PyTypeError::new_err(
+                            "Password was not given but private key is encrypted",
+                        ),
+                    ))
+                }
+                Some(p) => p,
+            };
+
+            let cipher = match cipher_algorithm {
```
reaperhulk, line 107: Could we put a comment telling future readers why we only support these 3? (Just putting this here in addition to our chat)

reaperhulk (approved):

PR #12355, opened 2025-01-28
fixes #10808 -- allow empty plaintexts for aes-gcm-siv

### LINKED ISSUE #10808: AES_GCM_SIV does not accept empty plaintexts.
A test I have started failing recently when using AES-GCM-SIV with empty plaintexts.

The failing test is
```
from cryptography.hazmat.primitives.ciphers import aead
import cryptography
import sys

def test_empty_pt():
    key = bytes(range(16))
    nonce = bytes(range(12))
    pt = b""
    aad = b"associated data"
    crypter = aead.AESGCMSIV(key)
    ct = crypter.encrypt(nonce, pt, aad)

if __name__ == "__main__":
    print(f"{sys.version=}")
    print(f"{cryptography.__version__=}")
    test_empty_pt()

```
This gives the following output

sys.version='3.12.3 (tags/v3.12.3:f6650f9, Apr  9 2024, 14:05:25) [MSC v.1938 64 bit (AMD64)]'
cryptography.__version__='42.0.5'
  File "...pyca/aesgcmsiv.py", line 16, in <module>
    test_empty_pt()
  File "...pyca/aesgcmsiv.py", line 11, in test_empty_pt
    ct = crypter.encrypt(nonce, pt, aad)
ValueError: data must not be zero length

The exception is thrown for all supported key sizes of AES-GCM-SIV. No other AEAD mode seems to be affected. Other parameter sizes give the expected results. The test passed before updating to the Python version 3.12.3 and cryptography version 42.0.5.

alex:


reaperhulk:
Is master labeled 3.5 right now?

alex:
yes, you can look at the job to confirm if you want

reaperhulk (approved):

PR #12364, opened 2025-01-30
Update reference.rst:load_der_x509_certificate to note ValueError raised

ohemorange:
It raises the error and therefore should be noted in the documentation.

```
>>> x509.load_der_x509_certificate(b64.b64decode("foobar".encode()))
ValueError: error parsing asn1 value: ParseError { kind: InvalidLength }
```

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -223,6 +223,9 @@ Loading Certificates
 
     :returns: An instance of :class:`~cryptography.x509.Certificate`.
 
+    :raises ValueError: If there isn't at least one certificate, or if any
+        certificate is malformed.
+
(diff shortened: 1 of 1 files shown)


docs/x509/reference.rst
```diff
@@ -223,6 +223,9 @@ Loading Certificates
 
     :returns: An instance of :class:`~cryptography.x509.Certificate`.
 
+    :raises ValueError: If there isn't at least one certificate, or if any
+        certificate is malformed.
```
alex, line 227:
```suggestion
    :raises ValueError: If a certificate cannot be parsed from the provided
        data.
```

alex (approved):

alex:
thanks!

PR #12378, opened 2025-02-01
Minimal PoC for putting pyOpenSSL functionality in Rust

alex:


alex:
Problem: allows mutating the context in ways that are not sound.

Issue #12280, opened 2025-01-13
this is not the method chaining idiom you're looking for


cagney:
ref [Method Chining](https://en.wikipedia.org/wiki/Method_chaining)

The documentation for [Creating a self-signed certificate](https://cryptography.io/en/latest/x509/tutorial/#creating-a-self-signed-certificate) has the example:

```
cert = x509.CertificateBuilder().subject_name(
    subject
).issuer_name(
    issuer
).public_key(
    key.public_key()
).serial_nuber(
    x509.random_serial_number()
...
).sign(key, hashes.SHA256())
```

which sure looks a lot like method chaining where each call returns `self`.  When I read the documentation, for the individual functions, such as [CertificateBuilder.subnect_name()](https://cryptography.io/en/latest/x509/reference/#cryptography.x509.CertificateBuilder)

```
    Sets the subject’s distinguished name.
    Parameters:
        name – The [Name](https://cryptography.io/en/latest/x509/reference/#cryptography.x509.Name) that describes the subject.
```

the documentation states that the method will **set** the name, again consistent with method chaining.

However, when I try to invoke this code using the method chaining's alternative form:

```
cert_builder = x509.CertificateBuilder()
cert_builder.subject_name(subject)
```

things don't work.  It turns out that, rather than **setting** the existing builder, each of the above calls return a new instance of the builder.  Which means that:

```
cert_builder = x509.CertificateBuilder()
cert_builder = cert_builder.subject_name(subject)
...
```

must be used.

The only hint I've seen of this is in `CertificateBuilder` reference documentation which uses the above; but without explanation.

alex:
Yes, we should change the documentation to more clearly indicate this: "Returns a new builder with the certificate's subject to this value.", and with `:return:` annotations.

Would you be interested in submitting a pull request for this?

cagney:
I think it would be more useful to see the code made consistent with the idiom.

alex:
We think immutability is a better design pattern across the board, so we don't have an interest in changing it.

However, even if we did, it'd be wildly backwards incompatible and thus breaking of many downstream users, so it'd be difficult the justify.

tanmayghai18:
hope you folks don't mind, I took a stab and put up a PR to update the documentation: https://github.com/pyca/cryptography/pull/12386

PR #12386, opened 2025-02-03
Update CertBuilder docs

tanmayghai18:
As mentioned in https://github.com/pyca/cryptography/issues/12280, methods off of `cryptography.x509.base.CertificateBuilder` do not follow traditional "method chaining", rather each method will return a new builder instance with the corresponding updated value (depending on the method). 

This diff adds a note at the class level to highlight this + adds explicit `:return:` annotations to the various methods.

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -885,6 +885,11 @@ X.509 Certificate Builder
     :canonical: cryptography.x509.base.CertificateBuilder
 
     .. versionadded:: 1.0
+    
+    .. note::
+       All methods, except :meth:`sign`, return a **new** CertificateBuilder
+       instance with the corresponding updated value. They do not modify the
+       existing builder in place.
 
     .. doctest::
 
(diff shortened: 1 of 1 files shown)


docs/x509/reference.rst
```diff
@@ -990,6 +1006,8 @@ X.509 Certificate Builder
 
         :param critical: Set to ``True`` if the extension must be understood and
              handled by whoever reads the certificate.
+        
+        :return: A new :class:`CertificateBuilder` with the updated extension.
```
alex, line 1010:
```suggestion
        :return: A new :class:`CertificateBuilder` with the additional extension.
```

docs/x509/reference.rst
```diff
@@ -885,6 +885,10 @@ X.509 Certificate Builder
     :canonical: cryptography.x509.base.CertificateBuilder
 
     .. versionadded:: 1.0
+    
+    .. note::
+       Methods such as :meth:`issuer_name` return a **new** CertificateBuilder instance
+       with the corresponding updated value. They do not modify the existing builder in place.
```
alex, line 891:
```suggestion
       All methods, except :meth:`sign`, return a **new** CertificateBuilder
       instance with the corresponding updated value. They do not modify the
       existing builder in place.
```

alex (commented):
Two small suggested tweaks. Thanks for working on this!

alex (approved):
Thank you!

Issue #12388, opened 2025-02-03
Duplicate OID


vlaci:
Since https://github.com/pyca/cryptography/pull/10517, there is an `rsassaPss` public key algorithm OID name:

https://github.com/pyca/cryptography/blob/116c5af19dde67a043d4c4f5b82fde1b1ac330f2/src/cryptography/hazmat/_oid.py#L295

However, there is already a signature algorithm OID with a different name:

https://github.com/pyca/cryptography/blob/116c5af19dde67a043d4c4f5b82fde1b1ac330f2/src/cryptography/hazmat/_oid.py#L263

I don't know if it is intentional, that the names are spelled differently, but the one lower will overwrite the one above, so in both cases, the name will be `rsassaPss`.

alex:
No, it's not intentional. Nice catch.

Are you interested in submitting a PR, removing the duplicate entry?

vlaci:
> No, it's not intentional. Nice catch.
> 
> Are you interested in submitting a PR, removing the duplicate entry?

I'd be happy to. IMO the new name is the more correct, as we have long names e.g. `sha1WithRSAEncryption` instead of `RSA-SHA1`.

alex:
Sounds great!

PR #12389, opened 2025-02-03
Remove conflicting name of `rssassaPss` OID

vlaci:
The same OID is used as a signature algorithm and public key algorithm, so they have to be defined twice, but they should be added to `_OID_NAMES` only once.

Used the later introduced `rsassaPss` long name, instead of `RSASSA-PSS` short name, as other algorithms also use long names (i.e. `sha1WithRSAEncryption` instead of `RSA-SHA1`).

Resolves: #12388


src/cryptography/hazmat/_oid.py
```diff
@@ -292,7 +292,7 @@ class AttributeOID:
     PublicKeyAlgorithmOID.DSA: "dsaEncryption",
     PublicKeyAlgorithmOID.EC_PUBLIC_KEY: "id-ecPublicKey",
     PublicKeyAlgorithmOID.RSAES_PKCS1_v1_5: "rsaEncryption",
-    PublicKeyAlgorithmOID.RSASSA_PSS: "rsassaPss",
+    # PublicKeyAlgorithmOID.RSASSA_PSS: See SignatureAlgorithmOID.RSASSA_PSS
```
alex, line 295: Instead of commenting out, you can just remove the line.
```suggestion
```

alex (approved):

PR #12390, opened 2025-02-03
PyPolicyDefinition and related Verifier API changes

deivse:


deivse:
This is just the rust changes, I haven't adjusted the tests for the new API where things are accessible via `.policy` instead of directly, nor have I updated the `.pyi` files. I'll get to that tomorrow.

deivse:
Buut, this is a good time to discuss if we do want these things to be accessible only via the `policy` field, or to keep the old API alongside that for backwards-compatibility/convenience.  Not sure about that to be honest.

reaperhulk:
If they're just aliases then my opinion here is that we keep, but raise warnings and remove in 46. We have explicitly stated this isn't a stable API yet and is not subject to our backwards compatibility policy, but we do have consumers so we should give them a window for migration. @alex do you have an opinion?

alex:
What @reaperhulk said sounds fine.

deivse:
Any pointers on what to look at for implementing deprecation warnings for python getters in rust? I think I have seen something along these lines in the codebase but I have no idea where anymore. (Will update the PR tomorrow - it's already evening here in Prague)

reaperhulk:
https://github.com/pyca/cryptography/blob/d860aaa54305c2089b1bbf49a3afadabf90c0e14/src/rust/src/x509/crl.rs#L262-L264 should point you the right direction

deivse:
Exactly what I was looking for, thank you! Will report back tomorrow.

deivse:
I decided to rename `PyPolicyDefinition` to `PyPolicy`. The original name was to better reflect that it holds a `PolicyDefinition`, not a `Policy`, but I think I prefer to maintain the convention of `PyXXX` name directly corresponding to `XXX` in python.

PS: Will add the deprecated getters back in in a bit.

src/cryptography/hazmat/bindings/_rust/x509.pyi
```diff
@@ -206,6 +206,18 @@ class PolicyBuilder:
         self, subject: x509.verification.Subject
     ) -> ServerVerifier: ...
 
+class Policy:
+    @property
+    def max_chain_depth(self) -> int: ...
+    @property
+    def subject(self) -> x509.verification.Subject: ...
```
alex, line 213: This is `Subject | None`, right? A client validator doesn't have one?

src/rust/src/x509/verify.rs
```diff
@@ -11,21 +11,24 @@ use cryptography_x509_verification::trust_store::Store;
 use cryptography_x509_verification::types::{DNSName, IPAddress};
 use pyo3::types::{PyAnyMethods, PyListMethods};
 
+mod py_policy;
```
alex, line 14: Let's rename this file to `verify/mod.rs`, I find the presence of both `verify.rs` and a `verify/` diretory confusing.

I think this `mod` should also just be `policy`, not `py_policy`.

src/rust/src/x509/verify.rs
```diff
@@ -433,6 +463,9 @@ fn build_subject<'a>(
 
             Ok(Subject::IP(ip_addr))
         }
+        SubjectOwner::None => Err(pyo3::exceptions::PyRuntimeError::new_err(
+            "subject must be set for ServerVerifiers",
+        )),
```
alex, line 468: Hmm, this should be unreachable from the Python API, right?

alex: If yes, this can just be a panic.

src/rust/src/x509/verify/py_policy.rs
```diff
@@ -0,0 +1,71 @@
...
+
+#[pyo3::pymethods]
+impl PyPolicy {
+    #[getter]
+    pub(super) fn max_chain_depth(&self) -> u8 {
+        self.0.borrow_dependent().max_chain_depth
+    }
+
+    #[getter]
+    pub(super) fn subject(&self, py: pyo3::Python<'_>) -> pyo3::PyResult<pyo3::PyObject> {
```
alex, line 25: Previously we just kept the original subject object around, any reason not to do that here?

src/cryptography/hazmat/bindings/_rust/x509.pyi
deivse: Yes, thx!

src/rust/src/x509/verify.rs
deivse: Up to you - I went this way since I am slightly bothered by having many identically named mod.rs files in a project (makes finding the right one slightly harder). The reasoning behind `py_policy` was the same. But this is totally a matter of preference, so I'll just change it.

src/rust/src/x509/verify.rs
deivse: Take a look and tell me if u prefer it that way. If not I can revert back to old behaviour with a panic instead of returning pyerror.

src/rust/src/x509/verify/py_policy.rs
deivse: No reason at all, good call. Implemented in last commit. I'll resolve, feel free to re-open if there are any issues with my impl.

alex (approved):

PR #12410, opened 2025-02-06
Bump versions in wheel builder pins

alex:
For some reason dependabot isn't handling them properly


alex:
Needs GHA to bump its PyPy version for the wheel builder.

reaperhulk (approved):

PR #12417, opened 2025-02-08
ExtensionPolicy changes required for Python callback support.

deivse:
Should be merged after #12416, will rebase if there are any changes there. Creating as draft for now.

This PR changes the validator callback types to allow storing python callbacks in them. 
The callback types have to be `Send` and `Sync` to allow the `ExtensionPolicy` to be stored inside a `PyExtensionPolicy` pyclass. They are  `Arc` to be able to clone the `ExtensionPolicy`, which is required for the builder-style API of `PyExtensionPolicy`.

Hopefully this makes sense on it's own. I was originally going to include this as part of a larger PR adding all the functionality required to expose extension policies to python, but that was getting too big.

### DIFF
--- a/src/rust/cryptography-x509-verification/src/lib.rs
+++ b/src/rust/cryptography-x509-verification/src/lib.rs
@@ -463,7 +463,7 @@ impl<'a, 'chain, B: CryptoOps> ChainBuilder<'a, 'chain, B> {
         let leaf_extensions = leaf.certificate().extensions()?;
 
         self.policy
-            .permits_ee(leaf.certificate(), &leaf_extensions)
+            .permits_ee(leaf, &leaf_extensions)
             .map_err(|e| e.set_cert(leaf.clone()))?;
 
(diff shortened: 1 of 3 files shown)


deivse:
The just added VerificationCertificate change is so that extension validation callbacks have access to CertificateExtra. Specifically so that the original python certificate object is accessible when invoking a python extension validation callback. Originally I used a pretty awful hack, creating a PyCertificate from the rust certificate by serializing and deserializing from asn1.

alex:
Just wanted to give you a huge thank you for your work splitting this apart, it's making things much easier to review.

deivse:
No worries! Not sure why Github still shows the first commit as part of this PR though, since it's the same one as in the one you just merged

deivse:
Oh I get it, it's because you don't use fast-forward. I'll force push this in a second then.

alex:
Yeah, if you rebase on main it should do the right thing.

alex (approved):

PR #12432, opened 2025-02-11
x509 Verification: base Python support for extension policies

deivse:

PR #12433, opened 2025-02-11
x509 Verification: Python ExtensionPolicy builder API impl.

deivse:
~~Prepared for after #12432 is merged.~~

### DIFF
--- a/src/rust/src/x509/verify/extension_policy.rs
+++ b/src/rust/src/x509/verify/extension_policy.rs
@@ -23,9 +23,7 @@ use crate::asn1::py_oid_to_oid;
 
 use crate::types;
 use crate::x509::certificate::parse_cert_ext;
-use crate::x509::certificate::Certificate as PyCertificate;
 
-use super::policy::PyPolicy;
 use super::PyCryptoOps;
 
 #[pyo3::pyclass(
@@ -204,8 +202,15 @@ fn wrap_maybe_validator_callback(
               cert: &VerificationCertificate<'_, PyCryptoOps>,
               ext: Option<&Extension<'_>>| {
             pyo3::Python::with_gil(|py| {
-                let args = make_python_callback_args(py, policy, cert, ext)?;
-                invoke_py_validator_callback(py, &py_cb, args)
+                invoke_py_validator_callback(
+                    py,
+                    &py_cb,
+                    (
+                        policy.extra.clone_ref(py),
+                        cert.extra().clone_ref(py),
+                        make_py_extension(py, ext)?,
+                    ),
+                )
             })
         },
     )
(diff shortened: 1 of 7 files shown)


src/rust/src/x509/verify/extension_policy.rs
```diff
@@ -0,0 +1,274 @@
...
+    policy: &Policy<'_, PyCryptoOps>,
+    cert: &VerificationCertificate<'chain, PyCryptoOps>,
+    ext: Option<&Extension<'p>>,
+) -> ValidationResult<'chain, PyCallbackArgs<'p>, PyCryptoOps> {
+    let py_policy = policy.extra.clone_ref(py);
+    let py_cert = cert.extra().clone_ref(py);
+    let py_ext = match ext {
+        None => None,
+        Some(ext) => parse_cert_ext(py, ext).map_err(|e| {
+            make_validation_error(format!("{e} (while converting Extension to Python object)"))
```
deivse, line 250: Copying over from previous PR as I'm closing that to declutter:

Do you maybe have any ideas on how to handle this? This should be the last part that doesn't have coverage. I'm struggling to come up with a way to test this in a sane way, or to prove that this can not occur (and thus make this an unwrap). Suggestions much appreciated.

deivse:
Rebased on main.

src/rust/src/x509/verify/extension_policy.rs
deivse: I am starting to lean towards thinking that having an unwrap there should be fine. The extensions are guaranteed to have successfully been parsed from asn1 at this point, and the function parse_cert_ext function supports a superset of all extensions that an ExtensionPolicy can have validators for. So these are the only reasons I can see for parse_cert_ext to fail:
- Logic error in `parse_cert_ext` itself
- Some critical global condition which would make pyo3 functions in `parse_cert_ext` fail, e.g. out of memory.

Let me know if that makes sense to you and if you think making this an `unwrap` is fine.

deivse: Another part of me was saying it's risky to have an unwrap that deep in the callstack that is practically relying on  implementation details of the cryptography-x509-verification crate, so I'm trying a refactor to make the test saner.

src/rust/src/x509/verify/extension_policy.rs
```diff
@@ -36,4 +144,152 @@ impl PyExtensionPolicy {
...
+                    ),
+                )
+            })
+        },
+    )
+}
+
+fn make_validation_error(msg: String) -> ValidationError<'static, PyCryptoOps> {
+    ValidationError::new(ValidationErrorKind::Other(msg))
+}
```
alex, line 243: This isn't doing much, please inline into the callers

src/rust/src/x509/verify/extension_policy.rs
alex: Yeah, I think strictly speaking errors can happen here for like `MemoryError` and such.

alex (approved):

PR #12451, opened 2025-02-13
Refresh `ci-constraints-requirements.txt`

alex:


ci-constraints-requirements.txt
```diff
@@ -113,7 +117,7 @@ markupsafe==3.0.2 ; python_full_version >= '3.9'
     # via jinja2
 mypy==1.4.1 ; python_full_version < '3.8'
     # via cryptography (pyproject.toml)
-mypy==1.15.0 ; python_full_version == '3.8.*'
+mypy==1.14.1 ; python_full_version == '3.8.*'
```
reaperhulk, line 120: Why did this go down?

alex: Because the old one was wrong, but we never noticed because we don't test mypy on an old Python.

If you look below, you'll see 1.15 requires >=3.9

reaperhulk (approved):

PR #12455, opened 2025-02-14
ServerVerifier incorrect ext. policy early fail.

deivse:
Decided to do this in it's own PR since #12454 is already not that small. Will rebase on main once #12454 is merged.

### DIFF
--- a/src/rust/cryptography-x509-verification/src/policy/extension.rs
+++ b/src/rust/cryptography-x509-verification/src/policy/extension.rs
@@ -986,11 +986,8 @@ mod tests {
             &VerificationCertificate::new(&cert, ()),
             &extensions,
         );
-        let err = result.expect_err(
-            "ee_ext_policy must not permit certificate with no SAN if policy.subject is set",
-        );
         assert_eq!(
-            err.to_string(),
+            result.unwrap_err().to_string(),
             "leaf server certificate has no subjectAltName"
         );
     }
--- a/src/rust/cryptography-x509-verification/src/policy/extension.rs
+++ b/src/rust/cryptography-x509-verification/src/policy/extension.rs
@@ -702,10 +702,11 @@ mod tests {
     use cryptography_x509::extensions::{BasicConstraints, Extension};
     use cryptography_x509::oid::BASIC_CONSTRAINTS_OID;
 
-    use super::{Criticality, ExtensionValidator};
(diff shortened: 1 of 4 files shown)


deivse:
Rebased on main

src/rust/cryptography-x509-verification/src/policy/mod.rs
```diff
@@ -247,8 +247,8 @@ impl<'a, B: CryptoOps + 'a> PolicyDefinition<'a, B> {
         extended_key_usage: ObjectIdentifier,
         ca_extension_policy: Option<ExtensionPolicy<'a, B>>,
         ee_extension_policy: Option<ExtensionPolicy<'a, B>>,
-    ) -> Self {
-        Self {
+    ) -> Result<Self, &'static str> {
```
deivse, line 250: Not sure if some specific error type should be used here or if this is fine. None of the existing error types seemed to fit this usage.

src/rust/cryptography-x509-verification/src/policy/mod.rs
```diff
@@ -308,6 +310,28 @@ impl<'a, B: CryptoOps + 'a> PolicyDefinition<'a, B> {
...
+
+        match self.ee_extension_policy.subject_alternative_name {
+            ExtensionValidator::Present {
+                criticality: _,
+                validator: _,
+            } => Ok(()),
+            _ => Err(
+                "An EE extension policy used for server verification must require the subjectAltName extension to be present.",
+            ),
+        }
```
alex, line 333: You can write this as `if !matches!(self.ee_extension_policy.subject_alternative_name, ExtensionValidator::Present { .. }) { }`

src/rust/cryptography-x509-verification/src/policy/extension.rs
```diff
@@ -950,4 +956,42 @@ mod tests {
...
+            &VerificationCertificate::new(&cert, ()),
+            &extensions,
+        );
+        let err = result.expect_err(
+            "ee_ext_policy must not permit certificate with no SAN if policy.subject is set",
+        );
+        assert_eq!(
+            err.to_string(),
+            "leaf server certificate has no subjectAltName"
+        );
```
alex, line 995: `unwrap_err()` is preferrable here, since we know that case cannot be reached.

alex (approved):

PR #12458, opened 2025-02-14
docs: Add warning about ext. policies and ServerVerifier.

deivse:


deivse:
@alex regarding the changelog entry, should I write that? How verbose should it be, should it for example list every class/method added, every deprecated and new property (ClientVerifier and ServerVerifier), etc., or can it be more generic and refer people to the documentation for details, e.g. "Added API for customizing x509 verification behaviour via user-defined extension policies."

deivse:
PS: I guess including a full list of deprecated properties should be done in any case.

alex:
More general is fine: "Added support for configuring extension handling in X.509 certificate validation" or something, with a link to the `PolicyBuilder` method.

alex (approved):

PR #12476, opened 2025-02-16
Make `PolicyBuilder.extension_policies` keyword-only.

deivse:
https://github.com/pyca/cryptography/pull/12432#discussion_r1957088411

I decided to remove the `new_` prefix from the argument names, because that is a bit too annoying to write out for me. There is a slight inconsistency now with how the other arguments are named, but I can personally live with that.
If we do want to keep it consistent, I would rather remove `new_` from all `PolicyBuilder` argument names than force users to write out `new_ee_policy` and `new_ca_policy` every time 😅 But that's just my personal preference of course 😄


deivse:
Regarding the naming, I just searched through the code quickly, and I don't think any other builder APIs use the `new_xxx` naming scheme, so I guess removing the `new_` prefix from all parameter names would actually improve overall consistency. I'm going to do that preliminary, and I can revert if you don't agree.

alex (approved):

PR #12505, opened 2025-02-24
Enable strict bytes. This causes mypy to treat bytes as bytes, and not as a hodgepodge of buffer things

alex:
Then fix the 9000 places we violate this. I only updated things we had tests for, for better or worse.


src/cryptography/hazmat/bindings/_rust/openssl/poly1305.pyi
```diff
@@ -2,12 +2,14 @@
...
 class Poly1305:
-    def __init__(self, key: bytes) -> None: ...
+    def __init__(self, key: Buffer) -> None: ...
     @staticmethod
-    def generate_tag(key: bytes, data: bytes) -> bytes: ...
+    def generate_tag(key: Buffer, data: bytes) -> bytes: ...
     @staticmethod
-    def verify_tag(key: bytes, data: bytes, tag: bytes) -> None: ...
-    def update(self, data: bytes) -> None: ...
+    def verify_tag(key: Buffer, data: bytes, tag: bytes) -> None: ...
```
reaperhulk, line 12:
```suggestion
    def verify_tag(key: Buffer, data: Buffer, tag: bytes) -> None: ...
```

reaperhulk (approved):
I'm sure we'll find some more, but 👍

PR #12521, opened 2025-02-26
fixes #10958 -- support aes-siv with empty plaintext on latest openssl

### LINKED ISSUE #10958: AES-SIV does not accept empty plaintext.
This issue is possibly related to #10808, which observes a similar behavior for AES-GCM-SIV.

I'm using python version 3.12.3
and  cryptography version 42.0.5

The following code
```
from cryptography.hazmat.primitives.ciphers import aead
import cryptography
import sys

def test_empty_pt():
    key = bytes(range(32))
    nonce = bytes(range(12))
    pt = b""
    aads = [b"row1", b"col2", b"timestamp"]
    crypter = aead.AESSIV(key)
    ct = crypter.encrypt(pt, aads)

if __name__ == "__main__":
    print(f"{sys.version=}")
    print(f"{cryptography.__version__=}")
    test_empty_pt()

```
gives the following result:

> 
> sys.version='3.12.3 (tags/v3.12.3:f6650f9, Apr  9 2024, 14:05:25) [MSC v.1938 64 bit (AMD64)]'
> cryptography.__version__='42.0.5'
> Traceback (most recent call last):
>   File ".../aes_siv_test.py", line 16, in <module>
>     test_empty_pt()
>   File ".../aes_siv_test.py", line 11, in test_empty_pt
>     ct = crypter.encrypt(pt, aads)
> ValueError: data must not be zero length
> 
>

The only restriction I could find is RFC 5297, section 6, where there is a lower bound of 1 byte for the nonce, when AES-SIV is being used as AEAD. However, AES-SIV is useful in many contexts, e.g., deterministically encrypting cells in a database using row and column as AAD. For such uses it would be helpful if all fields can be empty.

alex:


alex:
Hmm, is something screwed up? This test case is right out of wycheproof...

alex:
(Sigh, difference between `ad=[]` and `ad=[b""]`.

reaperhulk (approved):

PR #12534, opened 2025-02-28
enable test contexts

reaperhulk:
this allows us to see what tests exercised what lines of code


alex:
is this for review? does it work? what's the interface in coverage like?

reaperhulk:
Yes, it works fine. You can pull and run locally to see coverage or we could change coverage to always upload the zip I guess. It looks like a disclosure option on each line with a list of tests that hit the line.

alex (approved):

Issue #12488, opened 2025-02-19
Update documentation to clearly state that Fernet is thread safe


dan-corneanu:
I recently used Fernet in a multithreaded environment and struggled to determine from the documentation whether it is thread-safe.

After reviewing the implementation, it appears that Fernet is indeed thread-safe:
[Source Code](https://github.com/pyca/cryptography/blob/e6eb2579373f6d1d06f2cf2b3c1c29cbdaae8f09/src/cryptography/fernet.py#L28)

Would it be appropriate to update the documentation to clarify this?

alex:
Every immutable object in cryptography is safe to share between multiple threads. (Mutable objects, like hash contexts, can also safely be shared, but concurrent mutation will produce indefinite results or raise an exception).

This is by way of saying: this isn't specific to fernet, so I think if we want to add something to the docs, we probably want to make it more general purpose.

Are you interested in submitting a PR?

dan-corneanu:
I'll submit a PR for the documentation.

alex:
Great, thanks.

ngoldbaum:
I'm also happy to help out with this in the context of free-threaded Python.

dan-corneanu:
Could you provide some guidance on:
- The best section in the documentation to amend with this additional information.
- The process for generating the documentation in my local development environment.

PR #12550, opened 2025-03-04
docs: clarify thread safety of immutable objects

dan-corneanu:
This PR is trying to address https://github.com/pyca/cryptography/issues/12488

### DIFF
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -107,6 +107,15 @@ Glossary
         This is a term used to describe an operation where the user must
         ensure that the input is correct. Failure to do so can result in
         crashes, hangs, and other security issues.
+    
+    thread safety
+        All immutable objects in ``cryptography`` are safe to use in
+        multi-threaded environments. This means they can be shared across
+        threads without requiring additional synchronization. Mutable objects,
+        such as hash contexts, can also be shared, but concurrent modification
+        may lead to exceptions or incorrect results. When working with
+        cryptographic operations in a multi-threaded application, ensure that
+        any mutable objects are used in a thread-safe manner.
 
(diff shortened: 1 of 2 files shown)


docs/index.rst
```diff
@@ -45,6 +45,16 @@ See :doc:`Installation <installation>` for more information.
 .. _cryptography-layout:
 
 
+Thread Safety
```
alex, line 48: You asked where the right place for this would be: My suggestion is adding a new term in the glossary, which we can then reference in various places. (e.g., you asked about fernet, so we can add a reference there).

alex:
Are you interested in completing this?

docs/index.rst
dan-corneanu: Just moved the section to the glossary. Is this what you had in mind?

docs/glossary.rst
```diff
@@ -107,6 +107,15 @@ Glossary
         This is a term used to describe an operation where the user must
         ensure that the input is correct. Failure to do so can result in
         crashes, hangs, and other security issues.
+    
+    thread safety
+        All immutable objects in ``cryptography`` are safe to use in
+        multi-threaded environments. This means they can be shared across
+        threads without requiring additional synchronization. Mutable objects,
+        such as hash contexts, can also be shared, but concurrent modification
+        may lead to undefined behavior or exceptions. When working with
```
alex, line 116: Let's avoid the phrase undefined behavior, as it implies the possibility of security issues.

How about "exceptions or incorrect results."

dan-corneanu: Updated

alex (approved):
Thanks!

PR #12590, opened 2025-03-07
port ANSIX923 padding and unpadding contexts to Rust

### LINKED ISSUE #12553: `_ANSIX923PaddingContext` isn't thread-safe
While writing tests in support of #12489, I ran across a case where the GIL is implicitly locking some data stored in Python objects, but in a manner that can be worked around, even with the GIL. 

Consider the following script, which updates a padder in multiple threads in a random order using a different chunk size for each thread. The data are chosen such that the final state of the padder should be the same no matter which order the threads updated it.

```python
import sys
import threading

from cryptography.hazmat.primitives.padding import ANSIX923

# force threads to switch more quickly to make the race more likely
sys.setswitchinterval(.0000001)

num_threads = 4
padder = ANSIX923(num_threads * 256).padder()
validate_padder = ANSIX923(num_threads * 256).padder()
chunk = b"abcd1234"
data = chunk * 16
validate_padder.update(data * num_threads)
expected_pad = validate_padder.finalize()

b = threading.Barrier(num_threads)

def pad_in_chunks(chunk_size):
    index = 0
    b.wait()
    while index < len(data):
        padder.update(data[index : index + chunk_size])
        index += chunk_size

threads = []
for threadnum in range(num_threads):
    chunk_size = len(data) // (2**threadnum)
    thread = threading.Thread(target=pad_in_chunks, args=(chunk_size,))
    threads.append(thread)

for thread in threads:
    thread.start()
for thread in threads:
    thread.join()

calculated_pad = padder.finalize()

assert expected_pad == calculated_pad, (expected_pad, calculated_pad)
```

On My M3 Macbook Pro with the standard build of CPython 3.13.2, this test sometimes runs to completion and sometimes the final assert fails with errors like:

```
goldbaum at Nathans-MBP in ~/Documents/test
$  python test.py
  File "/Users/goldbaum/Documents/test/test.py", line 39, in <module>
    assert expected_pad == calculated_pad, (expected_pad, calculated_pad)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: (b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x80', b'abcd1234abcd1234abcd1234abcd1234abcd1234abcd1234abcd1234abcd1234abcd1234abcd1234abcd1234abcd1234abcd1234abcd1234\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x10')
```

This code is adapted from the `test_threaded_hashing` test in CPython's hashlib - I've already written a similar test for the hash digest context and that worked fine.

This is happening because `_ANSIX923PaddingContext` stores its state in a bytestring and since updates to the bytestring aren't atomic, if the thread switch interval is short enough (or someone gets really unlucky in the default configuration) then there is a race to update the `_buffer`.

https://github.com/pyca/cryptography/blob/0ce9f53c9842c1d279ae5323ee42a0bd1899d464/src/cryptography/hazmat/primitives/padding.py#L137-L159

One solution is to add a `threading.lock` in the python implementation. Another is to move the code into Rust, where runtime borrow checking would turn this silently incorrect result into a runtime borrow-check error.

ngoldbaum:
Fixes #12553.

Also adds multithreaded tests for the padding and unpadding algorithms along with some test infrastructure to support that.


src/rust/src/padding.rs
```diff
@@ -42,7 +42,6 @@ pub(crate) fn check_pkcs7_padding(data: &[u8]) -> bool {
     (mismatch & 1) == 0
 }
 
-#[pyo3::pyfunction]
 pub(crate) fn check_ansix923_padding(data: &[u8]) -> bool {
```
alex, line 45: I don't think it needs to be `pub(rate)` anymore?

src/rust/src/padding.rs
```diff
@@ -110,6 +109,62 @@ impl PKCS7PaddingContext {
...
+        }
+    }
+
+    pub(crate) fn update<'p>(
+        &mut self,
+        buf: CffiBuf<'p>,
+        py: pyo3::Python<'p>,
+    ) -> CryptographyResult<pyo3::Bound<'p, pyo3::types::PyBytes>> {
+        match self.buffer.as_mut() {
+            Some(v) => {
```
alex, line 134: I don't think any of this is necessary, if you look at pkcs7padder, we can jsut return the buffer and count the length when padding.

src/rust/src/padding.rs
```diff
@@ -110,6 +109,62 @@ impl PKCS7PaddingContext {
...
+                let pad_size = (self.block_size).saturating_sub(v.len());
+                let mut count = 0;
+                v.extend(std::iter::from_fn(move || {
+                    count += 1;
+                    match count.cmp(&pad_size) {
+                        std::cmp::Ordering::Less => Some(0),
+                        std::cmp::Ordering::Equal => Some(pad_size as u8),
+                        std::cmp::Ordering::Greater => None,
+                    }
+                }));
```
alex, line 160: I think  `for _ in 0..(pad_size - 1) { v.push(0) }; v.push(pad_size);` would be simpler and easier to read.

src/rust/src/padding.rs
```diff
@@ -169,6 +224,67 @@ impl PKCS7UnpaddingContext {
...
+    pub(crate) fn update<'p>(
+        &mut self,
+        buf: CffiBuf<'p>,
+        py: pyo3::Python<'p>,
+    ) -> CryptographyResult<pyo3::Bound<'p, pyo3::types::PyBytes>> {
+        match self.buffer.as_mut() {
+            Some(v) => {
+                v.extend_from_slice(buf.as_bytes());
+                let finished_blocks =
+                    std::cmp::max((v.len() / self.block_size).saturating_sub(1), 0);
```
alex, line 252: Is a negative value possible here? If not, why do we have the `max()` call? (If it is possible, we have the same bug in pkcs7unpadder I guess)

tests/hazmat/primitives/test_padding.py
```diff
@@ -240,3 +247,96 @@ def test_bytearray(self):
         unpadder = padding.ANSIX923(128).unpadder()
         final = unpadder.update(padded) + unpadder.finalize()
         assert final == unpadded + unpadded
+
```
alex, line 250: I'm very ambivalent about the utility of these tests.

Part of the benefit of moving stuff to Rust is that we can have confidence that these are correct by construction. As is these are basically tests that pyo3's borrow checking code is correct.

src/rust/src/padding.rs
ngoldbaum: This is a direct port of this line in the Python implementation:

https://github.com/pyca/cryptography/blob/46a8693ad634d7a2a606a4cbb0ee759147dd911c/src/cryptography/hazmat/primitives/padding.py#L81

I'll think about whether or not the `max` was ever really necessary.

alex: python doesn't have saturating sub :-)

tests/hazmat/primitives/test_padding.py
ngoldbaum: If you're comfortable with (eventually) shipping free-threaded wheels without explicit multithreaded tests for mutable primitives implemented in rest, I'm happy to delete the new tests.

alex: That's my strong inclination yes. (Ordinarily I'd bug @reaperhulk for a second opinion, but he just had a kid so I'm not going to nag him in too many reviews.)

Let me mull on this for a day or two before we make any decisions.

tests/hazmat/primitives/test_padding.py
```diff
@@ -172,9 +179,9 @@ def __str__(self):
 
         str(mybytes())
         padder = padding.ANSIX923(128).padder()
-        padder.update(mybytes(b"abc"))
+        data = padder.update(mybytes(b"abc")) + padder.finalize()
         unpadder = padding.ANSIX923(128).unpadder()
-        unpadder.update(mybytes(padder.finalize()))
+        unpadder.update(mybytes(data))
```
ngoldbaum, line 184: It looks like a similar change had to happen back in da45641e46 for the PKCS7 padder.

src/rust/src/padding.rs
ngoldbaum: Did a slight variation on this: create a zero-filled vector and then do `v.push(pad_size)`

tests/hazmat/primitives/test_padding.py
alex: Yeah, after some review, I don't think we should carry tests that are really just tests of pyo3's invariants.

alex (approved):

PR #12591, opened 2025-03-09
feat(repo): Be more strict in the cffi compat

do3cc:
Testing uv in aother project with the option
`--resolution lowest`
showed that current python versions would not work at all with the minimum required version of cffi, 1.12.

This has very likely no impact on production systems, but it empowers other projects to validate that the minimum requirements of themselves and their dependencies are correct.

I have tested these version specifiers by fiddling with pyproject.toml and uv sync.
As soon as the Python version compiled with the minor version, I updated the dependency specification.
I have not run the nox tests. It didn´t work, and I started to do something on my blog end ended up writing PRs to the cryptography project. I am trying to force myself to limit my yak shaving at least a bit.

I have tried to figure out, why I get compilation errors with, for example, python 3.10 and cffi 1.13, but looking for the error I got in the cffi repo, I did not find any results. I presume that this means that the problem I am solving does not really exist in production installations.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ requires = [
     "maturin>=1,<2",
 
     # Must be kept in sync with `project.dependencies`
-    "cffi>=1.12; platform_python_implementation != 'PyPy'",
+    "cffi>=1.14; platform_python_implementation != 'PyPy'",
     # Used by cffi (which import distutils, and in Python 3.12, distutils has
     # been removed from the stdlib, but installing setuptools puts it back) as
     # well as our build.rs for the rust/cffi bridge.
@@ -50,7 +50,7 @@ classifiers = [
 requires-python = ">=3.7,!=3.9.0,!=3.9.1"
 dependencies = [
     # Must be kept in sync with `build-system.requires`
-    "cffi>=1.12; platform_python_implementation != 'PyPy'",
+    "cffi>=1.14; platform_python_implementation != 'PyPy'",
 ]
 
 [project.urls]


alex:
I don't understand the motivation here, we should work with cffi 1.12 on every version of python, right?

do3cc:
Hi, I am not 100% sure if I understood your question. If think you meant all python versions should work with cffi 1.12.0.

That would be good, but is unfortunately not the case.

A simple example with just pip in a docker container shows how it is failing:
```bash
docker run -it --rm python:3.13 pip install cryptography cffi==1.12.0
```
Searching for the failure, I see this bugfix: https://foss.heptapod.net/pypy/cffi/-/issues/458

However, I can´t reproduce the problem with python 3.13 and cffi <= 1.16. So I can reduce the complexity of the PR.

Ideally, CFFI would declare what version of python they explicitly support in each version. But if they´d change that, it would only come into effect in 1.18.0.

With regards to motivation, I would like to use a feature of uv, that allows me to test a package with the oldest supported versions of each library. That mostly works surprisingly well, cryptography with cffi>=1.12.0 being one of the few exceptions.
This change wil most likely not improve any actual deployment, but help other projects who want to test their minimum supported versions.

As a temporary workaround, I will put a cffi dependency into the project itself, but this is just a kludge.

If that makes sense and you see the value of this, I'll update the PR to remove the cffi>=1.16 dependency for Python 3.13. That seems to be a variant of "works on my machine" issue.

alex:
Hmm, I see what you mean.

It feels like this ought to be handled by cffi's own metadata, though I also recognize it's not (and there's reasonable reasons for that).

I'm going to mull on this for a bit. On its face it seems fine, but this also implies that _every_ cffi consumer needs to carry version bounds like this, and that can't be right.

do3cc:
I started to write a PR for CFFI to add the upper bound and found this article against upper bounds:
https://iscinumpy.dev/post/bound-version-constraints/#pinning-the-python-version-is-special
It also points to a long discussion on discourse. The end result is that some packages would prefer to specify an upper bound, but the official documentation of pyproject.toml slightly discourages from this.

I've stepped away a bit and considered this: There are not multiple branches of CFFI being supported or maintained. If you want to use CFFI and you want to support Python 3.10 and newer. You must use CFFI 1.14 or newer. So it may be reasonable to just update the minimum requirement to 1.14.0 and be done with it.

Also, if CFFI would start to define upper bounds on python, they would have to do releases more often than they do right now, so it is an additional burden for them.

alex:
Yeah, just bumping up to 1.14 as the minimum is fine with me, since it doesn't look like that drops support for anything.

alex (approved):

PR #12601, opened 2025-03-11
fixes #11450 -- warn on legacy provider load

### LINKED ISSUE #11450: OpenSSL legacy provider isn't always installed on Debian unstable
This is an attempt to get a conversation started.  I wasn't responsible for the original change here, but I've been cleaning up after it in Debian's Python team and I'm not convinced my current approach is really ideal.

The OpenSSL packages in Debian unstable recently [changed](https://bugs.debian.org/965041) to move the legacy provider into a separate package, `openssl-provider-legacy`.  There's a `Recommends` relationship from the main library package (`libssl3t64`) which means that it's installed on typical user systems, but it's not a `Depends` so it's now possible for users not to have the legacy provider installed.  In particular, `Recommends` are not normally installed in package builds and automatic test environments (deliberately, to ensure that declared dependencies are sufficient), and as a result we find ourselves having to apply workarounds to a fairly large number of Python packages in Debian to get their tests working again.

I'd like to find out what approach you'd recommend here.  I can think of a bunch of possibilities:

* Just set `CRYPTOGRAPHY_OPENSSL_NO_LEGACY=1` in the Debian build rules for lots of Python packages (this is pretty much what I'm doing at the moment, but it's repetitive)
* Make affected Python packages in Debian depend/build-depend on `openssl-provider-legacy` (also repetitive and feels heavy-handed)
* Make Debian's `python3-cryptography` package depend on `openssl-provider-legacy`
* Effectively revert the Debian OpenSSL packaging change (perhaps just by upgrading the `Recommends` to `Depends`)
* Change the logic in `cryptography/src/rust/src/lib.rs` in some way
* Something else I haven't thought of

Do you have any opinions on this?  Ideally I wouldn't be passing messages back and forward, so I'll see if I can get Debian's OpenSSL maintainers to show up on this issue.

alex:
enable disabling the legacy provider entirely


alex:
(PR generated almost entirely with claude-code)

reaperhulk:
This adds a job to trigger the warning when it can't load legacy, do we need a job that validates `CRYPTOGRAPHY_BUILD_OPENSSL_NO_LEGACY` doesn't attempt to load? (I want the answer to be no since adding entire CI jobs for this stuff is not a path to performant CI 😆)

PR #12618, opened 2025-03-14
12612 Conditional cipher registry install for RC4

szelenka:
This is a patch to only add the RC4 cipher when/if the OpenSSL library being linked against supports it.

Addresses issue #12612


src/rust/src/backend/cipher_registry.rs
```diff
@@ -298,7 +299,9 @@ fn get_cipher_registry(
                 m.add(&idea, &cfb, Some(128), Cipher::idea_cfb64())?;
             }
 
-            m.add(&arc4, none_type.as_any(), None, Cipher::rc4())?;
+            #[cfg(not(CRYPTOGRAPHY_OSSLCONF = "OPENSSL_NO_RC4"))]
+                m.add(&arc4, none_type.as_any(), None, Cipher::rc4())?;
+            }
```
alex, line 304: You've got a syntax error here, you want:
```suggestion
            #[cfg(not(CRYPTOGRAPHY_OSSLCONF = "OPENSSL_NO_RC4"))]
            m.add(&arc4, none_type.as_any(), None, Cipher::rc4())?;
```

alex (commented):
You'll need to add this to `src/rust/Cargo.toml` as well (see the long list of other `OPENSSL_NO` values)

alex:
I suspect you'll also need to put the same `cfg()` on the line with `pub static ARC4` in `src/rust/src/types.rs` (because its now unused on those builds). Besides that this LGTM! Thanks for your contribution.

alex (approved):

PR #12631, opened 2025-03-17
pyo3: bump version to v0.24.0

SilverBzH:
pyo3 version prior to 0.24.0 are not able to parse
properly exotic python version. This is fix since
https://github.com/PyO3/pyo3/pull/4949 available on the
v0.24.0 version

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ rust-version = "1.65.0"
 
 [workspace.dependencies]
 asn1 = { version = "0.20.0", default-features = false }
-pyo3 = { version = "0.23.5", features = ["abi3"] }
+pyo3 = { version = "0.24.0", features = ["abi3"] }
 openssl = "0.10.71"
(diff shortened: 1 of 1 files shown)


SilverBzH:
Ah, indeed. Let's wait until 0.25.0 then I guess. I'll properly push the `Cargo.lock` as well if that's works for you ?

alex:
Sure, go for it. There's at least one other change that will be required, even once a release pyo3 with a fix is out.

SilverBzH:
Do you have a link the change needed ? Than I can follow it.

alex:
If you read the failing logs, you'll see that there's three compilation errors. Two are related to the issue I linked, one requires a change from us.

SilverBzH:
Ok sounds good. Once the issue on `pyo3` side is solved, I'll update this PR with the right version. Then I'll wait for your green flag from there :+1:

PR #12673, opened 2025-03-24
fixes #12132 -- implement ssh public key fingerprints

### LINKED ISSUE #12132: feature request - ssh public key fingerprints
Perhaps the feature is there but I cannot find it.

Is there a way to obtain the md5 or sha256 fingerprint of a public key?

With ed25519 it can be done as:

` hashlib.md5(key.public_bytes_raw()).hexdigest() `

Other format do not support Raw encoding. 

Seems wasteful to have to base64 or DER decode the key bytes, again, to get the raw bytes for hashing.

thank you

alex:


alex:
(This was mostly developed by claude-code, but eventually I got bored of baby sitting it and so I just cleaned it up myself)

alex:
an alternative here would be to _just_ compute the hash, no serialization at all.

src/cryptography/hazmat/primitives/serialization/ssh.py
```diff
@@ -1152,6 +1153,46 @@ def _parse_exts_opts(exts_opts: memoryview) -> dict[bytes, bytes]:
     return result
 
 
+def ssh_key_fingerprint(
+    key: SSHPublicKeyTypes,
+    hash_algorithm: hashes.MD5 | hashes.SHA256,
+) -> bytes:
+    """
```
reaperhulk, line 1160: I assume claude code did this, but we definitely don't use comment docs like this!

reaperhulk (commented):
Does the SSH spec define the md5 fingerprint as hex encoded and sha256 as base64 with no padding? If so, can we put that in a comment somewhere? 😄 

Assuming that is a documented property of these fingerprints then I think it's fine to do it for the caller rather than presenting back raw digest bytes.

alex:
So, I'm not aware of a spec for this, I based this on the behavior of `ssh-keygen -l` with each hash, and a review of the OpenSSH source code. The more I think about this, the more I want to just do raw hash bytes, ala `Certificate.fingerprint()`. Do you disagree?

reaperhulk:
Yeah if it's just a current implementation detail I'd rather just do digest bytes with examples of how to encode it to match current ssh-keygen behavior.

reaperhulk (approved):

PR #12693, opened 2025-03-27
fixes #11955 -- added OCSPResponseBuilder.add_response_by_hash

### LINKED ISSUE #11955: Add OCSP Response Builder APIs to add certificates by hash
[Creating an OCSP Response](https://cryptography.io/en/latest/x509/ocsp/#creating-responses) seems to require unnecessary verbose objects "issuer" and "cert" in add_response. 
I guess the Serial Number,  IssuerNameHash and IssuerKeyHash would be enough. 
This change would simplify the implementation of OCSP responders and align better with scenarios where responders operate based on limited certificate information.

alex:


alex:
claude-code wrote the scaffholding, docs, tests, and changelog. I did the impl. Prompt:

> > I would like to add a new API to OCSPResponseBuilder: add_certificate_by_hash. It should be similar to 
  OCSPResponseBuilder.add_certificate, but take the hashes it needs instead of full certificates. This is analogous to 
  OCSPRequestBuilder's add_certificate and add_certificate_by_hash. I would like you to: a) Write a changelog entry, b) 
  write docs, c) write tests, d) write a no-op implementation of the API (i.e., stub out the function). Do not attempt 
  to _actually_ implement it.

Cost:  $1.11

docs/x509/ocsp.rst
```diff
@@ -157,7 +157,7 @@ Creating Requests
         .. versionadded:: 39.0.0
 
         Adds a request using the issuer's name hash, key hash, the certificate
-        serial number and hash algorithm. You can call this method or
+        serial number, and hash algorithm. You can call this method or
```
reaperhulk, line 160: So claude code believes in the oxford comma I guess?

alex: Fascinating, I didn't notice this, but there was some place in their generated docs that _didn't_ have an oxford comma and I added one!

docs/x509/ocsp.rst
```diff
@@ -285,6 +286,55 @@ Creating Responses
...
+            :class:`~cryptography.hazmat.primitives.hashes.SHA1`,
+            :class:`~cryptography.hazmat.primitives.hashes.SHA224`,
+            :class:`~cryptography.hazmat.primitives.hashes.SHA256`,
+            :class:`~cryptography.hazmat.primitives.hashes.SHA384`, and
+            :class:`~cryptography.hazmat.primitives.hashes.SHA512` are allowed.
+
+        :param cert_status: An item from the
+            :class:`~cryptography.x509.ocsp.OCSPCertStatus` enumeration.
+
+        :param this_update: A naïve :class:`datetime.datetime` object
```
reaperhulk, line 322: Do we want to be doing new naïve APIs?

docs/x509/ocsp.rst
```diff
@@ -285,6 +286,55 @@ Creating Responses
             :class:`~cryptography.x509.ReasonFlags` enumeration or ``None`` if
             the ``cert`` is not revoked.
 
+    .. method:: add_response_by_hash(issuer_name_hash, issuer_key_hash, serial_number, algorithm, cert_status, this_update, next_update, revocation_time, revocation_reason)
+
+        .. versionadded:: 45.0.0
+
+        This method adds status information about the certificate that was
```
reaperhulk, line 293: I think we should add an example that shows correct generation of the hashes for this.

docs/x509/ocsp.rst
alex: I'm actually a bit unsure. Our choices are to make it inconsistent with `add_response`, or to make it inconsistent with how we want APIs to work in the future :-/

reaperhulk: Should we add add_response_utc, deprecate add_response, and then just name the new one add_response_by_hash_utc? Although that latter name is pretty incoherent.

docs/x509/ocsp.rst
alex: We don't have that for the OCSPRequest side, fwiw

reaperhulk: We can always be better 😇

alex: shiiit

alex: Because we lack an API for getting the SPKI value bytes (i.e., excluding the OID), it's not particularly clear how to write an example for this.

Issue #10529, opened 2024-03-04
Cypher text


testlorenzospa:
Hello.

I want to encrypt a simple text (not a byte string) in AES256 CBC.
Is it possible?

In a script I use these instruction

`cipher = Cipher(algorithm=AES256(cryptoKey), mode=CBC(iv))`
`encText = cipher.encryptor().update(b"Ciao")`

But if I cange the parameter to the update method in a simple text ("Ciao", without sign "b") I get an error from python

`TypeError: from_buffer() cannot return the address of a unicode object`

What can I do to encrypt a simple text?

Thanks.

alex:
Encryption is an operation performed on bytes, you are passing a unicode string. The correct solution is for you to encode your string to bytes with an encoding such as utf8.

PR #12694, opened 2025-03-27
friendlier error message converting a string to a buffer

ngoldbaum:
Consider this script:

```
import os

key = os.urandom(32)
iv = os.urandom(16)
cipher = Cipher(algorithms.AES(key), modes.CBC(iv))
encryptor = cipher.encryptor()

encryptor.update('abcd1234')
```

Running it will generate this error right now:

```
  File "/Users/goldbaum/Documents/test/test2.py", line 10, in <module>
    encryptor.update('abcd1234')
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^
TypeError: argument 'data': from_buffer() cannot return the address of a unicode object
```

This error message has confused users in the past. See [here](https://stackoverflow.com/questions/59766666/typeerror-from-buffer-cannot-return-the-address-of-a-unicode-object), [here](https://github.com/pyca/cryptography/issues/10529), and [here](https://github.com/pyca/cryptography/issues/6687).

When I hit this I got to the correct solution quickly, but the fact that `from_buffer` isn't a function in user code threw me for a loop. 

We can also suggest the correct solution for what is IMO 99% of use-cases as well:

```
  File "/Users/goldbaum/Documents/test/test2.py", line 10, in <module>
    encryptor.update('abcd1234')
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^
TypeError: argument 'data':
Cannot convert "<class 'str'>" instance to a buffer.
Did you mean to pass a bytestring instead?
```

### DIFF
--- a/src/rust/src/buf.rs
+++ b/src/rust/src/buf.rs
@@ -4,7 +4,9 @@
 
 use std::slice;
 
+use pyo3::exceptions::PyTypeError;
 use pyo3::types::{IntoPyDict, PyAnyMethods};
+use pyo3::PyErr;
 
 use crate::types;
 
@@ -19,6 +21,13 @@ fn _extract_buffer_length<'p>(
     mutable: bool,
 ) -> pyo3::PyResult<(pyo3::Bound<'p, pyo3::PyAny>, usize)> {
     let py = pyobj.py();
+    if pyobj.is_instance_of::<pyo3::types::PyString>() {
+        return Err(PyErr::new::<PyTypeError, _>(
+            "\n\
+             Cannot create a buffer from a string.\n\
+             Did you mean to pass a bytestring instead?",
+        ));
+    }
     let bufobj = if mutable {
         let kwargs = [(pyo3::intern!(py, "require_writable"), true)].into_py_dict(py)?;
         types::FFI_FROM_BUFFER
--- a/src/rust/src/buf.rs
+++ b/src/rust/src/buf.rs
@@ -4,9 +4,7 @@
 
 use std::slice;
 
-use pyo3::exceptions::PyTypeError;
 use pyo3::types::{IntoPyDict, PyAnyMethods};
-use pyo3::PyErr;
 
 use crate::types;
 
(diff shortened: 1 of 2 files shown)


src/rust/src/buf.rs
```diff
@@ -19,6 +21,13 @@ fn _extract_buffer_length<'p>(
     mutable: bool,
 ) -> pyo3::PyResult<(pyo3::Bound<'p, pyo3::PyAny>, usize)> {
     let py = pyobj.py();
+    if pyobj.is_instance_of::<pyo3::types::PyString>() {
+        return Err(PyErr::new::<PyTypeError, _>(
+            "\n\
+             Cannot create a buffer from a string.\n\
+             Did you mean to pass a bytestring instead?",
+        ));
+    }
```
alex, line 30: Hmm, is there a reason to special case strings, as opposed to just raising our own error  anytime `FFI_FROM_BUFFER` fails?

src/rust/src/buf.rs
```diff
@@ -19,6 +21,13 @@ fn _extract_buffer_length<'p>(
     mutable: bool,
 ) -> pyo3::PyResult<(pyo3::Bound<'p, pyo3::PyAny>, usize)> {
     let py = pyobj.py();
+    if pyobj.is_instance_of::<pyo3::types::PyString>() {
+        return Err(PyErr::new::<PyTypeError, _>(
```
alex, line 25: You can use `PyTypeError::new_err("...")` instead.

We generally write out the full path `pyo3::exception::PyTypeError::new_err()`.

alex (commented):
thanks

src/rust/src/buf.rs
ngoldbaum: Good call, replaced this with a `map_err` on the `Result` returned by calling `get_buffer`.

src/rust/src/buf.rs
```diff
@@ -23,10 +23,21 @@ fn _extract_buffer_length<'p>(
...
+            .call((pyobj,), Some(&kwargs))
     } else {
-        types::FFI_FROM_BUFFER.get(py)?.call1((pyobj,))?
-    };
+        types::FFI_FROM_BUFFER.get(py)?.call1((pyobj,))
+    }
+    .map_err(|_| {
+        let errmsg = if pyobj.is_instance_of::<pyo3::types::PyString>() {
+            format!(
+                "\nCannot convert \"{}\" instance to a buffer.\nDid you mean to pass a bytestring instead?",
```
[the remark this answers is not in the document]
ngoldbaum: I thought it looked slightly nicer in my terminal. But on second thought it's not so bad, dropped the leading carriage return.

alex (approved):

PR #12702, opened 2025-03-29
CRL hash algorithm getter breaks with PSS padding

maddinat0r:
Calling `crl.signature_hash_algorithm` on a PSS-padded CRL object would result in following exception:
```
cryptography.exceptions.UnsupportedAlgorithm: Signature algorithm OID: 1.2.840.113549.1.1.10 not recognized
```
This PR fixes this issue.


alex (approved):

alex:
Thanks for the fix!

PR #12807, opened 2025-04-25
Update license metadata in `pyproject.toml` according to PEP 639

facutuesca:
`setuptools` [now warns](https://github.com/pyca/cryptography/actions/runs/14663972603/job/41154408745#step:10:973) when building a project that uses license classifiers:

```
DEBUG /github/home/.cache/uv/builds-v0/.tmpjaMf3z/lib/python3.13/site-packages/setuptools/dist.py:759: SetuptoolsDeprecationWarning: License classifiers are deprecated.
DEBUG !!
DEBUG 
DEBUG         ********************************************************************************
DEBUG         Please consider removing the following classifiers in favor of a SPDX license expression:
DEBUG 
DEBUG         License :: OSI Approved :: MIT License
DEBUG 
DEBUG         See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
DEBUG         ********************************************************************************
DEBUG 
DEBUG !!
```

This is due to [PEP 639](https://peps.python.org/pep-0639/), which changes the way the licenses (and their files) are specified in the project metadata (see also the packaging guide [here](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files)). ~This [is supported](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#id9) since version `77.0.3` of `setuptools`.~

This PR changes the metadata accordingly, ~and bumps the minimum version of `setuptools` to ensure it's supported.~

cc @woodruffw @alex

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,12 +21,11 @@ authors = [
 ]
 description = "cryptography is a package which provides cryptographic recipes and primitives to Python developers."
 readme = "README.rst"
-license = { text = "Apache-2.0 OR BSD-3-Clause" }
+license = "Apache-2.0 OR BSD-3-Clause"
+license-files = [ "LICENSE", "LICENSE.APACHE", "LICENSE.BSD" ]
 classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Intended Audience :: Developers",
-    "License :: OSI Approved :: Apache Software License",
-    "License :: OSI Approved :: BSD License",
     "Natural Language :: English",
     "Operating System :: MacOS :: MacOS X",
     "Operating System :: POSIX",
@@ -105,9 +104,6 @@ features = ["pyo3/abi3-py37"]
 include = [
     "CHANGELOG.rst",
     "CONTRIBUTING.rst",
-    "LICENSE",
-    "LICENSE.APACHE",
-    "LICENSE.BSD",
 
     "docs/**/*",
 


alex:
I think we want to leave the setuptools version alone (for 3.7 support) and just update the metadata. People on old setuptools will get marginally wrong stuff, and I guess that's ok.

facutuesca:
ah but I think that it will cause their builds to outright fail:

> If you get a build error that license should be a dict/table, your build backend doesn’t yet support the new format. See the [above section](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files) for more context. The now deprecated format is [described in PEP 621](https://peps.python.org/pep-0621/#license).

alex:
Actually, why is setuptools even processing this? We use maturin for builidng the package?

facutuesca:
> Actually, why is setuptools even processing this? We use maturin for builidng the package?

it seems to originate from this `uv pip install` call: https://github.com/pyca/cryptography/actions/runs/14663972603/job/41154408745#step:10:76

facutuesca:
🤦 the warning is not coming from `cryptography`, but rather while building `cffi`: https://github.com/pyca/cryptography/actions/runs/14663972603/job/41154408745#step:10:928

`cffi` is the package that has that `License :: OSI Approved :: MIT License` classifier

alex:
heh. well, probably good to do this cleanup either way, since maturin appears to handle it properly in all versions.

alex (approved):

PR #12839, opened 2025-05-01
FIPS fixes

reaperhulk:
This resolves the changes the OpenSSL 3.5 FIPS provider made:
* PSS salt length must match digest length
* OAEP decryptions will reject ciphertext if it's not in the range of `1 < ct < (n - 1)`
* For PSS inner MGF1 hash must match outer hash.

This also updates an X509 PSS cert test vector to replace the max salt length with a digest salt length.

Depends on https://github.com/C2SP/wycheproof/pull/145 (at which point I can update the wycheproof hash in this PR and remove the workaround)


src/cryptography/hazmat/backends/openssl/backend.py
```diff
@@ -181,14 +181,13 @@ def rsa_padding_supported(self, padding: AsymmetricPadding) -> bool:
...
             return True
         elif isinstance(padding, PSS) and isinstance(padding._mgf, MGF1):
-            # SHA1 is permissible in MGF1 in FIPS even when SHA1 is blocked
-            # as signature algorithm.
-            if self._fips_enabled and isinstance(
-                padding._mgf._algorithm, hashes.SHA1
+            # FIPS 186-4 only allows salt length == digest length for PSS
+            if (
+                self._fips_enabled
+                and padding._salt_length != PSS.DIGEST_LENGTH
```
alex, line 187: teeecchnicallly I guess its ok to manually pass a salt length equal to teh digest length? I guess we don't care?

tests/wycheproof/test_rsa.py
```diff
@@ -138,25 +138,32 @@ def test_rsa_pkcs1v15_signature_generation(backend, wycheproof):
...
     mgf_digest = _DIGESTS[wycheproof.testgroup["mgfSha"]]
-
     if digest is None or mgf_digest is None:
         pytest.skip(
             "PSS with digest={} and MGF digest={} not supported".format(
                 wycheproof.testgroup["sha"],
                 wycheproof.testgroup["mgfSha"],
             )
         )
+    salt_length = int(wycheproof.testgroup["sLen"])
```
alex, line 149: is `int()` required here? if so, why isnt' it required below -- and can you reuse teh local variable down below.

src/cryptography/hazmat/backends/openssl/backend.py
reaperhulk: Yeah if we ever do this in our tests it'll fail and we'll fix it 😄

alex: can you add a comment for future us?

tests/wycheproof/test_rsa.py
reaperhulk: Fixed (it's not required so we don't need this var at all)

tests/hazmat/primitives/test_rsa.py
```diff
@@ -503,16 +503,25 @@ def test_pss_signing(self, subtests, backend):
...
         [hashes.SHA224(), hashes.SHA256(), hashes.SHA384(), hashes.SHA512()],
     )
-    def test_pss_signing_sha2(self, rsa_key_2048, hash_alg, backend):
+    def test_pss_sha2_max_length(self, rsa_key_2048, hash_alg, backend):
         _skip_pss_hash_algorithm_unsupported(backend, hash_alg)
         private_key = rsa_key_2048
         public_key = private_key.public_key()
         pss = padding.PSS(
-            mgf=padding.MGF1(hash_alg), salt_length=padding.PSS.MAX_LENGTH
+            mgf=padding.MGF1(hash_alg), salt_length=padding.PSS.DIGEST_LENGTH
```
alex, line 524: This uses `MAX_LENGTH` in only if but `DIGEST_LENGTH` in the test. The test function is max length, so I think the digest is wrong?

reaperhulk: good catch, fix pushed

alex:
Ok, looks good pending wycheproof merging the changes

PR #12853, opened 2025-05-03
Align our handling of non-positive serial numbers.

### LINKED ISSUE #12748: Cryptography parsed a CRL file with the authorityCertSerialNumber set to 0.
Version:
cryptography-44.0.2

I have a CRL file where the authorityCertSerialNumber field in the AKI extension is set to 0. Cryptography successfully parsed the authorityCertSerialNumber field without any errors. However, according to RFC5280, the authorityCertSerialNumber field stores the certificate serial number, which must be a positive integer.So, is this a bug?

Code:
```
from cryptography.x509 import load_pem_x509_crl, load_der_x509_crl
from cryptography.x509 import ExtensionNotFound
import sys

def load_crl(file_path):
    with open(file_path, "rb") as f:
        crl_data = f.read()
    try:
        crl = load_pem_x509_crl(crl_data)
    except ValueError:
        crl = load_der_x509_crl(crl_data)
    return crl

def print_crl_issuer(file_path):
    aki_extension = None
    try:
        for ext in crl.extensions:
             if ext.oid == x509.oid.ExtensionOID.AUTHORITY_KEY_IDENTIFIER:
                    aki=ext.value
                    print(aki.authority_cert_issuer)
    except Exception as e:
        print(f"Error parsing CRL: {e}")

file_path = 'crl_aki_serial_0.der'
print_crl_issuer(file_path)
```
Test Case:

[crl_aki_serial_0.zip](https://github.com/user-attachments/files/19668466/crl_aki_serial_0.zip)

### LINKED ISSUE #12771: The cryptography system parsed a CRL file that contains a revoked certificate with a serial number of 0.
Version:
cryptography-44.0.2

Hello Developer,

I have a CRL file that contains a revoked certificate with a serial number of 0. Cryptography successfully parsed this serial number without any errors. However, according to RFC 5280, the certificate serial number must be a non-negative integer. At the same time, tests show that cryptography will reject the parsing of revoked certificates with negative serial numbers.

Code:
```
from cryptography.x509 import load_pem_x509_crl, load_der_x509_crl
from cryptography.x509 import ExtensionNotFound
import sys

def load_crl(file_path):
    with open(file_path, "rb") as f:
        crl_data = f.read()
    try:
        crl = load_pem_x509_crl(crl_data)
    except ValueError:
        crl = load_der_x509_crl(crl_data)
    return crl

def print_crl_issuer(file_path):
    crl=load_crl(file_path)
    try:
        for entry in crl:
               print(entry)
    except Exception as e:
        print(f"Error occurred: {e}")

file_path = 'crl_revoked_serial_0.der'
print_crl_issuer(file_path)
```
Test Case:

[crl_revoked_serial_0.zip](https://github.com/user-attachments/files/19759738/crl_revoked_serial_0.zip)

Another Test Case:

[crl_revoked_serial_-24.zip](https://github.com/user-attachments/files/19759747/crl_revoked_serial_-24.zip)

alex:
Basically: for now we allow them, but when we drop them they'll be dropped everywhere.

Closes #12771
Closes #12748

### DIFF
--- a/src/rust/cryptography-x509/src/certificate.rs
+++ b/src/rust/cryptography-x509/src/certificate.rs
@@ -31,12 +31,17 @@ impl<'a> Certificate<'a> {
     }
 }
 
+// This should really be a wrapper around `BigUint` that rejects 0s, however
+// for the time being we support invalid serial numbers (mostly because the MS
+// trust store has a certificate with a negative serial number).
+pub type SerialNumber<'a> = asn1::BigInt<'a>;
+
 #[derive(asn1::Asn1Read, asn1::Asn1Write, Hash, PartialEq, Eq, Clone)]
 pub struct TbsCertificate<'a> {
     #[explicit(0)]
     #[default(0)]
     pub version: u8,
-    pub serial: asn1::BigInt<'a>,
+    pub serial: SerialNumber<'a>,
     pub signature_alg: common::AlgorithmIdentifier<'a>,
 
     pub issuer: name::Name<'a>,
--- a/src/rust/cryptography-x509/src/crl.rs
+++ b/src/rust/cryptography-x509/src/crl.rs
@@ -2,6 +2,7 @@
 // 2.0, and the BSD License. See the LICENSE file in the root of this repository
 // for complete details.
 
+use crate::certificate::SerialNumber;
 use crate::common::Asn1Operation;
 use crate::{common, extensions, name};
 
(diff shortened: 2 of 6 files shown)


reaperhulk:
Do we want to raise warnings when we detect these invalid values? As is, this PR actually makes the parser for CRL more lax and doesn't raise warnings (or else I misread it, which is always possible)

alex:
Yeah, let me see what I can do.

AGWA:
FYI, rejecting non-positive revoked certificate serial numbers in CRLs came up in Go recently, and the consensus emerged to continue accepting them indefinitely due to the need for CAs to be able to revoke misissued certificates with negative serial numbers: https://github.com/golang/go/issues/73433#issuecomment-2819738114

reaperhulk:
@AGWA Thanks for the link. Such a shame this format even allows encoding the wrong thing 😭

alex:
I hate computers so much. Thanks, that's a good point

PR #12855, opened 2025-05-04
fixes #12094 -- added support for argon2id phc format

### LINKED ISSUE #12094: Support "PHC String Format" for argon2(id) hashing
The [reference implementation](https://github.com/P-H-C/phc-winner-argon2) of argon2 supports an output format that encodes the hashing parameters, along with the hash itself. This is useful, for example, if you're storing password hashes in a database and you want to upgrade the parameters (or even the algorithm itself) over time.

The format is documented here: https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md

The format is supported by other argon2 libraries/implementations, including https://github.com/hynek/argon2-cffi

Similar serialisation formats exist for most hashing schemes, and [passlib](https://passlib.readthedocs.io/en/stable/) supports many of them (including argon2).

alex:
This was implemented using `claude-code` and then non-trivial manual cleanup. The prompt used was:

    Please implement the feature requested in https://github.com/pyca/cryptography/issues/12094.
    
    This means adding the following methods to the existing `Argon2id` class:
    
    ```
    class Argon2id:
        def derive_phc_encoded(self, key_material: bytes) -> str: ...
        @classmethod
        def verify_phc_encoded(key_material: bytes, phc_encoded: str) -> None: ...
    ```
    
    The specification can be found at https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md
    
    You should plan to implement:
    
    - These two methods (in Rust, on the existing class)
    - The python stub file declarations for them
    - Tests for these
    - Documentation
    - A changelog entry
    
    Before you start to write code, do the required research and planning and present me with a detailed plan of what you will do.

The total cost was $2.64


reaperhulk:
This generally looks fine, but can we add the test vector from the spec as one final sanity check? https://github.com/P-H-C/phc-string-format/blob/e8fbd333dcc9a8b0843fac6b33371cf157e91a48/phc-sf-spec.md

reaperhulk (approved):

PR #12857, opened 2025-05-05
Add `py311` wheels with faster `PyBuffer` API for significant buffer performance gains

### LINKED ISSUE #12854: cryptography_rust::buf::_extract_buffer_length overhead for small messages
Hi there! 👋

First off, thank you for the fantastic work on the Rust-backed cryptography components — the performance improvements have been impressive and very much appreciated. While running some benchmarks, I noticed that cryptography_rust::buf::_extract_buffer_length appears to introduce noticeable overhead for smaller messages. It might just be the name giving me the wrong impression, but it seems a bit more expensive than expected for what sounds like a lightweight operation.

For example:

**1 KiB messages (aioesphomeapi)**  
https://codspeed.io/esphome/aioesphomeapi/benchmarks/tests/benchmarks/test_noise.py::test_noise_messages%5B1024%5D  
1KiB benchmark

**8 KiB messages (snitun)**  
https://codspeed.io/NabuCasa/snitun/benchmarks/tests/benchmarks/test_server_sni.py::test_server_send_message%5B1000@8KiB%5D  
8KiB benchmark

Is there anything that might help reduce the runtime for smaller payloads in this path?  
Happy to help test ideas or provide more profiling detail if useful.

Thanks again for all the amazing work on this project!

bdraco:
## Summary of Changes

This pull request adds support for optimized buffer handling using `pyo3::buffer::PyBuffer` on Python 3.11, along with several related build system improvements.

> ⚠️ **Note:** These optimizations are available only in wheels built with the `abi3-py311` feature (or later). To include them in manual or vendor builds, use the following command:  
> `pip wheel --config-settings="build-args=--features=pyo3/abi3-py311"`
> We understand that passing custom arguments to your wheel builder system may not always be feasible. We are actively working on improvements to eliminate the need for this manual step in the future.

### Features & Enhancements

- **Optimized buffer access**: On Python 3.11, the `CffiBuf` type now uses `pyo3::buffer::PyBuffer` to significantly speed up operations on small buffers.
- **New wheel builds**: Added publishing of `py311` wheels targeting `abi3-py311` using `pyo3::buffer` which replaces the `abi3-py39` wheels.

### Build System Changes

- **Added `pyo3-build-config`** as a build dependency to handle feature configuration cleanly.
- **Updated `build.rs`** to print relevant PyO3 build script outputs.

### CI & Testing

- Extended `.github/workflows/wheel-builder.yml` to include `py311` ABI wheels for CPython 3.11.
  
### Changelog

- Documented the addition of `py311` wheels with improved buffer performance.

### Testing

- [x] Home Assistant production instance - 1 - install and restart, verify HomeKit and other uses of this library
- [x] Home Assistant production instance - 2 - install and restart, verify HomeKit and other uses of this library
- [x] https://github.com/esphome/aioesphomeapi/pull/1174 - test CI run

### Benchmark

| Benchmark | Base | PR | Delta |
| --------- | ---- | -- | ----- |
| `tests/bench/test_aead.py::test_aesccm_decrypt` | 2735.0 ns | 1455.25 ns | 47% faster |
| `tests/bench/test_aead.py::test_aesccm_encrypt` | 2765.0 ns | 1593.0 ns | 42% faster |
| `tests/bench/test_aead.py::test_aesgcm_decrypt` | 2845.0 ns | 1603.0 ns | 44% faster |
| `tests/bench/test_aead.py::test_aesgcm_encrypt` | 2825.0 ns | 1583.0 ns | 44% faster |
| `tests/bench/test_aead.py::test_aesocb3_decrypt` | 1883.33 ns | 956.83 ns | 49% faster |
| `tests/bench/test_aead.py::test_aesocb3_encrypt` | 1984.0 ns | 1102.0 ns | 44% faster |
| `tests/bench/test_aead.py::test_aessiv_decrypt` | 1973.33 ns | 1375.25 ns | 30% faster |
| `tests/bench/test_aead.py::test_aessiv_encrypt` | 2054.0 ns | 1493.0 ns | 27% faster |
| `tests/bench/test_aead.py::test_chacha20poly1305_decrypt` | 3537.0 ns | 1863.0 ns | 47% faster |
| `tests/bench/test_aead.py::test_chacha20poly1305_encrypt` | 3527.0 ns | 1843.0 ns | 48% faster |
| `tests/bench/test_ec_load.py::test_load_ec_private_numbers` | 26299.0 ns | 25527.0 ns | 3% faster |
| `tests/bench/test_ec_load.py::test_load_ec_public_numbers` | 15579.0 ns | 15118.0 ns | 3% faster |
| `tests/bench/test_fernet.py::test_fernet_encrypt` | 20298.0 ns | 16060.0 ns | 21% faster |
| `tests/bench/test_hashes.py::test_sha256` | 2474.0 ns | 1934.0 ns | 22% faster |
| `tests/bench/test_hmac.py::test_hmac_sha256` | 3636.0 ns | 2574.0 ns | 29% faster |
| `tests/bench/test_x509.py::test_aki_public_bytes` | 672.0 ns | 601.0 ns | 11% faster |
| `tests/bench/test_x509.py::test_load_der_certificate` | 2946.0 ns | 2926.0 ns | 1% faster |
| `tests/bench/test_x509.py::test_load_pem_certificate` | 14397.0 ns | 14417.0 ns | 0% slower |
| `tests/bench/test_x509.py::test_object_identifier_constructor` | 491.0 ns | 302.05 ns | 38% faster |
| `tests/bench/test_x509.py::test_verify_docs_python_org` | 190255.0 ns | 182380.0 ns | 4% faster |

closes #12854


src/rust/src/buf.rs
```diff
@@ -46,47 +40,130 @@ fn _extract_buffer_length<'p>(
...
+                } else {
+                    format!("Cannot convert \"{}\" instance to a buffer.", pyobj.get_type())
+                };
+                pyo3::exceptions::PyTypeError::new_err(errmsg)
+            })?;
+            let len = pyobj.len()?;
+            let buf = if len == 0 {
+                &[]
+            } else {
+                // SAFETY: pyo3::buffer::PyBuffer ensures that we have a valid ptr
```
bdraco, line 93: Probably much safer now since we know its not released until `drop`
https://github.com/PyO3/pyo3/blob/c25c6f773dd023720b4e6b3969d5e99fef57433d/src/buffer.rs#L198
https://github.com/PyO3/pyo3/blob/c25c6f773dd023720b4e6b3969d5e99fef57433d/src/buffer.rs#L613

src/rust/src/buf.rs
```diff
@@ -46,47 +40,130 @@ fn _extract_buffer_length<'p>(
...
+                    format!(
+                        "Cannot convert \"{}\" instance to a buffer.\nDid you mean to pass a bytestring instead?",
+                        pyobj.get_type()
+                    )
+                } else {
+                    format!("Cannot convert \"{}\" instance to a buffer.", pyobj.get_type())
+                };
+                pyo3::exceptions::PyTypeError::new_err(errmsg)
+            })?;
+            let len = pyobj.len()?;
```
bdraco, line 89: Would it be better to do `bufobj.len_bytes()`?
I'm so used to working with Cython that I usually look at the length of the Python object because Cython length can be affected by a null bytes.

bdraco: That does seem to be safe https://github.com/PyO3/pyo3/blob/c25c6f773dd023720b4e6b3969d5e99fef57433d/src/buffer.rs#L288

bdraco: adjusted to that in 09e3b128c8921b2b5bec110ee6c7abcafbecb537

.github/workflows/ci.yml
```diff
@@ -26,8 +26,11 @@ jobs:
         PYTHON:
           - {VERSION: "3.13", NOXSESSION: "flake"}
           - {VERSION: "3.13", NOXSESSION: "rust"}
+          - {VERSION: "3.13", NOXSESSION: "rust,tests", NOX_INSTALL_ARGS: "--features='pyo3/abi3-py37'"}
+          - {VERSION: "3.13", NOXSESSION: "rust,tests", NOX_INSTALL_ARGS: "--features='pyo3/abi3-py39'"}
```
bdraco, line 30: Can py39 be removed per https://github.com/pyca/cryptography/issues/12854#issuecomment-2849394290?

.github/workflows/wheel-builder.yml
```diff
@@ -295,6 +308,7 @@ jobs:
         PYTHON:
           - {VERSION: "3.11", "ABI_VERSION": "py37"}
           - {VERSION: "3.11", "ABI_VERSION": "py39"}
```
bdraco, line 310: Can py39 be removed per https://github.com/pyca/cryptography/issues/12854#issuecomment-2849394290?

src/rust/src/buf.rs
```diff
@@ -46,47 +40,130 @@ fn _extract_buffer_length<'p>(
...
+            };
+
+            Ok(CffiBuf {
+                pyobj: pyobj.clone(),
+                _bufobj: bufobj,
+                buf,
+            })
+        }
     }
 }
```
bdraco, line 168: `CffiMutBuf` was not touched as `get` is implemented as `ffi::PyObject_GetBuffer(obj.as_ptr(), buf.as_mut_ptr(), ffi::PyBUF_FULL_RO)` so it would happily accept `PyBytes` objects and write into them.  I wasn't sure what kind of guard was needed there and if something like `obj.is_instance_of::<pyo3::types::PyBytes>` would have been enough so I did not implement it but likely could be done in a future PR.

.github/workflows/wheel-builder.yml
```diff
@@ -67,6 +67,7 @@ jobs:
         PYTHON:
           - { VERSION: "cp311-cp311", ABI_VERSION: 'py37' }
           - { VERSION: "cp311-cp311", ABI_VERSION: 'py39' }
```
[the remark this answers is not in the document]
alex: Yeah, we want to replace py39 with py311, not supplement it.

.github/workflows/wheel-builder.yml
```diff
@@ -199,6 +200,18 @@ jobs:
             # This will change in the future as we change the base Python we
```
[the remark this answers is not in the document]
alex: Yup

src/rust/src/buf.rs
alex: I believe all that's required is checking the `readonly` property on the `Buffer` object and raising if its true.

alex:
(I've got this on my TODO list to give a comprehensive review this evening. Thanks for your work!)

.github/workflows/ci.yml
alex: Yes, I don't think we need py39 here (there's no actual branches for it, we build py39 wheels due to some branch in pyo3 itself)

.github/workflows/ci.yml
```diff
@@ -26,8 +26,11 @@ jobs:
...
+          - {VERSION: "3.13", NOXSESSION: "rust,tests", NOX_INSTALL_ARGS: "--features='pyo3/abi3-py311'"}
           - {VERSION: "3.12", NOXSESSION: "docs", OPENSSL: {TYPE: "openssl", VERSION: "3.5.0"}}
-          - {VERSION: "3.14-dev", NOXSESSION: "tests"}
+          - {VERSION: "3.14-dev", NOXSESSION: "tests", NOX_INSTALL_ARGS: "--features='pyo3/abi3-py313'"}
```
alex, line 33: Hmm, is this required because without it, the build fails?

.github/workflows/wheel-builder.yml
alex: Same here

Cargo.toml
```diff
@@ -21,7 +21,8 @@ rust-version = "1.65.0"
 
 [workspace.dependencies]
 asn1 = { version = "0.21.1", default-features = false }
-pyo3 = { version = "0.24.2", features = ["abi3"] }
+pyo3 = { version = "0.24.2" }
```
alex, line 24: So we really want to always have the abi3 feature enabled, we never want to build per-python wheels. Is this change necessary?

docs/installation.rst
```diff
@@ -234,6 +234,22 @@ dependencies.
     cd ..
     OPENSSL_DIR="${CWD}/openssl" pip wheel --no-cache-dir --no-binary cryptography cryptography
 
+Building a wheel with the Limited API
```
alex, line 237: Yeah, in the same vein, we shouldn't need to document this because it should be the default :-)

noxfile.py
```diff
@@ -31,10 +31,24 @@ def install(
 ) -> None:
     if verbose:
         args += ("-v",)
+    # Check for any --features= argument passed to nox
```
alex, line 34: Ugly, but I don't have a better suggestion :-/

src/rust/build.rs
```diff
@@ -6,6 +6,10 @@ use std::env;
 
 #[allow(clippy::unusual_byte_groupings)]
 fn main() {
+    for cfg in pyo3_build_config::get().build_script_outputs() {
+        println!("{cfg}");
+    }
```
alex, line 11:
```suggestion
    pyo3_build_config::use_pyo3_cfgs();
```

alex: This change also obviants the need to add it to `unexpected_cfgs`.

src/rust/src/buf.rs
```diff
@@ -46,47 +40,130 @@ fn _extract_buffer_length<'p>(
...
 
-impl<'a> CffiBuf<'a> {
-    pub(crate) fn from_bytes(py: pyo3::Python<'a>, buf: &'a [u8]) -> Self {
-        CffiBuf {
-            pyobj: py.None().into_bound(py),
-            _bufobj: py.None().into_bound(py),
-            buf,
-        }
+#[cfg(Py_3_11)]
```
alex, line 43: So, instead of this approach, I think you shoudl be able to have only `_extract_buffer_length` and the `_bufobj` field in `CffiBuf` be contingent on the version -- basically have those return `(PyBuffer, ptr)` on py311 instead. It shoudl result in a smaller diff.

Let me know if this is unclear.

src/rust/src/buf.rs
```diff
@@ -46,47 +40,130 @@ fn _extract_buffer_length<'p>(
...
+        pub(crate) fn into_pyobj(self) -> pyo3::Bound<'a, pyo3::PyAny> {
+            self.pyobj
+        }
     }
 
-    pub(crate) fn into_pyobj(self) -> pyo3::Bound<'a, pyo3::PyAny> {
-        self.pyobj
+    impl<'a> pyo3::conversion::FromPyObject<'a> for CffiBuf<'a> {
+        fn extract_bound(pyobj: &pyo3::Bound<'a, pyo3::PyAny>) -> pyo3::PyResult<Self> {
+            let bufobj = PyBuffer::<u8>::get(pyobj).map_err(|_| {
```
alex, line 78: Does this reject multi-dimension buffers, if not we should check for and reject them? (I'm 99% sure our existing APIs reject those)

alex (commented):
Left a handful of comments. But this looks great.

The biggest thing to figure out is the build/ABI version changes. Our ideal behavior is:

- Builds always use abi3
- The default is to use "abi3 of the current python versions"
- In CI we build abi3 for some specific version

On `main`, right now we don't have (2) -- we default to always py37. With this PR we default to non-abi3 builds. Would like to find a way to reconcile this, if it doesn't seem possible we can ask for a feature on maturin.

bdraco:
> The biggest thing to figure out is the build/ABI version changes. Our ideal behavior is:
> 
> * Builds always use abi3
> * The default is to use "abi3 of the current python versions"
> * In CI we build abi3 for some specific version
> 
> On `main`, right now we don't have (2) -- we default to always py37. With this PR we default to non-abi3 builds. Would like to find a way to reconcile this, if it doesn't seem possible we can ask for a feature on maturin.

Would you mind sharing a bit more context on why builds should default to `abi3`? I'm having a hard time understanding the motivation, particularly from the perspective of downstream users. My assumption has been that most people building or installing wheels expect a version tailored to their specific Python environment.

As far as I can tell, there isn't a way to say "default to `abi3` for the current Python version," since that will always fail unless we explicitly select a version:  
https://github.com/PyO3/maturin/blob/2f305aaa4ef79b9887ee50a4dcb7c81a4721e109/src/build_options.rs#L984

.github/workflows/ci.yml
bdraco: Yes since 3.14 is not a currently supported version by the production tooling

src/rust/src/buf.rs
bdraco: I had tried that first but `PyBytes` is not `readonly`

.github/workflows/wheel-builder.yml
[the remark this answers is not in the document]
bdraco: removed in 8f34f134e

.github/workflows/wheel-builder.yml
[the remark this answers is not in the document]
bdraco: removed in 8f34f134e

.github/workflows/ci.yml
bdraco: removed in 8f34f134e

.github/workflows/ci.yml
alex: This is one of those consequences of dropping abi3 by default -- with abi3 we can build fine against py314. (I should have included this in my original list)

src/rust/Cargo.toml
```diff
@@ -33,4 +36,4 @@ name = "cryptography_rust"
...
 [lints.rust]
-unexpected_cfgs = { level = "warn", check-cfg = ['cfg(CRYPTOGRAPHY_OPENSSL_300_OR_GREATER)', 'cfg(CRYPTOGRAPHY_OPENSSL_309_OR_GREATER)', 'cfg(CRYPTOGRAPHY_OPENSSL_320_OR_GREATER)', 'cfg(CRYPTOGRAPHY_OPENSSL_330_OR_GREATER)', 'cfg(CRYPTOGRAPHY_OPENSSL_350_OR_GREATER)', 'cfg(CRYPTOGRAPHY_IS_LIBRESSL)', 'cfg(CRYPTOGRAPHY_IS_BORINGSSL)', 'cfg(CRYPTOGRAPHY_IS_AWSLC)', 'cfg(CRYPTOGRAPHY_BUILD_OPENSSL_NO_LEGACY)', 'cfg(CRYPTOGRAPHY_OSSLCONF, values("OPENSSL_NO_IDEA", "OPENSSL_NO_CAST", "OPENSSL_NO_BF", "OPENSSL_NO_CAMELLIA", "OPENSSL_NO_SEED", "OPENSSL_NO_SM4", "OPENSSL_NO_RC4"))'] }
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(Py_3_11)', 'cfg(CRYPTOGRAPHY_OPENSSL_300_OR_GREATER)', 'cfg(CRYPTOGRAPHY_OPENSSL_309_OR_GREATER)', 'cfg(CRYPTOGRAPHY_OPENSSL_320_OR_GREATER)', 'cfg(CRYPTOGRAPHY_OPENSSL_330_OR_GREATER)', 'cfg(CRYPTOGRAPHY_OPENSSL_350_OR_GREATER)', 'cfg(CRYPTOGRAPHY_IS_LIBRESSL)', 'cfg(CRYPTOGRAPHY_IS_BORINGSSL)', 'cfg(CRYPTOGRAPHY_IS_AWSLC)', 'cfg(CRYPTOGRAPHY_BUILD_OPENSSL_NO_LEGACY)', 'cfg(CRYPTOGRAPHY_OSSLCONF, values("OPENSSL_NO_IDEA", "OPENSSL_NO_CAST", "OPENSSL_NO_BF", "OPENSSL_NO_CAMELLIA", "OPENSSL_NO_SEED", "OPENSSL_NO_SM4", "OPENSSL_NO_RC4"))'] }
```
alex, line 39: This is no longer needed now.

bdraco: Dropped in 0026cc4b2

src/rust/src/buf.rs
alex: wat... how can that possibly be right?

src/rust/src/buf.rs
alex: Good point, we can extend `_extract_buffer_length` to return `(buf, ptr, length)`, that has the side benefit of making the name accurate.

src/rust/src/buf.rs
alex: I can't reproduce this:

```
fn main() -> anyhow::Result<()> {
    pyo3::prepare_freethreaded_python();

    pyo3::Python::with_gil(|py| -> pyo3::PyResult<()> {
        let b = pyo3::types::PyBytes::new(py, b"hello world");
        let buf = pyo3::buffer::PyBuffer::<u8>::get(&b)?;
        dbg!(buf.readonly());

        let b = pyo3::types::PyByteArray::new(py, b"hello world");
        let buf = pyo3::buffer::PyBuffer::<u8>::get(&b)?;
        dbg!(buf.readonly());

        Ok(())
    })?;

    Ok(())
}
```

alex: Err, forgot to include the result:

```
[src/main.rs:7:9] buf.readonly() = true
[src/main.rs:11:9] buf.readonly() = false
```

bdraco: This test will fail: https://github.com/pyca/cryptography/blob/7a49948ebf22bc8050cd34bc6cce7154634fadc8/tests/hazmat/primitives/test_ciphers.py#L231. It looks like the `readonly` flag isn’t automatically derived from Python-level mutability, objects like bytes are immutable to Python, but the buffer interface doesn’t appear to reflect that.

bdraco: Let me try it again. Hopefully I got something wrong the first time

bdraco:
> The motivation for abi3 by default is that a) it ensures a more consistent experience, b) its more resilient (in that users are less likely to inadvertantly develop a reliance on it), c) it keeps consistency between our wheel builders, CI, and local dev (without annotating all CI jobs with the abi3 annotation)

Do you have any ideas on how we could ensure that downstreams building wheels with Python 3.11+ automatically pick up the PyBuffer path? The main challenge is that many downstreams, such as Home Assistant or piwheels, don’t customize their wheel builds and just use whatever `pip wheel` outputs by default.

src/rust/src/buf.rs
alex: Oh, if the existing impl is fine with them, than we shoudl be consistent with that -- I assumed it rejected them.

alex:
So, `--build-option` should do it, I'm not sure how to tell `pip wheel` that a given option is for a specific package though.

reaperhulk:
So the concern is that ecosystems that build their own wheels rather than using the ones we supply on PyPI are likely to build the default rather than an optimized one?

With our current approach, users installing from PyPI would receive the correct optimized wheel including armv7l since we ship those wheels now. However, would users on raspbian see those wheels or would piwheels take precedence? Similarly, I don't know if HAOS has other package indices with higher priority.

bdraco:
> So the concern is that ecosystems that build their own wheels rather than using the ones we supply on PyPI are likely to build the default rather than an optimized one?
> 
> With our current approach, users installing from PyPI would receive the correct optimized wheel including armv7l since we ship those wheels now. However, would users on raspbian see those wheels or would piwheels take precedence? Similarly, I don't know if HAOS has other package indices with higher priority.

For anyone building their own wheels, the vendor index typically takes precedence over PyPI. As a result, they’d likely pull in the vendor-built wheel, which is currently the one generated by `pip wheel`, limited to `py37`, so if we leave the default as-is, they end up with the FFI implementation.

src/rust/src/buf.rs
bdraco: I added
```
            if bufobj.dimensions() != 1 {
                return Err(pyo3::exceptions::PyTypeError::new_err(
                    "Only 1-dimensional byte buffers are supported.",
                ));
            }
```

And the above test passes on the Buffer branch, and fails on the FFI branch so it looks like the test is good.

I'll mark this resolves since the behavior is the same without adding this change

bdraco: Test was run on my fork to avoid eating up CI resources here https://github.com/bdraco/cryptography/pull/7

src/rust/src/buf.rs
bdraco: You're right, I must've messed something up on my first try. I just tested it again and it works exactly as you described.

Pushed in 6bf9f981c885171d5d29d2cd86bb8a1780ac0a20. I'll follow up with some cleanup to make it more DRY.

src/rust/src/buf.rs
bdraco: I've refactored the code to reduce duplication. Let me know if there's anything else you'd suggest cleaning up or simplifying further.

alex: Yes, I think it could be simplified further. Here's what I'm imaginging:

```diff
diff --git a/src/rust/src/buf.rs b/src/rust/src/buf.rs
index 35ab7753b..c2d4ed059 100644
--- a/src/rust/src/buf.rs
+++ b/src/rust/src/buf.rs
@@ -4,20 +4,27 @@
 
 use std::slice;
 
-use pyo3::types::{IntoPyDict, PyAnyMethods};
+#[cfg(not(Py_3_11))]
+use pyo3::types::IntoPyDict;
+use pyo3::types::PyAnyMethods;
 
+#[cfg(not(Py_3_11))]
 use crate::types;
 
 pub(crate) struct CffiBuf<'p> {
     pyobj: pyo3::Bound<'p, pyo3::PyAny>,
+    #[cfg(not(Py_3_11))]
     _bufobj: pyo3::Bound<'p, pyo3::PyAny>,
+    #[cfg(Py_3_11)]
+    _bufobj: Option<pyo3::buffer::PyBuffer<u8>>,
     buf: &'p [u8],
 }
 
+#[cfg(not(Py_3_11))]
 fn _extract_buffer_length<'p>(
     pyobj: &pyo3::Bound<'p, pyo3::PyAny>,
     mutable: bool,
-) -> pyo3::PyResult<(pyo3::Bound<'p, pyo3::PyAny>, usize)> {
+) -> pyo3::PyResult<(pyo3::Bound<'p, pyo3::PyAny>, usize, usize)> {
     let py = pyobj.py();
     let bufobj = if mutable {
         let kwargs = [(pyo3::intern!(py, "require_writable"), true)].into_py_dict(py)?;
@@ -43,13 +50,44 @@ fn _extract_buffer_length<'p>(
         .call1((pyo3::intern!(py, "uintptr_t"), bufobj.clone()))?
         .call_method0(pyo3::intern!(py, "__int__"))?
         .extract::<usize>()?;
-    Ok((bufobj, ptrval))
+    let len = bufobj.len()?;
+    Ok((bufobj, ptrval, len))
+}
+
+#[cfg(Py_3_11)]
+fn _extract_buffer_length(
+    pyobj: &pyo3::Bound<'_, pyo3::PyAny>,
+    mutable: bool,
+) -> pyo3::PyResult<(Option<pyo3::buffer::PyBuffer<u8>>, usize, usize)> {
+    let buf = pyo3::buffer::PyBuffer::get(pyobj)
+        .map_err(|_| {
+            let errmsg = if pyobj.is_instance_of::<pyo3::types::PyString>() {
+                format!(
+                    "Cannot convert \"{}\" instance to a buffer.\nDid you mean to pass a bytestring instead?",
+                    pyobj.get_type()
+                )
+            } else {
+                format!("Cannot convert \"{}\" instance to a buffer.", pyobj.get_type())
+            };
+            pyo3::exceptions::PyTypeError::new_err(errmsg)
+        })?;
+
+    if mutable && buf.readonly() {
+        todo!("return appropriate error");
+    }
+
+    let ptr = buf.buf_ptr() as usize;
+    let len = buf.len_bytes();
+    Ok((Some(buf), ptr, len))
 }
 
 impl<'a> CffiBuf<'a> {
     pub(crate) fn from_bytes(py: pyo3::Python<'a>, buf: &'a [u8]) -> Self {
         CffiBuf {
             pyobj: py.None().into_bound(py),
+            #[cfg(Py_3_11)]
+            _bufobj: None,
+            #[cfg(not(Py_3_11))]
             _bufobj: py.None().into_bound(py),
             buf,
         }
@@ -66,8 +104,7 @@ impl<'a> CffiBuf<'a> {
 
 impl<'a> pyo3::conversion::FromPyObject<'a> for CffiBuf<'a> {
     fn extract_bound(pyobj: &pyo3::Bound<'a, pyo3::PyAny>) -> pyo3::PyResult<Self> {
-        let (bufobj, ptrval) = _extract_buffer_length(pyobj, false)?;
-        let len = bufobj.len()?;
+        let (bufobj, ptrval, len) = _extract_buffer_length(pyobj, false)?;
         let buf = if len == 0 {
             &[]
         } else {
@@ -92,7 +129,10 @@ impl<'a> pyo3::conversion::FromPyObject<'a> for CffiBuf<'a> {
 
 pub(crate) struct CffiMutBuf<'p> {
     _pyobj: pyo3::Bound<'p, pyo3::PyAny>,
+    #[cfg(not(Py_3_11))]
     _bufobj: pyo3::Bound<'p, pyo3::PyAny>,
+    #[cfg(Py_3_11)]
+    _bufobj: Option<pyo3::buffer::PyBuffer<u8>>,
     buf: &'p mut [u8],
 }
 
@@ -104,9 +144,8 @@ impl CffiMutBuf<'_> {
 
 impl<'a> pyo3::conversion::FromPyObject<'a> for CffiMutBuf<'a> {
     fn extract_bound(pyobj: &pyo3::Bound<'a, pyo3::PyAny>) -> pyo3::PyResult<Self> {
-        let (bufobj, ptrval) = _extract_buffer_length(pyobj, true)?;
+        let (bufobj, ptrval, len) = _extract_buffer_length(pyobj, true)?;
 
-        let len = bufobj.len()?;
         let buf = if len == 0 {
             &mut []
         } else {
```

bdraco: I was originally trying to keep the implementations more separated, but I think switching to more fine-grained `#[cfg]`s inline like this is a better strategy overall as we can get a lot more reduction in duplication. I'm working on implementing that approach now.

bdraco: Just pushed 8107b408a. Let me know if this matches what you were envisioning or if you'd like to see any changes.

src/rust/src/buf.rs
```diff
@@ -2,55 +2,97 @@
...
 }
 
 impl<'a> CffiBuf<'a> {
     pub(crate) fn from_bytes(py: pyo3::Python<'a>, buf: &'a [u8]) -> Self {
+        #[cfg(Py_3_11)]
+        let bufobj = {
+            let py_bytes = pyo3::types::PyBytes::new(py, buf);
+            pyo3::buffer::PyBuffer::get(&py_bytes)
+                .expect("Cannot convert \"bytes\" instance to a buffer.")
+        };
```
alex, line 90: This gives us extra allocations that defeat the point of the methods :-) My PoC just makes `bufobj` be `Option<Buffer>` on py311+ and stores a none there for this reason.

bdraco: Thanks for the feedback, and thank you for your patience! I overlooked the `Option` change in your diff while focusing on the `cfg` updates. I'll make the necessary adjustments and push an updated change shortly.

bdraco: I've implemented the changes in commit 1024811bb. Thanks again for your patience and understanding!

src/rust/src/buf.rs
bdraco: conversation moved to https://github.com/pyca/cryptography/pull/12857#discussion_r2074147807

bdraco:
With the latest changes, benchmark performance remains stable—any differences appear to be within typical measurement variance.

| Benchmark | Base | PR | Delta |
| --------- | ---- | -- | ----- |
| `tests/bench/test_aead.py::test_aesccm_decrypt` | 2865.0 ns | 1425.0 ns | 50% faster |
| `tests/bench/test_aead.py::test_aesccm_encrypt` | 2895.0 ns | 1563.0 ns | 46% faster |
| `tests/bench/test_aead.py::test_aesgcm_decrypt` | 3035.0 ns | 1653.0 ns | 46% faster |
| `tests/bench/test_aead.py::test_aesgcm_encrypt` | 3025.0 ns | 1643.0 ns | 46% faster |
| `tests/bench/test_aead.py::test_aesocb3_decrypt` | 1983.0 ns | 955.8 ns | 52% faster |
| `tests/bench/test_aead.py::test_aesocb3_encrypt` | 2013.0 ns | 1062.0 ns | 47% faster |
| `tests/bench/test_aead.py::test_aessiv_decrypt` | 2008.5 ns | 1380.0 ns | 31% faster |
| `tests/bench/test_aead.py::test_aessiv_encrypt` | 2093.0 ns | 1483.0 ns | 29% faster |
| `tests/bench/test_aead.py::test_chacha20poly1305_decrypt` | 3536.0 ns | 1863.0 ns | 47% faster |
| `tests/bench/test_aead.py::test_chacha20poly1305_encrypt` | 3616.0 ns | 1854.0 ns | 49% faster |
| `tests/bench/test_ec_load.py::test_load_ec_private_numbers` | 26249.0 ns | 25488.0 ns | 3% faster |
| `tests/bench/test_ec_load.py::test_load_ec_public_numbers` | 15930.0 ns | 15068.0 ns | 5% faster |
| `tests/bench/test_fernet.py::test_fernet_encrypt` | 20699.0 ns | 16410.0 ns | 21% faster |
| `tests/bench/test_hashes.py::test_sha256` | 2595.0 ns | 1944.0 ns | 25% faster |
| `tests/bench/test_hmac.py::test_hmac_sha256` | 3687.0 ns | 2585.0 ns | 30% faster |
| `tests/bench/test_x509.py::test_aki_public_bytes` | 691.0 ns | 581.0 ns | 16% faster |
| `tests/bench/test_x509.py::test_load_der_certificate` | 2925.0 ns | 2815.0 ns | 4% faster |
| `tests/bench/test_x509.py::test_load_pem_certificate` | 14517.0 ns | 14277.0 ns | 2% faster |
| `tests/bench/test_x509.py::test_object_identifier_constructor` | 363.15 ns | 286.5 ns | 21% faster |
| `tests/bench/test_x509.py::test_verify_docs_python_org` | 189874.0 ns | 182640.0 ns | 4% faster |

alex:
All teh code looks good -- thanks for your work on this! We just need to sort out the build system/abi3 changes now.

src/rust/src/buf.rs
bdraco: Resolving per https://github.com/pyca/cryptography/pull/12857#issuecomment-2852370426

bdraco:
> All teh code looks good -- thanks for your work on this! We just need to sort out the build system/abi3 changes now.

Thanks! From what I can tell, the cleanest solution to meet your expectations would be for `maturin` to support something like an `abi3-auto` option that sets the correct `abi3-pyXXX` tag based on the installed Python version. That would avoid hardcoding and make things much more maintainable. I imagine you might have more influence with the `maturin` maintainers than I do if you think that's a worthwhile direction.

The only alternatives I see are either adding a `build.py` script to handle the logic, which feels a bit clunky, or keeping it as I’ve written it now, which I understand isn’t ideal.

Happy to adjust however you prefer. As long as downstream ends up with the better performing wheels, I'm a happy camper.

alex:
Yeah, setuptools-rust had a `abi3=auto` option, having that in maturin would be great. I'll file a bug to request it for. For now my preference would be to leave teh build stuff unchanged -- still defaulting to abi3-py37 and having py311 in wheel builder. And then we adopt auto once its available.

alex:
hysterical, apparently I asked for this last year: https://github.com/PyO3/maturin/issues/1894#issuecomment-2852639577

bdraco:
That doesn’t seem to be moving forward anytime soon, so I’ll push the PEP 517 build backend I put together. It turned out better than I expected. If it’s not what you had in mind, I’m happy to revert or make changes.

alex:
My preference would be: lets do the bare minimum in this PR, so no changes to the build backend or anything like that. Then we can have a seprate discussion about changes.

bdraco:
> My preference would be: lets do the bare minimum in this PR, so no changes to the build backend or anything like that. Then we can have a seprate discussion about changes.

Reverted in 1292ebdae1eed154cbd751b1eefbdd1d7e66f3ff

bdraco:
Updated the PR description to match the current state.

Benchmarks will show no change from `main`, which is expected because the benchmark runner builds with Python 3.7 limited wheels, so the improvements don't apply in that environment.

alex:
Ok, we need one CI job built with abi3-py311, lets me think about what the easiest way to do that would be.

bdraco:
CI is now failing because the 3.11 code path isn't exercised during coverage collection. It only runs on the PyPy runner, which doesn't collect coverage. I'll restore the `noxfile.py` changes to address this.

alex:
My suggestion would be add `@nox.session(name="tests-abi3-py311")` and use that.

noxfile.py
```diff
@@ -70,52 +84,59 @@ def tests(session: nox.Session) -> None:
...
-    if session.name != "tests-nocoverage":
+    if coverage:
         [rust_so] = glob.glob(
             f"{session.virtualenv.location}/lib/**/cryptography/hazmat/bindings/_rust.*",
             recursive=True,
         )
         process_rust_coverage(session, [rust_so], prof_location)
 
 
+@nox.session(name="tests-abi3-py311")
```
alex, line 130: Is there a reason this didn't work as another `if` statement in teh same function?

bdraco: I can do it that way instead.  Pushed 69e697641

alex (approved):
Thank you for your extreme patience with us. This is an exceptional improvement and I'm thrilled.

Next up, figuring out how to improve the build process, so even more people benefit!

bdraco:
Thank you!

PR #12872, opened 2025-05-08
fixes #12866 -- add license field to all rust crates

### LINKED ISSUE #12866: Define license for cryptography-x509
Xray detects cryptography-x509 as unknown license. Jfrog support says this is because "In the [rust code](https://github.com/pyca/cryptography/tree/main/src/rust/cryptography-x509) defined, there is no license file specified"

Image

alex:


alex:
Done with zed, prompt:

> Add license indicator to the workspace root cargo.toml, matching our existing python license, and then mark all other rust crates as inheriting the workspace license.

reaperhulk (approved):

PR #12873, opened 2025-05-08
prohibit snowballstemmer 3.0 which breaks sphinx

alex:
See: https://github.com/sphinx-doc/sphinx/issues/13533
See: https://github.com/snowballstem/snowball/issues/229


alex:
this is needed for RTD

PR #12896, opened 2025-05-13
Use the current Python version as the abi3 default when building from source

alex:


alex:
Hmm, https://github.com/PyO3/pyo3/issues/5093 discusses this issue. We could fix by setting `PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1`.

reaperhulk:
Any reason we shouldn't set this to unblock us and remove it when the changed behavior lands in the next pyo3 release?

alex:
Seems fine to me

reaperhulk (approved):

PR #12916, opened 2025-05-16
support explicit curves mapping to named curves (for some curves)

reaperhulk:
We now support loading EC keys with explicit encodings of named curves.  This requires that the explicit curve encoding match exactly (but allows the omission of the seed parameter to maintain compatibility with the -no_seed OpenSSL flag). This is only supported for secp256r1, secp384r1, secp521r1, and secp256k1 at this time.

This is not arbitrary explicit curve support; we still have no plans to implement that.

This PR also updates MSRV to 1.83 as we need that version for Option::unwrap() in const contexts and thus is blocked until we feel comfortable with that bump (As of late May 2025 75% of our sdist downloads have 1.83+)


reaperhulk:
We're going to need a 1.83 MSRV due to needing Option<T>::unwrap in const context or else we'll have to change the approach to avoid use const values.

MFM007 (approved):

reaperhulk:
5 months later we're above 90% on 1.83 so we've moved up the MSRV and this is ready for review 😄

src/rust/cryptography-x509/src/common.rs
```diff
@@ -473,7 +473,59 @@ pub const PSS_SHA512_MASK_GEN_ALG: MaskGenAlgorithm<'_> = MaskGenAlgorithm {
...
+    #[defined_by(field_type)]
+    pub parameters: FieldParameters<'a>,
+}
+
+#[derive(asn1::Asn1DefinedByRead, asn1::Asn1DefinedByWrite, Hash, Clone, PartialEq, Eq, Debug)]
+pub enum FieldParameters<'a> {
+    #[defined_by(oid::PRIME_FIELD_OID)]
+    PrimeField(asn1::BigUint<'a>),
+    #[defined_by(oid::CHARACTERISTIC_TWO_FIELD_OID)]
+    CharacteristicTwo(asn1::Tlv<'a>),
```
alex:
```suggestion
    CharacteristicTwo(asn1::Sequence<'a>),
```

src/rust/src/error.rs
```diff
@@ -70,7 +70,7 @@ impl From<cryptography_key_parsing::KeyParsingError> for CryptographyError {
             }
             cryptography_key_parsing::KeyParsingError::ExplicitCurveUnsupported => {
                 CryptographyError::Py(pyo3::exceptions::PyValueError::new_err(
-                    "ECDSA keys with explicit parameters are unsupported at this time",
+                    "ECDSA keys with explicit parameters are only supported when they map to secp256r1, secp384r1, secp521r1, or secp256k1. No custom curves are supported.",
```
alex:
```suggestion
                    "ECDSA keys with explicit parameters are only supported when they map to secp256r1, secp384r1, or secp521r1. No custom curves are supported.",
```

tests/hazmat/primitives/test_ec.py
```diff
@@ -1086,6 +1088,54 @@ def test_load_private_key_explicit_parameters(self):
                 mode="rb",
             )
 
+    @pytest.mark.parametrize(
+        ("curve", "file"),
+        [
+            # secp256k1 has no seed value
+            (ec.SECP256K1, "secp256k1-explicit-no-seed.pem"),
```
alex:
```suggestion
```

tests/hazmat/primitives/test_ec.py
```diff
@@ -1086,6 +1088,54 @@ def test_load_private_key_explicit_parameters(self):
...
+            mode="rb",
+        )
+        assert isinstance(key, ec.EllipticCurvePrivateKey)
+        assert isinstance(key.curve, curve)
+
+    @pytest.mark.parametrize(
+        ("curve", "file"),
+        [
+            # secp256k1 has no seed value
+            (ec.SECP256K1, "secp256k1-pub-explicit-no-seed.pem"),
```
alex:
```suggestion
```

CHANGELOG.rst
```diff
@@ -12,6 +12,8 @@ Changelog
   OpenSSL 3.0.0 or later is now required. LibreSSL, BoringSSL, and AWS-LC
   continue to be supported.
 * Updated the minimum supported Rust version (MSRV) to 1.83.0, from 1.74.0.
+* Added support for loading elliptic curve keys that contain explicit encodings
+  of the curves ``secp256r1``, ``secp384r1``, ``secp521r1``, and ``secp256k1``.
```
alex:
```suggestion
  of the curves ``secp256r1``, ``secp384r1``, or ``secp521r1``.
```

PR #12917, opened 2025-05-16
suppress new clippy warning

reaperhulk:


reaperhulk:
This warning doesn't actually happen on main, but only on the explicit branch. However, that branch doesn't touch this enum so I'm not sure what's going on.

alex:
uhh, if master is green I'm disinclined to land a supression that it can't trigger 😂

PR #12919, opened 2025-05-17
bump MSRV to 1.74

reaperhulk:


alex (dismissed):

alex (approved):

alex (approved):

h-vetinari:
Curious that you did this just before the release - no criticism implied, just that the whole development cycle was done with the old MSRV, so it would seem to have cost ~nothing to release with that and bump afterwards.

reaperhulk:
That's not a bad point -- we could have done it in any order and I simply didn't think about it to be honest. We did also check the download data on pypi and saw we had ~98% on 1.74+.

PR #12927, opened 2025-05-17
fix(x509): use backward compatible syntax for types

shcheklein:
An attempt to fix the issue we are hitting on `mypy` Python 3.9 (but also I think it won't work with Python 3.9 itself as well):

```
.nox/lint/lib/python3.9/site-packages/cryptography/x509/certificate_transparency.py:8: note: In module imported here,
.nox/lint/lib/python3.9/site-packages/cryptography/x509/__init__.py:7: note: ... from here,
.nox/lint/lib/python3.9/site-packages/werkzeug/serving.py:93: note: ... from here,
.nox/lint/lib/python3.9/site-packages/werkzeug/__init__.py:1: note: ... from here:
.nox/lint/lib/python3.9/site-packages/cryptography/hazmat/bindings/_rust/x509.pyi:232:7: error:
invalid syntax  [syntax]
    type MaybeExtensionValidatorCallback[T: x509.ExtensionType] = typing.C...
          ^
Found 1 error in 1 file (errors prevented further checking)

nox > Command pre-commit run --show-diff-on-failure --all-files failed with exit code 1
nox > Session lint failed.
```

It was introduced originally here: https://github.com/pyca/cryptography/commit/4cfd7b1208e3b806cfa20f8e0f8d21d01bfd768f

I would appreciate a review since I'm definitely not a Python types expert.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         PYTHON:
+          - {VERSION: "3.8", NOXSESSION: "flake"}
           - {VERSION: "3.13", NOXSESSION: "flake"}
           - {VERSION: "3.13", NOXSESSION: "rust"}
           - {VERSION: "3.12", NOXSESSION: "docs", OPENSSL: {TYPE: "openssl", VERSION: "3.5.0"}}
--- a/src/cryptography/hazmat/bindings/_rust/x509.pyi
+++ b/src/cryptography/hazmat/bindings/_rust/x509.pyi
@@ -229,7 +229,9 @@ class Criticality:
     AGNOSTIC: Criticality
     NON_CRITICAL: Criticality
 
-type MaybeExtensionValidatorCallback[T: x509.ExtensionType] = typing.Callable[
+T = typing.TypeVar("T", contravariant=True, bound=x509.ExtensionType)
+
+MaybeExtensionValidatorCallback = typing.Callable[
     [
         Policy,
         x509.Certificate,
@@ -238,12 +240,10 @@ type MaybeExtensionValidatorCallback[T: x509.ExtensionType] = typing.Callable[
     None,
 ]
 
-type PresentExtensionValidatorCallback[T: x509.ExtensionType] = (
-    typing.Callable[
-        [Policy, x509.Certificate, T],
-        None,
-    ]
-)
+PresentExtensionValidatorCallback = typing.Callable[
+    [Policy, x509.Certificate, T],
+    None,
+]
 
 class ExtensionPolicy:
     @staticmethod
@@ -255,13 +255,13 @@ class ExtensionPolicy:
     def require_not_present(
         self, extension_type: type[x509.ExtensionType]
     ) -> ExtensionPolicy: ...
-    def may_be_present[T: x509.ExtensionType](
+    def may_be_present(
         self,
         extension_type: type[T],
         criticality: Criticality,
         validator: MaybeExtensionValidatorCallback[T] | None,
     ) -> ExtensionPolicy: ...
-    def require_present[T: x509.ExtensionType](
+    def require_present(
         self,
         extension_type: type[T],
         criticality: Criticality,


shcheklein:
@deivse @alex @reaperhulk I would appreciate a review / suggestions when you a minute. Thanks 🙏

deivse:
I remember thinking if this is fine when making the original PR, yeah... I think we agreed during review that using the newer syntax shouldn't cause any issues, since it's an interface file, but it actually can, as I have encountered since. Normally mypy won't complain about this, but if you're running with min python version specified in mypy config (I think it can also pick it up from pyproject.toml), it will. I assume you have the min python version explicitly specified somewhere?

Fortunately this should not be an issue outside of type checking - the .pyi file is purely for interface definitions and type checking information. I definitely agree this should be fixed though.

alex (commented):
can you add a job to `ci.yml`, probably something like:

```
          - {VERSION: "3.8", NOXSESSION: "flake"}
```

src/cryptography/hazmat/bindings/_rust/x509.pyi
```diff
@@ -229,7 +229,9 @@ class Criticality:
     AGNOSTIC: Criticality
     NON_CRITICAL: Criticality
 
-type MaybeExtensionValidatorCallback[T: x509.ExtensionType] = typing.Callable[
+T = typing.TypeVar("T", covariant=True, bound=x509.ExtensionType)
```
deivse, line 232: I think this should actually be contravariant, not covariant - this doesn't affect usage in the method definitions, but for the type aliases, since they are callables, we want them to be contravariant, I think. Please correct me if I'm wrong - my first encounter with covariance and friends.

From what I understood, making this covariant means that wherever a MaybeExtensionValidator[A] is accepted, a MaybeExtensionValidator[B] would also be accepted, where B is derived from A. But I don't think that's a good idea. Contravariance would mean that MaybeExtensionValidator[A] would be accepted wherever MaybeExtensionValidator[B] is accepted, and that seems potentially useful in some specific cases - i.e. for example the user might want to create some extension validation callback that accepts any type and re-routes the validation to some other structure in their code.

deivse (changes requested):

src/cryptography/hazmat/bindings/_rust/x509.pyi
deivse: The [mypy docs](https://mypy.readthedocs.io/en/stable/generics.html#variance-of-generic-types:~:text=Callable%20is%20an%20example%20of%20type%20that%20behaves%20contravariant%20in%20types%20of%20arguments.%20That%20is%2C%20Callable%5B%5BShape%5D%2C%20int%5D%20is%20a%20subtype%20of%20Callable%5B%5BTriangle%5D%2C%20int%5D%2C%20despite%20Shape%20being%20a%20supertype%20of%20Triangle.%20To%20understand%20this%2C%20consider%3A) describe why contravariance makes sense for callables better than I do.

shcheklein:
@alex added, seems there are other issue on Python 3.8, probably unrelated to this one.

alex:
Grumble. Ok I think what I'll do is send a separate PR to resolve those, so you don't have to debug all of our (apparently broken) type annotations.

src/cryptography/hazmat/bindings/_rust/x509.pyi
shcheklein: right. so, do you feel we need to put explicit `contravariant=True` or leave it default?

deivse: Explicitly mark contravariant I think, because with the old type syntax it's neither by default, and the new one defaulted this to contravariant. I think it's best to keep it that way, since the use case of passing in an extension validator that accepts any extension can be valid under some very specific circumstances (and I'm pretty sure I did that in tests).

@alex can you please double check that my understanding of co/contravariance is correct when you have a chance? 😄

alex: I think leaving it blank makes it invariant, and that's fine?

alex:
Ok you should be able to merge/rebase main and get green.

src/cryptography/hazmat/bindings/_rust/x509.pyi
deivse: Imo that's a bit needlessly prohibitive since then you wouldn't be able to pass in an extension validator that accepts any extension. I'm pretty sure that would make type checking fail for the tests. The real world use cases I can come up with are extremely contrived, but I think it's worth keeping it contravariant just for the tests since it doesn't hurt anything imo.

deivse: Not sure about the tests tbh, basically either nothing or explicitly contravariant should be fine. Not a deal breaker imo.

alex:
(And we need to bring back the 3.8 flake job)

src/cryptography/hazmat/bindings/_rust/x509.pyi
alex: My understanding is that inavariant is the same behavior as the prior type annotation:

```
>>> def f[T: datetime.datetime](t: T): pass
... 
>>> f.__annotations__
{'t': T}
>>> t = f.__annotations__["t"]
>>> t.__covariant__
False
>>> t.__contravariant__
False
```

alex:
doh. missed one https://github.com/pyca/cryptography/pull/12929

shcheklein:
@alex still getting some issues on Py 3.8:

```
nox > mypy src/cryptography/ vectors/cryptography_vectors/ tests/ release.py noxfile.py
release.py:10: error: Cannot find implementation or library stub for module named "tomllib"  [import-not-found]
release.py:10: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 205 source files)
pyproject.toml: [mypy]: Unrecognized option: strict_bytes = True
nox > Command mypy src/cryptography/ vectors/cryptography_vectors/ tests/ release.py noxfile.py failed with exit code 1
```

shcheklein:
I'll rebase when it is merged ...

alex:
Ok, it's merged

src/cryptography/hazmat/bindings/_rust/x509.pyi
deivse: this is different since it's a function accepting a type, there any sort of variance is not defined since the function itself is not generic, it just accepts a generic argument. If you defined a type alias for a callable with the new syntax I'm quite sure it would be contravariant. (Or alternatively a function which accepts a generic callable.) 

The idea is that when u accept a callable you want to accept a callable for a specific type or a callable for a more generic type - all you care about is that you can call it with your type.

deivse: (at least mypy docs explicitly say that generic callables are contravariant by default with the new syntax)

.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         PYTHON:
-          - {VERSION: "3.13", NOXSESSION: "flake"}
+          - {VERSION: "3.8", NOXSESSION: "flake"}
```
alex, line 27: We want to run both, not replace

shcheklein: oops, my bad ...

alex (approved):

PR #12934, opened 2025-05-18
Convert several constants to the new declarative format

alex:


alex:
Done with zed+claude using the following prompt:

```
Identify all places in `#[pyo3::pymodule]` where constants are added to a module imperatively (i.e., in a fucntion with `module.add()`) and convert those to the new declarative form:

#[pymodule_export]
const NAME: TYPE = VALUE;
```

reaperhulk (approved):

PR #12946, opened 2025-05-19
add missing 45.0.0 changelog entry

reaperhulk:
@yossarian pointed out that we missed this breaking change https://github.com/pyca/cryptography/pull/12110

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -52,6 +52,9 @@ Changelog
   provided (previously no exception was raised), and raises a ``TypeError`` if
   the key is encrypted but no password is provided (previously a ``ValueError``
   was raised).
+* **BACKWARDS INCOMPATIBLE:** Added ``__copy__`` to the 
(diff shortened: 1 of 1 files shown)


reaperhulk:
We should probably backport this to 45.0.x as well so that any subsequent 45.0.x releases have the corrected changelog.

alex:
We're calling this backwards incompatible because external implementers need to add it? I don't think we've historically documented additions this way -- I'm happy to add it, but I"m not wild about the bold BACKWARDS INCOMPATIBLE.

We did all the ABCs right, not just RSA?

reaperhulk:
We did add it to all keys. I'm okay with saying we should document this without the BACKWARDS INCOMPATIBLE, although any time we add to an ABC we are going to break external implementations and we even document how to do this type of external implementation now 😄

alex:
:+1:

alex (approved):

PR #12957, opened 2025-05-20
ci: update wycheproof convergence branch to main

cpu:
This updates the CI machinery that fetches Wycheproof to reflect that the upstream moved the convergence branch from 'master' to 'main' (https://github.com/C2SP/wycheproof/issues/136). 

I suspect the existing code would keep working through the redirects GitHub leaves in place, but thought I'd offer this patch just in case.


reaperhulk (approved):
thanks!

PR #12978, opened 2025-05-25
fixes #12949 -- added support for decrypting des-cbc-md5 keys

### LINKED ISSUE #12949: Current key not working with version 45.0.2
Existing key that was working with cryptography version 44.0.3 no longer works with 45.0.2 (basically 45.0.0 and above)

serialization.load_pem_private_key(..) returns `Unknown Key Encryption Algorithm : 1.2.840.113549.1.5.3`

alex:
Implemented by Zed (claude sonnet 4) + a bit of cleanup by me with the following prompt:

> Add support to the existing private key parsing APIs for decrypting keys that have the OID 1.2.840.113549.1.5.3. There's an existing vector you can add to our tests for this (documented in test-vectors.rst). You can run tests + formatters + linters with `nox -e local`

### DIFF
--- a/src/cryptography/hazmat/decrepit/ciphers/algorithms.py
+++ b/src/cryptography/hazmat/decrepit/ciphers/algorithms.py
@@ -40,6 +40,11 @@ class TripleDES(BlockCipherAlgorithm):
         return len(self.key) * 8
 
 
+# Not actually supported, marker for tests
+class _DES:
+    key_size = 64
+
+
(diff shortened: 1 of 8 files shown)


alex:
sigh, our `cipher_supported()` check doesn't work because we turn the key into a triple des key, but only single des requires the legacy provider

we need some way to check if we have single des in the tests

reaperhulk (approved):

PR #12988, opened 2025-05-26
Apply assorted ruff rules

DimitriPapadopoulos:
These rules do seem to simplify/improve the source code. How about enforcing them?


alex:
Please revert the PERF401 change, it has a negative readability impact.

DimitriPapadopoulos:
Once you're used to it, readability is not that bad and really faster since you avoid memory allocation at each iteration. But I'm fine with removing PERF401.

PR #12989, opened 2025-05-26
bump mypy

DimitriPapadopoulos:
Required to fix this CI warning:
```
pyproject.toml: [mypy]: Unrecognized option: strict_bytes = True
```

Indeed option `--strict-bytes` was added in mypy 1.15:
* https://mypy.readthedocs.io/en/stable/changelog.html#strict-bytes
* https://github.com/python/mypy/blob/master/CHANGELOG.md#--strict-bytes

### DIFF
--- a/ci-constraints-requirements.txt
+++ b/ci-constraints-requirements.txt
@@ -87,7 +87,7 @@ markupsafe==2.1.5 ; python_full_version < '3.9'
     # via jinja2
 markupsafe==3.0.2 ; python_full_version >= '3.9'
     # via jinja2
-mypy==1.14.1 ; python_full_version < '3.9'
+mypy==1.15.0 ; python_full_version < '3.9'
     # via cryptography (pyproject.toml)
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Yes, I've seen that, but the CI warning with Python 3.8 should be avoided if possible:
* You shouldn't need to actually run mypy under each supported version of Python, instead you could option `--python-version=3.8` with a recent version of Python.
* But then what's the point of running mypy with different versions of Python? I may be mistaken, but the source code seems identical for all versions of Python, with a couple exceptions which seem unrelated to typing:
  https://github.com/pyca/cryptography/blob/5cf5edf1b36afc8400e5b6d3975d0b15380f3eff/release.py#L13-L16
  https://github.com/pyca/cryptography/blob/5cf5edf1b36afc8400e5b6d3975d0b15380f3eff/noxfile.py#L18-L21

DimitriPapadopoulos:
Superseded by #13001.

PR #12995, opened 2025-05-27
hoist HKDF into rust for even more speed

reaperhulk:
This PR was done via zed+claude sonnet 4 (with some cleanup and additions) using the following prompt:

Please implement HKDF in rust. You can follow the example in hmac.rs to see similar structure, but implement the general HKDF algorithm as seen in hkdf.py

### DIFF
--- a/src/rust/src/backend/kdf.rs
+++ b/src/rust/src/backend/kdf.rs
@@ -6,12 +6,14 @@
 use base64::engine::general_purpose::STANDARD_NO_PAD;
 #[cfg(CRYPTOGRAPHY_OPENSSL_320_OR_GREATER)]
 use base64::engine::Engine;
+use pyo3::prelude::PyAnyMethods;
 #[cfg(not(CRYPTOGRAPHY_IS_LIBRESSL))]
 use cryptography_crypto::constant_time;
 #[cfg(not(CRYPTOGRAPHY_IS_LIBRESSL))]
 use pyo3::types::PyBytesMethods;
 
 use crate::backend::hashes;
+use crate::backend::hmac::Hmac;
 use crate::buf::CffiBuf;
 use crate::error::{CryptographyError, CryptographyResult};
 use crate::exceptions;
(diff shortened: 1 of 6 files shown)


src/rust/src/backend/kdf.rs
```diff
@@ -447,12 +449,239 @@ impl Argon2id {
...
+        algorithm: pyo3::Py<pyo3::PyAny>,
+        length: usize,
+        salt: Option<pyo3::Py<pyo3::types::PyBytes>>,
+        info: Option<pyo3::Py<pyo3::types::PyBytes>>,
+        backend: Option<pyo3::Bound<'_, pyo3::PyAny>>,
+    ) -> CryptographyResult<Self> {
+        _ = backend;
+
+        let algorithm_bound = algorithm.bind(py);
+        let digest_size = algorithm_bound
```
alex, line 476:
```suggestion
        let digest_size = algorithm.bind(py)
```

src/rust/src/backend/kdf.rs
```diff
@@ -447,12 +449,239 @@ impl Argon2id {
...
+            .getattr(pyo3::intern!(py, "digest_size"))?
+            .extract::<usize>()?;
+
+        let max_length = 255 * digest_size;
+        if length > max_length {
```
alex, line 481: nit: need to handle overflows here.

src/rust/src/backend/kdf.rs
```diff
@@ -447,12 +449,239 @@ impl Argon2id {
...
+                    max_length
+                )),
+            ));
+        }
+
+        let salt = if let Some(salt) = salt {
+            salt
+        } else {
+            let zero_salt = vec![0u8; digest_size];
+            pyo3::types::PyBytes::new(py, &zero_salt).into()
```
alex, line 494:
```suggestion
            pyo3::types::PyBytes::new_with(py, digest_size, |_| {}).into()
```

no need to allocate

src/rust/src/backend/kdf.rs
```diff
@@ -447,12 +449,239 @@ impl Argon2id {
...
+        py: pyo3::Python<'_>,
+        algorithm: pyo3::Py<pyo3::PyAny>,
+        length: usize,
+        info: Option<pyo3::Py<pyo3::types::PyBytes>>,
+        backend: Option<pyo3::Bound<'_, pyo3::PyAny>>,
+    ) -> CryptographyResult<Self> {
+        _ = backend;
+
+        let algorithm_bound = algorithm.bind(py);
+        let digest_size = algorithm_bound
```
alex, line 590:
```suggestion
        let digest_size = algorithm.bind(py)
```

src/rust/src/backend/kdf.rs
```diff
@@ -447,12 +449,239 @@ impl Argon2id {
...
+            .getattr(pyo3::intern!(py, "digest_size"))?
+            .extract::<usize>()?;
+
+        let max_length = 255 * digest_size;
+        if length > max_length {
```
alex, line 595: overflow

src/rust/src/backend/kdf.rs
```diff
@@ -447,12 +449,239 @@ impl Argon2id {
...
+        let h_prime = Hmac::new_bytes(py, key_material.as_bytes(), algorithm_bound)?;
+        while output.len() < self.length {
+            let mut h = h_prime.copy(py)?;
+            h.update_bytes(&previous_output)?;
+            h.update_bytes(self.info.as_bytes(py))?;
+            h.update_bytes(&[counter])?;
+
+            let block = h.finalize(py)?;
+            let block_bytes = block.as_bytes();
+            previous_output = block_bytes.to_vec();
```
alex, line 643: We don't want to convert to vec, we want to just have each iteration slice the back off of previous output

src/rust/src/backend/kdf.rs
```diff
@@ -447,12 +449,239 @@ impl Argon2id {
...
+            output.extend_from_slice(block_bytes);
+
+            counter += 1;
+        }
+
+        output.truncate(self.length);
+        Ok(pyo3::types::PyBytes::new(py, &output))
```
alex, line 650: We can save an allocation by using `PyBytes::new_with(self.length, |b| { ... })` and then writing directly into the output.

tests/hazmat/primitives/utils.py
```diff
@@ -353,7 +353,7 @@ def hkdf_extract_test(backend, algorithm, params):
         backend=backend,
     )
 
-    prk = hkdf._extract(binascii.unhexlify(params["ikm"]))
+    prk = hkdf._extract(binascii.unhexlify(params["ikm"]))  # type: ignore[attr-defined]
```
alex, line 356: Uhh, is there a reason we have this test? I'd rather stop exposing this random (trivial) implementation detail

reaperhulk: We probably only have it because it was possible to test intermediate values so we did. I'll remove it in a separate PR and then rebase this one.

alex:
(you've got failing tests)

src/rust/src/backend/kdf.rs
reaperhulk: Revisiting this, are you looking for something like
```rust
        let max_length = match 255usize.checked_mul(digest_size) {
            Some(max_len) => max_len,
            None => {
                return Err(CryptographyError::from(
                    pyo3::exceptions::PyValueError::new_err(
                        "Digest size too large, would cause overflow in max length calculation"
                    ),
                ));
            }
        };
```
to protect against user-supplied hash function objects causing overflow?

alex: Yeah, you can make it a bit shorter with `let max_length = 255.checked_mul(digest_size).ok_or_else(|| pyo3::exceptions::PyValueError::new_err(...))?;`

src/rust/src/backend/kdf.rs
```diff
@@ -8,10 +8,11 @@ use base64::engine::general_purpose::STANDARD_NO_PAD;
 use base64::engine::Engine;
 #[cfg(not(CRYPTOGRAPHY_IS_LIBRESSL))]
 use cryptography_crypto::constant_time;
-#[cfg(not(CRYPTOGRAPHY_IS_LIBRESSL))]
+use pyo3::prelude::PyAnyMethods;
 use pyo3::types::PyBytesMethods;
```
alex:
```suggestion
use pyo3::types::{PyAnyMethods, PyBytesMethods};
```

src/rust/src/backend/kdf.rs
```diff
@@ -443,12 +444,248 @@ impl Argon2id {
...
+        py: pyo3::Python<'_>,
+        key_material: CffiBuf<'_>,
+        expected_key: CffiBuf<'_>,
+    ) -> CryptographyResult<()> {
+        let actual = self.derive(py, key_material)?;
+        let actual_bytes = actual.as_bytes();
+        let expected_bytes = expected_key.as_bytes();
+
+        if actual_bytes.len() != expected_bytes.len()
+            || !openssl::memcmp::eq(actual_bytes, expected_bytes)
```
alex: there's a utility function for this.

src/rust/src/backend/kdf.rs
```diff
@@ -443,12 +444,248 @@ impl Argon2id {
...
+        {
```
alex: same

PR #13001, opened 2025-05-28
A couple mypy/ruff fixes

### LINKED ISSUE #13000: No need to run `flake` test session with multiple Python versions?
The `flake` test session runs:
* the ruff linter
* the ruff formatter
* mypy
* check-sdist

https://github.com/pyca/cryptography/blob/f5fe8209d85a0c9a030749652402ee1c3090ac45/noxfile.py#L205-L215

This test session is run twice during CI, with two different versions of Python:

https://github.com/pyca/cryptography/blob/f5fe8209d85a0c9a030749652402ee1c3090ac45/.github/workflows/ci.yml#L26-L27

I don't see the need for that:
* [ruff](https://docs.astral.sh/ruff/) is a Rust program that does not depend on the available version of Python. Instead, the target Python version is defined in:
  https://github.com/pyca/cryptography/blob/f5fe8209d85a0c9a030749652402ee1c3090ac45/pyproject.toml#L48
* While [mypy](https://mypy.readthedocs.io/en/stable/) does target the version of Python that is running mypy by default, the version of Python does not seem relevant:
  * Option `--python-version` could be used to modify the target version of Python.
  * Most importantly, _cryptography_ lacks checks based on  [sys.version_info](https://docs.python.org/3/library/sys.html#sys.version_info), except for a couple checks detailed in https://github.com/pyca/cryptography/pull/12989#issuecomment-2910574775 that are irrelevant for typing. Such checks define different code paths when targeting different versions of Python according to section [Python version and system platform checks](https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks) of the documentation.
  * Even if there were more code paths specific to Python 3.8, testing typing with previous versions of Python doesn't seem useful. Such “legacy”  code paths usually do not change much until they disappear and should have been tested at some point in the past with previous versions of Python.
* I understand the version of Python used to run [check-sdist](https://pypi.org/project/check-sdist/) is irrelevant too.

I suggest running the `flake` test session with Python 3.13 only. If you insist on running mypy with different target versions of Python, I suggest splitting it into `ruff`, `mypy` and `check-sdist` sessions.

DimitriPapadopoulos:
~~Fixes #13000.~~
~~Supersedes #12989.~~

**EDIT:**
* Bump ruff/mypy in `pyproject,toml` to match versions in `ci-constraints-requirements.txt`.
* Option `warn-unused-configs = true` requires `incremental = false` according to the [`--warn-unused-configs`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-warn-redundant-casts) documentation:
  > This requires turning off incremental mode using [`--no-incremental`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-incremental).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,7 +86,7 @@ docstest = [
 sdist = ["build >=1.0.0"]
 # `click` included because its needed to type check `release.py`
 pep8test = [
-    "ruff >=0.3.6",
+    "ruff >=0.11.11",
     "mypy >=1.14",
     "check-sdist",
     "click >=8.0.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,7 +87,7 @@ sdist = ["build >=1.0.0"]
 # `click` included because its needed to type check `release.py`
 pep8test = [
     "ruff >=0.3.6",
-    "mypy >=1.4",
+    "mypy >=1.14",
     "check-sdist",
     "click >=8.0.1",
 ]
@@ -138,6 +138,7 @@ markers = [
 [tool.mypy]
 show_error_codes = true
 check_untyped_defs = true
+incremental = false
 no_implicit_reexport = true
 warn_redundant_casts = true
 warn_unused_ignores = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -138,7 +138,6 @@ markers = [
 [tool.mypy]
 show_error_codes = true
 check_untyped_defs = true
-incremental = false
 no_implicit_reexport = true
 warn_redundant_casts = true
 warn_unused_ignores = true
--- a/noxfile.py
+++ b/noxfile.py
@@ -202,8 +202,8 @@ def flake(session: nox.Session) -> None:
         *pyproject_data["project"]["optional-dependencies"]["nox"],
     )
 
-    session.run("ruff", "check", ".")
-    session.run("ruff", "format", "--check", ".")
+    session.run("ruff", "check")
+    session.run("ruff", "format", "--check")
     session.run(
         "mypy",
         "src/cryptography/",
@@ -286,8 +286,8 @@ def local(session: nox.Session):
         verbose=False,
     )
 
-    session.run("ruff", "format", ".")
-    session.run("ruff", "check", ".")
+    session.run("ruff", "format")
+    session.run("ruff", "check")
 
     session.run("cargo", "fmt", "--all", external=True)
     session.run("cargo", "check", "--all", "--tests", external=True)


alex:
I don't believe that's correct. As the original issue describes, the issue that this test was added to catch reproduces on 3.9, which supports mypy 1.15. (You'll also see that the changelog for mypy 1.15 doesn't mention this feature: https://mypy-lang.blogspot.com/2025/02/mypy-115-released.html).

The point of this test is to verify that users will be able to typecheck their programs against cryptography.

AFAIK there's no way to resolve that warning without regressing this.

DimitriPapadopoulos:
> It's about supporting all the configurations we expect to work for our users.

Ate they expected to work for end-users of the _cryptography_ module or developers/users that run mypy on the source code? That's the distinction I am making above.

DimitriPapadopoulos:
My point was that it's not useful to let developers run mypy under Python 3.8. However, after reading [Issues with code at runtime](https://mypy.readthedocs.io/en/stable/common_issues.html), I realise there may be **runtime** issues after all:
> Idiomatic use of type annotations can sometimes run up against what a given version of Python considers legal code.

I wonder whether these might be addressed using [`--python-version 3.8`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-python-version): is that option the same as running under Python 3.8, but without the CI warning?

Any way, let me read more about it and stop annoying you. I'll modify this PR to apply two very simple changes:
* commit 4a8e8410ddc717c4f7c19f264f177b5331c946d8 for ruff,
* turn off incremental mode as per the [`--warn-unused-configs`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-warn_unused_configs) documentation.

pyproject.toml
```diff
@@ -138,6 +138,7 @@ markers = [
 [tool.mypy]
 show_error_codes = true
 check_untyped_defs = true
+incremental = false
```
alex, line 141: this pessimizes the experience for local runs, so I don't think we want it.

DimitriPapadopoulos: I admit I don't know much about mypy internals, but the documentation of [`warn_unused_configs`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-warn_unused_configs) suggests it is required:
> This requires turning off incremental mode using [`incremental = False`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-incremental).

alex: If we have to choose between them, I think we'd prefer to keep incremental. (But I suspect what this really means is that unused config warnings only happen on initial runs, and are ignored on subsequent incremental runs).

DimitriPapadopoulos: Which is OK because the cache is gone when restarting CI tests?

alex: That's my guess. In any event, we need incremental to remain on. So I would drop this change.

alex:
were you interested in making the change here?

PR #13011, opened 2025-05-30
fixed #12965 -- document that OPENSSL_DIR is the way to build with boring or aws-lc

### LINKED ISSUE #12965: Build python cryptography with aws-lc as a backend cryptographic provider
Hi,

Python cryptography has released support for building with [aws-lc](https://github.com/aws/aws-lc) in 45.0.0 release. There is no official documentation available to try this out yet. Today I tried to build cryptography with [aws-lc](https://github.com/aws/aws-lc) as a backend crypto provider and was successfully able to do it. Creating this issue to provide a pull request for building with aws-lc as cryptographic provider.

Created this thread on the email list earlier - https://mail.python.org/archives/list/cryptography-dev@python.org/thread/VIRPURXDUF4LRLVSISI4BJZRGEPKWKLS/

alex:
hard work done by claude

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -234,6 +234,12 @@ dependencies.
     cd ..
     OPENSSL_DIR="${CWD}/openssl" pip wheel --no-cache-dir --no-binary cryptography cryptography
 
+.. note::
+
+    To build against BoringSSL or AWS-LC instead of OpenSSL, you can set the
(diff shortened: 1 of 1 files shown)


reaperhulk:
This doesn’t look like the correct place to put this note. It applies to all platforms so maybe we should just have an “boring/libre/aws-lc target” section that has this body?

PR #13016, opened 2025-05-30
remove nonsense constant

alex:


alex:
Oh, I this dates back to when we accidentally put `salt_length` on `MGF1` instead of `PSS`. That's been literally a decade.

reaperhulk:
Yeah I guess we forgot to remove this when we finished that deprecation.

reaperhulk (approved):

PR #13022, opened 2025-06-01
Use mitmproxy's uv.lock for downstream tests

mhils:
This should make things more stable going forward. :-)

### DIFF
--- a/.github/downstream.d/mitmproxy.sh
+++ b/.github/downstream.d/mitmproxy.sh
@@ -6,7 +6,7 @@ case "${1}" in
         git clone --depth=1 https://github.com/mitmproxy/mitmproxy
         cd mitmproxy
         git rev-parse HEAD
-        uv pip install --system --group dev -e .
+        uv pip install --system --requirements <(uv export) -e .
         ;;
     run)
         cd mitmproxy


.github/downstream.d/mitmproxy.sh
```diff
@@ -6,7 +6,7 @@ case "${1}" in
         git clone --depth=1 https://github.com/mitmproxy/mitmproxy
         cd mitmproxy
         git rev-parse HEAD
-        uv pip install --system --group dev -e .
+        uv pip install --system --requirements <(uv export) -e .
```
alex: Given we're already doing `uv pip install`, can we just install directly from the `uv.lock`?

mhils: `uv pip install` does not seem to have any support for uv.lock as far as I can see, and `uv sync` does not like installing into the system env. I considered echoing the venv into GITHUB_ENV, but that did not seem simpler than what we have now?

alex: nah, if there's not a trivial way we'll just roll with this

alex:
```suggestion
        uv pip install --system --requirements <(uv export --group dev) -e .
```

Should we do this?

mhils: `dev` is a default group, so those dependencies are already included. We can do this if you think it's more readable, but it's functionally equivalent. :)

alex (approved):

PR #13044, opened 2025-06-06
Return ValueError instead of panicking in x509 encode_authority_key_identifier

sekuso:
In the vanishingly rare case that Python passes invalid bytes as serials, send back a ValueError exception instead of panicking.

Adding a test case is probably impractical. This is an extreme edge case that only applies when dealing with really messy bulk data.


alex:
Can you explain the scenario in which you think a panic is possible here?

sekuso:
As far as I can tell in normal Python usage it shouldn't be possible. The panic I got here came from serializing `public_bytes` via PySpark in a UDF, and for lack of a better understanding, I think the only reason it's possible is because the bytes aren't being serialized in true Python.

alex:
Are you saying that serial number is some sort of integer subclass that has different behavior on its `bit_length` or `to_bytes` method?

PR #13081, opened 2025-06-15
fixes #13080 -- test and build wheels for musl armv7l

### LINKED ISSUE #13080: Wheel package of musl armv7l architecture
I found that cryptography only provides x86_64 and aarch64 architecture support on musl. In some embedded Linux devices, due to the volume problem, they will use musl as the c library, such as openwrt, etc. 

I want to know why cryptography does not currently provide the wheel pre-compiled package of musl armv7l architecture. I can work normally through cross-compilation, but the construction process is quite complicated. Is this a technical limitation?

alex:


alex:
this is going to be blocked on https://github.com/rust-lang/rustup/issues/3398

alex:
at this point going to close this, it's going to atrophy until the target is promoted

PR #13137, opened 2025-07-02
Switch to uv build backend

alex:
It's probably faster, I didn't measure.


sthen:
This is quite a nuisance for at least some OS packagers, building from the uv_build sdist on pypi requires downloading source during the build, which is not allowed by build infrastructure on many OS, and building the python pieces from the original uv source distribution looks rather fiddly too.

As building cryptography_vectors using python -m build only takes a matter of seconds with the (much easier to use downstream) flit_core, could this be reconsidered please?

sthen:
..also I see that the pypi sdist has incorrect (1970-01-01) timestamps which may also be due to this

alex:
Given the popularity of uv, I'm not sure "nothing can use uv-build" is a realistic path forward. Have you filed issues upstream with uv? I think getting them to resolve them is the best path forward because it fixes this for everyone. We think distributions are important, so if you file those and flag them here, I'm more than happy to reach out to folks at Astral to push them on implementing.

We spend a _lot_ of time waiting on local builds and on CI, so literally anything that speeds up our devloop is really high ROI for us as OSS maintainers, so I'm hoping we can channel our energy into getting things fixed upstream.

sthen:
Ignoring packaging issues for now (and yes we will need that sometimes, but it doesn't seem a high priority for now, as some other prominent OS don't have uv_build either, and at least Debian doesn't have uv at all in native packages yet), if I run "python -m build" in a venv with cryptography_vectors extracted (on a machine where uv_build can be built, but isn't in the set of machine-dependent wheels on https://pypi.org/project/uv-build/#files), it takes around 20min here, compared to about 10 seconds for 45.0.1 (obviously flit_core is pure-python so wheels are available everywhere). So if your CI tooling is running on OS other than Linux/MacOS that may be a consideration too.

(Will look at uv_build anyway but the build times are rather frustrating, especially when it seems many iterations will be needed to get it to work)

alex:
Sorry to make sure I understand the scenario, the comparison is between installing fit-core + cryptography-vectors vs. building uv-build from source + cryptography-vectors?

sthen:
Starting from a clean venv, pip install build, extract cryptography_vectors source, python -m build. I guess this is a fairly common scenario for a CI type workflow?

Depending on the OS/cpu arch, that will download a pre-built uv_build if available on pypi (yes that's obviously fairly quick on an arch where that binary is available), or build it if not.

So comparing those commands using 46.0.1, between an OS where uv_build is available on pypi, and one where it isn't, you either have a few seconds, or possibly ~20 minutes.

alex:
Yeah I think the high order bit here is: if you're on a platform that doesn't have a uv-build wheel, there's an expensive compilation phase. That said, they have a _very_ extensive set of wheels. In fact, I'm reasonably certain they have wheels for every single platform we test cryptography on. Can you share what platform you're using here?

sthen:
I'm using OpenBSD - in terms of cryptography, I woudn't really expect you to be testing there, unless you're doing libressl tests regularly (where it's probably the easiest OS).

pypi has binary uv_build wheels for macos, linux, windows only - no freebsd, solaris derivatives or openbsd. (it's unlikely that pypi will ever have them for openbsd as it's too much of a moving target).

After talking to another OpenBSD port dev we've managed to come up with a working way to build and use uv_build in ports. Comparing times of cryptography_vectors 46.0.1 builds with uv_build and flit_core (i.e. with #13137 backed out), and with uv_build preinstalled, I'm seeing build times of 9 seconds using uv_build vs 11 seconds for flit_core.

alex:
Thanks for the context -- for some reason I'd inferred that this was about an obscure CPU architecture, not simply a different OS. My OpenBSD-using friends would never let me live it down if I compared it to PA-RISC or something :-)

Glad to hear you guys were able to find a build solution that works for the ports process!

While 2 seconds might not sound like a ton, depending on how warm caches are and other such things, it can be up to 12% of a test run locally.

omnivagant:
In Alpinelinux `uv` is not yet available to build packages in `main/`, like `main/py3-cryprography`, but reverting this for now was easy enough. Not advocating that you should but, if someone runs into this, here it is https://git.alpinelinux.org/aports/tree/main/py3-cryptography/vectors-build-backend-flit_core_patch?id=d48c3b9fc84940df91171116cb51ec293be60c11

PR #13145, opened 2025-07-04
Drop support for OpenSSL 1.1.x

alex:


alex:
(Probably don't merge yet due to the coverage nonsense we're seeing on main)

reaperhulk (approved):

PR #13215, opened 2025-07-27
remove coverage exclude: this is coverage's default

alex:


alex:
doh, this behavior is new in 7.10.0 and we don't have that 3.8

PR #13240, opened 2025-08-04
Switch paramiko.sh to be fully uv-based

bitprophet:
This is ~= what I do in local dev after the 4.0 update. I have some local idiosyncracies re: where UV sticks its venv, and whether the venv has pip in it (I create the venvs via `uv venv --seed`), _but_ I don't believe either of those matter for this environment/use case.

### DIFF
--- a/.github/downstream.d/paramiko.sh
+++ b/.github/downstream.d/paramiko.sh
@@ -5,9 +5,7 @@ case "${1}" in
         git clone --depth=1 https://github.com/paramiko/paramiko
         cd paramiko
         git rev-parse HEAD
-        uv pip install pip
-        uv pip install -e . --group dev
-        pip install . --group dev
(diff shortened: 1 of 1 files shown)


bitprophet:
Not immediately obvious from the workflows but [this uv warning](https://github.com/pyca/cryptography/actions/runs/16726800027/job/47344479980?pr=13240#step:7:21) implies that ".venv" is not the same as whatever uv in this env is trying to use for its implicit venv. I do like its suggestion of `--active` but I also need to make sure it doesn't actually blow away the rest of the venv.

Checking for other flags now but may just decide to drop this in favor of what Alex did just prior, since that clearly worked fine.

EDIT: yea let's try `--active --inexact`, may also want `--locked/--frozen` but that's orthogonalish.

PR #13253, opened 2025-08-06
Use manylinux-entrypoint script for linux wheel builds

ngoldbaum:
See https://github.com/pypa/manylinux/issues/1825#issuecomment-3142130844 for an explanation of why we need to use this.

PR'd separately from #12555 per https://github.com/pyca/cryptography/pull/12555#discussion_r2255656343

### DIFF
--- a/.github/workflows/wheel-builder.yml
+++ b/.github/workflows/wheel-builder.yml
@@ -156,9 +156,12 @@ jobs:
               PY_LIMITED_API="--config-settings=build-args=--features=pyo3/abi3-${{ matrix.PYTHON.ABI_VERSION }}"
           fi
 
+          # The manylinux-entrypoint script accounts for uname issues, otherwise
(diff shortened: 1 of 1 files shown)


alex:
This makes sense and seems simple enough. My own question is: how did armv7 _ever_ work without this??? Do you have any idea?

alex (approved):

ngoldbaum:
> Do you have any idea?

abi3 builds are considered cross-compiles, so they go through `get_platform_tag()` in maturin, which already accounts for this:

https://github.com/PyO3/maturin/blob/df427600d13c9b3212c453b093028cef8fad9d30/src/target/mod.rs#L347-L352

The new cp314t wheel goes through a different code path in maturin, which defers to what's in sysconfig to calculate the ABI tags.

alex:
Ahh, that's teh magic bit. Thanks much!

PR #13259, opened 2025-08-06
test to see if we can run x86_64 on macos-15

reaperhulk:


reaperhulk:
This silently uses arm64 for everything.

PR #13271, opened 2025-08-09
Add __eq__ to HashAlgorithm and padding instances

mathiasertl:
This PR:

1. Adds `__eq__()` to all subclasses of HashAlgorithm.
2. Adds `__eq__()` to all padding classes.

My primary use case is testing. First, it's really un-pythonic to me (obviously very subjective, but it feels just not right to me) to write `assert isinstance(cert.cert.signature_hash_algorithm, hashes.SHA256())`, when I could write `assert cert.cert.signature_hash_algorithm == hashes.SHA256()`. But once you start comparing to other data structures, it gets even uglier:

```python
def test_something() -> None:
    ...
    assert isinstance(cert.signature_hash_algorithm, type(other.signature_hash_algorithm))
```

I think this just looks a lot better and readable:

```python
def test_something() -> None:
    ...
    assert cert.signature_hash_algorithm == other.signature_hash_algorithm
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -68,6 +68,8 @@ Changelog
   :class:`~cryptography.hazmat.primitives.ciphers.aead.AESSIV`, and
   :class:`~cryptography.hazmat.primitives.ciphers.aead.ChaCha20Poly1305` to
   allow encrypting directly into a pre-allocated buffer.
+* Builtin hash classes and instances of classes in
+  :mod:`~cryptography.hazmat.primitives.asymmetric.padding` can now be compared with `==`
 
 .. _v46-0-3:
 
--- a/src/cryptography/hazmat/primitives/asymmetric/padding.py
+++ b/src/cryptography/hazmat/primitives/asymmetric/padding.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import abc
+import typing
 
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives._asymmetric import (
@@ -15,6 +16,9 @@ from cryptography.hazmat.primitives.asymmetric import rsa
 
 class PKCS1v15(AsymmetricPadding):
     name = "EMSA-PKCS1-v1_5"
+
+    def __eq__(self, other: typing.Any) -> bool:
+        return isinstance(other, PKCS1v15)
 
 
 class _MaxLength:
@@ -56,6 +60,17 @@ class PSS(AsymmetricPadding):
 
         self._salt_length = salt_length
 
+    def __eq__(self, other: typing.Any) -> bool:
+        if not isinstance(other, PSS):
+            return False
+
+        if isinstance(self._salt_length, int):
+            eq_salt_length = self._salt_length == other._salt_length
+        else:
+            eq_salt_length = self._salt_length is other._salt_length
+
+        return eq_salt_length and self._mgf == other._mgf
+
     @property
     def mgf(self) -> MGF:
         return self._mgf
@@ -77,6 +92,14 @@ class OAEP(AsymmetricPadding):
         self._algorithm = algorithm
         self._label = label
 
+    def __eq__(self, other: typing.Any) -> bool:
+        return (
+            isinstance(other, OAEP)
+            and self._mgf == other._mgf
+            and self._algorithm == other._algorithm
+            and self._label == other._label
+        )
+
     @property
     def algorithm(self) -> hashes.HashAlgorithm:
         return self._algorithm
@@ -89,6 +112,13 @@ class OAEP(AsymmetricPadding):
 class MGF(metaclass=abc.ABCMeta):
     _algorithm: hashes.HashAlgorithm
 
+    @abc.abstractmethod
+    def __eq__(self, other: typing.Any) -> bool:
+        """
+        Implement equality checking.
+        """
+        ...
+
 
 class MGF1(MGF):
     def __init__(self, algorithm: hashes.HashAlgorithm):
@@ -96,6 +126,9 @@ class MGF1(MGF):
             raise TypeError("Expected instance of hashes.HashAlgorithm.")
 
         self._algorithm = algorithm
+
+    def __eq__(self, other: typing.Any) -> bool:
+        return isinstance(other, MGF1) and self._algorithm == other._algorithm
 
 
 def calculate_max_pss_salt_length(
--- a/tests/hazmat/primitives/test_rsa.py
+++ b/tests/hazmat/primitives/test_rsa.py
@@ -7,6 +7,7 @@ import binascii
 import copy
 import itertools
 import os
+import typing
 
 import pytest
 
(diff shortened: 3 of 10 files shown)


CHANGELOG.rst
```diff
@@ -18,6 +18,18 @@ Changelog
 * Removed the deprecated ``CAST5``, ``SEED``, ``IDEA``, and ``Blowfish``
   classes from the cipher module. These are still available in
   :doc:`/hazmat/decrepit/index`.
+* Make instances of
+  :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm` as well as
+  instances of classes in
+  :mod:`~cryptography.hazmat.primitives.asymmetric.padding`
+  comparable.
+* Added `salt_length` property to
```
reaperhulk: Let's put these in a separate PR. Whether or not we want to do `__eq__` is possibly debateable, but if you want these properties to be public that's a separate discussion.

mathiasertl: Hi @reaperhulk 

Thanks for the feedback - I've updated this PR to only include the `__eq__` instances. I'll make a separate PR for the extra properties.

kr, Mat

reaperhulk:
Looking at the test failures here the issue is that scapy creates its own HashAlgorithm, which now fails because they don't implement `__eq__`. We could avoid this by making `__eq__` a concrete method that raises NotImplementedError, but we'd need to cover those lines in a test as well.

mathiasertl:
Or it could be a breaking change and that lib would need to implement the method.

Open to implement any solution, just let me know what you prefer!

reaperhulk:
@alex do you have an opinion here? I'm reluctant to spend breakage budget on this when we could avoid it with a concrete implementation, but I'm interested in other opinions.

alex:
I'm pretty ambivalent about this. We have a few choices:

1) Simply implement `__eq__` for our own hashes, don't change the contract (all objects always have `__eq__`).
2) Add `__eq__` to the ABC and require people to implement it: breaking change for people defining their own hashes
3) Add `__eq__` to the ABC and provide a default impl: works for some users with their own hashes, but is _wrong_ for anyone with their own hash which has any attributes (e.g. `blake2`).


Of these I think (1) is probably best.

reaperhulk:
For 3 the default impl could raise NotImplementedError so you wouldn’t need to have a potentially incorrect impl

alex:
That can break existing applications, to the extent they're relying on the default `object.__eq__` behavior!

kmk142789 (approved):

kmk142789 (approved):

mathiasertl:
I'm also ambivalent, I defer to your call and will adapt as requested.

Just option three (default Implementation) seems like a bad idea to me.

Mat

reaperhulk:
After thinking for too long, I'm okay with option 1 here.

reaperhulk:
Are you still interested in this PR? Happy to review/land with option 1 + a rebase to fix the changelog conflict

src/cryptography/hazmat/primitives/asymmetric/padding.py
```diff
@@ -56,6 +60,18 @@ def __init__(
...
         self._salt_length = salt_length
 
+    def __eq__(self, other: typing.Any) -> bool:
+        if isinstance(self._salt_length, int):
+            eq_salt_length = self._salt_length == other._salt_length
+        else:
+            eq_salt_length = self._salt_length is other._salt_length
+
+        return (
+            isinstance(other, PSS)
```
alex: We need to do this check first (otherwise you get an AttributeError) and return NotImplemented

src/cryptography/hazmat/primitives/asymmetric/padding.py
```diff
@@ -77,6 +93,14 @@ def __init__(
         self._algorithm = algorithm
         self._label = label
 
+    def __eq__(self, other: typing.Any) -> bool:
+        return (
+            isinstance(other, OAEP)
```
alex, line 97: Same issue here.

CHANGELOG.rst
```diff
@@ -68,6 +68,11 @@ Changelog
   :class:`~cryptography.hazmat.primitives.ciphers.aead.AESSIV`, and
   :class:`~cryptography.hazmat.primitives.ciphers.aead.ChaCha20Poly1305` to
   allow encrypting directly into a pre-allocated buffer.
+* Make instances of
+  :class:`~cryptography.hazmat.primitives.hashes.HashAlgorithm` as well as
+  instances of classes in
+  :mod:`~cryptography.hazmat.primitives.asymmetric.padding`
+  comparable.
```
alex: This isn't quite right for the behavior we implemented. I'd probably document it as "builtin hash classes can now be compared with `==`"

alex (commented):
I think a question is whether we should be returning `NotImplemented` or `False` for diffefrent types

mathiasertl:
> I think a question is whether we should be returning `NotImplemented` or `False` for diffefrent types

I don't know of any classes that do that. Standard python certainly doesn't:

```
>>> {} == 0
False
```

src/cryptography/hazmat/primitives/asymmetric/padding.py
mathiasertl: Right, on it!

src/cryptography/hazmat/primitives/asymmetric/padding.py
mathiasertl: Actually not correct in this case:

```
>>> h = hashes.SHA256()
>>> mgf = MGF1(h)
>>> OAEP(mgf, h, None) == h
False
```

CHANGELOG.rst
mathiasertl: agreed. will be updated.

alex:
```
>>> {}.__eq__(0)
NotImplemented
```

:-)

mathiasertl:
> ```
> >>> {}.__eq__(0)
> NotImplemented
> ```
> 
> :-)

To quote [official documentation](https://docs.python.org/3/library/constants.html#NotImplemented) - emphasis mine:

> A rich comparison method **may** return the singleton [NotImplemented](https://docs.python.org/3/library/constants.html#NotImplemented) if it does not implement the operation for a given pair of arguments. By convention, False and True are returned for a successful comparison.

So I can implement it, it's just lots of extra code with no benefit that I'm aware of.

tests/hazmat/primitives/test_rsa.py
```diff
@@ -70,6 +71,9 @@ class DummyMGF(padding.MGF):
     _salt_length = 0
     _algorithm = hashes.SHA256()
 
+    def __eq__(self, other: typing.Any) -> bool:
+        return isinstance(other, DummyMGF)
+
```
mathiasertl, line 76: This class is more-or-less duplicated in `test_openssl` as well!

I would offer to move it to `tests.doubles.py` for shorter code - if you approve.

alex: Sounds great, thanks

reaperhulk:
Sorry to let this drop off our radar. After discussion @alex and I still think there's value in supporting `__eq__`, but we do want the `NotImplemented` behavior to match our other `__eq__` impls across the codebase.

mathiasertl:
Hi there!

 This also dropped of my radar, so no worries! I'll try to follow up soon, but busy in June.  I hope July is still okay.

Mat

reaperhulk:
https://github.com/pyca/cryptography/pull/15153 takes this over, thanks!

PR #13295, opened 2025-08-12
fixes #13293 -- clearly document that we care about rfc 5280 x.509

### LINKED ISSUE #13293: Docs should reflect that RFC 5280 policy is binding on X.509 components
Despite some features that provide for policy flexibility (custom extension verification), if the intent of the project is only to support internet PKI it would be nice if there was a very clear statement in the docs to that effect.  If cryptography isn't an appropriate path forward from pyOpenSSL for some use cases, it's better that this is as clear as possible so we don't waste your time with tickets.

alex:
and that enabling webpki is our main focus

### DIFF
--- a/docs/x509/index.rst
+++ b/docs/x509/index.rst
@@ -1,9 +1,13 @@
 X.509
 =====
 
-X.509 is an ITU-T standard for a `public key infrastructure`_. X.509v3 is
-defined in :rfc:`5280` (which obsoletes :rfc:`2459` and :rfc:`3280`). X.509
-certificates are commonly used in protocols like `TLS`_.
+X.509 is a standard for `public key infrastructure`_. ``cryptography``
+implements X.509 in accordance with :rfc:`5280`, is principally focused on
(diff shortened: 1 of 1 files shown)


docs/x509/index.rst
```diff
@@ -1,9 +1,13 @@
 X.509
 =====
 
-X.509 is an ITU-T standard for a `public key infrastructure`_. X.509v3 is
-defined in :rfc:`5280` (which obsoletes :rfc:`2459` and :rfc:`3280`). X.509
-certificates are commonly used in protocols like `TLS`_.
+X.509 is a standard for `public key infrastructure`_. ``cryptography``
+implements X.509 in accordance with :rfc:`5280`, is principally focused on
```
reaperhulk:
```suggestion
implements X.509 in accordance with :rfc:`5280`, which is principally focused on
```

alex: This didn't match my intent, fixed!

Issue #12283, opened 2025-01-14
[META] Declarative API for ASN.1 in Python


woodruffw:
This is a meta-issue/design issue for tracking a declarative ASN.1 API for Cryptography!

The goal: an importable Python API that users of Cryptography can define ASN.1 structures with, which can then be ser/de'd to and from DER (and **only** DER).

A rough sketch, demonstrating the basic idioms of the API:

```python
from cryptography.hazmat import asn1

# Corresponding to:
#
# Signature ::= Sequence {
#   r INTEGER,
#   s INTEGER
# }
@asn1.sequence
class Signature:
    r: int
    s: int

sig = Signature.from_der(b"...")
raw: bytes = sig.to_der()
```

This declarative API should have fully generality/expressivity with respect to ASN.1's own feature set, including qualifiers like `EXPLICIT` and `IMPLICIT`:

```python
@asn1.sequence
class Signature:
    r: Annotated[int, asn1.explicit(0)]
    s: Annotated[int, asn1.implicit(1)]
```

This would also (naturally) include generality over user-defined types:

```python
@asn1.sequence
class Frob:
    ...

@asn1.sequence
class FrobHolder:
    frob: Frob
```

Key design constraints:

* This API should be 100% declarative: there should be no imperative effects on ASN.1 ser/de

Open design questions:

* What's the best way to handle `ANY`? `asn1.Any` as a generic TLV type, similar to rust-asn1?
* What's the best way to handle `ANY DEFINED BY`?
  
    Maybe something like this:

	```python
	@asn1.sequence
	class VaryMe:
        content_type: asn1.ObjectIdentifier
        content: Annotated[Varied, asn1.defined_by("content_type")]

	# or maybe we can do this with the standard enum.Enum?
	# or maybe @asn1.varied?
	class Varied(asn1.Enum):
        foo: Annotated[VariantA, asn1.defined_by(SOME_OID)]
        bar: Annotated[VariantB, asn1.defined_by(ANOTHER_OID)]
	```

* To what extent/how can we best support trivial "native" Python types (`int`, `str`, etc.) versus "synthetic" types?
    * Moreover, what's the appropriate isomorph for `str`? Probably `UTF8String`, with all other string-ish types being `bytes`?
* What about non-trivial native types like `list[T]`, `set[T]`, etc? Should we support these with fixed mappings (e.g. `list[T] -> SEQUENCE OF`), or should we have our own types that don't require as much object conversion (e.g. `asn.List[T]`)?
[...]

PR #13325, opened 2025-08-20
Initial implementation of ASN.1 API

facutuesca:

## Overview
This PR adds the initial structure for the ASN.1 module, plus support for defining `SEQUENCE` types and using `INTEGER` fields. Only the encoding logic is included, to keep the PR size small and to keep the focus on the structure of the code.

Here's an example of what using the API looks like:
```python

@asn1.sequence
class Example:
    foo: int
    bar: int

value = Example(foo=9, bar=6)

encoded = asn1.encode_der(value)
assert encoded == b"\x30\x06\x02\x01\x09\x02\x01\x06"
```

## Structure
The code is divided into a `hazmat.asn1` Python module, and a Rust `_rust.asn1_exp` Rust module. This last one has a different name in order to avoid mixing it with the existing `asn1` Rust module (which can be done later, but for now doing it like this simplifies the PR).

## Details
Taking as an example the code snippet above, this is what happens:
1. While interpreting the class definition, the `@asn1.sequence` decorator is executed, which goes through the `Example` class and recursively reads its type annotations. This information is converted to a Rust object (`AnnotatedType`) and stored in the `Example.__asn1_root__` and `Example.__asn1_fields__` attributes. All of this happens in `hazmat/asn1/asn1.py`
2. An `Example` object is instantiated with some values, and `asn1.encode_der` is called on it. `asn1.encode_der` is a Rust function defined in `asn1_exp/asn1.rs`.
3. This function combines the type information (`AnnotatedType` object) and the value of the object into an `AnnotatedTypeObject` value, and encodes it using `rust-asn1`'s `asn1::write` API. 
4. This is possible because `AnnotatedTypeObject` implements the `asn1::Asn1Writable` trait. This is done in `asn1_exp/encoding.rs`
5. And that's it, the resulting bytes are now available in Python

Some things (like the empty `Annotation` struct) are there as placeholders for missing features I removed to keep this PR small. I removed their contents but left the empty struct so that the rest of the data and logic that depend on it will keep the same structure once we re-add the missing features.


facutuesca:
I added `typing-extensions` as an unconditional dependency, but it can probably be removed for newer Python versions if we use `typing` when we detect a new-enough Python version

alex:
looks like there's some missed coverage here -- do you want to take a look at the missing test cases before I review?

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -0,0 +1,101 @@
...
+import builtins
+from typing import (
+    Any,
+    ClassVar,
+    TypeVar,
+)
+
+import typing_extensions as te
+
+from cryptography.hazmat.bindings._rust import asn1_exp
```
alex: I don't love this module name, I'm not sure what `exp` even stands for :-)

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -0,0 +1,101 @@
...
+from __future__ import annotations
+
+import builtins
+from typing import (
+    Any,
+    ClassVar,
+    TypeVar,
+)
+
+import typing_extensions as te
```
alex: we don't usually alias imports, I know it's verbose but we prefer using the full name

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -0,0 +1,101 @@
...
+# for complete details.
+
+from __future__ import annotations
+
+import builtins
+from typing import (
+    Any,
+    ClassVar,
+    TypeVar,
+)
```
alex: Similarly, house style is to reference these as `typing.` for clarity (`Any` is particularly unclear without the `typing.`)

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -0,0 +1,101 @@
...
+
+encode_der = asn1_exp.encode_der
+
+
+class Asn1Decorated(te.Protocol[T]):
+    __asn1_root__: asn1_exp.AnnotatedType
+    __asn1_fields__: ClassVar[dict[str, asn1_exp.AnnotatedType]]
+
+    def to_der(self) -> bytes:
+        pass
```
alex: Hmm, what is this for, is it an abstract base class?

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def _normalize_field_type(
+    field_type: Any, field_name: str
+) -> asn1_exp.AnnotatedType:
+    annotation = asn1_exp.Annotation()
+
+    if field_type == builtins.int:
```
alex: Prefer `is` when comparing against types

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -0,0 +1,101 @@
...
+        rust_field_type = asn1_exp.Type.PyInt()
+    elif hasattr(field_type, "__asn1_root__"):
+        annotated_root = getattr(field_type, "__asn1_root__")
```
alex: This is just `field_type.__asn1_root__`, no `getattr` needed

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -0,0 +1,101 @@
...
+        # Rust code can understand.
+        annotated_field_type = _normalize_field_type(field_type, field_name)
+        fields[field_name] = annotated_field_type
+
+    return fields
+
+
+def _register_asn1_type(cls: type[U], root_type: asn1_exp.RootType) -> None:
+    raw_fields = te.get_type_hints(cls, include_extras=True)
+    setattr(cls, "__asn1_fields__", _annotate_fields(raw_fields))
```
alex: similarly: `cls.__asn1_fields__ = _annotate_fields(raw_fields)`

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -0,0 +1,101 @@
...
+T = TypeVar("T", covariant=True)
+U = TypeVar("U")
+
+
+encode_der = asn1_exp.encode_der
+
+
+class Asn1Decorated(te.Protocol[T]):
+    __asn1_root__: asn1_exp.AnnotatedType
+    __asn1_fields__: ClassVar[dict[str, asn1_exp.AnnotatedType]]
```
alex: Why a dict? I'd prefer any things we're hanging off types be completely opaque.

src/cryptography/hazmat/asn1/__init__.py
```diff
@@ -0,0 +1,10 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+from cryptography.hazmat.asn1.asn1 import encode_der, sequence
+
+__all__ = [
+    "encode_der",
```
alex, line 8: If we have `asn1.encode_der()` I don't think we need a `to_der()` method -- one entrypoint only.

src/rust/src/asn1_exp/types.rs
```diff
@@ -0,0 +1,103 @@
...
+        Self {}
+    }
+}
+
+/// Utility function for converting builtin Python types.
+/// This is needed when `encode_der` and `decode_der` are called
+/// with builtin Python types (`int`, `str`, etc), and we can't
+/// handle the conversion to the Rust `AnnotatedType` like we
+/// do for classes with `@sequence`.
+pub fn non_root_type_to_annotated<'p>(
```
alex: It seems like we should be able to de-duplicate this logic with the Python implementation that does roughly the same.

pyproject.toml
```diff
@@ -6,6 +6,7 @@ requires = [
 
     # Must be kept in sync with `project.dependencies`
     "cffi>=1.14; platform_python_implementation != 'PyPy'",
+    "typing-extensions>=4.13.2,<4.14.0",
```
alex: Why such a tight version bound?

src/cryptography/hazmat/asn1/asn1.py
facutuesca: ah lol, `experimental`, it was just meant to have a different name from the existing `asn1` module

src/cryptography/hazmat/asn1/asn1.py

src/cryptography/hazmat/asn1/asn1.py

src/cryptography/hazmat/asn1/asn1.py
facutuesca: yeah, it was meant originally to document the fields/methods available on classes that were modified by the `@sequence` operator. I removed it now, since `mypy` just ignores it.

src/cryptography/hazmat/asn1/asn1.py

src/cryptography/hazmat/asn1/asn1.py

src/cryptography/hazmat/asn1/asn1.py
facutuesca: `mypy` complains:
```
src/cryptography/hazmat/asn1/asn1.py:55: error: "type[U]" has no attribute "__asn1_fields__"  [attr-defined]
```

src/cryptography/hazmat/asn1/asn1.py
facutuesca: as mentioned above, I removed this class since it was not useful

src/cryptography/hazmat/asn1/__init__.py
facutuesca: makes sense, fixed

src/rust/src/asn1_exp/types.rs
facutuesca: this is only the case for now (where we only handle `sequence` and `int` with no other types and/or annotations. The Python code becomes more complex once we add more features (since it handles all types/annotations), while this Rust function remains simple, since it's just a mapping between Python and Rust types for simple (non-root) types.

pyproject.toml
facutuesca: ah yeah, the upper bound is not needed, fixed

facutuesca:
> looks like there's some missed coverage here -- do you want to take a look at the missing test cases before I review?

Missing tests added!

alex:
Will review again tomorrow. I should have said at the top: this looks like a solid start!

src/cryptography/hazmat/asn1/asn1.py
alex: We probably need to find a real name 😂

src/cryptography/hazmat/asn1/asn1.py
alex: Hmm, this error doesn't make much sense to me.

alex: Why do we need both `__asn1_fields__` and `__asn1_root__`? It seems like only one should be required.

src/rust/src/asn1_exp/asn1.rs
```diff
@@ -0,0 +1,44 @@
+// This file is dual licensed under the terms of the Apache License, Version
+// 2.0, and the BSD License. See the LICENSE file in the root of this repository
+// for complete details.
+
+use pyo3::prelude::*;
```
alex: We don't use `import *`

src/rust/src/asn1_exp/encode.rs
```diff
@@ -0,0 +1,69 @@
...
+// 2.0, and the BSD License. See the LICENSE file in the root of this repository
+// for complete details.
+
+use asn1::{SimpleAsn1Writable, Writer};
+use pyo3::prelude::*;
+use pyo3::types::{PyDict, PyString};
+
+use crate::asn1_exp::types::{AnnotatedType, AnnotatedTypeObject, Type};
+
+fn write_value<T: SimpleAsn1Writable>(
```
alex: This doesn't seem to do very much?

src/rust/src/asn1_exp/encode.rs
```diff
@@ -0,0 +1,69 @@
...
+        let value: Bound<'_, PyAny> = self.value.clone();
+        let py = value.py();
+        let annotated_type = self.annotated_type;
+
+        let inner = annotated_type.inner.get();
+        match &inner {
+            Type::Sequence(cls) => {
+                let fields = cls
+                    .getattr(py, "__asn1_fields__")
+                    .map_err(|_| asn1::WriteError::AllocationError)?;
```
alex: It seems like instead of a separate `__asn1_fields__` we could just store the field types in `Type::Sequence()`?

src/rust/src/asn1_exp/types.rs
alex: Surely there'll be a bunch of duplication for all of the primitive types? Is there a reason the Python code can't use this?

tests/hazmat/asn1/test_asn1.py
```diff
@@ -0,0 +1,108 @@
...
+        setattr(Invalid, "__asn1_root__", int)
+
+        with pytest.raises(TypeError, match="unsupported root type"):
+
+            @asn1.sequence
+            class Example:
+                foo: Invalid
+
+
+class TestEncoding:
```
alex: Let's put tests of specific types in their own test file.

ci-constraints-requirements.txt
```diff
@@ -253,13 +253,9 @@ tomli==2.2.1 ; python_full_version <= '3.11'
     #   nox
     #   pytest
     #   sphinx
-typing-extensions==4.13.2 ; python_full_version < '3.9'
-    # via
-    #   exceptiongroup
-    #   mypy
-    #   virtualenv
-typing-extensions==4.14.1 ; python_full_version >= '3.9'
+typing-extensions==4.13.2
```
alex: This file should be built in a way that uses the highest version for each python version:

`uv pip compile -U --universal --python-version 3.8 --all-extras --unsafe-package=cffi --unsafe-package=pycparser --unsafe-package=setuptools --unsafe-package=cryptography-vectors --unsafe-package=bcrypt pyproject.toml -o ci-constraints-requirements.txt -q` is the right way to build it.

pyproject.toml
```diff
@@ -6,6 +6,7 @@ requires = [
 
     # Must be kept in sync with `project.dependencies`
     "cffi>=1.14; platform_python_implementation != 'PyPy'",
+    "typing-extensions>=4.13.2",
```
alex: Why is this needed at build time?

src/rust/src/asn1_exp/encode.rs
facutuesca: yeah, it's because it's useful to avoid duplication only when all the other types are implemented. See [here](https://github.com/trailofbits/cryptography-asn1/blob/935f7458c23ea8c7ad14e5119a5d71e6fafd67ae/rust/src/encode.rs#L36-L46) (it saves us writing that `match` statement once for each type)

pyproject.toml
facutuesca: it's not, ha. Fixed

alex:
Looks like there's some merge conflicts

ci-constraints-requirements.txt
facutuesca: fixed

tests/hazmat/asn1/test_asn1.py
facutuesca: fixed

src/rust/src/asn1_exp/encode.rs
facutuesca: yup, good idea. Now `Type::Sequence` contains the root type and the fields.

src/rust/src/asn1_exp/asn1.rs
facutuesca: fixed

src/rust/src/asn1_exp/types.rs
facutuesca: yeah fair enough. I moved the duplicated logic (that maps builtin Python types to Rust enum values) to a separate function, and called it also from Python.

src/cryptography/hazmat/asn1/asn1.py
facutuesca: yup, see below, I removed `__asn1_fields__`

src/cryptography/hazmat/asn1/asn1.py
facutuesca: I'm open to suggestions. `asn1_api` maybe?

facutuesca:
> Looks like there's some merge conflicts

fixed!

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -0,0 +1,77 @@
...
+def _register_asn1_sequence(cls: type[U]) -> None:
+    raw_fields = typing_extensions.get_type_hints(cls, include_extras=True)
+    root = asn1_exp.AnnotatedType(
+        asn1_exp.Type.Sequence(cls, _annotate_fields(raw_fields)),
+        asn1_exp.Annotation(),
+    )
+
+    setattr(cls, "__asn1_root__", root)
+
+    def new_init(self: U, /, **kwargs: object) -> None:
```
alex: We should probably codegen this for performance, but doesn't have to happen in this PR. Can you add a TODO though?

tests/hazmat/asn1/test_encoding.py
```diff
@@ -0,0 +1,66 @@
+# This file is dual licensed under the terms of the Apache License, Version
+# 2.0, and the BSD License. See the LICENSE file in the root of this repository
+# for complete details.
+
+import pytest
+
+import cryptography.hazmat.asn1 as asn1
+
+
+class TestEncoding:
```
alex: One class per type we're testing teh encoding of (sequence vs. int so far)

alex (commented):
Two small comments, but otherwise this LGTM. I'd like @reaperhulk to have a look before we merge though.

tests/hazmat/asn1/test_encoding.py
facutuesca: fixed

src/cryptography/hazmat/asn1/asn1.py
facutuesca: added a TODO comment

pyproject.toml
```diff
@@ -49,6 +49,8 @@ requires-python = ">=3.8,!=3.9.0,!=3.9.1"
 dependencies = [
     # Must be kept in sync with `build-system.requires`
     "cffi>=1.14; platform_python_implementation != 'PyPy'",
+    # Must be kept in sync with ./.github/requirements/build-requirements.{in,txt}
+    "typing-extensions>=4.13.2",
```
reaperhulk: This is required because we need two functions that were added in Python 3.11 right? is it a requirement for >= 3.11?

src/cryptography/hazmat/asn1/asn1.py
reaperhulk: I suggest `declarative_asn1` for now since it provides the declarative ASN.1 API (or `decl_asn1` if we need it to be short)

pyproject.toml
alex: If it's not, let's use the `python_version` branch

src/cryptography/hazmat/asn1/asn1.py
alex: lets use the full word, bytes are free

pyproject.toml
reaperhulk: Looks like typing has this in 3.11+ (and Facundo did note that in a comment 5 days ago, but there's a lot of traffic on this PR 😄). Let's call it from `typing` where possible and set the `python_version` on the rest, so we know when we can remove this...in 4 years.

facutuesca: fixed! there was an extra `typing` function which we need to use `typing-extensions` for: `typing.get_type_hints` which we call with its `include_extras` parameter. That parameter was added in Python 3.9, so I also added it to the conditional.

Now it will use the `typing-extensions` variant for any Python < 3.11. I can make it even more precise (only use it for Python < 3.9), but for now I only left a comment saying to remove it when the min version is 3.9.

```python
if sys.version_info < (3, 11):
    import typing_extensions

    # We use the `include_extras` parameter of `get_type_hints`, which was
    #  added in Python 3.9, so this can be replaced by the `typing` version
    # once the min version is >= 3.9
    get_type_hints = typing_extensions.get_type_hints
else:
    get_type_hints = typing.get_type_hints
```

src/cryptography/hazmat/asn1/asn1.py

pyproject.toml
alex: Is there a reason not to make this branch on 3.9? Seems cleaner.

facutuesca: nope, fixed

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -0,0 +1,92 @@
...
+# for `dataclass_transform` that conditionally points to `typing` or
+# `typing_extensions` depending on the Python version (like we do for
+# `get_type_hints`).
+# We work around it by making the whole decorated class conditional on the
+# Python version.
+if sys.version_info < (3, 11):
+
+    @typing_extensions.dataclass_transform(kw_only_default=True)
+    def sequence(cls: type[U]) -> type[U]:
+        dataclass_cls = dataclasses.dataclass(cls)
```
alex: We don't want all the defaults that dataclass has, I think to start with the _only_ thing we want to define is `init`, everything else should be false.

src/rust/src/declarative_asn1/types.rs
```diff
@@ -0,0 +1,104 @@
...
+    #[pyo3(constructor = ())]
+    PyInt(),
+}
+
+/// A type that we know how to encode/decode, along with any
+/// annotations that influence encoding/decoding.
+#[pyo3::pyclass(frozen, module = "cryptography.hazmat.bindings._rust.asn1")]
+#[derive(Debug)]
+pub struct AnnotatedType {
+    #[pyo3(get)]
```
alex: is there a reason to expose these?

src/rust/src/declarative_asn1/types.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+#[pyo3::pyclass(module = "cryptography.hazmat.bindings._rust.asn1")]
+#[derive(Clone, Debug)]
+pub struct Annotation {}
+
+#[pyo3::pymethods]
+impl Annotation {
+    #[new]
+    #[pyo3(signature = ())]
```
alex: I don't think this is required?

src/rust/src/declarative_asn1/types.rs
facutuesca: Removed it for now, it will be needed in the future

src/rust/src/declarative_asn1/types.rs
facutuesca: Removed

src/cryptography/hazmat/asn1/asn1.py
facutuesca: Fixed. I had to add another Python version check, since one of the `dataclass` arguments (`match_args`) that defaults to True was added in Python 3.10.

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -0,0 +1,111 @@
...
+    def sequence(cls: type[U]) -> type[U]:
+        # We use `dataclasses.dataclass` to add an __init__ method
+        # to the class.
+        if sys.version_info >= (3, 10):
+            dataclass_cls = dataclasses.dataclass(
+                repr=False,
+                eq=False,
+                # `match_args` was added in Python 3.10 and defaults
+                # to True
+                match_args=False,
```
alex, line 90: Ooops, sorry one last thing, I think we _do_ want `kw_only=True`

facutuesca: `kw_only` was added in Python `3.10`. If we enable it only for those versions, the ASN.1 classes will behave differently depending the Python version:
```python
@sequence
class Example:
    foo: int

value = Example(3) # Works in Python < 3.10, raises an exception in Python >= 3.10.
```

alex: Frustrating, but I think we should enable it wherever possible. WDYT @reaperhulk ?

facutuesca: I enabled it (for Python >= 3.10), and added a test that checks the behavior

alex (dismissed):
will wait for @reaperhulk to confirm he's goood, but I think this is ready

reaperhulk (approved):

reaperhulk:
This needs a rebase now, sorry!

reaperhulk (approved):

PR #13328, opened 2025-08-20
scope --profile=dev to just cryptography

alex:


alex:
Suggested here: https://github.com/astral-sh/uv/issues/15386#issuecomment-3205498814

PR #13341, opened 2025-08-22
serialization.rst: pkcs7_decrypt_der (binary mode)

gav2xlin:
Hello. Please, mention that the ist of [PKCS7Options] is optional.

options = [pkcs7.PKCS7Options.Text]  # to decrypt text data
options = []  # to decrypt binary data (as is)

An example:

> :param options: An optional list of
>         :class:`~cryptography.hazmat.primitives.serialization.pkcs7.PKCS7Options`. For
>         this operation only
>         :attr:`~cryptography.hazmat.primitives.serialization.pkcs7.PKCS7Options.Text` is supported
>         or nothing for binary data.


reaperhulk:
`Optional` generally means you could pass`None`. This function requires that you pass a list, but that list can be empty. I'd suggest just stating that an empty list is allowable and handles binary data.

gav2xlin:
@alex @reaperhulk

Changed and fixed

[https://github.com/pyca/cryptography/blob/main/src/rust/src/pkcs7.rs](https://github.com/pyca/cryptography/blob/main/src/rust/src/pkcs7.rs)

The functions decrypt_smime() and decrypt_pem() call decrypt_der()

PR #13366, opened 2025-08-29
Upgrade to pyo3 0.26, fixing all warnings

alex:


alex:
@sandeepgupta12 not sure what's going on, but I'm now seeing _other_ jobs getting killed. For example:
Screenshot 2025-08-29 at 4 42 31 PM

This job is not running on the ppc64le running, so there's no reason that it should be getting killed. Is there a better place to file these bugs besides pinging you?

reaperhulk (approved):

sandeepgupta12:
@alex  Sorry for the late reply — I was on leave for the past few days. I’ll inform the GHA team about this. In the meantime, you can also open an issue anytime here: [https://github.com/IBM/actionspz/issues](https://github.com/IBM/actionspz/issues)

PR #13369, opened 2025-08-30
disable cov-contexts, they modestly slow things down and we don't use them

alex:


alex:
Cuts the coverarge merge step from 15s min to 6s. I'm sure there are perf wins in other stages as well.

reaperhulk:
Makes sense that they do slow things down, but I’ve used them to determine if a test covered a line before. I suppose we could disable them unless we explicitly want to check though…

alex:
Huh, really? I've literally never used them 😂 they're easy enough to enable if you want them `nox -e tests -- --cov-context=test` works fine.

reaperhulk (approved):

PR #13375, opened 2025-08-31
Add SSL_set_info_callback function declaration

alex:
we need this to have a `set_info_callback` on an SSL.Connection which twisted needs for reasons I don't understand


reaperhulk (approved):

adiroiban:
`set_info_callback` is used by Twisted to check that the connection hostname matches the connection certificate

The callback code is here

https://github.com/twisted/twisted/blob/aedee3ea72b2c206774fbeae7a54e71939699d54/src/twisted/internet/_sslverify.py#L1152-L1182

The `Context.set_verify` is not used, and instead the info callback is used so that the code can set the hostname `connection.set_tlsext_host_name(self._hostnameBytes)`

---------

I think that this is "legacy" code in Twisted, and the plan is to (re)use newer code to check for server identity.

I remember that @glyph has mentioned that he wants to rewrite this code... but I can't find the conversation

alex:
It seems like `connection.set_tlsext_host_name(self._hostnameBytes)` should be callable as soon as the connection is created?

PR #12465, opened 2025-02-15
PKCS#7 signing & verification - Certificate extension policies

nitneuqr:
As per #12267 and #12104, I'm opening this draft implementation of verification of PKCS#7 (S/MIME) certificates. Tests are not passing yet; still need to build an appropriate certificate for this.

The constraints I've gathered so far:
- Certificates used as end entities (i.e., the cert used to sign
  a PKCS#7/SMIME message) should not have ca=true in their basic
  constraints extension.
- EKU_CLIENT_AUTH_OID is not required
- EKU_EMAIL_PROTECTION_OID is required

@alex @woodruffw @prauscher feel free to add some more constraints, I'm not particularly familiar with CA verification.

PR #13376, opened 2025-08-31
add test vectors for pkcs#7 verification policies

prauscher:
Thanks to @briantist in #12465 I separated the vectors needed to test PKCS#7 verification. The files were created by @nitneuqr over in #12465


docs/development/test-vectors.rst
```diff
@@ -1002,6 +1002,12 @@ Custom PKCS7 Test Vectors
 * ``pkcs7/enveloped-no-content.der``- A DER encoded PKCS7 file with
   enveloped data, without encrypted content, with key encrypted under the
   public key of ``x509/custom/ca/rsa_ca.pem``.
+* ``pkcs7/ca.pem`` - A certificate adapted for S/MIME signature & verification.
```
alex: These shouldn't be named "ca" -- that'd be used for certificate authorities that are used to issue other cers, not for stand-alone certs.

briantist: What should they be named?

prauscher: After reading https://github.com/pyca/cryptography/pull/12465/files#diff-ee3cc87634fba3d98c17ee2dbcb575f63f33603a5ab7ac9fa7bcc3b17bced6a3R154 i recon only the description is missleading: The `ca.pem` is used to issue certificates used during testing. Maybe the description should be replaced with `A certificate adapted to sign certificates used for S/MIME signature & verification`?

alex: Hmm, that's a bit surprising to me since it has a Subject Alt Names extension which generally implies its an end-entity certificate.

prauscher: According to the code it is only used as a CA, so maybe the SAN is there by accident? @nitneuqr can you help out?

prauscher: Okay, I might have a better understanding now: While `ca.pem` actually is a CA-certificate, the test code will simply copy the SAN and AKI-Extensions from the CA-Cert to the (temporary generated) EE-certificate. Otherwise, it basically is `x509/custom/ca/ca.pem` (even sharing `ca_key.pem`). The basic question is: What should it be like? I think `ca.pem` could be replaced if the test code is changed to add the SAN and AKI-extensions in code instead of reading them from file. Would it be worth to propose such a change? Or should we change documentation?

prauscher: Okay, it was quite simple to change the test and avoid shipping a custom `ca.pem` and `ca_key.pem`. Will push the change here and submit a patch to #12465

alex (approved):

docs/development/test-vectors.rst
nitneuqr: Hey, sorry I'm late for that. I think I changed the ca.pem certificate for a reason, can't remember why. We will see in the next PRs :)

PR #13387, opened 2025-09-03
Upgrade to v1 of the wycheproof vectors

alex:
v0 was removed, RIP

### DIFF
--- a/tests/wycheproof/test_rsa.py
+++ b/tests/wycheproof/test_rsa.py
@@ -40,7 +40,6 @@ def should_verify(backend, wycheproof):
 
 
 @wycheproof_tests(
-    "rsa_signature_test.json",
     "rsa_signature_2048_sha224_test.json",
     "rsa_signature_2048_sha256_test.json",
     "rsa_signature_2048_sha384_test.json",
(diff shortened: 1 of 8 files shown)


alex:
I didn't do a comprehensive review for test files we should have but don't.

Also, FYI @cpu that we did a successful migration with only one upstream bug filed :-)

tests/wycheproof/test_rsa.py
```diff
@@ -58,15 +57,19 @@ def should_verify(backend, wycheproof):
     "rsa_signature_3072_sha3_256_test.json",
     "rsa_signature_3072_sha3_384_test.json",
     "rsa_signature_3072_sha3_512_test.json",
+    "rsa_signature_4096_sha256_test.json",
     "rsa_signature_4096_sha384_test.json",
     "rsa_signature_4096_sha512_test.json",
     "rsa_signature_4096_sha512_256_test.json",
+    "rsa_signature_8192_sha256_test.json",
```
reaperhulk, line 64: How slow do these make us 😒

cpu:
> Also, FYI @cpu that we did a successful migration with only one upstream bug filed :-)

Love to see it. Thanks for mentioning!

PR #13404, opened 2025-09-05
PKCS#7 signing & verification - Certificate extension policies

### LINKED ISSUE #12104: FR: Allow verification of SMIME-Certificates
Since cryptography supports signing and encrypting mails for smime, there should be an (easy) way to verify smime-certificates before using in encryption.
This is possibly a duplicate of #11165, which would allow custom extension policys. For SMIME-Support it probably would be sufficient to add something like `build_smime_verifier`. `build_client_verifier` cannot be used since it requires `clientAuth` Extended Key Usage.

As #11165 is already there, one discussion here could be weather a "quick access" `build_smime_verifier` is wanted or not. I believe implementing this feature would be fairly straightforward (and obviously simplier than crafting an API for custom EKU validation). Also I believe it would be beneficiary for "common" use cases to provide pre-configured verifiers to avoid each user having to copy-paste validation code (and making misstakes in the process).

prauscher:
This is a fork PR of #12465, which was created by @nitneuqr and adds verification of PKCS#7 (S/MIME) certificates. This fork rebases the PR to main and applies the patch given in the PR to avoid a custom `ca.pem` in tests.

forking as I lack permissions to apply the patch and rebase in @nitneuqr's repo. If we prefer to merge #12465 I will close this PR.

closes #12104

### DIFF
--- a/src/cryptography/hazmat/primitives/serialization/pkcs7.py
+++ b/src/cryptography/hazmat/primitives/serialization/pkcs7.py
@@ -21,6 +21,13 @@ from cryptography.hazmat.primitives.ciphers import (
     algorithms,
 )
 from cryptography.utils import _check_byteslike
+from cryptography.x509 import Certificate
+from cryptography.x509.oid import ExtendedKeyUsageOID
+from cryptography.x509.verification import (
+    Criticality,
+    ExtensionPolicy,
+    Policy,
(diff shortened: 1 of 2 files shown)


prauscher:
closed to continue in #12465 :)

PR #13417, opened 2025-09-10
fix: avoid cloning PEM tag when parsing

akx:
Cleans up the TODO from #12296 by using a Cow around the either-decrypted-or-maybe-not contents.

### DIFF
--- a/src/rust/src/backend/keys.rs
+++ b/src/rust/src/backend/keys.rs
@@ -3,6 +3,7 @@
 // for complete details.
 
 use pyo3::IntoPyObject;
+use std::borrow::Cow;
 
 use crate::buf::CffiBuf;
 use crate::error::{CryptographyError, CryptographyResult};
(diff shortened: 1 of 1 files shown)


src/rust/src/backend/keys.rs
```diff
@@ -149,19 +152,23 @@ fn load_pem_private_key<'p>(
...
-            assert_eq!(tag, "ENCRYPTED PRIVATE KEY");
+        "ENCRYPTED PRIVATE KEY" => {
             password_used = true;
             cryptography_key_parsing::pkcs8::parse_encrypted_private_key(&data, password)?
         }
+        _ => {
+            return Err(CryptographyError::from(
+                pyo3::exceptions::PyValueError::new_err("Unsupported PEM tag."),
+            ))
+        }
```
alex: This change seems distinct -- I'm a bit surprised coverage pases, I'm not sure how the `_` branch is reachable now?

akx: There's no local `tag` name that could be captured and asserted on anymore, hence the change here. I could make it a local and match on that to retain the previous behavior if you like - but feels like a catch-all + assert is... just worse? 😅 

As for coverage passing, I don't know enough about the CI infra here to say 😄

alex: Oh, I think this is something weird with this code being optimized out. For clarity, would you mind switching back to the previous formulation (`_ => ` + assert and simply using `p.tag()` in the assert?

akx: Sure, reverted that bit.

alex (approved):

alex:
thanks!

Issue #13331, opened 2025-08-21
The SHA3 OIDs are deprecated


Jakuje:
The x509 SHA3 OIDs listed here:

https://github.com/pyca/cryptography/blob/5e73a9138cf970e6ccf963125e37331dfa9b151d/src/rust/cryptography-x509/src/oid.rs#L130-L137

are deprecated values that should not be used anymore:

https://hosted.oidplus.com/viathinksoft/?goto=oid%3A1.3.6.1.4.1.37476.3.2.1.99.7.384

The NIST values `{joint-iso-itu-t(2) country(16) us(840) organization(1) gov(101) csor(3) nistAlgorithms(4) hashalgs(2) sha3-384(9)}` should be used instead, wherever possible.

If these are strictly necessary for some use case you need, they should have at least warning about their deprecation and there should be generally usable variants with standard OIDs.

Jakuje:
This diff worked for me:
```
 pub const SHA3_224_OID: asn1::ObjectIdentifier =
-    asn1::oid!(1, 3, 6, 1, 4, 1, 37476, 3, 2, 1, 99, 7, 224);
+    asn1::oid!(2, 16, 840, 1, 101, 3, 4, 2, 7);
 pub const SHA3_256_OID: asn1::ObjectIdentifier =
-    asn1::oid!(1, 3, 6, 1, 4, 1, 37476, 3, 2, 1, 99, 7, 256);
+    asn1::oid!(2, 16, 840, 1, 101, 3, 4, 2, 8);
 pub const SHA3_384_OID: asn1::ObjectIdentifier =
-    asn1::oid!(1, 3, 6, 1, 4, 1, 37476, 3, 2, 1, 99, 7, 384);
+    asn1::oid!(2, 16, 840, 1, 101, 3, 4, 2, 9);
 pub const SHA3_512_OID: asn1::ObjectIdentifier =
-    asn1::oid!(1, 3, 6, 1, 4, 1, 37476, 3, 2, 1, 99, 7, 512);
+    asn1::oid!(2, 16, 840, 1, 101, 3, 4, 2, 10);
```
I can submit a PR if this sounds like a good direction.

alex:
Hmm, we've got three problems here:

1) on parse we need to recognize both versions
2) on generate, we should use the new OIDs
3) we also have public Python APIs for these OIDs (`HashAlgorithmOID.SHA3_224` etc.), which are documented as having these particular OIDs. In these cases I'm not sure what the right behavior is.

[...]

PR #13428, opened 2025-09-14
Add NIST standard SHA3 OIDs alongside existing deprecated ones

alex:
Adds new SHA3_224_NIST, SHA3_256_NIST, SHA3_384_NIST, and SHA3_512_NIST OID constants using the official NIST OID values (2.16.840.1.101.3.4.2.x) in both Rust and Python APIs. The existing deprecated SHA3 OIDs are preserved for backward compatibility.

Related to #13331

🤖 Generated with [Claude Code](https://claude.ai/code)


reaperhulk (commented):
We should document these new oids and update the docs for the other oids to call them legacy and not preferred.

alex:
There's no docs for any of the hash OIDs, I checked.

reaperhulk:
That’s convenient 😄

reaperhulk (approved):

PR #13251, opened 2025-08-06
Introduce a script for doing GHA pin bumps

alex:

PR #13431, opened 2025-09-14
disable openssl main branch updating

reaperhulk:
they're transitioning to openssl 4 so we need to wait until we have a path in rust-openssl for testing against this. this can be reverted once we can compile with rust-openssl again


alex:
Can you please review https://github.com/pyca/cryptography/pull/13251 first, I want to avoid the merge conflicts.

PR #13442, opened 2025-09-16
we'll ship this with 3.5.3 why not

reaperhulk:


alex:
can you do the ci.yml bumps here as well?

PR #13447, opened 2025-09-16
Explicitly specify python in mac uv build invocation

ngoldbaum:
This should fix the missing wheels for cryptography 46 on MacOS.

### DIFF
--- a/.github/workflows/wheel-builder.yml
+++ b/.github/workflows/wheel-builder.yml
@@ -305,7 +305,7 @@ jobs:
 
           OPENSSL_DIR="$(readlink -f ../openssl-macos-universal2/)" \
               OPENSSL_STATIC=1 \
-              uv build --wheel --require-hashes --build-constraint=$BUILD_REQUIREMENTS_PATH $PY_LIMITED_API cryptography*.tar.gz -o wheelhouse/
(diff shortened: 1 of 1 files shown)


reaperhulk:
Looks like it's building the 3.14t properly now, but attempting to install it in a non-ft environment.

.github/workflows/wheel-builder.yml
```diff
@@ -305,20 +305,20 @@ jobs:
...
           OPENSSL_DIR="$(readlink -f ../openssl-macos-universal2/)" \
               OPENSSL_STATIC=1 \
-              uv build --wheel --require-hashes --build-constraint=$BUILD_REQUIREMENTS_PATH $PY_LIMITED_API cryptography*.tar.gz -o wheelhouse/
+              uv build --python='${{ matrix.PYTHON.BIN_PATH }}' --wheel --require-hashes --build-constraint=$BUILD_REQUIREMENTS_PATH $PY_LIMITED_API cryptography*.tar.gz -o wheelhouse/
         env:
           MACOSX_DEPLOYMENT_TARGET: ${{ matrix.PYTHON.DEPLOYMENT_TARGET }}
           ARCHFLAGS: ${{ matrix.PYTHON.ARCHFLAGS }}
           _PYTHON_HOST_PLATFORM: ${{ matrix.PYTHON._PYTHON_HOST_PLATFORM }}
 
       - run: uv venv
```
alex: Maybe this is dumb, but don't we just need the `--python` on the `uv venv` line and then all teh following commands will just use that?

ngoldbaum:
I think it's working now

PR #13449, opened 2025-09-16
Add decoding support to ASN.1 API

facutuesca:
Following https://github.com/pyca/cryptography/pull/13325, this PR adds support for decoding `int` and `Sequence` types to the ASN.1 API.

cc @alex @reaperhulk


src/rust/src/declarative_asn1/decode.rs
```diff
@@ -0,0 +1,109 @@
...
+                    let fields = fields.bind(py);
+                    for (name, ann_type) in fields.into_iter() {
+                        let name = name.extract::<&str>()?;
+                        let ann_type = ann_type.downcast::<AnnotatedType>().map_err(|_| {
+                            pyo3::exceptions::PyValueError::new_err(
+                                "target type has invalid annotations".to_string(),
+                            )
+                        })?;
+                        let value = ann_type.get().decode(py, d)?;
+                        kwargs.insert(name.to_string(), value);
```
alex: We shouldn't need to convert to a rust string here, this should just be a python string all the way through to avoid the copy/allocation.

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+impl<'a> SimpleAsn1ReadablePy<'a> for pyo3::types::PyInt {
+    type DerTarget = asn1::BigInt<'a>;
+
+    fn decode(py: pyo3::Python<'a>, parser: &mut Parser<'a>) -> ParseResult<pyo3::Bound<'a, Self>> {
+        let value = Self::get_value(parser)?;
+        let big = num_bigint::BigInt::from_signed_bytes_be(value.as_bytes());
+        Ok(big.into_pyobject(py)?)
```
alex: Why is this needed? We shouldn't need num just to convert from signed bytes, see `big_byte_slice_to_py_int`

alex:
gave a very cursory review, there's also some failing CI

src/rust/src/declarative_asn1/decode.rs
facutuesca: yeah good catch. I also made the `HashMap` a `PyDict`, since we don't need it as a rust type

src/rust/src/declarative_asn1/decode.rs

src/rust/src/declarative_asn1/asn1.rs
```diff
@@ -41,3 +43,28 @@ pub(crate) fn encode_der<'p>(
 
     Ok(pyo3::types::PyBytes::new(py, &encoded_bytes))
 }
+
+#[pyo3::pyfunction]
+pub(crate) fn decode_der<'p>(
+    py: pyo3::Python<'p>,
+    class: &pyo3::Bound<'p, pyo3::types::PyType>,
+    value: &'p pyo3::Bound<'p, pyo3::types::PyBytes>,
```
alex: You can just take `&[u8]` as an argument

src/rust/src/declarative_asn1/asn1.rs
```diff
@@ -41,3 +43,28 @@ pub(crate) fn encode_der<'p>(
...
+                pyo3::exceptions::PyValueError::new_err(
+                    "target type has invalid annotations".to_string(),
+                )
+            })?;
+            root.get().decode(py, parser)
+        } else {
+            // Handle simple types directly
+            let annotated_type = asn1_types::non_root_type_to_annotated(py, class, None)?;
+            annotated_type.decode(py, parser)
+        }
```
alex: This getattr + non_root_type_to_annotated call is the same as we have in encode, can we factor it out?

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -0,0 +1,108 @@
...
+// 2.0, and the BSD License. See the LICENSE file in the root of this repository
+// for complete details.
+
+use asn1::{Asn1Readable, Parser, SimpleAsn1Readable};
+use pyo3::types::PyAnyMethods;
+
+use crate::asn1::big_byte_slice_to_py_int;
+use crate::declarative_asn1::types::{AnnotatedType, Type};
+
+pub(crate) enum DecodeError {
```
alex: Do we need this in addition to CryptographyError?

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -0,0 +1,108 @@
...
+                            )
+                        })?;
+                        let value = ann_type.get().decode(py, d)?;
+                        kwargs.set_item(name, value)?;
+                    }
+                    let val = cls.call(py, (), Some(&kwargs))?.into_bound(py);
+                    Ok(val)
+                })
+            }
+            Type::PyInt() => pyo3::types::PyInt::decode(py, parser).map(|x| x.into_any()),
```
alex:
```suggestion
            Type::PyInt() => pyo3::types::PyInt::decode(py, parser)?.into_any(),
```

is much more our style

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -0,0 +1,108 @@
...
+        match self {
+            DecodeError::Asn1(e) => write!(f, "ASN.1 parse error: {e}"),
+            DecodeError::Py(e) => write!(f, "{e}"),
+        }
+    }
+}
+
+type ParseResult<T> = Result<T, DecodeError>;
+
+pub(crate) trait SimpleAsn1ReadablePy<'a>: Sized {
```
alex: I'm not sure I understand what this trait heirarchy is actually being used for -- right now there's no code that's either generic or dyn over it?

facutuesca: We use it to store the `rust-asn1` type corresponding to the Python type (e.g: `PyInt` -> `asn1::BigInt`), and the shared behavior `SimpleAsn1ReadablePy::get_value()`. Do you prefer to do it another way?

src/rust/src/declarative_asn1/decode.rs
facutuesca: fixed (also wrapped it in an `Ok()`)

src/rust/src/declarative_asn1/decode.rs
facutuesca: nope, it's a leftover from this being implemented as a separate library. Replaced with `CryptographyError`

src/rust/src/declarative_asn1/asn1.rs
facutuesca: yup, fixed

src/rust/src/declarative_asn1/asn1.rs
facutuesca: fixed

src/rust/src/declarative_asn1/decode.rs
alex: Unless I'm misreading, all the calls I see are direct calls, these could just be functions?

facutuesca: fixed, these are now two free functions for decoding `AnnotatedType` and `PyInt`

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -0,0 +1,55 @@
...
+use crate::declarative_asn1::types::{AnnotatedType, Type};
+use crate::error::CryptographyError;
+
+type ParseResult<T> = Result<T, CryptographyError>;
+
+fn decode_pyint<'a>(
+    py: pyo3::Python<'a>,
+    parser: &mut Parser<'a>,
+) -> ParseResult<pyo3::Bound<'a, pyo3::types::PyInt>> {
+    let value = asn1::BigInt::parse(parser)?;
```
alex: This should be `parser.read_element::<asn1::BigInt>()?`

alex (approved):
LGTM. One thought for the tests: we may want an assert_roundtrips() helper that takes a list of `[(obj, bytes)]` and verifies that obj serializes to bytes and bytes parses back to obj

PR #13476, opened 2025-09-19
Unify ASN.1 encoding and decoding tests

facutuesca:
As per https://github.com/pyca/cryptography/pull/13449#pullrequestreview-3242304881, test both encoding and decoding in a single roundtrip test (instead of separately)

cc @alex


tests/hazmat/asn1/test_serialization.py
```diff
@@ -2,14 +2,27 @@
...
 
 import cryptography.hazmat.asn1 as asn1
 
 
-class TestEncodeInteger:
+def assert_roundtrip(obj: typing.Any, obj_bytes: bytes) -> None:
+    encoded = asn1.encode_der(obj)
+    assert encoded == obj_bytes, (
+        f"Failed for {obj}: got {encoded.hex()}, expected{obj_bytes.hex()}"
+    )
```
alex: We don't generally use assertion messages -- they're pretty duplicative of what pytest already does, and strictly worse than just getting in there with a pdb. (In this case, I personally almost always prefer looking at the raw bytes rather than the hex, for example.)

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -109,7 +109,7 @@ def sequence(cls: type[U]) -> type[U]:
         # parameters.
         dataclass_cls = dataclasses.dataclass(
             repr=False,
-            eq=False,
+            eq=True,
```
alex: Do we need this? Could we just have the test types implement `__eq__` (perhaps with teh dataclass decorator themselves?)

tests/hazmat/asn1/test_serialization.py
```diff
@@ -23,35 +36,28 @@ class TestEncodeInteger:
             (-129, b"\x02\x02\xff\x7f"),
         ],
     )
-    def test_encode_int(self, value: int, expected: bytes) -> None:
-        encoded = asn1.encode_der(value)
-        assert encoded == expected, (
-            f"Failed for {value}: got {encoded.hex()}, expected"
-            f"{expected.hex()}"
-        )
+    def test_int(self, obj: int, obj_bytes: bytes) -> None:
```
alex: Instead of using `pytest.mark.parametrize`, can we just have `assert_roundtrips` take a list of test cases?

src/cryptography/hazmat/asn1/asn1.py
facutuesca: Do you mean something like this?:
```python
        @asn1.sequence
        @dataclass(eq=True)
        class Example:
            foo: int
```

I'm a bit worried that we would be testing classes that are in some way different than what users would use.

We could declare the dataclass here using the same parameters as the one in `sequence` (except for `eq`), so that no new behavior is added compared to the plain `@sequence` dataclass, but that requires the same version-dependent logic we implement [here](https://github.com/pyca/cryptography/blob/31edc05f7cbb77d2d39ca737e51838794bfd2588/src/cryptography/hazmat/asn1/asn1.py#L79-L117).

Is there a strong reason to not have the `sequence` dataclasses with equality comparison by default?

alex: Something like that, yes.

The reason not to do have all sequences implement eq is simply that we don't know what users would like, and our goal isn't to build dataclasses, it's to build an ASN.1 library :-)

tests/hazmat/asn1/test_serialization.py

src/cryptography/hazmat/asn1/asn1.py
facutuesca: fixed, now it looks like:
```python
        @asn1.sequence
        @_comparable_dataclass
        class Example:
            foo: int
```

where `_comparable_dataclass` contains the version-specific logic

tests/hazmat/asn1/test_serialization.py

PR #13482, opened 2025-09-21
asn1: Add support for bytes, str and bool

facutuesca:
This PR adds support for encoding/decoding the ASN.1 types `BOOLEAN`, `OCTET STRING` and `UTF8String`. They are mapped to the Python types `bool`, `bytes` and `str` respectively.

cc @alex @reaperhulk

### DIFF
--- a/src/cryptography/hazmat/bindings/_rust/declarative_asn1.pyi
+++ b/src/cryptography/hazmat/bindings/_rust/declarative_asn1.pyi
@@ -11,7 +11,10 @@ def non_root_python_to_rust(cls: type) -> Type: ...
 # annotations like this:
 class Type:
     Sequence: typing.ClassVar[type]
+    PyBool: typing.ClassVar[type]
     PyInt: typing.ClassVar[type]
+    PyBytes: typing.ClassVar[type]
+    PyStr: typing.ClassVar[type]
 
 class Annotation:
     def __new__(
--- a/src/rust/src/declarative_asn1/encode.rs
+++ b/src/rust/src/declarative_asn1/encode.rs
@@ -48,11 +48,30 @@ impl asn1::Asn1Writable for AnnotatedTypeObject<'_> {
                     Ok(())
                 }),
             ),
+            Type::PyBool() => {
+                let val: bool = value
+                    .extract()
+                    .map_err(|_| asn1::WriteError::AllocationError)?;
+                write_value(writer, &val)
+            }
             Type::PyInt() => {
                 let val: i64 = value
                     .extract()
                     .map_err(|_| asn1::WriteError::AllocationError)?;
                 write_value(writer, &val)
+            }
+            Type::PyBytes() => {
+                let val: &[u8] = value
+                    .extract()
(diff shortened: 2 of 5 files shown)


src/cryptography/hazmat/bindings/_rust/declarative_asn1.pyi
```diff
@@ -11,7 +11,10 @@ def non_root_python_to_rust(cls: type) -> Type: ...
 # annotations like this:
 class Type:
```
alex, line 12: Does this actually need to be exposed anymore? I don't think we use it in python.

src/rust/src/declarative_asn1/encode.rs
```diff
@@ -48,12 +48,31 @@ impl asn1::Asn1Writable for AnnotatedTypeObject<'_> {
...
                 write_value(writer, &val)
             }
+            Type::PyBytes() => {
+                let val: &[u8] = value
+                    .extract()
+                    .map_err(|_| asn1::WriteError::AllocationError)?;
+                write_value(writer, &val)
+            }
+            Type::PyStr() => {
+                let val: &str = value
```
alex: I think you'll need to use PyBackedStr here

src/cryptography/hazmat/bindings/_rust/declarative_asn1.pyi
facutuesca: we still use it here: https://github.com/pyca/cryptography/blob/cac6426aafd989e4d8d9c4f6b920a332a5280bba/src/cryptography/hazmat/asn1/asn1.py#L66

src/rust/src/declarative_asn1/encode.rs

alex:
@reaperhulk (copying from IRC): where'd we land: does str map to utf8string, or do we require users to always annotate with a specific string type?

alex (approved):
besides making sure we're in agreement on str -> utf8string, this lgtm

reaperhulk:
I can be fine with either. However, `str` being canonically utf8 makes sense as a default. When passing an, e.g., PrintableString, is the intent to say it has to be bytes with a declared encoding or are we going to allow `str` as long as it conforms to the allowable charset?

alex:
I think for PrintableString the API is `field: typing.Annotated[str, asn1.PrintableString]`? And maybe we optionally add an `asn1.PrintableString()` class that wraps a string?

facutuesca:
> I think for PrintableString the API is `field: typing.Annotated[str, asn1.PrintableString]`? And maybe we optionally add an `asn1.PrintableString()` class that wraps a string?

Our idea was to use `NewType`s for `PrintableString`:
```python
PrintableString = typing.NewType("PrintableString", str)
VisibleString = typing.NewType("VisibleString", str)
#... etc
```

and only use `typing.Annotated` for the encoding information associated with a field (e.g `EXPLICIT`, `IMPLICIT`, `DEFAULT`, `SIZE`, etc).

alex:
That works!

PR #13496, opened 2025-09-24
asn1: Add support for `PrintableString`

facutuesca:
This adds support for encoding/decoding `PrintableString`. 

```python
@asn1.sequence
class MySeq:
    field: asn1.PrintableString

object = MySeq(field=asn1.PrintableString("A string"))
encoded = encode_der(object)

assert decoded.field.as_str() == "A string"
```

A couple of implementation notes:

- In [this comment](https://github.com/pyca/cryptography/pull/13482#issuecomment-3321548452), I mentioned that the idea was to use `typing.NewType` to define the `PrintableString` type. However, I had forgotten that a `NewType` cannot be told apart from the original type during runtime:

    > Note that these checks are enforced only by the static type checker. At runtime, the statement Derived = NewType('Derived', Base) will make Derived a callable that immediately returns whatever parameter you pass it. 
[src](https://docs.python.org/3/library/typing.html#newtype)

  The actual solution was to use a Rust type `PrintableString` and expose it directly to Python. This type wraps a normal Python string, and can only be used through a restricted API (similar to [`rust-asn1`'s](https://github.com/alex/rust-asn1/blob/4790ee37682fafc094499dfaf8ded2c6fb3eb73a/src/types.rs#L378)):
```python
class PrintableString:
    def __new__(cls, inner: str) -> PrintableString: ...
    def as_str(self) -> str: ...
```

- The encoding errors coming from `rust-asn1` are still limited to `AllocationError`s without custom error messages. This is blocked by https://github.com/alex/rust-asn1/pull/567, so I'll work on that next.


src/rust/src/declarative_asn1/encode.rs
```diff
@@ -73,6 +73,20 @@ impl asn1::Asn1Writable for AnnotatedTypeObject<'_> {
...
             }
+            Type::PrintableString() => {
+                let val: &pyo3::Bound<'_, PrintableString> = value
+                    .downcast()
+                    .map_err(|_| asn1::WriteError::AllocationError)?;
+                let inner_str: pyo3::pybacked::PyBackedStr = val
+                    .get()
+                    .inner
+                    .extract(py)
+                    .map_err(|_| asn1::WriteError::AllocationError)?;
```
alex: We shouldn't need a pybacked string, `val.get().inner.bind(py)` should be all we need.

src/rust/src/declarative_asn1/types.rs
```diff
@@ -70,6 +73,26 @@ impl Annotation {
...
+    #[new]
+    #[pyo3(signature = (inner,))]
+    fn new(inner: pyo3::Py<pyo3::types::PyString>) -> Self {
+        PrintableString { inner }
+    }
+
+    #[pyo3(signature = ())]
+    pub fn as_str(&self, py: pyo3::Python<'_>) -> pyo3::PyResult<pyo3::Py<pyo3::types::PyString>> {
+        Ok(self.inner.clone_ref(py))
+    }
```
alex, line 100: We should probably give this a repr

facutuesca: should I use the repr to compare the values in the tests?

alex: No, but we should test the rep

tests/hazmat/asn1/test_api.py
```diff
@@ -9,7 +9,15 @@
 import cryptography.hazmat.asn1 as asn1
 
 
-class TestClassAPI:
+class TestTypesAPI:
+    def test_repr_printable_string(self) -> None:
+        assert (
+            repr(asn1.PrintableString("MyString"))
+            == "PrintableString(MyString)"
```
alex: I think we want `PrintableString("MyString")` (or single quotes) -- using the repr of the str

tests/hazmat/asn1/test_serialization.py
```diff
@@ -105,6 +117,28 @@ def test_string(self) -> None:
         )
 
 
+class TestPrintableString:
+    def test_ok_printable_string(self) -> None:
+        def decoded_eq(a: asn1.PrintableString, b: asn1.PrintableString):
```
alex: Oh, we should implement `__eq__` on `asn1.PrintableString`.

src/rust/src/declarative_asn1/encode.rs

tests/hazmat/asn1/test_api.py

tests/hazmat/asn1/test_serialization.py

src/rust/src/declarative_asn1/types.rs
```diff
@@ -70,6 +73,34 @@ impl Annotation {
...
+#[pyo3::pyclass(frozen, module = "cryptography.hazmat.bindings._rust.asn1")]
+pub struct PrintableString {
+    pub(crate) inner: pyo3::Py<pyo3::types::PyString>,
+}
+
+#[pyo3::pymethods]
+impl PrintableString {
+    #[new]
+    #[pyo3(signature = (inner,))]
+    fn new(inner: pyo3::Py<pyo3::types::PyString>) -> Self {
```
alex: We should validate that `inner` is a valid PrintableString, rather than defering that to serialization

facutuesca: Should I replicate the logic from [rust-asn1's PrintableString::verify](https://github.com/alex/rust-asn1/blob/7ff2acc222bf333d75ddda14c6bb0dc6704f5d8d/src/types.rs#L402) ? We could also make `verify` public.

alex: `asn1::PrintableString::new(s).is_none()` should be enough to check if its invalid

facutuesca: Or call rust-asn1's `PrintableString::new` and discard the result

alex (approved):

PR #13513, opened 2025-09-27
asn1: Add support for `UtcTime` and `GeneralizedTime`

facutuesca:
This adds support for encoding/decoding `UtcTime` and `GeneralizedTime`. 

```python
@asn1.sequence
class MySeq:
    field: asn1.UtcTime

dt = datetime.datetime(1990, 12, 31, 10, 10, 10, tzinfo=datetime.UTC)

object = MySeq(field=asn1.UtcTime(field=asn1.UtcTime(dt)))
encoded = encode_der(object)

assert decoded.field.as_datetime() == dt
```


src/rust/src/declarative_asn1/decode.rs
```diff
@@ -57,6 +59,50 @@ fn decode_printable_string<'a>(
...
+        dt.year().into(),
+        dt.month(),
+        dt.day(),
+        dt.hour(),
+        dt.minute(),
+        dt.second(),
+        0,
+        Some(&pyo3::types::PyTzInfo::utc(py)?.to_owned()),
+    )?
+    .unbind();
```
alex: I think we have a function that does this -- `datetime_to_py_utc`. Your implementation is better than that one, so assuming it works on all python versions, can you replace that impl with this one and call it here?

src/rust/src/declarative_asn1/encode.rs
```diff
@@ -87,6 +90,27 @@ impl asn1::Asn1Writable for AnnotatedTypeObject<'_> {
                         .ok_or(asn1::WriteError::AllocationError)?;
                 write_value(writer, &printable_string)
             }
+            Type::UtcTime() => {
+                let val: &pyo3::Bound<'_, UtcTime> = value
+                    .downcast()
+                    .map_err(|_| asn1::WriteError::AllocationError)?;
+                let datetime_info = get_datetime_info(py, &val.get().inner)
+                    .map_err(|_| asn1::WriteError::AllocationError)?;
```
alex: Similarly, we have `py_to_datetime`, but I think we want an impl using the pyo3 datetime types.

src/rust/src/declarative_asn1/decode.rs
alex: Actually -- if you can specifically do a standalone PR that changes those functions to use pyo3's types that'd be best.

src/rust/src/declarative_asn1/encode.rs
facutuesca: For this one we can't: the methods to access the `datetime` fields are not available when building with the limited API: https://docs.rs/pyo3/latest/pyo3/types/trait.PyDateAccess.html
> Available on non-Py_LIMITED_API only.
Trait for accessing the date components of a struct containing a date.

facutuesca: the tzinfo field is available though: https://docs.rs/pyo3/latest/pyo3/types/trait.PyTzInfoAccess.html

alex: Ah, well, we still want to use `py_to_datetime` here.

src/rust/src/declarative_asn1/decode.rs
facutuesca: PR open here: https://github.com/pyca/cryptography/pull/13514

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -57,6 +59,37 @@ fn decode_printable_string<'a>(
...
+fn decode_utc_time<'a>(
+    py: pyo3::Python<'a>,
+    parser: &mut Parser<'a>,
+) -> ParseResult<pyo3::Bound<'a, UtcTime>> {
+    let value = parser.read_element::<asn1::UtcTime>()?;
+    let dt = value.as_datetime();
+
+    let inner = crate::x509::datetime_to_py_utc(py, dt)?
+        .downcast::<pyo3::types::PyDateTime>()?
+        .clone()
```
alex: I don't think you need this clone.

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -57,6 +59,37 @@ fn decode_printable_string<'a>(
...
+    py: pyo3::Python<'a>,
+    parser: &mut Parser<'a>,
+) -> ParseResult<pyo3::Bound<'a, GeneralizedTime>> {
+    let value = parser.read_element::<asn1::GeneralizedTime>()?;
+    let dt = value.as_datetime();
+    let microseconds = value.nanoseconds().map_or(0, |nanos| nanos / 1_000);
+
+    let inner = crate::x509::datetime_to_py_utc_with_microseconds(py, dt, microseconds)?
+        .downcast::<pyo3::types::PyDateTime>()?
+        .clone()
```
alex: same

tests/hazmat/asn1/test_api.py
```diff
@@ -21,6 +22,68 @@ def test_invalid_printable_string(self) -> None:
...
+    def test_invalid_utc_time(self) -> None:
+        with pytest.raises(
+            ValueError,
+            match="cannot initialize with naive datetime object",
+        ):
+            # We don't allow naive datetime objects
+            asn1.UtcTime(datetime.datetime(2000, 1, 1, 10, 10, 10))
+
+        with pytest.raises(ValueError, match="invalid UtcTime"):
+            # UtcTime does not support dates before 2050
```
alex:
```suggestion
            # UtcTime does not support dates before 1950
```

src/rust/src/declarative_asn1/decode.rs

src/rust/src/declarative_asn1/decode.rs

src/rust/src/declarative_asn1/types.rs
```diff
@@ -108,6 +116,116 @@ impl PrintableString {
...
 }
 
+pub(crate) struct DateTimeInfo {
+    pub(crate) datetime: asn1::DateTime,
+    pub(crate) nanoseconds: Option<u32>,
+}
+
+// Internal utility function used to extract data from a Python datetime
+// object. Used by both GeneralizedTime and UtcTime.
+pub(crate) fn get_datetime_info(
```
alex: This should be organized next to `py_to_datetime` and have an analogous name. I don't think we need a `DateTimeInfo` type -- a tuple is fine.

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -57,6 +59,45 @@ fn decode_printable_string<'a>(
...
+    let value = parser.read_element::<asn1::UtcTime>()?;
+    let dt = value.as_datetime();
+
+    let inner = crate::x509::datetime_to_py_utc(py, dt)?
+        .downcast_into::<pyo3::types::PyDateTime>()
+        .map_err(|_| {
+            pyo3::exceptions::PyValueError::new_err(
+                "error converting to datetime object".to_string(),
+            )
+        })?
```
alex: The downcast should never fail, so we don't need a specific error message, you can just `?`

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -57,6 +59,45 @@ fn decode_printable_string<'a>(
...
+    let dt = value.as_datetime();
+    let microseconds = value.nanoseconds().map_or(0, |nanos| nanos / 1_000);
+
+    let inner = crate::x509::datetime_to_py_utc_with_microseconds(py, dt, microseconds)?
+        .downcast_into::<pyo3::types::PyDateTime>()
+        .map_err(|_| {
+            pyo3::exceptions::PyValueError::new_err(
+                "error converting to datetime object".to_string(),
+            )
+        })?
```
alex: same

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -57,6 +59,45 @@ fn decode_printable_string<'a>(
...
+    Ok(pyo3::Bound::new(py, UtcTime { inner })?)
+}
+
+fn decode_generalized_time<'a>(
+    py: pyo3::Python<'a>,
+    parser: &mut Parser<'a>,
+) -> ParseResult<pyo3::Bound<'a, GeneralizedTime>> {
+    let value = parser.read_element::<asn1::GeneralizedTime>()?;
+    let dt = value.as_datetime();
+    let microseconds = value.nanoseconds().map_or(0, |nanos| nanos / 1_000);
```
alex: hmm, this is going to truncate off some precision, which means things won't roundtrip. is this a problem?

src/rust/src/declarative_asn1/encode.rs
```diff
@@ -87,6 +89,29 @@ impl asn1::Asn1Writable for AnnotatedTypeObject<'_> {
                         .ok_or(asn1::WriteError::AllocationError)?;
                 write_value(writer, &printable_string)
             }
+            Type::UtcTime() => {
+                let val: &pyo3::Bound<'_, UtcTime> = value
+                    .downcast()
+                    .map_err(|_| asn1::WriteError::AllocationError)?;
+                let (datetime, _) =
+                    crate::x509::py_to_datetime_with_microseconds(py, val.get().inner.bind(py))
+                        .map_err(|_| asn1::WriteError::AllocationError)?;
```
alex: Why are we using the microsecond variant if we're going to ignore microseconds?

src/rust/src/x509/common.rs
```diff
@@ -520,34 +521,63 @@ pub(crate) fn datetime_to_py_utc<'p>(
...
+    let normalized: pyo3::Bound<'_, pyo3::types::PyAny>;
     let val_utc = if val.get_tzinfo().is_none() {
-        val.into_any()
+        val.as_any()
     } else {
         let utc = pyo3::types::PyTzInfo::utc(py)?;
-        val.call_method1(pyo3::intern!(py, "astimezone"), (utc,))?
+        normalized = val.call_method1(pyo3::intern!(py, "astimezone"), (utc,))?;
+        &normalized
     };
```
alex: I'm not sure I understand the intent here.

src/rust/src/declarative_asn1/decode.rs
facutuesca: mm yeah that's a good point. I think we should do what `rust-asn1` does, and fail if the decoded fraction exceeds our maximum precision (in our case, 999.999 microseconds).

src/rust/src/x509/common.rs
facutuesca: yeah this was me trying to workaround the fact that I changed the parameter type from a `Bound` to a `&Bound`, but ended up making things more complicated than needed.

I made it a `Bound` and now the logic is the same as before.

src/rust/src/declarative_asn1/decode.rs

src/rust/src/declarative_asn1/decode.rs

src/rust/src/declarative_asn1/decode.rs
facutuesca: implemented the above, and added a test for it

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -57,6 +59,42 @@ fn decode_printable_string<'a>(
...
+) -> ParseResult<pyo3::Bound<'a, GeneralizedTime>> {
+    let value = parser.read_element::<asn1::GeneralizedTime>()?;
+    let dt = value.as_datetime();
+
+    let microseconds = match value.nanoseconds() {
+        Some(x) if x % 1_000 == 0 => Ok(x / 1_000),
+        Some(_) => Err(pyo3::exceptions::PyValueError::new_err(
+            "decoded GeneralizedTime data has higher precision than supported".to_string(),
+        )),
+        None => Ok(0),
```
alex: This should be `None`, instead of `Ok(0)` I think?

src/rust/src/declarative_asn1/encode.rs

src/rust/src/declarative_asn1/decode.rs
alex: Err, wait sorry I see what's going. Instead of `?` on the match block, I'd just have a `return` in the Err branch and then drop the `Ok()`

facutuesca: No, we use it for `datetime_to_py_utc_with_microseconds`, which takes a `u32` for the microseconds.

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -57,6 +59,44 @@ fn decode_printable_string<'a>(
...
+) -> ParseResult<pyo3::Bound<'a, GeneralizedTime>> {
+    let value = parser.read_element::<asn1::GeneralizedTime>()?;
+    let dt = value.as_datetime();
+
+    let microseconds = match value.nanoseconds() {
+        Some(x) if x % 1_000 == 0 => x / 1_000,
+        Some(_) => {
+            return Err(pyo3::exceptions::PyValueError::new_err(
+                "decoded GeneralizedTime data has higher precision than supported".to_string(),
+            ))?
```
alex: no `?`, you're directly returning

alex (approved):

PR #13542, opened 2025-10-02
asn1: Add support for `OPTIONAL`

facutuesca:

Example:
```python
@asn1.sequence
class Example:
    a: typing.Union[bool, None] # or bool | None
    b: int

asn1.decode_der(Example, encoded_bytes)
```

Decoding is done by peeking the next tag, and returning `None` if it doesn't match the tag of the `T` inside `Option<T>`

### DIFF
--- a/src/cryptography/hazmat/asn1/asn1.py
+++ b/src/cryptography/hazmat/asn1/asn1.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import dataclasses
 import sys
+import types
 import typing
 
 if sys.version_info < (3, 11):
@@ -16,10 +17,13 @@ if sys.version_info < (3, 11):
     # once the min version is >= 3.9
     if sys.version_info < (3, 9):
         get_type_hints = typing_extensions.get_type_hints
+        get_type_args = typing_extensions.get_args
     else:
         get_type_hints = typing.get_type_hints
+        get_type_args = typing.get_args
 else:
     get_type_hints = typing.get_type_hints
+    get_type_args = typing.get_args
 
 from cryptography.hazmat.bindings._rust import declarative_asn1
 
@@ -29,6 +33,16 @@ U = typing.TypeVar("U")
 
 decode_der = declarative_asn1.decode_der
 encode_der = declarative_asn1.encode_der
+
+
+def _is_union(field_type: type) -> bool:
+    # NOTE: types.UnionType for `T | U`, typing.Union for `Union[T, U]`.
+    union_types = (
+        (types.UnionType, typing.Union)
+        if hasattr(types, "UnionType")
+        else (typing.Union,)
+    )
+    return typing.get_origin(field_type) in union_types
 
 
 def _normalize_field_type(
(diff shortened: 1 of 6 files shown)


src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -41,6 +55,19 @@ def _normalize_field_type(
         if not isinstance(annotated_root, declarative_asn1.AnnotatedType):
             raise TypeError(f"unsupported root type: {annotated_root}")
         return annotated_root
+    elif _is_union(field_type):
+        union_args = get_type_args(field_type)
+        if len(union_args) == 2 and type(None) in union_args:
```
alex: You can use `types.NoneType` here.

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -41,6 +55,19 @@ def _normalize_field_type(
...
+            # A Union between a type and None is an OPTIONAL
+            optional_type = (
+                union_args[0] if union_args[1] is type(None) else union_args[1]
+            )
+            annotated_type = _normalize_field_type(optional_type, field_name)
+            rust_field_type = declarative_asn1.Type.Option(annotated_type)
```
alex, line 73: Hmm, do we need to reject nesting (`(x | None) | None`)?

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -125,6 +125,16 @@ pub(crate) fn decode_annotated_type<'a>(
                 Ok(val)
             })
         }
+        Type::Option(cls) => {
+            let inner_tag = type_to_tag(cls.get().inner.get());
+            match parser.peek_tag() {
+                Some(t) if t == inner_tag => {
+                    let decoded_value = decode_annotated_type(py, parser, cls.get())?;
+                    Ok(decoded_value.into_any())
```
alex: Is `into_any` doing anything here?

facutuesca: nope, fixed

src/cryptography/hazmat/asn1/asn1.py
facutuesca: Union annotations are flattened and redundant types are skipped: https://docs.python.org/3.10/library/typing.html?highlight=union#typing.Union.
I added a test for it just in case

src/cryptography/hazmat/asn1/asn1.py

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -31,6 +40,16 @@
 encode_der = declarative_asn1.encode_der
 
 
+def _is_union(field_type: type) -> bool:
+    # NOTE: types.UnionType for `T | U`, typing.Union for `Union[T, U]`.
+    union_types = (
+        (types.UnionType, typing.Union)
+        if hasattr(types, "UnionType")
```
alex, line 49: Can you add a comment for when we drop the hasattr? (I assume its some new python version)

facutuesca: done!

PR #13560, opened 2025-10-04
convert pbkdf2hmac to rust

reaperhulk:


src/rust/src/backend/kdf.rs
```diff
@@ -15,23 +15,98 @@ use crate::buf::CffiBuf;
...
+    used: bool,
+}
+
+#[pyo3::pymethods]
+impl Pbkdf2Hmac {
+    #[new]
+    #[pyo3(signature = (algorithm, length, salt, iterations, backend=None))]
+    fn new(
+        py: pyo3::Python<'_>,
+        algorithm: pyo3::Py<pyo3::PyAny>,
```
alex: this should use  `pyo3::Bound` since we're not saving it

PR #13562, opened 2025-10-04
asn1: Add support for `DEFAULT`

facutuesca:
This PR adds support for `DEFAULT` fields to the ASN.1 API. The field types must be annotated with `asn1.Default(value=X)`:

```python
@asn1.sequence
class Example
  field: typing.Annotated[int, asn1.Default(value=3)]
```

During decoding, if the field is not found then the object is initialized with the default value. During encoding, if the field has the same value as the default value, it is skipped (not encoded).

Trying to declare a field that is both OPTIONAL and DEFAULT will fail, warning the user to use only one:

```python
IntWithDefault = typing.Annotated[int, asn1.Default(value=3)]

@asn1.sequence
class Invalid:
  # this will raise a TypeError
  invalid_field1: IntWithDefault | None   

  # this will also raise
  invalid_field2: typing.Annotated[int | None, asn1.Default(value=3)]  
```

### DIFF
--- a/src/cryptography/hazmat/asn1/asn1.py
+++ b/src/cryptography/hazmat/asn1/asn1.py
@@ -189,7 +189,7 @@ else:
 
 
 # TODO: replace with `Default[U]` once the min Python version is >= 3.12
-@dataclasses.dataclass
+@dataclasses.dataclass(frozen=True)
 class Default(typing.Generic[U]):
     value: U
 
--- a/src/cryptography/hazmat/asn1/asn1.py
+++ b/src/cryptography/hazmat/asn1/asn1.py
@@ -76,7 +76,7 @@ def _normalize_field_type(
     # from it if it exists.
     if get_type_origin(field_type) is Annotated:
         annotation = _extract_annotation(field_type.__metadata__)
-        field_type = get_type_args(field_type)[0]
+        field_type, _ = get_type_args(field_type)
     else:
         annotation = declarative_asn1.Annotation()
 
--- a/src/cryptography/hazmat/asn1/asn1.py
+++ b/src/cryptography/hazmat/asn1/asn1.py
@@ -61,8 +61,8 @@ def _is_union(field_type: type) -> bool:
 def _extract_annotation(metadata: tuple) -> declarative_asn1.Annotation:
     default = None
     for raw_annotation in metadata:
-        if isinstance(raw_annotation, declarative_asn1.Default):
-            default = raw_annotation
+        if isinstance(raw_annotation, Default):
+            default = raw_annotation.value
         else:
             raise TypeError(f"unsupported annotation: {raw_annotation}")
 
@@ -188,7 +188,12 @@ else:
         return dataclass_cls
 
 
-Default = declarative_asn1.Default
+# TODO: replace with `Default[U]` once the min Python version is >= 3.12
+@dataclasses.dataclass
+class Default(typing.Generic[U]):
+    value: U
+
+
 PrintableString = declarative_asn1.PrintableString
 UtcTime = declarative_asn1.UtcTime
 GeneralizedTime = declarative_asn1.GeneralizedTime
--- a/src/cryptography/hazmat/asn1/asn1.py
+++ b/src/cryptography/hazmat/asn1/asn1.py
@@ -62,7 +62,7 @@ def _extract_annotation(metadata: tuple) -> declarative_asn1.Annotation:
     default = None
     for raw_annotation in metadata:
         if isinstance(raw_annotation, declarative_asn1.Default):
-            default = declarative_asn1.Default(value=raw_annotation.value)
+            default = raw_annotation
         else:
             raise TypeError(f"unsupported annotation: {raw_annotation}")
 
--- a/src/cryptography/hazmat/asn1/asn1.py
+++ b/src/cryptography/hazmat/asn1/asn1.py
@@ -94,14 +94,19 @@ def _normalize_field_type(
             )
             annotated_type = _normalize_field_type(optional_type, field_name)
 
-            if (
-                annotation.default is not None
-                or annotated_type.annotation.default is not None
-            ):
+            if not annotated_type.annotation.is_empty():
+                raise TypeError(
+                    "optional (`X | None`) types cannot have `X` "
+                    "annotated: annotations must apply to the union "
+                    "(i.e: `Annotated[X | None, annotation]`)"
+                )
+
+            if annotation.default is not None:
                 raise TypeError(
                     "optional (`X | None`) types should not have a DEFAULT "
                     "annotation"
                 )
+
             rust_field_type = declarative_asn1.Type.Option(annotated_type)
         else:
             raise TypeError(
--- a/src/cryptography/hazmat/asn1/asn1.py
+++ b/src/cryptography/hazmat/asn1/asn1.py
@@ -18,12 +18,15 @@ if sys.version_info < (3, 11):
     if sys.version_info < (3, 9):
         get_type_hints = typing_extensions.get_type_hints
         get_type_args = typing_extensions.get_args
+        Annotated = typing_extensions.Annotated
     else:
         get_type_hints = typing.get_type_hints
         get_type_args = typing.get_args
+        Annotated = typing.Annotated
 else:
     get_type_hints = typing.get_type_hints
     get_type_args = typing.get_args
+    Annotated = typing.Annotated
 
 if sys.version_info < (3, 10):
     NoneType = type(None)
@@ -52,10 +55,25 @@ def _is_union(field_type: type) -> bool:
     return typing.get_origin(field_type) in union_types
 
 
+def _extract_annotation(metadata: tuple) -> declarative_asn1.Annotation:
+    default = None
+    for raw_annotation in metadata:
+        if isinstance(raw_annotation, declarative_asn1.Default):
+            default = declarative_asn1.Default(value=raw_annotation.value)
+        else:
+            raise TypeError(f"unsupported annotation: {raw_annotation}")
+
+    return declarative_asn1.Annotation(default=default)
+
+
 def _normalize_field_type(
     field_type: typing.Any, field_name: str
 ) -> declarative_asn1.AnnotatedType:
-    annotation = declarative_asn1.Annotation()
+    if typing.get_origin(field_type) is Annotated:
+        annotation = _extract_annotation(field_type.__metadata__)
+        field_type = get_type_args(field_type)[0]
+    else:
+        annotation = declarative_asn1.Annotation()
 
     if hasattr(field_type, "__asn1_root__"):
         annotated_root = field_type.__asn1_root__
@@ -70,6 +88,15 @@ def _normalize_field_type(
                 union_args[0] if union_args[1] is type(None) else union_args[1]
             )
             annotated_type = _normalize_field_type(optional_type, field_name)
+
+            if (
+                annotation.default is not None
+                or annotated_type.annotation.default is not None
+            ):
+                raise TypeError(
+                    "optional (`X | None`) types should not have a DEFAULT "
+                    "annotation"
+                )
             rust_field_type = declarative_asn1.Type.Option(annotated_type)
         else:
             raise TypeError(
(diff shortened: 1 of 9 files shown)


src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -49,13 +55,30 @@ def _is_union(field_type: type) -> bool:
         if hasattr(types, "UnionType")
         else (typing.Union,)
     )
-    return typing.get_origin(field_type) in union_types
+    return get_type_origin(field_type) in union_types
+
+
+def _extract_annotation(metadata: tuple) -> declarative_asn1.Annotation:
```
alex, line 61: Does this not have a more specific type than just tuple?

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -49,13 +55,30 @@ def _is_union(field_type: type) -> bool:
...
 
 def _normalize_field_type(
     field_type: typing.Any, field_name: str
 ) -> declarative_asn1.AnnotatedType:
-    annotation = declarative_asn1.Annotation()
+    # Strip the `Annotated[...]` off, and populate the annotation
+    # from it if it exists.
+    if get_type_origin(field_type) is Annotated:
+        annotation = _extract_annotation(field_type.__metadata__)
+        field_type = get_type_args(field_type)[0]
```
alex:
```suggestion
        field_type, _ = get_type_args(field_type)
```

src/cryptography/hazmat/bindings/_rust/declarative_asn1.pyi
```diff
@@ -20,10 +20,17 @@ class Type:
     PyStr: typing.ClassVar[type]
 
 class Annotation:
+    default: Default | None
     def __new__(
         cls,
+        default: Default | None = None,
     ) -> Annotation: ...
 
+class Default:
```
alex: Should this be `Default[T]`? Does it matter?

alex (commented):
It should be an error to parse a value with an explicitly encoded default.

src/cryptography/hazmat/asn1/asn1.py
facutuesca: I don't think so: https://github.com/python/cpython/blob/ac853ece0895628ce2bd9e506fd8e39e6de2a4dc/Lib/typing.py#L2223

src/cryptography/hazmat/bindings/_rust/declarative_asn1.pyi
facutuesca: Fixed. I don't think it matters much currently, but it does constrain the type of the field to be the same as the constructor, which is better than just using `Any` everywhere.

src/cryptography/hazmat/asn1/asn1.py

facutuesca:
> It should be an error to parse a value with an explicitly encoded default.

fixed, see last commit

alex:
(Conflicts + please rebase on main for coverage)

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -49,13 +55,30 @@ def _is_union(field_type: type) -> bool:
...
     )
-    return typing.get_origin(field_type) in union_types
+    return get_type_origin(field_type) in union_types
+
+
+def _extract_annotation(metadata: tuple) -> declarative_asn1.Annotation:
+    default = None
+    for raw_annotation in metadata:
+        if isinstance(raw_annotation, declarative_asn1.Default):
+            default = declarative_asn1.Default(value=raw_annotation.value)
```
alex: What's this actually doing? It looks like we already have an instance of the right type?

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -70,6 +93,15 @@ def _normalize_field_type(
                 union_args[0] if union_args[1] is type(None) else union_args[1]
             )
             annotated_type = _normalize_field_type(optional_type, field_name)
+
+            if (
+                annotation.default is not None
+                or annotated_type.annotation.default is not None
```
alex: Why do we need to check both of these?

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -116,24 +127,30 @@ pub(crate) fn decode_annotated_type<'a>(
...
+        Type::PyStr() => decode_pystr(py, parser)?.into_any(),
+        Type::PrintableString() => decode_printable_string(py, parser)?.into_any(),
+        Type::UtcTime() => decode_utc_time(py, parser)?.into_any(),
+        Type::GeneralizedTime() => decode_generalized_time(py, parser)?.into_any(),
+    };
+
+    match &ann_type.annotation.default {
+        Some(default) if decoded.eq(default.value.bind(py))? => Err(CryptographyError::Py(
+            pyo3::exceptions::PyValueError::new_err(
+                "invalid DER: DEFAULT value was present in encoded data".to_string(),
```
alex:
```suggestion
                "invalid DER: DEFAULT value was explicitly encoded".to_string(),
```

src/rust/src/declarative_asn1/decode.rs
alex, line 99: Idea for a potential refactor (separate PR). Create a:

```
fn read_value<T: Asn1Readable>(&mut Parser, default: Option<PyAny...>, cb: impl FnOnce(T) -> CryptographyResult<PyAny>) -> CryptographyResult<PyAny> {}
```

function and use that in each path so that we bring the default handling closer to the parse call site. It may also make things easier for Explicit/Implicit (but maybe not)

Food for thought.

src/rust/src/declarative_asn1/types.rs
```diff
@@ -66,13 +67,42 @@ pub struct AnnotatedTypeObject<'a> {
...
+#[pyo3::pymethods]
+impl Default {
+    #[new]
+    #[pyo3(signature = (value))]
+    fn new(value: pyo3::Py<pyo3::types::PyAny>) -> Self {
+        Self { value }
+    }
+}
+
+impl Clone for Default {
```
alex: Why do we need this?

src/cryptography/hazmat/asn1/asn1.py
facutuesca: yup, it was a leftover after a refactor, fixed.

src/cryptography/hazmat/asn1/asn1.py
facutuesca: first one covers:
```python
Annotated[T | None, Default(x)],
```

second one covers:
```python
Annotated[T, Default(x)]  |  None
```

src/rust/src/declarative_asn1/types.rs
facutuesca: Because we pass `Default` by value in `Annotation`'s constructor:

```python
#[pyo3::pymethods]
impl Annotation {
    #[new]
    #[pyo3(signature = (default = None))]
    fn new(default: Option<Default>) -> Self {
        Self { default }
    }
}
```


> Classes can also be passed by value if they can be cloned, i.e. they automatically implement FromPyObject if they implement Clone, e.g. via #[derive(Clone)]:


[src](https://pyo3.rs/main/class.html#:~:text=Classes%20can%20also%20be%20passed%20by%20value%20if%20they%20can%20be%20cloned%2C%20i.e.%20they%20automatically%20implement%20FromPyObject%20if%20they%20implement%20Clone%2C%20e.g.%20via%20%23%5Bderive(Clone)%5D%3A)

src/cryptography/hazmat/asn1/asn1.py
alex: Ooof, that second one seems fairly perplexing to me. Is there a strong argument that we should support it?

src/rust/src/declarative_asn1/types.rs
alex: I think it'd be better to take it as a reference and then do an explicit clone_ref on the value -- in general, we should avoid `attach()` where practical.

facutuesca: Also, creating a getter via `#[pyo3(get)]` requires `Default` to:

> For get the field type T must implement either &T: IntoPyObject or T: IntoPyObject + Clone.
[src](https://pyo3.rs/v0.26.0/class.html#object-properties-using-pyo3get-set)

Should I then implement `IntoPyObject` for `&Default` and `&Annotation`? (since we have getters for them in `Annotation` and `AnnotatedType` respectively)

There is a derive macro [`IntoPyObjectRef`](https://pyo3.rs/v0.26.0/conversions/traits.html#derive-macro), but the Python object it returns for structs is a dict, so fields are accessed as keys (`annotation["default"]` instead of `annotation.default`)

alex: For a gettable attribute, we should just store it as `Py<Default>`

src/cryptography/hazmat/asn1/asn1.py
facutuesca: not really. I changed it so that we raise if we find annotations in the inner type of the Option, and also added a test for it.

src/rust/src/declarative_asn1/types.rs

src/rust/src/declarative_asn1/types.rs
```diff
@@ -64,15 +65,40 @@ pub struct AnnotatedTypeObject<'a> {
...
+        Self { default }
+    }
+
+    #[pyo3(signature = ())]
+    fn is_empty(&self) -> bool {
+        self.default.is_none()
+    }
+}
+
+#[pyo3::pyclass(frozen, module = "cryptography.hazmat.bindings._rust.asn1")]
```
alex: Sorry, should have asked this earlier: why do we hvae a `Default` type at all? Why not just have the default field on the Annotation be a PyAny?

facutuesca: it's how the user specifies the default value. Otherwise, it would be something like:
```python
field: Annotated[int, 3, other_annotations...]
```

alex: duh, of course.

Once we have an asn1::Annotation object, we don't need to keep the default wrapper around, right?

facutuesca: Can we have an `asn1::Annotation` object? The thing about using `typing.Annotated` is that type checkers know that it's a special form and treat it accordingly. If we replaced it with our own `asn1::Annotation` object, we would lose that (for example, `asn1.Annotated[int, ...]` wouldn't be treated as an `int`)

alex: Let me read through again and see if I can ask a better question

src/cryptography/hazmat/bindings/_rust/declarative_asn1.pyi
```diff
@@ -20,9 +20,20 @@ class Type:
     PyStr: typing.ClassVar[type]
 
 class Annotation:
+    default: Default | None
     def __new__(
         cls,
+        default: Default | None = None,
```
alex: I guess my question is: what's the argument for this being a `Default` instance here vs. just a `typing.Any`?

alex (approved):

src/cryptography/hazmat/bindings/_rust/declarative_asn1.pyi
facutuesca: ah yeah that makes sense. I now changed it so that `Annotation.default` is a `typing.Any | None`, and made the `Default` class Python-only (moved to `asn1.py`), since it will only be used at the Python layer by the user.

```python
field: Annotated[int, asn1.Default(3)]

#.... which we convert to an Annotation object without wrapping into a Default tpe:

def _extract_annotation(metadata: tuple) -> declarative_asn1.Annotation:
    default = None
    for raw_annotation in metadata:
        if isinstance(raw_annotation, Default):
            default = raw_annotation.value
        else:
            raise TypeError(f"unsupported annotation: {raw_annotation}")

    return declarative_asn1.Annotation(default=default)



```

alex:
hmm, what's going on with CI?

facutuesca:
> I don't love teh `is_empty()` name -- perhaps it should be `has_default()` or something. Can be a follow up PR

It's because the Annotation object will have annotations other than `default`, and we need to check if any of them are `None`

facutuesca:
> hmm, what's going on with CI?

It was because of a bug in older Python versions: https://github.com/python/cpython/issues/112281. Making `Default` hashable solves it

PR #13591, opened 2025-10-07
Two minor fixes for code blocks in x509 docs

iyanmv:
This PR fixes two minor issues I noticed in the X.509 docs.

* A missing `import datetime` in the section [Creating a self-signed certificate](https://cryptography.io/en/latest/x509/tutorial/#creating-a-self-signed-certificate)
* A missing variable definition in the section [X.509 Verification](https://cryptography.io/en/latest/x509/verification/#module-cryptography.x509.verification)

### DIFF
--- a/docs/x509/verification.rst
+++ b/docs/x509/verification.rst
@@ -79,7 +79,7 @@ the root of trust:
     >>> # See the documentation on `time` below for more details. If
     >>> # significant time passes between creating a verifier and performing a
     >>> # verification, you may encounter issues with certificate expiration.
-    >>> verification_time = datetime.now(timezone.utc)
+    >>> verification_time = datetime.now(timezone.utc)  # doctest: +SKIP
     >>> builder = builder.time(verification_time)
     >>> verifier = builder.build_server_verifier(DNSName("cryptography.io"))
     >>> # NOTE: peer and untrusted_intermediates are Certificate and
--- a/docs/x509/verification.rst
+++ b/docs/x509/verification.rst
@@ -72,13 +72,14 @@ the root of trust:
     >>> from cryptography.x509 import Certificate, DNSName, load_pem_x509_certificates
     >>> from cryptography.x509.verification import PolicyBuilder, Store
     >>> import certifi
-    >>> from datetime import datetime
+    >>> from datetime import datetime, timezone
     >>> with open(certifi.where(), "rb") as pems:
     ...    store = Store(load_pem_x509_certificates(pems.read()))
     >>> builder = PolicyBuilder().store(store)
     >>> # See the documentation on `time` below for more details. If
     >>> # significant time passes between creating a verifier and performing a
     >>> # verification, you may encounter issues with certificate expiration.
+    >>> verification_time = datetime.now(timezone.utc)
     >>> builder = builder.time(verification_time)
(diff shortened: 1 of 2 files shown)


docs/x509/verification.rst
```diff
@@ -72,13 +72,14 @@ the root of trust:
...
     >>> import certifi
-    >>> from datetime import datetime
+    >>> from datetime import datetime, timezone
     >>> with open(certifi.where(), "rb") as pems:
     ...    store = Store(load_pem_x509_certificates(pems.read()))
     >>> builder = PolicyBuilder().store(store)
     >>> # See the documentation on `time` below for more details. If
     >>> # significant time passes between creating a verifier and performing a
     >>> # verification, you may encounter issues with certificate expiration.
+    >>> verification_time = datetime.now(timezone.utc)
```
alex: Unfortunately this isn't going to work -- these code blocks are run as part of the tests, and we need a fixed time to ensure the cert passes.

My recommendation would be to do something like:

```
>>> # verification_time = datetime.now(timezone.utc)
```
(i.e., have hte line be commented out)

iyanmv: What about using `# doctest: +SKIP` instead? Then it would still work for people copying-pasting the full block, but that line should be skipped by the tests.

alex: No, we'd like the doctest to actually run, that's how we know our examples work.

iyanmv: If I understood the docs correctly, that comment would skip only THAT particular line, not the whole code block. Isn't that exactly what you want?

alex: Does it skipping _running_ that line or merely skip assertions about it?

iyanmv: Just to be clear, this is what I'm suggesting:

```python
    >>> from cryptography.x509 import Certificate, DNSName, load_pem_x509_certificates
    >>> from cryptography.x509.verification import PolicyBuilder, Store
    >>> import certifi
    >>> from datetime import datetime, timezone
    >>> # See the documentation on `time` below for more details. If
    >>> # significant time passes between creating a verifier and performing a
    >>> # verification, you may encounter issues with certificate expiration.
    >>> verification_time = datetime.now(timezone.utc)  # doctest: +SKIP
    # (...)
```

iyanmv: I think [this](https://stackoverflow.com/questions/1809037/python-doctest-skip-entire-block) confirms my understanding that only that single line would be skipped when running the tests.

alex: I guess I'd be curious what that looks like rendered in the docs, but if the tests pass on it, I'd certainly be willing to consider it.

reaperhulk: Feel free to push a commit and we'll see if it works 😄

iyanmv: Alright, let's see.

iyanmv: The comment is removed from the code block in the html as desired, and I guess the line is not run since [CI / linux (3.12, docs, openssl, 3.6.0) (pull_request)](https://github.com/pyca/cryptography/actions/runs/18308784529/job/52132227643?pr=13591) passed.

alex (approved):
thanks!

PR #13615, opened 2025-10-08
bump python version for downstream testing

reaperhulk:


reaperhulk:
Updating to 3.13 breaks `awslabs/aws-dynamodb-encryption-python` and 3.14 breaks that as well as `twisted` and `awslabs/aws-encryption-sdk-python`

PR #13616, opened 2025-10-08
document pkcs7 decryption of binary data

reaperhulk:
Closes #13341


reaperhulk:
replaces https://github.com/pyca/cryptography/pull/13341

alex (approved):

gav2xlin:
@reaperhulk @alex  You have three functions, but not one.

pkcs7_decrypt_smime and pkcs7_decrypt_pem call pkcs7_decrypt_der

The fact that these two functions also accept an empty list for processing binary data is not described anywhere in the documentation at all.

PR #13619, opened 2025-10-08
Simplify wheel builds using new uv support for 3.14t

ngoldbaum:
uv 0.9.0 no longer considers 3.14t to be experimental: https://github.com/astral-sh/uv/pull/16142


alex:
FYI, I've already got https://github.com/pyca/cryptography/pull/13618 up -- I see you changed one more site, but I'm not sure that one is correct. I think we do need --python on linux?

ngoldbaum:
Sorry, missed that! I'll close this. And I think you're right, it still seems to be needed on Linux. Sorry for the noise.

PR #13630, opened 2025-10-09
Fix string interpolation for rust 1.88.0.

jahkosha:
Greetings,

This fix an error that occur when trying to build the library using rustc 1.88.0.
Following a discussion on IRC the problem was replicated by @alex 

Thank you for this great library!


reaperhulk:
Thanks for the contribution!

PR #13634, opened 2025-10-10
Add client hello callback related functions

schwabe:
This exposes the OpenSSL functions SSL_CTX_set_client_hello_cb, SSL_client_hello_get0_ext and SSL_client_hello_get1_extensions_present.

These are required to implement to the client hello callback functionality in pyOpenSSL.


schwabe:
@Tarnum-tst I assume that is about not having names for the parameter? If yes I fixed that. If there was something else wrong, please point it out to me.

Issue #13245, opened 2025-08-05
Support mutable outputs for key derivation and ciphers to allow secure memory wiping


HaZeShade:
## Issue: Lack of Mutable Output for Secrets
Most (if not all) of Cryptography's systems that handle and/or generate sensitive material **do not support a mutable output.** `HKDF().derive()` for example:
```python
def derive(self, key_material: utils.Buffer) -> bytes:
        utils._check_byteslike("key_material", key_material)
        return self._hkdf_expand.derive(self._extract(key_material))
```
This returns a `bytes` object, which is **immutable**. This is also true for other subsystems such as `hazmat.primitives.ciphers.aead.AESGCM`.

This entirely destroys any attempt at pure memory hygiene for anyone using the library, as an immutable object **can not be effectively zeroed in place**, creating a vulnerability, since this memory will persist at least until Python garbage collects, and at most until the program is closed.

## Theoretical Solution
I am aware the Cryptography HKDF implementation is in Rust rather than C, and while I am not well versed in Rust, I will provide a proof of concept solution that I have created wrapping OpenSSL. I am not a professional at any of this, so though my implementation may not be perfect, it should work as a proof of concept. Below this wrapper I will also leave a zeroing function, as it is used in my glue code in Python.
```c
#include <openssl/kdf.h>
#include <openssl/evp.h>

int hkdf(
    const uint8_t* ikm,  uint32_t ikmlen,
    const uint8_t* salt, uint32_t saltlen,
    uint8_t* out,        uint32_t outlen,
    uint8_t* info,       uint32_t infolen
) {
    EVP_PKEY_CTX *pctx = EVP_PKEY_CTX_new_id(EVP_PKEY_HKDF, NULL);
    if (!pctx) return -1;

    if (EVP_PKEY_derive_init(pctx) <= 0) goto err;
    if (EVP_PKEY_CTX_set_hkdf_md(pctx, EVP_sha256()) <= 0) goto err;
    if (EVP_PKEY_CTX_set1_hkdf_salt(pctx, salt, saltlen) <= 0) goto err;
    if (EVP_PKEY_CTX_set1_hkdf_key(pctx, ikm, ikmlen) <= 0) goto err;
    if (EVP_PKEY_CTX_add1_hkdf_info(pctx, info, infolen) <= 0) goto err;
    if (EVP_PKEY_derive(pctx, out, (size_t*)&outlen) <= 0) goto err;

    EVP_PKEY_CTX_free(pctx);
    return 0;

[...]

PR #13643, opened 2025-10-11
implement HKDF and HKDFExpand derive_into

reaperhulk:
If we decide we like this we can implement it for every KDF. refs #13245

### DIFF
--- a/src/rust/src/backend/hmac.rs
+++ b/src/rust/src/backend/hmac.rs
@@ -94,7 +94,7 @@ impl Hmac {
         Ok(pyo3::types::PyBytes::new(py, &data))
     }
 
-    fn verify(&mut self, _py: pyo3::Python<'_>, signature: &[u8]) -> CryptographyResult<()> {
+    fn verify(&mut self, signature: &[u8]) -> CryptographyResult<()> {
         let actual = self.finalize_bytes()?;
         if !constant_time::bytes_eq(&actual, signature) {
             return Err(CryptographyError::from(


src/rust/src/backend/kdf.rs
```diff
@@ -526,6 +526,40 @@ struct Hkdf {
...
+        py: pyo3::Python<'_>,
+        key_material: &[u8],
+        output: &mut [u8],
+    ) -> CryptographyResult<usize> {
+        if self.used {
+            return Err(exceptions::already_finalized_error());
+        }
+        self.used = true;
+
+        if output.len() < self.length {
```
alex: Feels like it should just be an error for these to be different -- people wanting to derive into longer buffers should just slice what they pass here.

reaperhulk: Seems fine -- do we want to have `derive_into` return `None` then rather than the bytes written? Or leave it as is?

alex: I don't have a strong instinct. Either seems ok.

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -646,6 +646,27 @@ HKDF
...
+        :param key_material: The input key material.
+        :type key_material: :term:`bytes-like`
+        :param buffer: A writable buffer to write the derived key into.
+        :return int: The number of bytes written to the buffer.
+        :raises TypeError: This exception is raised if ``key_material`` is not
+                           ``bytes``.
+        :raises ValueError: This exception is raised if the buffer is too small
+                           for the derived key.
+        :raises cryptography.exceptions.AlreadyFinalized: This is raised when
+                                                          :meth:`derive_into`
```
alex: Hmm, its really when any combination of the two methods is called more than once.

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -729,6 +750,25 @@ HKDF
...
+
+        :param bytes key_material: The input key material.
+        :param buffer: A writable buffer to write the derived key into.
+        :return int: The number of bytes written to the buffer.
+        :raises TypeError: This exception is raised if ``key_material`` is not
+                           ``bytes``.
+        :raises ValueError: This exception is raised if the buffer is too small
+                           for the derived key.
+        :raises cryptography.exceptions.AlreadyFinalized: This is raised when
+                                                          :meth:`derive_into`
```
alex: same

src/rust/src/backend/kdf.rs
```diff
@@ -526,6 +526,40 @@ struct Hkdf {
...
+        if output.len() != self.length {
+            return Err(CryptographyError::from(
+                pyo3::exceptions::PyValueError::new_err(format!(
+                    "buffer must be {} bytes",
+                    self.length
+                )),
+            ));
+        }
+
+        let prk = self._extract(py, key_material)?;
```
alex: Does it undermind the point at all that we still end up with the prk in some buffer that the caller doesn't control?

reaperhulk: Sort of? It's not ideal. We could require that the buffer provided be >= underlying HMAC output size, but wow that sounds confusing.

reaperhulk: After investigation prk is stack allocated.

alex: To be clear: its stack allocated in OpenSSL/BoringSSL. This implementation is very much not stack allocated.

alex: You could fix that by having `_extract` return some internal type rather than a pyo3 PyByes

reaperhulk: IIRC we left _extract as a python accessible method because while it's not public there was a caller of some form I was aware of. I have now forgotten the details. We could choose to break it since it's not public API surface though.

alex: We do not have any public tests for this. If we wanted this to he public: yikes

But ok you can have an internal method that does the stack thing and then still have the public _extract (but then you'll need a test)

src/rust/src/backend/hmac.rs
```diff
@@ -92,14 +91,12 @@ impl Hmac {
...
         let data = self.finalize_bytes()?;
-        self.ctx = None;
         Ok(pyo3::types::PyBytes::new(py, &data))
     }
 
-    fn verify(&mut self, py: pyo3::Python<'_>, signature: &[u8]) -> CryptographyResult<()> {
-        let actual_bound = self.finalize(py)?;
-        let actual = actual_bound.as_bytes();
-        if !constant_time::bytes_eq(actual, signature) {
+    fn verify(&mut self, _py: pyo3::Python<'_>, signature: &[u8]) -> CryptographyResult<()> {
```
alex: You can drop the py if we're not using it

Issue #13548, opened 2025-10-03
"ExtensionType" has no attribute "get_values_for_type"


mimi89999:
Hello,

I used the example from https://cryptography.io/en/latest/x509/reference/#cryptography.x509.SubjectAlternativeName.get_values_for_type:

```
from cryptography import x509
from cryptography.x509.oid import ExtensionOID

cert = x509.load_pem_x509_certificate(b"")
san_ext = cert.extensions.get_extension_for_oid(ExtensionOID.SUBJECT_ALTERNATIVE_NAME)
san_names = san_ext.value.get_values_for_type(x509.DNSName)

print(san_names)
```

When I run mypy on it, I'm getting the following error:
`error: "ExtensionType" has no attribute "get_values_for_type"  [attr-defined]`

Versions:
```
cryptography==46.0.2
mypy==1.18.2
```

alex:
The issue here is that `get_extension_for_oid` doesn't return a specifically typed extension.

You can either use a type assertion on `san_ext`, or you can use `cert.extensions.get_extension_for_class(x509.SubjectAlternativeName)`.

mimi89999:
Thanks. After using `get_extension_for_class` I have the correct type. Maybe that should be explained in the doc?

reaperhulk:
We'd be happy to take a docs PR that explains this, but glad you were able to resolve your issue!

PR #13653, opened 2025-10-12
Document return type differences in Extensions lookup methods

mimi89999:
I think that it would be useful since I run into https://github.com/pyca/cryptography/issues/13548

### DIFF
--- a/docs/x509/reference.rst
+++ b/docs/x509/reference.rst
@@ -1950,6 +1950,15 @@ X.509 Extensions
 
     .. method:: get_extension_for_oid(oid)
 
+        .. note::
+
+            For type-checking purposes, this method returns ``Extension[ExtensionType]``,
+            which is a generic type. To access methods specific to a particular extension
+            class, you'll need to add a type assertion or a cast. When you know the
+            specific extension type you need, prefer :meth:`get_extension_for_class`
+            which returns a properly typed ``Extension[ExtensionTypeVar]`` and allows
+            direct access to extension-specific methods.
+
         :param oid: An :class:`ObjectIdentifier` instance.
 
         :returns: An instance of :class:`Extension`.


docs/x509/reference.rst
```diff
@@ -1980,6 +1980,18 @@ X.509 Extensions
             >>> cert.extensions.get_extension_for_class(x509.BasicConstraints)
             <Extension(oid=<ObjectIdentifier(oid=2.5.29.19, name=basicConstraints)>, critical=True, value=<BasicConstraints(ca=True, path_length=None)>)>
 
+    .. note::
```
alex: I think it might be clearer to write this specifically as a note on `get_extension_for_oid` -- that for type checking purposes it returns a generic `Extension`, and therefore `get_extension_for_class` is preferred since it'll be properly typed.

mimi89999: Thanks for the feedback. I considered doing that at the beginning, but decided against it. It could be interpreted as if `get_extension_for_class` was generally preferred which I think is not the case. I think that my proposal is more neutral.

alex: I think `get_extension_for_class` _is_ generally preferred where possible.

docs/x509/reference.rst
```diff
@@ -1963,6 +1963,15 @@ X.509 Extensions
             >>> cert.extensions.get_extension_for_oid(ExtensionOID.BASIC_CONSTRAINTS)
             <Extension(oid=<ObjectIdentifier(oid=2.5.29.19, name=basicConstraints)>, critical=True, value=<BasicConstraints(ca=True, path_length=None)>)>
 
+        .. note::
```
reaperhulk: Could you move the note up so it's right after the method decl? Other than that this LGTM now, thanks!

alex:
thank you!

PR #13656, opened 2025-10-13
add some additional typing help

reaperhulk:
this reduces our errors with the current ty alpha to 32 while staying clean on mypy

### DIFF
--- a/tests/hazmat/bindings/test_openssl.py
+++ b/tests/hazmat/bindings/test_openssl.py
@@ -79,7 +79,6 @@ class TestOpenSSL:
         with pytest.raises(InternalError) as exc_info:
             _openssl_assert(False)
 
-        assert isinstance(exc_info.value, InternalError)
         error = exc_info.value.err_code[0]
         assert error.lib == b.lib.ERR_LIB_EVP
         assert error.reason == b.lib.EVP_R_DATA_NOT_MULTIPLE_OF_BLOCK_LENGTH
--- a/tests/x509/test_x509.py
+++ b/tests/x509/test_x509.py
@@ -1490,6 +1490,7 @@ class TestRSACertificate:
                 os.path.join("x509", "custom", "invalid_version.pem"),
                 x509.load_pem_x509_certificate,
             )
+
         assert exc.value.parsed_version == 7
 
     def test_invalid_visiblestring_in_explicit_text(self, backend):


tests/hazmat/bindings/test_openssl.py
```diff
@@ -79,6 +79,7 @@ def test_openssl_assert_error_on_stack(self):
         with pytest.raises(InternalError) as exc_info:
             _openssl_assert(False)
 
+        assert isinstance(exc_info.value, InternalError)
```
alex: hmm, I think this one might be a ty bug -- it should be properly typed AFAICT.

tests/wycheproof/test_ecdh.py
```diff
@@ -64,6 +65,7 @@ def test_ecdh(backend, wycheproof):
         pytest.skip(
             "Unsupported curve ({})".format(wycheproof.testgroup["curve"])
         )
+    assert isinstance(curve, ec.EllipticCurve)
```
alex: this assert shouldn't be needed -- `ty` should be able to know that `pytest.skip` diverges and thus this path has `curve is not None` which makes it `ec.EllipticCurve`

tests/wycheproof/test_rsa.py
```diff
@@ -154,6 +155,8 @@ def test_rsa_pss_signature(backend, wycheproof):
                 wycheproof.testgroup["mgfSha"],
             )
         )
+    assert isinstance(digest, hashes.HashAlgorithm)
+    assert isinstance(mgf_digest, hashes.HashAlgorithm)
```
alex: same

tests/utils.py
```diff
@@ -491,6 +498,8 @@ def load_fips_ecdsa_signing_vectors(vector_data):
     Loads data out of the FIPS ECDSA SigGen vector files.
     """
     vectors = []
+    curve_name = ""
+    digest_name = ""
```
alex: feels like these should be initialized to `None` and then we assert the values aren't `None` by the time we use them

tests/utils.py
```diff
@@ -805,7 +820,7 @@ def load_nist_kbkdf_vectors(vector_data):
     Load NIST SP 800-108 KDF Vectors
     """
     vectors = []
-    test_data = None
+    test_data: typing.Dict[str, typing.Any] = {}
```
alex: I think it'd be better to leave this as is and then assert it's not None at the right place?

tests/hazmat/bindings/test_openssl.py
alex: this is still here.

tests/x509/test_x509.py
```diff
@@ -1490,7 +1490,6 @@ def test_invalid_version_cert(self, backend):
                 os.path.join("x509", "custom", "invalid_version.pem"),
                 x509.load_pem_x509_certificate,
             )
-
```
alex, line 1493: meh

PR #13657, opened 2025-10-13
When serializing EC private keys, ensure the key is padded out to the group degree

alex:
noticed in conversation with @davidben

### DIFF
--- a/src/rust/cryptography-key-parsing/src/ec.rs
+++ b/src/rust/cryptography-key-parsing/src/ec.rs
@@ -130,7 +130,9 @@ pub fn serialize_pkcs1_private_key(
         None
     };
 
-    let private_key_bytes = ec.private_key().to_vec();
+    let private_key_bytes = ec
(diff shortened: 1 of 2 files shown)


alex:
needs: https://github.com/rust-openssl/rust-openssl/pull/2487

davidben:
Optional: This works too, I believe, but I just remembered `EC_KEY_priv2oct` and `EC_KEY_priv2buf` exist, which would also size it correctly for you. Whatever's convenient.

(There was a time I wanted to drop `EC_KEY_get0_private_key` because it, rather inconveniently for constant-time, forces the private key into a `BIGNUM` when we'd actually gotten `BIGNUM` mostly out of that path. But since then, we've figured out how to keep `BIGNUM` from chopping off leading zeros in memory, so `EC_KEY_get0_private_key`'s existence is not so big of a deal.)

reaperhulk (approved):

PR #13658, opened 2025-10-14
add HKDF.extract as a public API and deprecate HKDF._extract

reaperhulk:


alex:
Should this be a method on `HKDF`, or a seperate type like `HKDFExpand` is?

docs/hazmat/primitives/key-derivation-functions.rst
```diff
@@ -630,6 +630,33 @@ HKDF
     :raises TypeError: This exception is raised if ``salt`` or ``info`` is not
                        ``bytes``.
 
+    .. staticmethod:: extract(algorithm, salt, key_material)
+
+        .. versionadded:: 47.0.0
+
+        .. note::
+            Extract is a component of the complete HKDF algorithm.
+            Unless required, users should ignore this method and just
```
alex: "Unless required" is kind of vague, I think you want "Unless needed for implementing an existing protocol" or something.

src/rust/src/backend/hmac.rs
alex: this should call `finalie_bytes`

src/rust/src/backend/kdf.rs
```diff
@@ -520,12 +520,35 @@ impl Argon2id {
 #[pyo3::pyclass(module = "cryptography.hazmat.primitives.kdf.hkdf", name = "HKDF")]
 struct Hkdf {
     algorithm: pyo3::Py<pyo3::PyAny>,
-    salt: pyo3::Py<pyo3::types::PyBytes>,
+    salt: Option<pyo3::Py<pyo3::types::PyBytes>>,
```
alex, line 523: Err, why did this change?

reaperhulk: This changed because for extract we need an optional, so we either store the option and compute the default lazily or we end up having to put the logic in two places.

PR #13665, opened 2025-10-14
disable coverage on macos python 3.8

reaperhulk:
maybe it's faster, maybe it's not


reaperhulk:
We don't cover 3.8 anywhere else and it's not worth adding a linux job for this.

PR #13691, opened 2025-10-17
support encrypt_into on AesSiv

reaperhulk:
this is the beginning of a refactor that allows encrypt_into/decrypt_into on all AEADs.

### DIFF
--- a/src/rust/src/backend/aead.rs
+++ b/src/rust/src/backend/aead.rs
@@ -156,6 +156,28 @@ impl EvpCipherAead {
         )?)
     }
 
+    fn encrypt_into(
+        &self,
+        plaintext: &[u8],
+        aad: Option<Aad<'_>>,
+        nonce: Option<&[u8]>,
+        buf: &mut [u8],
+    ) -> CryptographyResult<()> {
+        let mut ctx = openssl::cipher_ctx::CipherCtx::new()?;
+        ctx.copy(&self.base_encryption_ctx)?;
+
+        Self::encrypt_with_context(
+            ctx,
+            plaintext,
+            aad,
+            nonce,
+            self.tag_len,
+            self.tag_first,
+            false,
+            buf,
+        )
+    }
+
     #[allow(clippy::too_many_arguments)]
     fn encrypt_with_context(
         mut ctx: openssl::cipher_ctx::CipherCtx,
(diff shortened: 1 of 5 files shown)


alex:
this can be rebased now

src/rust/src/backend/aead.rs
```diff
@@ -917,7 +939,50 @@ impl AesSiv {
...
-        self.ctx.encrypt(py, data_bytes, aad, None)
+        Ok(pyo3::types::PyBytes::new_with(
+            py,
+            data_bytes.len() + self.ctx.tag_len,
+            |b| {
+                self.ctx.encrypt_into(data_bytes, aad, None, b)?;
+                Ok(())
+            },
+        )?)
+    }
```
alex, line 950: Hmm, why did this need to change?

src/rust/src/backend/aead.rs
```diff
@@ -917,7 +939,50 @@ impl AesSiv {
...
+        if data_bytes.is_empty() {
+            return Err(CryptographyError::from(
+                pyo3::exceptions::PyValueError::new_err("data must not be zero length"),
+            ));
+        };
+
+        // Check this early so we know we can add tag_len without overflow
+        // check_length requires that the length be 2 ** 31 - 1 or smaller.
+        check_length(data_bytes)?;
+        let expected_len = data_bytes.len() + self.ctx.tag_len;
```
alex, line 972: Can this overflow?

src/rust/src/backend/aead.rs
reaperhulk: Eventually the idea is to remove the encrypt method from ctx entirely so this just starts it

src/rust/src/backend/aead.rs
reaperhulk: Read the comment 😄

alex: 32-bit ;-)

reaperhulk: But since check length validates a max signed int value there’s still 2 billion left before overflow on a 32 bit usize, right?

alex: Sigh, yeah. Sorry.

PR #13735, opened 2025-10-24
asn1: Add support for `IMPLICIT` and `EXPLICIT`

facutuesca:
This PR adds support for `IMPLICIT` and `EXPLICIT` fields to the ASN.1 API. The field types must be annotated with `asn1.Implicit(tag=X)` or `asn1.Explicit(tag=Y)`:

```python
@asn1.sequence
class Example
  a: Annotated[typing.Union[int, None], asn1.Implicit(0)]
  b: Annotated[typing.Union[int, None], asn1.Implicit(1)]
```

### DIFF
--- a/src/rust/src/declarative_asn1/decode.rs
+++ b/src/rust/src/declarative_asn1/decode.rs
@@ -155,8 +155,9 @@ pub(crate) fn decode_annotated_type<'a>(
             let inner_tag = type_to_tag(cls.get().inner.get(), encoding);
             match parser.peek_tag() {
                 Some(t) if t == inner_tag => {
-                    // Since for optional types the annotations are enforced to be associated with the Option
-                    // (instead of the inner type), when decoding the inner type we add the annotations of the Option
+                    // For optional types, annotations will always be associated to the `Optional` type
+                    // i.e: `Annotated[Optional[T], annotation]`, as opposed to the inner `T` type.
+                    // Therefore, when decoding the inner type `T` we must pass the annotation of the `Optional`
                     let inner_ann_type = AnnotatedType {
                         inner: cls.get().inner.clone_ref(py),
                         annotation: ann_type.annotation.clone_ref(py),


src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -60,13 +60,18 @@ def _is_union(field_type: type) -> bool:
...
 def _extract_annotation(metadata: tuple) -> declarative_asn1.Annotation:
     default = None
+    encoding = None
     for raw_annotation in metadata:
         if isinstance(raw_annotation, Default):
             default = raw_annotation.value
+        elif isinstance(raw_annotation, Explicit):
+            encoding = declarative_asn1.Encoding.Explicit(raw_annotation.tag)
+        elif isinstance(raw_annotation, Implicit):
+            encoding = declarative_asn1.Encoding.Implicit(raw_annotation.tag)
```
alex: You need to prohibit duplicates (I guess this is true of defaults as well!)

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -194,6 +199,16 @@ class Default(typing.Generic[U]):
     value: U
 
 
+@dataclasses.dataclass(frozen=True)
```
alex: Do we need these, or could we expose `Encoding.Explicit` and `Encoding.Implicit` as these types?

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -130,19 +152,27 @@ pub(crate) fn decode_annotated_type<'a>(
             })?
         }
         Type::Option(cls) => {
-            let inner_tag = type_to_tag(cls.get().inner.get());
+            let inner_tag = type_to_tag(cls.get().inner.get(), encoding);
             match parser.peek_tag() {
-                Some(t) if t == inner_tag => decode_annotated_type(py, parser, cls.get())?,
+                Some(t) if t == inner_tag => {
+                    // Since for optional types the annotations are enforced to be associated with the Option
+                    // (instead of the inner type), when decoding the inner type we add the annotations of the Option
```
alex: I'm not sure I follow.

src/rust/src/declarative_asn1/types.rs
```diff
@@ -266,17 +282,25 @@ pub(crate) fn python_class_to_annotated<'p>(
...
         Type::PyStr() => asn1::Utf8String::TAG,
         Type::PrintableString() => asn1::PrintableString::TAG,
         Type::UtcTime() => asn1::UtcTime::TAG,
         Type::GeneralizedTime() => asn1::GeneralizedTime::TAG,
+    };
+
+    match encoding {
+        Some(e) => match e.get() {
+            Encoding::Implicit(n) => asn1::implicit_tag(*n, inner_tag),
+            Encoding::Explicit(n) => asn1::explicit_tag(*n),
```
alex, line 301: These are currenyly `#[doc(hidden)]` and thus not part of the stable API. We need to either change that, or represent these a different way.

reaperhulk:
@facutuesca Just a friendly ping since we're probably only a few weeks away from 47.0.0. If we need to slip to 48 that's fine of course.

src/rust/src/declarative_asn1/decode.rs
facutuesca: I modified the comment to try to make it clearer:
```rust
// For optional types, annotations will always be associated to the `Optional` type
// i.e: `Annotated[Optional[T], annotation]`, as opposed to the inner `T` type.
// Therefore, when decoding the inner type `T` we must pass the annotation of the `Optional`
```

src/cryptography/hazmat/asn1/asn1.py
facutuesca: fixed!

src/cryptography/hazmat/asn1/asn1.py
facutuesca: fixed!

src/rust/src/declarative_asn1/types.rs
facutuesca: I'll create a PR in `rust-asn1` to make them part of the stable API, unless there's a better solution

alex: I wonder if instead of doing this we should be using `can_parse` on the `Asn1Readable` trait?

facutuesca: That would work for the simple types (`int`, `bool`, etc). But for things like `Option`, where at the moment of parsing we only know that `Option[T]` has a `T: AnnotatedType`, we would need to define `Asn1Readable` for `AnnotatedType`.

And for that, we would need to define:
```rust
impl<'a> asn1::Asn1Readable<'a> for AnnotatedType {
    fn parse(parser: &mut Parser<'a>) -> ParseResult<Self> { }

    .....
```

But the structure of the type we're parsing is dynamic, and that structure is stored inside the `AnnotatedType` object, so `parse` would need to take a `self` parameter in order to know how to parse the type.

This was the original reason why we didn't implement `Asn1Readable` for `AnnotatedType`. We didn't have the same issue for `Asn1Writable`, because `Asn1Writable::write()` takes a `self` parameter so we can decide how to encode an object at runtime, depending on the contents of `self`

alex: Hmm, so if you push `Option` handling into `read_value` (e.g., it takes an `is_optional` argument or something), it seems like it should be possible. Requires a non-trivial refactor and I'd suggest splitting that out if you think it makes sense.

facutuesca: A couple of notes:
- It'd also mean passing an `is_optional` argument to `decode_annotated_type()`, since when parsing `Optional` we call it recursively
- To use `Asn1Readable::can_parse(t: Tag)`, wouldn't we need to make public the [`Tag::new`](https://docs.rs/asn1/latest/src/asn1/tag.rs.html#75) API?

alex: Hmm, I think if we push everything into `read_value`, we no longer need to manually check tags at all?

facutuesca: if we have a `field [0] EXPLICIT INTEGER DEFAULT 3`.

How do we determine if `field` is not present in order to return the default value?

alex: `p.read_element::<Option<Explicit<T>>()` and then if it's None, use the default

facutuesca: > `p.read_element::<Option<Explicit<T>>()` and then if it's None, use the default

Doesn't `Explicit` require specifying the tag during compile time? i.e: `Option<Explicit<int, 0>>`

alex: Oh. Right. Ughhhhhhh.

Issue #10226, opened 2024-01-22
Add support for RSA sign/verify with raw PKCS #1 data without DigestInfo


misterzed88:
There are non-standard RSA signature formats where the PKCS <span>#</span>1-padded data is without DigestInfo. A few examples:
- Some old Thawte and Verisign timestamp certificates.

- Signed MD5/SHA1 hashes in TLS 1.1 or earlier ([RFC 4346](https://datatracker.ietf.org/doc/html/rfc4346.html), section 4.7).

- IKE version 1 signatures ([RFC 2409](https://datatracker.ietf.org/doc/html/rfc2409.html), section 5.1).

By making it possible to support these formats the Cryptography library would be even more versatile and useful, broadening its user base.

The `RSAPublicKey.recover_data_from_signature` already supports these type of non-standard formats. (See issue #5495).

A straightfoward solution would be to extend the `RSAPrivateKey.sign` and `RSAPublicKey.verify` APIs to allow `None` in the in the `algorithm` parameter.  This would also make these APIs symmetrical with `RSAPublicKey.recover_data_from_signature`, which accepts `None` for the exact same purpose.

reaperhulk:
Are there any test vectors we can use to confirm functionality here?

misterzed88:
The best I have found for now are NIST test vectors for RSA signatures, part of their Cryptographic Algorithm Validation Program ([CAVP](https://csrc.nist.gov/projects/cryptographic-algorithm-validation-program/digital-signatures#rsa2vs)). The RSA test vector files are linked from the "Test Vectors" section.

The SigGen15_186-*.txt files contain the private exponent which makes it possible to create signatures in addition to verifying them.

Unfortunately the vector files do not contain data for the intermediary steps. So the test code has perform the low-level message encoding according to section 9.2 in RFC 3447. (It is quite straightforward).

I have also tried to find public test data for RSA-signatures of 36-byte TLS 1.1 MD5+SHA1 hash values, but so far without any luck.

misterzed88:
After spending considerable effort, I could not find any nice, public test vectors for the SSL/TLS 1.1 signature case (RSA-signed 36-bytes MD5+SHA1 hashes).

As an alternative, I could create a test vector for this case, by running a TLS handshake trace with OpenSSL, documenting the steps so it can be reproduced by others.

Would you have any use of a test vector created this way? Is there anything else I can help with to advance this feature request?

reaperhulk:
[...]

Issue #5495, opened 2020-10-24
Add support for non-standard RSA signature encodings


misterzed88:
There are use cases when there is a need to encode and decode RSA signature data using odd formats inside the PKCS #1-padding, such as:
- Encoding the raw digest without a DigestInfo structure
- Using a proprietary or invalid AlgorithmIdentifier in the DigestInfo structure

One example is described in https://vcsjones.dev/2019/07/18/sometimes-valid-rsa-dotnet/

In order to support these cases, a proposal for API extension is described below.

Extension in the RSAPrivateKey.sign function:
- The algorithm parameter is allowed to be None, meaning that data is independent of hash algorithm.
- Only supported for some of the paddings (currently only PKCS1v15).
- The data is encoded as the raw signature data (without DigestInfo).

Extension in the RSAPublicKey.verify function:
- Allow None in the algorithm parameter with the same semantics as described for the sign function above.

Add a new function RSAPublicKey.verify_recover:
- Used to recover the digest (or raw signature data).
- Only supported for some of the paddings (currently only PKCS1v15).
- Returns the recovered data as bytes.
- Uses the same parameters as the verify function but removing the data parameter.
- The padding parameter is kept for future extensibility, even though only PKCS1v15 works for now.
- The algorithm parameter can None to return raw signature data (the same semantics as described above).
- When algorithm is a HashAlg, the signature is checked to contain that alg in DigestInfo and only the digest is returned.
- The Prehashed value may be kept as valid input for orthogonality, in which case it is only used as an alias for HashAlg.

From an API user point of view, I think the proposed changes will be simple to understand and to use, without breaking compatibility and introducing too much complexity. This is the best I could come up with for now. It is not straightforward to find a solution which keeps the current nice, simple API while adding the functionality needed. But please feel free to comment or improve the proposal. (For example, should we replace the None algorithm value with a more specific value similar to Prehashed, such as RawData?)

Note that this proposal only goes so far: it does not help if an application uses a a non-standard RSA padding, for example.

Implementation notes:
- OpenSSL handles most of the complexity and error handling of border cases.
- The None algorithm value can simply be mapped to skipping the call to EVP_PKEY_CTX_set_signature_md, in which case OpenSSL uses data without DigestInfo (both during signature creation and during verification).
- OpenSSL returns an error when using a padding (such as PSS) which requires the hash algorithm to be set.
- The None value also works as expected with some of the other algorithms, such as DSA.
[...]

PR #13740, opened 2025-10-24
Allow RSA signing with raw data (without a DigestInfo)

### LINKED ISSUE #3713: Allow signing to exclude hash constant
Hi lovely cryptography folks. I just added the ability to [sign server descriptors](https://gitweb.torproject.org/stem.git/commit/?id=9a947c6) similar to what the tor process does but ran into one rough bit. In particular when cryptography makes an RSA signature it embeds a constant indicating the hash it uses whereas tor does not. This is buried pretty deep within cryptography's innards so I had to **[do an icky hack to be compatible with tor](https://gitweb.torproject.org/stem.git/commit/?id=a41767c)**.

No doubt cryptography is doing the perfectly right thing here, but would you mind adding an option either to RSAPrivateKey's sign() method or the MultiBackend to omit this so we can drop the hack? In particular it's just two lines in the **_rsa_sig_sign() method of cryptography/hazmat/backends/openssl/rsa.py** we'd like to avoid...

```
197     res = backend._lib.EVP_PKEY_CTX_set_signature_md(
198         pkey_ctx, evp_md)
199     backend.openssl_assert(res > 0)
```
... to...

```
if include_hash_constant:
  res = backend._lib.EVP_PKEY_CTX_set_signature_md(pkey_ctx, evp_md)
  backend.openssl_assert(res > 0)
```

Cheers! -Damian

jahkosha:
This fixes #3713 and #10226.

Instead of using the script written by @misterzed88 in #10226 to generate modified tests vectors, I did directly implement the same logic in the test infrastructure, so we can reuse directly the NIST (or others...) vectors stored in the tests.

### DIFF
--- a/tests/hazmat/primitives/test_rsa.py
+++ b/tests/hazmat/primitives/test_rsa.py
@@ -562,6 +562,52 @@ class TestRSASignature:
     @pytest.mark.supported(
         only_if=lambda backend: backend.rsa_padding_supported(
             padding.PSS(
+                mgf=padding.MGF1(hashes.SHA1()),
+                salt_length=padding.PSS.MAX_LENGTH,
+            )
+        ),
+        skip_message="Does not support PSS.",
+    )
+    @pytest.mark.supported(
+        only_if=lambda backend: backend.signature_hash_supported(
+            hashes.SHA1()
+        ),
+        skip_message="Does not support SHA1 signature.",
+    )
+    def test_pss_signing_without_digest(self, backend, subtests):
+        for private, public, example in _flatten_pkcs1_examples(
+            load_vectors_from_file(
+                os.path.join(
+                    "asymmetric", "RSA", "pkcs-1v2-1d2-vec", "pss-vect.txt"
+                ),
+                load_pkcs1_vectors,
+            )
+        ):
+            with subtests.test():
+                private_key = rsa.RSAPrivateNumbers(
+                    p=private["p"],
+                    q=private["q"],
+                    d=private["private_exponent"],
+                    dmp1=private["dmp1"],
+                    dmq1=private["dmq1"],
+                    iqmp=private["iqmp"],
+                    public_numbers=rsa.RSAPublicNumbers(
+                        e=private["public_exponent"], n=private["modulus"]
+                    ),
+                ).private_key(backend, unsafe_skip_rsa_key_validation=True)
+                with pytest.raises(TypeError):
+                    private_key.sign(
+                        binascii.unhexlify(example["message"]),
+                        padding.PSS(
+                            mgf=padding.MGF1(algorithm=hashes.SHA1()),
+                            salt_length=padding.PSS.MAX_LENGTH,
+                        ),
+                        asym_utils.NoDigestInfo(),
+                    )
+
+    @pytest.mark.supported(
+        only_if=lambda backend: backend.rsa_padding_supported(
+            padding.PSS(
                 mgf=padding.MGF1(hashes.SHA256()),
                 salt_length=padding.PSS.MAX_LENGTH,
             )


alex:
I haven't reviewed in depth -- but I don't think we should use `None` to signify this, I think we should have a dedicated sentinel, `Raw<Something>` perhaps. `None` looks like a reasonable default, but this is the opposite of that :-)

@reaperhulk wdyt?

jahkosha:
@alex yeah that is a fair concern :+1:  in some early attempt I did add a `NoHash` hashes instead of using `None`, but I was afraid that would cause even more confusion so eventually I went with `None`.

I'll be happy to apply what you folks think is the best, if possible, please provide pointers where in the code-base a similar pattern is used so I can take inspiration from it.

misterzed88:
> I haven't reviewed in depth -- but I don't think we should use `None` to signify this, I think we should have a dedicated sentinel, `Raw<Something>` perhaps. `None` looks like a reasonable default, but this is the opposite of that :-)
> 
> @reaperhulk wdyt?

I would like to highlight that the API already uses None for this purpose (for the RSA signature recover functionality, ref. issue #5495). So whatever you decide, you may want to use the same method in all the API functions to make them symmetric.

reaperhulk:
Hmm, the inconsistency is a bit unfortunate. I'd be inclined to do a `RawNoDigestInfo` (or perhaps just `NoDigestInfo`?) sentinel here and then also add that as an option on the signature recovery method while retaining `None` for backwards compatibility. So this signature would look like `sign(value, PKCS1v15(), RawNoDigestInfo())` or `sign(value, PKCS1v15(), NoDigestInfo())`.

jahkosha:
@alex @reaperhulk I did add a sentinel `NoDigestInfo` and support it also on `recover_data_from_signature`

Please have an other look and let me know what you think

src/rust/src/backend/rsa.rs
```diff
@@ -481,6 +497,11 @@ impl RsaPublicKey {
         padding: &pyo3::Bound<'_, pyo3::PyAny>,
         algorithm: &pyo3::Bound<'_, pyo3::PyAny>,
     ) -> CryptographyResult<pyo3::Bound<'p, pyo3::types::PyBytes>> {
+        let algorithm = if algorithm.is_instance(&types::NO_DIGEST_INFO.get(py)?)? {
+            &pyo3::types::PyNone::get(py).to_owned().into_any()
```
reaperhulk, line 508: This branch isn't covered so CI is failing, but should be an easy test to add.

jahkosha: @reaperhulk it's done! I did squash the fix in the latest commit (see `test_pkcs1v15_verification`)

Thanks for the guidance, I was confused what was the issue.

src/cryptography/hazmat/primitives/asymmetric/utils.py
```diff
@@ -11,6 +11,10 @@
 encode_dss_signature = asn1.encode_dss_signature
 
 
+class NoDigestInfo:
```
reaperhulk, line 14: We need to document this in the same section as `Prehashed`

tests/hazmat/primitives/test_rsa.py
```diff
@@ -442,6 +444,49 @@ def test_pkcs1v15_signing(self, backend, subtests):
...
+                    dmq1=private["dmq1"],
+                    iqmp=private["iqmp"],
+                    public_numbers=rsa.RSAPublicNumbers(
+                        e=private["public_exponent"], n=private["modulus"]
+                    ),
+                ).private_key(backend, unsafe_skip_rsa_key_validation=True)
+                signature = private_key.sign(
+                    binascii.unhexlify(
+                        compute_rsa_hash_digest(
+                            backend, hashes.SHA1(), example["message"]
```
reaperhulk: I'd prefer we used SHA256 for the tests here because otherwise we'll increasingly have coverage challenges as more and more things disable SHA1.

tests/hazmat/primitives/test_rsa.py
```diff
@@ -910,7 +955,7 @@ def test_pkcs1v15_verification(self, backend, subtests):
                 # Test recovery of all data (full DigestInfo) with hash alg. as
                 # None
                 rec_sig_data = public_key.recover_data_from_signature(
```
reaperhulk, line 1014: We need to document that `recover_data_from_signature` now takes this new argument.

tests/hazmat/primitives/test_rsa.py
```diff
@@ -1522,6 +1567,26 @@ class TestRSAPKCS1Verification:
         )
     )
 
+    test_rsa_pkcs1v15_verify_sha1_without_digest = pytest.mark.supported(
+        only_if=lambda backend: (
+            backend.signature_hash_supported(hashes.SHA1())
+            and backend.rsa_padding_supported(padding.PKCS1v15())
+        ),
+        skip_message="Does not support SHA1 and PKCS1v1.5.",
```
reaperhulk: Same SHA1 concern

reaperhulk (commented):
We also need docs for this new value in `RSAPrivateKey.sign`

src/rust/src/backend/rsa.rs
```diff
@@ -481,6 +497,11 @@ impl RsaPublicKey {
         padding: &pyo3::Bound<'_, pyo3::PyAny>,
         algorithm: &pyo3::Bound<'_, pyo3::PyAny>,
     ) -> CryptographyResult<pyo3::Bound<'p, pyo3::types::PyBytes>> {
+        let algorithm = if algorithm.is_instance(&types::NO_DIGEST_INFO.get(py)?)? {
```
Tarnum-tst, line 507:
```
error[E0716]: temporary value dropped while borrowed
   --> src\backend\rsa.rs:501:14
    |
500 |         let algorithm = if algorithm.is_instance(&types::NO_DIGEST_INFO.get(py)?)? {
    |             --------- borrow later stored here
501 |             &pyo3::types::PyNone::get(py).to_owned().into_any()
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creates a temporary value which is freed while still in use
502 |         } else {
    |         - temporary value is freed at the end of this statement
    |
    = note: consider using a `let` binding to create a longer lived value

For more information about this error, try `rustc --explain E0716`.
error: could not compile `cryptography-rust` (lib) due to previous error
```

alex: What is this from? CI shows it builds fine.

Tarnum-tst: @alex 
Python 3.8.10 32-bit, **i686-pc-windows-gnu**, **Rust 1.74** (because I'm still on Windows 7 SP1). The only compatibility change from main branch I'm made so far - used const unbind fn's in file ec_constants.rs for commit ac63eab2a12e357671139ef809e6a17dbf680c8f

alex: We don't support Windows 7 and our MSRV is 1.83.

Please do not bother contributors with complaints about not supporting unsupported platforms (particularly not with comments that provide _absolutely no context_)

tests/hazmat/primitives/test_rsa.py
jahkosha: The problem with that change is that the test vectors we use here (pkcs1v15sign-vectors.txt) do use SHA1.

tests/hazmat/primitives/test_rsa.py
jahkosha: Fixed in https://github.com/jahkosha/cryptography/commit/7147c8658bb7560a48897a6a4f0844b939c33c55

tests/hazmat/primitives/test_rsa.py
jahkosha: Fixed in https://github.com/jahkosha/cryptography/commit/5cc4ab6da28094736ba044e97cefa3a95eaf7a71

src/cryptography/hazmat/primitives/asymmetric/utils.py
jahkosha: Fixed in https://github.com/jahkosha/cryptography/commit/5cc4ab6da28094736ba044e97cefa3a95eaf7a71

jahkosha:
Hey @reaperhulk, I finally had some time to have an other pass on this.

I could apply your feedback, there is just one comment I could not fix (SHA1 vs SHA256) because SHA1 is used by the test vectors themself (https://github.com/pyca/cryptography/pull/13740#discussion_r2495197471).

Please have a look at the last two commits and let me know if some more polish is needed.

Many thanks!

docs/hazmat/primitives/asymmetric/utils.rst
```diff
@@ -29,6 +29,19 @@ Asymmetric Utilities
 
     :return bytes: The encoded signature.
 
+.. class:: NoDigestInfo()
+
+    .. versionadded:: 47.0
```
reaperhulk:
```suggestion
    .. versionadded:: 47.0.0
```

tests/hazmat/primitives/test_rsa.py
reaperhulk: We don't need to use `pkcs1v15sign-vectors.txt` here, we can use `SigVer15_186-3.rsp` and filter to passing non-SHA1 tests.

reaperhulk:
Oh, forgot one thing: we need a changelog entry in `CHANGELOG.rst` for this. Something like:
> Added support for PKCS1v15 signing without `DigestInfo` using :class:`~cryptography.path.to.NoDigestInfo`.

jahkosha:
@reaperhulk I gave a try at using `SigVer15_186-3.rsp` instead of `pkcs1v15sign-vectors.txt`, but the fields does not match between `load_pkcs1_vectors` and `load_rsa_nist_vectors`.

I could add the code to parse `signature` (`S`) and message (`Msg`), but I don't see in `SigVer15_186-3.rsp`  some of the fields used to create the private key, namely: dmp1, dmq1 and iqmp.

Also I'm not sure how to differentiate private/public exponent... could you please give some guidance?

For reference, here is the field used by the test:
```
                private_key = rsa.RSAPrivateNumbers(
                    p=private["p"],
                    q=private["q"],
                    d=private["private_exponent"],
                    dmp1=private["dmp1"],
                    dmq1=private["dmq1"],
                    iqmp=private["iqmp"],
                    public_numbers=rsa.RSAPublicNumbers(
                        e=private["public_exponent"], n=private["modulus"]
                    ),
```

I looked in the rest of the codebase but I could not find any test that use `SigVer15_186-3.rsp` to build private keys.

misterzed88:
> I could add the code to parse `signature` (`S`) and message (`Msg`), but I don't see in `SigVer15_186-3.rsp` some of the fields used to create the private key, namely: dmp1, dmq1 and iqmp.

In a straightforward RSA implementation it would have been sufficient with the modulus `n` and the private exponent `d` for computations with the private key (as illustrated in the test script `rsavec.py`). There is also a more efficient computation, called the Chinese Reminder Theorem (CRT), which requires the parameters `p`, `q`,  `dmp1`, `dmq1` and `iqmp`.

The cryptography library always requires the CRT parameters to be available. Since they are not present in  `SigVer15_186-3.rsp` they must be computed from `n`, `e` and `d` as follows:
```
p, q = rsa_recover_prime_factors(n, e, d)
dmp1 = rsa_crt_dmp1(d, p)
dmq1 = rsa_crt_dmq1(d, q)
iqmp = rsa_crt_iqmp(p, q)
```

It would of course be nice if the lib supported private RSA with only `d` being available, using it directly or with optional CRT precomputation support (for better efficiency in case the key is used multiple times). But that is a diffent story.

Perhaps the maintainers have more context or a better answer.

jahkosha:
@misterzed88 thanks for your kind explanation, I was able to write code to recompute those values, and that was an opportunity to understand something new, much appreciated.

@reaperhulk sadly, even with the help from @misterzed88 I am not able to make it work, there is two things that I find particularly odd and concerning:

1.  When I use `rsa_recover_prime_factors` (which fundamentally I don't need, it is there just for debugging) the value that I recover does not match the one defined in the test vectors, please have a look at [those lines](https://github.com/pyca/cryptography/pull/13740/commits/2eaf3d9cedb747c025af429ffa116f2c5eaa01f0#diff-eaf812a71facb603d3a294e14cc4997e69cc207e594731f0b9f4fab10c92663bR468-R475).

2. When I try to do generate the signature in the normal way (without the NoDigestInfo logic - again just for debbuging), the signature does not match. As [you can see here](https://github.com/pyca/cryptography/pull/13740/commits/2eaf3d9cedb747c025af429ffa116f2c5eaa01f0#diff-eaf812a71facb603d3a294e14cc4997e69cc207e594731f0b9f4fab10c92663bR506).

This prevents me from writing the required tests, do you have any ideas what I'm doing wrong here? I'm just confused after doing those other tests... at this point, I'm not sure where could be the issue.

(I did fix the other points though)

docs/hazmat/primitives/asymmetric/utils.rst
jahkosha: Fixed and squashed in https://github.com/pyca/cryptography/pull/13740/commits/405302f56330af6645a29e7e3dbc027cae87f786

tests/hazmat/primitives/test_rsa.py
```diff
@@ -442,6 +444,73 @@ def test_pkcs1v15_signing(self, backend, subtests):
...
+    def test_pkcs1v15_signing_without_digest(self, backend, subtests):
+        vectors = load_vectors_from_file(
+            os.path.join(
+                "asymmetric", "RSA", "FIPS_186-2", "SigVer15_186-3.rsp"
+            ),
+            load_rsa_nist_vectors,
+        )
+        for params in vectors:
+            with subtests.test():
+                # TODO Why those not match?
```
reaperhulk: This is failing because p and q are nominally a set (e.g., unordered) primes, but in practice many implementations require `p > q` so we changed how this recovery function returns (see the note in https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.rsa.rsa_recover_prime_factors). Since you're asserting that the recovered `p == params["p"]` that fails when that value is contained in `params["q"]`

tests/hazmat/primitives/test_rsa.py
```diff
@@ -442,6 +444,73 @@ def test_pkcs1v15_signing(self, backend, subtests):
...
+                #     binascii.unhexlify(
+                #         compute_rsa_hash_digest(
+                #             backend, hashes.SHA256(), params["msg"]
+                #         )
+                #     ),
+                #     padding.PKCS1v15(),
+                #     asym_utils.NoDigestInfo(),
+                # )
+
+                # TODO Why even this does not works?
```
reaperhulk: This is failing because you assumed the vector file only contained vectors for SHA256. If you look at it you'll see there's an entry called SHAAlg which says which hash function was used for each vector. For this you could just do `if params["SHAAlg"] != "SHA256": continue`. You can combine this with the `continue` for fails.

tests/hazmat/primitives/test_rsa.py
```diff
@@ -442,6 +444,73 @@ def test_pkcs1v15_signing(self, backend, subtests):
...
+    @pytest.mark.supported(
+        only_if=lambda backend: backend.signature_hash_supported(
+            hashes.SHA256()
+        ),
+        skip_message="Does not support SHA256 signature.",
+    )
+    def test_pkcs1v15_signing_without_digest(self, backend, subtests):
+        vectors = load_vectors_from_file(
+            os.path.join(
+                "asymmetric", "RSA", "FIPS_186-2", "SigVer15_186-3.rsp"
```
reaperhulk, line 462: This vector file contains failed vectors too, which we don't care about and should skip for this. You can skip with `if params["fail"]: continue`

tests/hazmat/primitives/test_rsa.py
jahkosha: I see! I was conflating the filter for the backend with the filtering of the the tests vectors.

jahkosha:
@reaperhulk thanks for your guidance and patience!

I was able to address the tests related feedaback: https://github.com/pyca/cryptography/pull/13740/commits/e47bf817919197341d183d49fdedf1eacea7961f

I did also update the CHANGELOG as requested and add the proper version number in the documentation.

Ready for an other pass when you have time, thank you :)

CHANGELOG.rst
```diff
@@ -73,6 +73,8 @@ Changelog
   :class:`~cryptography.hazmat.primitives.ciphers.aead.AESSIV`, and
   :class:`~cryptography.hazmat.primitives.ciphers.aead.ChaCha20Poly1305` to
   allow encrypting directly into a pre-allocated buffer.
+* Added support for PKCS1v15 signing without DigestInfo using
+  :class:~cryptography.path.to.NoDigestInfo.
```
reaperhulk: This was just intended as an example. You need to define the actual path to NoDigestInfo here. I believe it's in cryptography.hazmat.primitives.asymmetric.utils

tests/hazmat/primitives/test_rsa.py
```diff
@@ -442,6 +444,59 @@ def test_pkcs1v15_signing(self, backend, subtests):
...
+                    dmp1=dmp1,
+                    dmq1=dmq1,
+                    iqmp=iqmp,
+                    public_numbers=rsa.RSAPublicNumbers(
+                        e=params["public_exponent"], n=params["modulus"]
+                    ),
+                ).private_key(backend, unsafe_skip_rsa_key_validation=True)
+                signature = private_key.sign(
+                    binascii.unhexlify(
+                        compute_rsa_hash_digest(
```
reaperhulk: This is the only call site for this so I don't think we need this to be a function.

tests/hazmat/primitives/utils.py
```diff
@@ -47,6 +56,13 @@ def _load_all_params(path, file_names, param_loader):
     return all_params
 
 
+def compute_rsa_hash_digest(backend, hash_alg, msg):
+    oid = _hash_alg_oids[hash_alg.name]
```
reaperhulk: This isn't necessary if we're only doing SHA256.

reaperhulk:
Almost there!

tests/hazmat/primitives/test_rsa.py
jahkosha: It is **not** the only call site, it is used for other tests thru: https://github.com/pyca/cryptography/pull/13740/files#diff-adca63e2e988b9f614480a194688af9467b6b54710a658d4aa958e48b2bc18c2R526

tests/hazmat/primitives/utils.py
jahkosha: Fixed in https://github.com/pyca/cryptography/pull/13740/commits/1f359948ff5915e2a325dc11300649845c3a3f17

CHANGELOG.rst
jahkosha: Oops, I was so focused an having the other test working that I did not even read carefully!

Anyway, this is fixed in -> https://github.com/pyca/cryptography/pull/13740/commits/5ee18d4c767a1c20a882fba95c748e7da5226662

This time I did copy/pasted a line for `Prehashed` so it must be right.

jahkosha:
@reaperhulk done applying your last feedback!

The only thing that I did not do is inlining `compute_rsa_hash_digest_sha256` because as I said it is used in two different places.

Please have an other look :-)

reaperhulk:
Tests are failing only on the FIPS builders, almost certainly due to key size restrictions. You can either skip this test on FIPS or try adding a check for whether fips is enabled and skip if the modulus is < 2048-bit and see if that fixes it. You can see examples of both of these ideas in test_rsa.py

jahkosha:
@reaperhulk you were right! 

It is now all green (and freshly rebased), let me know if anything else remain to be fixed?

thanks a lot

(please see the last commit, I just reused the same check used in other tests)

src/rust/src/backend/rsa.rs
```diff
@@ -290,8 +290,16 @@ impl RsaPrivateKey {
         padding: &pyo3::Bound<'p, pyo3::PyAny>,
         algorithm: &pyo3::Bound<'p, pyo3::PyAny>,
     ) -> CryptographyResult<pyo3::Bound<'p, pyo3::types::PyAny>> {
-        let (data, algorithm) =
-            utils::calculate_digest_and_algorithm(py, data.as_bytes(), algorithm)?;
+        let (data, algorithm) = {
```
reaperhulk, line 293: What happens if I pass NoDigestInfo and also PSS instead of PKCS1v15 for padding? Do we properly handle that case (it should be a ValueError that NoDigestInfo is only allowed for PKCS1v15)? I don't see logic for it. Assuming it's missing, we'll need a test for that once the check is added..

tests/hazmat/primitives/utils.py
```diff
@@ -37,6 +38,14 @@
 
 from ...utils import load_vectors_from_file
 
+_hash_alg_oids = {
```
reaperhulk: This came back again, mistaken rebase?

src/rust/src/backend/rsa.rs
jahkosha: Currently If you pass `NoDigestInfo` and also `PSS` you get a `TypeError: Expected instance of hashes.HashAlgorithm.`.

I think this is even better than a `ValueError`but your mileage might vary.

I added a test case for this code path in https://github.com/pyca/cryptography/pull/13740/commits/712d5e28f7cd05c367c2c97c4d054febd45407af

tests/hazmat/primitives/utils.py
jahkosha: Yes! Fixed by squashing the removal in https://github.com/pyca/cryptography/pull/13740/commits/6a8b12974acca7f3f73c8006983ce7c8dc3bf92c

jahkosha:
@reaperhulk ready for an other round of review, thanks!

reaperhulk:
Thanks for your work here!

misterzed88:
Wow, finally done! Big thanks @jahkosha for your persistence and positivity!
Also thanks to Paul and Alex for all your efforts with maintaining this great library! (And for the swift guidance to external contributors).

jahkosha:
Thanks for your guidance @reaperhulk, that was fun to learn about how a python library can use types, fuzzing and testing to have robust code, I did learn a lot along the way!

Also thanks to @misterzed88 for explaining some of the subtle cryptographic math and providing the initial conversion script, that was very interesting to learn about.

Long live pyca :)

Issue #13700, opened 2025-10-18
key_size and group_order don't match for some obscure EC curves


davidben:
@hannob and I were talking about EC private key sizes, and he noticed that the `key_size` value in some of cryptography's `EllipticCurve` objects don't match the generated keys. They seem to be reporting the size of the field, but the docs say:

> Size (in [bits](https://cryptography.io/en/latest/glossary/#term-bits)) of a secret scalar for the curve (as generated by [generate_private_key()](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#cryptography.hazmat.primitives.asymmetric.ec.generate_private_key)).

That suggests that it should be the group order. Here's a little script that flags curves whose `group_order` and `key_size` fields don't match.

```
$ cat curve_sizes.py
from cryptography.hazmat.primitives.asymmetric import ec

curves = [
    ec.SECP256R1,
    ec.SECP384R1,
    ec.SECP521R1,
    ec.SECP224R1,
    ec.SECP192R1,
    ec.SECP256K1,
    ec.BrainpoolP256R1,
    ec.BrainpoolP384R1,
    ec.BrainpoolP512R1,
    ec.SECT571K1,
    ec.SECT409K1,
    ec.SECT283K1,
    ec.SECT233K1,
    ec.SECT163K1,
    ec.SECT571R1,
    ec.SECT409R1,
    ec.SECT283R1,
    ec.SECT233R1,
    ec.SECT163R2,
]

for curve in curves:
    # These should be equal. According to the docs, key_size is:
    # > Size (in bits) of a secret scalar for the curve (as generated
    # > by generate_private_key()).
    if curve.key_size != curve.group_order.bit_length():
        print(f"{curve.name} MISMATCH: key_size = {curve.key_size}; group_order.bit_length() = {curve.group_order.bit_length()}")

    # Confirm curve.group_order is indeed how private keys are generated.
    for _ in range(100):
        key = ec.generate_private_key(curve())
        assert key.private_numbers().private_value < curve.group_order

$ python3 curve_sizes.py 
sect571k1 MISMATCH: key_size = 571; group_order.bit_length() = 570
sect409k1 MISMATCH: key_size = 409; group_order.bit_length() = 407
sect283k1 MISMATCH: key_size = 283; group_order.bit_length() = 281
sect233k1 MISMATCH: key_size = 233; group_order.bit_length() = 232
sect283r1 MISMATCH: key_size = 283; group_order.bit_length() = 282
```

[...]

PR #13750, opened 2025-10-27
Correct key_size values for sect curves

hannob:
Some of the sect curves used the size of the field as their key size, not the group order.

Closes: https://github.com/pyca/cryptography/issues/13700


alex:
thanks!

PR #13759, opened 2025-10-29
Support for copy.deepcopy keys

Courtcircuits:
This adds support for deep copying keys. This resolves issue #13588.

I tried to follow as close as possible what was done in PR #12110. Hope it fits what was expected for this issue !

Basically, this PR adds deep copy support for :
- RSA
- DH
- DSA
- EC
- ED25519
- ED448
- X25519
- X448

### DIFF
--- a/src/rust/src/backend/dh.rs
+++ b/src/rust/src/backend/dh.rs
@@ -250,7 +250,7 @@ impl DHPrivateKey {
 
     fn __deepcopy__<'p>(
         slf: pyo3::PyRef<'p, Self>,
-        _memo: &pyo3::Bound<'p, pyo3::types::PyDict>,
+        _memo: &pyo3::Bound<'p, pyo3::PyAny>,
     ) -> pyo3::PyRef<'p, Self> {
         slf
     }
@@ -322,7 +322,7 @@ impl DHPublicKey {
 
     fn __deepcopy__<'p>(
         slf: pyo3::PyRef<'p, Self>,
-        _memo: &pyo3::Bound<'p, pyo3::types::PyDict>,
+        _memo: &pyo3::Bound<'p, pyo3::types::PyAny>,
     ) -> pyo3::PyRef<'p, Self> {
         slf
     }
--- a/src/rust/src/backend/dsa.rs
+++ b/src/rust/src/backend/dsa.rs
@@ -159,7 +159,7 @@ impl DsaPrivateKey {
 
     fn __deepcopy__<'p>(
         slf: pyo3::PyRef<'p, Self>,
-        _memo: &pyo3::Bound<'p, pyo3::types::PyDict>,
+        _memo: &pyo3::Bound<'p, pyo3::PyAny>,
     ) -> pyo3::PyRef<'p, Self> {
         slf
     }
@@ -239,7 +239,7 @@ impl DsaPublicKey {
 
     fn __deepcopy__<'p>(
         slf: pyo3::PyRef<'p, Self>,
-        _memo: &pyo3::Bound<'p, pyo3::types::PyDict>,
+        _memo: &pyo3::Bound<'p, pyo3::PyAny>,
     ) -> pyo3::PyRef<'p, Self> {
         slf
     }
--- a/src/rust/src/backend/ec.rs
+++ b/src/rust/src/backend/ec.rs
@@ -376,7 +376,7 @@ impl ECPrivateKey {
 
     fn __deepcopy__<'p>(
         slf: pyo3::PyRef<'p, Self>,
-        _memo: &pyo3::Bound<'p, pyo3::types::PyDict>,
+        _memo: &pyo3::Bound<'p, pyo3::PyAny>,
     ) -> pyo3::PyRef<'p, Self> {
         slf
     }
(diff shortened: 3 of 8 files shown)


src/rust/src/backend/dh.rs
```diff
@@ -247,6 +247,17 @@ impl DHPrivateKey {
     fn __copy__(slf: pyo3::PyRef<'_, Self>) -> pyo3::PyRef<'_, Self> {
         slf
     }
+
+    fn __deepcopy__<'p>(
+        slf: pyo3::PyRef<'p, Self>,
+        _memo: &pyo3::Bound<'p, pyo3::types::PyDict>,
+    ) -> CryptographyResult<Self> {
+        let new_key = Self {
```
alex: You shouldn't need to do this, because keys are immutable a version of `return self` (like `__copy__`) should be enough.

src/rust/src/backend/dsa.rs
```diff
@@ -153,9 +153,22 @@ impl DsaPrivateKey {
         )
     }
 
+    fn __eq__(&self, other: pyo3::PyRef<'_, Self>) -> bool {
```
alex: What's going on here? Why is this needed?

src/rust/src/backend/dh.rs
Courtcircuits: Ok I didn't know that, I'll basically use the same as `__copy__` then. Thanks for the quick review !

src/rust/src/backend/dsa.rs
Courtcircuits: I forgot to uncommit it. I removed it.

src/rust/src/backend/dh.rs
```diff
@@ -247,6 +247,13 @@ impl DHPrivateKey {
     fn __copy__(slf: pyo3::PyRef<'_, Self>) -> pyo3::PyRef<'_, Self> {
         slf
     }
+
+    fn __deepcopy__<'p>(
+        slf: pyo3::PyRef<'p, Self>,
+        _memo: &pyo3::Bound<'p, pyo3::types::PyDict>,
```
alex: nit: given these are unused, it's probably better to just type them as PyAny

tests/hazmat/primitives/test_dsa.py
```diff
@@ -409,6 +409,18 @@ def test_public_key_copy(self):
...
+    def test_public_key_deepcopy(self):
+        key_bytes = load_vectors_from_file(
+            os.path.join("asymmetric", "PKCS8", "unenc-dsa-pkcs8.pem"),
+            lambda pemfile: pemfile.read().encode(),
+        )
+        key1 = serialization.load_pem_private_key(key_bytes, None).public_key()
+        key2 = copy.deepcopy(key1)
+        assert key1 == key2
+
+        key1 = dsa.generate_private_key(2048).public_key()
```
alex: IIRC dsa key generation is quite slow, is there some other key we can load to test here?

tests/hazmat/primitives/test_rsa.py
```diff
@@ -2795,3 +2796,25 @@ def test_private_key_copy(self, rsa_key_2048: rsa.RSAPrivateKey):
...
+
+    def test_public_key_deepcopy(self, rsa_key_2048: rsa.RSAPrivateKey):
+        key1 = rsa_key_2048.public_key()
+        key2 = copy.deepcopy(key1)
+
+        assert key1.public_numbers() == key2.public_numbers()
+
+        key1 = generate_private_key(
+            public_exponent=65537, key_size=2048
+        ).public_key()
```
alex: same, RSA key gen also takes appreciable time, there should be other vectors available

tests/hazmat/primitives/test_dsa.py
Courtcircuits: I replaced the generated key by a key in an already existing PEM file (this one : `vectors/cryptography_vectors/asymmetric/Traditional_OpenSSL_Serialization/dsa.1024.pem`). The goal of this test is mainly to make sure that deepcopy really creates a deep copy of the key. But if I understand correctly, @joakimnordling in issue #13588 is just asking for `copy.deepcopy` to work on keys.

tests/hazmat/primitives/test_rsa.py
Courtcircuits: For this I replaced the generated key with the test fixture `rsa_key_512`.

src/rust/src/backend/dh.rs
Courtcircuits: Done !

src/cryptography/hazmat/primitives/asymmetric/dh.py
```diff
@@ -87,6 +87,12 @@ def __copy__(self) -> DHPublicKey:
         Returns a copy.
         """
 
+    @abc.abstractmethod
+    def __deepcopy__(self) -> DHPublicKey:
```
joakimnordling, line 91: Isn't the `__deepcopy__` expected to take in the `memo` as a second parameter as described on https://docs.python.org/3/library/copy.html#object.__deepcopy__ ?

alex: oof yes, not sure how I missed this. Will send a fix.

joakimnordling: In case you missed it, the version in `docs/hazmat/primitives/asymmetric/cloudhsm.rst` has a `memodict` (not sure if anything cares about if it has that name of `memo`, but I'd at least change it to match the name in the documentation. And also note that the parameter is missing in not just this one I commented on, but all the *.py files.

PR #13789, opened 2025-11-02
migrate X963KDF to rust

reaperhulk:


src/rust/src/backend/kdf.rs
```diff
@@ -834,8 +834,120 @@ impl HkdfExpand {
     }
 }
 
+// NO-COVERAGE-START
```
alex, line 837: Did you verify if this was actually needed?

alex (approved):

src/rust/src/backend/kdf.rs
```diff
@@ -834,8 +834,116 @@ impl HkdfExpand {
...
+        _ = backend;
+
+        let digest_size = algorithm
+            .bind(py)
+            .getattr(pyo3::intern!(py, "digest_size"))?
+            .extract::<usize>()?;
+
+        let max_len = digest_size.saturating_mul(u32::MAX as usize);
+
+        if length > max_len {
```
alex, line 870: This is wrong for length=usize::MAX, right?

tests/hazmat/primitives/test_x963kdf.py
```diff
@@ -13,6 +14,10 @@
 
 
 class TestX963KDF:
+    @pytest.mark.skipif(
+        sys.maxsize <= 2**31,
+        reason="Skip on 32-bit systems",
+    )
```
alex: Why skip? We should just assert the correct behavior.

alex (approved):

PR #13798, opened 2025-11-03
Fix deepcopy signature on ABCs

alex:


src/cryptography/hazmat/primitives/asymmetric/dh.py
```diff
@@ -88,7 +88,7 @@ def __copy__(self) -> DHPublicKey:
         """
 
     @abc.abstractmethod
-    def __deepcopy__(self) -> DHPublicKey:
+    def __deepcopy__(self, memodict: dict) -> DHPublicKey:
```
joakimnordling: Why `memodict` and not `memo`, like in the [Python docs](https://docs.python.org/3/library/copy.html#object.__deepcopy__) for `object.__deepcopy__(self, memo)`? Also note the same name was also used in `docs/hazmat/primitives/asymmetric/cloudhsm.rst` which is not touched in this PR.

alex: 🙃 I copied from cloudhsm, which is why this matches. Using the same name as python upstrema makes sense.

reaperhulk (approved):

PR #13800, opened 2025-11-03
migrate kbkdfhmac to rust

reaperhulk:
Since this only migrates the HMAC part there is a lot of duplication, which will be removed when CMAC gets migrated. To review the logic you'll need to look at the _KBKDFDeriver class (which isn't removed since KBKDFCMAC still depends on it)

### DIFF
--- a/src/rust/src/backend/kdf.rs
+++ b/src/rust/src/backend/kdf.rs
@@ -1764,12 +1764,14 @@ fn validate_kbkdf_parameters(
         CounterLocation::AfterFixed
     } else {
         // There are only 3 options so this is MiddleFixed
-        if break_location.is_none() {
-            return Err(CryptographyError::from(
-                pyo3::exceptions::PyValueError::new_err("Please specify a break_location"),
-            ));
+        match break_location {
+            Some(break_location) => CounterLocation::MiddleFixed(break_location),
+            None => {
+                return Err(CryptographyError::from(
+                    pyo3::exceptions::PyValueError::new_err("Please specify a break_location"),
+                ))
+            }
         }
-        CounterLocation::MiddleFixed(break_location.unwrap())
     };
 
     if break_location.is_some() && !matches!(rust_location, CounterLocation::MiddleFixed(_)) {
@@ -1848,8 +1850,6 @@ impl KbkdfHmac {
             CounterLocation::BeforeFixed => (&b""[..], &fixed[..]),
             CounterLocation::AfterFixed => (&fixed[..], &b""[..]),
             CounterLocation::MiddleFixed(break_location) => {
-                // We validate break_location is Some when counter_location is MiddleFixed
-                // in the validate function
                 if break_location > fixed.len() {
                     return Err(CryptographyError::from(
                         pyo3::exceptions::PyValueError::new_err(


reaperhulk:
These commits can be reviewed separately to make it a bit easier to follow. The second also provides coverage since we need a test to pass an incorrect buffer size.

src/rust/src/backend/kdf.rs
```diff
@@ -1696,11 +1696,306 @@ impl ConcatKdfHmac {
...
+    }
+
+    if fixed.is_none() && llen.is_none() {
+        return Err(CryptographyError::from(
+            pyo3::exceptions::PyValueError::new_err("Please specify an llen"),
+        ));
+    }
+
+    if let Some(l) = llen {
+        if l == 0 {
```
alex: `if llen == Some(0)`

src/rust/src/backend/kdf.rs
```diff
@@ -1696,11 +1696,306 @@ impl ConcatKdfHmac {
...
+            return Err(CryptographyError::from(
+                pyo3::exceptions::PyValueError::new_err("llen must be non-zero"),
+            ));
+        }
+    }
+
+    Ok(KbkdfValidatedParams {
```
alex: why validated? why not just kbkdfparams

src/rust/src/backend/kdf.rs
```diff
@@ -1696,11 +1696,306 @@ impl ConcatKdfHmac {
...
+        let counter_location = self.params.location.bind(py);
+        let counter_location_before_fixed = crate::types::KBKDF_COUNTER_LOCATION
+            .get(py)?
+            .getattr(pyo3::intern!(py, "BeforeFixed"))?;
+        let counter_location_after_fixed = crate::types::KBKDF_COUNTER_LOCATION
+            .get(py)?
+            .getattr(pyo3::intern!(py, "AfterFixed"))?;
+
+        let (data_before_ctr, data_after_ctr) = if counter_location
+            .eq(&counter_location_before_fixed)?
```
alex: Should we translate this into an internal enum?

src/rust/src/backend/kdf.rs
```diff
@@ -1696,11 +1696,321 @@ impl ConcatKdfHmac {
...
+        Ok(self.length)
+    }
+
+    fn generate_fixed_input(&self, py: pyo3::Python<'_>) -> CryptographyResult<Vec<u8>> {
+        if let Some(ref fixed_data) = self.params.fixed {
+            return Ok(fixed_data.as_bytes(py).to_vec());
+        }
+
+        // llen will exist if fixed data is not provided
+        let l_val = int_to_bytes(self.length * 8, self.params.llen.unwrap());
```
alex: We don't ever verify the range of llen, but `int_to_bytes` will panic if it's too big: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a4c1b2f7067040c7a9f906a14fe15f0b

src/rust/src/backend/kdf.rs
```diff
@@ -1696,11 +1697,317 @@ impl ConcatKdfHmac {
...
+        ));
+    }
+
+    if break_location.is_some() && rust_location != CounterLocation::MiddleFixed {
+        return Err(CryptographyError::from(
+            pyo3::exceptions::PyValueError::new_err(
+                "break_location is ignored when location is not CounterLocation.MiddleFixed",
+            ),
+        ));
+    }
```
alex: `break_location` should probably be an element inn `MiddleFixed` rather than a separate Option

src/rust/src/backend/kdf.rs
```diff
@@ -1696,11 +1697,317 @@ impl ConcatKdfHmac {
...
+
+    fn generate_fixed_input(&self, py: pyo3::Python<'_>) -> CryptographyResult<Vec<u8>> {
+        if let Some(ref fixed_data) = self.params.fixed {
+            return Ok(fixed_data.as_bytes(py).to_vec());
+        }
+
+        // llen will exist if fixed data is not provided
+        let py_bitlength = pyo3::types::PyInt::new(py, self.length)
+            .mul(8)?
+            .extract::<pyo3::Bound<'_, pyo3::types::PyInt>>()?;
```
alex: Uhh, can we just multiply by 8 in rust? :D

src/rust/src/asn1.rs
```diff
@@ -71,22 +71,38 @@ pub(crate) fn py_uint_to_big_endian_bytes<'p>(
     py: pyo3::Python<'p>,
     v: pyo3::Bound<'p, pyo3::types::PyInt>,
 ) -> pyo3::PyResult<PyBackedBytes> {
-    if v.lt(0)? {
-        return Err(pyo3::exceptions::PyValueError::new_err(
-            "Negative integers are not supported",
-        ));
-    }
-
+    reject_negative_integer(&v)?;
```
alex: I don't think you need this call, since `py_uint_to_be_bytes_with_length` does the check

src/rust/src/backend/kdf.rs
reaperhulk: You mean in the internal rust mapping? Because that ship has sailed as public API unfortunately.

alex: yes

alex (approved):

PR #13835, opened 2025-11-08
Refactor CRL to use PyBytes ownership instead of Arc

alex:
Simplifies the CRL implementation by removing Arc indirection - OwnedRevokedCertificate and OwnedCRLIteratorData now own Py<PyBytes> directly. Adds sound abstraction functions for lifetime rebinding following the existing try_map_arc_data_mut_crl_iterator pattern.


reaperhulk (approved):
well that was confusing 🥳

PR #13839, opened 2025-11-09
kbkdfcmac oxidized

reaperhulk:


src/rust/src/backend/cmac.rs
```diff
@@ -14,22 +14,54 @@ use crate::{exceptions, types};
     module = "cryptography.hazmat.bindings._rust.openssl.cmac",
```
alex, line 14: if there ends up being other review feedback, can this be split out?

src/rust/src/backend/kdf.rs
```diff
@@ -1981,7 +1978,159 @@ impl KbkdfHmac {
...
+        }
+        self.used = true;
+        kbkdf_derive_into_buffer(
+            py,
+            self.length,
+            self.digest_size,
+            &self.params,
+            buf.as_mut_bytes(),
+            |data| {
+                let mut hmac =
```
alex: our previous impl also had this proble, but shouldn't we be creating an hmac once and then copying it each iteration, since the key never changes?

PR #13894, opened 2025-11-19
asn1: Add support for `SEQUENCE OF`

facutuesca:
This PR adds support for `SEQUENCE OF` fields to the ASN.1 API. The fields must be annotated as Python lists:

```python
@asn1.sequence
class Example
  a: typing.List[int]
```

### DIFF
--- a/src/rust/src/declarative_asn1/encode.rs
+++ b/src/rust/src/declarative_asn1/encode.rs
@@ -4,6 +4,7 @@
 
 use asn1::{SimpleAsn1Writable, Writer};
 use pyo3::types::PyAnyMethods;
+use pyo3::types::PyListMethods;
 
 use crate::declarative_asn1::types::{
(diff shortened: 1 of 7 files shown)


src/rust/src/declarative_asn1/encode.rs
```diff
@@ -70,6 +71,18 @@ impl asn1::Asn1Writable for AnnotatedTypeObject<'_> {
                 }),
                 encoding,
             ),
+            Type::SequenceOf(cls) => {
+                let values: Vec<AnnotatedTypeObject<'_>> = value
```
alex, line 75: Is it practical to not have to allocate a whole vec and instead just iterate?

facutuesca: can that be done with the existing `asn1::SequenceOf` API?

alex: mmm, probably not, no

PR #13899, opened 2025-11-20
asn1: Add support for `SIZE` to `SEQUENCE OF`

facutuesca:
This PR adds support for `SIZE` annotations to the ASN.1 API, starting with fields of type `SEQUENCE OF`.

```python
@asn1.sequence
class Example
  a: Annotated[typing.List[int], asn1.Size(min=1, max=5)]  # SIZE(1..5)

@asn1.sequence
class Example
  a: Annotated[typing.List[int], asn1.Size(min=1, max=None)]  # SIZE(1..MAX)


@asn1.sequence
class Example
  a: Annotated[typing.List[int], asn1.Size.exact(6)]  # SIZE(6)
```

### DIFF
--- a/src/rust/src/declarative_asn1/decode.rs
+++ b/src/rust/src/declarative_asn1/decode.rs
@@ -166,7 +166,7 @@ pub(crate) fn decode_annotated_type<'a>(
                     let list_len = list.len();
                     let min = size.get().min;
                     let max = size.get().max.unwrap_or(usize::MAX);
-                    if list_len < min || list_len > max {
+                    if !(min..=max).contains(&list_len) {
                         return Err(CryptographyError::Py(
                             pyo3::exceptions::PyValueError::new_err(format!(
                                 "SEQUENCE OF has size {0}, expected size in [{1}, {2}]",
--- a/src/rust/src/declarative_asn1/encode.rs
+++ b/src/rust/src/declarative_asn1/encode.rs
@@ -85,7 +85,7 @@ impl asn1::Asn1Writable for AnnotatedTypeObject<'_> {
                 if let Some(size) = &annotated_type.annotation.get().size {
                     let min = size.get().min;
                     let max = size.get().max.unwrap_or(usize::MAX);
-                    if values.len() < min || values.len() > max {
+                    if !(min..=max).contains(&values.len()) {
                         return Err(asn1::WriteError::AllocationError);
                     }
                 }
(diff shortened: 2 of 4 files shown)


src/cryptography/hazmat/bindings/_rust/declarative_asn1.pyi
```diff
@@ -36,6 +38,14 @@ class Encoding:
     Implicit: typing.ClassVar[type]
     Explicit: typing.ClassVar[type]
 
+class Size:
+    min: int
+    max: int | None
+
+    def __new__(cls, min: int, max: int | None) -> Size: ...
```
alex, line 45: I guess if you want no min you just do min=0?

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -162,6 +162,19 @@ pub(crate) fn decode_annotated_type<'a>(
                     let val = decode_annotated_type(py, d, inner_ann_type)?;
                     list.append(val)?;
                 }
+                if let Some(size) = &ann_type.annotation.get().size {
+                    let list_len = list.len();
+                    let min = size.get().min;
+                    let max = size.get().max.unwrap_or(usize::MAX);
+                    if list_len < min || list_len > max {
```
alex: I think this is like, 2% easier to read as `if !(min..=max).contains(&list_len)` since thats it unavoidable that it's an inclusive range

src/rust/src/declarative_asn1/encode.rs
```diff
@@ -81,6 +81,14 @@ impl asn1::Asn1Writable for AnnotatedTypeObject<'_> {
                         value: e,
                     })
                     .collect();
+
+                if let Some(size) = &annotated_type.annotation.get().size {
+                    let min = size.get().min;
+                    let max = size.get().max.unwrap_or(usize::MAX);
+                    if values.len() < min || values.len() > max {
```
alex: Same

alex (commented):
we should block size annotations with types we don't do anything with them

src/cryptography/hazmat/bindings/_rust/declarative_asn1.pyi
facutuesca: yup

src/rust/src/declarative_asn1/decode.rs

src/rust/src/declarative_asn1/encode.rs

facutuesca:
> we should block size annotations with types we don't do anything with them

done!

alex (approved):

PR #13908, opened 2025-11-21
asn1: Add support for `BIT STRING`

facutuesca:
This PR adds support for `BIT STRING` to the ASN.1 API

```python
@asn1.sequence
class Example
  field: asn1.BitString

bt = asn1.BitString(data=b"\x01\x10", padding_bits=2)

object = Example(field=bt)
encoded = encode_der(object)

decoded = decode_der(Example, encoded)
assert decoded.field.as_bytes() == b"\x01\x10"
assert decoded.field.padding_bits() == 2
```

### DIFF
--- a/src/rust/src/declarative_asn1/types.rs
+++ b/src/rust/src/declarative_asn1/types.rs
@@ -96,7 +96,6 @@ impl Annotation {
         }
     }
 
-    #[pyo3(signature = ())]
     fn is_empty(&self) -> bool {
         self.default.is_none() && self.encoding.is_none() && self.size.is_none()
     }
@@ -150,7 +149,6 @@ impl PrintableString {
         Ok(PrintableString { inner })
     }
 
-    #[pyo3(signature = ())]
     pub fn as_str(&self, py: pyo3::Python<'_>) -> pyo3::PyResult<pyo3::Py<pyo3::types::PyString>> {
         Ok(self.inner.clone_ref(py))
     }
@@ -194,7 +192,6 @@ impl UtcTime {
         Ok(UtcTime { inner })
     }
 
-    #[pyo3(signature = ())]
     pub fn as_datetime(
         &self,
         py: pyo3::Python<'_>,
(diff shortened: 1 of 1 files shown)


alex:
@reaperhulk does a BitString type (with API that looks like rust-asn1: as_byes() + padding_bits()) look right to you?

reaperhulk:
There are two primary ways bit strings are used. One is the “wrong type”, as in a signature in x509, where you really just want the byte value. The other is an actual bit field, like key usage or CRL revocation reasons. This API seems fine for the former and lacking for the latter.

alex:
Nothing prevents us from introducing another representation of bitstring for that in the future. Will go ahead and review this PR as-is.

src/rust/src/declarative_asn1/types.rs
```diff
@@ -249,6 +251,54 @@ impl GeneralizedTime {
...
+
+    fn __eq__(&self, py: pyo3::Python<'_>, other: pyo3::PyRef<'_, Self>) -> pyo3::PyResult<bool> {
+        Ok((**self.data.bind(py)).eq(other.data.bind(py))?
+            && self.padding_bits == other.padding_bits)
+    }
+
+    pub fn __repr__(&self) -> pyo3::PyResult<String> {
+        Ok(format!(
+            "BitString(data: {}, padding_bits: {})",
+            self.data, self.padding_bits,
```
alex, line 291: repr should be formatted like the constructor, so =, not :

src/rust/src/declarative_asn1/types.rs
```diff
@@ -249,6 +251,54 @@ impl GeneralizedTime {
...
+
+        Ok(BitString { data, padding_bits })
+    }
+
+    #[pyo3(signature = ())]
+    pub fn as_bytes(&self, py: pyo3::Python<'_>) -> pyo3::Py<pyo3::types::PyBytes> {
+        self.data.clone_ref(py)
+    }
+
+    #[pyo3(signature = ())]
```
alex: Is this necessary for anything?

src/rust/src/declarative_asn1/types.rs
facutuesca: fixed

src/rust/src/declarative_asn1/types.rs
facutuesca: nope, fixed

PR #13910, opened 2025-11-22
asn1.py: treat asn1_string_st as opaque

botovq:
OpenSSL 4 plans to [make ASN1_STRING opaque][1]. Using a forward declaration rather than a redefinition avoids the build breakage in cryptography's cffi when it tries to validate the sizes of the members.

[1]: https://github.com/openssl/openssl/issues/29117

### DIFF
--- a/src/_cffi_src/openssl/asn1.py
+++ b/src/_cffi_src/openssl/asn1.py
@@ -12,19 +12,11 @@ TYPES = """
 typedef int... time_t;
 
 typedef ... ASN1_INTEGER;
-
-struct asn1_string_st {
-    int length;
-    int type;
-    unsigned char *data;
-    long flags;
-};
-
-typedef struct asn1_string_st ASN1_OCTET_STRING;
-typedef struct asn1_string_st ASN1_IA5STRING;
-typedef struct asn1_string_st ASN1_TIME;
+typedef ... ASN1_OCTET_STRING;
+typedef ... ASN1_IA5STRING;
+typedef ... ASN1_TIME;
 typedef ... ASN1_OBJECT;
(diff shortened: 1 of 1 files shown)


alex:
oh my god, apparently pyopenssl still uses internals 😭. Looks like we have `ASN1_STRING_get0_data` and `ASN1_STRING_length` exposed, I'll go fix there.

botovq:
I see this causes issues in pyopenssl which will need to be addressed first.

src/_cffi_src/openssl/asn1.py
```diff
@@ -13,12 +13,7 @@
 
 typedef ... ASN1_INTEGER;
 
-struct asn1_string_st {
-    int length;
-    int type;
-    unsigned char *data;
-    long flags;
-};
+struct asn1_string_st;
```
alex: I think you can probably remove this entirely and change the tyepdefs below to just use `...` in place of `asn1_string_st`.

alex:
https://github.com/pyca/pyopenssl/pull/1462 has the fix -- will mark this for the next milestone and expect to rebase/merge after our next pyopenssl release.

src/_cffi_src/openssl/asn1.py
botovq: I was thinking about this. I believe it loses the information that they all have the same underlying type, which may be useful for calling `ASN1_STRING_length()` on an `ASN1_INTEGER` for example. But if you don't need ugly things like that I'll of course change to `...`.

alex:
ooops, forgot you had a PR open for this so I did it myself, it's in our latest release.

botovq:
cool, thanks.

PR #13928, opened 2025-11-26
Enforce 192-bit key length for TripleDES

alex:
TripleDES now only accepts 24-byte (192-bit) keys. Users needing single DES (1-key) or two-key Triple DES compatibility must expand the key themselves:
- Single DES (8 bytes): key + key + key
- Two-key 3DES (16 bytes): key + key[:8]

This change removes the automatic key expansion that was previously done in the TripleDES constructor, making the key length requirement explicit and preventing accidental use of weaker key configurations.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,11 @@ Changelog
   such keys are impossible to process in a constant-time manner. We do not
   believe keys with this problem are in wide use, however we may revert this
   change based on the feedback we receive.
+* Deprecated passing 64-bit (8-byte) and 128-bit (16-byte) keys to
+  :class:`~cryptography.hazmat.decrepit.ciphers.algorithms.TripleDES`. In a
+  future release, only 192-bit (24-byte) keys will be accepted. Users should
+  expand shorter keys themselves (e.g., for single DES: ``key + key + key``,
+  for two-key: ``key + key[:8]``).
 * Updated the minimum supported Rust version (MSRV) to 1.83.0, from 1.74.0.
 * Added support for loading elliptic curve keys that contain explicit encodings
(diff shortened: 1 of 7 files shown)


reaperhulk:
Have we tagged our downstreams to let them know about this?

alex:
No, I was going to let the warning play that role. (Since this is a warning and not an immediate breakage.)

alex:
FYI @glyph, @gpotter2 -- this appears to trigger warnings in twisted and scapy. 

For twisted I think there's some tests that just use a short key for no reason. For scapy I think there's code that really does want Single DES (in which case this behavior can be obtained by concatting the key 3x).

glyph:
We definitely have a recurring issue where people just generate the default, or sometimes even the shortest available, key for tests. And we don't have good infrastructure for responsiveness to warnings 😬

alex:
@glyph good news, the problem is far dumber: https://github.com/twisted/twisted/blob/506d17fb71abf213dc77cfe7b55eb1fafb5f2437/src/twisted/conch/test/test_transport.py#L702-L705 this assumes all keys have a size of 16 bytes, but that's wrong for 3des, it's key size is 24 bytes. If you fix that, I think twisted will be good.

glyph:
OK, unless I hear otherwise, I assume Twisted is good now

gpotter2:
Thanks a lot for the heads up, I've drafted a PR on our side to have this ready.

> I think there's code that really does want Single DES

Heh yeah, for compat with ancient IPSEC, Kerberos and TLS cases.

reaperhulk (approved):

alex:
thanks for jumping on this!

PR #13985, opened 2025-12-08
asn1: Add support for `IA5String`

facutuesca:
This adds support for encoding/decoding `IA5String`. 

```python
@asn1.sequence
class MySeq:
    field: asn1.IA5String

object = MySeq(field=asn1.IA5String("A string"))
encoded = encode_der(object)

assert decoded.field.as_str() == "A string"
```


alex:
Another follow up: can you have us use `to_str()` on py310+ and `to_cow()` only on older Pythons? We do 3.11 wheels, so they'll get the perf win.

PR #13986, opened 2025-12-08
asn1: Add TODO comment for uses of `PyStringMethods::to_cow`

facutuesca:
From https://github.com/pyca/cryptography/pull/13985#issuecomment-3627669112:

> Another follow up: can you have us use to_str() on py310+ and to_cow() only on older Pythons? We do 3.11 wheels, so they'll get the perf win.

Since `pyo3` [already internally uses](https://docs.rs/pyo3/latest/src/pyo3/types/string.rs.html#346-361) `to_str` when calling `to_cow` in Python 3.10+, we just add a comment to migrate once the minimum Python version is sufficient


facutuesca:
> I think we're going to have this pattern in several places, let's put it in a function.

Changed the PR to only add a TODO comment because `pyo3` [already internally uses](https://docs.rs/pyo3/latest/src/pyo3/types/string.rs.html#346-361) `to_str` when calling `to_cow` in Python 3.10+.

PR #13999, opened 2025-12-10
asn1: Add `SIZE` support to `BIT STRING`

facutuesca:
Similar to https://github.com/pyca/cryptography/pull/13899, this PR extends the support for `SIZE` to `BIT STRING`

The logic to check the size constraint is moved to its own function (`check_size_constraint()`), so that it can be reused by all types that support `SIZE` annotations.

I've also moved the existing tests for `SIZE` to their own class.

### DIFF
--- a/src/rust/src/declarative_asn1/types.rs
+++ b/src/rust/src/declarative_asn1/types.rs
@@ -134,6 +134,13 @@ impl Size {
     }
 }
 
+// TODO: Once the minimum Python version is >= 3.10, use a `self_cell`
+// to store the owned PyString along with the dependent Asn1PrintableString
+// in order to avoid verifying the string twice (once during construction,
+// and again during serialization).
+// This is because for Python < 3.10 getting an Asn1PrintableString object
+// from a PyString requires calling `to_cow()`, which creates an intermediate
+// `Cow` object with a different lifetime from the PyString.
 #[derive(pyo3::FromPyObject)]
 #[pyo3::pyclass(frozen, module = "cryptography.hazmat.bindings._rust.asn1")]
 pub struct PrintableString {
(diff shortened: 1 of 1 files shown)


src/rust/src/declarative_asn1/encode.rs
```diff
@@ -191,10 +188,12 @@ impl asn1::Asn1Writable for AnnotatedTypeObject<'_> {
                 let val: &pyo3::Bound<'_, BitString> = value
                     .cast()
                     .map_err(|_| asn1::WriteError::AllocationError)?;
-
                 let bitstring: asn1::BitString<'_> =
                     asn1::BitString::new(val.get().data.as_bytes(py), val.get().padding_bits)
                         .ok_or(asn1::WriteError::AllocationError)?;
+                let n_bits = bitstring.as_bytes().len() * 8 - (bitstring.padding_bits() as usize);
```
alex: You can use `into()` instead of `as` so its clear there's no truncation risk

tests/hazmat/asn1/test_api.py
```diff
@@ -238,8 +238,11 @@ class Example:
     def test_fail_unsupported_size_annotation(self) -> None:
         with pytest.raises(
             TypeError,
-            match="field invalid has a SIZE annotation, but SIZE "
-            "annotations are only supported for SEQUENCE OF fields",
+            match=re.escape(
+                "field invalid has a SIZE annotation, but SIZE "
+                "annotations are only supported for fields of types: "
+                "[SEQUENCE OF, BIT STRING]"
```
alex: We can probably simplify this to just the first part of the error, so it doesn't need to be updated for every type

facutuesca:
> @reaperhulk this uses bits as the length unit for a size annotation, WDYT?
> 
> On the on one hand, bits are stupid.
> 
> @facutuesca is that waht RFCs do when SIZE OF annotations are on BIT STRING?

They should. From [ITU-T X.680](https://www.itu.int/rec/T-REC-X.680-202102-I/en):
image

src/rust/src/declarative_asn1/encode.rs
facutuesca: fixed (with `usize::from()`)

tests/hazmat/asn1/test_api.py

PR #14003, opened 2025-12-11
Removed indirect dependency on OpenSSL from cryptography-x509-verification

TaaviE:
This PR should decouple `cryptography-x509-verification` from `cryptography-key-parsing` and subsequently OpenSSL.

This makes it possible to (cross-)build the library for targets that would otherwise have difficulties with OpenSSL. It also makes the PKCS#1 parsing pure Rust.

Tests seem to all pass when ran locally.

### DIFF
--- a/src/rust/cryptography-x509-verification/Cargo.toml
+++ b/src/rust/cryptography-x509-verification/Cargo.toml
@@ -10,7 +10,7 @@ license.workspace = true
 [dependencies]
 asn1.workspace = true
 cryptography-x509 = { path = "../cryptography-x509" }
-cryptography-key-parsing = { path = "../cryptography-key-parsing" }
+pkcs1 = "0.7.5"
 
 [dev-dependencies]
 pem.workspace = true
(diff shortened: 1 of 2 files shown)


alex:
We're not going to depend on an entirely different ASN.1 stack for this. If there's an interest in the objective we can consider it, but this isn't going to be the right approach.

TaaviE:
@alex Yeah, that makes sense. I just stumbled upon this when trying to use that code standalone (and without OpenSSL).

It might be worthwhile to consider switching to a memory-safe ASN.1 stack even outside the context of build targets that OpenSSL isn't so pleasant to use with.

alex:
I'm not sure I follow, is there any place we don't use our own rust asn1
stack for der?

TaaviE:
> I'm not sure I follow, is there any place we don't use our own rust asn1 stack for der?

Yeah I misremembered/misphrased, OSSL is just intertwined with the Rust asn1 stack currently used in `cryptography-key-parsing`, including [rsa.rs](https://github.com/pyca/cryptography/blob/main/src/rust/cryptography-key-parsing/src/rsa.rs). Which made me make a (silly) leap into assuming it's used throughout.

alex:
going to close this since it doesn't make sense as-is, but we're generally happy to take improvements to this code.

PR #14025, opened 2025-12-17
Improve error message if neither EKU nor anyEKU could be found

TaaviE:
This function also checks for `|| eku == EKU_ANY_KEY_USAGE_OID` so the error message should contain it.

### DIFF
--- a/src/rust/cryptography-x509-verification/src/policy/extension.rs
+++ b/src/rust/cryptography-x509-verification/src/policy/extension.rs
@@ -642,7 +642,7 @@ mod ca {
                 Ok(())
             } else {
                 Err(ValidationError::new(ValidationErrorKind::Other(
-                    "EKU and anyEKU not found".to_string(),
+                    "Neither EKU nor anyEKU could be found".to_string(),
                 )))
             }
         } else {


src/rust/cryptography-x509-verification/src/policy/extension.rs
```diff
@@ -642,7 +642,7 @@ mod ca {
                 Ok(())
             } else {
                 Err(ValidationError::new(ValidationErrorKind::Other(
-                    "required EKU not found".to_string(),
+                    "EKU not found or anyEKU found".to_string(),
```
alex: this doesn't look quite right -- unless I'm misreading the code, we allow any eku

src/rust/cryptography-x509-verification/src/policy/extension.rs
```diff
@@ -642,7 +642,7 @@ mod ca {
                 Ok(())
             } else {
                 Err(ValidationError::new(ValidationErrorKind::Other(
-                    "required EKU not found".to_string(),
+                    "EKU and anyEKU not found".to_string(),
```
alex: I think this is more accurately an "or"

TaaviE: The error case should be reached only if both are missing?

alex: Correct, but to me "EKU and anyEKU" can be mis-read as "we didn't find _both_"

TaaviE: Hm, true. `Neither EKU or anyEKU could be found`?

alex: "nor" instead of "or", but that sounds good to me.

Thank you!

TaaviE: Made the changes. It's probably worth squashing all the commits to avoid noise.

alex (approved):
we squash on merge, thansk

PR #14052, opened 2025-12-23
Update build docs to work on OpenBSD

gpotter2:
Minor documentation updates on how to build cryptography with custom OpenSSL versions :

- the link now redirects to github, add `-L` option to follow redirections.
- tar requires `-z` on OpenBSD, unlike other plateformes. Since it also works on other plateforms, let's just add it


alex (approved):
thanks

PR #14061, opened 2025-12-28
Add AES-DNDK-GCM AEAD

script3r:
Summary:
- Implement AESDNDKGCM (32-byte root key, 24-byte nonce, no key commitment)
- Derive per-message AES-256-GCM key/IV per draft-gueron-cfrg-dndkgcm
- Expose bindings and add Appendix A2 vector + basic tests

Spec: https://datatracker.ietf.org/doc/draft-gueron-cfrg-dndkgcm/

Testing: python -m pytest tests/hazmat/primitives/test_aead.py -k AESDNDKGCM

### DIFF
--- a/src/rust/src/backend/aead.rs
+++ b/src/rust/src/backend/aead.rs
@@ -56,11 +56,7 @@ fn aes256_ecb_block(key: &[u8], block: &[u8; 16]) -> CryptographyResult<[u8; 16]
 }
 
 fn derive_dndk_gcm_key_iv(key: &[u8], nonce: &[u8]) -> CryptographyResult<([u8; 12], [u8; 32])> {
-    if key.len() != DNDK_GCM_KEY_LEN {
-        return Err(CryptographyError::from(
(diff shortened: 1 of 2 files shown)


alex:
Thanks for submitting this.

Given the draft RFC is still early days, I don't think we'll consider this for merging at this time. For us to add something we generally want there to be a degree of adoption (or sufficient benefits that we want to encourage the adoption!).

In this case there's also the benefit that this can be implemented on top of our existing AES-GCM API.

script3r:
> Thanks for submitting this.
> 
> 
> 
> Given the draft RFC is still early days, I don't think we'll consider this for merging at this time. For us to add something we generally want there to be a degree of adoption (or sufficient benefits that we want to encourage the adoption!).
> 
> 
> 
> In this case there's also the benefit that this can be implemented on top of our existing AES-GCM API.

Thanks @alex! That makes sense. We were looking for extended nonce setups and between xaes-gcm, xchacha-poly1305 and aes-gcm-dndk, we ended up choosing the latter.

As you said, it's pretty simple to construct on top of existing aes-gcm setup, which is one of the reasons we chose it.

PR #14090, opened 2026-01-05
asn1: Add support for `OBJECT IDENTIFIER`

facutuesca:
This adds support for encoding/decoding `OBJECT IDENTIFIER`. 

```python
@asn1.sequence
class MySeq:
    field: asn1.ObjectIdentifier

object = MySeq(field=asn1.ObjectIdentifier("1.3.6.1.4.1.55738.3"))
encoded = encode_der(object)

assert decoded.field == asn1.ObjectIdentifier("1.3.6.1.4.1.55738.3")
```

### DIFF
--- a/tests/hazmat/asn1/test_serialization.py
+++ b/tests/hazmat/asn1/test_serialization.py
@@ -15,8 +15,8 @@ if sys.version_info < (3, 9):
 else:
     from typing import Annotated
 
-import cryptography.hazmat.asn1 as asn1
-import cryptography.x509 as x509
+from cryptography import x509
+from cryptography.hazmat import asn1
 
 U = typing.TypeVar("U")
 
--- a/tests/hazmat/asn1/test_api.py
+++ b/tests/hazmat/asn1/test_api.py
@@ -14,7 +14,7 @@ if sys.version_info < (3, 9):
 else:
     from typing import Annotated
 
-import cryptography.hazmat.asn1 as asn1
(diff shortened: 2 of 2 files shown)


src/rust/src/declarative_asn1/types.rs
```diff
@@ -216,6 +218,29 @@ impl IA5String {
     }
 }
 
+#[pyo3::pyclass(frozen, eq, module = "cryptography.hazmat.bindings._rust.asn1")]
+#[derive(PartialEq)]
```
alex: We shouldn't need a new OID type, this should just work with the existing cryptography OID type

facutuesca: fixed! Should I create an alias for it, so that `x509.ObjectIdentifier` is also available under `asn1.ObjectIdentifier`?

alex: No alias.

tests/hazmat/asn1/test_serialization.py
```diff
@@ -16,6 +16,7 @@
     from typing import Annotated
 
 import cryptography.hazmat.asn1 as asn1
+import cryptography.x509 as x509
```
alex: The right way to do these imports is: `from cryptogrpahy import x509`. sorry I didn't notice this earlier.

facutuesca: fixed!

PR #14127, opened 2026-01-10
Bump x509-limbo and/or wycheproof in CI

woodruffw:
Subsumes #14092.

For context: the 9 cases that are being skipped here were added with https://github.com/C2SP/x509-limbo/pull/533, and involve CABF consistency cross-checks between a cert's CN and its SAN(s). No major validators in the current matrix appear to perform these checks, they're more conformance checks for CAs themselves I suspect.

(We *could* add them as checks though, if you two want!)


alex:
Merging since lierally no one else enforces.

PR #14130, opened 2026-01-10
Add type annotations to vector loading code

alex:


tests/utils.py
```diff
@@ -493,15 +500,17 @@ def load_fips_ecdsa_key_pair_vectors(vector_data):
...
+def load_fips_ecdsa_signing_vectors(
+    vector_data: typing.Iterable[str],
+) -> list[dict]:
     """
     Loads data out of the FIPS ECDSA SigGen vector files.
     """
     vectors = []
-    curve_name: typing.Union[str, None] = None
-    digest_name: typing.Union[str, None] = None
+    curve_name: str | None = None
```
reaperhulk, line 510: I vaguely recall us having some problems assuming we could use this syntax. Did we drop the Python where this was a problem?

alex: `from __future__ import annotations` makes it legal, you can see all of CI is green :-)

reaperhulk: All of CI was green last time too! I can’t remember what happened though 😅

alex: I do not! (It's really fine though, I promise!)

reaperhulk: https://github.com/pyca/cryptography/pull/12927 Might be what I was remembering, which shouldn’t be an issue here

alex: We added a CI job to prevent the recurrence.

reaperhulk (approved):

PR #14190, opened 2026-01-22
Add CABF S/MIME certificate policies

TaaviE:
This PR intends to add the base for S/MIME certificate validation based on the CA/B Forum S/MIME BR 1.0.12.

The biggest change is likely the addition of `cRLDistributionPoints` validation for both WebPKI and S/MIME, to the extent that is the lowest common denominator.

I did not introduce ML-DSA or ML-KEM support, even though the S/MIME BR allows them. Nor did I implement S/MIME signature validation or anything CMS-related. This is intended to be the first "brick" that can be built upon.‡

I took the Server TLS policies as the base, to be as lax or strict as those are. So they generally do not differ a lot. There's also work ongoing to further align S/MIME BR with Server TLS ones, so it might be possible to simplify futher later on.

‡ - Such as strict verification based on specific S/MIME certificate profiles or the signature/signing code in https://github.com/pyca/cryptography/pull/12465 or https://github.com/pyca/cryptography/pull/12267

### DIFF
--- a/src/rust/cryptography-x509-verification/src/policy/extension.rs
+++ b/src/rust/cryptography-x509-verification/src/policy/extension.rs
@@ -265,10 +265,10 @@ impl<'cb, B: CryptoOps + 'cb> ExtensionPolicy<'cb, B> {
                 None,
             ),
             // 7.1.2.3.e: keyUsage (SHALL be present)
-            key_usage: ExtensionValidator::present(
+            key_usage: ExtensionValidator::maybe_present(
                 KEY_USAGE_OID,
                 Criticality::Critical,
-                Some(Arc::new(ca::key_usage)),
+                Some(Arc::new(ee::key_usage)),
             ),
             // 7.1.2.3.g: subjectAlternativeName (SHALL be present)
             subject_alternative_name: ExtensionValidator::maybe_present(


alex:
cc: @woodruffw 

If I'm reading this correctly, the only two substantive differences are CRL DP validation, and two additional key types?

The biggest thing we'd need for this are test cases -- in https://github.com/c2sp/x509-limbo.

TaaviE:
> The biggest thing we'd need for this are test cases -- in https://github.com/c2sp/x509-limbo.

Is that something you have to do or should I? If I can do it, is there documentation about it that would help me get started?

alex:
Anyone can contribute to that repo! We don't really have docs for it,
but https://github.com/C2SP/x509-limbo/tree/main/limbo/testcases if
you look in a file you'll see what it looks like to add a test case.

On Tue, Jan 27, 2026 at 8:46 AM Taavi Eomäe ***@***.***> wrote:
>
> TaaviE left a comment (pyca/cryptography#14190)
>
> The biggest thing we'd need for this are test cases -- in https://github.com/c2sp/x509-limbo.
>
> Is that something you have to do or should I? If I can do it, is there documentation about it that would help me get started?
>
> —



-- 

TaaviE:
Test cases to [x509-limbo](https://github.com/C2SP/x509-limbo/pull/562) are now available in a pending PR.

PR #14201, opened 2026-01-24
asn1: Add support for CHOICE fields

facutuesca:
Following the discussion in https://github.com/pyca/cryptography/pull/14183, here's an alternative implementation for CHOICE fields.
If one of the types inside the union is a `asn1.Variant`, all the other types should be too.

```python
# all types in the union are asn1.Variant
@asn1.sequence
class Example:
    foo: (
        Annotated[Variant[int, typing.Literal["Tag1"]], asn1.Implicit(0)]
        | Annotated[Variant[int, typing.Literal["Tag2"]], asn1.Implicit(1)]
        | Annotated[Variant[str, typing.Literal["Tag3"]], asn1.Implicit(2)]
        | Annotated[Variant[bool, typing.Literal["Tag4"]], asn1.Implicit(3)]
    )

obj = Example(foo=Variant(42, "Tag1"))
encoded = encode_der(obj)

decoded = decode(Example, encoded)
assert isinstance(decoded.foo.value, int)
assert decoded.foo.tag == "Tag1"
assert decoded.foo.value == 42


# none of the types in the union are asn1.Variant
@asn1.sequence
class Example:
    foo: (
        | Annotated[str, asn1.Implicit(0)]
        | Annotated[bool, asn1.Implicit(1)]
    )
```

### DIFF
--- a/tests/hazmat/asn1/test_serialization.py
+++ b/tests/hazmat/asn1/test_serialization.py
@@ -699,11 +699,13 @@ class TestSequence:
             bar: int
 
         assert_roundtrips(
-            [(Example(foo=True, bar=1), b"\x30\x06\x01\x01\xff\x02\x01\x01")]
-        )
-
-        assert_roundtrips(
-            [(Example(foo=None, bar=1), b"\x30\x03\x02\x01\x01")]
+            [
+                (
+                    Example(foo=True, bar=1),
+                    b"\x30\x06\x01\x01\xff\x02\x01\x01",
+                ),
+                (Example(foo=None, bar=1), b"\x30\x03\x02\x01\x01"),
+            ]
         )
 
     def test_fail_sequence_with_choice_decode_nonexistent_variant(


alex:
I haven't reviewed the impl closely yet, but this is probably my favorite of the various APIs we've brainstormed. @reaperhulk wdyt?

reaperhulk:
Hmm, what does this look like when applied to the basic real world example of a generalname? (I would actually do this myself but am currently mobile)

facutuesca:
> Hmm, what does this look like when applied to the basic real world example of a generalname? (I would actually do this myself but am currently mobile)

Something like this (assuming the non-trivial types are defined elsewhere):

```python
#   GeneralName ::= CHOICE {
#        otherName                       [0]     OtherName,
#        rfc822Name                      [1]     IA5String,
#        dNSName                         [2]     IA5String,
#        x400Address                     [3]     ORAddress,
#        directoryName                   [4]     Name,
#        ediPartyName                    [5]     EDIPartyName,
#        uniformResourceIdentifier       [6]     IA5String,
#        iPAddress                       [7]     OCTET STRING,
#        registeredID                    [8]     OBJECT IDENTIFIER }

GeneralName: typing.TypeAlias = (
    Annotated[OtherName, asn1.Implicit(0)]
    | Annotated[
        asn1.Variant[asn1.IA5String, typing.Literal["rfc822Name"]],
        asn1.Implicit(1),
    ]
    | Annotated[
        asn1.Variant[asn1.IA5String, typing.Literal["dNSName"]],
        asn1.Implicit(2),
    ]
    | Annotated[ORAddress, asn1.Implicit(3)]
    | Annotated[Name, asn1.Implicit(4)]
    | Annotated[EDIPartyName, asn1.Implicit(5)]
    | Annotated[
        asn1.Variant[
            asn1.IA5String, typing.Literal["uniformResourceIdentifier"]
        ],
        asn1.Implicit(6),
    ]
    | Annotated[bytes, asn1.Implicit(7)]
    | Annotated[x509.ObjectIdentifier, asn1.Implicit(8)]
)

@asn1.sequence
class Example:
    name: GeneralName

obj = Example(name=asn1.Variant(asn1.IA5String("abcd"), "dNSName"))
encoded = encode_der(obj)

decoded = decode(Example, encoded)
assert isinstance(decoded.name.value, asn1.IA5String)
assert decoded.name.tag == "dNSName"
assert decoded.name.value.as_str() == "abcd"
```

facutuesca:
I wonder if maybe we should require the use of `Variant` with a tag for all the possible types? Makes the API more uniform, both when declaring the types but also when the user needs to get the type and value from a decoded variant. Downside is that it's more verbose.


```python
# current API, this would be the `iPAddress [7] OCTET STRING` variant
obj = Example(name=b"\x01\x02")
encoded = encode_der(obj)

decoded = decode(Example, encoded)
assert isinstance(decoded.name, bytes)
assert decoded.name == b"\x01\x02"


# API where every variant is defined with a `Variant(type, tag)` annotation
GeneralName: typing.TypeAlias = (
#...
    | Annotated[
        asn1.Variant[bytes, typing.Literal["iPAddress"]],
        asn1.Implicit(7),
    ]
# ...

obj = Example(name=asn1.Variant(b"\x01\x02", "iPAddress"))
encoded = encode_der(obj)

decoded = decode(Example, encoded)

assert isinstance(decoded.name.value, bytes)
assert decoded.name.tag == "ipAddress"
assert decoded.name.value == b"\x01\x02"
```

alex:
Either all variant or no-variant perhaps.

facutuesca:
> Either all variant or no-variant perhaps.

how would no-variant work?

alex:
Would be fine for cases where all the variants have different types, and therefore don't need anything to disambiguate them.

facutuesca:
> Would be fine for cases where all the variants have different types, and therefore don't need anything to disambiguate them.

Done, I also updated the PR description with the changes.

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -41,6 +41,30 @@
...
+        )
+
+    Usage:
+        example = Example(foo=Variant(5, "IntA"))
+        decoded.foo.value  # The int value
+        decoded.foo.tag    # "IntA" or "IntB"
+    """
+
+    value: U
+    tag: str
```
alex: Does `Tag` either need to be annotated to indicate its a `Literal[str]`, or should we just allow arbitrary tag types?

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -150,10 +174,52 @@ def _normalize_field_type(
...
+                for v in variants
+            ):
+                raise TypeError(
+                    "When using `asn1.Variant` in a union, all the other "
+                    "types in the union must also be `asn1.Variant`"
+                )
+
+            if are_union_types_tagged:
+                tags = [v.tag_name for v in variants]
+                if len(tags) != len(set(tags)):
```
alex:
```suggestion
                tags = {v.tag_name for v in variants}
                if len(variants) != len(tags):
```

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -172,10 +213,10 @@ pub(crate) fn decode_annotated_type<'a>(
...
     // returning the default value)
     if let Some(default) = &ann_type.annotation.get().default {
-        let expected_tag = type_to_tag(inner, encoding);
-        let next_tag = parser.peek_tag();
-        if next_tag != Some(expected_tag) {
-            return Ok(default.clone_ref(py).into_bound(py));
+        let expected_tags = expected_tags_for_type(py, inner, encoding);
+        match parser.peek_tag() {
+            Some(next_tag) if expected_tags.contains(&next_tag) => (),
+            _ => return Ok(default.clone_ref(py).into_bound(py)),
```
alex: this seems bad, in that it's going to result in tons of allocations, can we instead have an `is_tag_valid_for_type` or something?

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -160,6 +161,46 @@ fn decode_bitstring<'a>(
...
+    parser: &mut Parser<'a>,
+    ann_type: &AnnotatedType,
+    encoding: &Encoding,
+) -> ParseResult<pyo3::Bound<'a, pyo3::PyAny>> {
+    match encoding {
+        Encoding::Implicit(_) => Err(CryptographyError::Py(
+            pyo3::exceptions::PyValueError::new_err(
+                "invalid type definition: CHOICE fields cannot be implicitly encoded".to_string(),
+            ),
+        ))?,
```
alex, line 178: WHy is this prohibited? You can have a choice with implicit, in fact you can mix and match on different elements! This happens in general name!

tests/hazmat/asn1/test_serialization.py
```diff
@@ -601,6 +601,236 @@ class Example:
...
 
+    def test_sequence_with_choice(self) -> None:
+        @asn1.sequence
+        @_comparable_dataclass
+        class Example:
+            foo: typing.Union[int, bool, str]
+
+        assert_roundtrips([(Example(foo=9), b"\x30\x03\x02\x01\x09")])
+        assert_roundtrips([(Example(foo=True), b"\x30\x03\x01\x01\xff")])
+        assert_roundtrips([(Example(foo="a"), b"\x30\x03\x0c\x01a")])
```
alex: Why three cals to assert_roundtrips, rather than passing a list to the function.

src/rust/src/declarative_asn1/decode.rs
facutuesca: > WHy is this prohibited? You can have a choice with implicit, in fact you can mix and match on different elements! This happens in general name!

The elements of a CHOICE can be implicit or explicit. The CHOICE itself cannot be IMPLICIT
```
GeneralName ::= CHOICE {....}

# this is not allowed
Example ::= SEQUENCE {
    myField   [0]   IMPLICIT GeneralName
}
```

From [X.680](https://www.itu.int/rec/T-REC-X.680-202102-I/en):

>  31.2.7 The tagging construction specifies explicit tagging if any of the following holds:
>
> a) the "Tag EXPLICIT Type" alternative is used;
> b) the "Tag Type" alternative is used and the value of "TagDefault" for the module is either EXPLICIT
> TAGS or is empty;
> c) the "Tag Type" alternative is used and the value of "TagDefault" for the module is IMPLICIT TAGS or
> AUTOMATIC TAGS, but the type defined by "Type" is an **untagged choice type**, an untagged open type, or
> an untagged "DummyReference" (see Rec. ITU-T X.683 | ISO/IEC 8824-4, 8.3).
> The tagging construction specifies implicit tagging otherwise.
>(...) 
> 31.2.9 The IMPLICIT alternative shall not be used if the type defined by "Type" is an **untagged choice type** or an
untagged open type or an untagged "DummyReference" (see Rec. ITU-T X.683 | ISO/IEC 8824-4, 8.3).

alex: Ah, can you add a comment to that effect please.

tests/hazmat/asn1/test_serialization.py

src/rust/src/declarative_asn1/decode.rs
facutuesca: done

src/rust/src/declarative_asn1/decode.rs
facutuesca: ah good catch, fixed

src/cryptography/hazmat/asn1/asn1.py
facutuesca: the docstring is wrong, fixed

alex: My question was more about the Tag TypeVar.

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -160,6 +161,47 @@ fn decode_bitstring<'a>(
...
+                    Annotation {
+                        default: None,
+                        encoding: None,
+                        size: None,
+                    },
+                )?,
+            };
+            // Parse the TLV data (which contains the field without the EXPLICIT
+            // wrapper)
+            asn1::parse(tlv.full_data(), |d| {
```
alex, line 198: This is `tlv.parse()`

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -246,3 +313,33 @@ pub(crate) fn decode_annotated_type<'a>(
         _ => Ok(decoded),
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::declarative_asn1::types::{AnnotatedType, Annotation, Encoding, Type, Variant};
+    #[test]
+    fn test_decode_implicit_choice() {
```
alex, line 321: Hmm, why do we need a rust test case for this? Not reachable from the PYthon API?

src/rust/src/declarative_asn1/types.rs
```diff
@@ -419,29 +449,80 @@ pub(crate) fn python_class_to_annotated<'p>(
...
+            Encoding::Explicit(n) => asn1::explicit_tag(*n),
+        },
+        None => tag_without_encoding,
     };
+    tag_with_encoding == tag
+}
 
-    match encoding {
+// Utility function to see if a tag matches an unnanotated variant.
+pub(crate) fn is_tag_valid_for_variant(
```
alex, line 470: Can you refactor this out into a seperate PR for ease of review.

src/cryptography/hazmat/asn1/asn1.py
facutuesca: How would you annotate it? `Literal[str]` is not a valid expression

src/rust/src/declarative_asn1/decode.rs
facutuesca: yeah, we fail first at the Python level if someone annotates a `CHOICE` with `IMPLICIT`

src/cryptography/hazmat/asn1/asn1.py
alex: I'm not sure! The answer is maybe that this is all correct -- but I was wondering if the Tag typevar should have a bound of some sort on it?

src/rust/src/declarative_asn1/decode.rs
facutuesca: `tlv.parse()` requires knowing the type at compile time:
```rust
    pub fn parse<T: Asn1Readable<'a>>(&self) -> ParseResult<T> {
```

src/cryptography/hazmat/asn1/asn1.py
facutuesca: We can add a bound of [`LiteralString`](https://typing.python.org/en/latest/spec/literal.html#literalstring). It's not perfect, `str` and any subclass of `str` will pass the type check if passed as a `Tag`, but we can catch those at runtime.

So something like:
```python
Tag = typing.TypeVar("Tag", bound=typing.LiteralString)

@dataclasses.dataclass(frozen=True)
class Variant(typing.Generic[U, Tag]):
```

and we keep the runtime check for `Tag` being a `typing.Literal`.

facutuesca: I pushed a commit with this ^ change

src/rust/src/declarative_asn1/types.rs
facutuesca: done, see https://github.com/pyca/cryptography/pull/14231

tests/hazmat/asn1/test_serialization.py
```diff
@@ -607,6 +677,244 @@ class Example:
...
+            foo: typing.Union[bool, str, None]
+            bar: int
+
+        assert_roundtrips(
+            [(Example(foo=True, bar=1), b"\x30\x06\x01\x01\xff\x02\x01\x01")]
+        )
+
+        assert_roundtrips(
+            [(Example(foo=None, bar=1), b"\x30\x03\x02\x01\x01")]
+        )
```
alex, line 709: ni: merge into single call to assert_roundtrips

facutuesca: fixed!

PR #14203, opened 2026-01-25
Remove aad parameter from HPKE one-shot public API

### LINKED ISSUE #14073: Add HPKE support
We've gotten some feedback that it'd be valuable for us to add HPKE support.

TODO:

- [x] Review RFC, decide what subset (if any) we think makes sense to implemen
- [ ] Design an API
  - [ ] Is this sufficiently misuse resistant that it could be considered a recipe (i.e., not hazmat)?
- [x] Gather test vectors/KATs
- [ ] Implement (including tests + docs)

alex:
Per RFC 9180 Section 8.1, applications using single-shot APIs should use the info parameter for auxiliary authenticated information rather than aad. This change:

- Creates internal _encrypt/_decrypt methods that accept aad for running RFC test vectors
- Removes aad parameter from public encrypt/decrypt methods
- Updates tests to use internal API for vector validation
- Updates documentation to reflect the API change
- Adds CHANGELOG entry for HPKE support

Fixes #14073

### DIFF
--- a/src/rust/src/backend/hpke.rs
+++ b/src/rust/src/backend/hpke.rs
@@ -2,14 +2,14 @@
 // 2.0, and the BSD License. See the LICENSE file in the root of this repository
 // for complete details.
 
+use pyo3::types::{PyAnyMethods, PyBytesMethods};
+
 use crate::backend::aead::AesGcm;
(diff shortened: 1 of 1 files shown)


reaperhulk:
apparently needs a rebase

reaperhulk (dismissed):

PR #14272, opened 2026-02-07
x509-verification: drop cryptography-key-parsing dependency

edef1c:
Move the Pkcs1RsaPublicKey ASN.1 struct into cryptography-x509::common (next to SubjectPublicKeyInfo, where it naturally belongs) so that cryptography-x509-verification no longer depends on cryptography-key-parsing.

This removes the transitive dependency on openssl-sys, making cryptography-x509-verification usable with alternative crypto backends (e.g. aws-lc-rs) without requiring OpenSSL headers or libraries.

### DIFF
--- a/src/rust/cryptography-key-parsing/src/rsa.rs
+++ b/src/rust/cryptography-key-parsing/src/rsa.rs
@@ -4,11 +4,7 @@
 
 use crate::{KeyParsingError, KeyParsingResult, KeySerializationResult};
 
-#[derive(asn1::Asn1Read, asn1::Asn1Write)]
-pub struct Pkcs1RsaPublicKey<'a> {
-    pub n: asn1::BigUint<'a>,
-    pub e: asn1::BigUint<'a>,
-}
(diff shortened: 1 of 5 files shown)


src/rust/cryptography-key-parsing/src/rsa.rs
```diff
@@ -4,11 +4,7 @@
 
 use crate::{KeyParsingError, KeyParsingResult, KeySerializationResult};
 
-#[derive(asn1::Asn1Read, asn1::Asn1Write)]
-pub struct Pkcs1RsaPublicKey<'a> {
-    pub n: asn1::BigUint<'a>,
-    pub e: asn1::BigUint<'a>,
-}
+pub use cryptography_x509::common::Pkcs1RsaPublicKey;
```
alex:
```suggestion
use cryptography_x509::common::Pkcs1RsaPublicKey;
```

no need to re-export this

reaperhulk:
One step closer to publishing these crates separately 😎

PR #14286, opened 2026-02-10
Update extremely outdated structures to tolerate relevant versions of CMS and S/MIME

TaaviE:
This PR updates relevant ASN.1 structures to tolerate S/MIME above version 2 (so newer than 2004) and the respective CMS version. All the structure changes are backwards compatible if none of the "new" features are used.

In general it should now tolerate RFC 5652 (CMS "3") for RFC 8551 (S/MIME 4.0) and RFC 5940 (Additional CMS Revocation Information Choices) options. Most notably `RevocationInfoChoice` in `SignedData` `crls`, `CertificateChoices` in `ContentInfo` `certificates` and `SignerIdentifier` in `SignerInfo` `issuer_and_serial_number`.

~~Few very minor cosmetic data type changes to OCSP.~~ I did try to turn `OCSPResponseStatus` and `CRLReason` into nice `enum`s, but I couldn't win the fight against `asn1` and the tests. (It could be that the test cases contain something odd.)

In general the PR just fixes decoding/encoding but adds no new features nor affects how older CMS/PKCS#7 is interpreted (none of the tests have been modified, but they all pass).

### DIFF
--- a/src/rust/cryptography-x509/src/pkcs7.rs
+++ b/src/rust/cryptography-x509/src/pkcs7.rs
@@ -39,21 +39,10 @@ pub struct SignedData<'a> {
     >,
     pub content_info: ContentInfo<'a>,
     #[implicit(0)]
-    pub certificates: Option<
-        common::Asn1ReadableOrWritable<
-            asn1::SetOf<'a, certificate::Certificate<'a>>,
-            asn1::SetOfWriter<'a, certificate::Certificate<'a>>,
-        >,
-    >,
+    pub certificates: Option<CertificateSet<'a>>,
 
-    // We don't ever supply any of these, so for now, don't fill out the fields.
     #[implicit(1)]
-    pub crls: Option<
-        common::Asn1ReadableOrWritable<
-            asn1::SetOf<'a, asn1::Sequence<'a>>,
-            asn1::SetOfWriter<'a, asn1::Sequence<'a>>,
-        >,
-    >,
+    pub crls: Option<RevocationInfoChoices<'a>>,
 
     pub signer_infos: common::Asn1ReadableOrWritable<
         asn1::SetOf<'a, SignerInfo<'a>>,
@@ -61,10 +50,53 @@ pub struct SignedData<'a> {
     >,
 }
 
+pub type CertificateSet<'a> = common::Asn1ReadableOrWritable<
+    asn1::SetOf<'a, CertificateChoices<'a>>,
+    asn1::SetOfWriter<'a, CertificateChoices<'a>, Vec<CertificateChoices<'a>>>,
+>;
+
+pub type RevocationInfoChoices<'a> = common::Asn1ReadableOrWritable<
+    asn1::SetOf<'a, RevocationInfoChoice<'a>>,
+    asn1::SetOfWriter<'a, RevocationInfoChoice<'a>, Vec<RevocationInfoChoice<'a>>>,
+>;
+
+#[allow(clippy::large_enum_variant)]
+#[derive(asn1::Asn1Write, asn1::Asn1Read)]
+pub enum CertificateChoices<'a> {
+    Certificate(certificate::Certificate<'a>),
+    #[implicit(0)]
+    ExtendedCertificate(asn1::Sequence<'a>),
+    #[implicit(1)]
+    V1AttrCert(asn1::Sequence<'a>),
+    #[implicit(2)]
+    V2AttrCert(asn1::Sequence<'a>),
+    #[implicit(3)]
+    OtherCertificate(OtherCertificateFormat<'a>),
+}
+
+#[derive(asn1::Asn1Write, asn1::Asn1Read)]
+pub struct OtherCertificateFormat<'a> {
+    pub other_cert_format: asn1::ObjectIdentifier,
+    pub other_cert: asn1::Tlv<'a>,
+}
+
+#[derive(asn1::Asn1Write, asn1::Asn1Read)]
+pub enum RevocationInfoChoice<'a> {
+    Crl(asn1::Sequence<'a>),
+    #[implicit(1)]
+    Other(OtherRevocationInfoFormat<'a>),
+}
+
+#[derive(asn1::Asn1Write, asn1::Asn1Read)]
+pub struct OtherRevocationInfoFormat<'a> {
+    pub other_rev_info_format: asn1::ObjectIdentifier,
+    pub other_rev_info: asn1::Tlv<'a>,
+}
+
 #[derive(asn1::Asn1Write, asn1::Asn1Read)]
 pub struct SignerInfo<'a> {
     pub version: u8,
-    pub issuer_and_serial_number: IssuerAndSerialNumber<'a>,
+    pub issuer_and_serial_number: SignerIdentifier<'a>,
     pub digest_algorithm: common::AlgorithmIdentifier<'a>,
     #[implicit(0)]
     pub authenticated_attributes: Option<csr::Attributes<'a>>,
@@ -74,6 +106,13 @@ pub struct SignerInfo<'a> {
 
     #[implicit(1)]
     pub unauthenticated_attributes: Option<csr::Attributes<'a>>,
+}
+
+#[derive(asn1::Asn1Write, asn1::Asn1Read)]
+pub enum SignerIdentifier<'a> {
+    IssuerAndSerialNumber(IssuerAndSerialNumber<'a>),
+    #[implicit(0)]
+    SubjectKeyIdentifier(&'a [u8]),
 }
 
 #[derive(asn1::Asn1Write, asn1::Asn1Read)]
--- a/src/rust/src/pkcs7.rs
+++ b/src/rust/src/pkcs7.rs
@@ -61,7 +61,7 @@ fn serialize_certificates<'p>(
 
     let raw_certs = py_certs
         .iter()
-        .map(|c| c.raw.borrow_dependent().clone())
+        .map(|c| pkcs7::CertificateChoices::Certificate(c.raw.borrow_dependent().clone()))
         .collect::<Vec<_>>();
 
     let signed_data = pkcs7::SignedData {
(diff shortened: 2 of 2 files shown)


alex:
We _strongly_ encourage contributors to break out separate parts of their changes. In particularly, cosmetic changes to OCSP should be split out from this, it makes it harder to review any PKCS#7 changes.

TaaviE:
The cosmetic OCSP change is really just two lines, but I can move it to a separate commit.

TaaviE:
@alex I've split the code up into separate commits for ease-of-review, I hope it's sufficient.

alex:
We prefer things split into seperate PRs so that they can be merged and thus dispensed with individually. (If Github supported stacked diffs, this would be much easier for everyone!)

The refactors in src/rust/cryptography-x509/src/ocsp_resp.rs appear orthagonal to the purpose of this, unless I'm misunderstanding.

TaaviE:
Well, no. Proper handling of RevocationInfoChoices also means proper OCSP structures (because OtherRevocationFormat can contain it). It's just part of reaching a more recent CMS and S/MIME baseline.

As changes to ocsp_resp.rs depend on OID changes in oid.rs and affect crs.rs, it would also be rather tedious to separate them further and then deal with rebasing and whatnot.

src/rust/cryptography-x509/src/ocsp_resp.rs
alex, line 35: This seems like a perfectly fine cleanup, but it's orthagonal to the purpose of the PR as I understand it. Please submit as a seperate PR.

src/rust/src/x509/ocsp_resp.rs
```diff
@@ -28,7 +25,7 @@ pub(crate) fn load_der_ocsp_response(
     match response.response_status.value() {
         SUCCESSFUL_RESPONSE => match response.response_bytes {
             Some(ref bytes) => {
-                if bytes.response_type != BASIC_RESPONSE_OID {
+                if *bytes.oid() != oid::OCSP_BASIC_OID {
```
alex: Now that we enforce this on parse, I don't think this check is even required? Indeed, coverage tests are failing precisely because this is now dead code.

(This is why we insist on splitting things out.)

alex (commented):
Left some comments to explain our position on the OCSP changes.

My bigger question is: what is the compataibility goal for this? It adds support for a number of new OIDs and types, but ultimately mostly for the purposes of ignoring those values. We need to understand better what the goal is here (and what products this adds compatibility for). Finally, we're going to need a lot more tests -- these are how we ensure things don't regress.

src/rust/src/x509/ocsp_resp.rs
TaaviE: Possibly, but I wanted to reduce the amount of changes. There are other places as well that could use a cleanup. Specifically the sections around signature and digest algorithms.

TaaviE:
> My bigger question is: what is the compataibility goal for this? It adds support for a number of new OIDs and types, but ultimately mostly for the purposes of ignoring those values.

> We need to understand better what the goal is here (and what products this adds compatibility for). 

Yes, precisely, they are just to up-to-date CMS and S/MIME, not just something from the 90s that has been obsolete for 27 years. The product that is supported are currently relevant standards (from the 2000s). Things even OpenSSL is able to read. 

Doing anything in even smaller increments would likely make the overall goal (that's already split between two PRs right now) even harder to see - "modern" S/MIME support and RFC compliance. For example I'd take on PQ support next, but `ring` is actually terrible (it has public traits that can't be implemented) and it's very poorly documented. It would have to be replaced with something more sane and that's a decision you'd have to reach first.

> Finally, we're going to need a lot more tests -- these are how we ensure things don't regress.

It wouldn't be too difficult to add a few examples for code coverage. However a proper test suite would rely on all the relevant structures and code paths being up-to-date (this probably should include S/MIME BR policies being supported).

reaperhulk:
"Up to date standards that aren't 27 years old" is not a reason for us to merge something. The reason we ask about what product/capability this enables is because CMS and SMIME adjacent things are an absolutely enormous spec, the vast majority of which is not actually used by anything. We are interested in expanding support to fit use cases that make sense, but this project has learned 13 years of hard lessons that "support it because it exists" is a bad path.

You can help us learn what matters here by splitting pieces of this PR off into smaller, highly focused pieces. Not everything can be done this way, but much of it can, and in doing so you can also add test vectors that show why expanded parsing is valuable/useful (for example, because a major vendor encodes something in that form).

Rebasing like this definitely increases work on your end and will result in some annoying de-duplication as you go, but it's going to be necessary for us to help you get this merged.

In the long term, as @alex noted, PR stacks will make this radically simpler for contributors (and GH did recently announce their intent to support them: https://github.com/github/roadmap/issues/1218), but for now this is what we have.

TaaviE:
> "Up to date standards that aren't 27 years old" is not a reason for us to merge something. The reason we ask about what product/capability this enables is because CMS and SMIME adjacent things are an absolutely enormous spec, the vast majority of which is not actually used by anything.

Most importantly, implementations relying on pyca/cryptography not crashing when they're interpreting things incorrectly in the wild. This PR intentionally refrains from using any of the new data or introducing any changes to the interpretation of existing structures.

It just ensures proper behaviour when something does follow the RFC and does use the options given. I dread seeing the other PRs adding "S/MIME" support when it's actually S/MIME 2 from the 90s. Some of structures have contained a `CHOICE` for at least two decades, whether or not the data is used, not handling it is incorrect. In its current state it can be called buggy.

Functionality-wise, being able to "staple" an OCSP response to a CMS structure is useful, being able to use SKI for finding the signer is useful. That's why they were added in the first place.

The changes actually are already split up, I have a proper S/MIME implementation already, these actually are the small pieces.

alex:
> The changes actually are already split up, I have a proper S/MIME implementation already, these actually are the small pieces.

This is manifestly not true. I've already flagged a piece that is separable, and the reason CI is failing (which will require more changes to solve, which compounds the motivation for separating it out).

You're not obligated to do any of this work, but it is a pre-requisite for it being mergable.

TaaviE:
> This is manifestly not true. I've already flagged a piece that is separable, and the reason CI is failing (which will require more changes to solve, which compounds the motivation for separating it out).

I will put that one commit in an another PR when I get the chance. Albeit it still makes more sense as a cohesive whole in the current set.

TaaviE:
The OSCP PR was merged, is there anything else that should be split up to an another PR?

src/rust/cryptography-x509/src/ocsp_resp.rs
```diff
@@ -6,11 +6,13 @@ use crate::{certificate, common, crl, extensions, name, ocsp_req};
 
 #[derive(asn1::Asn1Read, asn1::Asn1Write)]
 pub struct OCSPResponse<'a> {
-    pub response_status: asn1::Enumerated,
+    pub response_status: OCSPResponseStatus,
     #[explicit(0)]
     pub response_bytes: Option<ResponseBytes<'a>>,
 }
 
+pub type OCSPResponseStatus = asn1::Enumerated;
```
alex, line 14: A type alias for this could split out.

src/rust/cryptography-x509/src/pkcs7.rs
```diff
@@ -39,32 +39,64 @@ pub struct SignedData<'a> {
...
 
+pub type CertificateSet<'a> = common::Asn1ReadableOrWritable<
+    asn1::SetOf<'a, CertificateChoices<'a>>,
+    asn1::SetOfWriter<'a, CertificateChoices<'a>, Vec<CertificateChoices<'a>>>,
+>;
+
+pub type RevocationInfoChoices<'a> = common::Asn1ReadableOrWritable<
+    asn1::SetOf<'a, RevocationInfoChoice<'a>>,
+    asn1::SetOfWriter<'a, RevocationInfoChoice<'a>, Vec<RevocationInfoChoice<'a>>>,
+>;
```
alex, line 61: As could type aliases .

src/rust/cryptography-x509/src/ocsp_resp.rs
TaaviE: Split here: https://github.com/pyca/cryptography/pull/14528

src/rust/cryptography-x509/src/pkcs7.rs
TaaviE: Split here: https://github.com/pyca/cryptography/pull/14528

TaaviE: Though this PR does depend on this definition

TaaviE:
Beyond `pub type OCSPResponseStatus = asn1::Enumerated;` is there actually anything that could be split to an another PR that could work out alone?

In general these changes aren't very big, there aren't any functional changes. All it does is make pyca/cryptography tolerate up-to-date CMS structures used for up-to-date S/MIME.

These are also just a subset, there are other (crash) bugs with obsolete defined structures ☹️

alex:
As we said in the beginning "just tolerate different structures" is not an API goal. You need to articulate what we're trying to support, why, and have tests to ensure we don't regress them and behave correctly.

TaaviE:
The goal is to tolerate and implement correct structures following S/MIME 3+, at least when decoding, is it not?

There are code comments (for example under encryption) that try to imply that it follows S/MIME 3.2, but factually it does not. Look at those structures, the current definitions can't handle an ECDSA recipient even when decoding because the structures are so old.

TaaviE:
So when the [documentation currently says](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/#cryptography.hazmat.primitives.serialization.pkcs7.pkcs7_decrypt_smime) "Deserialize and decrypt a S/MIME-encoded PKCS7 message. PKCS7 (or S/MIME) has multiple versions, but this supports a subset of [RFC 5751](https://datatracker.ietf.org/doc/html/rfc5751.html), also known as S/MIME Version 3.2." it should actually say "[...] this supports RFC 2311 (1998) also known as S/MIME 2.0" 

I can also make the documentation update.

PR #14287, opened 2026-02-10
EC check key on cofactor > 1

reaperhulk:


src/rust/src/backend/ec.rs
```diff
@@ -367,6 +362,33 @@ impl ECPrivateKey {
     }
 }
 
+fn cofactor_check(group: &openssl::ec::EcGroupRef) -> CryptographyResult<bool> {
```
alex: this is a terrible name tbh. I don't see why you wouldn't just inline this into the caller either?

src/rust/src/backend/ec.rs
```diff
@@ -367,6 +362,33 @@ impl ECPrivateKey {
...
+    let mut bn_ctx = openssl::bn::BigNumContext::new()?;
+    let mut cofactor = openssl::bn::BigNum::new()?;
+    group.cofactor(&mut cofactor, &mut bn_ctx)?;
+    let one = openssl::bn::BigNum::from_u32(1)?;
+    Ok(cofactor.ucmp(&one) == std::cmp::Ordering::Greater)
```
alex: you can just do `cofactor == one`

CHANGELOG.rst
```diff
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+.. _v46-0-5:
+
+46.0.5 - 2026-02-10
+~~~~~~~~~~~~~~~~~~~
+
+* Support for ``SECT*`` elliptic curves is deprecated and will be removed in
+  the next release.
```
alex: We need to document the fix here.

alex (approved):

PR #14288, opened 2026-02-10
Expose set_groups/set_curves functions

schwabe:
This allows pyOpenSSL to restrict the groups allowed to be used. E.g. for restricting the groups to post-quantum hybrid groups (e.g. X25519MLKEM768) to always ensure that post-quantum cryptography is used.

This commit uses the set_curves instead of the set_groups variant since that naming is available on all OpenSSL versions and forks.

### DIFF
--- a/src/_cffi_src/openssl/ssl.py
+++ b/src/_cffi_src/openssl/ssl.py
@@ -346,6 +346,13 @@ int SSL_version(const SSL *);
 
 const char *SSL_get0_group_name(SSL *);
 
+/* both functions are have int return type according to the man page but
+ * long in its implementation. Newer versions also have the set1_groups
+ * function but the curves variants are available on all forks and OpenSSL
+ * variants */
+long SSL_CTX_set1_curves_list(SSL_CTX *, char *);
+long SSL_set1_curves_list(SSL *, char *);
+
 void SSL_set_tlsext_host_name(SSL *, char *);
 void SSL_CTX_set_tlsext_servername_callback(
     SSL_CTX *,


alex:
Before we land new CFFI APIs in cryptography, we prefer to have a PoC up for pyOpenSSL, just to make sure the APIs we're adding support a reasonable design in pyOpenSSL.

schwabe:
Sure, the PR for pyOpenSSL is here: https://github.com/pyca/pyopenssl/pull/1477

I still struggle a bit to figure out how to tell the CI for pyOpenSSL/Github/nox to tell that it needs to build cryptography from this PR to make the tests work. Installing manually the cryptography version in the venv make the tests work at least on OpenSSL.

alex:
That's fine, the goal is demonstrate that this can work (and that we'll actually getting a pyOpenSSL PR out of it).

Unfortunately for a while we had a lot of folks send PRs to add new bindings, but never follow up with the pyOpenSSL PRs.

src/_cffi_src/openssl/ssl.py
```diff
@@ -346,6 +346,13 @@
 
 const char *SSL_get0_group_name(SSL *);
 
+/* both functions are have int return type according to the man page but
+ * long in its implementation. Newer versions also have the set1_groups
+ * function but the curves variants are availalbe on all forks and OpenSSL
```
alex:
```suggestion
 * function but the curves variants are available on all forks and OpenSSL
```

alex (approved):

PR #14291, opened 2026-02-10
remove binary curve support

reaperhulk:
binary curves are rarely used, contain additional dangers, and may not be supported in some of our alternate backends.

### DIFF
--- a/src/rust/src/backend/ec.rs
+++ b/src/rust/src/backend/ec.rs
@@ -236,9 +236,11 @@ impl ECPrivateKey {
 
         let len = deriver.len()?;
         Ok(pyo3::types::PyBytes::new_with(py, len, |b| {
-            let n = py.detach(|| deriver.derive(b)).map_err(|_| {
-                pyo3::exceptions::PyValueError::new_err("Error computing shared key.")
-            })?;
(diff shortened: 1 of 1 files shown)


src/rust/src/backend/ec.rs
```diff
@@ -371,7 +371,13 @@ impl ECPublicKey {
...
         let ec = pkey.ec_key()?;
         check_key_infinity(&ec)?;
-
+        let mut bn_ctx = openssl::bn::BigNumContext::new()?;
+        let mut cofactor = openssl::bn::BigNum::new()?;
+        ec.group().cofactor(&mut cofactor, &mut bn_ctx)?;
+        let one = openssl::bn::BigNum::from_u32(1)?;
+        // We only support curves with a cofactor of 1.
+        // Any change here requires more careful key checking
+        assert!(cofactor == one);
```
alex: nit: assert_eq

alex (dismissed):

alex:
I think you should be able to rebase this now!

alex:
src/rust/src/backend/ec.rs line 240 is apparently uncovered now

PR #14292, opened 2026-02-10
asn1: Add support for decoding TLVs

facutuesca:
This adds support for decoding TLVs

```python
@asn1.sequence
class Example:
    foo: Annotated[asn1.Tlv, asn1.Explicit(0)]
    bar: Annotated[asn1.Tlv, asn1.Explicit(1)]


encoded = b"\x30\x0a\xa0\x03\x02\x01\x08\xa1\x03\x02\x01\x09"
decoded = asn1.decode_der(Example, encoded)

assert isinstance(decoded.foo, asn1.Tlv)
assert decoded.foo.tag_bytes == b"\x02"
assert decoded.bar.data == b"\x09"
```

Once we have a `Tlv` object, we can parse it if we know its type:

```python
decoded_tlv = asn1.decode_der(asn1.Tlv, b"\x30\x03\x02\x01\x09")
assert isinstance(decoded_tlv, asn1.Tlv)

@asn1.sequence
class Example:
    foo: int

decoded_example = decoded_tlv.parse(Example)
assert isinstance(decoded_example, Example)
assert decoded_example.foo == 9
```

## Implementation details
* Only `EXPLICIT` annotations are supported. `DEFAULT`, `IMPLICIT` and `OPTIONAL` TLVs are not supported.
* I implemented the Rust struct so that it stores the full bytes of the TLV, so that implementing `tlv.parse()` is easy (we can just parse `full_data`, instead of reconstructing it from the other fields). There might be a better way of doing this though.
```rust
pub struct Tlv {
    #[pyo3(get)]
    pub tag_bytes: pyo3::Py<pyo3::types::PyBytes>,

    // We store the bytes of the entire TLV, and to access the Value part
    // we store the index where it starts.
    pub data_index: usize,
    pub full_data: pyo3::Py<pyo3::types::PyBytes>,
}
````

### CODE REVIEW

**Comment by alex** [comment]
Is there a reason to have a length attribute? It's just `len(tlv.data)`, right?

**Comment by facutuesca** [comment]
> Is there a reason to have a length attribute? It's just `len(tlv.data)`, right?

true, removed!

File: src/cryptography/hazmat/asn1/asn1.py [hunk 1]
```diff
@@ -365,3 +383,4 @@ class Default(typing.Generic[U]):
 UtcTime = declarative_asn1.UtcTime
 GeneralizedTime = declarative_asn1.GeneralizedTime
 BitString = declarative_asn1.BitString
+Tlv = declarative_asn1.Tlv
```
alex: In Python, this type should idiomatically be TLV, not the rust style Tlv. (This also applies to UTCTime, sorry I missed that before.)

src/cryptography/hazmat/bindings/_rust/declarative_asn1.pyi
```diff
@@ -103,3 +103,10 @@ class BitString:
     def __eq__(self, other: object) -> bool: ...
     def as_bytes(self) -> bytes: ...
     def padding_bits(self) -> int: ...
+
+class Tlv:
+    @property
+    def tag(self) -> int: ...
```
alex: We should think for a second about if int makes sense here, or if we should have a Tag type.

What value does an integer tag have for a >1-byte tag with (say) a constructed flag set?

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -161,6 +161,34 @@ fn decode_bitstring<'a>(
...
+    parser: &mut Parser<'a>,
+    encoding: &Option<pyo3::Py<Encoding>>,
+) -> ParseResult<pyo3::Bound<'a, Tlv>> {
+    let tlv = match encoding {
+        Some(e) => match e.get() {
+            Encoding::Implicit(_) => Err(CryptographyError::Py(
+                // We don't support IMPLICIT TLV
+                pyo3::exceptions::PyValueError::new_err(
+                    "invalid type definition: TLV/ANY fields cannot be implicitly encoded"
+                        .to_string(),
```
alex: You don't actually need to call to_string here, AFAIK.

src/rust/src/declarative_asn1/types.rs
```diff
@@ -165,6 +167,33 @@ impl Variant {
...
+        py: pyo3::Python<'p>,
+        class: &pyo3::Bound<'p, pyo3::types::PyType>,
+    ) -> pyo3::PyResult<pyo3::Bound<'p, pyo3::PyAny>> {
+        crate::declarative_asn1::asn1::decode_der(py, class, self.full_data.as_bytes(py))
+    }
+
+    #[getter]
+    pub fn data<'p>(&self, py: pyo3::Python<'p>) -> pyo3::Bound<'p, pyo3::types::PyBytes> {
+        pyo3::types::PyBytes::new(py, &self.full_data.as_bytes(py)[self.data_index..])
+    }
```
alex, line 204: It seems unfortunate that every time you access data, you get an allocation and copy. Should we be returning a memory view or something?

facutuesca: is there a way of doing that if we're slicing the bytes before we return them? The pyo3 API for memory view only supports initializing from an existing Python object: https://docs.rs/pyo3/latest/pyo3/types/struct.PyMemoryView.html#method.from

src/rust/src/declarative_asn1/decode.rs

src/cryptography/hazmat/asn1/asn1.py

src/rust/src/declarative_asn1/types.rs
alex: Can we just pass the bytes object and then slice the memoryview?

facutuesca: ah ofc, fixed!

src/cryptography/hazmat/bindings/_rust/declarative_asn1.pyi
facutuesca: fixing this is blocked on https://github.com/alex/rust-asn1/pull/596

facutuesca:
Rebased, still blocked on making the change to `rust-asn1` to get the encoded bytes of a `Tag`

src/cryptography/hazmat/bindings/_rust/declarative_asn1.pyi
facutuesca: fixed, now we have `def tag_bytes(self) -> bytes` instead

facutuesca:
Bumped `rust-asn1` and now we expose the TLV tag with a `tag_bytes` property

PR #14319, opened 2026-02-15
Fix installing stray files into site-packages

mgorny:
Fix the `include` pattern in `pyproject.toml` not to install stray files such as `CHANGELOG.rst`, `CONTRIBUTING.rst`, `docs` and `tests` straight into site-packages.  Apparently Maturin did not install them before due to a bug, but it was fixed in maturin 1.12.0, leading to the files being suddenly installed.

Originally reported as https://bugs.gentoo.org/970090.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,10 +106,10 @@ module-name = "cryptography.hazmat.bindings._rust"
 locked = true
 sdist-generator = "git"
 include = [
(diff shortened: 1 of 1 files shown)


reaperhulk:
Thanks @mgorny!

mgorny:
Thanks!

PR #14320, opened 2026-02-15
More misc. zizmor fixes

woodruffw:
Fixes a handful of default-persona zizmor findings. The main ones here are template injections, which aren't exploitable in practice (since the only calls to the action don't allow those inputs to vary at the moment).

I've also added cooldowns to the cargo and uv dep groups in Dependabot, but these could be removed (and optionally suppressed) if you prefer not to have cooldowns on these. (The cooldowns don't affect security updates, only normal bumps.)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,8 +24,6 @@ updates:
       # Also update indirect dependencies
       - dependency-type: all
     open-pull-requests-limit: 1024
-    cooldown:
-      default-days: 7
 
   - package-ecosystem: uv
     directories:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -37,3 +39,5 @@ updates:
       # Also update indirect dependencies
       - dependency-type: all
     open-pull-requests-limit: 1024
+    cooldown:
+      default-days: 7
```
alex: I'm not sure we want these. We review the diffs for any dep bumps, so I'm not sure this is removing risk for us.

woodruffw: Cool, removed.

alex (approved):

PR #14327, opened 2026-02-16
Include OpenSSL SBOMs in built wheels

alex:
Pass --sbom-include to maturin via build-args config setting to include the OpenSSL CycloneDX SBOM (generated by pyca/infra) in the wheel's .dist-info/sboms directory for all platforms (Linux, macOS, Windows).


alex:
https://github.com/PyO3/maturin/issues/3003

reaperhulk:
Does maturin fail if the path doesn’t exist? I want to be sure we don’t silent fail including this in the future.

alex:
Yeah, it will.

PR #14216, opened 2026-01-27
drop building and testing against windows arm64

reaperhulk:
When we originally added win arm64 wheels part of our decision was that we would remove it if the underlying CI infrastructure proved problematic. Unfortunately, that time has already arrived. We need to do a release today for an OpenSSL security release and the arm64 windows image has been [broken with Python 3.14t for over two weeks](
https://github.com/actions/runner-images/issues/14058). Working around issues like this in our CI is not desirable or acceptable. Accordingly, we are removing wheels effective with our next point release.

Our apologies to the users of these wheels, but the most effective thing you can do is advocate with GitHub to have windows arm64 treated like a first class citizen in Actions.

PR #14346, opened 2026-02-19
[CI][Wheels] Make the win_arm64 platform silent on failures

khmyznikov:
- Reverting the PR https://github.com/pyca/cryptography/pull/14216
- added continue-on-error and allowed-failures for windows-arm64.

In result, failures on windows-arm64 side will not disturb other platforms release.


reaperhulk:
> First of all, the GitHub team is committed to bringing the Windows 11 ARM image under its ownership

This is great news. We will be much more interested in merging this and resuming win_arm64 support once this transition has occurred!

reaperhulk:
Just for reference as we track this: Our desired end state would be being able to treat the WoA builders exactly as we do x86_64 Windows. Bringing it back in-house and making sure there aren't basic image differences/broken core actions will be enormously helpful.

alex:
Thanks for working on this. Just to hone in on what Paul said, we really don't want to carry a big "if arm64" block :-) And ideally we wouldn't have to carry that much windows specific code (if PyManager is the right thing, maybe setup-python should use it?)

khmyznikov:
@alex Yeah, I don't like this big chunk neither. I've asked maintainer if we can have `pymanager` included in the image, that should trim down this fix a lot.

khmyznikov:
I have good news I can't share yet, but just want to let you know we are not abandoned this and working on full resolution.

alex:
@khmyznikov FYI, it looks like there's been another regression in win-arm64 + setup-python -- see https://github.com/pyca/bcrypt/pull/1193. I haven't fully diagnosed the cause.

I do see that the image has moved from partner images to the primary image repo, which is an encouraging signal. But to be very direct: repeated breakages like this are precisely the concern we had that led to us not re-adding arm64 to our CI after the original issue was fixed.

khmyznikov:
@alex Thank you for keeping your finger on the pulse! As a first step, we expect the image to be moved to the main repository very soon. I’m discussing with the team how we can best publicly demonstrate the readiness of new versions. I would be happy to hear any suggestions regarding CI testing.

alex:
I don't know a ton about what y'all already do, so there's a big risk of suggesting things that y'all already do. That said, here are some things I suspect are not currently done:

- Smoke tests with live actions. When a new image is ready to be pushed, deploy it with some internal `ubuntu-22.04-secret-next` or whatever label, and then run some real jobs. You can build up a set of workflows that excersise common things: test every supported python version with `setup-python` on the new image, for example. If those all pass, then it can move on to next gates.
- This is harder, but some sort of shadow/canary builds: pick some real repos+workflows, and when they're triggered, add shadow versions of jobs using the new image and run them. If the new image fails and the old one passed, that's a potential indicator of problems.

reaperhulk:
Let me add that working rapidly to fix major breakage is also critical. I say this because the basic and total breakage of setup-python for a subset of versions on windows arm64, which we noted 2 days ago, is not fixed and we have no idea when it will be. That's so far beyond unacceptable as to border on farce. Since we've had our entire CI blocked in both PyNaCl and bcrypt due to this we'll be dropping Windows ARM support there today as well.

khmyznikov:
@reaperhulk please hold. GitHub team reproduces the issue and willing to address it ASAP. The issues was they didn't tested the latest package of the setup python, which was resolved and now will be included into test matrix.

paveliak:
@reaperhulk Sorry for the troubles you are experiencing with Python on ARM. We are investigating this problem. The workaround is to uninstall Python version that comes cached in the image
```
    - run: |
        Start-Process -FilePath 'c:\hostedtoolcache\windows\Python\3.14.4\arm64\python-3.14.4-arm64.exe' -ArgumentList @('/uninstall', '/quiet') -Wait
        Remove-Item -Recurse -Force 'c:\hostedtoolcache\windows\Python\3.14.4'
        
    - uses: actions/setup-python@v6.2.0
      with:
        python-version: '3.14t'
```

In future please open an issue in https://github.com/actions/runner-images and https://github.com/actions/setup-python 🙏

alex:
To be clear, we're simply dropping Windows+ARM64 from all projects we maintain until we have confidence it will be stable. We're not going to carry workarounds for this.

reaperhulk:
We've removed support across all our projects so this is no longer an issue for us. It has become clear over the past months that the set of repositories, processes, and procedures that encompass the Windows Arm64 GitHub Actions hosted runner feature lacks basic testing rigor and the organization is not capable of quickly or efficiently addressing those deficiencies. This is not a commentary on any individual contributor, as I'm sure everyone is trying hard within the structural constraints imposed by your employer, but the outcome is that we are unwilling to continue supporting it.

As before, we'll revisit at some point in the future, but for now no PyCA project will support Windows Arm64 wheels.

paveliak:
@alex @reaperhulk This issue should be fixed. Please take a look and let us know if you still observe any problems. Again sorry for all the hassle 🙏

reaperhulk:
I'm glad you've resolved the problem, but we won't be revisiting adding wheels until we see a long term track record of reliability. It takes relatively little time to destroy trust and much more to rebuild it.

khmyznikov:
@lazka we are working with Paul and Alex to get back the support as "zero-commitment" aka "test period". While Core Collective will monitor the actual status of the WoA pipelines, contributors will not be distracted if platform get back to sudden failure once again. Feel free to help with the review.

alex:
So, this PR currently contains both re-introducing ARM64 _and_ substantial refactors of our windows wheel building and tests. I would prefer we handle these as seperate matters. If refactoring is needed to enable re-adding ARM64, let's do the refactors in seperate PRs which significantly aids review.

khmyznikov:
@alex it's mostly separation rather than refactoring, but I'm cool with splitting 👍

khmyznikov:
@alex yeah it def cleaner now 😊

.github/workflows/ci.yml
```diff
@@ -360,6 +360,34 @@ jobs:
 
       - uses: ./.github/actions/upload-coverage
 
+  windows-arm64:
```
alex: Is it practical to just have thse be an additional entry in the windows matrix, rather than a seperate step?

docs/installation.rst
```diff
@@ -32,6 +32,7 @@ operating systems.
 * x86-64 Debian Bookworm (12.x), Trixie (13.x), and Sid (unstable)
 * x86-64 and ARM64 Alpine (latest)
 * 64-bit Python on 64-bit Windows Server 2022
+* ARM64 Windows 11
```
alex: We don't want to document this at all until it's stable.

.github/workflows/ci.yml
khmyznikov: Hmm that's a good idea.

.github/workflows/wheel-builder.yml
```diff
@@ -318,7 +327,9 @@ jobs:
             .github/actions/wheel-smoketest/
           sparse-checkout-cone-mode: false
 
-      - uses: ./.github/actions/windows-wheel
+      - id: build
+        continue-on-error: ${{ matrix.WINDOWS.ARCH == 'arm64' }}
+        uses: ./.github/actions/windows-wheel
```
alex, line 332: We need to not upload the wheel here, even if it builds successfully on arm64.

khmyznikov: Yeah that's what I wanted to do as well, let me verify

khmyznikov: @alex do you want to skip the arm64 wheels upload completely? Despite of successful build and test?

alex: Correct. All wheels that get produced here are published to PyPI (when run on a tag), and we don't want to publish arm64 wheels until we're committed to keeping them working.

khmyznikov: @alex maybe we can gate the pypi-publish for WoA based on CI tests? The stability of the wheel itself was not under any concerns, right? Problem was the build tools flakiness. If both build and tests are successful, why not pushing it for the people? Prev successful versions are there on pypi.

alex: Because people will come to rely on it and then complain when it goes away if it breaks again.

khmyznikov: Cool, gated it here: https://github.com/pyca/cryptography/pull/14346/changes/1f49c06a9c358e8418571e47910c0d21fa0de717

alex (approved):

khmyznikov:
@alex @reaperhulk thanks for considering the trial. I'll monitor how it's going.

alex:
Great. Thanks for working through this with us.

PR #14353, opened 2026-02-20
Add MLDSA Test vectors

DarkaMaul:
Following https://github.com/pyca/cryptography/issues/14337, this initial PR just adds the test vectors to the repo.
They are not consumed by any tests yet.

The source is:

- https://raw.githubusercontent.com/usnistgov/ACVP-Server/master/gen-val/json-files/ML-DSA-keyGen-FIPS204/internalProjection.json
- https://raw.githubusercontent.com/usnistgov/ACVP-Server/master/gen-val/json-files/ML-DSA-sigGen-FIPS204/internalProjection.json
- https://raw.githubusercontent.com/usnistgov/ACVP-Server/master/gen-val/json-files/ML-DSA-sigVer-FIPS204/internalProjection.json

I filtered out:
- Non MLDSA65 test vectors
- Pre-Hash vectors
- Non-deterministic ones

The vector files are kept _as-is_ from upstream, including ACVP metadata fields we don't use (e.g., `vsId`, `deferred`). This keeps the format close to the original so we can easily regenerate (e.g. redownload) them.

I'm happy to massage the files into a more compact format tailored for this project instead if that's preferred.


alex:
I just skimmed the tests, but it looks like these all use the expanded key format, and not the seed based format. Does that match your understanding?

We intend to only support seed based keys, so if these are expanded form unfortunately they don't work for us :-(

DarkaMaul:
> I just skimmed the tests, but it looks like these all use the expanded key format, and not the seed based format. Does that match your understanding?
> 
> We intend to only support seed based keys, so if these are expanded form unfortunately they don't work for us :-(

🤦 The seeds are only included for the `keyGen` tests - not the two others.

I've updated the PR with the one from [KAT](https://github.com/post-quantum-cryptography/KAT) by @kriskwiatkowski

While they are non-official, they have been tested by the broader community ([thread](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/63upUY9hYPI)).
They contain 100 instances of a `(seed, pk, sk, msg, sm)`.

[wycheproof](https://github.com/C2SP/wycheproof/blob/main/testvectors_v1/mldsa_65_sign_seed_test.json) also has test vectors, including invalid test cases. They can be used in addition to the other ones to ensure better coverage.

abbra:
@DarkaMaul as I mentioned in the issue https://github.com/pyca/cryptography/issues/14337, I already have wycheproof support in my ML-DSA branches.

alex:
sorry, I think I caused a merge conflict :-(

these seem like a reasonable starting point for vectors, and we'll of course use wycheproof as well. (normally we don't copy wycheproof vectors in the repo, we pull them in via git, the one exception is if wycheproof is the _onyl_ source of vectors)

facutuesca:
> sorry, I think I caused a merge conflict :-(

conflict fixed!

docs/development/test-vectors.rst
```diff
@@ -64,7 +64,7 @@ Asymmetric ciphers
   encrypted with ``AES-192-CBC``. The password is ``PolarSSLTest``.
 * ``asymmetric/PKCS8/ed25519-scrypt.pem`` a PKCS8 encoded Ed25519 key from
   RustCrypto using scrypt as the KDF. The password is ``hunter42``.
-
+* FIPS 204 ML-DSA-65 KAT vectors from `post-quantum-cryptography/KAT`_.
```
alex: is there a reason to do only ML-DSA-65 and not 87?

facutuesca: The idea was to scope the PRs as [minimally](https://github.com/pyca/cryptography/issues/12824#issuecomment-3779398375) as possible, so only ML-DSA-65 to start with.

> Let's pick MLDSA, strip it down to **a single key size**, and land that. Doing that has this ordering of PRs:
> 
> Vectors for ML-DSA testing
Land impl for boring or aws-lc + docs + tests
Followups for other impls.

facutuesca:
the Windows GHA runner is being flaky

image

docs/development/test-vectors.rst
alex: agreed in general -- let's get all the vectors at once though, that makes it easier

facutuesca: done!

DarkaMaul: Thank you @facutuesca for updating this PR!

PR #14370, opened 2026-02-24
Release the GIL during core EC and RSA cryptographic operations

alex:
Wrap CPU-intensive OpenSSL calls in py.detach() to release the GIL, allowing other Python threads to run during cryptographic operations.

Operations covered:
- EC key generation, ECDH key exchange, ECDSA sign/verify
- RSA key generation, sign/verify, encrypt/decrypt

https://claude.ai/code/session_01TbN4mmRwmt1boDWU8Gwb5E

### DIFF
--- a/src/rust/src/backend/ec.rs
+++ b/src/rust/src/backend/ec.rs
@@ -149,10 +149,8 @@ fn generate_private_key(
     let _ = backend;
 
     let ossl_curve = curve_from_py_curve(py, curve)?;
-    let pkey = py.detach(|| {
-        let key = openssl::ec::EcKey::generate(&ossl_curve)?;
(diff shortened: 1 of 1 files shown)


src/rust/src/backend/ec.rs
```diff
@@ -149,10 +149,13 @@ fn generate_private_key(
     let _ = backend;
 
     let ossl_curve = curve_from_py_curve(py, curve)?;
-    let key = openssl::ec::EcKey::generate(&ossl_curve)?;
+    let pkey = py.detach(|| {
+        let key = openssl::ec::EcKey::generate(&ossl_curve)?;
```
reaperhulk: I suspect this isn't worth doing, EC key generation is **fast**

PR #14390, opened 2026-02-27
HPKE: Convert decap errors to InvalidTag to prevent oracle leaks

alex:
## Summary
This change hardens the HPKE decryption implementation against timing and error-based side-channel attacks by converting all decapsulation errors to `InvalidTag` exceptions, preventing attackers from distinguishing between small-order point rejections and other cryptographic failures.

## Key Changes
- Modified `Suite.decrypt()` in the Rust backend to catch all `decap()` errors and convert them to `InvalidTag` exceptions instead of propagating the original error
- Added comprehensive test coverage with all 8 known small-order points on Curve25519 to verify that decryption with these points raises `InvalidTag` rather than `ValueError`
- This prevents error oracles that could leak information about the validity of ephemeral public keys during decapsulation

## Implementation Details
- The change uses `.map_err()` to transform any decapsulation error into a cryptographic authentication failure
- Test cases cover all small-order points including the zero point, points with various orders (1, 2, 4, 8), and field boundary cases (p, p+1, p-1)
- By treating all decap failures uniformly as authentication failures, the implementation avoids leaking whether a point was rejected due to small order versus other reasons

https://claude.ai/code/session_019QtgYxKBdQahyunYWCxDjb

### DIFF
--- a/tests/hazmat/primitives/test_hpke.py
+++ b/tests/hazmat/primitives/test_hpke.py
@@ -173,6 +173,50 @@ class TestHPKE:
         with pytest.raises(InvalidTag):
             suite.decrypt(b"\x00", sk_r)
 
+    @pytest.mark.parametrize(
+        "small_order_point",
+        [
+            # All 8 known small-order points on Curve25519
+            bytes(32),  # Zero point (order 1)
+            bytes([1] + [0] * 31),  # Order 4
+            bytes.fromhex(
+                "ecffffffffffffffffffffffffffffffffffff"
+                "ffffffffffffffffffffffffffff7f"
+            ),  # Order 8
+            bytes.fromhex(
+                "5f9c95bca3508c24b1d0b1559c83ef5b04445cc"
+                "4581c8e86d8224eddd09f1157"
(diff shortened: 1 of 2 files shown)


tests/hazmat/primitives/test_hpke.py
```diff
@@ -173,6 +173,50 @@ def test_truncated_ciphertext(self):
...
             suite.decrypt(b"\x00", sk_r)
 
+    @pytest.mark.parametrize(
+        "small_order_point",
+        [
+            # All 8 known small-order points on Curve25519
+            bytes(32),  # Zero point (order 1)
+            bytes([1] + [0] * 31),  # Order 4
+            bytes.fromhex(
+                "ecffffffffffffffffffffffffffffffffffff"
```
reaperhulk, line 183: Do we have a canonical source for these low order points? And are there only 8? wycheproof appears to have more.

alex: I don't _super_ thing it matters? x25519 has cofactor 8 though...

PR #14398, opened 2026-03-01
Add DHKEM(P-256, HKDF-SHA256) support to HPKE implementation

alex:
Add P256 KEM variant alongside the existing X25519, following the same abstraction patterns used for KDF and AEAD. KEM-specific operations (key generation, public key serialization/deserialization, DH exchange) are methods on the KEM enum, keeping encap/decap unified with no duplication.

Includes key type validation against Python ABCs (TypeError on mismatch), RFC 9180 test vector validation for kem_id=0x0010, and comprehensive wrong-key/wrong-type/wrong-curve tests.

https://claude.ai/code/session_01FJG426sLhnaeMytWntj9aJ


bitloi:
@alex can I help you?

PR #14430, opened 2026-03-05
ML-DSA-44 support (OpenSSL backend)

abbra:
I'm submitting first of ML-DSA branches for consideration of higher level API. All branches build up on each other, so ML-DSA-65 version is built on top of this one, and ML-DSA-87 is on top of ML-DSA-65 branch.

The branch is not supposed for merge but rather for discussing whether higher level API is what PyCA wants to see.

We have validated that with these patches we are able to enable FreeIPA to operate as ML-DSA CA.


alex:
I had a cursory review, the Python API here looks pretty consistent with the one in #14404 which is good. The major differences look to be: a) whether context comes via a seperate method or as an arg to sign/verify (we landed on arg, rather than seperate method in the other PR), and the exact spelling of some serialization methods.

This is a good sign and indicates that we're headed in the right direction.

alex:
All of ML-DSA for AWS-LC+BoringSSL has landed, so it should now be possible to rebase this on that.

abbra:
Thanks, yes, it is in plan. @frasertweedale is working on submitting rust-openssl patches that we depend on first.

reaperhulk:
We've fully released mldsa/mlkem including for openssl now. @frasertweedale is working on some better APIs up in rust-openssl which we will adopt when they land/are released.

PR #14441, opened 2026-03-08
Add more property-caching optimizations to x509 Rust backend

abbra:
# PyCA x509 Rust backend — caching optimizations

I was working on my general purpose ASN.1 library and when creating Python bindings, I tested against PyCA code. Some operations in PyCA were slow compared to my code so I wanted to look into what could be improved. With the help of Claude Code I've got some repeatable patterns improved using the same approach PyCA already had in place for some properties.

Below is a report Claude created.

## Background

The x509 Rust backend (`src/rust/src/x509/`) converts parsed ASN.1 data into Python objects on every property access. Operations like name parsing (`parse_name`), public-key loading, OID conversion, and serial-number iteration are not cheap: each one allocates Python objects, traverses ASN.1 sequences, and crosses the Rust/Python FFI boundary. In workloads that touch the same property more than once on the same object (chain building, path validation, CRL checking, OCSP processing) this cost is paid repeatedly and unnecessarily.

The existing mitigation is `pyo3::sync::PyOnceLock<pyo3::Py<pyo3::PyAny>>`: a thread-safe write-once cell that stores the Python object after the first computation. It was already used for extension lists everywhere. The work described here extends that pattern to the remaining uncached properties.

## Caching pattern

Every cached getter follows the same idiom:

```rust
// struct field
cached_foo: pyo3::sync::PyOnceLock<pyo3::Py<pyo3::PyAny>>,

// getter
fn foo<'p>(&self, py: Python<'p>) -> PyResult<Bound<'p, PyAny>> {
    Ok(self.cached_foo
        .get_or_try_init(py, || expensive_computation(py).map(|v| v.unbind()))?
        .bind(py)
        .clone())
}
```

`get_or_try_init` is a no-op on every call after the first; the atomic check costs ~50 ns, the cached result is returned without any allocation.

## What was implemented

Ten changes were made across five files, committed individually on the `performance-improvements` branch.

| Commit | File | Change |
|--------|------|--------|
| `e7cc638` | `csr.rs` | Cache `CertificateSigningRequest.attributes` |
| `75451dc` | `ocsp_req.rs` | Cache `OCSPRequest` `issuer_name_hash`, `issuer_key_hash`, `hash_algorithm`, `serial_number` |
| `1830ca3` | `certificate.rs`, `pkcs7.rs`, `ocsp_resp.rs` | Cache `Certificate` `issuer`, `subject`, `public_key`, `signature_algorithm_oid`, `signature_hash_algorithm` |
| `e90f4e5` | `ocsp_resp.rs` | Fix O(n²) `certificates` iteration; cache the resulting list |
| `986298b` | `ocsp_resp.rs` | Add `OCSPSingleResponse.extensions` getter with caching |
| `f18d144` | `crl.rs` | Cache `CRL` `issuer`, `signature_algorithm_oid`, `signature_hash_algorithm` |
| `d149aaf` | `crl.rs` | Replace `get_revoked_certificate_by_serial_number` linear scan with O(1) `HashMap` |

The `OCSPResponse.certificates` getter additionally had a documented O(n²) bug (each certificate extracted via `clone().nth(i)` restarted the iterator). It was replaced with a single linear pass using `asn1::write_single` to produce independent DER bytes for each certificate, eliminating the need for the `map_arc_data_ocsp_response` unsafe helper.

## Benchmark results

Benchmarks measure repeated access on a single pre-loaded object — the workload that caching is designed to accelerate. Each benchmark creates the object once outside the timed loop, then calls the getter in a tight loop.

Comparison: `main` (baseline) vs `abbra-p-f` (PR), both built with `maturin develop --release`, Python 3.14, OpenSSL 3.5.

| Benchmark | Baseline (median) | PR (median) | Speedup |
|-----------|------------------:|------------:|---------|
| `certificate_subject` | 7141 ns | 56 ns | **99% faster** |
| `certificate_issuer` | 5626 ns | 56 ns | **99% faster** |
| `crl_issuer` | 5460 ns | 56 ns | **99% faster** |
| `certificate_public_key` | 1361 ns | 55 ns | **96% faster** |
| `ocsp_request_properties` | 1525 ns | 119 ns | **92% faster** |
| `crl_serial_number_lookup_miss` | 2159 ns | 224 ns | **90% faster** |
| `certificate_signature_hash_algorithm` | 190 ns | 56 ns | **71% faster** |
| `crl_serial_number_lookup_hit` | 448 ns | 295 ns | **34% faster** |
| `certificate_signature_algorithm_oid` | 109 ns | 56 ns | **49% faster** |
| `ocsp_response_properties` | 1190 ns | 1150 ns | ~3% (noise) |

The subject/issuer/CRL-issuer gains are ~100× because `parse_name` is the most expensive operation — it constructs a full Python `Name` object tree from ASN.1 on every call. The cached path costs only an atomic load plus a Python reference clone (~50 ns regardless of name complexity).

`crl_serial_number_lookup_hit` is 34% faster rather than near-zero because `get_revoked_certificate_by_serial_number` must still construct a new `RevokedCertificate` Python object on each hit (the `HashMap` stores `OwnedRevokedCertificate` values that are cloned per call). The miss path (90% faster) avoids iterating the whole list and drops from O(n) to O(1).

`ocsp_response_properties` shows no meaningful change because the properties benchmarked there (`issuer_key_hash`, `serial_number`, `signature_hash_algorithm` on the response-level object) were already relatively cheap and the test exercises only a few iterations of the warm path.

## Why the existing load benchmarks showed no improvement

`test_load_der_certificate` and `test_load_pem_certificate` each call `x509.load_der_x509_certificate(bytes)` per iteration, creating a fresh object with empty caches each time. The cache is always cold; caching adds zero benefit and a tiny overhead (extra `PyOnceLock::new()` fields). These benchmarks measure parsing throughput, not property-access throughput, so they are unaffected by this work.

## Benchmark reproduction

```bash
uv venv /tmp/bench-venv --python python3.14
uv pip install --python /tmp/bench-venv/bin/python \
    maturin pytest pytest-benchmark certifi setuptools cffi
uv pip install --python /tmp/bench-venv/bin/python -e vectors/

# baseline (main branch)
git checkout main
cp tests/bench/test_x509.py /tmp/bench_test.py   # copy new benchmarks over
VIRTUAL_ENV=/tmp/bench-venv maturin develop --release
/tmp/bench-venv/bin/python -m pytest tests/bench/test_x509.py \
    -k "subject or issuer or public_key or signature or crl_serial or ocsp" \
    --benchmark-json=/tmp/bench_base.json --benchmark-enable \
    --benchmark-warmup=on --benchmark-min-rounds=200 -q

# PR branch
git checkout abbra-p-f
VIRTUAL_ENV=/tmp/bench-venv maturin develop --release
/tmp/bench-venv/bin/python -m pytest tests/bench/test_x509.py \
    -k "subject or issuer or public_key or signature or crl_serial or ocsp" \
    --benchmark-json=/tmp/bench_pr.json --benchmark-enable \
    --benchmark-warmup=on --benchmark-min-rounds=200 -q

python3 .github/bin/compare_benchmarks.py /tmp/bench_base.json /tmp/bench_pr.json
```


alex:
Thanks for submitting this -- for ease of review, can you split this into a few smaller PRs? My suggsetion would be to start with splitting out:

1. The subject/issuer properties
2. the public key properties

and we can go from there. Thanks

abbra:
@alex thanks, I opened https://github.com/pyca/cryptography/pull/14442 for the first one. Since all other PRs would depend on the previous ones being merged, should I wait with the remaining ones?

reaperhulk:
Yeah since GH hasn’t shipped dependent PRs yet you should just submit one and once it merges rebase and submit the next.

alex:
We've landed a bunch of the constituent PRs here, so I'm going to close this.

PR #14445, opened 2026-03-08
add attributes caching to OCSP request

abbra:
Continuation of https://github.com/pyca/cryptography/pull/14441

### DIFF
--- a/src/rust/src/x509/ocsp_req.rs
+++ b/src/rust/src/x509/ocsp_req.rs
@@ -46,6 +46,7 @@ pub(crate) fn load_der_ocsp_request(
         cached_extensions: pyo3::sync::PyOnceLock::new(),
         cached_issuer_name_hash: pyo3::sync::PyOnceLock::new(),
         cached_issuer_key_hash: pyo3::sync::PyOnceLock::new(),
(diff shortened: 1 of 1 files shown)


src/rust/src/x509/ocsp_req.rs
```diff
@@ -71,40 +79,86 @@ impl OCSPRequest {
 #[pyo3::pymethods]
 impl OCSPRequest {
     #[getter]
-    fn issuer_name_hash(&self) -> &[u8] {
-        self.cert_id().issuer_name_hash
```
alex, line 75: I wonder if we should be returning some sort of buffer here that points into self (while ensuring self is alive, of course). Feels excessive to need to cache something that's just constructing a bytes object...

WDYT @reaperhulk

alex (commented):
hmm, caching things that are just type conversions/

src/rust/src/x509/ocsp_req.rs
alex: (filed https://github.com/PyO3/pyo3/issues/5871 to try to make this cheaper!)

reaperhulk:
I think this optimization isn't worth it at this time, but thank you for the other contributions!

Issue #10393, opened 2024-02-14
PyOpenSSL X509Store / SSLContext parity in Cryptography


vEpiphyte:
Hello!

I've been working on updating some code to utilize cryptography in favor of PyOpenSSL due to the API deprecation in the older project.

The only code that I can not currently remove is related to the use of X509Store and X509StoreContext. That is utilized for doing certificate validation. For example:

```
    def verifyACertificate(self, cert: c_x509.Certificate):
    	'''
    	Check if a Certificate is trusted by a set of CAs and is not revoked.
    	'''
        crls = self.getCaCrls()  # type: List[c_x509.CertificateRevocationList]
        cacerts = self.getCaCerts()  # type: List[c_x509.Certificate]

        store = crypto.X509Store()  # This is pyopenssl...
        [store.add_cert(crypto.X509.from_cryptography(cacert)) for cacert in cacerts]

        if crls:
            # Setting flags is important. Other uses use things such as PARTIAL_CHAIN flags.
            store.set_flags(crypto.X509StoreFlags.CRL_CHECK | crypto.X509StoreFlags.CRL_CHECK_ALL)
            [store.add_crl(crypto.CRL.from_cryptography(crl)) for crl in crls]

        ctx = crypto.X509StoreContext(store, crypto.X509.from_cryptography(cert))
        try:
            ctx.verify_certificate()  # raises X509StoreContextError if unable to verify
        except crypto.X509StoreContextError as e:
            mesg = _unpackContextError(e)  # helper function to unpack the X509StoreContextError, not important here.  
            raise Exception(mesg)
        return cert
```

I believe my use case aligns with https://github.com/pyca/cryptography/issues/10276 ( doing code signing and/or user cert verification ). Current docs for the verification APIS ( https://cryptography.io/en/42.0.2/x509/verification/ ) don't seem to support setting CRLs or flag setting.

Is this type of use case in scope for work in https://github.com/pyca/cryptography/pull/10345 ?

alex:
cc: @woodruffw 

Looks like there's two pieces you need here:

1. Doing verification without checking for a particular SAN, which is what @woodruffw has in his PR
2. Checking CLRs.

Both of these are tracked in https://github.com/pyca/cryptography/issues/10034

Figuring out CLRs is going to require some API design work on our part. It looks like your implementation relies on having pre-fetched the CRLs, and not loading them on-demand.

vEpiphyte:
[...]

PR #14501, opened 2026-03-18
verification: add skeleton for revocation checking

tnytown:
This is an outline of the CRL revocation API that was discussed in #10393. I'd like to get some feedback on the general approach before we go further and fill in the revocation algorithm.


src/rust/src/x509/verify/mod.rs
tnytown, line 1: I'd appreciate thoughts on the general approach here with revocation checker ownership. I assume that we eventually want to instantiate `RevocationChecker` for OCSP, so I patterned everything after the existing `Subject` type. It feels a little bit clunky.

src/cryptography/x509/verification.py
```diff
@@ -32,3 +35,13 @@
 ExtensionPolicy = rust_x509.ExtensionPolicy
 Criticality = rust_x509.Criticality
 VerificationError = rust_x509.VerificationError
+CRLRevocationChecker = rust_x509.CRLRevocationChecker
+
+
+class RevocationChecker(metaclass=abc.ABCMeta):
+    """
+    An interface for revocation checkers.
+    """
```
alex, line 44: So right now the idea is that we won't actually have a Python API you can implement for revocation checking, it'll be handled internally and these are more markers, is that the idea?

src/rust/src/x509/verify/mod.rs
alex: It's not clear to me that the RevocationCheckerOwner is actually doing any work here? It only ever lives on the stack AFAICT, which suggests it shouldn't be needed.

src/cryptography/x509/verification.py
tnytown: ~~Correct, sorry that this isn't more clear—I experimented with having an extensible Python API but it was a bit complicated to shim over considering that the implementation for CRL is going to be in Rust. I can sketch out a shim if you'd like?~~ see latest diff which supports Python-extensible revocation checkers

src/rust/src/x509/verify/mod.rs
tnytown: ~~Maybe there's a better name for it, but `RevocationCheckerOwner` helps us by holding the concrete `PyCrlRevocationChecker`, which we then can borrow a `CrlRevocationChecker` from to pass to `cryptography-x509-verification`. Another way we could do this is by holding a plain `PyCrlRevocationChecker`, but we'd have to figure something out in the future to accept multiple RevocationChecker impls. I originally wanted to directly borrow a `CrlRevocationChecker` out of the `PyAny` but couldn't find a way to; I'm not very handy with pyo3 😅~~

tnytown: Scratch the above, I had an idea that makes everything simpler and allows for custom Python revocation checkers :)

src/rust/src/x509/verify/revocation.rs
tnytown, line 1: The idea here is to have `CrlRevocationChecker` and `PyRevocationChecker` both implement the trait that is used in the verification crate. `PyRevocationChecker` is a marker class that implements the `CheckRevocation` trait and dispatches to the underlying Python method for Python revocation checker implementations.

src/cryptography/x509/verification.py
```diff
@@ -32,3 +35,22 @@
...
+    def is_revoked(
+        self,
+        leaf: rust_x509.Certificate,
+        issuer: rust_x509.Certificate,
+        policy: Policy,
+    ) -> bool | None:
+        """
+        Returns whether the certificate is revoked. If the revocation status
+        cannot be determined, the revocation checker may return None.
+        """
```
tnytown, line 56: This signature is a bit subtle. There are situations in which a revocation checker cannot determine the revocation status for a given certificate: for the CRL checker, the checker might not contain a relevant CRL; and for OCSP, the responder may be offline. I figure that it's best to signal this instead of collapsing that error state into "not revoked".

src/cryptography/x509/verification.py
```diff
@@ -32,3 +35,22 @@
 ExtensionPolicy = rust_x509.ExtensionPolicy
 Criticality = rust_x509.Criticality
 VerificationError = rust_x509.VerificationError
+CRLRevocationChecker = rust_x509.CRLRevocationChecker
+
+
+class RevocationChecker(rust_x509.RevocationChecker, metaclass=abc.ABCMeta):
```
tnytown, line 41: note that we now have an ABC that inherits from the Rust class that implements the `CheckRevocation` trait

tnytown (commented):
@alex this is ready for another round of review :)

alex:
What would it be useful to review for? Do we have x509-limbo test cases for this yet?

tnytown:
> What would it be useful to review for? Do we have x509-limbo test cases for this yet?

@alex I'd like your eyes on the approach wrt. bridging Python revocation checkers, just to make sure that I'm not totally off base. We do have a few Limbo tests for CRLs already (https://github.com/C2SP/x509-limbo/pull/538), but I'm working on a few more and will get them up to cover the meat of the `CRLRevocationChecker`.

alex:
I briefly skimmed, and I think it'd be helpful if you wrote a few paragraphs/examples of what the design/goals are. Right now it's very unclear which parts of the inheritance/ABCs/etc. are incidental vs. the intended design.

tnytown:
@alex Sure, so the whole implementation here is predicated on a couple assumptions, namely that we'd like to (1) implement the revocation checker for CRLs in Rust while (2) allowing for custom revocation checker implementations from Python. Let me know if we don't need the extensibility on the Python end, it'd allow us to get rid of a lot of this infrastructure.

I started off by defining a revocation checker trait [`CheckRevocation`](https://github.com/pyca/cryptography/pull/14501/changes#diff-9cfd45504507680587a5601614ecf64764adbbca188701ebb6a4f456b1ce7659R13-R20) in `cryptography-x509-verification`. This is the basic interface that the path building logic uses, and instances of it are provided to the public API via a new argument in `cryptography_x509_verification::verify`. The CRL revocation checker [`CrlRevocationChecker`](https://github.com/pyca/cryptography/pull/14501/changes#diff-9cfd45504507680587a5601614ecf64764adbbca188701ebb6a4f456b1ce7659R22-R42) also lives in `cryptography-x509-verification` and implements `CheckRevocation`. We can implement other (OCSP?) revocation checkers in Rust in the future with the same basic pattern, they just need to implement `CheckRevocation`. Constructing a CRL revocation checker from Rust and using it with verification is pretty straightforward, to wit:

```rust
// `::new` accepts an iterator of `CertificateRevocationList`
let checker = CrlRevocationChecker::new(&crls);
cryptography_x509_verification::verify(&leaf, &intermediates, &policy, &checker, &store);
```

At this point, we have to figure out how to expose `CrlRevocationChecker` to Python and make it constructible as a Python class. Luckily, this was done before with `PyStore`, so I just constructed a [`PyCrlRevocationChecker`](https://github.com/pyca/cryptography/pull/14501/changes#diff-583af7328f25d005b0496f0f5a0696474531cadf1cb8043b472a96aeb91d319eR19-R50) patterned off of that API. The new API [`PolicyBuilder.revocation_checker`](https://github.com/pyca/cryptography/pull/14501/changes#diff-797165641e2b15003aebf5f011e31335bb7c9cd160636ddd5afb30c1c2cca1bbR196-R208) accepts instances of these, so we can construct a verifier with a revocation checker like so:

```python
builder = PolicyBuilder().store(store).time(validation_time)
builder = builder.revocation_checker(CRLRevocationChecker([crl]))
verifier = builder.build_client_verifier()
```

The question then becomes, what should the signature of `PolicyBuilder.revocation_checker` be? If we want to allow for further instantiations like an OCSP revocation checker, we can't just accept `PyCrlRevocationChecker`. I decided to make a marker class, [`PyRevocationChecker`](https://github.com/pyca/cryptography/pull/14501/changes#diff-583af7328f25d005b0496f0f5a0696474531cadf1cb8043b472a96aeb91d319eR52-R61), that `PyCrlRevocationChecker` inherits from. This allows us to be stricter in the signature of `revocation_checker`, which `pyo3` enforces:
```rust
fn revocation_checker(
    &self,
    py: pyo3::Python<'_>,
    revocation_checker: pyo3::Py<PyRevocationChecker>,
) -> CryptographyResult<PolicyBuilder>
```

This is good enough for the Rust-implemented revocation checkers, but we also want Python extensibility. Starting with the user-facing API, we want to support passing Python-implemented checkers to `PolicyBuilder.revocation_checker`, just like Rust-implemented checkers. I started with a [`RevocationChecker` ABC](https://github.com/pyca/cryptography/pull/14501/changes#diff-510c826084eed421841cb286d02bfaa703f87ff0747841c7a219d62decbd61bbR41-R56) that roughly mirrors the `CheckRevocation` trait. This ABC inherits from our `PyCrlRevocationChecker` marker class, which allows `PolicyBuilder.revocation_checker` to take instances of it.

The question now becomes, how do we bridge the Python ABC and the Rust `CheckRevocation` trait? [Unfortunately, PyO3 doesn't do this for us yet](https://github.com/PyO3/pyo3/issues/1190). The solution I settled on was [implementing the `CheckRevocation` trait on `Py<PyRevocationChecker>`](https://github.com/pyca/cryptography/pull/14501/changes#diff-583af7328f25d005b0496f0f5a0696474531cadf1cb8043b472a96aeb91d319eR72-R105). This allows us to call the concrete `is_revoked` method implemented in Python with `call_method1` from Rust.

Now, from the perspective of the `cryptography-rust` bindings, we can get a `CheckRevocation` trait object by trying to cast to `PyCrlRevocationChecker` and falling through to the implementation on `Py<PyRevocationChecker>`. [This is what `build_rust_revocation_checker` does.](https://github.com/pyca/cryptography/pull/14501/changes#diff-583af7328f25d005b0496f0f5a0696474531cadf1cb8043b472a96aeb91d319eR108-R118) If we wanted to separate things out more, I could add another `pyclass` for the ABC to subclass from instead of having it subclass from the `PyRevocationChecker`.

Sorry for all the complexity, hopefully this clears things up a bit. Let me know if there's anything I can clarify.

reaperhulk:
Does this need input from us @tnytown? @alex and I are sprinting on cryptography this week so we've got bandwidth to talk things through if needed.

tnytown:
@reaperhulk Yeah, I'd appreciate it if you guys take a quick look at the API surface I outlined in my comment above and ACK/NACK it. To recap a bit, there's some complexity around the design that arises from accepting both Python and Rust implementations for revocation checkers. The CRL revocation checker that I'm trying to land is targeting the WebPKI only, so if folks want to do stuff with iCRLs or other features outside of that scope, my feeling is that we can give them an escape hatch with Python extensibility.

Coincidentally, I'm working on this for the next few days as well; I hope to have the actual implementation up soon for you guys to review.

alex:
My instinct is that we should start with revocation checker only taking a concrete CRLRevocationChecker type -- that'll let us build the relevant pieces. We can always make the `revocation_checker()` more general later.

tnytown:
Summarizing our discussion on Slack: the goal is to review and merge a concrete implementation first before extending it for multiple revocation checker implementations.

PR #14638, opened 2026-04-11
Convert development extras to dependency groups

sirosen:
closes #14622

I had to make a couple of decisions here, which I'm happy to change in response to feedback.

1.  I didn't preserve empty extras.

    Some projects keep old extras but make them empty, e.g. defining
    `cryptography[test]` as `[]`.

    It's not totally clear to me why.[^1]

    `pip install cryptography[bogus]` will warn, but not error.
    Lacking any clear justification for it, I have not kept them.

2.  Only `test-randomorder` is modified to include another group.

    The `docs` and `docstest` groups would be an obvious option to change in
    this way, but the usage in `nox` makes it nonobvious that it's worth doing
    so. The dependency groups could be rearranged in any way in the future.

3.  Added a `dev` group for all groups.

    `dev` is a semi-standard name for a default or catch-all group, under
    `uv` and a few other workflow tools. The exact meaning of `dev` is not
    set or standardized. In this case, it was added because it's useful to
    have some "all" group for `uv pip compile --group=...`, and there's no
    obviously correct name. (`all` would also work)

[^1]: I'm fairly experienced and involved in packaging specifications, but it's hard to know and remember everything. :)

---

Move development extras to dependency groups, thereby removing them from
downstream dependency resolution on `cryptography` itself.

The dev extras are completely removed, not kept as empty extras.

A new `dev` extra, including all of the others, is defined primarily for
use in `uv pip compile`. `ci-constraints-requirements.txt` is regenerated
using this group (which appears in the comes-from annotation comments).
This updates `pygments` data in a minor way.

`test-randomorder` includes `test`, but no other groups are modified in
the conversion.

`noxfile.py` and `.readthedocs.yml` are updated to load these dependency groups as appropriate.

edit: I forgot RTD; amended to include it.


noxfile.py
```diff
@@ -50,11 +50,12 @@ def load_pyproject_toml() -> dict:
 @nox.session(name="tests-nocoverage")
 @nox.session(name="tests-rust-debug")
 def tests(session: nox.Session) -> None:
-    extras = "test"
+    extras: tuple[str, ...] = ()
+    groups: tuple[str, ...] = ("test",)
```
alex: tuples should not be used for homogenous sequences of values, lists should be used instead.

noxfile.py
```diff
@@ -69,15 +70,19 @@ def tests(session: nox.Session) -> None:
...
             "--config-settings-package=cryptography:build-args=--profile=dev",
-            f".[{extras}]",
+            install_spec,
         )
     else:
-        install(session, f".[{extras}]")
+        install(session, install_spec)
+
+    pyproject_data = load_pyproject_toml()
+    install(session, *nox.project.dependency_groups(pyproject_data, *groups))
```
alex, line 86: is there a reason we have to have nox compute these, is it not possible to delegate to the installer (as we do with etxras)?

sirosen: It isn't possible...yet.

I didn't get `--group` added until pip 25.1 , which is also the first pip version which dropped 3.8. As soon as 3.8 support ends and 3.9 becomes the lowest target, `install(session, "--group=dev")` and similar should work.

Would you like me to add a comment to note this? I could include the desired future invocation, e.g.,
```python
*(f"--group={g}" for g in groups)
```

noxfile.py
sirosen: It didn't seem worth putting an extra commit in the history (unless external PRs get squashed?), so I did this in an amend.

noxfile.py
alex: just a comment saying it can be updated once we drop 3.8 is great, thanks

alex (commented):
thanks!

alex (approved):
thanks!

PR #14647, opened 2026-04-12
Add ML-DSA support for BoringSSL

alex:
## Summary

Extends ML-DSA (FIPS 204) support to BoringSSL. Previously ML-DSA was only enabled when building against AWS-LC; BoringSSL now has ML-DSA support but uses a different API than AWS-LC, so this PR adds conditional compilation paths for both.

## Key differences

**BoringSSL API:**
- `EVP_PKEY_from_private_seed()` / `EVP_PKEY_from_raw_public_key()` with `EVP_pkey_ml_dsa_{44,65,87}()` for key creation
- `EVP_DigestSign` / `EVP_DigestVerify` (via `openssl::md_ctx::MdCtx`) plus `EVP_PKEY_CTX_set1_signature_context_string` for signing/verifying with context strings
- Per-variant NIDs (`NID_ML_DSA_{44,65,87}`) identify ML-DSA keys via `pkey.id()`

**AWS-LC API (unchanged):**
- `EVP_PKEY_pqdsa_new_raw_private_key` / `EVP_PKEY_pqdsa_new_raw_public_key` for key creation
- `ml_dsa_{44,65,87}_sign/verify` direct functions for signing (AWS-LC's EVP PQDSA path hardcodes context to `NULL`, so the low-level functions are needed to preserve context string support)
- Shared `NID_PQDSA` with `EVP_PKEY_pqdsa_get_type` for variant identification

## Implementation notes

- `mldsa_seed_from_pkey` in `cryptography-key-parsing` now works for both backends, since BoringSSL and AWS-LC both serialize ML-DSA private keys as a seed in PKCS#8
- Signing and verification on BoringSSL use `MdCtx::digest_sign_init` / `digest_verify_init` (which return `&mut PkeyCtxRef`) so the ML-DSA context string can be set via a single unsafe call to `EVP_PKEY_CTX_set1_signature_context_string`, and then `digest_sign_to_vec` / `digest_verify` handle the actual operation safely (this requires openssl 0.10.77, which #14644 already bumped to)
- `is_mldsa_pkey_type` helper replaces the single `PKEY_ID` constant, since BoringSSL uses per-variant NIDs for the EVP pkey type

## Test plan

- [x] `OPENSSL_DIR=<boringssl> nox -e local` passes on BoringSSL (3149 passed, 931 skipped)
- [ ] CI: BoringSSL build job
- [ ] CI: AWS-LC build job (confirm no regression)
- [ ] CI: OpenSSL build jobs (confirm no regression — ML-DSA remains unsupported there)

https://claude.ai/code/session_01YVJGsasgsDfKxpkG91NWAw

### DIFF
--- a/src/rust/cryptography-openssl/src/mldsa.rs
+++ b/src/rust/cryptography-openssl/src/mldsa.rs
@@ -86,13 +86,14 @@ fn set_context_string<T>(
 ) -> OpenSSLResult<()> {
     // SAFETY: EVP_PKEY_CTX_set1_signature_context_string sets the ML-DSA
     // context string on an initialized sign/verify context.
-    unsafe {
-        cvt(ffi::EVP_PKEY_CTX_set1_signature_context_string(
+    let res = unsafe {
+        ffi::EVP_PKEY_CTX_set1_signature_context_string(
             pkey_ctx.as_ptr(),
             context.as_ptr(),
             context.len(),
-        ))?;
-    }
+        )
+    };
+    cvt(res)?;
     Ok(())
 }
 
--- a/src/rust/cryptography-openssl/src/mldsa.rs
+++ b/src/rust/cryptography-openssl/src/mldsa.rs
@@ -79,40 +79,19 @@ fn evp_pkey_alg(variant: MlDsaVariant) -> *const ffi::EVP_PKEY_ALG {
     }
 }
 
-/// A wrapper around a `PkeyCtxRef` that has been initialized for either a
-/// sign or verify digest operation. The only way to construct one is via
-/// [`InitializedPkeyCtxRef::sign_init`] or [`InitializedPkeyCtxRef::verify_init`],
-/// which perform the underlying `EVP_DigestSignInit`/`EVP_DigestVerifyInit`
-/// call. This lets `set_context_string` assume the underlying context is
-/// already initialized without having to document an implicit precondition.
+/// A marker type for a `PkeyCtxRef` that has been initialized for a sign or
+/// verify operation. Callers explicitly wrap the ref in this type immediately
+/// after calling `digest_sign_init` / `digest_verify_init`, which makes the
+/// precondition required by [`InitializedPkeyCtxRef::set_context_string`]
+/// visible in the source.
 #[cfg(CRYPTOGRAPHY_IS_BORINGSSL)]
 struct InitializedPkeyCtxRef<'a, T>(&'a mut openssl::pkey_ctx::PkeyCtxRef<T>);
 
 #[cfg(CRYPTOGRAPHY_IS_BORINGSSL)]
-impl<'a, T> InitializedPkeyCtxRef<'a, T> {
-    fn sign_init(
-        md_ctx: &'a mut openssl::md_ctx::MdCtx,
-        pkey: &openssl::pkey::PKeyRef<T>,
-    ) -> OpenSSLResult<Self>
-    where
-        T: openssl::pkey::HasPrivate,
-    {
-        Ok(Self(md_ctx.digest_sign_init(None, pkey)?))
-    }
-
-    fn verify_init(
-        md_ctx: &'a mut openssl::md_ctx::MdCtx,
-        pkey: &openssl::pkey::PKeyRef<T>,
-    ) -> OpenSSLResult<Self>
-    where
-        T: openssl::pkey::HasPublic,
-    {
-        Ok(Self(md_ctx.digest_verify_init(None, pkey)?))
-    }
-
+impl<T> InitializedPkeyCtxRef<'_, T> {
     fn set_context_string(&mut self, context: &[u8]) -> OpenSSLResult<()> {
-        // SAFETY: By construction, the wrapped pkey_ctx has been initialized
-        // via digest_sign_init or digest_verify_init.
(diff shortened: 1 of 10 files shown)


src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -143,54 +213,76 @@ pub fn sign(
...
+            let pkey_ctx = md_ctx.digest_sign_init(None, pkey)?;
+            if !context.is_empty() {
+                // SAFETY: EVP_PKEY_CTX_set1_signature_context_string sets the
+                // ML-DSA context string on an initialized signing context.
+                unsafe {
+                    cvt(ffi::EVP_PKEY_CTX_set1_signature_context_string(
+                        pkey_ctx.as_ptr(),
+                        context.as_ptr(),
+                        context.len(),
+                    ))?;
```
alex: this is missing coverage (and the same on the other invocation). I suggest splitting the unsafe function call into a `fn set_context_string(pkey, context) -> Result`. the reformatting should resolve this.

src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -29,18 +43,60 @@ impl MlDsaVariant {
...
+    context: &[u8],
+) -> OpenSSLResult<()> {
+    // SAFETY: EVP_PKEY_CTX_set1_signature_context_string sets the ML-DSA
+    // context string on an initialized sign/verify context.
+    unsafe {
+        cvt(ffi::EVP_PKEY_CTX_set1_signature_context_string(
+            pkey_ctx.as_ptr(),
+            context.as_ptr(),
+            context.len(),
+        ))?;
```
alex: do `res = ffi::..;` and then `cvt(res)?` on its own line

src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -29,18 +43,61 @@ impl MlDsaVariant {
...
+        }
+    }
+}
+
+#[cfg(CRYPTOGRAPHY_IS_BORINGSSL)]
+fn set_context_string<T>(
+    pkey_ctx: &mut openssl::pkey_ctx::PkeyCtxRef<T>,
+    context: &[u8],
+) -> OpenSSLResult<()> {
+    // SAFETY: EVP_PKEY_CTX_set1_signature_context_string sets the ML-DSA
```
reaperhulk: This is a pretty big safety precondition. Can we wrap the pkeyctxref in a struct like InitializedPkeyRef so that you have to construct that type to call this or is that too much/not actually helpful?

src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -79,22 +79,50 @@ fn evp_pkey_alg(variant: MlDsaVariant) -> *const ffi::EVP_PKEY_ALG {
...
-            context.as_ptr(),
-            context.len(),
-        )
-    };
-    cvt(res)?;
-    Ok(())
+struct InitializedPkeyCtxRef<'a, T>(&'a mut openssl::pkey_ctx::PkeyCtxRef<T>);
+
+#[cfg(CRYPTOGRAPHY_IS_BORINGSSL)]
+impl<'a, T> InitializedPkeyCtxRef<'a, T> {
```
reaperhulk: This is overkill. I meant to kick off a discussion, not immediately implement the most heavyweight form of this. If we were to do this at all I really just want the caller to have to wrap it in something so that we can have an explicit type.

src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -29,18 +43,68 @@ impl MlDsaVariant {
...
+/// precondition required by [`InitializedPkeyCtxRef::set_context_string`]
+/// visible in the source.
+#[cfg(CRYPTOGRAPHY_IS_BORINGSSL)]
+struct InitializedPkeyCtxRef<'a, T>(&'a mut openssl::pkey_ctx::PkeyCtxRef<T>);
+
+#[cfg(CRYPTOGRAPHY_IS_BORINGSSL)]
+impl<T> InitializedPkeyCtxRef<'_, T> {
+    fn set_context_string(&mut self, context: &[u8]) -> OpenSSLResult<()> {
+        // SAFETY: By construction, the caller has wrapped a pkey_ctx that
+        // has been initialized for a sign/verify operation.
```
alex: this whole thing is pointless and should be reverted, PkeyCtxRef are always initialized, it's already an invariant of the type

src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -110,31 +166,62 @@ pub fn new_raw_private_key(
...
-            variant.nid(),
-            data.as_ptr(),
-            data.len(),
-        ))?;
-        Ok(openssl::pkey::PKey::from_ptr(pkey))
+    cfg_if::cfg_if! {
+        if #[cfg(CRYPTOGRAPHY_IS_BORINGSSL)] {
+            // SAFETY: EVP_PKEY_from_raw_public_key creates a new EVP_PKEY
+            // from raw public key bytes.
+            unsafe {
```
alex: Can this use the existing `PKey::public_key_from_raw_bytes` and be entirely safe (using the nid for the variant)?

reaperhulk (approved):

PR #14666, opened 2026-04-15
fix: AES-KWP allows wrapping empty `key_to_wrap`, inconsistent with RFC 5649

### LINKED ISSUE #14664: AES-KWP allows wrapping empty `key_to_wrap`, inconsistent with RFC 5649
Hi,

while testing the pyca/cryptography library I noticed that `aes_key_wrap_with_padding` allows wrapping an empty key. The resulting wrapped key is invalid and is rejected by `aes_key_unwrap_with_padding`:

```python
>>> import cryptography; cryptography.__version__
'46.0.5'
>>> from cryptography.hazmat.primitives.keywrap import *
>>> kek = b"A"*16
>>> wrapped = aes_key_wrap_with_padding(kek, b"")
>>> wrapped.hex()
'a65959a600000000'
>>> aes_key_unwrap_with_padding(kek, wrapped)
  File "<python-input>", line 1, in <module>
    aes_key_unwrap_with_padding(kek, wrapped)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File ".../cryptography/hazmat/primitives/keywrap.py", line 112, in aes_key_unwrap_with_padding
    raise InvalidUnwrap("Must be at least 16 bytes")
cryptography.hazmat.primitives.keywrap.InvalidUnwrap: Must be at least 16 bytes
```

[RFC 5649](https://www.rfc-editor.org/rfc/rfc5649) specifies that »input data may be as short as one octet« in the introduction and also mentions »the plaintext consists of between one and 2^32 octets« in Section 4.1.

The current implementation in pyca/cryptography doesn't check for this:

https://github.com/pyca/cryptography/blob/cfe93c50f16f8539a6c67f558236103642f3702c/src/cryptography/hazmat/primitives/keywrap.py#L81-L103

When called with `key_to_wrap=b""`, `aes_key_wrap_with_padding` ends up calling `_wrap_core` with `r=[]`, which then just returns the Alternative IV. The unwrapping function correctly requires two AES blocks of input.

Other implementations like OpenSSL's wrap128.c do check for this and reject empty plaintexts:

https://github.com/openssl/openssl/blob/9e26ae32a757178d6862e0951c76615fe10a3d1d/crypto/modes/wrap128.c#L208-L210

Notably, this issue only affects the padded variant of AES key wrapping. It also doesn't seem like a security vulnerability to me, as the edge case doesn't reveal the KEK and the discrepancy between wrapping succeeding and unwrapping failing doesn't seem like it could cause issues in real-world use to me.

I'll send a PR to fix 🙂

SillyZir:
## Summary

AES-KWP allows wrapping empty `key_to_wrap`, inconsistent with RFC 5649 — modified `tests/hazmat/primitives/test_keywrap.py`.

Fixes #14664

## Changes

- tests/hazmat/primitives/test_keywrap.py (+6 lines, 4 modified)

## Rationale

Applied fix for AES-KWP allows wrapping empty `key_to_wrap`, inconsistent with RFC 5649 in `test_keywrap.py`, where the affected behavior originates.

### DIFF
--- a/tests/hazmat/primitives/test_keywrap.py
+++ b/tests/hazmat/primitives/test_keywrap.py
@@ -198,6 +198,12 @@ class TestAESKeyWrapWithPadding:
         with pytest.raises(ValueError, match="must be a valid AES key length"):
             keywrap.aes_key_wrap_with_padding(b"badkey", b"\x00", backend)
 
+    def test_wrap_empty_key_to_wrap(self, backend):
+        with pytest.raises(ValueError, match="key_to_wrap must not be empty"):
+            keywrap.aes_key_wrap_with_padding(
+                b"sixteen_byte_key", b"", backend
+            )
+
     def test_unwrap_invalid_key_length(self, backend):
         with pytest.raises(ValueError, match="must be a valid AES key length"):
             keywrap.aes_key_unwrap_with_padding(


reaperhulk:
This duplicates #14665 but doesn't actually fix it.

PR #14671, opened 2026-04-16
Revert "Add ML-KEM-512 support"

alex:
Reverts pyca/cryptography#14670


alex:
Not in BoringSSL or Go suggests we don't want it

reaperhulk (approved):

PR #14674, opened 2026-04-16
Add documentation for MLKEM

DarkaMaul:
Followup of #14598

### DIFF
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -36,6 +36,9 @@ cryptographically
 de
 Debian
 deallocated
+decapsulate
+Decapsulate
+Decapsulation
 declaratively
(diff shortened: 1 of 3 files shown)


alex:
Looks like there's some spelling issues

docs/spelling_wordlist.txt
```diff
@@ -50,7 +53,6 @@ deserialized
 Deserialization
 deserializing
 Diffie
-Diffie
```
DarkaMaul, line 53: FLAG: I removed a duplicate entry here (I can revert the change and open a separate PR if that's prefered)

reaperhulk: This is fine.

reaperhulk (approved):

PR #14695, opened 2026-04-19
Add MLKEM768-X25519 hybrid KEM support in HPKE

alex:
## Summary

- Adds the MLKEM768-X25519 hybrid KEM (X-Wing, HPKE KEM ID `0x647A`) as `KEM.MLKEM768_X25519`, specified in draft-connolly-cfrg-xwing-kem and draft-ietf-hpke-pq.
- Introduces minimal `MLKEM768X25519PrivateKey(mlkem_key, x25519_key)` / `MLKEM768X25519PublicKey(mlkem_key, x25519_key)` constructor-only types in `cryptography.hazmat.primitives.hpke` that compose existing ML-KEM-768 and X25519 key objects. The private key exposes `public_key()`.
- Encap/decap, the SHA3-256 X-Wing combiner, and all component key access live inside the HPKE backend and go through the same Python-level interfaces (`encapsulate`, `decapsulate`, `exchange`, `public_key`, `public_bytes_raw`) already used by the pure ML-KEM and X25519 HPKE paths.

## Test plan

- [x] All existing HPKE vectors (including `kem_id: 0x647A` from `hpke-pq-test-vectors.json`) decrypt correctly via `test_vector_decryption`; the X-Wing 32-byte seed is expanded to component keys in the test via `hashlib.shake_256`.
- [x] New unit tests cover hybrid roundtrip, ciphertext length (1120 + 4 + 16), wrong-key / wrong-type decryption failures, and using the hybrid public key with a non-hybrid KEM suite.
- [x] `OPENSSL_DIR=/opt/aws-lc nox -e local` passes (ruff, clippy, mypy, cargo check, 3612 pytest passed / 634 skipped, cargo test).

https://claude.ai/code/session_019VW3s2nq5QVCdpUnTvPJDv


alex:
@reaperhulk should the Hybrid types have a `from_seed()` API (currently in the tests) that uses the X-wing expansion? (Claude: do not implement anything until we give you a clear instruction)

src/rust/src/backend/hpke.rs
```diff
@@ -1001,12 +1036,228 @@ fn _decrypt_with_aad<'p>(
...
+        py: pyo3::Python<'p>,
+        enc: &[u8],
+    ) -> CryptographyResult<pyo3::Bound<'p, pyo3::types::PyBytes>> {
+        let expected_len = MLKEM768_X25519_MLKEM_CT_LENGTH + MLKEM768_X25519_X25519_CT_LENGTH;
+        if enc.len() != expected_len {
+            return Err(CryptographyError::from(
+                pyo3::exceptions::PyValueError::new_err(
+                    "An ML-KEM-768/X25519 ciphertext is 1120 bytes long",
+                ),
+            ));
```
alex: There's no test case for this -- is it reachable or does the caller always enforce?

src/rust/src/backend/hpke.rs
```diff
@@ -170,6 +177,7 @@ impl KEM {
             KEM::P521 => kem_params::P521_NSECRET,
             KEM::MLKEM768 => kem_params::MLKEM768_NSECRET,
             KEM::MLKEM1024 => kem_params::MLKEM1024_NSECRET,
+            KEM::MLKEM768_X25519 => kem_params::MLKEM768_X25519_NSECRET,
```
alex, line 180: Need a rust test for this (see how we handle MLKEM below)

src/rust/src/backend/hpke.rs
```diff
@@ -1001,12 +1036,228 @@ fn _decrypt_with_aad<'p>(
...
+        mlkem_key: pyo3::Py<pyo3::PyAny>,
+        x25519_key: pyo3::Py<pyo3::PyAny>,
+    ) -> CryptographyResult<Self> {
+        if !mlkem_key
+            .bind(py)
+            .is_instance(&types::MLKEM768_PRIVATE_KEY.get(py)?)?
+        {
+            return Err(CryptographyError::from(
+                pyo3::exceptions::PyTypeError::new_err("Expected MLKEM768PrivateKey for mlkem_key"),
+            ));
```
alex, line 1122: Missing test coverage for this

src/rust/src/backend/hpke.rs
```diff
@@ -1001,12 +1036,228 @@ fn _decrypt_with_aad<'p>(
...
+        }
+        if !x25519_key
+            .bind(py)
+            .is_instance(&types::X25519_PRIVATE_KEY.get(py)?)?
+        {
+            return Err(CryptographyError::from(
+                pyo3::exceptions::PyTypeError::new_err("Expected X25519PrivateKey for x25519_key"),
+            ));
```
alex, line 1130: And this

src/rust/src/backend/hpke.rs
```diff
@@ -1001,12 +1036,228 @@ fn _decrypt_with_aad<'p>(
...
+        mlkem_key: pyo3::Py<pyo3::PyAny>,
+        x25519_key: pyo3::Py<pyo3::PyAny>,
+    ) -> CryptographyResult<Self> {
+        if !mlkem_key
+            .bind(py)
+            .is_instance(&types::MLKEM768_PUBLIC_KEY.get(py)?)?
+        {
+            return Err(CryptographyError::from(
+                pyo3::exceptions::PyTypeError::new_err("Expected MLKEM768PublicKey for mlkem_key"),
+            ));
```
alex, line 1230: coverage

src/rust/src/backend/hpke.rs
```diff
@@ -1001,12 +1036,228 @@ fn _decrypt_with_aad<'p>(
...
+        }
+        if !x25519_key
+            .bind(py)
+            .is_instance(&types::X25519_PUBLIC_KEY.get(py)?)?
+        {
+            return Err(CryptographyError::from(
+                pyo3::exceptions::PyTypeError::new_err("Expected X25519PublicKey for x25519_key"),
+            ));
```
alex, line 1238: coverage

alex:
@reaperhulk oh, other question: do we want a dedicated type per hybrid (after this there'll be MLKEM1024/P384), or should it be one shared Hybrid{Public,Private}Key type? (I think I prefer this way, but it's more verbose)

reaperhulk:
How many types are we going to have? I prefer the verbose ones unless we're going to have a large proliferation.

Re: `from_seed` -- since we're following Go here I'd be inclined to do what they do. Is there a public seed construction on their APIs?

alex:
It's 2x, this and mlkem1024/p384. Not sure, I don't think so though

reaperhulk:
Okay let's do the verbose ones then and not have from_seed in our public API for now?

alex:
Great, this is r4r then

reaperhulk (approved):

PR #14700, opened 2026-04-20
Add SLH-DSA test vectors and test-vectors documentation

DarkaMaul:
Adds ACVP test vectors for SLH-DSA from the [NIST ACVP-Server](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files).

### Parameter sets

Only the two parameter sets exposed by the BoringSSL public API are included ([`include/openssl/slhdsa.h` @ `8313e23`](https://github.com/google/boringssl/blob/8313e239e50e4d240ac7fc04c8e1aa05d58ca0fc/include/openssl/slhdsa.h)):

- `SLH-DSA-SHA2-128s`
- `SLH-DSA-SHAKE-256f`

### Vector filtering rationale

- **keyGen**: all vectors for the two parameter sets are kept.
- **sigVer**: only `pure` + `external` interface vectors are kept.   BoringSSL's `SLHDSA_*_verify` accepts only the external-interface and I don't think we want to support prehashing.
- **sigGen**: only `deterministic=false` + `pure` + `external` interface vectors are kept.  BoringSSL's signature are non deterministic. The same `pure` / `external` interface restrictions apply as for sigVer.

### Why filtering?

- Without filtering, the two files are 38 MB and 31 MB.
- With filtering, we are at 1 MB and 2 MB.

SLHDSA parameter set is expansive:
- 2 hash functions (SHA-2, SHAKE)
- 3 security levels (128 192 256)
- 2 tree / size tradeoff (small / fast)

### Some notes

- The private key size is `4*n` (with `n` the security level), so much smaller than MLKEM/MLDSA. We don't have to do the dance with seed form only there (and the standard mandates that the key is stored in full).

```
Key checks. [...], the owner of the private key shall check that the private key is 4𝑛
bytes in length and shall use SK.seed and PK.seed to recompute PK.root and compare the
newly generated value with the value in the private key currently held.
```

([FIPS 205 - Section 3.1](https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.205.pdf))


reaperhulk:
Does aws-lc have SLH-DSA support?

DarkaMaul:
> Does aws-lc have SLH-DSA support?

No ([source](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/PQREADME.md))

reaperhulk:
Hmm, I'm reluctant to merge support when aws-lc doesn't have it. Looking at their repo I don't actually even see a request, and I'm not sure where it's in use given the signature sizes.

DarkaMaul:
I searched to see where the support for SLH-DSA landed:

- [OpenSSL](https://github.com/openssl/openssl/tree/master/crypto/slh_dsa)
- [BoringSSL](https://github.com/google/boringssl/tree/main/crypto/slhdsa)

It is missing from:

- AWS-LC (the [search](https://github.com/aws/aws-lc/issues?q=SLH-DSA) returns no results)
- LibreSSL

I can open an issue on AWS-LC to ask if they plan to support SLH-DSA in the future.

reaperhulk:
Yes we should file one to find out plans there.

PR #14724, opened 2026-04-23
Recommend Argon2id over PBKDF2HMAC as KDF

### LINKED ISSUE #14723: Please update PBKDF2HMAC minimum iterations recommendation
The [Fernet documentation](https://cryptography.io/en/latest/fernet/#using-passwords-with-fernet) states:

> The iteration count used should be adjusted to be as high as your server can tolerate. A good default is at least 1,200,000 iterations, which is what [Django recommends as of January 2025](https://github.com/django/django/blob/main/django/contrib/auth/hashers.py).

(and the code example indeed has `iterations=1_200_000`)

[Django has updated their recommendation to 1,500,000](https://github.com/django/django/commit/b83204a06e2a4b01f8580b5452058729786ee1d5); please update your documentation accordingly.

tmo1:
Update the recommended PBKDF2HMAC iteration count from 1,200,000 to 1,500,000 to reflect the latest Django recommendations.

Closes: #14723

### DIFF
--- a/docs/fernet.rst
+++ b/docs/fernet.rst
@@ -231,7 +231,7 @@ such functions; it is generally recommended to use
     >>> import os
     >>> from cryptography.fernet import Fernet
     >>> from cryptography.hazmat.primitives import hashes
-    >>> from cryptography.hazmat.primitives.kdf.pbkdf2 import Argon2id
+    >>> from cryptography.hazmat.primitives.kdf.argon2 import Argon2id
     >>> password = b"password"
     >>> salt = os.urandom(16)
     >>> kdf = Argon2id(
--- a/docs/fernet.rst
+++ b/docs/fernet.rst
@@ -285,4 +285,4 @@ unsuitable for very large files at this time.
 
 .. _`Fernet`: https://github.com/fernet/spec/
 .. _`specification`: https://github.com/fernet/spec/blob/master/Spec.md
- .._`IRTF RFC 9106`: https://datatracker.ietf.org/doc/html/rfc9106#name-parameter-choice
+.. _`IRTF RFC 9106`: https://datatracker.ietf.org/doc/html/rfc9106#name-parameter-choice
--- a/docs/fernet.rst
+++ b/docs/fernet.rst
@@ -239,7 +239,7 @@ password through a key derivation function such as
     ...     algorithm=hashes.SHA256(),
     ...     length=32,
     ...     salt=salt,
-    ...     iterations=1_200_000,
+    ...     iterations=1_500_000,
     ... )
     >>> key = base64.urlsafe_b64encode(kdf.derive(password))
     >>> f = Fernet(key)
@@ -253,8 +253,8 @@ In this scheme, the salt has to be stored in a retrievable location in order
 to derive the same key from the password in the future.
 
 The iteration count used should be adjusted to be as high as your server can
-tolerate. A good default is at least 1,200,000 iterations, which is what `Django
-recommends as of January 2025`_.
+tolerate. A good default is at least 1,500,000 iterations, which is what `Django
+recommends as of September 2025`_.
 
 Implementation
 --------------
--- a/docs/fernet.rst
+++ b/docs/fernet.rst
@@ -221,10 +221,9 @@ Using passwords with Fernet
 ---------------------------
 
 It is possible to use passwords with Fernet. To do this, you need to run the
-password through a key derivation function such as
-:class:`~cryptography.hazmat.primitives.kdf.pbkdf2.PBKDF2HMAC`,
-:class:`~cryptography.hazmat.primitives.kdf.argon2.Argon2id` or
-:class:`~cryptography.hazmat.primitives.kdf.scrypt.Scrypt`.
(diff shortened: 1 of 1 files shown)


docs/fernet.rst
reaperhulk: You'll need to update the text here to match the link

tmo1: Okay - should I do that, or just follow [your other suggestion](https://github.com/pyca/cryptography/issues/14723#issuecomment-4308097512) and rewrite with Argon2id?

reaperhulk: I'd say let's just rewrite to use argon2id and leave PBKDF2 in the rear view mirror where it belongs.

tmo1: Done. I'm new to RST, so I likely made some syntax errors - let me know if I need to make any changes.

docs/fernet.rst
```diff
@@ -221,25 +221,25 @@ Using passwords with Fernet
...
+:class:`~cryptography.hazmat.primitives.kdf.argon2.Argon2id`.
 
 .. doctest::
 
     >>> import base64
     >>> import os
     >>> from cryptography.fernet import Fernet
     >>> from cryptography.hazmat.primitives import hashes
-    >>> from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
+    >>> from cryptography.hazmat.primitives.kdf.pbkdf2 import Argon2id
```
60x20:
```suggestion
    >>> from cryptography.hazmat.primitives.kdf.argon2 import Argon2id
```
Fix import

docs/fernet.rst
```diff
@@ -221,25 +221,25 @@ Using passwords with Fernet
...
     >>> password = b"password"
     >>> salt = os.urandom(16)
-    >>> kdf = PBKDF2HMAC(
-    ...     algorithm=hashes.SHA256(),
+    >>> kdf = Argon2id(
+    ...     salt=salt,
```
60x20:
```suggestion
    ...     salt=salt,
```
Remove unnecessary whitespace

docs/fernet.rst
```diff
@@ -221,25 +221,25 @@ Using passwords with Fernet
...
-    >>> kdf = PBKDF2HMAC(
-    ...     algorithm=hashes.SHA256(),
+    >>> kdf = Argon2id(
+    ...     salt=salt,    
     ...     length=32,
-    ...     salt=salt,
-    ...     iterations=1_200_000,
+    ...     iterations=1,
+    ...     lanes=4,
+    ...     memory_cost=2**21
```
60x20, line 242:
```suggestion
    ...     memory_cost=2048*1024
```
Wouldn't it be better if it was in a more human-readable form?

docs/fernet.rst
```diff
@@ -252,9 +252,11 @@ password through a key derivation function such as
 In this scheme, the salt has to be stored in a retrievable location in order
 to derive the same key from the password in the future.
 
-The iteration count used should be adjusted to be as high as your server can
-tolerate. A good default is at least 1,200,000 iterations, which is what `Django
-recommends as of January 2025`_.
+The :class:`~cryptography.hazmat.primitives.kdf.argon2.Argon2id` parameters
+in the above code example are based on the recommendations of `IRTF RFC 9106`_
+for general applications. For memory-constrained applications, the RFC
+recommends ``iterations=3`` and ``memory_cost=2**16``. See that document for
```
60x20, line 258:
```suggestion
recommends ``iterations=3`` and ``memory_cost=64*1024``. See that document for
```
Like the previous suggestion

60x20 (changes requested):
I suggest using 2048*1024 rather than 2**21 to make it more human-readable just like it is on the [Argon2 docs](https://cryptography.io/en/latest/hazmat/primitives/key-derivation-functions/#cryptography.hazmat.primitives.kdf.argon2.Argon2id).
Also the PR title should probably be changed to something like "Recommend Argon2 over PBKDF2 as KDF"

docs/fernet.rst
tmo1: I'll obviously defer to whoever is in charge here, but the form I used is that of the RFC. I actually don't find either form that readable, since you have to remember that the figure is in KiB, and then mentally convert that into something like MiB or Gib :|

docs/fernet.rst
```diff
@@ -282,5 +284,5 @@ unsuitable for very large files at this time.
 
 
 .. _`Fernet`: https://github.com/fernet/spec/
-.. _`Django recommends as of January 2025`: https://github.com/django/django/blob/main/django/contrib/auth/hashers.py
 .. _`specification`: https://github.com/fernet/spec/blob/master/Spec.md
+ .._`IRTF RFC 9106`: https://datatracker.ietf.org/doc/html/rfc9106#name-parameter-choice
```
60x20:
```suggestion
.. _`IRTF RFC 9106`: https://datatracker.ietf.org/doc/html/rfc9106#name-parameter-choice
```
I believe the test is failing because of the missing space

reaperhulk (approved):

reaperhulk:
Thank you!

Issue #7847, opened 2022-11-26
[macOS] "New Mach-O header is too large to relocate" when using py2app


albertosottile:
### Description:
When creating an application bundle (`.app`) from Python code on macOS, some tools like [py2app](https://github.com/ronaldoussoren/py2app) rewrite the library linking to make the bundle self consistent and not linked to system libraries. 

This process works well with `cryptography` up to 38.0.1. However, version 38.0.3 raises this error in `py2app`
```
ValueError: New Mach-O header is too large to relocate in '/Users/runner/work/syncplay/syncplay/dist/Syncplay.app/Contents/Resources/lib/python3.10/lib-dynload/cryptography/hazmat/bindings/_rust.so' (new size=2264, max size=2260, delta=72)
```

This means that the header space allocated in the shared library is not enough to accommodate the new (probably longer) destination path. According to the [py2app FAQ](https://py2app.readthedocs.io/en/latest/faq.html) 

> This error can be avoided by rebuilding binaries with enough space in the Mach-O headers, either by using the linker flag “-headerpad_max_install_names” or by installing shared libraries in a deeply nested location (the path for the install root needs to be at least 30 characters long).

Indeed, I built `cryptography==38.0.3` from sources with
```
env LDFLAGS="-L$(brew --prefix openssl@3)/lib -Wl -headerpad_max_install_names" CFLAGS="-I$(brew --prefix openssl@3)/include" pip install --no-binary :all: cryptography
```

and I can confirm this fixes the issue in `py2app` and allows again to create bundles that depend on `cryptography`.

Would it be possible for you to add this linker flag in your build chain?

### Impact:
This bug prevents creation of any Python-based application bundle on macOS if the bundle itself depends on `cryptography`. Current workaround is to pin the dependency on 38.0.1. 

### Environment:
macOS 12.5
Python 3.10.6
cryptography 38.0.3
cffi 1.15.1
pip 22.3.1
setuptools 65.6.3

reaperhulk:
Fascinating. I don't have any issue with adding that flag to our wheel builders going forward, but I'm surprised 38.0.1 -> 38.0.3 caused this problem since no paths should have changed. The `_rust.abi3.so` we ship only links against CoreFoundation, libiconv, libSystem, and libresolv:

```
_rust.abi3.so (architecture x86_64):
	@rpath/_rust.abi3.so (compatibility version 0.0.0, current version 0.0.0)
[...]

PR #14777, opened 2026-05-01
fix: add -headerpad_max_install_names for macOS wheels

washingtoneg:
## What

Add `-headerpad_max_install_names` to the macOS linker invocation via `build.rs`.

## Why

macOS 15 CI runners use Apple's new linker (`ld_prime`), which reserves significantly less header padding in Mach-O binaries than the old `ld` did. Homebrew's post-install fixup (`install_name_tool`) needs that padding to rewrite dylib paths — when it's absent, the install fails:

```
Error: Failed changing dylib ID of .../_rust.abi3.so
Updated load commands do not fit in the header of .../_rust.abi3.so.
..._rust.abi3.so needs to be relinked, possibly with -headerpad or -headerpad_max_install_names
```

This regressed in `cryptography 46.0.4`, when the wheel-builder CI moved from `macos-13` to `macos-15` runners.

## Prior art

This was discussed in #7847 (2022). At the time, @reaperhulk said:

> "I don't have any issue with adding that flag to our wheel builders going forward."

The regression went away on its own then (a rustc bump changed the default padding). The macOS 15 runner shift has brought it back permanently.

## Fix

Three lines in `build.rs`, gated to macOS:

```rust
if env::var("CARGO_CFG_TARGET_OS").as_deref() == Ok("macos") {
    println!("cargo:rustc-link-arg=-Wl,-headerpad_max_install_names");
}
```

`cargo:rustc-link-arg` passes the flag directly to the final `cdylib` link step. This covers published wheels, source builds, and downstream Homebrew formulas in one shot.

### DIFF
--- a/src/rust/build.rs
+++ b/src/rust/build.rs
@@ -46,4 +46,8 @@ fn main() {
             println!("cargo:rustc-cfg=CRYPTOGRAPHY_OSSLCONF=\"{var}\"");
         }
     }
+
+    if env::var("CARGO_CFG_TARGET_OS").as_deref() == Ok("macos") {
+        println!("cargo:rustc-link-arg=-Wl,-headerpad_max_install_names");
+    }
 }
--- a/src/rust/build.rs
+++ b/src/rust/build.rs
@@ -47,6 +47,13 @@ fn main() {
         }
     }
 
+    // macOS 15 CI runners use Apple's new linker (ld_prime), which reserves
+    // significantly less Mach-O header padding than the old ld. Tools like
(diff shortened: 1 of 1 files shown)


reaperhulk:
This generally LGTM, but could we add a comment explaining why we have this branch and flag as well?

washingtoneg:
> This generally LGTM, but could we add a comment explaining why we have this branch and flag as well?

Thanks for the review @reaperhulk! I added a comment explaining why in 033bd31.

reaperhulk (approved):

PR #14782, opened 2026-05-02
Raise MemoryError on Argon2 allocation failure

### LINKED ISSUE #14778: Please handle memory allocation failure during Argon2 hashing more gracefully
My application uses memory locking on Linux ( mlockall() ) to protect secrets held in memory. This causes Argon2id operations (when using standard parameters) to crash when run under normal max locked memory ulimits (e.g., 8MiB on Debian).

Here's an MRE:

```
import os
from cryptography.hazmat.primitives.kdf.argon2 import Argon2id
from ctypes import CDLL
libc = CDLL(None)
libc.syscall(151, 2)
kdf = Argon2id(salt=os.urandom(16), length=32, iterations=3, lanes=4, memory_cost=2**16)
key = kdf.derive(os.urandom(16))
```
This crashes with:
```
                    when another library is not cleaning up the OpenSSL error                                                                                                                                     
                    stack. If you are using cryptography with another library                                                                                                                                     
                    that uses OpenSSL try disabling it before reporting a bug.                                                                                                                                    
                    Otherwise please file an issue at                                                                                                                                                             
                    https://github.com/pyca/cryptography/issues with                                                                                                                                              
                    information on how to reproduce this. (error:078C0100:common libcrypto routines::malloc failure:providers/implementations/kdfs/argon2.c:736:, error:1C8000EB:Provider routines:initialize:invalid memory size:providers/implementations/kdfs/argon2.c:743:cannot allocate required memory)
```
The final lines are clear enough, but it would be nice if a proper, documented [Cryptography exception](https://cryptography.io/en/latest/exceptions/), that could be checked for programmatically, would be thrown.

alex:
Closes #14778.

Argon2 derivations that exceed available memory (e.g. when running under `mlockall` with a low max-locked-memory ulimit, or when given an unreasonably large `memory_cost`) currently surface as an opaque `InternalError` like:

```
cryptography.exceptions.InternalError: Unknown OpenSSL error. ... (error:078C0100:common libcrypto routines::malloc failure:providers/implementations/kdfs/argon2.c:736:, error:1C8000EB:Provider routines:initialize:invalid memory size:providers/implementations/kdfs/argon2.c:743:cannot allocate required memory)
```

This mirrors `Scrypt`'s behavior: any error from the OpenSSL argon2 derive call is mapped to `MemoryError` with a message that includes the configured `memory_cost`, so callers can handle the failure programmatically.

In theory other init failures (e.g. `PROV_R_INVALID_THREAD_POOL_SIZE` on builds without thread-pool support) can also reach this path, but in practice the only realistic failure mode after our constructor validation is memory allocation, so the unconditional mapping matches what `Scrypt` does.

## Test plan

- [x] `cargo build` and `cargo clippy` clean
- [x] Added `test_argon2_malloc_failure` parameterized over Argon2d/i/id, modeled on `test_scrypt_malloc_failure`, using `memory_cost=2**32 - 1` (~4 TiB)
- [ ] CI exercises the new test on an OpenSSL build with argon2 support (local OpenSSL is 3.0.13)
- [x] Documented `MemoryError` in the `derive` / `derive_into` raises sections

---
_Generated by [Claude Code](https://claude.ai/code/session_01N5DJsHxpAxRpiNHnDGBfxT)_

### DIFF
--- a/tests/hazmat/primitives/test_argon2.py
+++ b/tests/hazmat/primitives/test_argon2.py
@@ -140,6 +140,19 @@ class TestArgon2:
                 memory_cost=memory_cost,
             )
 
+    def test_argon2_malloc_failure(self, clazz, backend):
+        # memory_cost is in KiB, so 2**32 - 1 KiB is ~4 TiB. This should
+        # fail to allocate on any reasonable system.
+        argon2 = clazz(
+            salt=b"salt" * 2,
+            length=32,
+            iterations=1,
+            lanes=1,
+            memory_cost=2**32 - 1,
+        )
+        with pytest.raises(MemoryError):
+            argon2.derive(b"password")
+
     def test_already_finalized(self, clazz, backend):
         argon2id = clazz(
             salt=b"salt" * 2, length=32, iterations=1, lanes=1, memory_cost=32
--- a/tests/hazmat/primitives/test_argon2.py
+++ b/tests/hazmat/primitives/test_argon2.py
@@ -6,6 +6,7 @@
 import base64
 import binascii
 import os
+import sys
 
 import pytest
 
@@ -140,6 +141,14 @@ class TestArgon2:
                 memory_cost=memory_cost,
             )
 
+    @pytest.mark.skipif(
+        sys.platform == "darwin",
+        reason=(
+            "macOS overcommits the 4 TiB virtual reservation, then the "
+            "subsequent memset triggers an OOM-kill before MemoryError "
+            "can be raised."
+        ),
+    )
     def test_argon2_malloc_failure(self, clazz, backend):
         # memory_cost is in KiB, so 2**32 - 1 KiB is ~4 TiB. This should
         # fail to allocate on any reasonable system.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Changelog
   ``TBSCertList.signature`` algorithm does not match the outer
   ``signatureAlgorithm`` now raises ``ValueError``. Previously, such CRLs
   were parsed successfully and only rejected during signature validation.
+* :class:`~cryptography.hazmat.primitives.kdf.argon2.Argon2id`,
+  :class:`~cryptography.hazmat.primitives.kdf.argon2.Argon2i`, and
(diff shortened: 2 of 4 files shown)


reaperhulk (approved):

mgorny:
How much memory is the added test expected to require? Because it OOMs on my 32 GiB Gentoo Linux amd64 system. Against OpenSSL 3.6.2, if that matters. Linux 7.0.3.

holdenweb:
And, as we all know, theory and practice are the same in theory, but can differ in practice. :)

alex:
Pretty sure it's 1TiB, which is to say it's expected to be rejected :-)

tests/hazmat/primitives/test_argon2.py
```diff
@@ -140,6 +141,27 @@ def test_invalid_values(self, clazz, params, backend):
...
+    @pytest.mark.skipif(
+        sys.platform == "darwin",
+        reason=(
+            "macOS overcommits the 4 TiB virtual reservation, then the "
+            "subsequent memset triggers an OOM-kill before MemoryError "
+            "can be raised."
+        ),
+    )
+    def test_argon2_malloc_failure(self, clazz, backend):
+        # memory_cost is in KiB, so 2**32 - 1 KiB is ~4 TiB. This should
```
Dr-Emann, line 153: I've run on a system with 4 TiB of memory. I'm sure there's systems with more. I'm not sure this is outrageous enough to guarantee failure.

Dr-Emann (commented):
If this fails correctly, or actually tries to use a ton of memory and gets oom killed depends on the overcommit_memory policy on Linux. https://www.kernel.org/doc/Documentation/vm/overcommit-accounting
In the default mode (0), it'll still refuse to allocate "obvious"ly too huge allocations, which I believe was the behavior expected by the test. In mode (2), it won't allow overcommit at all. In mode (1), it will always allow the allocation to succeed, then it'll presumably try to actually use that memory, and the OOM killer will take it out.

mgorny:
Indeed I have it to set 1. I'm not sure why exactly, but a quick search suggests that Redis expects that.

CertainLach:
Having overcommit enabled is not something unusual, tests should not expect it to be disabled, especially on linux (malloc is allowed to never fail, because of that the test is already explicitly disabled on macos)

PR #14787, opened 2026-05-03
pyemscripten/pyodide wheel builds

greateggsgreg:
Enable building pyemscripten/pyodide wheels. Now that https://peps.python.org/pep-0783/ is accepted, Pyodide/PyEmscripten wheels (specifically the Pyemscripten ABI (2026_0) ) are allowed to be published on PyPI.

This should also remove the need for pyodide-recipies to maintain a cryptography wheel build script out of tree.

Remaning tasks / discussion in #14788


alex:
We don't ship wheels for any platform for which we don't run tests, so that's the first step here if you're interested in pursuing this.

greateggsgreg:
> We don't ship wheels for any platform for which we don't run tests, so that's the first step here if you're interested in pursuing this.

Definitely interested seeing this through. Started #14788 to track discussion and remaining TODOs. Just pushed tests and looking to improve them.

alex:
To be clear, when we say running the tests, we mean as a part of our CI workflow, on every PR, with acceptable performance.

.github/workflows/wheel-builder.yml
```diff
@@ -387,3 +387,147 @@ jobs:
...
+
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.14"
+        timeout-minutes: 3
+
+      - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9
+        with:
+          # Pyodide's emscripten build passes nightly-only `-Z` flags
+          # (link-native-libraries, emscripten-wasm-eh).
```
hoodmane, line 408: It should be possible to build on stable soon, since neither of these flags are needed any longer. It seems like it might still be passing link-native-libraries=yes somewhere which is the default but it's still nightly only.

greateggsgreg: Wrote a TODO near this line and in ci.yml to make sure we validate if we still need nightly by the time pyodide 314.0 is released.

Also updated the issue task list to make sure we don't forget about this.

.github/workflows/ci.yml
```diff
@@ -398,6 +398,152 @@ jobs:
...
+
+      - name: Install emsdk (Pyodide-pinned)
+        # Pyodide's xbuildenv ships only patches, not emsdk binaries.
+        if: steps.emsdk-cache.outputs.cache-hit != 'true'
+        run: |
+          set -e
+          git clone --depth 1 https://github.com/emscripten-core/emsdk.git /opt/emsdk
+          cd /opt/emsdk
+          ./emsdk install "${EMSDK_VERSION}"
+          ./emsdk activate "${EMSDK_VERSION}"
```
agriyakhetarpal: We have a `pyodide xbuildenv install-emscripten` command that you can use. It installs a pinned version of Emscripten guaranteed to be compatible with the xbuildenv you install, and then patches it too.

.github/workflows/ci.yml
```diff
@@ -398,6 +398,152 @@ jobs:
...
+      - name: Create Pyodide venv and install wheel + test deps
+        run: |
+          set -e
+          pyodide venv .venv-pyodide
+          # shellcheck disable=SC1091
+          source .venv-pyodide/bin/activate
+          # pytest-xdist (no multiprocessing under Emscripten), pytest-cov
+          # (no Rust wasm coverage backend), and pytest-pyodide (host-side
+          # plugin; pulls playwright which has no wasm32 distribution) are
+          # deliberately omitted. pytest-benchmark satisfies pyproject.toml's
```
agriyakhetarpal: Yes, this is correct, but one thing I'll add is that `pytest-pyodide` is a helper plugin for testing Pyodide code via Playwright from a Linux environment, not in a Pyodide venv (and also obviously because Playwright doesn't have WASM builds) and we don't advertise it for out-of-tree testing. It would be better to leave it out here as well, just to be clearer this isn't going to be part of a testing environment.

pyproject.toml
```diff
@@ -179,6 +180,8 @@ exclude_lines = [
     "@abc.abstractmethod",
     "@typing.overload",
     "if typing.TYPE_CHECKING",
+    # coverage lacks Rust wasm backend
+    'if sys.platform == "emscripten":'
```
agriyakhetarpal: coverage and pytest-cov runs fine in other projects where we previously added Pyodide builds – is there an issue with cryptography in specific?

.github/workflows/ci.yml
greateggsgreg: Removed reference to this, thanks!

.github/workflows/ci.yml
greateggsgreg: Appreciate the suggestion, fixed

pyproject.toml
greateggsgreg: Fixed, thanks - was running into issues before but looks like things are fine now

.github/workflows/ci.yml
```diff
@@ -398,6 +398,144 @@ jobs:
 
       - uses: ./.github/actions/upload-coverage
 
+  pyemscripten:
```
alex, line 401: I've taken a read through of this, and here is my concern: this is basically a significant amount of configuration, that none of hte pyca/cryptography knows anything about. We don't know when versions should be bumped, don't know what compatibility looks like, don't know how to debug. It duplicates a ton of our other CI code as well.

For this to be a realistic thing we were willing to maintain, we need to get this down much closer to "this is a new entry in the existing linux matrix that shares the vast majority of the code with the other CI workflows"

greateggsgreg: Thanks for the read. Maintainability concerns are fair. Two paths I'd like your call on before cutting code:

**Option A — Fold into the linux matrix**

New row:
```yaml
- VERSION: "3.14"
  NOXSESSION: "tests-pyemscripten"
  OPENSSL: {TYPE: "pyemscripten", VERSION: "3.6.2"}
  PYODIDE: {VERSION: "314.0.0a1", EMSDK: "5.0.3"}
  RUST: "nightly"
  RUST_TARGETS: "wasm32-unknown-emscripten"
```

Pros: checkout/setup-python/rust-toolchain/cache/upload-coverage reuse linux's steps; YAML shrinks; pyemscripten appears as a row in the linux matrix.
Cons: linux job grows 4-5 `if: matrix.PYTHON.PYODIDE` branches (install + Tests must diverge — nox can't drive a pyodide venv since uv has no wasm build).

**Option B — Standalone job, shrunk via shared building blocks**

- New composite action `.github/actions/setup-pyodide-build/` for the `pyodide-build[resolve]` + `xbuildenv install` + `install-emscripten` trio.
- Existing `build_openssl.sh` (TYPE=pyemscripten already supported).
- New `.github/bin/test_pyemscripten.sh` for `pyodide build .` → `pyodide venv` → install wheel → `pytest --cov`. Same script runs locally.
- Local reproducer (Containerfile + driver, already prototyped) so a maintainer can repro CI's pipeline with one command.

Pros: YAML shrinks; no conditional proliferation; unique logic in one action + one script, both grep-able and locally runnable.
Cons: still its own job rather than a row in the linux matrix.

I lean B — fewer conditionals, better local-debug story — but it's your call.

**Independent of A vs B**, on the maintainability gap: I'd add `docs/development/pyemscripten.rst` covering the Pyodide ↔ Emscripten ↔ Python compatibility triple (and where Pyodide pins it in `Makefile.envs`), bump cadence, and 4-5 failure-triage entries. I'll do this regardless of which option we go with.

What do you think?

agriyakhetarpal: I think one option could be to just use cibuildwheel in all workflow files, as we ensure in our Pyodide platform code that the right Emscripten toolchain is picked up, and we won't need to worry about the pyodide-build version, etc. All dependencies are pinned/constrained, too, making it more reliable and reducing the bespoke configuration involved. The only thing to pick up and update would be based on changes between cibuildwheel releases, i.e., which Pyodide ABIs are available to build and whether to enable or disable them as needed. That said, I'd recommend building only for the 2026 ABI, as the 314.0.0 stable release is coming soon ([issue](https://github.com/pyodide/pyodide/issues/6233), [milestone](https://github.com/pyodide/pyodide/milestone/37)). The second alpha is out as of two days ago.

alex:
See https://github.com/pyca/cryptography/issues/14788#issuecomment-4640198941 for why I'm closing.

PR #14794, opened 2026-05-04
Add AGENTS.md with CLAUDE.md symlink

reaperhulk:
Document project layout and the canonical commands (nox -e local, uv) for agent-based tooling. Exclude both files from the sdist via check-sdist's git-only list.

you're welcome Alex

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,7 +4,7 @@ Hybrid Python + Rust project. The Python package lives in `src/cryptography/`; t
 
 ## Running tests
 
-Use `nox -e local` to run the full local check: it formats and lints Python (`ruff`) and Rust (`cargo fmt`/`check`/`clippy`), type-checks with `mypy`, then runs the pytest suite and `cargo test`. This is the canonical command — prefer it over invoking `pytest` or `cargo test` directly.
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -0,0 +1,17 @@
+# cryptography
+
+Hybrid Python + Rust project. The Python package lives in `src/cryptography/`; the Rust extension (PyO3) lives in `src/rust/`. CFFI bindings to OpenSSL/BoringSSL/AWS-LC are generated from `src/_cffi_src/`. Test vectors live in `vectors/` (an installable sub-package).
+
+## Running tests
+
+Use `nox -e local` to run the full local check: it formats and lints Python (`ruff`) and Rust (`cargo fmt`/`check`/`clippy`), type-checks with `mypy`, then runs the pytest suite and `cargo test`. This is the canonical command — prefer it over invoking `pytest` or `cargo test` directly.
```
alex:
```suggestion
Use `nox -e local` to run the full local check: it formats and lints Python (`ruff`) and Rust (`cargo fmt`/`check`/`clippy`), type-checks with `mypy`, then runs the pytest suite and `cargo test`. This is the canonical command — never invoke `pytest` or `cargo test` directly.
```

alex (approved):

dadrian (approved):

PR #14798, opened 2026-05-04
Remove unused CFFI bindings without specific consumer

alex:
Drops bindings whose git history shows no specific user/consumer request — they were added in bulk during the cffi 1.0 conversion or as generic library setup, and are no longer referenced by cryptography itself. Bindings that were explicitly requested by external consumers (pyOpenSSL, mitmproxy, Tribler, etc.) are retained.

Also removes the now-unused USE_CONST_X509 env var plumbing in build.rs, since the X509V3_CTX struct that depended on it is gone.


alex:
@mhils there's a few cffi things that mitmproxy is poking around for -- can we add proper pyopenssl APIs for those so mitmproxy isn't reliant on implementation details that will break

reaperhulk (approved):

PR #14799, opened 2026-05-05
Fix publish

woodruffw:
https://github.com/pyca/cryptography/actions/runs/25347944931/job/74321055630 failed because `ci-constraints-requirements.txt` wasn't checked out.

This checks it out, which is sad because checkouts are slow. We could possibly make it a sparse checkout to be a little faster, although not sure if it matters in the publish job.

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -51,6 +51,8 @@ jobs:
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
+          sparse-checkout: ci-constraints-requirements.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi-publish.yml
```diff
@@ -49,6 +49,10 @@ jobs:
           python-version: '3.14'
         timeout-minutes: 3
 
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
```
alex, line 56: can you add
```
          sparse-checkout: <path>
          sparse-checkout-cone-mode: false
```

alex (approved):

PR #14819, opened 2026-05-06
fix: use runtime probe to select AES-GCM context strategy under FIPS

### LINKED ISSUE #14795: AESGCM fails on RHEL9 in FIPS mode
Precondition: Have RHEL9 running in FIPS mode and cryptography 46

Small unit test minimal example:

```
def test_aes_gcm_encrypt():
        from cryptography.hazmat.primitives.ciphers.aead import AESGCM
        import os
    
        key = os.urandom(16)
        aesgcm = AESGCM(key)
    
        iv = os.urandom(12)
>       encrypted = aesgcm.encrypt(iv, b"I am a test", None)
E       cryptography.exceptions.InternalError: Unknown OpenSSL error. This error is commonly encountered
E                           when another library is not cleaning up the OpenSSL error
E                           stack. If you are using cryptography with another library
E                           that uses OpenSSL try disabling it before reporting a bug.
E                           Otherwise please file an issue at
E                           https://github.com/pyca/cryptography/issues with
E                           information on how to reproduce this. (error:030000BE:digital envelope routines:EVP_CIPHER_CTX_copy:not able to copy ctx:crypto/evp/evp_enc.c:1792:)
```

This has been test with `46.0.7` but is likely a regression added by 98dfafeb8dcdf5e640c9612841f07da66586509b. The version 41.0.7 from the Red Hat package python3.12-cryptography does not exhibit this problem.

A possible workaround is to use `aesgcm = Cipher(algorithms.AES(key), modes.GCM(iv)).encryptor()` instead as that still works.

abbra:
EvpCipherAead pre-initialises a cipher context with the key and then clones it per operation via EVP_CIPHER_CTX_copy.  When an older FIPS provider (e.g. OpenSSL 3.0.7 FIPS module) is loaded alongside a newer main library (>= 3.2), EVP_CIPHER_CTX_copy may fail because that provider version does not support copying a pre-keyed context.

Instead of relying on FIPS as an imperfect proxy, introduce aes_gcm_ctx_copy_supported(), which probes the capability directly by initialising a dummy AES-128-GCM context and attempting to copy it. The result is cached in a OnceLock so the probe runs at most once per process.  On OpenSSL >= 3.2, AesGcm::new() consults the cache and falls back to LazyEvpCipherAead (fresh context per operation) when the copy is not supported.

The new AesGcmCtx enum holds either implementation so the choice can be made at runtime rather than baked in at compile time.  AES_GCM_TAG_LEN replaces the magic 16 literals throughout the AesGcm methods.

Fixes: https://github.com/pyca/cryptography/issues/14795

Assisted-by: Claude Code (Sonnet 4.6)

### DIFF
--- a/src/rust/src/backend/aead.rs
+++ b/src/rust/src/backend/aead.rs
@@ -31,9 +31,14 @@ pub(crate) struct EvpCipherAead {
     base_decryption_ctx: openssl::cipher_ctx::CipherCtx,
     tag_len: usize,
     tag_first: bool,
+    // Cipher and key stored to reinitialise a fresh context when
+    // EVP_CIPHER_CTX_copy fails (e.g. an older FIPS provider loaded alongside
+    // a newer main library).  None for algorithms where copy always succeeds.
+    copy_fallback: Option<(&'static openssl::cipher::CipherRef, Vec<u8>)>,
 }
 
 impl EvpCipherAead {
+    #[cfg(not(any(CRYPTOGRAPHY_IS_BORINGSSL, CRYPTOGRAPHY_IS_AWSLC)))]
     pub(crate) fn new(
         cipher: &openssl::cipher::CipherRef,
         key: &[u8],
@@ -50,6 +55,33 @@ impl EvpCipherAead {
             base_decryption_ctx,
             tag_len,
             tag_first,
+            copy_fallback: None,
+        })
+    }
+
+    #[cfg(any(
+        CRYPTOGRAPHY_OPENSSL_320_OR_GREATER,
+        CRYPTOGRAPHY_IS_BORINGSSL,
+        CRYPTOGRAPHY_IS_LIBRESSL,
+        CRYPTOGRAPHY_IS_AWSLC
+    ))]
+    pub(crate) fn new_with_copy_fallback(
+        cipher: &'static openssl::cipher::CipherRef,
+        key: &[u8],
+        tag_len: usize,
+        tag_first: bool,
+    ) -> CryptographyResult<EvpCipherAead> {
+        let mut base_encryption_ctx = openssl::cipher_ctx::CipherCtx::new()?;
+        base_encryption_ctx.encrypt_init(Some(cipher), Some(key), None)?;
+        let mut base_decryption_ctx = openssl::cipher_ctx::CipherCtx::new()?;
+        base_decryption_ctx.decrypt_init(Some(cipher), Some(key), None)?;
+
+        Ok(EvpCipherAead {
+            base_encryption_ctx,
+            base_decryption_ctx,
+            tag_len,
+            tag_first,
+            copy_fallback: Some((cipher, key.to_vec())),
         })
     }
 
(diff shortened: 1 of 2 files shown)


schwabe:
I tested this on my RHEL9 distribution running in FIPS mode and this PR fixes the problem for me.

src/rust/src/backend/aead.rs
```diff
@@ -622,6 +622,105 @@ impl ChaCha20Poly1305 {
...
+    static COPY_SUPPORTED: OnceLock<bool> = OnceLock::new();
+    *COPY_SUPPORTED.get_or_init(|| {
+        // Test hook: force the lazy-context fallback without a real FIPS
+        // provider that rejects EVP_CIPHER_CTX_copy.
+        if std::env::var_os("CRYPTOGRAPHY_TEST_DISABLE_AES_GCM_CTX_COPY").is_some() {
+            return false;
+        }
+        let key = [0u8; 16];
+        let Ok(mut ctx) = openssl::cipher_ctx::CipherCtx::new() else {
+            // NO-COVERAGE-START
```
alex: We never use coverage ignroes for real code -- they are only acceptable for macro-code that's uncoverable.

src/rust/src/backend/aead.rs
```diff
@@ -713,7 +838,7 @@ impl AesGcm {
         check_length(data_bytes)?;
         Ok(pyo3::types::PyBytes::new_with(
             py,
-            data_bytes.len() + self.ctx.tag_len,
+            data_bytes.len() + AES_GCM_TAG_LEN,
```
alex: is this change at all related?

src/rust/src/backend/aead.rs
```diff
@@ -622,6 +622,105 @@ impl ChaCha20Poly1305 {
...
+        CRYPTOGRAPHY_IS_LIBRESSL,
+        CRYPTOGRAPHY_IS_BORINGSSL,
+        CRYPTOGRAPHY_IS_AWSLC
+    ))
+))]
+fn aes_gcm_ctx_copy_supported() -> bool {
+    use std::sync::OnceLock;
+    static COPY_SUPPORTED: OnceLock<bool> = OnceLock::new();
+    *COPY_SUPPORTED.get_or_init(|| {
+        // Test hook: force the lazy-context fallback without a real FIPS
```
alex: This isn't an acceptable testing strategy.

alex (changes requested):
this whole approach seems overly fragile to me. If we want to support this (and honestly I still think this is a dumb thing to support), it seems much simpler to have the EvpCipherAead simply store sufficient parameters that if copy() fails it just allocates a new EvpCipher for it. That'd make the code much simpler, and also remove the needs for these hacky tests

src/rust/src/backend/aead.rs
```diff
@@ -50,6 +55,33 @@ impl EvpCipherAead {
...
+        })
+    }
+
+    #[cfg(any(
+        CRYPTOGRAPHY_OPENSSL_320_OR_GREATER,
+        CRYPTOGRAPHY_IS_BORINGSSL,
+        CRYPTOGRAPHY_IS_LIBRESSL,
+        CRYPTOGRAPHY_IS_AWSLC
+    ))]
+    pub(crate) fn new_with_copy_fallback(
```
alex, line 68: Why is this optional behavior instead of just being the default?

If this behavior was comprehensive, we shouldn't need the Lazy type at all, right?

abbra: Lazy type is not used with OpenSSL.

FIPS validation is a long process, so for some time old validated tokens will be available along with upgraded OpenSSL version. Some downstreams have their own FIPS validated tokens available before the token passes the validation, so we have to deal with presence of both old and new at the same time. However, it is not something that would change over the runtime of a single application -- multiple app executions could have different tokens in their environments, though.

reaperhulk:
https://github.com/pyca/cryptography/pull/15004 fixes this

PR #14830, opened 2026-05-09
Simplify ML-DSA and ML-KEM AWS-LC paths using AWS-LC 1.73.0 APIs

alex:
AWS-LC 1.73.0 added three APIs that let us drop AWS-LC-specific
workarounds in our ML-DSA and ML-KEM code. CI is already pinned to
v1.73.0.

* **`EVP_PKEY_CTX_set1_signature_context_string`** ([aws/aws-lc#3135](https://github.com/aws/aws-lc/pull/3135)) for ML-DSA
  EVP sign/verify. Previously the EVP PQDSA path hardcoded the context
  to `(NULL, 0)`, so AWS-LC had to call the raw `ml_dsa_{44,65,87}_sign`
  / `_verify` functions directly. AWS-LC now joins the BoringSSL /
  OpenSSL 3.5 `EVP_DigestSign` / `EVP_DigestVerify` path, removing
  ~150 lines of `extern "C"` declarations, hardcoded signature-size
  tables and manual error-stack draining.

* **`EVP_PKEY_kem_get_type`** ([aws/aws-lc#3179](https://github.com/aws/aws-lc/pull/3179)) returns the KEM key NID, so
  `MlKemVariant::from_pkey` can do a direct NID lookup on AWS-LC
  instead of discriminating ML-KEM-768 vs ML-KEM-1024 by public-key
  byte length.

* **`EVP_PKEY_get_private_seed`** ([aws/aws-lc#3200](https://github.com/aws/aws-lc/pull/3200)) for KEM and PQDSA
  keys, with the same signature as BoringSSL's. The AWS-LC PKCS#8
  round-trip in `mldsa_seed_from_pkey` / `mlkem_seed_from_pkey` is
  gone — both wrappers fold into a single helper call. Since the
  five backend callers only need the raw seed bytes, the wrappers are
  removed entirely; the two ASN.1 serialization sites construct
  `MlKemPrivateKey::Seed(...)` / `MlDsaPrivateKey::Seed(...)` inline.

Net change across the three commits: **+45 / −283**.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PZKqdgDtqU9eSf3Xsv2cVF)_

### DIFF
--- a/src/rust/cryptography-openssl/src/mldsa.rs
+++ b/src/rust/cryptography-openssl/src/mldsa.rs
@@ -136,16 +136,17 @@ fn set_context_string<T>(
 }
 
 /// Extract the raw 32-byte seed from an ML-DSA private key.
-///
-/// Avoids the PKCS#8 round-trip that vanilla OpenSSL 3.5 encodes
-/// differently from BoringSSL/AWS-LC.
-#[cfg(any(CRYPTOGRAPHY_IS_BORINGSSL, CRYPTOGRAPHY_OPENSSL_350_OR_GREATER))]
+#[cfg(any(
+    CRYPTOGRAPHY_IS_BORINGSSL,
+    CRYPTOGRAPHY_IS_AWSLC,
+    CRYPTOGRAPHY_OPENSSL_350_OR_GREATER
+))]
 pub fn mldsa_seed_raw(
     pkey: &openssl::pkey::PKeyRef<openssl::pkey::Private>,
 ) -> OpenSSLResult<[u8; 32]> {
     let mut seed = [0u8; 32];
     cfg_if::cfg_if! {
-        if #[cfg(CRYPTOGRAPHY_IS_BORINGSSL)] {
+        if #[cfg(any(CRYPTOGRAPHY_IS_BORINGSSL, CRYPTOGRAPHY_IS_AWSLC))] {
             let mut seed_len = seed.len();
             // SAFETY: pkey is a valid EVP_PKEY and seed is a 32-byte buffer.
             unsafe {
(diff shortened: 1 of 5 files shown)


src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -136,83 +135,18 @@ fn set_context_string<T>(
...
 /// Extract the raw 32-byte seed from an ML-DSA private key.
-///
-/// Avoids the PKCS#8 round-trip that vanilla OpenSSL 3.5 encodes
-/// differently from BoringSSL/AWS-LC.
-#[cfg(any(CRYPTOGRAPHY_IS_BORINGSSL, CRYPTOGRAPHY_OPENSSL_350_OR_GREATER))]
+#[cfg(any(
+    CRYPTOGRAPHY_IS_BORINGSSL,
+    CRYPTOGRAPHY_IS_AWSLC,
+    CRYPTOGRAPHY_OPENSSL_350_OR_GREATER
+))]
```
alex: this is a no-op, the entire file has this condition

src/rust/cryptography-openssl/src/mlkem.rs
```diff
@@ -120,16 +116,17 @@ extern "C" {
...
 /// Extract the raw 64-byte seed from an ML-KEM private key.
-///
-/// Avoids the PKCS#8 round-trip that vanilla OpenSSL 3.5 encodes
-/// differently from BoringSSL/AWS-LC.
-#[cfg(any(CRYPTOGRAPHY_IS_BORINGSSL, CRYPTOGRAPHY_OPENSSL_350_OR_GREATER))]
+#[cfg(any(
+    CRYPTOGRAPHY_IS_BORINGSSL,
+    CRYPTOGRAPHY_IS_AWSLC,
+    CRYPTOGRAPHY_OPENSSL_350_OR_GREATER
+))]
```
alex: same

reaperhulk (approved):

PR #14851, opened 2026-05-13
Lower cffi version requirement from >=2.0.0 to >=1.16

frasertweedale:
Some distros don't have cffi >= 2.0.0 yet.  v1.16 still works, so
reduce the lower bound.

### DIFF
--- a/.github/requirements/build-requirements.in
+++ b/.github/requirements/build-requirements.in
@@ -1,6 +1,7 @@
 # Must be kept sync with build-system.requires at pyproject.toml
 setuptools!=74.0.0
-cffi>=2.0.0; platform_python_implementation != 'PyPy'
+cffi>=1.16; python_version < '3.13' and platform_python_implementation != 'PyPy'
+cffi>=2.0.0; python_version >= '3.13' and platform_python_implementation != 'PyPy'
 maturin>=1,<2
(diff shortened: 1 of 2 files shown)


alex:
What's the motivation for this? Distros not having cffi 2.0 is not a sufficient motivation: mixing old deps with new ones is better solved by distros updating their cffis.

Further, this now gives you the wrong behavior on free-threaded Pythons, where it permits you to install a cffi that doesn't support it.

frasertweedale:
The motivation is: RHEL has old cffi for now, and it's one less thing to patch.
I added environment marker to restrict free-threaded python versions to cffi >= 2;
thanks for pointing that out.

alex:
If RHEL wants to use a very recent cryptography, why can't you upgrade your cffi?

frasertweedale:
@alex because change introduces risk; so if things works fine with the earlier cffi version, I'd prefer not to.

Offering it upstream because it would be one less thing to patch on our side.  If you're not keen on it, that's OK, just close the PR and no hard feelings :)

reaperhulk:
We're going to keep it at cffi 2 (but we're also have a goal of removing cffi entirely so this will hopefully become moot). Thanks though!

PR #14887, opened 2026-05-21
[CI] Move Windows CI/wheel jobs into composite actions

khmyznikov:
Extracts the inline step sequences of the `windows` jobs in `ci.yml` and `wheel-builder.yml` into two reusable composite actions:

- `.github/actions/windows-tests/` — setup Python + Rust, fetch OpenSSL, install nox, fetch test vectors, run the test suite.
- `.github/actions/windows-wheel/` — setup Python + Rust, fetch OpenSSL, build the wheel with `uv build`, smoke-test, upload artifact.

The `windows` jobs in both workflows now consist of `checkout` + a single `uses:` step, parameterized by `arch`, `openssl-name`, etc.

### Motivation

Carved out from #14346 per maintainer feedback to keep the refactor and the new arm64 platform as separate changes. Adding a new Windows architecture (or any future variant) becomes a few lines of YAML instead of duplicating ~60–80 lines of steps.

### Notes

- Pure refactor — no behavior change for the existing `x64` matrix.
- Composite actions follow the same convention as existing `.github/actions/cache/`, `fetch-vectors/`, `wheel-smoketest/`, `upload-coverage/`.
- `secrets.GITHUB_TOKEN` is replaced with `github.token` inside the composite actions (composite steps can't read `secrets.*` directly).

### DIFF
--- a/.github/actions/windows-tests/action.yml
+++ b/.github/actions/windows-tests/action.yml
@@ -1,0 +1,70 @@
+name: Windows tests
+description: |
+  Runs the cryptography test suite on Windows. Used by ci.yml for both
+  x64 and arm64 jobs to avoid duplication.
+
+inputs:
+  python-version:
+    description: "Python version to test (e.g. 3.9, 3.14, 3.14t)"
+    required: true
+  noxsession:
+    description: "Nox session to run (e.g. tests, tests-nocoverage)"
+    required: true
+  arch:
+    description: "actions/setup-python architecture (x86, x64, arm64)"
+    required: true
+  openssl-name:
+    description: "OpenSSL artifact name suffix (win32, win64, arm64)"
+    required: true
+
+runs:
+  using: "composite"
+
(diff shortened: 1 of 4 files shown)


khmyznikov:
@alex as discussed 👍

alex:
It looks like this loses the timeouts we had on several of the steps. I'm assuming there's a reason for this?

khmyznikov:
@alex yeah sadly it's known limitation https://github.com/actions/runner/issues/1979. But there's still `timeout-minutes: 15` on upper level

alex:
Why!!!

The windows wheel builder doesn't have a top-level timeout, it's only got the step-specific ones. Can you add a top-level timeout on windows there (15 minutes as well, I guess).

khmyznikov:
@alex oh jeez, looked at wrong place.

alex (approved):

PR #14898, opened 2026-05-22
Add support for PEP 695 type aliases

### LINKED ISSUE #14892: asn1: Support type aliases
While working with the new asn1 module i stubled upon one limitation when using type aliases. For example this excerpt from the documentation:

```python
@asn1.sequence
class Validity:
    not_before: asn1.UTCTime | asn1.GeneralizedTime
    not_after: asn1.UTCTime | asn1.GeneralizedTime
```

I would be nice to instead write:

```python
type Time = asn1.UTCTime | asn1.GeneralizedTime

@asn1.sequence
class Validity:
    not_before: Time
    not_after: Time
```

However this is currently not supported by cryptography. I think it would be nice to also support type aliases, so common combinations can be declared and reused.

loewexy:
Fixes #14892

However, currently the implementation is missing tests for that specific feature.

### DIFF
--- a/src/cryptography/hazmat/asn1/asn1.py
+++ b/src/cryptography/hazmat/asn1/asn1.py
@@ -106,6 +106,12 @@ def _extract_annotation(
 def _normalize_field_type(
     field_type: typing.Any, field_name: str
 ) -> declarative_asn1.AnnotatedType:
+    # Resolve PEP 695 `type X = ...` aliases (Python 3.12+) to their
+    # underlying type before any further processing.
+    if sys.version_info >= (3, 12):
+        while isinstance(field_type, typing.TypeAliasType):
+            field_type = field_type.__value__
+
     # Strip the `Annotated[...]` off, and populate the annotation
     # from it if it exists.
     if typing.get_origin(field_type) is typing.Annotated:
@@ -113,6 +119,13 @@ def _normalize_field_type(
         field_type, *_ = typing.get_args(field_type)
     else:
         annotation = declarative_asn1.Annotation()
+
+    # Resolve again: The Annotated wrapper may have hidden a
+    # TypeAliasType, or the first resolution may have exposed an
+    # Annotated wrapper whose inner type is itself a TypeAliasType.
+    if sys.version_info >= (3, 12):
+        while isinstance(field_type, typing.TypeAliasType):
+            field_type = field_type.__value__
 
     if annotation.size is not None and (
         typing.get_origin(field_type) not in (builtins.list, SetOf)


alex:
@facutuesca do you have a) opinions on if this is the right place for this? b) thoughts on how to test?

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -106,6 +106,12 @@ def _extract_annotation(
 def _normalize_field_type(
     field_type: typing.Any, field_name: str
 ) -> declarative_asn1.AnnotatedType:
+    # Resolve PEP 695 `type X = ...` aliases (Python 3.12+) to their
+    # underlying type before any further processing.
+    if sys.version_info >= (3,12):
+        while isinstance(field_type, typing.TypeAliasType):
```
facutuesca: This could also be present in previous versions by using [`typing_extensions.TypeAliasType`](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.TypeAliasType). See the top of the file for how we handle conditionally importing from `typing` vs `typing_extensions` for different Python versions.

loewexy:
I have to setup a complete development environment for cryptography to test so that i can also run the linters and stuff, however i will not manage to do this today. But I will try to do it next week. Otherwise also feel free to use my draft and update it as you like.

facutuesca:
> b) thoughts on how to test?

If the only concern about testing is the newer Python versions needed for the feature (mentioned [here](https://github.com/pyca/cryptography/issues/14892#issuecomment-4518488396)), see my review comment about supporting `typing_extensions.TypeAliasType` for older Python versions.

loewexy:
So no i have a local test setup an fixed the linter errors. On my machine also the existing tests run without a problem. However adding a test for that case an maybe also the edge cases in combination with annotated would make sense.

src/cryptography/hazmat/asn1/asn1.py
Viicos: While I'd assume cryptography can't easily add a new dependency, we provide and use https://typing-inspection.pydantic.dev/latest/ in Pydantic. It takes care of all the possible variants that are easy to miss when dealing with `typing` objects (in particular `TypeAliasType`, see our [implementation](https://github.com/pydantic/typing-inspection/blob/2afb37ca85617282877828bb8a118e60903830cd/src/typing_inspection/typing_objects.py#L491-L499)). The [usage guide](https://typing-inspection.pydantic.dev/latest/usage/) provides useful information too.

alex:
I think the best way to test this would be a pytest test that uses exec/eval to put the code in a string and not run on older python versions.

PR #14904, opened 2026-05-26
cffi: derive Python include dir from PYO3_CROSS_LIB_DIR when cross-compiling

commodo:
When cross-compiling, ``cryptography-cffi/build.rs`` queries the host interpreter via ``setuptools.command.build_ext.build_ext.include_dirs`` and feeds the result into ``cc-rs`` as ``-I /usr/include/python3.X``. On hosts that happen to ship Python development headers for the same 3.X line as the target (e.g. an Arch Linux build host with Python 3.14 installed), the host headers leak into the cross build and the target compiler aborts with::

    /usr/include/python3.14/pyport.h:429:2: error: #error "LONG_BIT
    definition appears wrong for platform (bad gcc/glibc config?)."

``PYO3_CROSS_LIB_DIR`` is the documented PyO3 cross-compile signal and points at the target's libpython directory. When it is set, derive the matching Python include dir from it (``<prefix>/include/<py_ver>``) and skip the host setuptools probe entirely. Native builds keep the existing setuptools-based behaviour, so this is a no-op outside of PyO3-style cross builds.

This is consistent with the existing ``PYO3_PYTHON`` handling in the same file: cryptography-cffi already respects PyO3's cross-compile environment for picking the interpreter, and this extends that to the matching header directory.

### DIFF
--- a/src/rust/cryptography-cffi/build.rs
+++ b/src/rust/cryptography-cffi/build.rs
@@ -56,16 +56,40 @@ fn main() {
     )
     .unwrap();
     println!("cargo:rustc-cfg=python_implementation=\"{python_impl}\"");
-    let python_includes = run_python_script(
-        &python,
-        "import os; \
-         import setuptools.dist; \
-         import setuptools.command.build_ext; \
-         b = setuptools.command.build_ext.build_ext(setuptools.dist.Distribution()); \
-         b.finalize_options(); \
-         print(os.pathsep.join(b.include_dirs), end='')",
-    )
-    .unwrap();
+    println!("cargo:rerun-if-env-changed=PYO3_CROSS_LIB_DIR");
+    // When cross-compiling, PyO3 expects the build system to point
+    // PYO3_CROSS_LIB_DIR at the target's libpython directory. Derive the
+    // matching include dir from it instead of querying the host
+    // interpreter's setuptools, which returns host headers (e.g.
+    // /usr/include/python3.x) and breaks the cross build whenever the host
+    // happens to have same-version Python development headers installed.
+    let python_includes = if let Ok(lib_dir) = env::var("PYO3_CROSS_LIB_DIR") {
+        let lib = Path::new(&lib_dir);
+        let py_ver = lib
+            .file_name()
+            .and_then(|s| s.to_str())
+            .unwrap_or("python3");
+        let prefix = lib
+            .parent()
+            .and_then(|p| p.parent())
+            .expect("PYO3_CROSS_LIB_DIR has unexpected layout");
+        prefix
+            .join("include")
+            .join(py_ver)
+            .to_string_lossy()
+            .into_owned()
+    } else {
+        run_python_script(
+            &python,
+            "import os; \
+             import setuptools.dist; \
+             import setuptools.command.build_ext; \
+             b = setuptools.command.build_ext.build_ext(setuptools.dist.Distribution()); \
+             b.finalize_options(); \
+             print(os.pathsep.join(b.include_dirs), end='')",
+        )
+        .unwrap()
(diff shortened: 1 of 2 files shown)


commodo:
For some context: https://github.com/openwrt/packages/issues/29521

It started with an issue in OpenWrt

src/rust/cryptography-cffi/build.rs
```diff
@@ -56,16 +56,40 @@ fn main() {
...
+            .unwrap_or("python3");
+        let prefix = lib
+            .parent()
+            .and_then(|p| p.parent())
+            .expect("PYO3_CROSS_LIB_DIR has unexpected layout");
+        prefix
+            .join("include")
+            .join(py_ver)
+            .to_string_lossy()
+            .into_owned()
```
alex, line 80: I'm not 100% sure I understand the path logic that's intended here, can you explain?

commodo: for context, cross-compiling Python and Python packages requires we build the Python  interpreter twice (for the same version); one for the host-side (and to be used during cross-build) and the target Python itself;
for some packages, we also need to do 2 builds; but some packages are host-only (like `maturin` or `pyo3` for building target side packages);

// if this is know info, apologies for the noise;

in OpenWrt's build system, we already set  something like `PYO3_CROSS_LIB_DIR=/builder/staging_dir/target-mips_24kc_musl/usr/lib/python3.14`
we set it here: https://github.com/openwrt/packages/blob/master/lang/python/python3-package.mk#L66

this is the directory where Python3.14 for the target system (MIPS24 in this case), installs target-side Python libraries
the logic is a bit specific;
short explanation of a flow is:
```
  let lib = Path::new(&lib_dir);
  // lib    = /builder/staging_dir/target-mips_24kc_musl/usr/lib/python3.14
  
  let py_ver = lib.file_name().and_then(|s| s.to_str()).unwrap_or("python3");
  // py_ver = "python3.14"   (the last component, preserved verbatim — also handles
  //                          free-threaded names like "python3.14t" without parsing)

  let prefix = lib.parent().and_then(|p| p.parent())
                  .expect("PYO3_CROSS_LIB_DIR has unexpected layout");
  // .parent()        = /builder/staging_dir/target-mips_24kc_musl/usr/lib
  // .parent().parent() = /builder/staging_dir/target-mips_24kc_musl/usr   ← the FHS "prefix"

  prefix.join("include").join(py_ver).into_owned()
  // result = /builder/staging_dir/target-mips_24kc_musl/usr/include/python3.14
```

I'll admit this logic is quirky at best.
And it's a dumb attempt at fixing the issue with just having `PYO3_CROSS_LIB_DIR`
Looking at https://pyo3.rs/v0.28.3/building-and-distribution  (latest IIRC?) there's no `PYO3_CROSS_INCLUDE_DIR`  which would have made things waaaaay easier.

Though, I can go that route and add a `PYO3_CROSS_INCLUDE_DIR` here and we just use that.

As you can tell, this needs to solve finding the correct Python headers which can fail if they find 32 bit headers on 64-bit systems or vice-versa.

alex: Is this directory hierarchy guaranteed to be the same for all cross-compilation setups?

commodo: for most embedded systems i know of, it's the same;

but re-thinking about `PYO3_CROSS_INCLUDE_DIR` ;  would that be something worth pushing for also for `pyo3`?
it looks the less quirky-est

alex: I don't have a strong opinion on that.

alex (approved):

PR #14934, opened 2026-06-03
count characters not bytes for UTF8String SIZE constraints

dxbjavid:
decode_pystr and the PyStr encode arm measure SIZE against `str.len()`, which is the UTF-8 byte count rather than the character count X.680 51.5 specifies. A two-character value like "é€" is five bytes, so it slips past `Size(min=3)` on decode and is wrongly rejected under `Size(max=2)` on encode. Count with `chars().count()` instead; BIT STRING already uses bit count and SEQUENCE OF uses element count, and PrintableString/IA5String are ASCII so they're unaffected.


alex (approved):
good catch

dxbjavid:
Thanks for the review and merge! Appreciated

PR #14960, opened 2026-06-06
Add a signature validation budget during path construction

woodruffw:
This extends our existing NC budget check to include a budget for signature validations. If a path construction exceeds the budget by performing more than the allowed number of signature validation steps, the entire construction fails.

For now, our budget is 128 signature validations. This is consistent with (higher than) Go and rustls-webpki, which both set a limit of 100. Like Go, we attempt to make the "best" use of our signature budget by ordering by likelihood, using AKI/SKI match as the strongest signal of fitness.

### DIFF
--- a/src/rust/cryptography-x509-verification/src/lib.rs
+++ b/src/rust/cryptography-x509-verification/src/lib.rs
@@ -18,10 +18,14 @@ use std::vec;
 use asn1::ObjectIdentifier;
 use cryptography_x509::common::Asn1Read;
 use cryptography_x509::extensions::{
-    DuplicateExtensionsError, Extensions, NameConstraints, SubjectAlternativeName,
+    AuthorityKeyIdentifier, DuplicateExtensionsError, Extensions, NameConstraints,
+    SubjectAlternativeName,
 };
 use cryptography_x509::name::GeneralName;
-use cryptography_x509::oid::{NAME_CONSTRAINTS_OID, SUBJECT_ALTERNATIVE_NAME_OID};
+use cryptography_x509::oid::{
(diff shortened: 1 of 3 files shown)


src/rust/cryptography-x509-verification/src/lib.rs
```diff
@@ -528,4 +595,102 @@ mod tests {
...
+        fn clone_public_key(_key: &Self::Key) -> Self::Key {}
+
+        fn clone_extra(_extra: &Self::CertificateExtra) -> Self::CertificateExtra {}
+    }
+
+    #[test]
+    fn test_clone() {
+        assert_eq!(NullOps::clone_public_key(&()), ());
+        assert_eq!(NullOps::clone_extra(&()), ());
+    }
```
woodruffw, line 631: We need `NullOps` for the coverage below, which means we need to "test" the empty trait impl bodies we add.

(We already do this for another test-only impl, `PulicKeyErrorOps`.)

src/rust/cryptography-x509-verification/src/lib.rs
```diff
@@ -528,4 +595,102 @@ mod tests {
...
+        let mut budget = Budget {
+            name_constraint_checks: usize::MAX,
+            signature_checks: usize::MAX,
+        };
+
+        let name_chain = NameChain::new::<NullOps>(None, &ca_exts, false)
+            .ok()
+            .unwrap();
+        let err = builder
+            .build_chain_inner(&working, u8::MAX, &ca_exts, name_chain, &mut budget)
```
woodruffw, line 688: This single line is the coverage-bearing one.

alex (approved):

PR #14962, opened 2026-06-07
Add asn1.value_set for enum-based ASN.1 value sets

### LINKED ISSUE #14891: asn1: Allow Enum Like Object identifiers
While working with the new asn1 implementation, i missed one specific feature that would make working with OIDs much simpler.

Currently there is just `x509.ObjectIdentifier`, which means i can build a sequence like the following:

```python
@asn1.sequence
class Test:
  a: x509.ObjectIdentifier
```

However this has the disadvantage that for every object creation i have to build an object identifier from a string like so.

```python
Test(a=x509.ObjectIdentifier('1.2.3.4'))
```

It is possible to create Constants for that or put them as Class Variables on a class however it would be much nicer from a type safety approach if we could have something like the following.

```python
class Algorithm(OIDEnum):
  SHA_256 = '1.2.3.4'
  SHA_384 = '1.2.3.5'
  SHA_512 = '1.2.3.6'

@asn1.sequence
class Test:
  a: Algorithm

obj = Test(a=Algorithm.SHA_256)

obj.a is Algorithm.SHA_256
```

The `Algorithm` class should behave like a Enum with singleton members and stuff. This would make it explicitly clear that the field `a` has to be an OID that represents exactly one of the given options. Also the type checker could catch errors where invalid OIDs are provided. Also when we have parsed an object we can be sure that in the field `a` must be exactly one of the declared OIDs. So parsing should fail if an invalid OID is provided.

I really are not sure if making a separate OIDEnum is required or we could just use Enum and declare the values as x509.ObjectIdentifier, or automatically convert the string values of needed. Or a decorator instead of subclassing from OIDEnum would be better. But some sort of type safe integrated mapping of names to OIDs would be very helpful.

alex:
Adds a new `@asn1.value_set(value_type)` class decorator that registers an `enum.Enum` subclass as an ASN.1 value set: a named set of values of a single governing type (X.680 §16.7).

```python
@asn1.value_set(x509.ObjectIdentifier)
class HashAlgorithm(enum.Enum):
    SHA_256 = x509.ObjectIdentifier("2.16.840.1.101.3.4.2.1")
    SHA_384 = x509.ObjectIdentifier("2.16.840.1.101.3.4.2.2")

@asn1.sequence
class Example:
    algorithm: HashAlgorithm
```

Members are encoded exactly as their underlying value; decoding maps the value back to the corresponding enum member and fails with `ValueError` if it doesn't match any member. All member values are validated against the declared `value_type` at class-definition time.

Works with any supported ASN.1 value type (OIDs, ints, strings, ...) and composes with `Implicit`/`Explicit`/`Default` annotations, `OPTIONAL` (`X | None`), `CHOICE` unions, and top-level `encode_der`/`decode_der`.

Implementation notes:
- New `Type::ValueSet(class, inner)` variant on the Rust side; the field's encoding annotation is propagated to the underlying value, while `DEFAULT` is handled at the member level.
- Decoding matches members by equality (not hash-based enum lookup), so wrapper types like `PrintableString` that define `__eq__` without `__hash__` work as values.

fixes #14891

https://claude.ai/code/session_01AbKLBGrHu13j6gKNpKGD2e

---
_Generated by [Claude Code](https://claude.ai/code/session_01AbKLBGrHu13j6gKNpKGD2e)_

### DIFF
--- a/src/cryptography/hazmat/asn1/asn1.py
+++ b/src/cryptography/hazmat/asn1/asn1.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import builtins
 import dataclasses
+import enum
 import sys
 import types
 import typing
@@ -170,7 +171,11 @@ def _normalize_field_type(
         root_type = field_type.__asn1_root__
         if not isinstance(
             root_type,
-            (declarative_asn1.Type.Sequence, declarative_asn1.Type.Set),
+            (
+                declarative_asn1.Type.Sequence,
+                declarative_asn1.Type.Set,
+                declarative_asn1.Type.ValueSet,
+            ),
         ):
             raise TypeError(f"unsupported root type: {root_type}")
         return declarative_asn1.AnnotatedType(
@@ -423,6 +428,45 @@ else:
         )(cls)
         _register_asn1_set(dataclass_cls)
         return dataclass_cls
+
+
+def value_set(cls: type[U]) -> type[U]:
+    """
+    A class decorator that registers an `enum.Enum` subclass as an
+    ASN.1 value set. All the member values must be instances of a
+    single ASN.1 type. Members are encoded as their value; decoding
+    fails if the decoded value does not match any member.
+    """
+    if not issubclass(cls, enum.Enum):
+        raise TypeError(
+            "value sets can only be defined from enum.Enum subclasses"
+        )
+    members = list(cls)
+    if not members:
+        raise TypeError(
+            f"value set '{cls.__name__}' must have at least one member"
+        )
+    value_type = type(members[0].value)
+    for member in members:
+        if type(member.value) is not value_type:
+            raise TypeError(
+                f"all members of value set '{cls.__name__}' must have "
+                "values of the same type"
+            )
+    try:
+        rust_type = declarative_asn1.non_root_python_to_rust(value_type)
+    except TypeError:
+        raise TypeError(
+            f"unsupported value type for value set '{cls.__name__}': "
+            f"{value_type!r}"
+        )
+    inner = declarative_asn1.AnnotatedType(
+        rust_type, declarative_asn1.Annotation()
+    )
+    root = declarative_asn1.Type.ValueSet(cls, inner)
+
+    setattr(cls, "__asn1_root__", root)
+    return cls
 
 
 # TODO: replace with `Default[U]` once the min Python version is >= 3.12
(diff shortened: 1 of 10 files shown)


src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -425,6 +430,52 @@ def set(cls: type[U]) -> type[U]:
...
+    """
+    A class decorator that registers an `enum.Enum` subclass as an
+    ASN.1 value set of the given underlying type. All the member
+    values must be instances of `value_type`. Members are encoded as
+    their value; decoding fails if the decoded value does not match
+    any member.
+    """
+    try:
+        rust_type = declarative_asn1.non_root_python_to_rust(value_type)
+    except TypeError:
```
alex: I don't think we need to re-raise here, the original exception is fine

src/cryptography/hazmat/asn1/asn1.py
```diff
@@ -425,6 +430,52 @@ def set(cls: type[U]) -> type[U]:
...
+            raise TypeError(
+                "value sets can only be defined from enum.Enum subclasses"
+            )
+        members = list(cls)
+        if not members:
+            raise TypeError(
+                f"value set '{cls.__name__}' must have at least one member"
+            )
+        for member in members:
+            if type(member.value) is not value_type:
```
alex: Should this be an `isinstance`? I guess theoretically subclasses are fine?

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -254,6 +254,33 @@ fn decode_null<'a>(
...
+fn decode_value_set<'a>(
+    py: pyo3::Python<'a>,
+    parser: &mut Parser<'a>,
+    cls: &pyo3::Py<pyo3::types::PyType>,
+    inner_type: &AnnotatedType,
+    annotation: &Annotation,
+) -> ParseResult<pyo3::Bound<'a, pyo3::PyAny>> {
+    let inner_ann_type = value_set_inner_type(py, inner_type, annotation)?;
+    let decoded = decode_annotated_type(py, parser, &inner_ann_type)?;
+    for member in cls.bind(py).try_iter()? {
```
alex, line 274: Seems like we should probably build a map and store it in the `Type::ValueSet` to avoid the linear scan

reaperhulk (approved):

PR #14964, opened 2026-06-07
Support arbitrary extension types in ExtensionPolicy

### LINKED ISSUE #14850: Unable to parse client certificates with critical 2.5.29.32 CertificatePolicies extension
I'm trying to verify a XAdES-signed XML document using signxml, currently being upgraded from version 3.x to 4.x, which involves switching the certificate parser from pyopenssl to pyca/cryptography, and I'm hitting a wall.

The client certificate embedded in the XAdES signature contains a critical OID 2.5.29.32 CertificatePolicies extension. Trying to parse it normally raises an `x509.verification.VerificationError` error: `validation failed: candidates exhausted: invalid extension: 2.5.29.32: certificate contains unaccounted-for critical extensions`

I thought the solution is to amend the PolicyBuilder with something like this:

```python
from cryptography.x509.extensions import SubjectAlternativeName, CertificatePolicies, BasicConstraints
from cryptography.x509.verification import PolicyBuilder, ExtensionPolicy, Criticality

ca_policy = (ExtensionPolicy.webpki_defaults_ca()
             .may_be_present(SubjectAlternativeName, Criticality.AGNOSTIC, None)  # MS KRK certs don't have SAN extensions
             .may_be_present(CertificatePolicies, Criticality.AGNOSTIC, None)  # <- This is the problem :(
)
ee_policy = (ExtensionPolicy.webpki_defaults_ee()
             .may_be_present(SubjectAlternativeName, Criticality.AGNOSTIC, None)
             .may_be_present(CertificatePolicies, Criticality.AGNOSTIC, None)
)

builder = (
    PolicyBuilder()
        .extension_policies(ca_policy=ca_policy, ee_policy=ee_policy)
        .store(self.store)
        .time(self.verification_time or datetime.now()
)
```

The builder only ever creates a client verifier, not the server verifier. Trying to configure the policy like this raises an exception when setting `may_be_present(CertificatePolicies, ...)`: `ValueError: Unsupported extension OID: 2.5.29.32`

I tried to work around the thing with just:

```python
ca_policy = ExtensionPolicy.permit_all().require_present(BasicConstraints, Criticality.CRITICAL, None)
ee_policy = ExtensionPolicy.permit_all()
```

But it still fails on `certificate contains unaccounted-for critical extensions` as it seems like the validator has a hardcoded list of extensions that [may be critical](https://github.com/pyca/cryptography/blob/main/src/rust/cryptography-x509-verification/src/policy/extension.rs#L231).

Is it possible to make it parse the certificate regardless of that?

reaperhulk:
The ExtensionPolicy builder methods (require_present, may_be_present, require_not_present) previously only accepted a fixed set of eight extension types and rejected everything else with "Unsupported extension OID". This made it impossible to account for otherwise unrecognized critical extensions (such as certificatePolicies), which would cause path validation to fail with "certificate contains unaccounted-for critical extensions".

ExtensionPolicy now keeps the eight default-handled extensions in their dedicated fields and stores any other user-added extension validators in a separate list, so any extension type can be configured.

A present extension whose value can't be parsed into a known Python object is now passed to the validator callback as an UnrecognizedExtension.

### DIFF
--- a/src/rust/cryptography-x509-verification/src/policy/extension.rs
+++ b/src/rust/cryptography-x509-verification/src/policy/extension.rs
@@ -2,6 +2,7 @@
 // 2.0, and the BSD License. See the LICENSE file in the root of this repository
 // for complete details.
 
+use std::collections::HashSet;
 use std::sync::Arc;
 
 use cryptography_x509::extensions::{Extension, Extensions};
@@ -15,7 +16,6 @@ use crate::ops::{CryptoOps, VerificationCertificate};
 use crate::policy::Policy;
 use crate::{ValidationError, ValidationErrorKind, ValidationResult};
 
-#[derive(Clone)]
 pub struct ExtensionPolicy<'cb, B: CryptoOps> {
     pub authority_information_access: ExtensionValidator<'cb, B>,
     pub authority_key_identifier: ExtensionValidator<'cb, B>,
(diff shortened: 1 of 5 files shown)


src/rust/cryptography-x509-verification/src/policy/extension.rs
```diff
@@ -168,7 +191,40 @@ impl<'cb, B: CryptoOps + 'cb> ExtensionPolicy<'cb, B> {
...
+            SUBJECT_KEY_IDENTIFIER_OID => policy.subject_key_identifier = validator,
+            KEY_USAGE_OID => policy.key_usage = validator,
+            SUBJECT_ALTERNATIVE_NAME_OID => policy.subject_alternative_name = validator,
+            BASIC_CONSTRAINTS_OID => policy.basic_constraints = validator,
+            NAME_CONSTRAINTS_OID => policy.name_constraints = validator,
+            EXTENDED_KEY_USAGE_OID => policy.extended_key_usage = validator,
+            _ => {
+                policy
+                    .additional_extensions
+                    .retain(|existing| existing.oid() != &oid);
```
alex: I don't think this retain is required -- we said registering these must be done only 1x.

src/rust/cryptography-x509-verification/src/policy/extension.rs
```diff
@@ -168,9 +199,48 @@ impl<'cb, B: CryptoOps + 'cb> ExtensionPolicy<'cb, B> {
...
+            AUTHORITY_INFORMATION_ACCESS_OID => policy.authority_information_access = validator,
+            AUTHORITY_KEY_IDENTIFIER_OID => policy.authority_key_identifier = validator,
+            SUBJECT_KEY_IDENTIFIER_OID => policy.subject_key_identifier = validator,
+            KEY_USAGE_OID => policy.key_usage = validator,
+            SUBJECT_ALTERNATIVE_NAME_OID => policy.subject_alternative_name = validator,
+            BASIC_CONSTRAINTS_OID => policy.basic_constraints = validator,
+            NAME_CONSTRAINTS_OID => policy.name_constraints = validator,
+            EXTENDED_KEY_USAGE_OID => policy.extended_key_usage = validator,
+            // `configured_oids` guarantees we haven't already stored a
+            // validator for this OID, so we can push without deduplicating.
```
alex: slop comment, remove

alex (approved):

ngnpope:
Thank you for this. A pleasant surprise to come back and find this was resolved 🙂

PR #14977, opened 2026-06-07
Track ChaCha20 32-bit counter and error on overflow

### LINKED ISSUE #9017: ChaCha20: change the current API (and recommendations)
This is a tie-together for Cryptography's side of the discussions in https://github.com/google/wycheproof/issues/90 and https://bugs.chromium.org/p/boringssl/issues/detail?id=614.

## Status quo

The current `ChaCha20` API in Cryptography is defined as follows:

```python
algorithms.ChaCha20(key, nonce)
```

...where `nonce` is a `bytes` of exactly `len(nonce) == 16`. This means that it contains both the "nonce" and "counter" fields used by ChaCha20, and leaves it to the underlying implementation to determine the appropriate nonce/counter split.

As part of that, the current documentation also encourages uses to fully randomize the `nonce` input, with `os.urandom(16)`:

https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#cryptography.hazmat.primitives.ciphers.algorithms.ChaCha20

Finally, the API implies that the underlying `ChaCha20` implementation implementation is RFC 7539, when the two current underlying implementations (OpenSSL and LibreSSL) are not (due to a 64/64 nonce/ctr split rather than 96/32).

Put together, these three points mean that the current `ChaCha20` API's internal implementations are more likely to diverge from RFC 7539 than they otherwise would be: if a user randomly initializes `nonce`, then the randomly initialized interior counter may be much closer to RFC 7539's block limit than a user might otherwise expect. 

I have two proposals for resolving this: one that forces all implementations to behave as if they're RFC 7539 compliant, and another that explicitly separates the API into 64/64 and 96/32 implementations.

## Proposal 1

My first proposal is a semi-breaking change. Behaviorally:

1. The API itself will change, with `nonce` becoming a 12-byte nonce rather than a 16-byte nonce;
2. Internally, the `ChaCha20` API will initialize the 32-bit counter (to 0) and combine it with the nonce.
3. _Optionally_, the API could grow a new `counter=` argument, which would allow the user to explicitly initialize the counter to a specific value in `[0, 2^32)`. 

In effect, this would make the non-RFC-compliant implementations of ChaCha20 behave as if they're compliant within the well-defined parameters of the RFC (i.e., up to 256 GB of enciphering), and would ensure that the majority of users never exceed those parameters (since most users will not need to manually initialize the counter).

Notably, this would be a breaking API change in terms of public API parameters, but a non-breaking change in terms of *documented* behavior (since it would be a bugfix to enforce compliance with RFC 7539, as documented). 

### Proposal 2

My second proposal is non-breaking, but requires a documentation correction and a new API for the RFC 7539-compliant version of ChaCha20. Behaviorally:

1. The existing ChaCha20 API's documentation will be updated to emphasize that it isn't intended to be RFC 7539 compliant. We'd also add test cases to ensure compliance with the 64/64 variant, e.g. at what would be the 32-bit counter overflow point. 
2. We'd add a new `ChaCha20RFC` API (with a better name than that), which would then support the RFC 7539 variant. This API would then use the pre-existing `ChaCha20` API with a fixed 32-bit nonce, forcing all implementations to behave as if they're compliant (at least within the bounds defined by the RFC -- exceeding those bounds will still exhibit divergences).

This would be non-breaking. However, if a _little_ bit of breakage is acceptable, I think combining this with a small change to `ChaCha20` to make `nonce` only 8 bytes (and allow an optional 64-bit counter) would be the best of both worlds: doing so would eliminate any confusion over initial high counter values, and would make the RFC-compliant version's interior use of the non-RFC version slightly more explicit.

See also #8956.

cc @facutuesca

reaperhulk:
ChaCha20 now extracts the 32-bit little-endian block counter from the first 4 bytes of the 128-bit nonce and tracks the number of bytes processed. Encrypting or decrypting more than (2**32 - counter) * 64 bytes would overflow that counter, at which point the underlying implementation silently diverges from RFC 7539. We now raise ValueError instead.

Setting the counter portion of the nonce to zero allows encrypting up to 256 GiB with a given nonce. The docs are updated accordingly and no longer suggest randomizing the full value.

The counter-overflow.txt vectors (and their generator/verifier docs) captured the old carry behavior, which can no longer be reproduced, so they are removed in favor of explicit overflow tests.


docs/hazmat/primitives/symmetric-encryption.rst
```diff
@@ -161,30 +158,30 @@ Algorithms
...
-        ``struct.pack("<Q", 2) + nonce``.
-
+        The block counter occupies the first 4 bytes of the 128-bit value and
+        is a 32-bit little-endian integer. Each ChaCha20 block encrypts 64
+        bytes, so an initial counter value of ``n`` allows up to
+        ``(2 ** 32 - n) * 64`` bytes to be encrypted before the counter would
+        overflow. We recommend setting the counter portion to zero, which
+        allows encrypting up to 256 GiB with a given nonce. Attempting to
+        encrypt or decrypt more data than the counter allows raises a
+        :class:`ValueError`.
```
alex: The "We recommend" language probably needs to somehow be more scenario driven.

docs/hazmat/primitives/symmetric-encryption.rst
```diff
@@ -161,30 +158,30 @@ Algorithms
...
-        >>> import struct, os
+        >>> import os
         >>> from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
         >>> key = os.urandom(32)
-        >>> nonce = os.urandom(8)
-        >>> counter = 0
-        >>> full_nonce = struct.pack("<Q", counter) + nonce
-        >>> algorithm = algorithms.ChaCha20(key, full_nonce)
+        >>> counter = b"\x00\x00\x00\x00"
+        >>> nonce = counter + os.urandom(12)
```
alex: I think the `struct.pack` form was better.

src/rust/src/backend/ciphers.rs
```diff
@@ -228,12 +228,67 @@ impl CipherContext {
...
+// ChaCha20 block counter would overflow, or `None` if `algorithm` is not
+// ChaCha20.
+fn chacha20_initial_byte_limit(
+    py: pyo3::Python<'_>,
+    algorithm: &pyo3::Bound<'_, pyo3::PyAny>,
+) -> CryptographyResult<Option<u64>> {
+    if algorithm.is_instance(&types::CHACHA20.get(py)?)? {
+        let nonce = algorithm
+            .getattr(pyo3::intern!(py, "nonce"))?
+            .extract::<CffiBuf<'_>>()?;
```
alex, line 251: I think you can directly extract to `[u8; 16]`

src/rust/src/backend/ciphers.rs
```diff
@@ -270,11 +325,25 @@ impl PyCipherContext {
...
+        // nonce length) succeeds.
+        let counter_bytes: Option<[u8; 4]> = if self.bytes_remaining.is_some() {
+            nonce.as_bytes().get(..4).map(|b| b.try_into().unwrap())
+        } else {
+            None
+        };
+        get_mut_ctx(self.ctx.as_mut())?.reset_nonce(py, nonce)?;
+        if let Some(counter_bytes) = counter_bytes {
+            self.bytes_remaining = Some(chacha20_byte_limit(u32::from_le_bytes(counter_bytes)));
+        }
```
alex: Seems like this should share code with the version?

src/rust/src/backend/ciphers.rs
```diff
@@ -228,12 +228,67 @@ impl CipherContext {
...
+// ChaCha20.
+fn chacha20_initial_byte_limit(
+    py: pyo3::Python<'_>,
+    algorithm: &pyo3::Bound<'_, pyo3::PyAny>,
+) -> CryptographyResult<Option<u64>> {
+    if algorithm.is_instance(&types::CHACHA20.get(py)?)? {
+        // The nonce is guaranteed to be 16 bytes by the ChaCha20 constructor.
+        let nonce = algorithm
+            .getattr(pyo3::intern!(py, "nonce"))?
+            .extract::<[u8; 16]>()?;
```
alex: This is wrong in the presence of nonce being a non-bytes buffer I think.

vectors/cryptography_vectors/ciphers/ChaCha20/rfc7539.txt
```diff
@@ -1,7 +1,7 @@
 # The vectors are from RFC 7539 Appendix A.2. They are reformatted into NIST
-# form for our vector loaders, and adapted to use a 64/64 bit counter/nonce
-# split (matching our implementation), rather than the 32/96 split defined
-# in the RFC.
+# form for our vector loaders. The INITIAL_BLOCK_COUNTER and NONCE columns are
+# loaded as the 32-bit little-endian block counter and 96-bit nonce defined in
+# the RFC, concatenated to form the 128-bit value passed to ChaCha20.
```
alex, line 6: I don't understand the changes here -- did we modify the vectors in some way?

alex (dismissed):

src/rust/src/backend/ciphers.rs
```diff
@@ -228,12 +228,63 @@ impl CipherContext {
...
+// ChaCha20 generates 64 bytes of keystream per 32-bit block counter value.
+const CHACHA20_BLOCK_SIZE: u64 = 64;
+
+// The 16-byte ChaCha20 nonce begins with a 32-bit little-endian block counter.
+// Encrypting more than `(2**32 - counter) * 64` bytes would overflow that
+// counter, at which point the OpenSSL implementation silently diverges from
+// RFC 7539 (the counter carries into the rest of the nonce), so we instead
+// refuse to encrypt past that point.
+fn chacha20_byte_limit(nonce: &CffiBuf<'_>) -> Option<u64> {
+    let counter = u32::from_le_bytes(nonce.as_bytes().get(..4)?.try_into().unwrap());
```
alex: Sorry, on further review it's wrong that this returns an Option and use `get(..4)?`. It should use `[..4]` and return just a `u64` and the caller should wrap it in `Some`.

This is called exclusively when our nonce is from chacha, and therefore we always have a limit.

src/rust/src/backend/ciphers.rs
```diff
@@ -228,12 +228,63 @@ impl CipherContext {
...
+            self.bytes_remaining = Some(
+                remaining
+                    .checked_sub(u64::try_from(n).unwrap())
+                    .ok_or_else(|| {
+                        pyo3::exceptions::PyValueError::new_err(
+                            "Exceeded the maximum number of bytes that can be \
+                             encrypted with ChaCha20 for this nonce. The 32-bit \
+                             counter portion of the nonce would overflow; set the \
+                             counter portion of the nonce to zero to allow \
+                             encrypting up to 256 GiB.",
```
alex: This last sentence is probably a wrong this to include.

src/rust/src/backend/ciphers.rs
```diff
@@ -270,11 +321,25 @@ impl PyCipherContext {
...
+        let data = data.as_bytes();
+        self.decrement_bytes_remaining(data.len())?;
+        get_mut_ctx(self.ctx.as_mut())?.update(py, data)
     }
 
     fn reset_nonce(&mut self, py: pyo3::Python<'_>, nonce: CffiBuf<'_>) -> CryptographyResult<()> {
-        get_mut_ctx(self.ctx.as_mut())?.reset_nonce(py, nonce)
+        // Compute the new limit before the buffer is moved into the inner
+        // reset, but only apply it once that reset (which validates the nonce
+        // length) succeeds.
```
alex: There has got to be a better way to structure this, it makes the logic correct -- we can just `derive(Clone)` on `CffiBuf` and then do all ops afterwards.

alex: eh - clone isn't possible, but I think the broader point holds. just change in the inner `reset_nonce` to take a `&[u8]` and this becomes trivial

alex (approved):

PR #14978, opened 2026-06-07
run downstream tests under 3.14

alex:


alex:
This needs https://github.com/aws/aws-encryption-sdk-python/pull/808

reaperhulk:
I propose we give these folks one more week and then we consider dropping this as a downstream.

alex:
We should probably ping someone, I suppose.

reaperhulk (approved):

PR #14979, opened 2026-06-07
Add external mu (message representative) support for ML-DSA

### LINKED ISSUE #14827: Provide HashML-DSA
Python cryptography already supports ML-DSA.

I need support for HashML-DSA as well. In our secure boot implementation, we hash the firmware binary and then provide the hash for verification to different digitial signature algorithms. Pure ML-DSA would require to pass the complete data to the hash function, which is inefficient regarding startup times. Just signing the hash with Pure ML-DSA does not make use of the clean context separation provided by HashML-DSA.

Instead of the complete message, HashML-DSA can create a signature based on the given hash of some input data.

According to FIPS 204 (https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf), the HashML-DSA sign function consumes a digest as message and the hash algorithm used to generate the digest as additional parameter.

Optional contexts are supported for both ML-DSA and HashML-DSA.

reaperhulk:
This PR adds support for the "external mu" variant of ML-DSA signing and verification as defined in FIPS 204. This allows applications to precompute the message representative (mu) and sign/verify it directly, which is useful for protocols that need to work with the intermediate mu value.

## Key Changes

- **New constant**: `MLDSA_MU_BYTES` (64 bytes) defines the length of an ML-DSA external mu value
- **New signing method**: `sign_mu(mu)` on all three private key classes (MlDsa44, MlDsa65, MlDsa87) to sign a precomputed 64-byte mu
- **New verification method**: `verify_mu(signature, mu)` on all three public key classes to verify a signature over a precomputed mu
- **OpenSSL 3.5.0+ support**: Added `CRYPTOGRAPHY_OPENSSL_350_OR_GREATER` to the conditional compilation flags for the new functions
- **Backend implementations**: 
  - BoringSSL: Uses low-level ML-DSA API with private key reconstruction from seed
  - AWS-LC: Uses EVP_PKEY_sign/verify with "ExternalMu" format
  - OpenSSL 3.5.0+: Uses EVP with the "mu" signature parameter
- **Input validation**: Both methods validate that mu is exactly 64 bytes, raising `ValueError` if not
- **Comprehensive tests**: Added unit tests and Wycheproof test vector validation for the external mu interface

## Implementation Details

The external mu interface is implemented with conditional compilation to support three different backends:
- For BoringSSL, the implementation reconstructs the private key from its 32-byte seed and calls the low-level `MLDSA*_sign_message_representative` functions
- For AWS-LC, the EVP interface natively supports external mu through the "ExternalMu" format
- For OpenSSL 3.5.0+, a helper function `set_mu()` configures the signature context with the "mu" parameter before signing/verifying

The external mu signatures are fully compatible with ordinary ML-DSA signatures—a signature produced via `sign_mu()` can be verified with the standard `verify()` method if the message and context are provided, and vice versa.

https://claude.ai/code/session_01MmjphxZ6ookRpjUhQouKjf

### DIFF
--- a/docs/hazmat/primitives/asymmetric/mldsa.rst
+++ b/docs/hazmat/primitives/asymmetric/mldsa.rst
@@ -100,8 +100,12 @@ Key interfaces
         .. versionadded:: 49.0.0
 
         Sign a precomputed ``mu`` (message representative) using ML-DSA-44,
-        the "external mu" variant from FIPS 204. ``mu`` already incorporates
-        the public key and any context string, so no context is accepted here.
+        the "external mu" variant from FIPS 204. ``mu`` is computed as
+        ``SHAKE256(SHAKE256(public_key, 64) || M', 64)``, where ``M'`` encodes
+        the context and message; because it already binds the public key and
+        context, no context is accepted here. ``cryptography`` does not
+        currently provide an API to compute ``mu``, so you must compute it
+        yourself.
 
         :param mu: The 64-byte message representative.
         :type mu: :term:`bytes-like`
@@ -247,8 +251,10 @@ Key interfaces
         .. versionadded:: 49.0.0
 
         Verify a signature over a precomputed ``mu`` (message representative),
-        the "external mu" variant from FIPS 204. ``mu`` already incorporates
-        the public key and any context string.
+        the "external mu" variant from FIPS 204. ``mu`` is computed as
+        ``SHAKE256(SHAKE256(public_key, 64) || M', 64)``, where ``M'`` encodes
+        the context and message. ``cryptography`` does not currently provide an
+        API to compute ``mu``, so you must compute it yourself.
 
         :param signature: The signature to verify.
         :type signature: :term:`bytes-like`
@@ -320,8 +326,12 @@ Key interfaces
         .. versionadded:: 49.0.0
 
         Sign a precomputed ``mu`` (message representative) using ML-DSA-65,
-        the "external mu" variant from FIPS 204. ``mu`` already incorporates
-        the public key and any context string, so no context is accepted here.
+        the "external mu" variant from FIPS 204. ``mu`` is computed as
+        ``SHAKE256(SHAKE256(public_key, 64) || M', 64)``, where ``M'`` encodes
+        the context and message; because it already binds the public key and
+        context, no context is accepted here. ``cryptography`` does not
+        currently provide an API to compute ``mu``, so you must compute it
+        yourself.
 
         :param mu: The 64-byte message representative.
         :type mu: :term:`bytes-like`
@@ -467,8 +477,10 @@ Key interfaces
         .. versionadded:: 49.0.0
 
         Verify a signature over a precomputed ``mu`` (message representative),
-        the "external mu" variant from FIPS 204. ``mu`` already incorporates
-        the public key and any context string.
+        the "external mu" variant from FIPS 204. ``mu`` is computed as
+        ``SHAKE256(SHAKE256(public_key, 64) || M', 64)``, where ``M'`` encodes
+        the context and message. ``cryptography`` does not currently provide an
+        API to compute ``mu``, so you must compute it yourself.
 
         :param signature: The signature to verify.
         :type signature: :term:`bytes-like`
@@ -540,8 +552,12 @@ Key interfaces
         .. versionadded:: 49.0.0
 
         Sign a precomputed ``mu`` (message representative) using ML-DSA-87,
-        the "external mu" variant from FIPS 204. ``mu`` already incorporates
-        the public key and any context string, so no context is accepted here.
+        the "external mu" variant from FIPS 204. ``mu`` is computed as
+        ``SHAKE256(SHAKE256(public_key, 64) || M', 64)``, where ``M'`` encodes
+        the context and message; because it already binds the public key and
+        context, no context is accepted here. ``cryptography`` does not
+        currently provide an API to compute ``mu``, so you must compute it
+        yourself.
 
         :param mu: The 64-byte message representative.
         :type mu: :term:`bytes-like`
@@ -687,8 +703,10 @@ Key interfaces
         .. versionadded:: 49.0.0
 
         Verify a signature over a precomputed ``mu`` (message representative),
-        the "external mu" variant from FIPS 204. ``mu`` already incorporates
-        the public key and any context string.
+        the "external mu" variant from FIPS 204. ``mu`` is computed as
+        ``SHAKE256(SHAKE256(public_key, 64) || M', 64)``, where ``M'`` encodes
+        the context and message. ``cryptography`` does not currently provide an
+        API to compute ``mu``, so you must compute it yourself.
 
         :param signature: The signature to verify.
         :type signature: :term:`bytes-like`
--- a/docs/hazmat/primitives/asymmetric/mldsa.rst
+++ b/docs/hazmat/primitives/asymmetric/mldsa.rst
@@ -95,6 +95,21 @@ Key interfaces
 
         :raises ValueError: If the context is longer than 255 bytes.
 
+    .. method:: sign_mu(mu)
+
+        .. versionadded:: 49.0.0
+
+        Sign a precomputed ``mu`` (message representative) using ML-DSA-44,
+        the "external mu" variant from FIPS 204. ``mu`` already incorporates
+        the public key and any context string, so no context is accepted here.
+
+        :param mu: The 64-byte message representative.
+        :type mu: :term:`bytes-like`
+
+        :returns bytes: The signature (2420 bytes).
+
+        :raises ValueError: If ``mu`` is not 64 bytes.
+
     .. method:: private_bytes(encoding, format, encryption_algorithm)
 
         Allows serialization of the key to bytes. Encoding (
@@ -227,6 +242,25 @@ Key interfaces
             signature cannot be verified.
         :raises ValueError: If the context is longer than 255 bytes.
 
+    .. method:: verify_mu(signature, mu)
+
+        .. versionadded:: 49.0.0
+
+        Verify a signature over a precomputed ``mu`` (message representative),
+        the "external mu" variant from FIPS 204. ``mu`` already incorporates
+        the public key and any context string.
+
+        :param signature: The signature to verify.
+        :type signature: :term:`bytes-like`
+
+        :param mu: The 64-byte message representative.
+        :type mu: :term:`bytes-like`
+
+        :returns: None
+        :raises cryptography.exceptions.InvalidSignature: Raised when the
+            signature cannot be verified.
+        :raises ValueError: If ``mu`` is not 64 bytes.
+
 .. class:: MLDSA65PrivateKey
 
     .. versionadded:: 47.0.0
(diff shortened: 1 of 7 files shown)


docs/hazmat/primitives/asymmetric/mldsa.rst
```diff
@@ -95,6 +95,21 @@ Key interfaces
 
         :raises ValueError: If the context is longer than 255 bytes.
 
+    .. method:: sign_mu(mu)
+
+        .. versionadded:: 49.0.0
+
+        Sign a precomputed ``mu`` (message representative) using ML-DSA-44,
+        the "external mu" variant from FIPS 204. ``mu`` already incorporates
+        the public key and any context string, so no context is accepted here.
```
alex: I feel like this second sentence is low value, and instead there should be _something_ about how you compuete mu (even if just to say we don't have an API for it yet)

docs/hazmat/primitives/asymmetric/mldsa.rst
```diff
@@ -227,6 +242,25 @@ Key interfaces
             signature cannot be verified.
         :raises ValueError: If the context is longer than 255 bytes.
 
+    .. method:: verify_mu(signature, mu)
+
+        .. versionadded:: 49.0.0
+
+        Verify a signature over a precomputed ``mu`` (message representative),
+        the "external mu" variant from FIPS 204. ``mu`` already incorporates
+        the public key and any context string.
```
alex: Same

src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -264,6 +280,196 @@ pub fn verify(
...
+/// [`MLDSA_MU_BYTES`] long, and already incorporates any context string, so no
+/// context is accepted here.
+pub fn sign_mu(
+    pkey: &openssl::pkey::PKeyRef<openssl::pkey::Private>,
+    variant: MlDsaVariant,
+    mu: &[u8],
+) -> OpenSSLResult<Vec<u8>> {
+    cfg_if::cfg_if! {
+        if #[cfg(CRYPTOGRAPHY_IS_BORINGSSL)] {
+            // BoringSSL has no EVP-level external mu support, so we drop down to
```
alex, line 321: if humanly possible, I'd like to wait on that, since it makes this 100x shorter

src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -264,6 +280,196 @@ pub fn verify(
...
+            ctx.sign_init()?;
+            let mut sig = vec![];
+            ctx.sign_to_vec(mu, &mut sig)?;
+            Ok(sig)
+        } else if #[cfg(CRYPTOGRAPHY_OPENSSL_350_OR_GREATER)] {
+            // OpenSSL signs an external mu by setting the "mu" parameter and
+            // passing the 64-byte mu in place of the message.
+            let _ = variant;
+            let mut md_ctx = openssl::md_ctx::MdCtx::new()?;
+            let pkey_ctx = md_ctx.digest_sign_init(None, pkey)?;
```
alex, line 394: Does this work if you do a pkeyctx with sign_init on openssl? (with or without set_mu)? that'd nicely let us consolidate with the aws-lc path

src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -264,6 +280,196 @@ pub fn verify(
...
+    }
+    Ok(())
+}
+
+/// Sign a precomputed external mu (message representative). `mu` must be
+/// [`MLDSA_MU_BYTES`] long, and already incorporates any context string, so no
+/// context is accepted here.
+pub fn sign_mu(
+    pkey: &openssl::pkey::PKeyRef<openssl::pkey::Private>,
+    variant: MlDsaVariant,
```
alex: this should not take variant as an argument, it should be inferred from pkey, as `sign` does. otherwise this is unsound

src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -264,6 +280,196 @@ pub fn verify(
...
+            Ok(sig)
+        }
+    }
+}
+
+/// Verify a signature over a precomputed external mu (message representative).
+/// `mu` must be [`MLDSA_MU_BYTES`] long.
+pub fn verify_mu(
+    pkey: &openssl::pkey::PKeyRef<openssl::pkey::Public>,
+    variant: MlDsaVariant,
```
alex: same

src/rust/src/backend/mldsa.rs
```diff
@@ -561,6 +654,30 @@ impl MlDsa87PrivateKey {
 
 #[pyo3::pymethods]
 impl MlDsa87PublicKey {
+    #[pyo3(signature = (signature, mu))]
```
alex: why does this _one_ have a `signature = ` block?

tests/wycheproof/test_mldsa.py
```diff
@@ -161,6 +162,71 @@ def test_mldsa65_sign_seed(backend, wycheproof):
...
+
+
+def _external_mu_test(public_key_class, wycheproof):
+    # The sign vectors carry a precomputed mu ("External Mu") for every case
+    # that has a valid signature, including the "Internal" cases that NIST
+    # provides as bare mu values with no message or context. Those are
+    # skipped by the signing tests above (we don't expose Sign_internal) but
+    # exercise the precomputed-mu verification interface here.
+    if "mu" not in wycheproof.testcase or not wycheproof.valid:
+        return
```
alex: This comment makes no sense to me, if mu exists in all of them, what's the first clause doing?

docs/hazmat/primitives/asymmetric/mldsa.rst
```diff
@@ -100,8 +100,12 @@ Key interfaces
         .. versionadded:: 49.0.0
 
         Sign a precomputed ``mu`` (message representative) using ML-DSA-44,
-        the "external mu" variant from FIPS 204. ``mu`` already incorporates
-        the public key and any context string, so no context is accepted here.
+        the "external mu" variant from FIPS 204. ``mu`` is computed as
+        ``SHAKE256(SHAKE256(public_key, 64) || M', 64)``, where ``M'`` encodes
```
alex: This is now way too verbose.

src/rust/cryptography-openssl/src/mldsa.rs
reaperhulk: I tried this against OpenSSL 4.0 and it doesn't work: `EVP_PKEY_sign_init` (and `verify_init`) fail for ML-DSA with `provider signature not supported` (`crypto/evp/signature.c`). The ML-DSA provider only registers the digest-sign / sign-message flow (`OSSL_FUNC_SIGNATURE_DIGEST_SIGN*`, plus a one-shot `SIGN` with no `SIGN_INIT`), so we can't drive it through a `PkeyCtx` like the AWS-LC path — it has to go through `digest_sign_init` + the `mu` param. I left a comment to that effect.

---

alex:
Claude you've got merge conflicts, can you rebase?

src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -264,6 +280,187 @@ pub fn verify(
...
+/// Enable "external mu" mode on an OpenSSL signing/verification context by
+/// setting the integer `mu` signature parameter (OSSL_SIGNATURE_PARAM_MU) to 1.
+#[cfg(CRYPTOGRAPHY_OPENSSL_350_OR_GREATER)]
+fn set_mu<T>(pkey_ctx: &mut openssl::pkey_ctx::PkeyCtxRef<T>) -> OpenSSLResult<()> {
+    // A fixed OSSL_PARAM array holding the integer "mu" parameter set to 1
+    // enables external mu mode. The provider reads it back with
+    // OSSL_PARAM_get_int, which accepts an unsigned integer param, so we can
+    // build the array on the stack rather than via an allocating
+    // OSSL_PARAM_BLD.
+    let mut mu: std::os::raw::c_uint = 1;
```
alex: use `std::ffi::c_uint` please

src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -264,6 +280,187 @@ pub fn verify(
...
+) -> OpenSSLResult<Vec<u8>> {
+    cfg_if::cfg_if! {
+        if #[cfg(CRYPTOGRAPHY_IS_BORINGSSL)] {
+            // BoringSSL has no EVP-level external mu support, so we drop down to
+            // the low-level ML-DSA API, reconstructing the private key from its
+            // 32-byte seed.
+            let seed = mldsa_seed_raw(pkey)?;
+            // SAFETY: `seed` is a valid 32-byte seed and `mu` is a valid
+            // MLDSA_MU_BYTES buffer; both outlive the calls below.
+            unsafe {
```
alex, line 327: This needs an assert that `mu.len() == MLDSA_MU_BYTES`, right now it's unenforced which makes this unsound. (In practice callers verify it, but we should assert it for soundness)

src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -264,6 +280,187 @@ pub fn verify(
...
+    pkey: &openssl::pkey::PKeyRef<openssl::pkey::Public>,
+    signature: &[u8],
+    mu: &[u8],
+) -> OpenSSLResult<bool> {
+    cfg_if::cfg_if! {
+        if #[cfg(CRYPTOGRAPHY_IS_BORINGSSL)] {
+            let raw = pkey.raw_public_key()?;
+            // SAFETY: We parse the low-level public key from its encoded form
+            // and verify the signature over the MLDSA_MU_BYTES `mu`.
+            unsafe {
```
alex, line 419: same mu length issue here

src/rust/cryptography-openssl/src/mldsa.rs
```diff
@@ -264,6 +280,187 @@ pub fn verify(
...
+                match MlDsaVariant::from_pkey(pkey) {
+                    MlDsaVariant::MlDsa44 => {
+                        let mut key = std::mem::MaybeUninit::<ffi::MLDSA44_public_key>::uninit();
+                        let mut cbs = ffi::CBS { data: raw.as_ptr(), len: raw.len() };
```
alex: Should this be using `CBS_init()`?

tests/wycheproof/test_mldsa.py
```diff
@@ -161,6 +162,73 @@ def test_mldsa65_sign_seed(backend, wycheproof):
...
+        pub.verify(sig, msg, ctx)
+
+
+@pytest.mark.supported(
+    only_if=lambda backend: backend.mldsa_supported(),
+    skip_message="Requires a backend with ML-DSA support",
+)
+@wycheproof_tests("mldsa_44_sign_seed_test.json")
+def test_mldsa44_external_mu(backend, wycheproof):
+    _external_mu_test(MLDSA44PublicKey, wycheproof)
```
alex: Why is this written in this weird style, we usually pass all paths to `wycheproof_tests` and have it compute the key types internally

Cargo.toml
```diff
@@ -36,3 +36,9 @@ self_cell = "1"
 
 [profile.release]
 overflow-checks = true
+
+[patch.crates-io]
+# Temporarily track rust-openssl master for the BoringSSL ML-DSA bindgen
+# additions (mldsa.h) until a release including them is published.
+openssl = { git = "https://github.com/sfackler/rust-openssl", branch = "master" }
+openssl-sys = { git = "https://github.com/sfackler/rust-openssl", branch = "master" }
```
alex: Do we need a release here? This isn't landable

alex:
Also there's a merge conflict in need of resolution

Cargo.toml
alex: we've now done a release and upgraded. this can be dropped

alex (approved):

PR #14984, opened 2026-06-08
Key cffi build-script cache on PYO3_BASE_PYTHON when available

alex:
Under PEP 517 build isolation, frontends (pip/uv/build) create venvs at random paths and point `PYO3_PYTHON` at them. Because `cryptography-cffi`'s `build.rs` declares `cargo:rerun-if-env-changed=PYO3_PYTHON`, that path churn makes the build script re-run on every invocation, defeating build caching. This is the same problem tracked upstream in [PyO3/pyo3#6113](https://github.com/PyO3/pyo3/issues/6113), which proposes a stable `PYO3_BASE_PYTHON` (≈ `sys._base_executable`).

This change mirrors the proposed "prefer BASE if present" behavior for this crate's build script: when `PYO3_BASE_PYTHON` is set, key the rerun on it; otherwise fall back to `PYO3_PYTHON`.

Why keying on the base interpreter is sound here: a venv cannot change the implementation (CPython/PyPy), version, ABI (free-threaded vs. not), or C headers of its base — and those are the only interpreter-derived inputs this script has. We still invoke `PYO3_PYTHON` to locate the binary, since `build_openssl.py` needs cffi, which lives in the build venv rather than the base interpreter. The fallback preserves correct rebuild behavior for manual interpreter switches (e.g. setting `PYO3_PYTHON` in a dev checkout) when `PYO3_BASE_PYTHON` is absent.

Note: this is forward-looking — it's harmless until a frontend actually sets `PYO3_BASE_PYTHON`, at which point caching improves.

### DIFF
--- a/src/rust/cryptography-cffi/build.rs
+++ b/src/rust/cryptography-cffi/build.rs
@@ -25,7 +25,21 @@ fn main() {
     let out_dir = env::var("OUT_DIR").unwrap();
     // FIXME: maybe pyo3-build-config should provide a way to do this?
     let python = env::var("PYO3_PYTHON").unwrap_or_else(|_| "python3".to_string());
-    println!("cargo:rerun-if-env-changed=PYO3_PYTHON");
+    // PEP 517 build frontends create build isolation venvs at random paths and
+    // point PYO3_PYTHON at them, so keying the build-script cache on PYO3_PYTHON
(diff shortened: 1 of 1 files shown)


alex:
(Shouldn't land until we depend on a pyo3 release with this)

PR #14992, opened 2026-06-09
Add more robust key processing: DSA

sjudson:
This PR adds more robust key processing for DSA. In particular:

- a new check, `check_dsa_public_numbers`, is added to verify that for DSA the public `y` value both `1 < y < p` and `y ** q mod p = 1`, as keys failing validation may be weak to forgery; and
- both that new and existing checks on key structure are added to the PEM and DER load paths, alongside where they were already used for direct constructions.

### DIFF
--- a/src/rust/src/backend/dsa.rs
+++ b/src/rust/src/backend/dsa.rs
@@ -2,6 +2,8 @@
 // 2.0, and the BSD License. See the LICENSE file in the root of this repository
 // for complete details.
 
+use std::cmp::Ordering;
+
 use pyo3::types::PyAnyMethods;
 
 use crate::backend::utils;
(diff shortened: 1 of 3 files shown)


src/rust/src/backend/dsa.rs
```diff
@@ -37,19 +37,23 @@ struct DsaParameters {
 }
 
 pub(crate) fn private_key_from_pkey(
+    _py: pyo3::Python<'_>,
```
alex: If we don't need this arg, we can just drop it.

src/rust/src/backend/rsa.rs
```diff
@@ -55,11 +55,15 @@ pub(crate) fn private_key_from_pkey(
...
     pkey: &openssl::pkey::PKeyRef<openssl::pkey::Public>,
-) -> RsaPublicKey {
-    RsaPublicKey {
+) -> CryptographyResult<RsaPublicKey> {
+    let key = RsaPublicKey {
         pkey: pkey.to_owned(),
-    }
+    };
+    let numbers = key.public_numbers(py)?;
+    check_public_key_components(numbers.e.bind(py), numbers.n.bind(py))?;
```
alex: Hmm, ideally we wouldn't have to round trip via the public numbers to create a public key... is there some way to structure this to avoid that?

alex (commented):
All the new vectors need to be documented in test-vectors.rst -- please split those out into their own PR.

For ease of review, I think it'd be best to split the actual changes into one PR per algorithm.

src/rust/src/backend/dsa.rs
sjudson: Fixed.

src/rust/src/backend/rsa.rs
sjudson: @alex thoughts on this alternative? Modifies the existing check to work on bignums as well.

sjudson: I'll leave this as the RSA PR btw, and split out the DH and DSA and vectors as requested

reaperhulk:
If you load a private key (via DER/PEM) with bad public key values this will raise a ValueError when calling `key.public_key()`, which is not desirable. We should run checks for all the values on private loading.

sjudson:
@reaperhulk updated to resolve this, and in a way consistent with your comment on https://github.com/pyca/cryptography/pull/15015 as well: the validation has been moved down to the binding layer, where it is then applied across the various code paths so that there is a single consistent implementation.

However, this has broken one of the certbot-josepy downstream tests, which loads a 512-bit private key (expecting no error at the cryptography layer) and then rejects it. With the stricter validation, that key is now rejected on load. 

Do you have a sense of how you'd prefer to proceed? Obviously one approach would be to accept being breaking here. I could also modify the validation to accept 512-bit key lengths, despite their insecure/legacy nature. Or even stronger I could drop the bitlength checks from the PEM/DER load paths (but leave them where they already exist) for maximal compatibility.

reaperhulk:
After sitting on this for a long time I think we're actually going to deprecate DSA. Thanks for the work here though!

PR #15004, opened 2026-06-10
Merge LazyEvpCipherAead into EvpCipherAead

### LINKED ISSUE #14795: AESGCM fails on RHEL9 in FIPS mode

alex:
`EvpCipherAead` now stores the cipher, key (as a `Py<PyAny>`, re-extracted on the fallback path like the lazy struct did), and an `is_ccm` flag. `encrypt_into`/`decrypt_into` attempt to copy the pre-keyed base context; if that fails (providers prior to OpenSSL 3.2 don't support copying AEAD contexts), or for CCM (whose IV and tag lengths must be set before the key, so pre-keyed contexts can't be used at all -- its base contexts are left uninitialized), they initialize a fresh context from scratch.

Since most callers use static ciphers but AESSIV/AESGCMSIV use `Cipher::fetch`, there's a small `AeadCipher` enum to hold either. The `*_with_context` helpers are inlined into `encrypt_into`/`decrypt_into`, the cfg-conditional `ctx` fields on `ChaCha20Poly1305`/`AESGCM`/`AESCCM` collapse to a single type, and `AESSIV`/`AESOCB3`/`AESGCMSIV` constructors take the key as `Py<PyAny>` so they can hand it to `EvpCipherAead::new`.

### DIFF
--- a/src/rust/src/backend/aead.rs
+++ b/src/rust/src/backend/aead.rs
@@ -26,30 +26,71 @@ pub(crate) enum Aad<'a> {
     List(pyo3::Bound<'a, pyo3::types::PyList>),
 }
 
+pub(crate) enum AeadCipher {
+    Static(&'static openssl::cipher::CipherRef),
+    #[cfg(not(any(
+        CRYPTOGRAPHY_IS_LIBRESSL,
+        CRYPTOGRAPHY_IS_BORINGSSL,
+        CRYPTOGRAPHY_IS_AWSLC
+    )))]
+    Fetched(openssl::cipher::Cipher),
+}
+
+impl std::ops::Deref for AeadCipher {
+    type Target = openssl::cipher::CipherRef;
+
+    fn deref(&self) -> &openssl::cipher::CipherRef {
+        match self {
+            AeadCipher::Static(c) => c,
+            #[cfg(not(any(
+                CRYPTOGRAPHY_IS_LIBRESSL,
+                CRYPTOGRAPHY_IS_BORINGSSL,
+                CRYPTOGRAPHY_IS_AWSLC
+            )))]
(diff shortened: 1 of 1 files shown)


alex:
Benchmark numbers claim big speedups. I'm very confused why...

alex:
Perf gain is real! Turns out OpenSSL backported the dupctx fix to 3.0.13

reaperhulk (approved):

PR #15005, opened 2026-06-10
use markdown output format for coverage

alex:


alex:
meh, doesn't actually look better

PR #15008, opened 2026-06-11
encode arbitrary-precision integers in declarative asn1

dxbjavid:
the declarative asn1 decoder reads INTEGER fields as arbitrary-precision values, but the encoder narrows them to an i64, so an integer outside the signed 64-bit range that round-trips through decode_der cannot be re-encoded and instead raises OverflowError. this trips up otherwise valid structures whose integers exceed 64 bits, like RSA moduli or large serial numbers. i've changed the INTEGER arm to emit minimal two's-complement bytes through a BigInt, matching what the decoder already accepts.

### DIFF
--- a/src/rust/src/asn1.rs
+++ b/src/rust/src/asn1.rs
@@ -67,6 +67,37 @@ fn decode_dss_signature<'p>(
         .into_any())
 }
 
+// Encodes a Python `int` of arbitrary size as the minimal big-endian
+// two's complement bytes expected by a DER INTEGER (i.e. suitable for
+// `asn1::BigInt::new`). Unlike `py_uint_to_big_endian_bytes` this accepts
+// negative values, which a general INTEGER field may hold.
+pub(crate) fn py_int_to_der_bytes<'p>(
+    py: pyo3::Python<'p>,
+    v: pyo3::Bound<'p, pyo3::types::PyInt>,
+) -> pyo3::PyResult<PyBackedBytes> {
+    // The number of significant magnitude bits. For negative values we use
+    // `~v` (== `-v - 1`), so that exact negative powers of two (e.g. -128)
+    // get the shorter length their two's complement encoding actually needs.
+    let magnitude = if v.lt(0)? {
+        v.call_method0(pyo3::intern!(py, "__invert__"))?
+    } else {
+        v.clone().into_any()
+    };
+    let bit_length = magnitude
+        .call_method0(pyo3::intern!(py, "bit_length"))?
+        .extract::<usize>()?;
+    // One extra octet leaves room for the sign bit and yields the minimal
+    // DER length once the high bit is accounted for.
+    let length = bit_length / 8 + 1;
+    let kwargs = [("signed", true)].into_py_dict(py)?;
+    Ok(v.call_method(
+        pyo3::intern!(py, "to_bytes"),
+        (length, "big"),
+        Some(&kwargs),
+    )?
+    .extract()?)
+}
+
 pub(crate) fn py_uint_to_big_endian_bytes<'p>(
     py: pyo3::Python<'p>,
     v: pyo3::Bound<'p, pyo3::types::PyInt>,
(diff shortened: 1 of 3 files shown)


src/rust/src/declarative_asn1/encode.rs
```diff
@@ -357,4 +363,49 @@ mod tests {
             assert!(result.is_err());
         });
     }
+
+    #[test]
```
alex: We don't need to add rust tests (except when that's the only way to test something) -- python tests only are fine.

src/rust/src/declarative_asn1/encode.rs
```diff
@@ -202,7 +202,13 @@ impl asn1::Asn1Writable for AnnotatedTypeObject<'_> {
                 Ok(write_value(writer, &val, encoding)?)
             }
             Type::PyInt() => {
-                let val: i64 = value.extract()?;
+                let int_value = value.cast::<pyo3::types::PyInt>()?;
+                let bytes = crate::asn1::py_int_to_der_bytes(py, int_value.clone())?;
```
alex:
```suggestion
                let bytes = crate::asn1::py_int_to_der_bytes(py, int_value)?;
```

there's no need to clone since this is the only user of this value.

src/rust/src/declarative_asn1/encode.rs
```diff
@@ -202,7 +202,13 @@ impl asn1::Asn1Writable for AnnotatedTypeObject<'_> {
...
+                let val = asn1::BigInt::new(&bytes).ok_or_else(|| {
+                    CryptographyError::Py(pyo3::exceptions::PyValueError::new_err(
+                        "invalid value for INTEGER",
+                    ))
+                })?;
```
alex: This or-else should be unreachable -- we should never be computing an invalid encoding. Therefore it should be an `unwrap()`

src/rust/src/asn1.rs
```diff
@@ -67,6 +67,37 @@ fn decode_dss_signature<'p>(
...
+        v.call_method0(pyo3::intern!(py, "__invert__"))?
+    } else {
+        v.clone().into_any()
+    };
+    let bit_length = magnitude
+        .call_method0(pyo3::intern!(py, "bit_length"))?
+        .extract::<usize>()?;
+    // One extra octet leaves room for the sign bit and yields the minimal
+    // DER length once the high bit is accounted for.
+    let length = bit_length / 8 + 1;
```
alex, line 91: I need to think a bit harder to ensure this is correct, but one note to self: a part of this correctness is that if the high bit is set in hte value, then bit_length becomes divisible by 8, otherwise its no (and / 8 rounds down)

alex (commented):
Please take a look at the CI results as well.

Thank you!

src/rust/src/declarative_asn1/encode.rs
dxbjavid: dropped it, the python round-trip cases cover the same values.

src/rust/src/declarative_asn1/encode.rs
dxbjavid: done, switched the cast to cast_into so the value is owned and the clone is gone.

src/rust/src/declarative_asn1/encode.rs
dxbjavid: done. it also acts as a minimality check now, since BigInt::new refuses non-minimal input, so a length bug upstream would fail loudly rather than emit padded DER.

src/rust/src/asn1.rs
dxbjavid: that's the crux of it, yes. with n as the bit length of the magnitude (v itself when non-negative, ~v when negative), the minimal signed length is ceil((n + 1) / 8), and n / 8 + 1 with flooring division equals that for every n, including n = 0 for the values 0 and -1. the python cases at the byte boundaries (127/128, 255/256, -128/-129) pin this down, and the failure modes are loud either way: a length one too short makes to_bytes raise OverflowError, one too long gets rejected by BigInt::new.

dxbjavid:
the CI failure was the flake job, one of the new test lines ran over 79 columns. fixed that and addressed the review comments; flake, the rust tests and the asn1 serialization tests all pass locally now.

alex (approved):

dxbjavid:
Thank you for the review and merge. Much appreciated

PR #15014, opened 2026-06-12
Reject critical nameConstraints extensions containing directoryName constraints

alex:
Per RFC 5280 §4.2.1.10:

> Restrictions of the form directoryName MUST be applied to the subject field in the certificate (when the certificate includes a non-empty subject field) and to any names of type directoryName in the subjectAltName extension.

and:

> If a name constraints extension that is marked as critical imposes constraints on a particular name form, and an instance of that name form appears in the subject field or subjectAltName extension of a subsequent certificate, then the application MUST either process the constraint or reject the certificate.

We don't implement directoryName constraint matching, and previously a directoryName constraint was only rejected (as unsupported) when a subsequent certificate had a directoryName *SAN* — the RFC-mandated application to the *subject* field was silently skipped, even for critical extensions.

With this change, when chain building processes a nameConstraints extension that is **critical** and contains a directoryName entry in either `permittedSubtrees` or `excludedSubtrees`, it returns a fatal error (`critical nameConstraints extension contains an unsupported directoryName constraint`) immediately, without inspecting subjects or SANs. Since subjects are essentially always non-empty, the only conforming alternative to actually processing the constraint is rejection, and rejecting unconditionally keeps the logic trivial. The fatal error also short-circuits path building rather than being swallowed into `CandidatesExhausted` and masked by alternate paths.

Caveats / judgment calls:

- **Non-critical nameConstraints with directoryName entries are ignored** (RFC 5280 only mandates process-or-reject for critical extensions; CABF relaxes nameConstraints criticality to MAY) — *except* that the pre-existing per-SAN evaluation still rejects a directoryName constraint evaluated against a directoryName SAN with the generic "unsupported name constraint" error. So non-critical + directoryName SAN still fails; non-critical + no directoryName SANs now validates.
- **rfc822Name has a similar (legacy) subject requirement we still don't implement**: RFC 5280 §4.2.1.10 says that when rfc822Name constraints are imposed and the certificate has *no* subjectAltName, the constraint MUST be applied to an `emailAddress` attribute in the subject DN. That's the only other name form with a subject-field requirement (x400Address/directoryName-in-SAN aside, all other forms apply to SANs only). Under this profile leaves must carry a SAN, so this can only arise for intermediates with `emailAddress` in their subject — left out of scope here, but a critical rfc822Name constraint over such a chain is technically processed incompletely.
- **Practical effect**: chains with critical directoryName constraints fail loudly instead of the subject-side application being silently unenforced. RFC 5280 requires CAs to mark nameConstraints critical, so most real-world directoryName-constrained chains will now be rejected; that's a behavior change for chains that previously validated.
- The `name-constraint-dn` skip in the x509-limbo suite remains, since those testcases expect real DN matching semantics, which this PR deliberately does not implement.

Tests: new `TestNameConstraints` cases cover critical directoryName constraints (permitted and excluded) being rejected, a critical extension with only DNS constraints still validating, and the two non-critical paths (rejected with a directoryName SAN, ignored without one).

https://claude.ai/code/session_01LJgFCNp7irHqqfQ6jXmPhE

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -61,12 +61,13 @@ Changelog
 * Added :meth:`~cryptography.x509.Name.from_bytes` for parsing a
   :class:`~cryptography.x509.Name` from DER bytes, the inverse of
   :meth:`~cryptography.x509.Name.public_bytes`.
-* The X.509 path validator no longer rejects chains containing a
-  ``directoryName`` name constraint outright. Matching ``directoryName``
-  constraints is still not implemented: such a constraint simply never
-  matches, so an excluded ``directoryName`` subtree never excludes anything
-  and a ``directoryName`` SAN can never satisfy a permitted ``directoryName``
-  subtree.
+* The X.509 path validator now rejects ``directoryName`` name constraints
+  when they apply to a subsequent certificate with a non-empty subject.
+  :rfc:`5280` requires ``directoryName`` constraints to be applied to the
+  subject field as well as ``directoryName`` SANs; since matching
+  ``directoryName`` constraints is not implemented, such chains were
+  previously accepted with the constraint silently ignored, unless the
+  certificate also had a ``directoryName`` SAN.
 * Added the ``rsa_padding`` keyword-only parameter to
   :meth:`~cryptography.x509.CertificateBuilder.public_key`. Passing the
   :class:`~cryptography.hazmat.primitives.asymmetric.padding.PSS` class
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -61,17 +61,6 @@ Changelog
 * Added :meth:`~cryptography.x509.Name.from_bytes` for parsing a
   :class:`~cryptography.x509.Name` from DER bytes, the inverse of
   :meth:`~cryptography.x509.Name.public_bytes`.
-* The X.509 path validator now rejects any chain containing a critical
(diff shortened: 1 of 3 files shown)


woodruffw (approved):
Seems right to me, but sanity-checking myself (and my read of the tests): 

- We _directly_ reject critical NCs that contain directoryName constraints
- We _indirectly_ reject noncritical NCs that contain directoryName constraints, since we don't support matching DNs
- We _ignore_ noncritical NCs that contain directoryName constraints if we don't have any SANs that contain DNs

If that's right then this looks good to me!

alex:
That's exactly right, on all three bullets. The only nuance worth spelling out: for the two noncritical cases, only SANs matter — the subject field plays no role in the noncritical paths at all. The RFC's requirement to apply directoryName constraints to the *subject* is what motivates the unconditional rejection in the critical case (since subjects are essentially always non-empty, a critical directoryName constraint would virtually always apply to something we can't match); for noncritical extensions we're allowed to skip what we can't process, so they just fall through to the existing per-SAN evaluation, which rejects on a directoryName-vs-directoryName-SAN pairing and otherwise ignores the constraint.

The tests cover the three cases in the same order: `test_critical_dirname_constraint_rejected`, `test_noncritical_dirname_constraint_with_dirname_san`, and `test_noncritical_dirname_constraint_ignored`.

---

reaperhulk (approved):

PR #15015, opened 2026-06-12
Add more robust key processing: RSA

sjudson:
This PR adds more robust key processing for RSA by extending checks (e.g., that `e != 1`) to the PEM and DER load paths. To maintain a single set of checks across all paths, it updates the existing direct construction flow to also validate over the public key components as `BigNum`s. 

See https://github.com/pyca/cryptography/pull/14992 for related changes to DSA processing.

### DIFF
--- a/src/rust/src/backend/rsa.rs
+++ b/src/rust/src/backend/rsa.rs
@@ -797,26 +797,15 @@ impl RsaPrivateNumbers {
     }
 }
 
-fn check_public_key_components_from_py(
+fn py_int_to_signed_bn(
     py: pyo3::Python<'_>,
-    e: &pyo3::Bound<'_, pyo3::types::PyInt>,
-    n: &pyo3::Bound<'_, pyo3::types::PyInt>,
-) -> CryptographyResult<()> {
-    if n.lt(0)? {
-        return Err(CryptographyError::from(
-            pyo3::exceptions::PyValueError::new_err("n must be >= 3."),
-        ));
-    }
-
-    if e.lt(0)? {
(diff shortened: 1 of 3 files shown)


src/rust/src/backend/rsa.rs
```diff
@@ -795,23 +797,47 @@ impl RsaPrivateNumbers {
...
 }
 
-fn check_public_key_components(
+fn check_public_key_components_from_py(
+    py: pyo3::Python<'_>,
     e: &pyo3::Bound<'_, pyo3::types::PyInt>,
     n: &pyo3::Bound<'_, pyo3::types::PyInt>,
 ) -> CryptographyResult<()> {
-    if n.lt(3)? {
+    if n.lt(0)? {
```
reaperhulk: `check_public_key_components_from_py` has error paths (one of which doesn't have the condition match the exception text), but then it calls `check_public_key_components`, which has the same error paths. This function also appears to only have one caller, so can we not just inline the py_int_to_bn calls and use the check_public_key_components path?

sjudson: @reaperhulk updated to remove the early exit, and consolidate all checks at the same layer.

CI is red, but as far as I can tell the error looks intermittent to me.

reaperhulk: This is a known bug in the RSA paths of Ubuntu 22.04 that they never backported the fix for, so yeah, just a flake. I'll review this soon, sorry for the delay

reaperhulk (approved):

PR #15016, opened 2026-06-12
Add more robust parameter processing: DH

sjudson:
This PR adds more robust key processing for DH by extending existing heuristic checks on parameter well-formedness to the PEM and DER load paths.

See https://github.com/pyca/cryptography/pull/14992 and https://github.com/pyca/cryptography/pull/15015 for related changes to DSA and RSA processing.

### DIFF
--- a/src/rust/src/backend/dh.rs
+++ b/src/rust/src/backend/dh.rs
@@ -55,18 +55,20 @@ fn generate_parameters(
 
 pub(crate) fn private_key_from_pkey(
     pkey: &openssl::pkey::PKeyRef<openssl::pkey::Private>,
-) -> DHPrivateKey {
-    DHPrivateKey {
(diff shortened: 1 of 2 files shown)


reaperhulk (approved):

src/rust/src/backend/dh.rs
```diff
@@ -119,13 +121,19 @@ fn dh_parameters_from_numbers(
     let g = utils::py_int_to_bn(py, numbers.g.bind(py))?;
 
     let dh = openssl::dh::Dh::from_pqg(p, q, g)?;
+    check_dh_parameters(&dh)?;
```
garbusbeach, line 124: @sjudson please see https://github.com/pyca/cryptography/issues/15042

PR #15018, opened 2026-06-12
Strengthen validation of end-entity key usage when present

sjudson:
This PR aligns the default behavior of the server and client certificate verifiers with modern CABF requirements. Along with the existing negative check that `keyCertSign` is not set, we add the requirement that `digitalSignature` is.

As before, this default validation can be overridden by way of the `ExtensionPolicy` interface.

### DIFF
--- a/src/rust/cryptography-x509-verification/src/policy/extension.rs
+++ b/src/rust/cryptography-x509-verification/src/policy/extension.rs
@@ -640,6 +640,12 @@ mod ee {
                     "EE keyUsage must not assert keyCertSign".to_string(),
                 )));
             }
+
+            if !key_usage.digital_signature() {
+                return Err(ValidationError::new(ValidationErrorKind::Other(
+                    "EE keyUsage must assert digitalSignature when present".to_string(),
+                )));
+            }
         }
 
         Ok(())
--- a/src/rust/cryptography-x509-verification/src/policy/extension.rs
+++ b/src/rust/cryptography-x509-verification/src/policy/extension.rs
@@ -551,6 +551,7 @@ impl<'cb, B: CryptoOps> ExtensionValidator<'cb, B> {
 }
 
 mod ee {
+    use cryptography_x509::common::AlgorithmParameters;
     use cryptography_x509::extensions::{BasicConstraints, ExtendedKeyUsage, Extension, KeyUsage};
 
     use crate::ops::{CryptoOps, VerificationCertificate};
@@ -629,7 +630,7 @@ mod ee {
 
     pub(crate) fn key_usage<'chain, B: CryptoOps>(
         _policy: &Policy<'_, B>,
-        _cert: &VerificationCertificate<'chain, B>,
+        cert: &VerificationCertificate<'chain, B>,
         extn: Option<&Extension<'_>>,
     ) -> ValidationResult<'chain, (), B> {
         if let Some(extn) = extn {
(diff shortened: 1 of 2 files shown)


src/rust/cryptography-x509-verification/src/policy/extension.rs
```diff
@@ -640,6 +640,12 @@ mod ee {
                     "EE keyUsage must not assert keyCertSign".to_string(),
                 )));
             }
+
+            if !key_usage.digital_signature() {
+                return Err(ValidationError::new(ValidationErrorKind::Other(
+                    "EE keyUsage must assert digitalSignature when present".to_string(),
```
woodruffw: At this point we have a `keyUsage`, so "when present" feels redundant 🙂 

```suggestion
                    "EE keyUsage must assert digitalSignature".to_string(),
```

src/rust/cryptography-x509-verification/src/policy/extension.rs
```diff
@@ -640,6 +640,12 @@ mod ee {
                     "EE keyUsage must not assert keyCertSign".to_string(),
                 )));
             }
+
+            if !key_usage.digital_signature() {
```
woodruffw: Note: CABF 7.1.2.7.11 requires `digitalSignature` for subscriber certs (i.e. EEs) that convey ECC keys, but does **not** require it for subscriber certs that convey RSA keys:

Image

This is an incredible mess in practice: the BR says that an RSA-conveying subscriber cert **MUST** have at least one keyUsage, but no specific one is a **MUST** in practice unless the validator knows the TLS version being used.

Given that, I think we have two options here:

1. Continue to just enforce `digitalSignature`. This...probably works fine in practice, even though it's technically stricter than what the BRs say?
2. Differentiate this check based on the subscriber's key type: for ECC keys it should be the current check, while for RSA keys it should at be at least one of `digitalSignature | keyEncipherment | dataEncipherment`.

Curious what @alex and @reaperhulk think about this as well 🙂

woodruffw (commented):
Thanks @sjudson!

src/rust/cryptography-x509-verification/src/policy/extension.rs
alex: I guess I'm still unsure if we want to enforce this at all, or if we want to allow configuring it on the policy builder? (Or just require users to provide their own extension validator for KU)

woodruffw: I think it makes sense to enforce in our default policy, since we're otherwise relatively strict about CABF (and the limbo results suggest that other validators check this). But it is indeed somewhat annoying to enforce correctly.

alex: Do we know if the other validators are key-type aware?

woodruffw: > Do we know if the other validators are key-type aware?

Didn't test that yet, will do so tonight 🙂

src/rust/cryptography-x509-verification/src/policy/extension.rs
sjudson: I've cut this.

src/rust/cryptography-x509-verification/src/policy/extension.rs
sjudson: Yes the word "modern" was doing perhaps a bit too much heavy lifting in my initial description. 

I've now added distinct handling for RSA that considers `keyEncipherment` sufficient (and allows both, despite being NOT RECOMMENDED) to be more permissive for the default validation.

reaperhulk: @woodruffw did you get a chance to test this?

PR #15046, opened 2026-06-16
reject trailing data when parsing signed certificate timestamps

dxbjavid:
Signed Certificate Timestamps embedded in a certificate are parsed by a small hand-rolled TLS reader rather than the DER machinery. The reader walks the length-prefixed list and then each fixed SCT structure, but it never checks that it reached the end of either, so any bytes left over after the list, or after the final signature field inside an individual SCT entry, were quietly discarded. Since the SCT bytes come straight from an untrusted certificate extension, a malformed encoding with appended padding parsed as a valid SCT, and the leftover bytes still ended up in the stored raw data used for equality and hashing. I noticed it while comparing this reader against the DER and SSH parsers, both of which reject leftover input. The fix verifies each reader is empty once the expected fields have been consumed. I kept it inside parse_scts and added a regression test covering both leftover positions.

### DIFF
--- a/src/rust/src/x509/sct.rs
+++ b/src/rust/src/x509/sct.rs
@@ -223,7 +223,15 @@ pub(crate) fn parse_scts<'p>(
     data: &[u8],
     entry_type: LogEntryType,
 ) -> CryptographyResult<pyo3::Bound<'p, pyo3::PyAny>> {
-    let mut reader = TLSReader::new(data).read_length_prefixed()?;
+    let mut outer = TLSReader::new(data);
+    let mut reader = outer.read_length_prefixed()?;
+    // The length-prefixed list must span the whole input; anything left over
+    // is a malformed encoding rather than a second list.
+    if !outer.is_empty() {
+        return Err(CryptographyError::from(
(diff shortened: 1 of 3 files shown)


src/rust/src/x509/sct.rs
```diff
@@ -262,6 +278,51 @@ pub(crate) fn parse_scts<'p>(
 mod tests {
     use super::*;
 
+    // Builds a minimal but well-formed SCT body (version, log id, timestamp,
+    // empty extensions, hash/signature algorithm and an empty signature).
+    fn minimal_sct_body() -> Vec<u8> {
```
alex: In general, unless there's a compelling reason to do otherwise, we prefer to test things from the Python API.

dxbjavid: makes sense, dropped the rust unit test and moved it to a python test in test_x509_ext.py. it builds a cert with a crafted precert SCT extension and checks the well-formed list still parses while a trailing byte either inside an SCT entry or after the list raises ValueError.

alex (approved):

dxbjavid:
Thank you for the review and merge. Much appreciated

PR #15063, opened 2026-06-18
Initial abi3t support

ngoldbaum:
Build abi3t wheels on python 3.15 and newer.


src/rust/cryptography-cffi/build.rs
```diff
@@ -106,16 +106,29 @@ fn main() {
         &python,
         "import sysconfig; print(bool(sysconfig.get_config_var('Py_GIL_DISABLED')), end='')",
     )
+    .unwrap()
+        == "True";
+    let at_least_py_315 = run_python_script(
```
alex: Do we need to be getting this from pyo3 build config, or is it fine to rely on the running interpreter.

src/rust/cryptography-cffi/build.rs
```diff
@@ -106,16 +106,29 @@ fn main() {
...
+        &python,
+        "import sys; print(sys.version_info > (3, 14), end='')",
+    )
     .unwrap()
         == "True";
 
-    // Enable abi3 mode if we're not using PyPy or the free-threaded build
-    if !(python_impl == "PyPy" || is_free_threaded) {
+    // Enable abi3 mode if we're not using PyPy or the free-threaded build on 3.14 and older
+    if !(python_impl == "PyPy" || is_free_threaded || at_least_py_315) {
```
alex: is this right? (It's very hard to follow)

src/rust/cryptography-cffi/build.rs
```diff
@@ -106,16 +106,29 @@ fn main() {
...
         // cp39 (Python 3.9 to help our grep when we some day drop 3.9 support)
-        build.define("Py_LIMITED_API", "0x030900f0");
+        build.define("Py_LIMITED_API", "0x030900F0");
+    }
+
+    // Enable abi3t mode if we're targeting Python 3.15 or newer
+    if at_least_py_315 {
+        // cp315 (Python 3.15 to help out grep when we some day drop 3.15 support)
+        build.define("Py_LIMITED_API", "0x030F0000");
+        build.define("Py_GIL_DISABLED", "1");
```
alex: Is GIL always disabled on 3.15?

alex:
Thanks for working on this!

src/rust/cryptography-cffi/build.rs
ngoldbaum: Setting `Py_LIMITED_API >= 0x030F0000` and `Py_GIL_DISABLED` is equivalent to selecting abi3t.

So no it's not always disabled but abi3t extensions have to be written as if it might be.

See https://docs.python.org/3.15/howto/abi3t-migration.html#avoid-build-time-conditionals

src/rust/cryptography-cffi/build.rs
ngoldbaum: I can probably make these conditionals less tortured. I'm relying on the fact that pypy 3.15 will probably never be a thing.

src/rust/cryptography-cffi/build.rs
ngoldbaum: That's a good question, I bet we can get it from PyO3 in case of cross-compiling...

src/rust/cryptography-cffi/build.rs
alex: Ok. I think this needs to be guarded on the target python version, not whether the host python is 3.15?

src/rust/cryptography-cffi/build.rs
alex: RIP 🥲

src/rust/cryptography-cffi/build.rs
```diff
@@ -106,16 +106,29 @@ fn main() {
         &python,
         "import sysconfig; print(bool(sysconfig.get_config_var('Py_GIL_DISABLED')), end='')",
     )
+    .unwrap()
+        == "True";
+    let at_least_py_315 = run_python_script(
+        &python,
+        "import sys; print(sys.version_info > (3, 14), end='')",
```
ngoldbaum: The 3.14 CI is failing because this should be a 15 instead of a 14. You can tell an AI didn't write this code because it has a dumb bug like this 🤣

pyproject.toml
```diff
@@ -2,10 +2,10 @@
 # These requirements must be kept sync with the requirements in
 # ./.github/requirements/build-requirements.{in,txt}
 requires = [
-    "maturin>=1.9.4,<2,!=1.12.0",
+    "maturin @ git+https://github.com/pyo3/maturin",
```
ngoldbaum: This is needed because adding both `abi3` and `abi3t` features to `cargo.toml` is broken in the released version of Maturin and we need a release with https://github.com/PyO3/maturin/pull/3226 included.

ngoldbaum: Actually I think there's a bug that https://github.com/PyO3/maturin/pull/3226 introduced. I have a fix that I'm testing now, along with a test crate for maturin that models how cryptography shells out to CFFI in a build script.

ngoldbaum:
OK great, it looks like the only failing CI jobs use a Rust compiler that's too old to build Maturin. There are two Windows jobs that timed out but I'm not worried about that because building Maturin on Windows is very slow.

I opened https://github.com/PyO3/maturin/pull/3233 to fix the bug in Maturin.

I'll try to work on getting the CFFI and Maturin PRs that I'm depending on here merged.

alex:
In practice those iwll be solved by a maturin release with wheels

ngoldbaum:
Maturin cut a 1.14.1 release so now all that's needed is CFFI.

ngoldbaum:
Looks like unconditionally requiring Maturin 1.14.1 or newer works in CI 🥳 

I can probably upstream part of this while we wait for CFFI to do a release but for now I think I'll just wait for that before finishing this up.

alex:
The `src/rust/cryptography-cffi/build.rs` bits look like they could be split out if you wanted.

ngoldbaum:
@alex I think this is ready now.

reaperhulk:
This is now a very small PR 😄 What's the next step after merging this?

alex (commented):
Do we need to update version bounds on cffi or anything?

ngoldbaum:
> Do we need to update version bounds on cffi or anything?

I don't think so. The first CFFI release with Python 3.15 support is CFFI 2.1, which is >= 2.0. 2.0 is fine for older Python versions.

> What's the next step after merging this?

Probably worth doing a dry-run of the release automation to verify that you do actually get three wheels per platform out to upload to PyPI. Other than that, I think just a release advertising Python 3.15 support.

alex:
Do we need >=2.1; python_version >= '3.15'?

ngoldbaum:
> Do we need >=2.1; python_version >= '3.15'?

Sure, let's add that just in case someone does something weird.

pyproject.toml
```diff
@@ -2,10 +2,11 @@
 # These requirements must be kept sync with the requirements in
 # ./.github/requirements/build-requirements.{in,txt}
 requires = [
-    "maturin>=1.9.4,<2,!=1.12.0",
+    "maturin>=1.14.1,<2",
 
     # Must be kept in sync with `project.dependencies`
     "cffi>=2.0.0; platform_python_implementation != 'PyPy'",
+    "cffi>=2.1; python_version >= '3.15'",
```
alex: ugh, do we need the `and platform_python_implementation != 'PyPy'` just in case?

ngoldbaum: I'm pretty sure we're safe given the status of Python 3.12 support in PyPy, but let's add it just to cover our bases I guess...

https://github.com/orgs/pypy/discussions/5145

alex:
better safe than sorry 🙃

alex (approved):

PR #15071, opened 2026-06-19
Use pyo3-build-config for cryptography-cffi build script

ngoldbaum:
Split off from #15063 to land separately from abi3t support.

I could make the match arm for abi3t panic instead of allowing the build but also it seems harmless to me to allow abi3t builds for now but leave them unexercised.


alex (approved):
nice cleanup!

PR #15076, opened 2026-06-21
raise ValueError for unsupported name attribute value tags

dxbjavid:
Parsing an X.509 name attribute whose value carries an ASN.1 tag outside the recognised string types looked the tag up in the _ASN1_TYPE_TO_ENUM dict without handling a miss, so a crafted certificate, CSR, CRL or OCSP response made accessing subject or issuer raise a bare KeyError rather than the ValueError every other malformed-input path raises. The value itself is already accepted into the catch-all AnyString variant, so only the type lookup was left unguarded. I map the miss to an invalid-value parse error so it surfaces as a ValueError with the usual field location, and add a regression test that mutates a built certificate's commonName tag.


tests/x509/test_x509.py
```diff
@@ -1334,6 +1334,34 @@ def test_invalid_unicode_name(self, backend):
...
+            .not_valid_before(datetime.datetime(2020, 1, 1))
+            .not_valid_after(datetime.datetime(2030, 1, 1))
+            .sign(key, hashes.SHA256())
+        )
+        der = bytearray(cert.public_bytes(serialization.Encoding.DER))
+        # Find id-at-commonName (2.5.4.3) and replace the following
+        # UTF8String tag with an unsupported tag value. The first occurrence
+        # is the issuer name.
+        idx = der.index(b"\x06\x03\x55\x04\x03")
+        der[idx + 5] = 0x69
```
alex: The right way to do these tests is with a checked in vector, not by generating and then manipulating the value

src/rust/src/x509/common.rs
```diff
@@ -234,7 +234,13 @@ fn parse_name_attribute<'p>(
...
         ))
     })?;
-    let py_tag = types::ASN1_TYPE_TO_ENUM.get(py)?.get_item(tag_val)?;
+    // A value whose tag is not one of the recognised string types is not a
+    // valid AttributeValue; surface it as a parse error rather than letting the
+    // dict lookup raise a bare KeyError.
+    let py_tag = types::ASN1_TYPE_TO_ENUM
+        .get(py)?
+        .get_item(tag_val)
+        .map_err(|_| asn1::ParseError::new(asn1::ParseErrorKind::InvalidValue))?;
```
alex: I think UnexpectedTag is a more correct error to return here.

alex (commented):
sorry about the slow review

src/rust/src/x509/common.rs
dxbjavid: switched it to UnexpectedTag, with the value's actual tag carried along. that does read better here.

tests/x509/test_x509.py
dxbjavid: fair point. dropped the generate-and-mutate and checked in a vector instead (name_attribute_unsupported_tag.pem), a cert with the commonName value tag set to an unsupported one in both names. test just loads it and checks subject and issuer both raise.

vectors/cryptography_vectors/x509/custom/name_attribute_unsupported_tag.pem
```diff
@@ -0,0 +1,8 @@
+-----BEGIN CERTIFICATE-----
```
alex, line 1: please document in test-vectors.rst

dxbjavid: done, added an entry for it under the x509 custom certificates in test-vectors.rst.

docs/development/test-vectors.rst
```diff
@@ -474,6 +474,9 @@ Custom X.509 Vectors
   "We heart UTF8!™".
 * ``invalid_utf8_common_name.pem`` - A certificate that contains a
   ``UTF8String`` common name with an invalid UTF-8 byte sequence.
+* ``name_attribute_unsupported_tag.pem`` - A certificate whose subject and
+  issuer common name values carry an ASN.1 tag (``0x69``) that is not one of
+  the recognised string types. The signature on this certificate is invalid.
```
alex:
```suggestion
  the recognized string types. The signature on this certificate is invalid.
```

dxbjavid: done, switched it to the american spelling to match the rest of the file.

alex (approved):
thanks

dxbjavid:
Thank you for your time. Appreciated.

PR #15097, opened 2026-06-25
fix: mark tests expecting malloc to fail as opt-in

CertainLach:
Those tests might overload machines with huge amount of RAM or with overcommit enabled, because of that - one test was already disabled on darwin, which enables overcommit by default.

Because of the potentially destructive overhead, I believe they should be opt-in.

Ref: Read the comments in https://github.com/pyca/cryptography/pull/14782

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -29,12 +29,21 @@ def pytest_addoption(parser):
     parser.addoption("--wycheproof-root", default=None)
     parser.addoption("--x509-limbo-root", default=None)
     parser.addoption("--enable-fips", default=False)
+    parser.addoption(
+        "--enable-malloc-failure", action="store_true", default=False
+    )
 
 
 def pytest_runtest_setup(item):
     if openssl_backend._fips_enabled:
         for marker in item.iter_markers(name="skip_fips"):
             pytest.skip(marker.kwargs["reason"])
+    if not item.config.getoption("--enable-malloc-failure"):
+        for _ in item.iter_markers(name="malloc_failure"):
+            pytest.skip(
+                "expects malloc to fail on pressure, "
+                "pass --enable-malloc-failure to run"
+            )
 
 
 @pytest.fixture(autouse=True)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,6 +3,8 @@
 # for complete details.
 
 import contextlib
+import functools
+import mmap
 
 import pytest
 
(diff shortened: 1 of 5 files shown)


CertainLach:
Cc: @alex as the author of https://github.com/pyca/cryptography/pull/14782

alex:
Is there a particular place this is causing a problem, or just in theory? Tests should (probably _must_ ) be on by default to be effective, so I don't want to disable things by default without an exceptionally good reason.

CertainLach:
> Is there a particular place this is causing a problem, or just in theory?

This test was already disabled on darwin, and it fails on any linux machine with overcommit enabled (`sysctl -w vm.overcommit_memory=1`)/machines with >4TiB of memory, because malloc never fails in this case, process is getting killed by OOM Killer without MemoryError being thrown.

I have noticed this problem because I was getting OOM during this package build in NixOS, but this may happen regardless of the distribution, the original PR has a comment telling about the OOM on Gentoo machine.

This test should not be run by default, because by spec, malloc only fails in case of system is unable to allocate memory, but with overcommit memory allocation always succeeds, regardless of the size.

alex:
I'd be happy to consider a PR that checks the sysctl for overcommit and skips in those cases, but right now this PR makes these tests less useless for the developers of this software.

CertainLach:
I'm not sure if checking sysctl is a good idea here, it might not be available in docker/with selinux/other sandboxing, while explicit `--enable-malloc-failure` allows the developer to manually specify the behavior they expect... This is also a source of impurity, as some sandboxes might expose this value, but then expect deterministic set of tests to pass, but I'm not sure if that would affect anyone.

Better behavior in my opinion would be trying to allocate 4TiB block, and then skipping `pytest.mark.malloc_failure` depending on this allocation succeeding or not, ~~but it would still fail on machines with 4TiB+ RAM (But I'm fine if this is the accepted solution, I don't have 4TiB of RAM to be affected by that :D)~~ - Actually, it won't, since we would disable this test for such machines and treat them as overcommitting, which is good, since it is not good when one test might occupy this much RAM.

CertainLach:
Implemented, but have not tested yet the second solution I proposed (probe for overcommit, skip tests if environment allows overcommits)

Just to be clear - OOM happens not because of malloc, but because it actually uses the allocated memory, so this probe won't cause the same problem

tests/conftest.py
```diff
@@ -25,6 +27,18 @@ def pytest_report_header(config):
     )
 
 
+@functools.cache
+def overcommit_probe() -> bool:
+    # Try to allocate 4TiB of memory, if it succeeds - environment either
+    # has more RAM, or overcommits.
+    try:
+        m = mmap.mmap(-1, (2**32 - 1) * 1024)
+    except (OSError, OverflowError):
```
CertainLach, line 36: CI was failing on armv7l with OverflowError here, because 4TiB can't be converted to ssize_t (2GB) on 32bit platforms
I don't think it makes sense to check for overcommit in such cases, all the malloc failure tests allocate much more that 2GB already and will never fail on overcommit enabled on 32bit platform

CertainLach:
Tested in overcommiting environment - the test gets properly skipped, and I don't see it being skipped in CI, so everything seems to be fine

CertainLach:
Also found out two downstreams were already disabling one of the failing tests, exactly because of overcommits:

https://github.com/NixOS/nixpkgs/commit/b8e8ef279380136c1ceccb4c442afec26d7ef41f

https://github.com/gentoo/gentoo/blob/f96c060220bb94c22f417d42bc9933843af18b45/dev-python/cryptography/cryptography-48.0.0.ebuild#L140

With the new pytest mark added, new such tests would no longer be disruptive for maintainers

alex (approved):

PR #15105, opened 2026-06-28
Add PolicyBuilder algorithm policy API for X509 verification

drmikecrypto:


alex:
This appears to be completely AI generated, and you ignored the fact that there was still open design questions.

We have no problem with using LLMs to work on this project, but you need to actually put in effort. In this case, you need to participate in the design question and decide what the correct API is before you generate some slop.

PR #15110, opened 2026-06-29
Support encode_der for top-level asn1.SetOf

### LINKED ISSUE #15107: Support asn1.encode_der for asn1.SetOf
Currently the following code fails:

```
from cryptography.hazmat import asn1

asn1.encode_der(asn1.SetOf([1,2,3]))
```

It would be very helpful when a SetOf could be encoded to asn1 directly. My use case is the Signed Attributes of a CMS Signature, they are a SetOf, but they must be encoded as DER to work as the input for the signature. My current workaround is having a dummy sequence with only one field, encoding that then parsing it to an `asn1.TLV` and extracting the data but direct support would be supper helpful here.

alex:
`asn1.encode_der` previously rejected a bare `SetOf` instance with a `TypeError`, because a `SetOf` carries no information about its element type (e.g. the `int` in `SetOf[int]`), and the encoder only dispatched on a value's Python type.

This infers the element type from the contained values when a `SetOf` is passed directly to `encode_der`, which is enough to encode it as a `SET OF`. This is useful for cases like producing the DER encoding of CMS signed attributes for signing.

```python
from cryptography.hazmat import asn1

asn1.encode_der(asn1.SetOf([3, 1, 2]))
# b'1\t\x02\x01\x01\x02\x01\x02\x02\x01\x03'
```

Decoding a top-level `SetOf` remains unsupported, since `decode_der` needs a concrete element type that a bare `SetOf` doesn't carry.

Closes #15107

---
_Generated by [Claude Code](https://claude.ai/code/session_01T3rrycoCFuiUTpM9AodStP)_

### DIFF
--- a/src/rust/src/declarative_asn1/asn1.rs
+++ b/src/rust/src/declarative_asn1/asn1.rs
@@ -3,7 +3,6 @@
 // for complete details.
 
 use asn1::Asn1Writable;
-use pyo3::types::PyAnyMethods;
 
 use crate::declarative_asn1::decode::decode_annotated_type;
 use crate::declarative_asn1::types as asn1_types;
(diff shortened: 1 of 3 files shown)


reaperhulk (approved):

loewexy:
I think this MR is missing the changelog entry.

PR #15113, opened 2026-06-29
Support encode_der for asn1.TLV values

### LINKED ISSUE #15109: Support asn1.encode_der for asn1.TLV
Currently the following code fails:

```
from cryptography.hazmat import asn1

tlv_obj = asn1.decode_der(asn1.TLV, asn1.encode_der(1))

asn1.encode_der(tlv_obj)
```

This would be helpful for cases where an arbitrary asn1 part is parsed using the TLV specifier to serialize only that part of the structure back to DER.

alex:
Closes #15109.

Previously, calling `asn1.encode_der` on a value decoded with the `asn1.TLV` specifier raised `NotImplementedError`, making it impossible to parse an arbitrary ASN.1 element as a `TLV` and then serialize just that element back to DER:

```python
from cryptography.hazmat import asn1

tlv_obj = asn1.decode_der(asn1.TLV, asn1.encode_der(1))
asn1.encode_der(tlv_obj)  # previously raised NotImplementedError
```

## Change

Implements the `Type::Tlv()` arm of the declarative ASN.1 encoder. A `TLV`'s `full_data` is a complete, already-validated DER element, so encoding re-parses it into an `asn1::Tlv` and writes it back out verbatim:

- **No annotation** (top-level / `ANY` field) → written as-is.
- **EXPLICIT annotation** → re-wrapped in the explicit tag, mirroring the existing decode path.
- **IMPLICIT annotation** → still rejected (consistent with decoding, and already rejected at definition time at the Python level).

The re-parse is necessary because the `TLV` Python object stores the element's raw DER bytes rather than a live `asn1::Tlv<'_>` (which borrows its input and can't be held in a pyclass); the writer needs an `Asn1Writable`, and `asn1::Tlv` has no public constructor, so re-parsing `full_data` is how we recover one. It only reads the tag/length header and can't fail in practice, since the bytes came from a successful decode.

## Tests

- `test_ok_encode_tlv` round-trips several elements (INTEGER, BIT STRING, SEQUENCE, NULL) through `decode_der(TLV, ...)` → `encode_der`.
- `test_ok_encode_tlv_issue_example` covers the exact example from the issue.
- Extended `test_sequence_with_tlv_with_explicit_annotation` to assert the EXPLICIT-tagged TLV fields re-encode to the original DER.
- Added a Rust unit test (`test_encode_implicit_tlv`) covering the defensive implicit-encoding branch, for parity with the decode side.

---
_Generated by [Claude Code](https://claude.ai/code/session_016iPBDfr6FxfNSm6nzakgwh)_

### DIFF
--- a/tests/hazmat/asn1/test_serialization.py
+++ b/tests/hazmat/asn1/test_serialization.py
@@ -328,14 +328,23 @@ class TestTLV:
         assert isinstance(decoded_example, Example)
         assert decoded_example.foo == 9
 
-    def test_fail_encode_tlv(self) -> None:
-        tlv = asn1.decode_der(asn1.TLV, b"\x03\x02\x07\x40")
-        assert isinstance(tlv, asn1.TLV)
+    def test_ok_encode_tlv(self) -> None:
+        for original in [
+            b"\x03\x02\x07\x40",
+            b"\x02\x01\x01",
+            b"\x30\x03\x02\x01\x09",
+            b"\x05\x00",
+        ]:
+            tlv = asn1.decode_der(asn1.TLV, original)
+            assert isinstance(tlv, asn1.TLV)
+            assert asn1.encode_der(tlv) == original
 
-        with pytest.raises(
-            NotImplementedError, match="TLV encoding currently not supported"
-        ):
-            asn1.encode_der(tlv)
+    def test_ok_encode_tlv_issue_example(self) -> None:
+        # The example from the original feature request: parse an
+        # arbitrary part with the TLV specifier, then serialize it back.
+        tlv = asn1.decode_der(asn1.TLV, asn1.encode_der(1))
+        assert isinstance(tlv, asn1.TLV)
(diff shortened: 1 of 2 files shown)


tests/hazmat/asn1/test_serialization.py
```diff
@@ -328,14 +328,23 @@ class Example:
...
+            b"\x03\x02\x07\x40",
+            b"\x02\x01\x01",
+            b"\x30\x03\x02\x01\x09",
+            b"\x05\x00",
+        ]:
+            tlv = asn1.decode_der(asn1.TLV, original)
+            assert isinstance(tlv, asn1.TLV)
+            assert asn1.encode_der(tlv) == original
+
+    def test_ok_encode_tlv_issue_example(self) -> None:
```
reaperhulk: This test name (and associated comment) isn't going to age well

reaperhulk (approved):

loewexy:
I think this MR is missing the changelog entry.

PR #15129, opened 2026-07-01
Add AWS-LC support for XOFHash

reaperhulk:
This PR extends XOFHash (Extendable Output Function) support to AWS-LC, in addition to the existing OpenSSL 3.3.0+ support.

## Summary

Previously, XOFHash was only supported on OpenSSL 3.3.0 or greater. This change adds support for AWS-LC while maintaining backward compatibility with existing backends.

## Key Changes

- **Rust backend (`src/rust/src/backend/hashes.rs`)**:
  - Updated the XOF support check to include AWS-LC alongside OpenSSL 3.3.0+
  - Simplified the conditional compilation logic by inverting the check: now enables XOF when `CRYPTOGRAPHY_OPENSSL_330_OR_GREATER` OR `CRYPTOGRAPHY_IS_AWSLC` is true
  - Updated error message to be backend-agnostic ("not supported on this backend" instead of specifically mentioning LibreSSL/BoringSSL)

- **Python tests (`tests/hazmat/primitives/test_xofhash.py`)**:
  - Extracted XOF support detection into a reusable `_xof_supported()` helper function
  - Updated all test markers to use the new helper function for consistency
  - Renamed `test_unsupported_boring_libre` to `test_unsupported_xof` to reflect the more general nature of the test

- **Hash tests (`tests/hazmat/primitives/test_hashes.py`)**:
  - Updated `test_hash_xof` to use the backend's `hash_supported()` method instead of directly checking OpenSSL version
  - Removed direct dependency on `rust_openssl` module in this test file

- **Changelog**: Added entry documenting XOFHash support for AWS-LC

## Implementation Details

The changes consolidate the XOF support logic by using a positive check (backends that support XOF) rather than a negative check (backends that don't support it). This makes the code more maintainable and easier to extend to additional backends in the future.

https://claude.ai/code/session_01SCjKRh9sZ4qXDfoVm5YAJ3

### DIFF
--- a/tests/hazmat/primitives/test_hashes.py
+++ b/tests/hazmat/primitives/test_hashes.py
@@ -8,7 +8,6 @@ import binascii
 import pytest
 
 from cryptography.exceptions import AlreadyFinalized, _Reasons
-from cryptography.hazmat.bindings._rust import openssl as rust_openssl
 from cryptography.hazmat.primitives import hashes
 
 from ...doubles import DummyHashAlgorithm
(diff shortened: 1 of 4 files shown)


tests/hazmat/primitives/test_hashes.py
```diff
@@ -192,10 +191,12 @@ def test_hash_unsupported_algorithm(self):
...
 
     @pytest.mark.supported(
-        only_if=lambda _: rust_openssl.CRYPTOGRAPHY_OPENSSL_330_OR_GREATER,
+        only_if=lambda backend: backend.hash_supported(
+            hashes.SHAKE256(digest_size=32)
+        ),
         skip_message="Requires backend with XOF support",
     )
-    def test_hash_xof(self):
+    def test_hash_xof(self, backend):
```
alex: I don't htink the backend arg is required, is it?

alex (approved):

PR #15143, opened 2026-07-03
Add MLDSAMuHasher for incremental mu computation

### LINKED ISSUE #15060: Provide example to calculate mu for ML-DSA
It would be nice to have some example code for calculating mu.

I wrote the attached one, which is inspired by https://github.com/nsmithuk/mldsa-mu-py/blob/main/src/mldsa_mu/mu.py.

I first generate a signature, then i generate mu manually and verify the signature using verify_mu. Then I also generate a hashmldsa signature and verify it using openssl (taken from wsl in my example).

[hashmldsa.py](https://github.com/user-attachments/files/29091846/hashmldsa.py)

reaperhulk:
Adds `MLDSAMuHasher`, an incremental (`update`/`copy`/`finalize`) API for computing the ML-DSA message representative (`mu`) consumed by the external-mu APIs (`sign_mu`/`verify_mu`), per FIPS 204: `mu = SHAKE256(SHAKE256(pk, 64) || 0x00 || len(ctx) || ctx || M, 64)`.

- Implemented in Rust on the EVP SHAKE256 XOF, mirroring the `Hash` interface. The public-key/context prefix is absorbed at construction; the message is streamed via `update()`, so large or streaming messages need not be buffered.
- Takes an ML-DSA public key only (mu is a public-key operation), plus an optional context of up to 255 bytes.
- On backends without SHAKE256 through the EVP interface (BoringSSL/LibreSSL), construction raises `UnsupportedAlgorithm`; `sign_mu`/`verify_mu` still work there with a mu computed elsewhere.
- Verified against the NIST KAT vectors and an independent SHAKE256 reference, and round-trips through `sign_mu`/`verify_mu`.

https://claude.ai/code/session_01EWdcBuM5mhzA43zsQaEfET


docs/hazmat/primitives/asymmetric/mldsa.rst
```diff
@@ -38,6 +38,32 @@ different contexts or protocols.
...
+
+ML-DSA supports an "external mu" mode, where the message representative
+``mu`` is computed separately from signing and verification (see
+:meth:`MLDSA65PrivateKey.sign_mu` and :meth:`MLDSA65PublicKey.verify_mu`).
+:class:`MLDSAMuHasher` computes ``mu`` incrementally, which is useful when the
+message is large or only available as a stream.
+
+On backends that do not support computing ``mu``, :class:`MLDSAMuHasher`
+raises :class:`~cryptography.exceptions.UnsupportedAlgorithm`; ``sign_mu``
+and ``verify_mu`` still work with a ``mu`` computed elsewhere.
```
alex: Drop this paragraph.

docs/hazmat/primitives/asymmetric/mldsa.rst
```diff
@@ -701,5 +727,58 @@ Key interfaces
...
+.. class:: MLDSAMuHasher(public_key, context=None)
+
+    .. versionadded:: 50.0.0
+
+    Incrementally computes the ML-DSA ``mu`` (message representative) for the
+    "external mu" signing and verification APIs. For pure ML-DSA, FIPS 204
+    defines ``mu = SHAKE256(SHAKE256(pk, 64) || 0x00 || len(ctx) || ctx || M,
+    64)``, where ``M`` is the message. The portion derived from the public key
+    and context is computed when the hasher is created, and the message is
+    supplied incrementally with :meth:`update`.
```
alex: You do not need to repeat the formula for mu, the caller does not care.

src/rust/src/backend/mldsa.rs
```diff
@@ -723,6 +723,134 @@ impl MlDsa87PublicKey {
     }
 }
 
+/// Extract the raw public key bytes from any ML-DSA public key object. Private
+/// keys (and other key types) are rejected: mu computation is a public-key
+/// operation.
```
alex: Everything after the first sentence is slop and should be remoed

tests/hazmat/primitives/test_mldsa.py
```diff
@@ -127,6 +133,9 @@ def test_mldsa_unsupported(backend):
     ):
         MLDSA87PrivateKey.generate()
 
+    # Without ML-DSA support the mu hasher isn't available at all.
+    assert not hasattr(mldsa, "MLDSAMuHasher")
```
alex: pointless assert, delete

alex (approved):

PR #15144, opened 2026-07-03
Add a Cobblestone recipe implementing c2sp.org/chunked-encryption

### LINKED ISSUE #2358: Implement (or document) streaming API

alex:
Resolves #2358

This adds `cryptography.cobblestone`, a top-level recipe (à la Fernet) for streaming authenticated encryption of large messages, implementing the [C2SP chunked-encryption specification](https://c2sp.org/chunked-encryption)'s two named instantiations (per [C2SP/C2SP#296](https://github.com/C2SP/C2SP/pull/296)): **Cobblestone-128** (SHA-512 + AES-128-GCM, recommended) and **Cobblestone-256** (SHA-512 + AES-256-GCM, for environments that mandate 256-bit keys).

```python
from cryptography.cobblestone import (
    Cobblestone128Decryptor, Cobblestone128Encryptor
)

key = Cobblestone128Encryptor.generate_key()
enc = Cobblestone128Encryptor(key, context=b"myapp backup encryption")
ciphertext = enc.update(b"a secret message") + enc.finalize()
dec = Cobblestone128Decryptor(key, context=b"myapp backup encryption")
assert dec.update(ciphertext) + dec.finalize() == b"a secret message"
```

Design notes:

* The core is implemented in Rust, on top of the existing `AesGcm` AEAD and `HkdfExpand` implementations. The four classes are simple compositions over internal `ChunkedEncryptor`/`ChunkedDecryptor` types parameterized by the AEAD (IANA name, key/tag lengths). For each message a fresh AEAD key, base nonce, and 32-byte key commitment are derived with HKDF-Expand-SHA-512 from the input key, a random 24-byte salt, and a caller-provided context; the message is encrypted in 16 KiB chunks, with the chunk counter XOR'd into the base nonce. The ciphertext is `salt || commitment || chunks`.
* Buffering is minimal: full 16 KiB chunks are encrypted/decrypted directly from the caller's input into the output, so only sub-chunk remainders are copied into the internal buffer. `update_into()` variants let callers supply their own output buffers.
* The decrypter checks the commitment before decrypting anything, releases plaintext only for authenticated chunks, and enforces the short-final-chunk truncation rule at `finalize()`.
* Auth failures raise `cryptography.exceptions.InvalidTag`; any further use of a finalized or failed context raises `AlreadyFinalized`.

Tests cover round-trip, streaming-granularity, tampering, reordering, truncation, extension, and API misuse cases for both variants, plus a Rust unit test for the chunk counter limit (unreachable from Python). Note: the vendored FiloSottile/chunked test vectors (#15145) predate C2SP/C2SP#296 — they were generated with HKDF-SHA-256 — so the vector tests are removed until the reference implementation regenerates them for Cobblestone; they should be re-added (and the vendored file refreshed) once that happens.

https://claude.ai/code/session_01Uyk58oD6F8BJwKHE4qCMA8


docs/chunked-encryption.rst
```diff
@@ -0,0 +1,199 @@
...
+    >>> ciphertext = encrypter.update(b"a secret message")
+    >>> ciphertext += encrypter.finalize()
+    >>> decrypter = Decrypter(key, context=b"example-app file encryption")
+    >>> decrypter.update(ciphertext) + decrypter.finalize()
+    b'a secret message'
+
+The ciphertext is 56 bytes (a salt and a key commitment), plus the
+message itself, plus 16 bytes for each started 16 KiB chunk of the
+message — an overhead of roughly 0.1%.
+
```
alex: Drop this paragraph

src/rust/src/cobblestone.rs
```diff
@@ -0,0 +1,613 @@
...
+        // leave a partially written output behind.
+        let n_chunks = ((self.buffer.len() + data.len()) / CHUNK_SIZE) as u64;
+        self.cipher.nonces.check_capacity(n_chunks)?;
+
+        let wire_chunk = CHUNK_SIZE + self.cipher.tag_len;
+        let mut written = 0;
+        if self.header_pending {
+            out[..HEADER_LEN].copy_from_slice(&self.header);
+            written += HEADER_LEN;
+            self.header_pending = false;
```
alex, line 274: What happens if you call `update_into()` with a buffer insufficiently large for the whole header?

alex: It raises `ValueError` before any state is modified: `update_into()` computes the exact output size for the call up front (`update_out_len()` includes the full 56-byte header while it's pending) and checks the buffer against that before calling `update_impl()`, so the header is never partially written and the context remains usable. `test_update_into_buffer_too_small` covers this (a `HEADER_LEN - 1` buffer raises, then the same context is used successfully). I've added a comment on `update_impl` documenting the invariant.

---

docs/chunked-encryption.rst
```diff
@@ -0,0 +1,220 @@
...
+instantiations: **Cobblestone-128** (SHA-512 and AES-128-GCM, the
+recommended choice) and **Cobblestone-256** (SHA-512 and AES-256-GCM,
+for environments that mandate 256-bit keys).
+
+A message is encrypted in 16 KiB chunks, each of which is individually
+authenticated, so decryption can also be performed as a stream:
+decrypted data is returned incrementally, and only authenticated
+plaintext is ever returned. Reordering, truncating, or extending the
+ciphertext is detected. The scheme is also key committing: a ciphertext
+can only be decrypted with the key it was encrypted with.
```
alex: None of this information is useful, delete this paragraph.

docs/chunked-encryption.rst
```diff
@@ -0,0 +1,220 @@
...
+    Encrypts a single message under ``key`` with Cobblestone-128. Each
+    instance must be used for exactly one message: call :meth:`update`
+    (or :meth:`update_into`) any number of times, then call
+    :meth:`finalize` exactly once. The concatenation of the returned
+    bytes is the ciphertext.
+
+    :param key: A 16-byte key. This **must** be kept secret, and
+        **must** be uniformly random (e.g. the output of
+        :meth:`generate_key`, :func:`os.urandom`, or a key derivation
+        function — never a password). A single key may be used to
```
alex: you don't need to mention os.urandom, `e.g. the output of generate key -- never a password` is enough

docs/chunked-encryption.rst
```diff
@@ -0,0 +1,220 @@
...
+        ``buf``, and returns the number of bytes written. This avoids
+        allocating a new buffer for each call.
+
+        :param data: The data to encrypt.
+        :type data: :term:`bytes-like`
+        :param buf: A writable buffer to write the ciphertext into. A
+            buffer of ``len(data) + len(data) // 1024 + 16456`` bytes
+            is always large enough; the exact number of bytes required
+            for a given call is included in the :class:`ValueError`
+            raised if the buffer is too small.
```
alex: Delete everything after hte semicolon

docs/chunked-encryption.rst
```diff
@@ -0,0 +1,220 @@
...
+        :type buf: :term:`bytes-like`
+        :return int: The number of bytes written to ``buf``.
+        :raises ValueError: If ``buf`` is too small.
+
+    .. method:: finalize()
+
+        Encrypts the final chunk and returns the last portion of the
+        ciphertext. This must always be called, even if ``update``
+        returned all but the last few bytes of ciphertext, and the
```
alex: delete everything between the first and second commas, it makes no sense

src/cryptography/hazmat/bindings/_rust/chunked_encryption.pyi
```diff
@@ -0,0 +1,37 @@
...
+    @staticmethod
+    def generate_key() -> bytes: ...
+    def update(self, data: Buffer) -> bytes: ...
+    def update_into(self, data: Buffer, buf: Buffer) -> int: ...
+    def finalize(self) -> bytes: ...
+
+class Cobblestone128Decryptor:
+    def __init__(self, key: Buffer, context: Buffer) -> None: ...
+    @staticmethod
+    def generate_key() -> bytes: ...
```
alex: Decryptors shouldn't have a generate key method, only the encryptor side

src/cryptography/cobblestone.py
```diff
@@ -0,0 +1,21 @@
+# This file is dual licensed under the terms of the Apache License, Version
```
alex, line 1: @reaperhulk should this module be named cobblestone? I think probably?

src/rust/src/cobblestone.rs
```diff
@@ -0,0 +1,801 @@
...
+pub(crate) mod chunked_encryption_mod {
+    #[pymodule_export]
+    use super::{
+        Cobblestone128Decryptor, Cobblestone128Encryptor, Cobblestone256Decryptor,
+        Cobblestone256Encryptor,
+    };
+}
+
+#[cfg(test)]
+mod tests {
```
alex, line 743: in general we should only have rust tests for things that are unreachable from python in normal operations, but needed for coverage. do these meet that?

alex: Partially: the counter-limit test does meet the bar (hitting the 2³⁸-chunk limit requires a 4 PiB message, and it's what covers the `check_capacity`/`message_too_large_error` error lines), but the nonce XOR test didn't — that path is fully exercised by the Python round-trip tests. Dropped the XOR test and kept only the limit test, with a comment explaining why it exists.

---

tests/test_chunked_encryption.py
```diff
@@ -358,7 +358,7 @@ def test_generate_key(self, variant):
         key = encryptor_cls.generate_key()
         assert isinstance(key, bytes)
         assert len(key) == key_len
-        assert len(decryptor_cls.generate_key()) == key_len
+        assert not hasattr(decryptor_cls, "generate_key")
```
alex: this is a very stupid assert, delete it.

docs/chunked-encryption.rst
```diff
@@ -0,0 +1,201 @@
...
+
+.. class:: Cobblestone256Decryptor(key, context)
+
+    .. versionadded:: 50.0.0
+
+    Exactly like :class:`Cobblestone128Decryptor`, but decrypts
+    messages produced by :class:`Cobblestone256Encryptor` with a
+    32-byte key.
+
+Implementation
```
reaperhulk: This whole section doesn’t feel useful since we already state that it’s an implementation of c2sp chunked encryption at the top

src/cryptography/cobblestone.py
reaperhulk: Yeah I think we should rename it all to just cobblestone.

alex:
At this point I think the only other thing we want is the wycheproof tests to land so we can use them.

alex:
The consuming test is written and validated: I extracted the vectors from C2SP/wycheproof#265 as it stands and ran them against this branch — all 70 pass (35 per variant: valid decryptions checked against `msgLength`/`msgSha512`, invalid ones rejected, `InvalidKeySize` rejected at construction, and failed contexts keep failing). Once #265 merges I'll push `tests/wycheproof/test_cobblestone.py` along with a `WYCHEPROOF_REF` bump so CI picks the files up.

---

reaperhulk:
The wycheproof pr is now merged

reaperhulk (approved):

PR #15146, opened 2026-07-03
Build without PEP 517 isolation in nox sessions to keep cargo caches warm

reaperhulk:
PEP 517 build isolation creates a temporary virtualenv with a randomized path for every build, so `PYO3_PYTHON` changes on each install. `cryptography-cffi`'s build script registers `rerun-if-env-changed=PYO3_PYTHON`, so `cryptography-cffi` and `cryptography-rust` recompile on every CI job even with a fully warm rust-cache — and on free-threaded (non-abi3) builds the pyo3 crates recompile too, since the maturin-generated interpreter config also embeds the ephemeral path. See https://github.com/PyO3/pyo3/issues/6113 (the upstream `PYO3_BASE_PYTHON` fix is deferred to pyo3 0.30).

This installs the `[build-system].requires` into the session venv (constrained by `ci-constraints-requirements.txt`, same as the `flake` session already does) and passes `--no-build-isolation`, so the build backend runs from the stable `.nox/<session>` interpreter and cargo fingerprints stay valid across runs.

Verified locally: with isolation, every reinstall recompiles the two crates (~30s of cargo on 4 cores); without it, a warm-cache reinstall with identical env compiles nothing (cargo finishes in 0.05s).

Note: this PR also bumps the rust-cache key. PR runs restore main's caches (whose fingerprints contain the old ephemeral paths) and rust-cache skips saving on an exact key hit, so without a key roll the stable-fingerprint cache would never get saved and the win would only appear once caches roll naturally. CI timing comparisons will be posted on this PR.

https://claude.ai/code/session_01Lj3BsMN8xmPQQ89d6nzDcZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lj3BsMN8xmPQQ89d6nzDcZ)_

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -19,4 +19,4 @@ runs:
         KEY: "${{ inputs.key }}"
     - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
       with:
-        key: ${{ steps.normalized-key.outputs.key }}-5
+        key: ${{ steps.normalized-key.outputs.key }}-6
--- a/noxfile.py
+++ b/noxfile.py
@@ -44,6 +44,18 @@ def load_pyproject_toml() -> dict:
         return tomllib.load(f)
 
 
+def install_cryptography(session: nox.Session, *args: str) -> None:
+    # Install the build requirements into the session's environment and
+    # then build without PEP 517 build isolation. Build isolation uses a
+    # temporary virtualenv with a randomized path, so PYO3_PYTHON differs
+    # on every build, which defeats cargo's caching of anything whose
+    # build script depends on it.
+    # See https://github.com/PyO3/pyo3/issues/6113
+    pyproject_data = load_pyproject_toml()
+    install(session, *pyproject_data["build-system"]["requires"])
+    install(session, "--no-build-isolation", *args)
+
+
 @nox.session
 @nox.session(name="tests-ssh")
 @nox.session(name="tests-randomorder")
@@ -73,13 +85,13 @@ def tests(session: nox.Session) -> None:
     install_spec = f".[{','.join(extras)}]"
     install(session, "-e", "./vectors")
     if session.name == "tests-rust-debug":
-        install(
+        install_cryptography(
             session,
             "--config-settings-package=cryptography:build-args=--profile=dev",
             install_spec,
         )
     else:
-        install(session, install_spec)
+        install_cryptography(session, install_spec)
 
     install(
         session,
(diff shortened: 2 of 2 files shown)


reaperhulk:
## CI timing results (4 measured runs on this PR)

| Run | Condition | Wall clock |
|---|---|---|
| main [28675190534](https://github.com/pyca/cryptography/actions/runs/28675190534) / [28659707361](https://github.com/pyca/cryptography/actions/runs/28659707361) | baseline (isolation, warm cache) | 5m47s / 7m16s |
| A [28678670499](https://github.com/pyca/cryptography/actions/runs/28678670499) | fix, stale `-5` caches from main | 5m03s |
| B [28679077977](https://github.com/pyca/cryptography/actions/runs/28679077977) | fix, cold caches (`-6` key roll) | 5m50s |
| C [28679399945](https://github.com/pyca/cryptography/actions/runs/28679399945) | fix, warm stable caches | 4m58s |
| D [28679807220](https://github.com/pyca/cryptography/actions/runs/28679807220) | fix, warm stable caches | 6m07s |

**What the logs prove:** the fix works as intended. In the warm runs (C/D), the pip-install build (`nox --install-only`) compiles **only the 8 workspace crates** — the pyo3 dependency chain is fully cache-stable on every platform, including the free-threaded jobs where it previously rebuilt every run. Example: windows 3.14t `nox --install-only` dropped from ~112s (baseline, compiles pyo3 + workspace) to ~80s (workspace only). A pleasant side-effect measured in run B: a full cache roll now only costs ~1 minute of wall clock, since the whole dependency tree builds in ~50s on the Linux runners.

**What it doesn't show:** total wall clock and per-job composites are within run-to-run noise (windows-tests composite: baseline 197–261s vs warm runs 168–249s; linux `Create nox environment` median ~40s in all runs). Two reasons, both now understood:

1. **rust-cache doesn't cache workspace crates** (default `cache-workspace-crates: false`), and those are the dominant remaining compile cost (~40–50s/job). This PR makes their fingerprints stable, which is the prerequisite — but enabling workspace-crate caching also requires solving cargo's mtime-based freshness on fresh checkouts (naively pinning mtimes to a fixed date would make cargo *miss real source changes*; it needs `git restore-mtime`-style per-file commit times, or sccache). Deliberately left as follow-up.
2. The `rust` session's `cargo test` build has its own pyo3/pyo3-ffi rebuild each run (no `PYO3_PYTHON` involved — pre-existing and unrelated to build isolation). Also follow-up.

So: correct fix, verified mechanism, ~30s of real redundant compile removed per job, but the headline wall-clock win is gated on the workspace-cache follow-up this enables.

https://claude.ai/code/session_01Lj3BsMN8xmPQQ89d6nzDcZ

---

PR #15149, opened 2026-07-04
reject fractional seconds in crl invalidity date

dxbjavid:
The InvalidityDate CRL entry extension parses its GeneralizedTime with the lenient asn1::GeneralizedTime, so a CRL whose invalidity date carries fractional seconds (or another non-DER form) is accepted and the fraction is silently dropped on the way to the Python datetime. Every other X.509 time field, and this extension's own encoder, already use the strict asn1::X509GeneralizedTime, which forbids fractional seconds as required by RFC 5280 and X.690 11.7. This moves the decode over to that strict type so the malformed value is rejected rather than quietly truncated.


alex (dismissed):

alex:
Please document the vector in test-vectors.rst

dxbjavid:
done, added an entry for it under the CRL vectors in test-vectors.rst.

alex (approved):

dxbjavid:
TY!

PR #15156, opened 2026-07-04
Add ML-KEM validation tests

Ap4sh:
## summary
- add coverage for ML-KEM `from_public_bytes()` rejecting invalid raw public key lengths and non-buffer inputs
- add coverage for ML-KEM `decapsulate()` rejecting invalid ciphertext lengths and non-buffer inputs

This just locks in the existing validation behavior around raw key and ciphertext inputs

## tests
- `PYTEST_XDIST_AUTO_NUM_WORKERS=2 python3 -m nox -e local -- tests/hazmat/primitives/test_mlkem.py`


reaperhulk:
This is not a useful PR. We already test `decapsulate` and `from_public_bytes` with incorrect lengths, and the TypeError tests re-exercise our standard CffiBuf branching.

PR #15171, opened 2026-07-05
Deprecate Diffie-Hellman over finite fields (FFDH)

alex:
Deprecates everything FFDH, per the request:

- Every public name in `cryptography.hazmat.primitives.asymmetric.dh` (the key/parameters classes, the numbers classes, the `WithSerialization` aliases, and `generate_parameters`) is deprecated via `utils.deprecated` with a new `DeprecatedIn50`, so any attribute access warns.
- Loading a DH key via `load_{pem,der}_{private,public}_key` (including PKCS#12 and certificate/CSR `public_key()` for DH SPKIs) emits the same warning from the Rust side, after the key passes validation, so failed loads raise their normal errors without a spurious warning. Non-DH key loads are unaffected.
- `load_pem_parameters`/`load_der_parameters` can only load FFDH parameters, so the functions themselves are deprecated.
- `types.py` uses private non-warning aliases for the DH entries in `PublicKeyTypes`/`PrivateKeyTypes` so that `import cryptography.x509` stays warning-free.
- `test_dh.py` ignores the warning module-wide via `pytestmark`, with `pytest.warns` at the parameter-loader callsites; the handful of DH touchpoints in `test_serialization.py`/`test_x509.py` use `pytest.warns`. The suite runs warning-free.
- CHANGELOG entry and `.. deprecated:: 50.0.0` notes in the docs.

https://claude.ai/code/session_017iUUcywCCXL51sy4WUGqbM

---
_Generated by [Claude Code](https://claude.ai/code/session_017iUUcywCCXL51sy4WUGqbM)_

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,20 +8,11 @@ Changelog
 
 .. note:: This version is not yet released and is under active development.
 
-* Deprecated Diffie-Hellman key exchange over finite fields (FFDH). The
-  classes and functions in ``cryptography.hazmat.primitives.asymmetric.dh``
-  now emit a ``CryptographyDeprecationWarning`` when used, as does loading a
-  DH key or DH parameters with
-  :func:`~cryptography.hazmat.primitives.serialization.load_pem_private_key`,
-  :func:`~cryptography.hazmat.primitives.serialization.load_der_private_key`,
-  :func:`~cryptography.hazmat.primitives.serialization.load_pem_public_key`,
-  :func:`~cryptography.hazmat.primitives.serialization.load_der_public_key`,
-  :func:`~cryptography.hazmat.primitives.serialization.load_pem_parameters`,
-  or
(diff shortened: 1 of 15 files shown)


alex:
@gpotter2 We plan to deprecate FFDH (though we do not have a timeline for removal and it's likely to be a while), it looks like the act of emitting the warning causes the scapy tests to fail. Are you able to take a look at that?

gpotter2:
Hi ! Sure I can take a look. As long as it remains in the decrepit folder it shouldn't be much of an issue on our side.

> though we do not have a timeline for removal

By removal do you mean entirely or just the move to decrepit ?

Have a good sunday

alex:
We weren't planning to migrate to decreptic, just deprecate (for a long time) and then remove.

Thanks so much for taking a look at this!

gpotter2:
This should be fixed on master ! Sorry for the delay

reaperhulk (dismissed):

reaperhulk (approved):

PR #15172, opened 2026-07-05
Skip slow DH_check tests on AWS-LC and LibreSSL for large moduli

reaperhulk:
## Summary
Add a test skip condition for DH parameter validation tests when using AWS-LC or LibreSSL with large moduli (≥3072 bits), as DH_check is known to be very slow on these backends for large key sizes.

## Changes
- Import `rust_openssl` bindings to detect AWS-LC and LibreSSL backends
- Add conditional skip in `test_dh_parameters_allows_rfc3526_groups` for large moduli on AWS-LC and LibreSSL
- Reference AWS-LC issue https://github.com/aws/aws-lc/issues/3335 tracking the performance problem

## Details
The DH_check operation is significantly slower on AWS-LC and LibreSSL when validating large Diffie-Hellman parameters (3072 bits and above). This change prevents test timeouts by skipping these specific test cases on affected backends while maintaining full test coverage on other OpenSSL implementations.

https://claude.ai/code/session_01PL9rchfs5nzFEamLrJnVTZ

### DIFF
--- a/tests/hazmat/primitives/test_dh.py
+++ b/tests/hazmat/primitives/test_dh.py
@@ -11,6 +11,7 @@ import typing
 
 import pytest
 
+from cryptography.hazmat.bindings._rust import openssl as rust_openssl
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import dh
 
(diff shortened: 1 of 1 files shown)


alex (approved):

tests/hazmat/primitives/test_dh.py
```diff
@@ -176,6 +177,16 @@ def test_dh_parameters_allows_rfc3526_groups(self, backend, vector):
         p = int.from_bytes(binascii.unhexlify(vector["p"]), "big")
         if backend._fips_enabled and p < backend._fips_dh_min_modulus:
             pytest.skip("modulus too small for FIPS mode")
+        # DH_check is very slow for large moduli on AWS-LC and LibreSSL.
+        # For AWS-LC this is tracked in
+        # https://github.com/aws/aws-lc/issues/3335
+        if (
+            rust_openssl.CRYPTOGRAPHY_IS_AWSLC
+            or rust_openssl.CRYPTOGRAPHY_IS_LIBRESSL
```
botovq, line 185: Why is this still needed?

To be clear: I have added special casing for the RFC 2409, RFC 3526, RFC 7919 primes in https://github.com/openbsd/src/commit/9ac14cf7a1b28db2437e849540c0abe6a436a03d and follow-ups, so this should not be slow on LibreSSL 4.3.0 and later.

alex: We still test against 4.2.1 (since it's still in an OpenBSD that's supported), and I believe that doesn't have this fix yet -- we should be able to remove it after the next OpenBSD release.

botovq: That's right. Fix is in 4.3.0 and later but was not backported to the 4.2 branch. Thanks.

PR #15174, opened 2026-07-05
Raise MSRV to 1.85

alex:
Raises the minimum supported Rust version from 1.83.0 to 1.85.0.

- `Cargo.toml`: bump the workspace `rust-version` to 1.85.0.
- `.github/workflows/ci.yml`: pin the three MSRV matrix builds (plain, AWS-LC, BoringSSL) to 1.85.0 and drop the now-satisfied "1.85: 2024 edition" line from the future-MSRV comment.
- `docs/installation.rst`: document the new minimum version and update the distro notes — RHEL/CentOS now requires 9.7+ (9.6 ships Rust 1.84.1) and Alpine now requires 3.22+ (3.21 ships Rust 1.83). The Debian note is unchanged since trixie ships rustc 1.85.0.
- `CHANGELOG.rst`: note the MSRV change in the unreleased 51.0.0 section.

## Coverage

This PR deliberately contains **no coverage workaround**. Starting with Rust 1.84, functions consisting entirely of proc-macro expanded code (`#[derive(...)]`, `#[pyo3::pyclass]`, `#[pyo3::pymethods]`) get defective coverage mappings: the residual attribute-line region is wired to a non-entry counter that never fires, so the line reports 0 even when the function runs. The 1.83 MSRV jobs were the last toolchain producing correct (entry-wired) mappings, so bumping them to 1.85 makes the combined-coverage gate fail with ~124 missed statements, all on attribute lines.

A compiler fix is up at rust-lang/rust#158833. Until it reaches the toolchains CI collects coverage from, the coverage job here is expected to fail — the red gate doubles as a canary: when it goes green vanilla, the upstream fix has landed in the relevant toolchains and this can merge as-is.

https://claude.ai/code/session_01X4EZ7B8sU7cLxPP4qnHfAm


alex:
Blocked on https://github.com/rust-lang/rust/pull/158833

PR #15213, opened 2026-07-06
Fetch test vectors via codeload tarballs instead of git clones

alex:
Tests whether codeload snapshot tarballs are faster than the two `actions/checkout` clones for fetching wycheproof and x509-limbo. Baseline from main's runs on 2026-07-06: the "Clone test vectors" step costs ~11–18s per job.

- The vectors are fetched as `https://codeload.github.com/<repo>/tar.gz/<sha>` downloads, extracted in parallel. This skips git ref negotiation, server-side pack generation, client-side object hashing, and writing ~40MB of `.git` directories. `curl --retry` handles transient failures, and a failed download fails the step.
- The pinned refs move from checkout `ref:` values to env vars on the download step. The `bump_dependency.py` patterns in `x509-limbo-version-bump.yml` are updated to match; verified by running the bump script against the new file (it matches the current SHAs).
- One thing this run answers: whether the container jobs' images (alpine, distros) all have `curl` — the checkout action brought its own node runtime, `curl`/`tar` come from the image.

To evaluate: compare the "Clone test vectors" step duration against main's ~11–18s. Note codeload only serves gzip (no zstd option — the `.tar.gz` is the payload, not a transport encoding, so `Accept-Encoding` can't upgrade it).

### DIFF
--- a/.github/actions/fetch-vectors/action.yml
+++ b/.github/actions/fetch-vectors/action.yml
@@ -1,22 +1,34 @@
-name: Clone test vectors
-description: Clones the wycheproof and x509-limbo repositories
+name: Fetch test vectors
+description: Fetches the wycheproof and x509-limbo repositories
 
 runs:
   using: "composite"
 
   steps:
-    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
-      with:
-        repository: "C2SP/wycheproof"
-        path: "wycheproof"
+    # Codeload snapshots skip the git protocol entirely: no ref
+    # negotiation, no server-side pack generation, no client-side object
+    # hashing, and no .git directory to write. The two downloads run in
+    # parallel.
+    - name: Download test vectors
+      shell: bash
+      run: |
+        download() {
+          curl --fail --silent --show-error \
+            --retry 4 --retry-all-errors --retry-delay 2 \
+            --output "$2.tar.gz" \
+            "https://codeload.github.com/$1/tar.gz/$3"
+          mkdir "$2"
+          tar -xzf "$2.tar.gz" --strip-components=1 -C "$2"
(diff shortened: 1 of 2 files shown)


alex:
Results are in. "Clone test vectors" step durations, comparing a main run from earlier today ([28816429921](https://github.com/pyca/cryptography/actions/runs/28816429921), two serial `actions/checkout` clones) against this PR's run ([28827263185](https://github.com/pyca/cryptography/actions/runs/28827263185), two parallel codeload tarball downloads):

| job family | main: git clones | this PR: tarballs | savings/job |
|---|---|---|---|
| macos (n=5) | 9–13s, avg **11.0s** | 1–2s, avg **1.4s** | ~10s |
| linux, ubuntu-latest (n=20) | 3–10s, avg **4.3s** | 1–2s, avg **1.2s** | ~3s |
| distros, containers (n=13) | 3–8s, avg **3.9s** | 0–5s, avg **1.7s** | ~2s |
| alpine, container (n=2) | 4–5s, avg **4.5s** | **2s** | ~2.5s |

Notes:

- Every container image turned out to have `curl`, so the tarball path works everywhere and no checkout fallback is needed.
- Windows step timings aren't visible through the jobs API (they're inside the `windows-tests` composite), but the jobs pass.
- The macOS win is the most valuable one: those 5 jobs sit in the 5-concurrent-runner pool, so ~10s per job comes straight off the queue backlog during merge bursts.
- Summed over a full CI run this is roughly 2 minutes of runner time, all from a step that produces identical bytes.

The delta is bigger than raw transfer would suggest because the clone path pays for ref negotiation, server-side pack generation, client-side object hashing, ~40MB of `.git` writes, and ran the two clones serially — the tarball path is just two concurrent HTTP streams into `tar -xz`.

---

reaperhulk (approved):

PR #15214, opened 2026-07-06
Declare rerun-if directives in build scripts that had none

alex:
A build script that emits no `rerun-if` directives makes cargo fall back to fingerprinting the entire package by its newest file mtime, so the script reruns — and the crate and its dependents recompile — whenever any file in the package is touched, even with no content change (visible with `CARGO_LOG=cargo::core::compiler::fingerprint=info` as `PrecalculatedComponentsChanged`). The three build scripts here (workspace root, cryptography-openssl, cryptography-key-parsing) depend only on themselves, `DEP_OPENSSL_*` metadata from openssl-sys (tracked by cargo as a unit dependency on its own), and in the root's case one environment variable — so `rerun-if-changed=build.rs` is the correct scope.

Two effects:

- The root `build.rs` now declares `rerun-if-env-changed=CRYPTOGRAPHY_BUILD_OPENSSL_NO_LEGACY`, which it reads but which cargo previously didn't track: changing it between builds didn't trigger a rerun unless something else happened to dirty the crate.
- Whole-package mtime fingerprinting is what makes these crates rebuild spuriously when file mtimes move without content changes — it's also a hard blocker for caching workspace-crate artifacts in CI (an equality check that no mtime restoration can satisfy), which is a follow-up I'd like to send.

`nox -e rust` (fmt, clippy, cargo test) passes.

---
_Generated by [Claude Code](https://claude.ai/code/session_013MTT3EVJrxAz3oJZhh6rhE)_

### DIFF
--- a/src/rust/build.rs
+++ b/src/rust/build.rs
@@ -6,6 +6,14 @@ use std::env;
 
 #[allow(clippy::unusual_byte_groupings)]
 fn main() {
+    // Without any rerun-if directives cargo reruns the build script (and
+    // recompiles the crate) whenever any mtime in the package changes,
+    // which defeats CI build caching. Everything below depends only on
+    // this file, the environment variables it reads, and metadata from
+    // openssl-sys (which cargo tracks as a dependency on its own).
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-env-changed=CRYPTOGRAPHY_BUILD_OPENSSL_NO_LEGACY");
+
     pyo3_build_config::use_pyo3_cfgs();
 
     if let Ok(version) = env::var("DEP_OPENSSL_VERSION_NUMBER") {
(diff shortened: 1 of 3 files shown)


reaperhulk:
Very curious if the workspace caching is going to work. If it does that’s going to save a lot of time…

reaperhulk (approved):

alex:
Local simulation says yes: with this PR plus #15215, a simulated warm-cache CI run (fresh-checkout mtimes everywhere, then pinning inputs that are unchanged since the cached build's commit to a fixed old mtime) went from recompiling all 8 workspace crates in ~27s to recompiling **nothing in 0.02s**; a commit touching one crate rebuilt only that crate plus the cdylib.

These two PRs are the prerequisites — without rerun-if directives, cargo fingerprints those packages by newest-mtime *equality*, which no mtime restoration can ever satisfy, and the `__pycache__` write re-dirtied `_cffi_src/` on every build.

The CI wiring (rust-cache `cache-workspace-crates`, the built commit recorded in the same cache entry so they can't desync, and a `git diff`-driven mtime restore) is written and validated locally; I'll send it as a follow-up once this lands. Failure modes degrade to a full rebuild, never stale artifacts — cargo's staleness check is strictly "source newer than fingerprint" (verified against cargo 1.83's `find_stale_file`), so pinned-old inputs can only ever be over-rebuilt.

---

PR #15216, opened 2026-07-06
Reuse cached Rust workspace builds in macOS CI

alex:
Reuse the workspace crates' compiled artifacts across macOS CI runs. Today every macOS job recompiles all 8 workspace crates (~40–60s of a ~190s job) despite a warm rust-cache; with macOS capped at 5 concurrent hosted runners, per-job seconds translate directly into queue time during merge bursts (~49 min queues on 2026-07-06).

**Measured on this PR's CI** (cold run, then a re-run restoring its caches): macOS "Build nox environment" went from **~62s cold to 3–4s warm with zero crates recompiled** (cargo: `Finished in 0.06s`) on the stable-Python jobs — full data in [the results comment](https://github.com/pyca/cryptography/pull/15216#issuecomment-4899665332). Prerequisites #15214, #15215, #15217, #15218 have all merged; this is now a single commit containing only the caching machinery.

**What's cached** (one rust-cache entry per macOS matrix job, in a `-ws` key namespace so other jobs' caches aren't touched):

- the workspace crates' artifacts and `.fingerprint` dirs (`cache-workspace-crates: true`; rust-cache normally strips these)
- `.rust-build-meta/anchor-commit` — the SHA the artifacts were built from, written at the end of the build step and stored in the *same* cache entry via `cache-directories`, so the two cannot desync
- the workspace cdylib, stashed next to the anchor and restored before the build — rust-cache's cleanup drops it even with `cache-workspace-crates`, which would otherwise force a rebuild+relink of the final crate every run (candidate for an upstream rust-cache issue)

**How it invalidates:**

- *Whole entry*: rust-cache's key rotates on any `Cargo.lock`/`Cargo.toml`/rustc change (prefix-restore keeps deps warm; the save refreshes the anchor).
- *Per crate*: cargo's own freshness rules, made meaningful by `.github/bin/prepare_rust_cache_mtimes.py`. Fresh checkouts stamp every file "now", so cargo would consider all cached fingerprints stale. On restore, the script pins inputs that are **unchanged since the anchor commit** (per `git diff --name-only <anchor> HEAD` — content-exact, tree-to-tree) to a fixed old mtime; changed paths keep fresh mtimes and cargo rebuilds exactly the crates that depend on them. Directories under the source roots are pinned too (cargo stats dir mtimes to detect adds/deletes in `rerun-if-changed` dirs), except directories containing a change. Requires `fetch-depth: 0` + `filter: blob:none` on the macOS checkout (~16MB of commit/tree metadata).

**Failure direction is always rebuild, never stale**: cargo's stale check is strictly "input newer than fingerprint" (`find_stale_file`, verified against cargo 1.83), so pinned-old files are only fresh when their content matches the anchor's tree byte-for-byte. A missing or unfetchable anchor (first run, evicted cache, a cache saved from an ephemeral PR merge commit) degrades to today's behavior: a full rebuild.

Note on post-merge behavior: caches saved by this PR's runs live in the PR's cache scope with ephemeral-merge-commit anchors, so the first runs on main after merging do one cold build per job (identical to today) and save main-scoped caches with resolvable anchors; every run after that gets the warm path.

### DIFF
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -60,6 +60,10 @@ runs:
     # the prior behavior.
     - name: Restore Rust build freshness
       run: |
+        # Container jobs run steps with a different HOME than the
+        # checkout action, so its safe.directory registration isn't
+        # visible here and git would refuse to read the workspace.
+        git config --global --add safe.directory "$(pwd)" 2>/dev/null || true
         if command -v python3 >/dev/null 2>&1; then PY=python3; else PY=python; fi
         "$PY" .github/bin/prepare_rust_cache_mtimes.py .rust-build-meta/anchor-commit
         if [ -d .rust-build-meta/artifacts/deps ]; then
--- a/.github/actions/cache/action.yml
+++ b/.github/actions/cache/action.yml
@@ -5,15 +5,6 @@ inputs:
     description: 'extra cache key components'
     required: false
     default: ''
-  cache-workspace-crates:
-    description: |
-      Also cache the workspace crates' own artifacts, not just
-      dependencies. Only useful for jobs that restore deterministic
-      source mtimes first (.github/bin/prepare_rust_cache_mtimes.py) and
-      record the built commit in .rust-build-meta/anchor-commit, which
-      is cached alongside the target directory below.
-    required: false
-    default: 'false'
 
 
 runs:
@@ -26,13 +17,16 @@ runs:
       shell: bash
       env:
         KEY: "${{ inputs.key }}"
-    # Workspace-crate caches live in their own key namespace: their target
-    # directories contain artifacts (and the anchor file) that plain jobs
-    # neither produce nor expect.
+    # cache-workspace-crates keeps the workspace crates' own artifacts
+    # and fingerprints, not just dependencies'. .rust-build-meta records
+    # which commit they were built from (and stashes the cdylib, which
+    # rust-cache's cleanup drops); it is cached in the same entry so the
+    # two cannot fall out of sync. Written by
+    # .github/bin/record_rust_cache_anchor.sh at the end of build steps.
     - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
       with:
-        key: ${{ steps.normalized-key.outputs.key }}${{ inputs.cache-workspace-crates == 'true' && '-ws' || '' }}-5
-        cache-workspace-crates: ${{ inputs.cache-workspace-crates }}
+        key: ${{ steps.normalized-key.outputs.key }}-6
+        cache-workspace-crates: true
         cache-directories: |
           ${{ github.workspace }}/.rust-build-meta
     # maturin writes the pyo3 config file to target/maturin/, and pyo3-ffi's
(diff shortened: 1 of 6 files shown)


alex:
Data collection complete — the warm path works end to end on real runners. Each cold run built and saved caches; each warm run is a re-run of the same head restoring them. "Build nox environment" durations for the stable-Python macOS jobs across the experiment:

| stage | 3.14 | 3.14t | 3.9 | what was left |
|---|---|---|---|---|
| cold (every head) | 62s | 53s | 59s | — |
| warm, initial | 33s | 33s | 32s | cffi + cdylib rebuilt |
| warm, + PYO3_PYTHON fix | 26s | 34s | 39s | cdylib relink only |
| warm, + cdylib preserved | **3s** | 51s* | **4s** | **nothing** — 0 `Compiling` lines, cargo `Finished in 0.06s` |

The mtime/anchor mechanism itself had zero findings in any run: every warm restore logged `pinned 142 files and 23 directories; 0 paths changed since <anchor>` with an exact cache-key hit, and every run's tests passed — no stale-artifact behavior anywhere.

The residual rebuilds were three distinct pre-existing bugs, each pinpointed by the (temporary) `CARGO_LOG=cargo::core::compiler::fingerprint=info` diagnostics:

1. **Stale maturin/pyo3 config for moving interpreter versions** → maturin rewrites the cached config every run, re-dirtying pyo3-ffi and everything above it. Split out as #15217. This is why the `-dev` jobs (and *3.14t in the last run — its patch version drifted past the cached config) still rebuild the pyo3 chain here; that column goes clean once #15217 lands.
2. **`rerun-if-env-changed=PYO3_PYTHON` vs uv's ephemeral build-env path** (`.cache/uv/builds-v0/.tmpXXXX/bin/python`, different every build) → cffi + cryptography-rust rebuilt unconditionally. Split out as #15218.
3. **rust-cache's cleanup drops the workspace cdylib even with `cache-workspace-crates`** (`FailedToReadMetadata { path: ".../deps/libcryptography_rust.dylib" }`) → rebuild+relink of the final crate every run. Worked around in this PR by stashing the cdylib in the cached metadata directory; this one's a candidate for an upstream rust-cache issue.

Housekeeping before this merges: the diagnostics commit (CARGO_LOG + the `-ws` namespace bumps used to force fresh anchors between experiment heads) gets dropped, and the PYO3_PYTHON commit gets rebased out now that it's #15218. Merge order that makes everything land clean: #15217, #15218, then this.

---

.github/actions/cache/action.yml
```diff
@@ -5,6 +5,15 @@ inputs:
     description: 'extra cache key components'
     required: false
     default: ''
+  cache-workspace-crates:
```
alex: I don't think this should be conditional, this should be on for all platforms.

alex:
Done in 62acb92 — the input is gone and workspace-crate caching is on unconditionally for every job. The cache action now owns the restore side (mtime pinning + cdylib put-back) so all callers get it automatically; the save side is a shared script (`record_rust_cache_anchor.sh`) called at the end of each build step across linux, distros, alpine, windows, macos, and linux-downstream, with full-history `blob:none` checkouts wherever Rust gets built. Jobs that never record an anchor (flake, docs, linkcheck) just rebuild as before. The key counter bump rotates all caches once since every entry changes shape.

One thing to keep an eye on post-merge: total cache storage grows by roughly 1.5–3GB compressed across ~40 keys against the 10GB repo budget — if eviction churn shows up (e.g. the OpenSSL build caches going cold), the fix is trimming which jobs keep workspace artifacts.

---

alex:
Spent a bunch of time staring at this. I do think we eventually want to figure out how to have caching of our local workspace crates, however the mtime hijinks required here are too hard to reason about, and I don't have enough confidence in correctness. We just need https://github.com/rust-lang/cargo/issues/14136 to stabilize.

alex:
Makes sense — checksum freshness eliminates the entire mtime layer, which was exactly the hard-to-trust part. For whenever cargo#14136 stabilizes, two findings from this PR remain relevant to a future version: rust-cache's cleanup drops the workspace cdylib even with `cache-workspace-crates` (the stash/put-back or an upstream fix will still be needed), and the fixes that made any of this possible (#15214, #15215, #15217, #15218) are merged and keep paying for themselves regardless.

---

kyle-basis:
@Alex @claude whenever cargo#14136 stabilizes, I have a working incremental cache for rust, that doesn't just tar the whole cargo directory into a large blob. LMK if you'd be interested.

kyle-basis:
I also went down this exact same mtime path and concluded it wasn't safe, especially around rerunning old builds, and if anything happened to pull in a dependency that wasn't in the git tree.

Issue #15241, opened 2026-07-09
Support for creating unsigned certificates with CertificateBuilder


magnuswatn:
[RFC 9925](https://datatracker.ietf.org/doc/html/rfc9925) specifies unsigned certificates, which are handy for trust anchors and other uses where one would traditionally use self signed certificates, but where the extra bytes used by the signature are unfortunate.

It would be nice if it was possible to create such certificates using the `x509.CertificateBuilder`, maybe with an `.create_unsiged()` method or something.

It is possible to create unsigned certs today using the asn1 module, so this is more a nice to have thing.

alex:
This seems like a reasonable thing, we'd have to figure out the API a bit, but, but prima facie we'd be happy to take a PR for this.

magnuswatn:
I went ahead and created a PR for this, just to have something to look at. The API now is just `.create_unsigned()`:

```python
    certificate = (
        x509.CertificateBuilder()
        .issuer_name(x509.Name([x509.NameAttribute(NameOID.UNSIGNED, "")]))
        .subject_name(x509.Name([x509.NameAttribute(NameOID.COMMON_NAME, common_name)]))
        .not_valid_before(now)
        .not_valid_after(now + timedelta(days=20 * 365))
        .serial_number(1)
        .public_key(private_key.public_key())
        .create_unsigned()
    )
```

I was unsure if it should validate that the issuer and/or extensions matches the RFC, but the builder is kinda lax on stuff like that today (since it cannot know the right issuer normally) so I opted not to.

magnuswatn:
Solved by #15242. Thanks!

PR #15242, opened 2026-07-10
Add support for creating unsigned certificates

magnuswatn:
First try for #15241.

I wasn't sure if there should be any validation for the subject or the extensions in the `create_unsigned()` method, but since the CertificateBuilder doesn't really do much validation for those today, I opted to skip it.

Also wasn't sure if the unsigned alg oid belongs in the `_SIG_OIDS_TO_HASH`.


reaperhulk (approved):

reaperhulk:
Thanks, this looks like a fine approach in general 😄

PR #15252, opened 2026-07-12
Migrate AES key wrap functions to Rust

reaperhulk:
This PR migrates the AES key wrap implementation from pure Python to Rust, leveraging OpenSSL's ECB cipher mode for improved performance.

## Summary

The AES key wrap functionality (`aes_key_wrap`, `aes_key_unwrap`, `aes_key_wrap_with_padding`, and `aes_key_unwrap_with_padding`) has been reimplemented in Rust using PyO3 bindings. The Python module now delegates to the Rust implementation while maintaining the same public API and exception behavior.

## Key Changes

- **New Rust module** (`src/rust/src/backend/keywrap.rs`): Implements RFC 3394 and RFC 5649 key wrap algorithms using OpenSSL's AES-ECB cipher context
  - `wrap_core()` and `unwrap_core()` helper functions for the core wrapping/unwrapping logic
  - Four public PyO3 functions matching the Python API
  - Proper validation of key sizes and wrapped key lengths
  - Constant-time comparison for integrity checks using `cryptography_crypto::constant_time`

- **Updated Python module** (`src/cryptography/hazmat/primitives/keywrap.py`): Simplified to delegate to Rust implementation
  - Removed ~150 lines of pure Python implementation
  - Kept `InvalidUnwrap` exception class for backward compatibility
  - Functions now call corresponding Rust implementations via `rust_openssl.keywrap`

- **Type stubs** (`src/cryptography/hazmat/bindings/_rust/openssl/keywrap.pyi`): Added for type checking support

- **Module registration**: Updated `src/rust/src/lib.rs` and `src/rust/src/backend/mod.rs` to export the new keywrap module

- **Exception handling**: Added `InvalidUnwrap` exception import in `src/rust/src/exceptions.rs`

- **Tests**: Added roundtrip tests for large keys (4096 bytes and 4099 bytes) to verify correctness with non-trivial data sizes

## Implementation Details

The Rust implementation follows the RFC specifications closely:
- Uses AES-ECB mode with padding disabled for direct block encryption/decryption
- Implements the index method for key wrap operations with 6 iterations
- Handles both standard (RFC 3394) and padded (RFC 5649) variants
- Validates input constraints (key length, wrapped key length, padding requirements)
- Uses constant-time comparison for authentication tag verification to prevent timing attacks

https://claude.ai/code/session_014ki2rU5KMuRkxtGEpkQc2w


src/rust/src/backend/keywrap.rs
```diff
@@ -0,0 +1,274 @@
...
+
+use crate::buf::CffiBuf;
+use crate::error::{CryptographyError, CryptographyResult};
+use crate::exceptions;
+
+fn ecb_ctx(
+    wrapping_key: &[u8],
+    side: openssl::symm::Mode,
+) -> CryptographyResult<openssl::cipher_ctx::CipherCtx> {
+    let cipher = match wrapping_key.len() {
```
alex, line 15: Hmm, should we be doing this in terms of the rust-openssl API vs. in terms of our Cipher class?

src/rust/src/backend/keywrap.rs
```diff
@@ -0,0 +1,274 @@
...
+        }
+    }
+    ctx.set_padding(false);
+    Ok(ctx)
+}
+
+fn wrap_core(
+    ctx: &mut openssl::cipher_ctx::CipherCtx,
+    mut a: [u8; 8],
+    r: &mut [u8],
```
alex: I'm confused why this is taking r as mutable and writing to it, and then at the end copying to new storage. Should it not be taking r as immutable and then writing directly to result mem?

src/rust/src/backend/keywrap.rs
```diff
@@ -0,0 +1,274 @@
...
+
+#[pyo3::pyfunction]
+#[pyo3(signature = (wrapping_key, key_to_wrap, backend=None))]
+fn aes_key_wrap_with_padding<'p>(
+    py: pyo3::Python<'p>,
+    wrapping_key: CffiBuf<'_>,
+    key_to_wrap: CffiBuf<'_>,
+    backend: Option<pyo3::Bound<'_, pyo3::PyAny>>,
+) -> CryptographyResult<pyo3::Bound<'p, pyo3::types::PyBytes>> {
+    let _ = backend;
```
alex: you don't actually need the `let` I'm pretty sure

src/rust/src/backend/keywrap.rs
reaperhulk: I followed the aead.rs pattern here (cipher selected by key length, raw `CipherCtx` for the block loop), since keywrap receives raw key bytes rather than Python algorithm objects — going through the `Cipher` class would mean constructing Python `AES`/`ECB` instances from Rust just to route back into a raw context. The middle option would be a `cipher_registry::get_cipher` lookup (like cmac.rs), which keeps the registry as the single source of cipher mappings but still requires building the Python objects for the lookup. Happy to switch to either if you prefer.

---

reaperhulk: I’m inclined to leave it as is unless we want to centralize our use of symmetric ciphers down to just the Cipher class + AEADs?

alex (approved):

PR #15253, opened 2026-07-13
reject unsupported version in ocsp requests and responses

dxbjavid:
**Unchecked version field in the OCSP loaders**

The request and response version is read straight from untrusted DER but never checked, so an OCSP request or response carrying a version other than v1 loads without complaint even though RFC 6960 defines no other value. Every other top-level loader here (certificate, CSR, CRL, SCT) already rejects an unknown version at parse time, so I've added the same check to the two OCSP loaders and left a regression vector for each.


alex (approved):

dxbjavid:
TY!

PR #15262, opened 2026-07-16
reject non-zero version in pkcs7 enveloped data decryption

dxbjavid:
When decrypting a PKCS7 EnvelopedData we read the EnvelopedData and RecipientInfo version fields but never check them, so a message whose version is not zero is accepted and decrypted even though it is malformed. For the shape we accept here, with no originatorInfo or unprotected attributes and a recipient identified by issuer and serial number, RFC 5652 6.1 and 6.2.1 fix both versions at zero, so any other value is out of spec. This rejects it in both places, matching the version validation the certificate, CSR, CRL and OCSP loaders already do.

### DIFF
--- a/tests/hazmat/primitives/test_pkcs7.py
+++ b/tests/hazmat/primitives/test_pkcs7.py
@@ -1326,6 +1326,37 @@ class TestPKCS7Decrypt:
         with pytest.raises(ValueError):
             pkcs7.pkcs7_decrypt_der(enveloped, certificate, private_key, [])
 
+    @pytest.mark.parametrize("occurrence", [0, 1])
+    def test_pkcs7_decrypt_der_invalid_version(
+        self, occurrence, data, certificate, private_key
+    ):
+        # occurrence 0 is the EnvelopedData version, occurrence 1 is the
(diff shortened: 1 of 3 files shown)


tests/hazmat/primitives/test_pkcs7.py
```diff
@@ -1326,6 +1326,37 @@ def test_pkcs7_decrypt_no_encrypted_content(
...
+            pkcs7.PKCS7EnvelopeBuilder()
+            .set_data(data)
+            .add_recipient(certificate)
+        )
+        enveloped = bytearray(builder.encrypt(serialization.Encoding.DER, []))
+
+        positions = []
+        start = 0
+        while True:
+            i = enveloped.find(b"\x02\x01\x00", start)
```
alex, line 1347: Please add a proper vector for this, not this weird style.

You've gotten this feedback before, which strongly makes me think this is just an LLM. That's fine, we use LLMs to. But if you don't put actual human effort into this, you're wasting our time.

alex:
No response for 3 weeks so closing. If you're still interested, feel free to reopen.

PR #15273, opened 2026-07-17
Hoist XTS mode check out of CipherContext update loop

reaperhulk:
CipherContext::update_into called back into Python with an
is_instance(XTS) check on every 2^29-byte chunk of every update call,
even though whether the mode is XTS is fixed at construction (where an
identical check already runs). Store the result as a bool on the
context and branch on that instead, removing a Python type lookup and
isinstance call from the streaming cipher hot path.

The py parameter to update_into existed only for that check, so it is
dropped from the internal method and its callers.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01XcTyprD7CsgzGHV9kEwksy

### DIFF
--- a/src/rust/src/backend/ciphers.rs
+++ b/src/rust/src/backend/ciphers.rs
@@ -15,6 +15,7 @@ pub(crate) struct CipherContext {
     py_mode: pyo3::Py<pyo3::PyAny>,
     py_algorithm: pyo3::Py<pyo3::PyAny>,
     side: openssl::symm::Mode,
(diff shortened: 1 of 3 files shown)


reaperhulk:
This summary isn’t super accurate. When doing large chunks it calls once per 2**29, which is irrelevant, but it also calls once per update no matter what, which is inelegant and pointless.

alex (approved):

PR #15322, opened 2026-07-24
Add random-access decryption to Cobblestone

reaperhulk:
Implements `decrypt_range()` method for Cobblestone128Decryptor and Cobblestone256Decryptor, enabling efficient decryption of arbitrary byte ranges from large messages without processing the entire ciphertext.

## Summary

This change adds random-access decryption capabilities to the Cobblestone chunked encryption implementation. The new `decrypt_range()` method allows decrypting a specific range of plaintext bytes by reading and authenticating only the 16 KiB chunks that cover that range, making it practical to work with large encrypted files without loading them entirely into memory.

## Key Changes

- **New `decrypt_range()` method**: Added to both `Cobblestone128Decryptor` and `Cobblestone256Decryptor` classes. Takes a source (buffer or file-like object), offset, and length; returns authenticated plaintext for the requested range.

- **Helper functions and utilities**:
  - `read_at()`: Reads bytes from either buffer-protocol objects (direct indexing) or file-like objects (via seek/read)
  - `mixing_error()` and `truncated_error()`: Error helpers for invalid usage patterns and truncation detection
  - `ChunkCipher::decrypt_chunk_at()`: Decrypts a single chunk at an explicit index without advancing the sequential counter
  - `ChunkNonces::nonce_for()`: Stateless nonce derivation for arbitrary chunk indices

- **State tracking**: Added fields to `ChunkedDecryptor` to:
  - Retain key and context for deriving the seek cipher on first `decrypt_range()` call
  - Cache the seek cipher for reuse across multiple range reads
  - Track whether streaming (`update`/`finalize`) or seeking (`decrypt_range`) has been used to enforce mutual exclusivity

- **Mutual exclusivity enforcement**: Streaming and random-access decryption cannot be mixed on a single instance; attempting to do so raises a `ValueError`.

- **Comprehensive test coverage**: Added 20+ test cases covering:
  - Various range positions and sizes (chunk boundaries, interior ranges, final chunks)
  - Multiple source types (bytes, bytearray, memoryview, BytesIO, custom file-like objects, mmap)
  - Error conditions (wrong key/context, tampering, truncation, out-of-range reads)
  - Instance reusability and mutual exclusivity constraints

- **Documentation**: Added detailed docstrings, type hints in `.pyi` file, and comprehensive user documentation with examples showing both buffer and file-like usage patterns, including a remote range-fetch adapter example.

## Implementation Details

- The seek cipher is derived from a 56-byte header (salt + commitment) read directly from the ciphertext source on the first `decrypt_range()` call and cached for subsequent calls.
- Requested ranges are silently expanded to whole chunk boundaries for authentication, then the requested sub-range is sliced from the authenticated plaintext.
- The implementation supports both buffer-protocol objects (bytes, mmap, etc.) for direct indexing and binary file-like objects implementing `seek()` and `read()` for remote or streaming sources.
- Multi-chunk ranges are read in a single contiguous operation for efficiency.
- Range reads authenticate the bytes they return but cannot detect truncation beyond the requested range; total message length must come from a trusted source.

https://claude.ai/code/session_0172rPrpx6h454zpXTGdzez6

### DIFF
--- a/docs/cobblestone.rst
+++ b/docs/cobblestone.rst
@@ -165,6 +165,58 @@ that mandate 256-bit keys).
         :raises cryptography.exceptions.AlreadyFinalized: If
             ``finalize`` has already been called.
 
+    .. method:: decrypt_range(source, offset, length)
+
+        .. versionadded:: 50.0.0
+
+        Decrypts and returns the ``length`` plaintext bytes beginning at
+        ``offset``, reading only the ciphertext needed to cover that
+        range rather than the whole message. See :ref:`random access
+        <cobblestone-random-access>` below.
+
+        The requested range is silently widened to whole 16 KiB chunk
+        boundaries so that every chunk it touches can be authenticated
+        by its tag; the requested sub-range is then sliced out of the
+        authenticated plaintext. Unauthenticated bytes are never
+        returned.
+
+        This method is an alternative to :meth:`update`/:meth:`finalize`:
+        a single instance may be used for streaming **or** random-access
+        decryption, not both. It does not consume the instance and may be
+        called repeatedly for different ranges.
+
+        .. warning::
+
+            A range read authenticates the bytes it returns, but **not
+            the message as a whole**. It cannot detect that chunks beyond
+            the requested range were removed, so it provides no
+            protection against truncation of the overall message. The
+            total plaintext length must come from a trusted source; do
+            not infer it from the (possibly truncated) size of
+            ``source``. Whole-message truncation protection is only
+            provided by streaming through :meth:`finalize`.
+
+        :param source: The ciphertext, as either a :term:`bytes-like`
+            object (for example an :class:`mmap.mmap` of a file, which
+            avoids reading it all into memory) or a binary file-like
+            object. A file-like ``source`` only needs to implement
+            ``seek(offset)`` and ``read(size)``; nothing else is used, so
+            an object that fetches byte ranges from remote storage on
+            demand works as well as an open file.
+        :param int offset: The plaintext offset to start at.
+        :param int length: The number of plaintext bytes to return.
+        :return bytes: The authenticated plaintext for the requested
+            range.
+        :raises cryptography.exceptions.InvalidTag: If a covering chunk
+            was encrypted with a different key or context, has been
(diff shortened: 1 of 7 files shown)


docs/cobblestone.rst
```diff
@@ -180,6 +232,76 @@ that mandate 256-bit keys).
...
+    >>> import io
+    >>> from cryptography.cobblestone import (
+    ...     Cobblestone128Decryptor, Cobblestone128Encryptor
+    ... )
+    >>> key = Cobblestone128Encryptor.generate_key()
+    >>> encryptor = Cobblestone128Encryptor(key, context=b"ranged")
+    >>> plaintext = b"the quick brown fox" * 10000
+    >>> ciphertext = encryptor.update(plaintext) + encryptor.finalize()
+    >>> decryptor = Cobblestone128Decryptor(key, context=b"ranged")
+    >>> decryptor.decrypt_range(io.BytesIO(ciphertext), 40005, 9)
```
alex: Having this be an instance method along-side regular decrypt seems super fraught, what happens when you interleave calls?

docs/cobblestone.rst
```diff
@@ -165,6 +165,58 @@ that mandate 256-bit keys).
...
+            ``source``. Whole-message truncation protection is only
+            provided by streaming through :meth:`finalize`.
+
+        :param source: The ciphertext, as either a :term:`bytes-like`
+            object (for example an :class:`mmap.mmap` of a file, which
+            avoids reading it all into memory) or a binary file-like
+            object. A file-like ``source`` only needs to implement
+            ``seek(offset)`` and ``read(size)``; nothing else is used, so
+            an object that fetches byte ranges from remote storage on
+            demand works as well as an open file.
```
alex: a) We should figure out if it really makes sense to do both bytes-like AND file-like. Certainly doing both of them in one-API isn't our usual style. I think you can always turn a bytes in a file-like with `io.BytesIO` (though maybe that doesn't work for arbitrary buffers?)

b) Gah, is there no python API for `pread` so we can avoid a reliance on `seek`? Seek is annoying because it blocks any sort of concurrency.

PR #15325, opened 2026-07-25
Guard release workflows against running from forks

duksh:
wheel-builder.yml triggers on any tag push, and pypi-publish.yml chains from its completion checking only the event type and conclusion. On a fork, pushing a tag therefore runs a full wheel build and then attempts an OIDC trusted publish (and attestation minting) — safe today only because no trusted publisher is configured for forks on PyPI's side.

This adds the same repository_owner guard the scheduled bot workflows already use: sdist (which every other wheel-builder job needs) skips on forks except for pull_request runs, and the publish job requires the pyca owner. No behavior change for this repository — verified by a full green run on a fork with the guard in place.

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -27,7 +27,9 @@ jobs:
     # We're not actually verifying that the triggering push event was for a
     # tag, because github doesn't expose enough information to do so.
     # wheel-builder.yml currently only has push events for tags.
-    if: github.event_name == 'workflow_dispatch' || (github.event.workflow_run.event == 'push' && github.event.workflow_run.conclusion == 'success')
+    # The repository_owner check keeps forks from attempting a publish (or
+    # requesting OIDC tokens/attestations) if a tag is ever pushed to one.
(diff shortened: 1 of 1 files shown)


.github/workflows/wheel-builder.yml
```diff
@@ -27,8 +27,12 @@ jobs:
   sdist:
     runs-on: ubuntu-latest
     name: sdists
+    # Keep forks from running release builds on tag pushes or dispatches
+    # (pull_request runs are kept so workflow changes stay testable). All
+    # other jobs need this one, so skipping it skips the entire workflow.
+    if: github.repository_owner == 'pyca' || github.event_name == 'pull_request'
```
alex: I'm not sure I Understand why we want to skip the wheel builder on forks. Skipping publish makes sense, but wheel builder seems harmless to run.

alex (approved):

.github/workflows/wheel-builder.yml
duksh: Good point — the wheel build itself is harmless. I've narrowed the PR to just the publish job guard and dropped the wheel-builder change.

PR #15326, opened 2026-07-25
Annotate dtolnay/rust-toolchain SHA pin with its ref

duksh:
Every other SHA-pinned action carries a version comment; this one lacked it because` dtolnay/rust-toolchain` publishes no tags. 

This annotates the three uses with the branch and commit date **(master, as of Aug 23, 2025)**, matching the style used for the BoringSSL commit pins.

### DIFF
--- a/.github/actions/windows-wheel/action.yml
+++ b/.github/actions/windows-wheel/action.yml
@@ -43,7 +43,7 @@ runs:
       with:
         python-version: ${{ inputs.python-version }}
         architecture: ${{ inputs.arch }}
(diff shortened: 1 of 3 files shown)


alex:
AFAIK this syntax will not be automatically updated by dependabot, which means it'll immediately fall out of date. Unless there's a way to have dependabot keep the comments up to date, this is better left unannotated.

duksh:
That's fair — Dependabot wouldn't keep this format current, so it would go stale on the first bump. Closing.

PR #15327, opened 2026-07-25
Deny unsafe_op_in_unsafe_fn in crates containing unsafe code

duksh:
Adds `unsafe_op_in_unsafe_fn` to the existing `#![deny(...)]` in the four crates that contain unsafe code. 

Every unsafe operation inside an `unsafe fn` must now sit in an explicit `unsafe` block, which — combined with the already-denied `clippy::undocumented_unsafe_blocks` — means each one also carries a `SAFETY` comment. 

The codebase already complies (zero violations; ` cargo check --locked --workspace, cargo fmt --check,` and clippy all clean), so this only prevents silent regressions.

### DIFF
--- a/src/rust/cryptography-cffi/src/lib.rs
+++ b/src/rust/cryptography-cffi/src/lib.rs
@@ -2,7 +2,11 @@
 // 2.0, and the BSD License. See the LICENSE file in the root of this repository
 // for complete details.
 
-#![deny(rust_2018_idioms, clippy::undocumented_unsafe_blocks)]
+#![deny(
+    rust_2018_idioms,
+    unsafe_op_in_unsafe_fn,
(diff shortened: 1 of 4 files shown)


alex:
I believe this warning is already on by default in the rust 2024 edition, so there's no need to explicitly enable it.

duksh:
In edition 2024 it's warn-by-default, agreed — but the workspace is currently edition 2021 (MSRV 1.83, and the 2024 edition needs 1.85), where this lint is allow-by-default, so it's not active today. The explicit deny covers the gap until the edition bump; happy to close if that bump is imminent.

alex:
MSRV bump is in #15174, so I think we'll just get this for free then, thanks.

PR #15332, opened 2026-07-26
Permit ML-DSA by default in the X.509 verifier

alex:
Adds ML-DSA-44, ML-DSA-65, and ML-DSA-87 to `WEBPKI_PERMITTED_SPKI_ALGORITHMS` and `WEBPKI_PERMITTED_SIGNATURE_ALGORITHMS`, so policies built by `PolicyBuilder` accept ML-DSA issuer public keys and ML-DSA certificate signatures without additional configuration. Both statics are what `PolicyDefinition::new` copies into every policy, so this covers `build_server_verifier` and `build_client_verifier` alike.

ML-DSA takes no parameters and isn't parameterized by a hash, so the `AlgorithmIdentifier` is byte-identical in the SPKI and signature positions — hence one constant per variant, rather than the `SPKI_*`/signature split that RSA and ECDSA need.

The existing canonical-encoding unit tests are extended to cover the new entries. End-to-end coverage will come from wycheproof rather than a Python test here.

Note that on a backend without ML-DSA support (OpenSSL < 3.5, non-BoringSSL/AWS-LC), an ML-DSA chain now fails slightly later in `valid_issuer`, at public-key parsing with `"issuer has malformed public key"` rather than `"Forbidden public key algorithm"`. `cryptography-x509-verification` has no OpenSSL dependency, so the permit-lists can't be gated on backend support.

Verified with `nox -e local`. Since the dev container's system OpenSSL is 3.0.13, I also built OpenSSL 3.5.4 and re-ran against it via `OPENSSL_DIR` to exercise the ML-DSA path.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DJ9Qa8WGToptmm9mr2XFkt)_

### DIFF
--- a/.github/actions/fetch-vectors/action.yml
+++ b/.github/actions/fetch-vectors/action.yml
@@ -30,5 +30,5 @@ runs:
       env:
         # Latest commit on the wycheproof main branch, as of Jul 21, 2026.
         WYCHEPROOF_REF: "b61843a9a5115bb758134b6a1f5d5e502d445342" # wycheproof-ref
-        # Latest commit on the x509-limbo main branch, as of Jul 13, 2026.
-        X509_LIMBO_REF: "3e000feb51285a8153509129a62248a6b4883181" # x509-limbo-ref
+        # Latest commit on the x509-limbo main branch, as of Jul 26, 2026.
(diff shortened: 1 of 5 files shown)


woodruffw (dismissed):
LGTM, I cross-checked the OIDs against RFC 9881.

woodruffw (approved):

PR #15359, opened 2026-07-30
Release the GIL in hash, HMAC, cipher, and AEAD operations

reaperhulk:
Hashing/HMAC/symmetric-cipher/AEAD work held the GIL for the entire operation, so multi-threaded workloads serialized completely (unlike stdlib hashlib, which releases the GIL for buffers > 2047 bytes).

This wraps the underlying OpenSSL calls in `Python::detach` when the input is at least `GIL_DETACH_THRESHOLD` (2048) bytes:

- `hashes.rs`: `Hash.update`, one-shot `Hash.hash`, `XOFHash.update`
- `hmac.rs`: `HMAC.update`
- `ciphers.rs`: `CipherContext` `update`/`update_into` (all cipher-context classes) and `authenticate_additional_data`
- `aead.rs`: `EvpCipherAead` encrypt/decrypt (covering the AESGCM, ChaCha20Poly1305, AESGCMSIV, AESOCB3, AESSIV, AESCCM one-shot APIs); AAD buffers are extracted and length-checked while attached, then the process_aad + process_data + tag region runs detached. The BoringSSL/AWS-LC `EvpAead` path is unchanged.

All borrowed slices remain owned by `CffiBuf`s held on the calling frame, and no Python object access happens inside the detached closures.

Benchmarks (Linux x86-64, 4 cores, CPython 3.11, release build):

    4 threads each SHA256-hashing 16MiB:
        43.4 ms -> 11.2 ms wall (3.9x scaling; previously 1.01x)
    4 threads each AESGCM-encrypting 16MiB:
        9.3 ms -> 5.1 ms wall (1.8x; memory-bandwidth bound)
    Small payloads (below threshold) are unchanged:
        SHA256 one-shot 64B: 1093 -> 1090-1120 ns (noise)
        HMAC-SHA256 64B: 1713 -> 1746 ns (noise)
        AESGCM encrypt 64B: 617 -> 630 ns (noise)
        CTR update 64B: 163 -> 166-180 ns (noise)
    Crossover check: a 2048B hash update pays ~75 ns (~5%) for the detach,
    4KiB and up is at parity single-threaded while enabling scaling.

The threshold matters for more than the fixed detach cost: re-attaching can wait up to `sys.getswitchinterval()` (5 ms default) when another thread is CPU-bound, so operations must be large enough to amortize that. The worst-case behavior with a greedy background thread was measured to match stdlib hashlib's for the same buffer sizes.

Also adds threaded correctness tests for the detached hash and AEAD paths, and a changelog entry.

Note: this will conflict trivially with #15358 in `CipherContext::update` — whichever lands second I'll rebase.

---

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,8 +13,8 @@ Changelog
   ``cryptography.hazmat.primitives.asymmetric.dh`` and loading FFDH keys or
   parameters with the key loading APIs. Users should migrate to a more
   modern key exchange algorithm.
-* Hashing, HMAC, symmetric cipher, and AEAD operations on payloads of 2KiB
-  or more now release the GIL, allowing other threads to run while the
+* Hashing, HMAC, and symmetric cipher operations on payloads of 2KiB or
+  more now release the GIL, allowing other threads to run while the
   cryptographic work happens. Multi-threaded hashing and encryption of
   large payloads now scale across CPU cores.
 * Added ``xof()`` class methods to
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,10 +13,6 @@ Changelog
   ``cryptography.hazmat.primitives.asymmetric.dh`` and loading FFDH keys or
   parameters with the key loading APIs. Users should migrate to a more
   modern key exchange algorithm.
-* Hashing, HMAC, and symmetric cipher operations on payloads of 2KiB or
-  more now release the GIL, allowing other threads to run while the
-  cryptographic work happens. Multi-threaded hashing and encryption of
-  large payloads now scale across CPU cores.
 * Added ``xof()`` class methods to
   :class:`~cryptography.hazmat.primitives.hashes.SHAKE128` and
   :class:`~cryptography.hazmat.primitives.hashes.SHAKE256` for constructing
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,10 @@ Changelog
   ``cryptography.hazmat.primitives.asymmetric.dh`` and loading FFDH keys or
   parameters with the key loading APIs. Users should migrate to a more
   modern key exchange algorithm.
+* Hashing, HMAC, symmetric cipher, and AEAD operations on payloads of 2KiB
+  or more now release the GIL, allowing other threads to run while the
+  cryptographic work happens. Multi-threaded hashing and encryption of
+  large payloads now scale across CPU cores.
 * Added ``xof()`` class methods to
   :class:`~cryptography.hazmat.primitives.hashes.SHAKE128` and
   :class:`~cryptography.hazmat.primitives.hashes.SHAKE256` for constructing
--- a/src/rust/src/backend/aead.rs
+++ b/src/rust/src/backend/aead.rs
@@ -26,34 +26,6 @@ pub(crate) enum Aad<'a> {
     List(pyo3::Bound<'a, pyo3::types::PyList>),
 }
 
-enum ExtractedAad<'a> {
-    None,
-    Single(CffiBuf<'a>),
-    List(Vec<CffiBuf<'a>>),
-}
-
-/// Extracts and length-checks every AAD buffer. This must happen while
-/// attached to the interpreter; the returned buffers keep the underlying
-/// Python objects alive so the borrowed slices stay valid afterwards.
-fn extract_aad(aad: Option<Aad<'_>>) -> CryptographyResult<ExtractedAad<'_>> {
-    match aad {
-        None => Ok(ExtractedAad::None),
-        Some(Aad::Single(ad)) => {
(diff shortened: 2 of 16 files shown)


tests/hazmat/primitives/test_hashes.py
```diff
@@ -156,6 +157,30 @@ def test_buffer_protocol_hash():
     )
 
 
+def test_hash_threaded():
```
alex: I think these tests are extremely low value. The whole point of Rust is that we don't need half-assed tests, we know these things are correct by construction.

src/rust/src/backend/hmac.rs
```diff
@@ -82,8 +82,18 @@ impl Hmac {
         Hmac::new_bytes(py, key.as_bytes(), algorithm)
     }
 
-    fn update(&mut self, data: CffiBuf<'_>) -> CryptographyResult<()> {
-        self.update_bytes(data.as_bytes())
+    fn update(&mut self, py: pyo3::Python<'_>, data: CffiBuf<'_>) -> CryptographyResult<()> {
+        let data = data.as_bytes();
+        let ctx = self.get_mut_ctx()?;
+        if data.len() >= crate::backend::GIL_DETACH_THRESHOLD {
```
alex: we should have a function run_with_gil_detached or something that takes a threshold, so we don't have to repeat the code block everywhere.

src/rust/src/backend/hmac.rs
```diff
@@ -82,8 +82,18 @@ impl Hmac {
...
+            // The `CffiBuf` holding `data` alive is owned by this frame, so
+            // the borrowed slice remains valid while detached.
```
alex: slop comment

src/rust/src/backend/hashes.rs
```diff
@@ -98,8 +98,17 @@ impl Hash {
...
     }
 
-    fn update(&mut self, data: CffiBuf<'_>) -> CryptographyResult<()> {
-        self.update_bytes(data.as_bytes())
+    fn update(&mut self, py: pyo3::Python<'_>, data: CffiBuf<'_>) -> CryptographyResult<()> {
+        let data = data.as_bytes();
+        let ctx = self.get_mut_ctx()?;
+        if data.len() >= crate::backend::GIL_DETACH_THRESHOLD {
+            // The `CffiBuf` holding `data` alive is owned by this frame, so
+            // the borrowed slice remains valid while detached.
```
alex: slop

src/rust/src/backend/aead.rs
```diff
@@ -26,6 +26,34 @@ pub(crate) enum Aad<'a> {
     List(pyo3::Bound<'a, pyo3::types::PyList>),
 }
 
+enum ExtractedAad<'a> {
```
alex: I can't tell if this refactor is related or not, but it should be in a seperate PR

reaperhulk: The refactor is related — the detached region must not touch Python objects, so the AAD `CffiBuf`s have to be extracted before detaching rather than inside `process_aad` — but it's only needed for the AEAD detach, so I've dropped all of the AEAD changes from this PR. This PR is now hashes/HMAC/ciphers only (plus the shared `run_with_gil_detached` helper), and the AEAD work is staged as a follow-up on `claude/perf-gil-release-aead`, to be opened once this merges.

---

CHANGELOG.rst
```diff
@@ -13,6 +13,10 @@ Changelog
   ``cryptography.hazmat.primitives.asymmetric.dh`` and loading FFDH keys or
   parameters with the key loading APIs. Users should migrate to a more
   modern key exchange algorithm.
+* Hashing, HMAC, and symmetric cipher operations on payloads of 2KiB or
```
alex: I don't think we've ever documented our GIL releases in CHANGELOG before

src/rust/src/backend/hashes.rs
```diff
@@ -98,8 +98,11 @@ impl Hash {
         })
     }
 
-    fn update(&mut self, data: CffiBuf<'_>) -> CryptographyResult<()> {
-        self.update_bytes(data.as_bytes())
+    fn update(&mut self, py: pyo3::Python<'_>, data: CffiBuf<'_>) -> CryptographyResult<()> {
+        let data = data.as_bytes();
+        let ctx = self.get_mut_ctx()?;
+        crate::backend::run_with_gil_detached(py, data.len(), || ctx.update(data))?;
```
alex: wat? why is this not inside `update_bytes`?

src/rust/src/backend/hashes.rs
```diff
@@ -173,13 +181,6 @@ pub(crate) struct XOFHash {
     squeezed: bool,
 }
 
-impl XOFHash {
```
alex, line 176: why did this change...

alex (approved):

PR #15361, opened 2026-07-30
Release the GIL in one-shot AEAD encrypt/decrypt

reaperhulk:
Follow-up to #15359, carrying the AEAD portion that was split out during review: `EvpCipherAead` encrypt/decrypt (the AESGCM, ChaCha20Poly1305, AESGCMSIV, AESOCB3, AESSIV, and AESCCM one-shot APIs) now run the process_aad + process_data + tag region via `run_with_gil_detached` when plaintext + AAD is at least the 2KiB threshold.

Because the detached region must not touch Python objects, the AAD buffers are extracted and length-checked up front (`extract_aad`), keeping the `CffiBuf`s alive on the calling frame while only borrowed slices cross into the detached closure. The BoringSSL/AWS-LC `EvpAead` path is unchanged.

Benchmarks (Linux x86-64, 4 cores, CPython 3.11, release build):

    4 threads each AESGCM-encrypting 16MiB:
        9.3 ms -> 5.1 ms wall (1.8x; memory-bandwidth bound)
    AESGCM encrypt 64B: 617 -> 630 ns (noise)

---

### DIFF
--- a/src/rust/src/backend/aead.rs
+++ b/src/rust/src/backend/aead.rs
@@ -26,6 +26,34 @@ pub(crate) enum Aad<'a> {
     List(pyo3::Bound<'a, pyo3::types::PyList>),
 }
 
+enum ExtractedAad<'a> {
+    None,
+    Single(CffiBuf<'a>),
+    List(Vec<CffiBuf<'a>>),
+}
+
+/// Extracts and length-checks every AAD buffer. This must happen while
+/// attached to the interpreter; the returned buffers keep the underlying
+/// Python objects alive so the borrowed slices stay valid afterwards.
+fn extract_aad(aad: Option<Aad<'_>>) -> CryptographyResult<ExtractedAad<'_>> {
+    match aad {
+        None => Ok(ExtractedAad::None),
+        Some(Aad::Single(ad)) => {
+            check_length(ad.as_bytes())?;
+            Ok(ExtractedAad::Single(ad))
+        }
+        Some(Aad::List(ads)) => {
+            let mut bufs = Vec::with_capacity(ads.len());
+            for ad in ads.iter() {
+                let buf = ad.extract::<CffiBuf<'_>>()?;
+                check_length(buf.as_bytes())?;
+                bufs.push(buf);
+            }
+            Ok(ExtractedAad::List(bufs))
+        }
+    }
+}
+
 pub(crate) enum AeadCipher {
     Static(&'static openssl::cipher::CipherRef),
     #[cfg(not(any(
@@ -96,21 +124,10 @@ impl EvpCipherAead {
 
     fn process_aad(
         ctx: &mut openssl::cipher_ctx::CipherCtx,
-        aad: Option<Aad<'_>>,
+        aad_slices: &[&[u8]],
     ) -> CryptographyResult<()> {
-        match aad {
-            Some(Aad::Single(ad)) => {
-                check_length(ad.as_bytes())?;
-                ctx.cipher_update(ad.as_bytes(), None)?;
-            }
-            Some(Aad::List(ads)) => {
-                for ad in ads.iter() {
-                    let ad = ad.extract::<CffiBuf<'_>>()?;
-                    check_length(ad.as_bytes())?;
-                    ctx.cipher_update(ad.as_bytes(), None)?;
-                }
-            }
-            None => {}
+        for ad in aad_slices {
+            ctx.cipher_update(ad, None)?;
         }
 
         Ok(())
@@ -205,8 +222,6 @@ impl EvpCipherAead {
             ctx.encrypt_init(None, None, nonce)?;
         }
 
-        Self::process_aad(&mut ctx, aad)?;
-
         let ciphertext;
         let tag;
         if self.tag_first {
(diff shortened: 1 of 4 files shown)


src/rust/src/backend/aead.rs
```diff
@@ -215,9 +230,34 @@ impl EvpCipherAead {
...
+        let single_storage;
+        let list_storage;
+        let aad_slices: &[&[u8]] = match &extracted_aad {
+            ExtractedAad::None => &[],
+            ExtractedAad::Single(b) => {
+                single_storage = [b.as_bytes()];
+                &single_storage
+            }
+            ExtractedAad::List(bufs) => {
+                list_storage = bufs.iter().map(|b| b.as_bytes()).collect::<Vec<_>>();
```
alex: Why are we doing _yet another_ vec allocation? can we not just pass the extracteaad to the right place?

src/rust/src/backend/aead.rs
```diff
@@ -270,10 +310,36 @@ impl EvpCipherAead {
...
+                &list_storage
+            }
+        };
+        let aad_len: usize = aad_slices.iter().map(|s| s.len()).sum();
```
alex: we should just be computing the length in extract_aad

src/rust/src/buf.rs
```diff
@@ -103,6 +103,14 @@ impl<'a> CffiBuf<'a> {
         self.buf
     }
 
+    /// Like `as_bytes`, but the returned slice carries the full `'p`
+    /// lifetime instead of borrowing from `self`. The caller is
+    /// responsible for keeping this `CffiBuf` (which keeps the underlying
+    /// buffer alive and pinned) around for as long as the slice is used.
+    pub(crate) fn as_bytes_full(&self) -> &'a [u8] {
+        self.buf
+    }
```
alex: Is there a reason not to extend the lifetime on teh regular as_bytes...

src/rust/src/backend/aead.rs
```diff
@@ -26,6 +26,77 @@ pub(crate) enum Aad<'a> {
...
+    fn slices(&self) -> &[&'a [u8]] {
+        match &self.buffers {
+            AadBuffers::None => &[],
+            AadBuffers::Single { slice, .. } => slice,
+            AadBuffers::List { slices, .. } => slices,
+        }
+    }
+}
+
+fn extract_aad(aad: Option<Aad<'_>>) -> CryptographyResult<ExtractedAad<'_>> {
```
alex: you can just return a tuple here, way simpler

src/rust/src/backend/aead.rs
```diff
@@ -96,21 +167,10 @@ impl EvpCipherAead {
 
     fn process_aad(
         ctx: &mut openssl::cipher_ctx::CipherCtx,
-        aad: Option<Aad<'_>>,
+        aad_slices: &[&[u8]],
```
alex: Why doers this not just take an ExtractedAad instance? makes the code 100x simper

src/rust/src/backend/aead.rs
```diff
@@ -215,9 +273,18 @@ impl EvpCipherAead {
...
+        let is_ccm = self.is_ccm;
+        let mut work = |ctx: &mut openssl::cipher_ctx::CipherCtx| -> CryptographyResult<()> {
+            Self::process_aad(ctx, aad_slices)?;
+            Self::process_data(ctx, plaintext, ciphertext, is_ccm)?;
+            ctx.tag(tag).map_err(CryptographyError::from)?;
+            Ok(())
+        };
 
-        ctx.tag(tag).map_err(CryptographyError::from)?;
+        crate::backend::run_with_gil_detached(py, plaintext.len() + aad.len, || work(&mut ctx))?;
```
alex: just write the callback inline here....

src/rust/src/backend/aead.rs
```diff
@@ -270,10 +337,20 @@ impl EvpCipherAead {
...
 
-        Self::process_data(&mut ctx, ciphertext_data, buf, self.is_ccm)
-            .map_err(|_| exceptions::InvalidTag::new_err(()))?;
+        let is_ccm = self.is_ccm;
+        let mut work = |ctx: &mut openssl::cipher_ctx::CipherCtx| -> CryptographyResult<()> {
+            Self::process_aad(ctx, aad_slices)?;
+            Self::process_data(ctx, ciphertext_data, buf, is_ccm)
+                .map_err(|_| exceptions::InvalidTag::new_err(()))?;
+            Ok(())
+        };
```
alex: same

src/rust/src/buf.rs
reaperhulk: The `&self`-tied lifetime on `as_bytes` is doing borrow-check work: it makes `something.extract::<CffiBuf>()?.as_bytes()` a compile error when the slice outlives the temporary. With the `'p` lifetime that pattern compiles and dangles — the `CffiBuf` drop releases the buffer view, so a bytearray can be resized/freed out from under the slice. Keeping the wide-lifetime version as a separate, documented method (only needed by callers that explicitly manage the owner's lifetime, like `extract_aad`) preserves that protection everywhere else. Happy to widen `as_bytes` itself instead if you'd rather have one method.

---

alex: Tha makes it sound like this API is completely unsound and shouldn't be offered at all....

src/rust/src/backend/aead.rs
```diff
@@ -26,6 +26,70 @@ pub(crate) enum Aad<'a> {
     List(pyo3::Bound<'a, pyo3::types::PyList>),
 }
 
+/// Keeps the Python buffer objects backing an `ExtractedAad` alive and
+/// pinned; must outlive all uses of the corresponding slices.
+enum AadOwner<'a> {
```
alex: ... THIS STRUCUTRE SHOULD BE ALL YOU NEED. CFFIBUF IS ENOUGH

alex: The whole point of passing the extractedaad is that you don't need 2x crap

src/rust/src/backend/aead.rs
```diff
@@ -26,6 +26,44 @@ pub(crate) enum Aad<'a> {
...
+    fn len(&self) -> usize {
+        match self {
+            ExtractedAad::None => 0,
+            ExtractedAad::Single(buf) => buf.as_bytes().len(),
+            ExtractedAad::List(bufs) => bufs.iter().map(|b| b.as_bytes().len()).sum(),
+        }
+    }
+}
+
+fn extract_aad(aad: Option<Aad<'_>>) -> CryptographyResult<ExtractedAad<'_>> {
```
alex: oh my god, I'm losing my mind: I told you to pre-return the length in a tuple. this is not that fucking complicated.

src/rust/src/buf.rs
```diff
@@ -78,10 +78,14 @@ fn _extract_buffer_length<'p>(
     Ok((bufobj, ptrval, len))
 }
 
+// Fields are deliberately GIL-independent (`Py` rather than `Bound`,
```
reaperhulk: Useless comment.

alex: Eh, core of this comment: that it's designed to be sound with GIL released should bre tained

alex (approved):

PR #15364, opened 2026-07-30
Release the GIL during DH and DSA parameter generation

reaperhulk:
DH parameter generation runs for seconds to minutes at common key sizes and DSA parameter generation for hundreds of milliseconds, all while holding the GIL and freezing every other thread in the process. Wrap the OpenSSL calls in `Python::detach`; neither closure touches Python objects.

Deliberately NOT detached: EC keygen (~20 µs), and DH/DSA `generate_private_key` (sub-millisecond to low-millisecond). Re-attaching after a detach can wait up to `sys.getswitchinterval()` (5 ms default) when another thread is CPU-bound, so detaching operations much shorter than the switch interval makes them dramatically slower under contention (measured: EC keygen 55 µs -> 5.1 ms per op with a busy background thread). Only operations that run for at least tens of milliseconds are worth detaching here.

Benchmark (Linux x86-64, CPython 3.11, release build; background Python thread's progress while the main thread generates):

    dsa.generate_parameters(2048): background thread throughput
        1.0M iters/s -> 27M iters/s (i.e. other threads now actually run)
    ec.generate_private_key / generate_private_key: unchanged

---

### DIFF
--- a/src/rust/src/backend/dh.rs
+++ b/src/rust/src/backend/dh.rs
@@ -55,8 +55,6 @@ fn generate_parameters(
         ));
     }
 
-    // Parameter generation runs for seconds to minutes at common key
-    // sizes; let other threads run in the meantime.
     let dh = py
         .detach(|| openssl::dh::Dh::generate_params(key_size, generator))
         .map_err(|_| pyo3::exceptions::PyValueError::new_err("Unable to generate DH parameters"))?;
--- a/src/rust/src/backend/dh.rs
+++ b/src/rust/src/backend/dh.rs
@@ -34,6 +34,7 @@ struct DHParameters {
 #[pyo3::pyfunction]
 #[pyo3(signature = (generator, key_size, backend=None))]
 fn generate_parameters(
+    py: pyo3::Python<'_>,
     generator: u32,
     key_size: u32,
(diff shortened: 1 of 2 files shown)


src/rust/src/backend/dh.rs
```diff
@@ -54,7 +55,10 @@ fn generate_parameters(
         ));
     }
 
-    let dh = openssl::dh::Dh::generate_params(key_size, generator)
+    // Parameter generation runs for seconds to minutes at common key
```
reaperhulk: remove this comment

src/rust/src/backend/dsa.rs
```diff
@@ -53,8 +53,10 @@ pub(crate) fn public_key_from_pkey(
 }
 
 #[pyo3::pyfunction]
-fn generate_parameters(key_size: u32) -> CryptographyResult<DsaParameters> {
-    let dsa = openssl::dsa::Dsa::generate_params(key_size)?;
+fn generate_parameters(py: pyo3::Python<'_>, key_size: u32) -> CryptographyResult<DsaParameters> {
+    // Parameter generation takes hundreds of milliseconds or more; let
```
reaperhulk: Remove this as well

alex (approved):

PR #15369, opened 2026-07-31
Don't leak how PKCS#7 encryptedKey decryption failed

alex:
pkcs7_decrypt_der and its PEM and S/MIME variants reported invalid RSA PKCS#1 v1.5 padding, a recovered key of the wrong length, and a wrong key as three distinct errors, and the middle one disclosed the exact recovered length. That is a Bleichenbacher oracle for any caller that decrypts untrusted messages. The wrong-length case also returned before the AES-CBC pass, so the same distinction was observable by timing, proportional to a content size the attacker chooses.

Substitute a random key of the expected length on failure, per RFC 3218, and continue down an identical path so that every outcome looks alike.

EnvelopedData still authenticates nothing, so revealing whether decryption succeeded remains a CBC padding oracle regardless. That is a property of PKCS#7 rather than of this implementation, and is now documented instead.

### DIFF
--- a/tests/hazmat/primitives/test_pkcs7.py
+++ b/tests/hazmat/primitives/test_pkcs7.py
@@ -1491,7 +1491,7 @@ class TestPKCS7Decrypt:
             .encrypt(serialization.Encoding.DER, [])
         )
 
-        with pytest.raises(TypeError, match="the smartcard is unplugged"):
+        with pytest.raises(TypeError, match="TypeError for testing"):
             pkcs7.pkcs7_decrypt_der(
                 enveloped, certificate, DummyRSAPrivateKey(), []
             )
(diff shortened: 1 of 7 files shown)


tests/hazmat/primitives/test_pkcs7.py
```diff
@@ -1434,6 +1435,67 @@ def test_smime_decrypt_smime_not_encrypted(self, certificate, private_key):
...
+        # therefore silenced. A private key implemented outside this library
+        # can fail in other ways, and those must still reach the caller.
+        enveloped = (
+            pkcs7.PKCS7EnvelopeBuilder()
+            .set_data(data)
+            .add_recipient(certificate)
+            .encrypt(serialization.Encoding.DER, [])
+        )
+
+        with pytest.raises(TypeError, match="the smartcard is unplugged"):
```
reaperhulk: Let's use a message that makes it clear this is a test rather than a plausible error message.

reaperhulk (approved):

PR #15374, opened 2026-08-01
ci: add names to dependency bump steps for readability

MGPOCKY:
## Problem

In the scheduled dependency-bump workflows, several `run` steps are unnamed and longer than **300 characters**. In the GitHub Actions UI those steps collapse into generic labels, which hurts readability when a bump job fails or needs review.

## Changes

Semantics are unchanged: same step `id`s, same `bump_dependency.py` invocations, and the same `HAS_UPDATES` / `COMMIT_MSG` output wiring used by later steps.

### `.github/workflows/boring-open-awslc-bump.yml`
- Add names to `bump-boringssl`, `bump-openssl`, and `bump-awslc`

### `.github/workflows/x509-limbo-version-bump.yml`
- Add names to `bump-x509-limbo` and `bump-wycheproof`

## Test plan
- [ ] Confirm later `if:` conditions still reference the same `steps.bump-*.outputs.HAS_UPDATES`
- [ ] Confirm PR body templates still use `steps.bump-*.outputs.COMMIT_MSG`

### DIFF
--- a/.github/workflows/boring-open-awslc-bump.yml
+++ b/.github/workflows/boring-open-awslc-bump.yml
@@ -17,7 +17,8 @@ jobs:
         with:
           # Needed so we can push back to the repo
           persist-credentials: true
-      - id: bump-boringssl
+      - name: Bump BoringSSL pin
+        id: bump-boringssl
         run: |
           python3 .github/bin/bump_dependency.py \
             --name "BoringSSL" \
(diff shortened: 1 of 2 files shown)


alex:
These steps have no consumers besides the maintainers of this project, and we don't need these.

While I appreciate you're trying to contribute, and we always welcome contributions, this is not useful.

PR #15394, opened 2026-08-04
Note why we don't reject all-zeros X/Ed 25519/448 keys

alex:
Adds a short comment above `from_private_bytes` in the X25519, X448, Ed25519, and Ed448 backend modules recording that we deliberately don't reject all-zeros keys, since there's no threat model in which accepting them is a risk.

Comment-only change; no behavior difference.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JtyEzx8vFtPnRLrf5TJzsw)_


reaperhulk (approved):
clanker inoculation

PR #15400, opened 2026-08-06
Allow separate public key in CSR

ronaldvanrij:
This PR:
* Allows any public key type to be included in a CSR, in particular ML-KEM public keys.


src/cryptography/hazmat/bindings/_rust/x509.pyi
```diff
@@ -210,6 +210,9 @@ class CertificateSigningRequest:
     def tbs_certrequest_bytes(self) -> bytes: ...
     @property
     def is_signature_valid(self) -> bool: ...
+    def validate(
```
reaperhulk: This needs to be documented

vectors/cryptography_vectors/x509/requests/mldsa-mlkem1024-signing-pubkey.pem
```diff
@@ -0,0 +1,44 @@
+-----BEGIN PUBLIC KEY-----
```
reaperhulk, line 1: All the vectors need to be documented in test-vectors.rst.

reaperhulk (commented):
Two comments, but also could you share the use case here? Are there specific systems that require CSR submission for public keys they'll encode in certs that are non-signature? We obviously support this in certs, but I'm curious what platforms require submission via CSR for this.

src/cryptography/hazmat/bindings/_rust/x509.pyi
reaperhulk: Looking at this more, why not refactor `is_signature_valid` to have it take an optional public key to handle this case?

alex: It can'tm because it's a property. But I do think it'd be more appropriate to name this `verify_directly_signed_by` or something so what it's checking is clear.

vectors/cryptography_vectors/x509/requests/mldsa-mlkem1024-signing-pubkey.pem
ronaldvanrij: I wasn't aware of that, I now have added descriptions.

src/cryptography/hazmat/bindings/_rust/x509.pyi
ronaldvanrij: Indeed, I considered allowing an optional public key (as with `CertificateRevocationList`) , but that would break existing code because the function would always evaluate to a truthy value.

I've renamed the `validate` function to `verify_directly_signed_by`, and aligned its behaviour with `Certificate.verify_directly_issued_by`: made the public key mandatory and raise exceptions instead of returning a boolean. I've also added appropriate test cases.

ronaldvanrij:
The primary use case is when both digital signature and encryption are used in digital communication, for example in AS2. With the caveat here that both ends need full control over their private keys so only they can read the content and assert integrity and origin. 

Within the Dutch Government we have a similar standard called Digikoppeling, which requires the use of both digital signature and payload encryption and enforces certificates issued by specific entities, which are part of the PKIoverheid ecosystem. To request certificates, organizations need to generate their own key pairs and issue a CSR towards said entities so they can issue appropriate certificates. 

Both digital signature and encryption can currently be done using RSA, but with the PQC algorithms having only a single purpose, logically they will need to generate two key pairs (signing and non-signing) and request the two associated certificates.

docs/x509/reference.rst
```diff
@@ -1615,6 +1615,23 @@ X.509 CSR (Certificate Signing Request) Builder Object
...
 
+    .. method:: verify_directly_signed_by(public_key)
+
+        .. versionadded:: 51.0.0
+
+        :param public_key: One of
+            :data:`~cryptography.hazmat.primitives.asymmetric.types.PublicKeyTypes`.
+
+        Validates that the request is signed by the private key belonging to 
+        provided public key.
```
alex: The docs need to explain what is actually going on here -- the conventional CSR is self-signed, so why would one need this?

src/cryptography/x509/base.py
```diff
@@ -202,7 +204,45 @@ def subject_name(self, name: Name) -> CertificateSigningRequestBuilder:
         if self._subject_name is not None:
             raise ValueError("The subject name may only be set once.")
         return CertificateSigningRequestBuilder(
-            name, self._extensions, self._attributes
+            name, self._public_key, self._extensions, self._attributes
+        )
+
+    def public_key(
```
alex, line 210: This needs to be documented (with teh same caveats of why you'd need it)

docs/development/test-vectors.rst
```diff
@@ -755,6 +755,12 @@ Custom X.509 Request Vectors
   attribute whose value's tag is encoded in the long form.
 * ``zero-element-attribute.pem`` - A certificate signing request containing an
   attribute whose value has zero elements.
+* ``mldsa-mlkem768.pem`` and ``mldsa-mlkem768-signing-pubkey.pem`` - A 
+  certificate signing request signed using an ML-DSA private key, and 
+  containing a ML-KEM-768 public key.
```
alex: I'm having troubled reading this, I think it's saying that mldsa-mlkem768.pem is the CSR and -pubkey.pem is the public key assosciated with it?

ronaldvanrij: That's right. I've modified it slightly to make this more clear.

docs/x509/reference.rst
ronaldvanrij: I've modified the docstring accordingly.

src/cryptography/x509/base.py
ronaldvanrij: Same here, added a bit of a description.

reaperhulk: This still isn't documented in the CertificateSigningRequestBuilder docs.

ronaldvanrij: I had only added some docstring within the Python code. Now also in the actual documentation.

PR #15415, opened 2026-08-07
Add ty type checking alongside mypy

reaperhulk:
This PR adds ty side-by-side mypy for typechecking.

ty and mypy use incompatible suppression comments: ty ignores any `# type: ignore[...]` that carries error codes in brackets, and mypy only recognizes `# type: ignore` when it is the first comment on a line. Lines that both checkers flag now carry a mypy comment followed by a ty one, which each tool parses as its own. A bare `# type: ignore` is honored by both, so the sites that already use one are left alone.

Add ty to the pep8test dependency group and run it in the `flake` and `local` nox sessions. ty resolves its target version from `project.requires-python` rather than the interpreter it runs on, so pass `--python-version` to keep it in step with mypy; otherwise the two disagree about which `sys.version_info` branch is reachable, and ty demands `tomli` even on interpreters where the fallback import is dead.

Disable ty's `unused-type-ignore-comment`. One arm of a `sys.version_info` pair is unreachable at any given version, and ty reports the blanket ignore mypy needs there as unused. mypy's `warn_unused_ignores` already covers those comments. `unused-ignore-comment` (for `# ty: ignore`) stays on, since mypy cannot check those.

Claude-Session: https://claude.ai/code/session_01FfdhCmHmddyxdZACBwjaeg

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -54,6 +54,26 @@ def load_pyproject_toml() -> dict:
         return tomllib.load(f)
 
 
+def run_ty(session: nox.Session) -> None:
+    # ty rejects every `pytest.skip("...")` under pytest 8, the newest pytest
+    # supporting 3.9 (https://github.com/astral-sh/ty/issues/2797).
+    if sys.version_info < (3, 10):
+        session.log("Skipping ty: needs pytest 9, which needs Python 3.10+")
+        return
+
+    session.run(
+        "ty",
+        "check",
+        # ty targets `project.requires-python`, not the running interpreter.
+        f"--python-version={sys.version_info[0]}.{sys.version_info[1]}",
+        "src/cryptography/",
+        "vectors/cryptography_vectors/",
+        "tests/",
+        "release.py",
+        "noxfile.py",
+    )
+
+
 @nox.session
 @nox.session(name="tests-ssh")
 @nox.session(name="tests-randomorder")
@@ -232,6 +252,7 @@ def flake(session: nox.Session) -> None:
         "release.py",
         "noxfile.py",
     )
+    run_ty(session)
     session.run("check-sdist", "--no-isolation")
 
 
@@ -323,14 +344,8 @@ def local(session: nox.Session):
         external=True,
     )
 
-    session.run(
-        "mypy",
-        "src/cryptography/",
-        "vectors/cryptography_vectors/",
-        "tests/",
-        "release.py",
-        "noxfile.py",
-    )
+    # `local` runs ty instead of mypy; `flake` runs both.
+    run_ty(session)
 
     session.run(
         "maturin",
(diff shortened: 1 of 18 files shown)


noxfile.py
```diff
@@ -23,6 +23,12 @@
 nox.options.reuse_existing_virtualenvs = True
 nox.options.default_venv_backend = "uv"
 
+# ty resolves its target version from `project.requires-python` (3.9) rather
+# than the interpreter it runs on. Pin it to the running interpreter so ty and
+# mypy agree on which `sys.version_info` branches are reachable. No session
+# pins a python, so this is the version the session venv uses.
+TY_PYTHON_VERSION = f"{sys.version_info[0]}.{sys.version_info[1]}"
```
alex: this does not need a constant or a long comment, the comment can be a single sentence

noxfile.py
```diff
@@ -332,6 +348,17 @@ def local(session: nox.Session):
         "noxfile.py",
     )
 
+    session.run(
```
alex: in local I think we should _replace_ mypy, and go ty only, since local is optimized for speed

noxfile.py
```diff
@@ -54,6 +54,33 @@ def load_pyproject_toml() -> dict:
...
 
+def run_ty(session: nox.Session) -> None:
+    # ty binds away the first parameter of a `__call__` typed through a
+    # Callable-bounded TypeVar (https://github.com/astral-sh/ty/issues/3981),
+    # which is how pytest 8's `_WithException` protocol declares it, so it
+    # rejects every `pytest.skip("...")` in the test suite
+    # (https://github.com/astral-sh/ty/issues/2797). pytest 9 dropped that
+    # pattern but requires Python 3.10+, so skip ty on 3.9.
+    if sys.version_info < (3, 10):
+        session.log("Skipping ty: needs pytest >= 9, which needs Python 3.10+")
```
alex: @reaperhulk I'm pretty sure this means we need to wait for a fix in ty, I don't love landing this with weird skips.

pyproject.toml
```diff
@@ -159,6 +160,21 @@ warn_unused_configs = true
...
+[tool.ty.rules]
+# In `sys.version_info`-gated code one branch is always unreachable, and ty
+# reports the blanket ignore there as unused while mypy stays quiet. Whichever
+# version we check, one arm of such a pair gets flagged. mypy's
+# `warn_unused_ignores` already covers `# type: ignore` comments, so turn this
+# off. `unused-ignore-comment` (for `# ty: ignore`) stays enabled, since mypy
+# can't check those.
+# Tracked upstream as https://github.com/astral-sh/ty/issues/2681; this can be
+# re-enabled once ty stops reporting suppressions in unreachable code.
+unused-type-ignore-comment = "ignore"
```
ngnpope, line 176: Instead of this you probably want to use:

```toml
[tool.ty.analysis]
respect-type-ignore-comments = false
```

That will make ty ignore `# type: ignore[...]` and only respect `# ty: ignore[...]` which I've found to be easier.

PR #15472, opened 2026-08-17
reject empty subjectAltName and issuerAltName sequences

dxbjavid:
RFC 5280 defines subjectAltName and issuerAltName as a GeneralNames value, which is a SEQUENCE of one or more GeneralName, but the parser accepted an empty sequence, so a malformed certificate or CRL carrying a zero-entry SAN or IAN parsed to an empty extension rather than being rejected. The check sits in parse_general_names rather than on the GeneralNames type alias because the path validation engine reuses that alias and relies on an empty sequence to stand in for a certificate with no SAN. This rejects an empty sequence when the extension is parsed, matching the constraint and the behaviour already in place for an empty extendedKeyUsage.


src/rust/src/x509/common.rs
```diff
@@ -340,6 +340,17 @@ pub(crate) fn parse_general_names<'a>(
     py: pyo3::Python<'a>,
     gn_seq: &asn1::SequenceOf<'a, GeneralName<'a>>,
 ) -> CryptographyResult<pyo3::Bound<'a, pyo3::PyAny>> {
+    // GeneralNames is defined as SEQUENCE SIZE (1..MAX) OF GeneralName in
+    // RFC 5280, so an empty sequence is malformed. The path validation engine
+    // represents a certificate with no subjectAltName using its own empty
+    // sequence and does not call this.
+    if gn_seq.clone().next().is_none() {
```
alex: `SequenceOf` takes a `MINIMUM_LEN`, is there a reason not to just set that to 1?

dxbjavid: i looked at that first, but the SubjectAlternativeName and IssuerAlternativeName aliases are shared with the path validation engine, which stands in for a certificate with no SAN by parsing an empty sequence into SubjectAlternativeName (the \x30\x00 parse in cryptography-x509-verification/src/lib.rs). setting MINIMUM_LEN to 1 on the alias would make that parse fail. keeping the check in parse_general_names also means the other GeneralNames users that go through it, authorityCertIssuer and the CRL distribution point names, pick up the same 1..MAX constraint, since 5280 defines them all the same way. happy to move it to a type-level constraint if you'd rather give verification its own zero-minimum type.

alex: Oh, I think we should probably fix the \x30\x00 callsites to just store an Option<SequenceOf<>>

dxbjavid: done. set MINIMUM_LEN to 1 on both aliases and made the verification NameChain store an Option<SubjectAlternativeName> instead of parsing the empty `\x30\x00` sequence, so the empty rejection now comes from the type. parse_general_names is generic over the minimum length now so the len-0 GeneralNames callers still go through unchanged. rust builds clean.

alex:
@mhils it looks like mitmproxy's tests rely on SANs that are 0 length, even though they're not permitted by the RFC. Is that intentional, or inadveratnat?

PR #15350, opened 2026-07-29
Promote Windows on ARM (win_arm64) to a fully supported platform

khmyznikov:
Windows on ARM was previously kept in a stability validation period: its wheel was built and smoke-tested but not uploaded, and the arm64 test and wheel-build steps used continue-on-error with soft-failure warnings so problems would not block CI or releases.

This promotes win_arm64 to full support:

- Upload the arm64 wheel like every other platform (drop the arch guard in the windows-wheel action).
- Remove continue-on-error and the "Report soft failure" steps for the arm64 entries in ci.yml and wheel-builder.yml, so arm64 failures are treated as real failures.
- Add Python 3.15-dev arm64 entries for both tests (ci.yml) and wheel builds (wheel-builder.yml).

Maintenance support:
- Python for win_arm64 got promoted [to a Tier 2](https://peps.python.org/pep-0011/#tier-2)
- No issues with wheel builds or tests [were observed](https://github.com/CoreCollective-dev/windows-on-arm-wg/actions/workflows/monitoring_cryptography.yml) in the last months.
- You have me @khmyznikov and @paveliak if you suddenly start observing an issues on gh actions/runner side, in addition to [CoreCollective working group](https://github.com/CoreCollective-dev/windows-on-arm-wg).

PR #15496, opened 2026-08-22
Build abi3.abi3t wheels for CPython 3.15+

mayeut:
This is a follow-up on #15063 which updates the wheel-builder workflow to build abi3.abi3t wheels for CPython 3.15+ now that CPython 3.15.0rc1 is available.

From #15063:
> What's the next step after merging this?
> > Probably worth doing a dry-run of the release automation to verify that you do actually get three wheels per platform out to upload to PyPI. Other than that, I think just a release advertising Python 3.15 support.

Release automation uses the wheel-builder workflow updated here.
Python 3.15 has been added in project classifiers.


.github/actions/windows-wheel/action.yml
```diff
@@ -43,6 +43,7 @@ runs:
       with:
         python-version: ${{ inputs.python-version }}
         architecture: ${{ inputs.arch }}
+        allow-prereleases: ${{ inputs.python-version == '3.15t' }}
```
alex: We do not want to ever use allow-prereleases. If you want to use a pre-release, you should specify the version.

alex (commented):
@reaperhulk do we want to drop 3.14t wheels as soon as we're doing abi3t wheels? I strongly prefer that, but curious for your views.

reaperhulk:
> @reaperhulk do we want to drop 3.14t wheels as soon as we're doing abi3t wheels? I strongly prefer that, but curious for your views.

Normally I’d be strongly against that, but I think free-threaded consumers are pretty aggressive in their upgrades right now.

However, if we’re shipping abi3t day 1 of 3.15 we’re cutting off wheels for every existing free-threaded consumer so I think we need one release with both?

alex:
Well, we'd need to choose when we want to do a release with this. Are we gonig to try to have it for 3.15 release day?

.github/actions/windows-wheel/action.yml
mayeut: I removed the usage of allow-prereleases.
Depending on the decision by maintainers regarding when this should be available, the modifications could be reduced for Windows if after 3.15.0.final (they can always be reverted after 3.15.0.final if this goes in before that)

alex:
Ok, we've discussed and we'd like to land this in the next month and then we'll backport it and do a point release. Can you fix the merge conflicts? THanks!

.github/workflows/wheel-builder.yml
```diff
@@ -299,17 +321,18 @@ jobs:
...
         include:
           - WINDOWS: {ARCH: 'arm64', WINDOWS: 'arm64', RUST_TRIPLE: 'aarch64-pc-windows-msvc', RUNNER: 'windows-11-arm'}
-            PYTHON: {VERSION: "3.14", "ABI_VERSION": "py311"}
+            PYTHON: {VERSION: "3.14", "ABI_VERSION": "abi3-py311", FREETHREADED: false}
           - WINDOWS: {ARCH: 'arm64', WINDOWS: 'arm64', RUST_TRIPLE: 'aarch64-pc-windows-msvc', RUNNER: 'windows-11-arm'}
-            PYTHON: {VERSION: "3.14t"}
+            PYTHON: {VERSION: "3.14t", FREETHREADED: true}
           - WINDOWS: {ARCH: 'arm64', WINDOWS: 'arm64', RUST_TRIPLE: 'aarch64-pc-windows-msvc', RUNNER: 'windows-11-arm'}
-            PYTHON: {VERSION: "3.15-dev", "ABI_VERSION": "py311"}
+            PYTHON: {VERSION: "3.15.0-rc.2", "ABI_VERSION": "abi3t-py315", FREETHREADED: true}
```
mayeut, line 335: Despite the `"ABI_VERSION": "py311"`,  this axis from #15350 was [building](https://github.com/pyca/cryptography/actions/runs/30497034618/job/90728238260#step:3:1936) a cp315-abi3.abi3t so I changed it to accurately reflect this.

mayeut:
Just some notes:
- Linux is still using 3.150rc1 and would need an image rebuild based on the latest manylinux images to pick-up 3.15.0rc2
- CI failures seem like transient network errors

alex:
Thanks! I just kicked off a rebuild of our docker images, so rc2 should be available momentarily.

mayeut:
Linux is now getting 3.15.0rc2 after the rebuild of docker images

.github/workflows/wheel-builder.yml
```diff
@@ -299,17 +321,18 @@ jobs:
...
           - {ARCH: 'x64', WINDOWS: 'win64', RUST_TRIPLE: 'x86_64-pc-windows-msvc', RUNNER: 'windows-latest'}
         PYTHON:
-          - {VERSION: "3.14", "ABI_VERSION": "py39"}
-          - {VERSION: "3.14", "ABI_VERSION": "py311"}
-          - {VERSION: "3.14t"}
-          - {VERSION: "pypy-3.11"}
+          - {VERSION: "3.14", "ABI_VERSION": "abi3-py39", FREETHREADED: false}
+          - {VERSION: "3.14", "ABI_VERSION": "abi3-py311", FREETHREADED: false}
+          - {VERSION: "3.14t", FREETHREADED: true}
+          - {VERSION: "3.15.0-rc.2", "ABI_VERSION": "abi3t-py315", FREETHREADED: true}
```
alex, line 327: hmm, why do these need the FREETHREADED: true, is the `t` in the version not sufficient?

alex (approved):
Besides that question, this lgtm

.github/workflows/wheel-builder.yml
mayeut: It's only for consistency when `t` is in the version spec but the `3.15.0-rc.2` version can't have `t` in its spec (setup-python limitation) which is why `FREETHREADED` was a needed addition in the first place.
Relates to https://github.com/pyca/cryptography/pull/15496#discussion_r3836335992 (although the comment was not explicit about which modification was added for rc vs final, this is the one).

PR #15510, opened 2026-08-24
docs: improve two-factor authentication documentation outline

wbolster:
This improves the document structure for the HOTP and TOTP docs. Both now have a heading at the same level, so that both appear in the navigation sidebar.

Before this change, HOTP had an autogenerated heading, while TOTP was incorrectly hidden as a subsection of the HOTP section, due to Sphinx automatically adding sub-headings for :class: blocks.

Also lift the section about the provisioning URI to a sibling level, since it applies equally to HOTP and TOTP.

### DIFF
--- a/docs/hazmat/primitives/twofactor.rst
+++ b/docs/hazmat/primitives/twofactor.rst
@@ -7,14 +7,21 @@ Two-factor authentication
 
 This module contains algorithms related to two-factor authentication.
 
-Currently, it contains an algorithm for generating and verifying
-one time password values based on Hash-based message authentication
-codes (HMAC).
+Currently, it contains an algorithm for generating and verifying one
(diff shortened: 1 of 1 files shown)


reaperhulk (approved):

wbolster:
for easier review, fully expanded outline before (incorrect):

```
Two-factor authentication
- InvalidToken
- HOTPHashTypes
- HOTP (class)
- Throttling
- Re-synchronization of the counter
  - TOTP (class)
- Provisioning URI
```

… and after (improved):


```
Two-factor authentication
- InvalidToken
- HOTP
  - HOTPHashTypes
  - HOTP (class)
  - Throttling
  - Re-synchronization of the counter
- TOTP
  - TOTP (class)
- Provisioning URI
```

Issue #6982, opened 2022-03-19
Possible bug: empty sequence in NameConstraints


mathiasertl:
Hi,

Now I've been going through various RFCs again and again, and I'm still not quite sure if this is a subtle bug in cryptography, or if this is actually valid x509.  Regarding `permittedSubtrees` and `excludedSubtrees`: Is an empty sequence valid at all (and if yes, what are the semantics?).

cryptography does not treat an empty list any different then any other list, but treats `None` special. That means, you can even create an extension with an empty _list_ of permitted/excluded subtrees, while of course passing None raises an exception:

```python
>>> x509.NameConstraints(permitted_subtrees=[], excluded_subtrees=[])
<NameConstraints(permitted_subtrees=[], excluded_subtrees=[])>
>>> x509.NameConstraints(permitted_subtrees=None, excluded_subtrees=None)
...
    ValueError: At least one of permitted_subtrees and excluded_subtrees must not be None
```

The public bytes generated for an empty list also differ to a `None` value:

```
>>> x509.NameConstraints(permitted_subtrees=[x509.DNSName('example.com')], excluded_subtrees=None).public_bytes()
b'0\x11\xa0\x0f0\r\x82\x0bexample.com'
>>> x509.NameConstraints(permitted_subtrees=[x509.DNSName('example.com')], excluded_subtrees=[]).public_bytes()
b'0\x13\xa0\x0f0\r\x82\x0bexample.com\xa1\x00'
```

So is an empty list something we need to support? Or is it actually a bug?

Slightly related side-note: can we make both values default to `None` in the constructor, that way you can also pass just one of the values when creating the extension. If you agree, I will submit an MR.

alex:
Yes, this is a bug, we should be rejecting non-none values that are less than 0 because:

```
GeneralSubtrees ::= SEQUENCE SIZE (1..MAX) OF GeneralSubtree
```

mathiasertl:
GREAT! (I was hoping for that). I will submit a PR.

alex:
Great, thank you!

mathiasertl:
> Slightly related side-note: can we make both values default to None in the constructor, that way you can also pass just one of the values when creating the extension. If you agree, I will submit an MR.

And what do you think? :-)

alex:
We generally avoid having defaults in APIs -- to ensure that consumers aren't accidentally missing details that are relevant. We know it's a bit more verbose, but we think it's the right trade-off.

PR #15560, opened 2026-08-31
Reject present-but-empty name constraint subtrees

avalyset:
Reported privately as GHSA-fw9w-89pp-fpmg. The advisory is not public, so there is nothing to link. The response there was that this is a regular bug rather than a security issue, with a request to open a public PR instead; this is that PR. We had offered no severity assessment either way.

## The rule is already enforced in the Python layer

`src/cryptography/x509/extensions.py:1344`:

```python
if not permitted_subtrees:
    raise ValueError(
        "permitted_subtrees must be a non-empty list or None"
    )
```

That check came from #6982, "Possible bug: empty sequence in NameConstraints" (opened and closed 2022-03-19), where the conclusion was explicit:

> Yes, this is a bug, we should be rejecting non-none values that are less than 0 [sic] because:
>
> `GeneralSubtrees ::= SEQUENCE SIZE (1..MAX) OF GeneralSubtree`

## The Rust path validator did not enforce it

`policy/extension.rs` rejected only when *both* subtree fields were empty, so an empty `permittedSubtrees` combined with a non-empty `excludedSubtrees` passed that check. In `lib.rs:258-274` the permit flag then keeps its default:

```rust
let mut permit = true;
if let Some(permitted_subtrees) = &constraints.permitted_subtrees {
    for p in permitted_subtrees.clone() {
        ...
    }
}
```

With an empty sequence the loop body never runs, so `permit` stays `true` and every SAN is accepted by the permitted side.

## Basis

RFC 5280 6.1.4 (g)(1) sets `permitted_subtrees` to the intersection of its previous value and the value in the extension; the intersection with an empty set is empty. RFC 5280 6.1.3 (b) requires the name to lie within `permitted_subtrees`, so with an empty one no name qualifies and every certificate below that CA should be rejected. It was accepting all of them instead.

Reaching this requires a signed CA certificate carrying such an extension, and RFC 5280 4.2.1.10 says conforming CAs must not issue one.

## The change

`GeneralSubtrees ::= SEQUENCE SIZE (1..MAX)` constrains both fields, so this rejects a present-but-empty `permittedSubtrees` and a present-but-empty `excludedSubtrees`. Only the first has any consequence; an empty `excludedSubtrees` excludes nothing, which is harmless. But it is the same ASN.1 requirement on the same type, and enforcing it for one field and not the other seemed arbitrary. Happy to drop the second check if you would rather keep the diff to the field that matters.

The existing both-empty check is kept. I checked whether it became redundant: it does not. It is the only one that covers a `NameConstraints` where *neither* field is present, which `is_some_and` by construction never matches.

## Behaviour

The probe from the advisory, through the public API. The CA's `NameConstraints` has to be built as raw DER, because the Python API refuses to construct this shape at all, which is itself part of the point.

Before, against 50.0.1 from PyPI:

```
Python API rejects empty permitted_subtrees -> ValueError: permitted_subtrees must be a non-empty list or None
NameConstraints DER used: 30 13 a0 00 a1 0f 30 0d 82 0b 62 61 64 2e 65 78 61 6d 70 6c 65

  permitted=EMPTY,        excluded=[bad.example] -> ACCEPTED
  permitted=[ok.example], excluded=[bad.example] -> rejected: validation failed: candidates exhausted: no permitted name constraints matched SAN
```

After, with this branch:

```
  permitted=EMPTY,        excluded=[bad.example] -> rejected: validation failed: candidates exhausted: nameConstraints permittedSubtrees must not be empty
  permitted=[ok.example], excluded=[bad.example] -> rejected: validation failed: candidates exhausted: no permitted name constraints matched SAN
```

The second line is the control, and it is unchanged.

## Tests

Three tests in `policy::extension::tests`, using raw DER because an empty sequence cannot be built through the writing API: empty `permittedSubtrees` with non-empty `excludedSubtrees`, empty `excludedSubtrees` with non-empty `permittedSubtrees`, and a control with both non-empty that must still be accepted.

Verified they fail without the fix: reverting the check and keeping the tests turns both rejection tests red while the control stays green.

## Checks

- `cargo test --all` green across the workspace, including the 43 tests in `cryptography-x509-verification`
- `cargo fmt --all -- --check` clean
- `cargo clippy -p cryptography-x509-verification --all-targets`: 8 warnings, the same 8 as on `main`, none on the changed lines
- `pytest tests/`: 4490 passed, 199 skipped (`wycheproof_root` and `x509_limbo_root` not available locally)

### DIFF
--- a/.github/actions/fetch-vectors/action.yml
+++ b/.github/actions/fetch-vectors/action.yml
@@ -30,5 +30,5 @@ runs:
       env:
         # Latest commit on the wycheproof main branch, as of Aug 19, 2026.
         WYCHEPROOF_REF: "dac1dd4729fd1f8dd9e1e9f3dce51d783da6c166" # wycheproof-ref
-        # Latest commit on the x509-limbo main branch, as of Aug 28, 2026.
-        X509_LIMBO_REF: "972626160c26b45426bbd8c935a605219bd93207" # x509-limbo-ref
+        # Latest commit on the x509-limbo main branch, as of Sep 02, 2026.
+        X509_LIMBO_REF: "21cc053f7edbd22e0e8d8a98a7fe13918912af9d" # x509-limbo-ref
--- a/src/rust/cryptography-x509-verification/src/policy/extension.rs
+++ b/src/rust/cryptography-x509-verification/src/policy/extension.rs
@@ -739,6 +739,32 @@ mod ca {
                 )));
             }
 
+            // `GeneralSubtrees ::= SEQUENCE SIZE (1..MAX) OF GeneralSubtree`,
+            // so a subtree field that is present must not be empty. This
+            // matters most for permittedSubtrees: RFC 5280 6.1.4 (g)(1)
+            // intersects it with its previous value, and 6.1.3 (b) requires
+            // each name to lie within it, so an empty permittedSubtrees admits
+            // no name at all. Accepting it here would instead permit every
+            // name, because the loop over the subtrees never runs.
+            if name_constraints
+                .permitted_subtrees
+                .as_ref()
+                .is_some_and(|pst| pst.is_empty())
+            {
+                return Err(ValidationError::new(ValidationErrorKind::Other(
+                    "nameConstraints permittedSubtrees must not be empty".to_string(),
+                )));
+            }
+            if name_constraints
+                .excluded_subtrees
+                .as_ref()
+                .is_some_and(|est| est.is_empty())
+            {
+                return Err(ValidationError::new(ValidationErrorKind::Other(
+                    "nameConstraints excludedSubtrees must not be empty".to_string(),
+                )));
+            }
+
             // NOTE: Both RFC 5280 and CABF require each `GeneralSubtree`
             // to have `minimum=0` and `maximum=NULL`, but experimentally
             // not many validators check for this.
@@ -798,9 +824,9 @@ mod tests {
 
     use asn1::{ObjectIdentifier, SimpleAsn1Writable};
     use cryptography_x509::extensions::{BasicConstraints, Extension};
-    use cryptography_x509::oid::BASIC_CONSTRAINTS_OID;
+    use cryptography_x509::oid::{BASIC_CONSTRAINTS_OID, NAME_CONSTRAINTS_OID};
 
-    use super::{Criticality, ExtensionValidator};
+    use super::{ca, Criticality, ExtensionValidator};
     use crate::certificate::tests::PublicKeyErrorOps;
     use crate::ops::tests::{cert, epoch, v1_cert_pem};
     use crate::ops::{CryptoOps, VerificationCertificate};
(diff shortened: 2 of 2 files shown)


alex:
It should be possible to write a test case for this against the Python APIs - that's preferrable to writing directly against the Rust APIs. Ideally this could just be a testcase in x509-limbo.

avalyset:
Done — the testcase is now C2SP/x509-limbo#658, as `rfc5280::nc::permitted-empty-sequence-excluded-nonempty`, with `rfc5280::nc::permitted-nonempty-excluded-nonempty` as the control. It builds the extension through the Python API using the private-attribute bypass the neighbouring `intermediate_permitted_excluded_subtrees_both_*` cases already use, so nothing is hand-encoded. Against 50.0.1 the control passes and the empty-permittedSubtrees chain is accepted where limbo expects it to fail.

Should the Rust tests in this PR come out once the limbo case lands, or stay as a unit test?

alex:
Yes, the rust tests should be removed once tha tlands.

avalyset:
Understood — I will drop the Rust tests here as soon as C2SP/x509-limbo#658 lands.

avalyset:
Removed `test_ca_name_constraints_non_empty_permitted_subtrees`, which `rfc5280::nc::permitted-nonempty-excluded-nonempty` now covers.

`test_ca_name_constraints_empty_permitted_subtrees` has to stay for now. `fetch-vectors` pins x509-limbo to `972626160c26b45426bbd8c935a605219bd93207` which predates the merge of C2SP/x509-limbo#658 on 2026-09-01, so neither new case is in the vectors CI actually runs — dropping the test takes `extension.rs:754-756` to 99% and fails the coverage gate. It can go once the pin moves.

`test_ca_name_constraints_empty_excluded_subtrees` stays regardless: no limbo case covers a non-empty `permittedSubtrees` with an empty `excludedSubtrees`, which is the second `is_empty` check in this PR. I can add that one to limbo too if you would rather have it there.

alex:
https://github.com/pyca/cryptography/pull/15580 has the fix, but there's a chicken and egg situation, if you cherry pick that it solves this.

avalyset:
Cherry-picked 319a8a4 from #15580, which moves the pin to 21cc053 — the merge commit for the limbo PR. With those vectors in CI, `test_ca_name_constraints_empty_permitted_subtrees` is covered by `rfc5280::nc::permitted-empty-sequence-excluded-nonempty` and is now removed.

`test_ca_name_constraints_empty_excluded_subtrees` still stays: no limbo case covers a non-empty `permittedSubtrees` with an empty `excludedSubtrees`. I can add that one to limbo as well if you would rather have it there.

alex:
Yes, we'd rather tests live in limbo

woodruffw:
I flagged this in https://github.com/C2SP/x509-limbo/pull/660 as well, but noting here too: AFAICT basically zero validators conform to this ATM. 

That also goes for the case added in https://github.com/C2SP/x509-limbo/pull/658: 

https://x509-limbo.com/testcases/rfc5280/#rfc5280ncpermitted-empty-sequence-excluded-nonempty

woodruffw:
For context, that would be consistent with that the validator currently does, i.e. we don't target requirements that are widely ignored by other validators:

https://github.com/pyca/cryptography/blob/5936b461718024e377713d875b6c59fe5a5df3f5/tests/x509/verification/test_limbo.py#L37-L39

avalyset:
Closing this — the case is now covered in x509-limbo (C2SP/x509-limbo#658 and #660), which is where you said you'd rather have it. Thanks for the look.

PR #15569, opened 2026-09-02
Don't build the test suites of the OpenSSL forks we compile in CI

reaperhulk:
Split out of #15568.

`ninja install` depends on `all`, which includes the test suites of BoringSSL, AWS-LC and LibreSSL unless they're turned off (`BUILD_TESTING` / `LIBRESSL_TESTS`), and OpenSSL's default target builds its test programs unless configured with `no-tests`. None of them are ever run in CI.

Build times measured for the from-source builds:

| Library | Before | After |
|---|---|---|
| BoringSSL, local 4 cores | 141s | 48s |
| AWS-LC, local 4 cores | 87s | 15s* |
| BoringSSL, CI | 179–187s | 98–115s* |
| AWS-LC, CI | 154–162s | 81–82s* |
| OpenSSL master, CI | 228s | 156s* |

\* measured on #15568 with a superset of this change (that branch also skips the AWS-LC `bssl` tool and OpenSSL's man pages), so the numbers here should be close but not identical.

The script hash is part of the OpenSSL cache key, so every cached library build is rebuilt once after this merges.

---

### DIFF
--- a/.github/bin/build_openssl.sh
+++ b/.github/bin/build_openssl.sh
@@ -17,8 +17,7 @@ if [[ "${TYPE}" == "openssl" ]]; then
   # linker doesn't load the system one.
   sed -i "s/^SHLIB_VERSION=.*/SHLIB_VERSION=100/" VERSION.dat
 
-  # CONFIG_FLAGS is a global coming from a previous step. no-tests skips
-  # building the test programs, which are never run here.
+  # CONFIG_FLAGS is a global coming from a previous step
   ./config ${CONFIG_FLAGS} no-tests -fPIC --prefix="${OSSL_PATH}"
 
   make depend
--- a/.github/bin/build_openssl.sh
+++ b/.github/bin/build_openssl.sh
@@ -17,8 +17,9 @@ if [[ "${TYPE}" == "openssl" ]]; then
   # linker doesn't load the system one.
   sed -i "s/^SHLIB_VERSION=.*/SHLIB_VERSION=100/" VERSION.dat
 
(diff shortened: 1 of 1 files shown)


.github/bin/build_openssl.sh
```diff
@@ -17,8 +17,9 @@ if [[ "${TYPE}" == "openssl" ]]; then
   # linker doesn't load the system one.
   sed -i "s/^SHLIB_VERSION=.*/SHLIB_VERSION=100/" VERSION.dat
 
-  # CONFIG_FLAGS is a global coming from a previous step
-  ./config ${CONFIG_FLAGS} -fPIC --prefix="${OSSL_PATH}"
+  # CONFIG_FLAGS is a global coming from a previous step. no-tests skips
+  # building the test programs, which are never run here.
```
alex: comment not rqeuired, its very obvious what no-tests does

.github/bin/build_openssl.sh
```diff
@@ -55,7 +55,9 @@ elif [[ "${TYPE}" == "boringssl" ]]; then
   git clone https://boringssl.googlesource.com/boringssl
   pushd boringssl
   git checkout "${VERSION}"
-  cmake -GNinja -B build -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_INSTALL_PREFIX="${OSSL_PATH}"
+  # install depends on all, which includes the (large) test suite unless
+  # BUILD_TESTING is off.
```
alex: what I said applies to all these comments

alex (approved):

PR #15570, opened 2026-09-02
Key the cargo cache on the OpenSSL build it was compiled against

reaperhulk:
Split out of #15568.

openssl-sys registers `cargo:rerun-if-changed` on the OpenSSL include directory, and cargo evaluates that by mtime. The cargo cache was keyed on the OpenSSL *configuration* but not on the specific *build*, so a cargo cache saved before the library was rebuilt sees "newer" headers on every run and recompiles openssl-sys, openssl, and everything above them, until the next Cargo.lock or rustc change happens to rotate the key.

- **macOS / Windows**: the pyca/infra artifact's mtimes are pinned to its creation time. On main today every macOS job recompiles openssl-sys, openssl (and, for unrelated reasons handled separately, pyo3-ffi and pyo3) because the artifact is newer than the cargo cache. The artifact id becomes part of the cache key, which means downloading before the cache step. The `fetch-openssl` action gains an `artifact-id` output.
- **Linux**: the OpenSSL cache key includes the build script hash, the cargo cache key only included the config hash. After #15569 every from-source library was rebuilt with fresh mtimes against an unchanged cargo cache, so the same recompile is now happening on every linux job with a custom OpenSSL. The cargo cache key now uses the OpenSSL cache key.

No cache-key suffix bump is needed: the keys change by construction.

Timing comparison to follow as a comment once the warm run is in.

---

### DIFF
--- a/.github/actions/build-openssl/action.yml
+++ b/.github/actions/build-openssl/action.yml
@@ -17,6 +17,9 @@ outputs:
   hash:
     description: "Hash of the build inputs, for use in other cache keys"
     value: ${{ steps.config.outputs.hash }}
+  cache-key:
+    description: "The key the build is cached under. Cache keys for things compiled against this build should include it, see the linux job in ci.yml"
+    value: ${{ steps.config.outputs.cache-key }}
 
 runs:
   using: "composite"
--- a/.github/actions/build-openssl/action.yml
+++ b/.github/actions/build-openssl/action.yml
@@ -18,7 +18,7 @@ outputs:
     description: "Hash of the build inputs, for use in other cache keys"
     value: ${{ steps.config.outputs.hash }}
   cache-key:
-    description: "The key the build is cached under. Cache keys for things compiled against this build should include it, see the linux job in ci.yml"
+    description: "The key the build is cached under"
     value: ${{ steps.config.outputs.cache-key }}
 
 runs:
--- a/.github/actions/fetch-openssl/action.yml
+++ b/.github/actions/fetch-openssl/action.yml
@@ -20,7 +20,7 @@ inputs:
 
 outputs:
   artifact-id:
-    description: "The id of the downloaded artifact, for use in cache keys"
+    description: "The id of the downloaded artifact"
     value: ${{ fromJSON(steps.download.outputs.artifacts)[0].id }}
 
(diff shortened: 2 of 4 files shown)


.github/actions/build-openssl/action.yml
```diff
@@ -17,6 +17,9 @@ outputs:
   hash:
     description: "Hash of the build inputs, for use in other cache keys"
     value: ${{ steps.config.outputs.hash }}
+  cache-key:
+    description: "The key the build is cached under. Cache keys for things compiled against this build should include it, see the linux job in ci.yml"
```
alex: Don't refer to other implementation details, it just goes out of date.

alex (commented):
Trim down all comments you added by 50%

alex (approved):

reaperhulk:
Timing, comparing two warm-cache runs on main: the #15571 merge run (33588813690, cargo caches from before this change) and a re-run of this PR's merge commit (33588967176 attempt 2, caches saved under the new keys).

Which non-workspace crates cargo recompiled in the build step:

| Job | Before | After |
|---|---|---|
| linux, openssl 3.0.22 / 3.4.7 / 3.5.8 | openssl-sys, openssl | none |
| linux, rust,tests, openssl 4.0.2 | openssl-sys, openssl, pyo3-ffi, pyo3 | pyo3-ffi, pyo3 |
| macos 3.9 / 3.14 | openssl-sys, openssl | none |
| macos 3.14t | openssl-sys, openssl, pyo3-ffi, pyo3 | pyo3-ffi, pyo3 |
| windows 3.14 | none | none |

The openssl-sys / openssl recompile is gone everywhere it was happening. The remaining pyo3-ffi / pyo3 recompile on the rust sessions and the free-threaded jobs has a different cause (the interpreter path pyo3-build-config fingerprints) and is a separate change.

Build-step medians across the affected job groups, before → after: linux custom-OpenSSL "Create nox environment" 39s → 41s (n=13), macOS "Build nox environment" 39s → 39s (n=5). In other words the recompile is a few seconds inside a step whose runner-to-runner noise is ±10s, so this is a correctness fix for the cache rather than a measurable wall-clock win on its own.

---

PR #15571, opened 2026-09-02
Build BoringSSL and AWS-LC with optimization in CI

reaperhulk:
Split out of #15568.

Neither project sets a default `CMAKE_BUILD_TYPE`, and the CI build didn't pass one, so CMake passed no optimization flags at all: the libraries under test were compiled at -O0 (checked by inspecting the generated `build.ninja`). `RelWithAsserts` is Release without `NDEBUG`, i.e. -O3 with asserts kept, and is what BoringSSL's own CI uses.

This changes what's under test (an optimized library rather than an unoptimized one), so it's its own PR.

Local build times on 4 cores, both with the test suites already disabled by #15569:

| Library | -O0 | RelWithAsserts |
|---|---|---|
| BoringSSL | 48s | 65s |
| AWS-LC | 15s | 29s |

The library builds are cached, so that cost is paid once per bump; the pytest run against them is paid on every run. Timing comparison to follow as a comment once the warm run is in.

---

### DIFF
--- a/.github/bin/build_openssl.sh
+++ b/.github/bin/build_openssl.sh
@@ -55,7 +55,10 @@ elif [[ "${TYPE}" == "boringssl" ]]; then
   git clone https://boringssl.googlesource.com/boringssl
   pushd boringssl
   git checkout "${VERSION}"
-  cmake -GNinja -B build -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_TESTING=OFF -DCMAKE_INSTALL_PREFIX="${OSSL_PATH}"
+  # Without a build type CMake passes no optimization flags at all.
+  # RelWithAsserts is Release with asserts kept, which is what BoringSSL's
+  # own CI uses.
+  cmake -GNinja -B build -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_TESTING=OFF -DCMAKE_BUILD_TYPE=RelWithAsserts -DCMAKE_INSTALL_PREFIX="${OSSL_PATH}"
   ninja -C build install
   # delete binaries we don't need
   rm -rf "${OSSL_PATH}/bin"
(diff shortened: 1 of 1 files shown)


.github/bin/build_openssl.sh
```diff
@@ -55,7 +55,10 @@ elif [[ "${TYPE}" == "boringssl" ]]; then
   git clone https://boringssl.googlesource.com/boringssl
   pushd boringssl
   git checkout "${VERSION}"
-  cmake -GNinja -B build -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_TESTING=OFF -DCMAKE_INSTALL_PREFIX="${OSSL_PATH}"
+  # Without a build type CMake passes no optimization flags at all.
+  # RelWithAsserts is Release with asserts kept, which is what BoringSSL's
+  # own CI uses.
```
alex: delete these pointless comments

alex (approved):

reaperhulk:
Timing from this PR's CI run (33588358341) versus the main run right after #15569 merged (33587494800), so both sides have the test suites disabled. Library builds are cold in both runs; pytest time is independent of cache state.

| Job | Library build, main (-O0) | Library build, this PR | pytest, main (-O0) | pytest, this PR |
|---|---|---|---|---|
| boringssl | 80s | 108s | 35.1s | 28.2s |
| boringssl, rust 1.83 | 75s | 100s | 34.7s | 27.0s |
| aws-lc | 71s | 94s | 48.7s | 66.2s |
| aws-lc, rust 1.83 | 81s | 87s | 76.7s | 70.4s |

- BoringSSL: about 25s more per from-source build (paid once per bump, then cached) for about 7s less pytest on every run of both jobs.
- AWS-LC: build cost goes up similarly, pytest is within noise (the two main-side numbers for the same test suite differ by 28s between themselves). Its slowest tests are the DH ones at 6–7s each, and they are just as slow at -O3, so that time is not optimization-bound.

---

PR #15572, opened 2026-09-02
Use uv-managed Python for the CI jobs

reaperhulk:
`actions/setup-python` on macOS downloads the python.org pkg and runs the installer, and the "Install Python binaries from prebuilt package" step takes 30–50s per job. On Windows the download and extraction is similarly the slowest part of the job. `astral-sh/setup-uv` installs a python-build-standalone interpreter in a few seconds.

Applies to every job that used `actions/setup-python` in `ci.yml`: linux, macOS, windows (x64 and arm64) and downstream.

- The uv version comes from `ci-constraints-requirements.txt` via `version-file`, so it stays in lockstep with the pinned `uv==` that dependabot bumps.
- `activate-environment` puts a venv on that interpreter first on `PATH`; nox is installed into it with `uv pip install`.
- The matrix names use uv's spellings: `3.15` / `3.15t` (uv resolves an unreleased minor to its latest prerelease) and `pypy3.11`.
- The cargo cache keys use setup-uv v10.1.0's `python-runtime-id` output (e.g. `cpython-3.15.0rc2-linux-x86_64-gnu`), which carries the implementation, resolved version and architecture, so the matrix name and arch components of the keys and the extra step that read the version from the interpreter go away.
- On Windows arm64 a bare version request gets uv the x86_64 (emulated) interpreter, so that job asks for `cpython-<version>-windows-aarch64-none` by name.
- The linux job puts the interpreter's `LIBDIR` on `LD_LIBRARY_PATH` for the rust session, whose `cargo test` binaries link libpython.
- The uv cache replaces setup-python's pip cache, keyed on the same constraints file.

The trade-off is that these jobs no longer exercise the python.org framework build on macOS; the wheel builder still installs and uses the python.org pkg.


reaperhulk:
Timing for the macOS jobs. "Before" is the two most recent warm main runs (33562151532 and the re-run of 33588967176); "after" is this PR's run (33589841415). The setup step is the direct effect; the build and test steps are unchanged by this PR and vary ±10s run to run.

| Job | Setup python, before | Setup python, after | Job total, before | Job total, after |
|---|---|---|---|---|
| 3.9 | 25s / 30s | 3s | 156s / 146s | 137s |
| 3.14 | 3s / 4s | 2s | 142s / 115s | 158s |
| 3.14t | 57s / 51s | 6s | 215s / 188s | 148s |
| 3.15 (was 3.15-dev) | 50s / 45s | 4s | 215s / 198s | 115s |
| 3.15t (was 3.15t-dev) | 36s / 32s | 9s | 173s / 133s | 167s |

3.14 was already fast because that version is preinstalled in the runner image's toolcache; every other version paid for the pkg installer. Summed over the five jobs the setup step goes from about 170s to about 24s per run. The 3.15 / 3.15t caches were cold in this run because their key changed (matrix name and resolved version), so their build and total numbers will settle on the next run.

---

.github/workflows/ci.yml
```diff
@@ -271,22 +271,26 @@ jobs:
...
+          - {VERSION: "3.15", NOXSESSION: "tests"}
+          - {VERSION: "3.15t", NOXSESSION: "tests"}
     timeout-minutes: 15
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         timeout-minutes: 3
         with:
           persist-credentials: false
+      # uv-managed Python rather than setup-python: the python.org pkg
+      # installer setup-python runs takes 30-50s per job.
```
alex: Comments should not reference past state

.github/workflows/ci.yml
```diff
@@ -271,22 +271,26 @@ jobs:
           - {VERSION: "3.9", NOXSESSION: "tests"}
           - {VERSION: "3.14", NOXSESSION: "tests"}
           - {VERSION: "3.14t", NOXSESSION: "tests"}
-          - {VERSION: "3.15-dev", NOXSESSION: "tests"}
-          - {VERSION: "3.15t-dev", NOXSESSION: "tests"}
+          - {VERSION: "3.15", NOXSESSION: "tests"}
+          - {VERSION: "3.15t", NOXSESSION: "tests"}
```
alex, line 279: -dev isn't required for pre-releases?

.github/workflows/ci.yml
```diff
@@ -271,22 +271,26 @@ jobs:
...
         with:
+          version-file: ci-constraints-requirements.txt
           python-version: ${{ matrix.PYTHON.VERSION }}
-          cache: pip
-          cache-dependency-path: ci-constraints-requirements.txt
+          activate-environment: true
+          cache-dependency-glob: ci-constraints-requirements.txt
         timeout-minutes: 3
+      - id: python-version
+        run: echo "version=$(python -c 'import platform; print(platform.python_version())')" >> $GITHUB_OUTPUT
```
alex, line 295: @woodruffw can you give us this as an output :-) Surely setup-uv already has this value internally

alex (commented):
@reaperhulk let's chat about this one

.github/workflows/ci.yml
reaperhulk: Right: uv has no `-dev` spelling (`3.15-dev` is rejected as an executable name), and a bare minor resolves to the latest prerelease when no stable release exists. Checked with the pinned uv 0.12.8: `uv python install 3.15` gives `cpython-3.15.0rc1`, `3.15t` gives `cpython-3.15.0rc1+freethreaded`, `3.14t` gives `3.14.7+freethreaded`. The run on this PR used those.

---

.github/workflows/ci.yml
```diff
@@ -296,11 +298,11 @@ jobs:
...
-      # windows jobs), not just the matrix name: the alpha that a -dev
-      # entry resolves to moves over time, and a cached maturin/pyo3
-      # config built by an older alpha gets rewritten by maturin on every
-      # run, which re-dirties pyo3-ffi and recompiles it, pyo3, and every
-      # crate above them despite a warm cache.
+      # windows jobs), not just the matrix name: the prerelease that an
+      # unreleased minor resolves to moves over time, and a cached
+      # maturin/pyo3 config built by an older one gets rewritten by
+      # maturin on every run, which re-dirties pyo3-ffi and recompiles it,
+      # pyo3, and every crate above them despite a warm cache.
```
alex: why did this get changed? this appears to be pure churn

alex:
Notes:
1. We want a uv fix for the output stuff
2. We'd lke but won't block on a -dev thing from uv
3. We want this for all OS

.github/workflows/ci.yml
reaperhulk: Only because the `-dev` spelling left this matrix; the sentence still reads correctly since it describes the linux and windows entries. Restored the original text.

---

.github/workflows/ci.yml
woodruffw: Yeah, I can look into this. I thought we did have this as a version output.

reaperhulk: In case it helps: in v10.0.1 the `python-version` output is the input string echoed back (`getPythonVersion` in `src/setup-uv.ts` returns `inputs.pythonVersion` when set), so `3.15t` comes out as `3.15t` rather than `3.15.0rc1`. `uv python find --show-version` after the venv is created would give the resolved value.

---

reaperhulk:
Extended to the linux, windows x64 and downstream jobs (c4c8e9f, plus 6932c44 which puts the interpreter's `LIBDIR` on `LD_LIBRARY_PATH` for the rust session: its `cargo test` binaries link libpython, which setup-python had been making findable implicitly). Windows arm64 stays on setup-python because uv has no Windows arm64 interpreters, so its matrix entries keep the `-dev` spelling.

Setup step times, last warm main run (33588967176) versus this head (33695131695); for Windows the whole `windows-tests` action, since the composite hides its steps:

| Job group | Before, median (max) | After, median (max) |
|---|---|---|
| linux "Setup python", 30 jobs | 5s (13s) | 3s (5s) |
| macos "Setup python", 5 jobs | 32s (51s) | 7s (11s) |
| windows x64, whole action, 5 jobs | 218s (241s) | 177s (185s) |
| downstream "Setup python", 10 jobs | 4.5s (7s) | 3s (3s) |

Linux was already fast because the runner image preinstalls those versions; macOS and Windows x64 are where the time was.

---

.github/actions/windows-tests/action.yml
```diff
@@ -21,14 +21,26 @@ runs:
   using: "composite"
 
   steps:
+    # uv has no Windows arm64 interpreters, so that job uses setup-python.
```
reaperhulk: This appears to be untrue, see: https://github.com/astral-sh/uv/releases/tag/0.12.9

reaperhulk: You're right, my check was wrong: `uv python list --all-platforms` only shows x86_64 for Windows, but asking for `cpython-3.14-windows-aarch64-none` explicitly finds it, and 3.14, 3.14t, 3.15 and 3.15t are all available for windows-aarch64 in both 0.12.8 and 0.12.9. ede2731 moves the arm64 job to setup-uv too and drops the setup-python branch entirely.

---

.github/workflows/ci.yml
eifinger: @alex https://github.com/astral-sh/setup-uv/releases/tag/v10.1.0 now has the output `python-runtime-id`

alex: Claude: you know what to do

reaperhulk: Done in 1706ca9: bumped to setup-uv v10.1.0 and the three cache keys now use `python-runtime-id` (e.g. `cpython-3.15.0rc2-linux-x86_64-gnu`), so the interpreter-probing steps are gone. Since the id also carries the implementation and architecture, the matrix name and arch components of the keys went with them. Also merged main.

The Windows arm64 job keeps asking for `cpython-<version>-windows-aarch64-none` explicitly: the run on ede2731 shows the aarch64 uv 0.12.9 downloading `cpython-3.15.0rc2-windows-x86_64-none` for a bare `3.15` request, and nothing in the uv changelog since has changed that.

---

alex (approved):

PR #15584, opened 2026-09-03
remove unused dep

alex:


alex:
ugh, this is in fact required, but causes rustc to complain. will investigate.

PR #15592, opened 2026-09-05
Implement Fernet in Rust

alex:
Moves `Fernet` and `MultiFernet` from Python to a Rust extension module. `cryptography.fernet` is now a thin shim that defines `InvalidToken` and re-exports the Rust classes. The public API is unchanged, including the private helpers the tests use (`_encrypt_from_parts`, `_get_unverified_token_data`).

Notable details:

* Each `Fernet` holds pre-keyed HMAC and AES-CBC contexts that are copied per operation, which is much cheaper than re-keying.
* Base64 decoding uses the `base64` crate's URL-safe alphabet with padding optional. CPython's decoder laxities (ignoring whitespace and other non-alphabet bytes, accepting the standard alphabet) are not replicated.
* `int(time.time())` is still obtained by calling into Python so that patching `time.time` keeps working.
* Inputs of 2 KiB and above release the GIL during encryption/decryption.
* `encrypt()` now accepts any bytes-like object. `MultiFernet` takes a sequence rather than an arbitrary iterable and type-checks its elements at construction.

Rough speedups over the Python implementation (OpenSSL 3.0, Python 3.11): 16-byte encrypt 4.5x, 16-byte decrypt 8x, 1 MiB encrypt/decrypt about 2x, `MultiFernet.rotate` 4x. `Fernet(key)` construction itself is slower (about 3 µs vs 1.3 µs) because it keys the contexts up front.

https://claude.ai/code/session_01XDr9nsqX5Vzcz6vmGyfqTx


reaperhulk:
changelog conflict, needs a rebase.

alex:
Rebased onto main. The changelog entry is gone entirely (it was supposed to be dropped in an earlier round but the revert hadn't actually been committed), so the PR no longer touches `CHANGELOG.rst`.

---

reaperhulk (approved):

PR #15593, opened 2026-09-05
use python 3.15.0 final release in wheel builder workflow

mayeut:
Follow up on https://github.com/pyca/cryptography/pull/15496#discussion_r3836335992

Needs Python 3.15.0 final release to be available which is planned in October.


alex:
Thanks for sending this as well!

PR #15621, opened 2026-09-10
Use uv-managed Python for the wheel builder

reaperhulk:
Follow-up to #15572, applying the same change to `wheel-builder.yml`: the sdist, macOS and Windows jobs get their interpreter from `astral-sh/setup-uv` instead of `actions/setup-python` or the python.org pkg installer.

- The uv version comes from `ci-constraints-requirements.txt` via `version-file`, as in `ci.yml`. The separate `.github/requirements/uv-requirements.{in,txt}` pair is deleted: dependabot only maintains the pin in `ci-constraints-requirements.txt`, so that pair had sat at 0.12.0, and nothing else consumed it. (This matters more than it used to: uv's interpreter manifest is compiled into the binary, so a stale uv means stale interpreters, which is what the first run here showed.)
- The macOS matrix shrinks to version and ABI: the download URLs, framework paths, PATH additions and the two installer choice XML files go away, and the deployment target and arch flags become literals on the build step. The abi3-py39 entry was labelled `3.11` but built with the python.org 3.14 pkg; it is now labelled and built with 3.14, like Windows.
- The Windows matrix drops the `freethreaded` flag and uses uv's spellings (`3.14t`, `pypy3.11`). The arm64 jobs ask for `cpython-<version>-windows-aarch64-none` by name, since a bare request gets uv the x86_64 build.
- The 3.15 free-threaded wheels are pinned to `3.15.0rc2t` by name on both platforms, as the pkg URL and setup-python version were before, so a stale manifest fails the build rather than quietly picking an older prerelease.
- The three jobs set `UV_PYTHON_PREFERENCE=only-managed`. In the #15572 runs a bare `3.14` made uv pick Homebrew's python3.14 on macOS and the toolcache 3.14.7 on Windows, which is fine for tests but not what a release build should depend on.
- No caching in these jobs (`enable-cache: false`), matching the previous behaviour.

macOS wheels are now built against python-build-standalone rather than the python.org framework build. The extension doesn't link libpython on macOS, and the build step still sets `MACOSX_DEPLOYMENT_TARGET=11.0` and `_PYTHON_HOST_PLATFORM=macosx-11.0-arm64`. The first run confirms the wheel tag (`macosx_11_0_arm64`) and `vtool` minimum OS (`minos 11.0`) are unchanged.

One trade-off: a uv bump no longer triggers the wheel builder on its PR, since `ci-constraints-requirements.txt` isn't in its `paths` filter and adding it would run the builder for every CI dependency bump.

After this, `actions/setup-python` remains only in `benchmark.yml`, `linkcheck.yml` and `pypi-publish.yml`.


reaperhulk:
The first run (d4fede7) was green everywhere, but the logs showed the 3.15t wheels built against `cpython-3.15.0b4+freethreaded` and the 3.14 wheels against 3.14.6, where the old config pinned the rc2 pkg and setup-python picked 3.14.7. Cause: uv's interpreter manifest is compiled into the binary, and `uv-requirements.txt` was still at 0.12.0 because dependabot only bumps the pin in `ci-constraints-requirements.txt`.

2f92580 takes the uv version from `ci-constraints-requirements.txt` instead (as `ci.yml` does), deletes the unmaintained `uv-requirements.{in,txt}` pair, and pins the 3.15 free-threaded wheels to `3.15.0rc2t` by name so a stale manifest fails loudly rather than picking an older prerelease. Everything else from the first run checked out: `macosx_11_0_arm64` tags, `minos 11.0` from vtool, `win_arm64` from the aarch64 interpreter.

---

alex:
please report the performance impact

reaperhulk:
Performance impact. "Before" is the last two green wheel-builder runs on the old config (33717972399 and 33838605550, Sep 3–4, same matrix); "after" is this PR's two runs (34529225558 and 34530402096). Setup = the Python install steps plus the pip install of uv (before) or the setup-uv step (after). Windows shows only job totals because the composite action hides its steps.

| Job | Setup before | Setup after | Job total before | Job total after |
|---|---|---|---|---|
| macOS 3.14 abi3-py39 (was labelled 3.11) | 23s / 15s | 1s / 0s | 119s / 94s | 84s / 95s |
| macOS 3.14 abi3-py311 | 15s / 16s | 0s / 1s | 76s / 84s | 68s / 63s |
| macOS 3.14t | 33s / 38s | 0s / 1s | 109s / 133s | 112s / 106s |
| macOS 3.15t | 33s / 28s | 1s / 1s | 109s / 94s | 83s / 106s |
| macOS pypy3.11 | 15s / 18s | 0s / 1s | 111s / 121s | 65s / 97s |
| win64 3.14 abi3-py39 | | | 133s / 145s | 145s / 185s |
| win64 3.14 abi3-py311 | | | 140s / 138s | 174s / 210s |
| win64 3.14t | | | 180s / 207s | 157s / 217s |
| win64 3.15t | | | 197s / 190s | 142s / 162s |
| win64 pypy3.11 | | | 151s / 160s | 138s / 167s |
| arm64 3.14 abi3-py311 | | | 143s / 154s | 155s / 158s |
| arm64 3.14t | | | 229s / 266s | 154s / 262s |
| arm64 3.15t | | | 222s / 247s | 167s / 164s |
| sdists | 3s / 2s | 1s / 1s | 21s / 22s | 28s / 21s |

macOS is where the time was: the pkg installer cost 15–38s per job and setup-uv takes about 1s. The interpreter download now happens inside `uv build` (2–5s, it's a 20MB tarball), so about 15–30s net per macOS job, roughly 120s summed over the five.

Windows is a wash, as expected: setup-python was already instant from the toolcache and the pip install of uv took about 5s, versus 2s for setup-uv plus a 2s download. The win64 3.14 jobs look slower in the table, so I compared the step timelines for `3.14 win64 abi3-py311`: the cargo compile itself is 69s / 74s / 77s across the three runs, and the difference is entirely in cargo's "Downloading crates" phase, 9s before versus 37s and 70s in the two PR runs. That's crates.io on the Windows runner, not this change.

---

alex (approved):

PR #15622, opened 2026-09-11
Use uv-managed Python for the remaining workflows

reaperhulk:
Follow-up to #15572 and #15621. `benchmark.yml`, `linkcheck.yml` and `pypi-publish.yml` were the last users of `actions/setup-python`; after this there are none. All three get uv from `astral-sh/setup-uv` with the version pinned by `ci-constraints-requirements.txt`.

- **benchmark**: setup-uv with Python 3.11 replaces setup-python plus the pip install of uv. The comparison script runs on the PR venv's interpreter instead of whatever `python` is on PATH, since nothing puts a specific one there any more.
- **linkcheck**: mirrors the `ci.yml` linux job (activated venv, nox via `uv pip install`, cache keyed on `python-runtime-id`). The comment saying its cache key matched the docs job's is dropped: docs runs on 3.12 with an OpenSSL component in its key, and rust-cache adds the job id regardless, so they haven't shared a cache for a while.
- **pypi-publish**: it only needed Python to pip-install uv for `uv publish`, so it now installs uv directly after the sparse checkout and sets up no interpreter at all. This one can't be exercised on a PR (it runs on wheel-builder completion or manual dispatch); the publish step itself is untouched.

benchmark and linkcheck both trigger on changes to their own file, so they run here.

---
_Generated by [Claude Code](https://claude.ai/code/session_0116sWJmeCaubZF2aMbR4kfc)_

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -43,13 +43,12 @@ jobs:
         uses: ./cryptography-base/.github/actions/fetch-vectors
 
       - name: Setup python
-        id: setup-python
-        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        uses: astral-sh/setup-uv@bec219d24cd3e171d82865faccec33120bb574f4 # v10.1.0
         with:
+          version-file: cryptography-pr/ci-constraints-requirements.txt
           python-version: "3.11"
+          enable-cache: false
 
-      - name: Install uv
-        run: pip install -c ./cryptography-pr/ci-constraints-requirements.txt uv
       - name: Create virtualenv (base)
         run: |
           uv venv .venv-base
@@ -67,4 +66,4 @@ jobs:
         run: .venv-pr/bin/pytest --benchmark-enable --benchmark-only ./cryptography-pr/tests/bench/ --benchmark-json=bench-pr.json --x509-limbo-root=x509-limbo/
 
       - name: Compare results
-        run: python ./cryptography-pr/.github/bin/compare_benchmarks.py bench-base.json bench-pr.json | tee -a $GITHUB_STEP_SUMMARY
+        run: .venv-pr/bin/python ./cryptography-pr/.github/bin/compare_benchmarks.py bench-base.json bench-pr.json | tee -a $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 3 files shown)


reaperhulk:
The `linkcheck` check is red here, but not because of this PR. The converted job got all the way through: setup-uv, the venv, building the extension and running sphinx's linkcheck. It then failed on one broken link, a 403 from `https://digitalcommons.unl.edu/cgi/viewcontent.cgi?article=1100&context=cseconfwork` (cited from `docs/hazmat/primitives/asymmetric/ec.rst` as "elliptic curve diffie-hellman is faster than diffie-hellman"). The scheduled linkcheck on main has failed on exactly that link for the last four weekly runs, most recently 34544896952 an hour before this PR was opened, so a re-run won't change it.

The PDF endpoint returns 403 to everything, browser user agents included, but the paper's landing page is fine. Proposed fix, kept out of this PR since it's a docs change:

```diff
--- a/docs/hazmat/primitives/asymmetric/ec.rst
+++ b/docs/hazmat/primitives/asymmetric/ec.rst
-.. _`elliptic curve diffie-hellman is faster than diffie-hellman`: https://digitalcommons.unl.edu/cgi/viewcontent.cgi?article=1100&context=cseconfwork
+.. _`elliptic curve diffie-hellman is faster than diffie-hellman`: https://digitalcommons.unl.edu/cseconfwork/105/
```

That's the landing page for the same paper ("The Performance of Elliptic Curve Based Group Diffie-Hellman Protocols for Secure Group Communication over Ad Hoc Networks"); its `citation_pdf_url` is the article=1100 URL.

---

reaperhulk:
The docs fix for the broken link is now #15623.

---

reaperhulk:
Rebased onto main with #15623; the digitalcommons link now reports `ok`. `linkcheck` is still red on 87cd2f6, this time on two `mail.python.org` links (`mailman/listinfo/cryptography-dev` and `mailman3/lists/python-announce-list.python.org/`) hitting sphinx's 5s read timeout. Both are fine on main's scheduled run from two hours ago; right now the site takes 6–16s to first byte from here too, so it's mail.python.org being slow rather than anything in this PR. I'll re-run the job once it responds normally. If it keeps recurring, bumping `linkcheck_timeout` in `docs/conf.py` would be the durable fix, separate from this PR.

---

alex (approved):

PR #15628, opened 2026-09-11
Switch downstream tests to Ubuntu 26.04

alex:
Pins the `linux-downstream` job's runner to `ubuntu-26.04` instead of `ubuntu-latest`.

Ubuntu 26.04 ships GCC 15, which turns the implicit declaration of `_PyErr_WriteUnraisableMsg` in cffi 1.15.1's C backend from a warning into an error, so the `cffi==1.15.1` pin that aws-encryption-sdk-python's test requirements used to carry no longer built for Python 3.14. That pin was fixed upstream in aws/aws-encryption-sdk-python#818, so this also bumps the aws-encryption-sdk-python downstream ref to the commit carrying that fix (the same bump as in #15627).

https://claude.ai/code/session_01EkX3LD8jYgtCyGDVRUq6cr

### DIFF
--- a/.github/downstream.d/aws-encryption-sdk-python.sh
+++ b/.github/downstream.d/aws-encryption-sdk-python.sh
@@ -4,7 +4,9 @@ case "${1}" in
     install)
         cd aws-encryption-sdk-python
         uv pip install -e .
-        uv pip install -r test/upstream-requirements-py311.txt
+        # cffi 1.15.1 cannot be built for Python 3.14 with GCC 14+, and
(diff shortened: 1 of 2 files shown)


alex:
depends on https://github.com/aws/aws-encryption-sdk-python/pull/818

alex:
Please remove your workaround and then grab just the aws-encryption-sdk bump from https://github.com/pyca/cryptography/pull/15627

reaperhulk (approved):

Issue #15603, opened 2026-09-07
Consider FFDH for decrepit — Apple Screen Sharing/ARD has no alternative


sibson:
Apple's Screen Sharing / Remote Desktop (ARD) authentication, part of the RFB/VNC protocol family, mandates classic finite-field Diffie-Hellman for its key exchange — there is no alternative to negotiate into, and Apple controls the protocol, not us. A client that wants to authenticate against ARD has no path other than FFDH.

Our usage: [`vncdotool/rfb.py`](https://github.com/sibson/vncdotool/blob/309cf0db2f14e6055148e50de28116245d3daa67/vncdotool/rfb.py#L221-L235) (`_encryptArd`) uses `dh.DHParameterNumbers`/`dh.DHPublicNumbers` for exactly this. We're tracking the deprecation on our end (sibson/vncdotool#388).

Following up on #15171, which deprecated FFDH in 50.0.0. @alex noted there:

> We weren't planning to migrate to decrepit, just deprecate (for a long time) and then remove.

That thread had no concrete interoperability case on the table. `decrepit` exists precisely for this: CAST5/SEED/IDEA/Blowfish went there in #6809, and 3DES/ARC4 were kept out of deprecation entirely, "due to their prevalence in existing systems." `decrepit`'s own docs: "algorithms that should not be used unless necessary for backwards compatibility or interoperability with legacy systems."

Given there's no migration target for a protocol that mandates FFDH by spec, please reconsider `decrepit` as FFDH's eventual home instead of outright removal.

reaperhulk:
Thanks for letting us know, we definitely want to understand use cases like this. This feels like a scenario where talking to SEAR folks might be helpful.

alex:
If I understand https://support.apple.com/guide/remote-desktop/encrypt-network-data-apdfe8e386b/mac correctly (which I may not, it's not super clearly written), it looks like more recent macOS support non-FFDH key exchange.

emillysilva65:
Regarding the Apple documentation Alex linked, ARD's "Encrypt Network Data" option strictly mandates the specific FFDH implementation currently used in `_encryptArd` for older but still pervasive deployments. Moving FFDH to `decrepit` preserves the interop path for these legacy macOS environments while maintaining the deprecation signal for modern consumers.

PR #15643, opened 2026-09-15
Move FFDH to decrepit

reaperhulk:
Addresses #15603.

Diffie-Hellman over finite fields (FFDH) was deprecated in 50.0.0 on the way to removal, but some protocols mandate it with no alternative to negotiate into. This moves it into `cryptography.hazmat.decrepit.asymmetric.dh`, following the pattern used for the decrepit ciphers and modes, and keeps every existing deprecation warning with reworded text: FFDH is deprecated and will only be available from the decrepit module starting in 53.0.0.

## What moved

- `cryptography.hazmat.decrepit.asymmetric.dh` now holds `generate_parameters`, the `DHParameters` / `DHPrivateKey` / `DHPublicKey` ABCs, the three numbers classes, and `load_pem_parameters` / `load_der_parameters` (these can only load FFDH parameters, so they belong with it). Nothing in the decrepit module warns.
- `cryptography.hazmat.primitives.asymmetric.dh` becomes a shim that re-exports the same objects via `utils.deprecated`, so `isinstance` checks against either path keep working. The warning class is `DeprecatedIn51` with removal targeted at 53.0.0, matching the two-release window.
- `serialization.load_pem_parameters` / `load_der_parameters` stay deprecated, with the "has been moved to ... and will be removed from ... in 53.0.0" wording the Camellia and CFB/OFB moves used.
- The Rust warning emitted when the generic key loading APIs (and X.509 `public_key()`) hand back a DH key is kept and reworded the same way. `DEPRECATED_IN_50` in `types.rs` was only used there and is removed.
- `asymmetric.types` now takes the DH types from the decrepit module instead of the `_DHPublicKey` / `_DHPrivateKey` private aliases, which are dropped.

## Bug fixed along the way

The old shim (and `dsa.py` today, not touched here) called `utils.deprecated(DHParameters, ..., name="DHParametersWithSerialization")` after `utils.deprecated(DHParameters, ..., name="DHParameters")` had already replaced the module global with the `_DeprecatedValue` marker, so the `...WithSerialization` aliases resolved to the marker object instead of the class. The new shim reads every value from the decrepit module directly, and the new test covers all ten names.

## Docs and tests

- New `docs/hazmat/decrepit/dh.rst` documents the decrepit module (with `versionadded:: 51.0.0` markers and doctests for the parameter loaders). The old `primitives/asymmetric/dh.rst` keeps its content with a "moved to decrepit" warning, as the cipher and mode pages do, so older changelog cross-references still resolve under nitpicky mode.
- Tests import from the decrepit module and no longer need to suppress warnings when building fixtures. `TestFFDHDeprecation` checks the old module names, the serialization loaders, the generic key loaders, and that the decrepit path is warning-free.

## One design point for reviewers

Loading FFDH keys through `load_pem_private_key` and friends still warns, as requested ("keep the warnings"). The decrepit module deliberately does not grow its own PEM/DER key loaders in this PR, so if the intent for 53.0.0 is that the generic loaders stop returning DH keys, a follow-up would need to add those; if instead the generic loaders should keep working and return the decrepit types, the Rust warning can simply be removed later.

`nox -e local` passes in full. `nox -e docs` builds cleanly under `-W`; the only doctest failures locally are in unrelated pages (Argon2, AES-GCM-SIV, SHAKE `squeeze`) because the container's system OpenSSL 3.0.13 lacks them.

https://claude.ai/code/session_01TmnZQpY2o4HeBNFN5tfqYE

---
_Generated by [Claude Code](https://claude.ai/code/session_01TmnZQpY2o4HeBNFN5tfqYE)_


docs/hazmat/primitives/asymmetric/dh.rst
```diff
@@ -5,15 +5,16 @@ Diffie-Hellman key exchange
...
-    where possible.
+.. warning::
+
+    Diffie-Hellman over finite fields (FFDH) has been deprecated and moved to
+    the :doc:`/hazmat/decrepit/index` module as
+    :mod:`cryptography.hazmat.decrepit.asymmetric.dh`. If you need to
+    continue using it then update your code to use the new module path.
+    Starting in 53.0.0 it will only be available from that module. Users
+    should migrate to a more modern key exchange algorithm such as
+    :class:`~cryptography.hazmat.primitives.asymmetric.ec.ECDH` where
```
reaperhulk: We should suggest MLKEM too

reaperhulk: Done in b984baf: the warning here and the intro on the new decrepit page now suggest ECDH, X25519, or ML-KEM (`MLKEM768PrivateKey`).

---

docs/hazmat/decrepit/dh.rst
```diff
@@ -0,0 +1,426 @@
+.. hazmat::
```
alex, line 1: This appears to full on duplicate the docs, I do not like that.

src/cryptography/hazmat/primitives/asymmetric/dh.py
```diff
@@ -4,250 +4,95 @@
...
-from cryptography.hazmat.primitives import _serialization
+from cryptography.hazmat.decrepit.asymmetric import dh as _decrepit_dh
 
 _FFDH_DEPRECATION_MSG = (
-    "Diffie-Hellman over finite fields (FFDH) is deprecated and support "
-    "will be removed in a future release. Use a more modern key exchange "
-    "algorithm."
+    "Diffie-Hellman over finite fields (FFDH) is deprecated and has been "
+    "moved to cryptography.hazmat.decrepit.asymmetric.dh. Starting in "
+    "53.0.0 it will only be available from that module."
```
alex, line 13: Ugh, I really feel like this is moving backwards on the deprecation, and therefore suggests we're prepared to carry it forever, which I really don't want to be.

docs/hazmat/decrepit/dh.rst
reaperhulk: Agreed, de-duplicated. The reference now lives only on the decrepit page; `primitives/asymmetric/dh.rst` is a short stub with the moved warning, and the historical changelog cross-references point at the decrepit page so the nitpicky build still resolves them. The full copy had only been kept for those changelog links.

---

PR #15656, opened 2026-09-17
Fix *_into methods producing incorrect output for overlapping buffers

alex:
Several `*_into` methods handed the caller's input and output buffers to the underlying implementation without considering whether they shared memory. When they did, the output could overwrite input that hadn't been read yet, silently producing incorrect results:

- **AEAD `encrypt_into`/`decrypt_into`** (all six classes): `buf` starting inside `data` (e.g. `encrypt_into(nonce, mv[:n], ad, mv[12:])`) produced wrong ciphertext with a *valid* tag, or, for `ChaCha20Poly1305.decrypt_into`, "authenticated" wrong plaintext. For the tag-first `AESSIV`, the obvious in-place layout with the plaintext at the front of `buf` was enough to trigger it.
- **`CipherContext.update_into`** in place with a block mode (CBC/ECB/...) silently corrupted output as soon as an earlier call left a partial block pending inside OpenSSL.
- **Cobblestone `update_into`** in place wrote the 56-byte header over the plaintext before reading it (and clobbered plaintext inside each chunk for messages ≥ 16 KiB), producing a valid ciphertext of the wrong message.
- **`ConcatKDFHash`/`ConcatKDFHMAC`/`X963KDF.derive_into`** with `key_material` aliasing `buffer` and `length > digest_size` hashed already-written output blocks for every block after the first.

Fix: every `*_into` method (AEAD `encrypt_into`/`decrypt_into`, `CipherContext.update_into`, Cobblestone `update_into`, and all KDF `derive_into` methods) now requires the input and output buffers to be disjoint and raises `ValueError("data and buf must not overlap")` otherwise, via a single `check_no_overlap` helper in `buf.rs`. No in-place special cases. The OpenSSH private key serializer, which encrypted in place, now passes a copy of the secrets section.

Tests assert overlap rejection (in place and shifted) for ChaCha20Poly1305/AESGCM/AESSIV, CBC and CTR, Cobblestone encrypt/decrypt, and ConcatKDFHash/ConcatKDFHMAC/X963KDF.

Discovered by Anthropic's OSS Scanner, as vulnerabilities ANT-2026-QFHQ4A4C, ANT-2026-ECRME9NI, ANT-2026-BF6D5V9V, ANT-2026-UDR61OGV.


alex:
We should be, across the board, rejecting _any_ overlaps between in/out, no special cases or anything like that.

alex:
Done in 9e06959: every `*_into` method (AEAD, `CipherContext.update_into`, Cobblestone, and all KDF `derive_into`) now rejects any overlap between input and output with `ValueError`, with no in-place special cases or copies. The OpenSSH serializer, which encrypted in place, now passes a copy of the secrets section.

---

CHANGELOG.rst
```diff
@@ -16,6 +16,14 @@ Changelog
   the key loading APIs (including from X.509 certificates and certificate
   signing requests). Users should migrate to a more modern signature
   algorithm.
+* The ``encrypt_into``/``decrypt_into`` methods in
```
alex: Don't enumerate everything, just describe the category of methods.

reaperhulk (approved):

PR #15658, opened 2026-09-17
Raise ValueError when an SCT list is too large to encode

alex:
`encode_scts` wrote the RFC 6962 `SignedCertificateTimestampList` length prefix with a truncating `length as u16` cast. An extension whose serialized SCTs total more than 65535 bytes (roughly 550 typical SCTs) was therefore silently encoded with a length wrapped modulo 65536: `public_bytes()` and `CertificateBuilder.sign()` succeeded, but the result is rejected by our own parser (`ValueError: Invalid SCT length`) and by OpenSSL.

A list that size has no valid encoding (`sct_list<1..2^16-1>`), so the list length now goes through `u16::try_from` and raises `ValueError`. The per-entry length is bounded by that check and uses `try_into().unwrap()`. The regression test repeats the SCT from `badssl-sct.pem` 1000 times and checks that `public_bytes()` raises for both extension classes.

Discovered by Anthropic's OSS Scanner, as vulnerability ANT-2026-H08T92Y2.


tests/x509/test_x509_ext.py
```diff
@@ -6137,6 +6137,57 @@ def test_init(self):
...
+        entry_len = (
+            len(
+                x509.PrecertificateSignedCertificateTimestamps(
+                    [sct]
+                ).public_bytes()
+            )
+            - 4
+        )
+        n = 65535 // entry_len
+        assert n * entry_len <= 65535 < (n + 1) * entry_len
```
alex: This assertion looks pointless. You can just use a large value of N for the below things, it doesn't have to be precisely sized.

tests/x509/test_x509_ext.py
```diff
@@ -6137,6 +6137,57 @@ def test_init(self):
...
+            == ext
+        )
+
+        with pytest.raises(ValueError, match="too large"):
+            x509.PrecertificateSignedCertificateTimestamps(
+                [sct] * (n + 1)
+            ).public_bytes()
+        with pytest.raises(ValueError, match="too large"):
+            x509.SignedCertificateTimestamps([sct] * (n + 1)).public_bytes()
+        with pytest.raises(ValueError, match="too large"):
```
alex: you don't need to create a cert, public_bytes assertions were sufficient

CHANGELOG.rst
```diff
@@ -16,6 +16,11 @@ Changelog
   the key loading APIs (including from X.509 certificates and certificate
   signing requests). Users should migrate to a more modern signature
   algorithm.
+* Serializing a
```
alex: no changelog

src/rust/src/x509/extensions.rs
```diff
@@ -415,12 +415,8 @@ fn encode_scts(ext: &pyo3::Bound<'_, pyo3::PyAny>) -> CryptographyResult<Vec<u8>
...
     for sct in ext.try_iter()? {
         let sct = sct?.cast::<sct::Sct>()?.clone();
-        let sct_length = u16::try_from(sct.get().sct_data.len()).map_err(|_| {
-            pyo3::exceptions::PyValueError::new_err(
-                "SCT is too large to encode (must be at most 65535 bytes)",
-            )
-        })?;
-        result.extend_from_slice(&sct_length.to_be_bytes());
+        // Each entry's length is bounded by the check on the total above.
+        result.extend_from_slice(&(sct.get().sct_data.len() as u16).to_be_bytes());
```
alex: use try_into().unwrap()

reaperhulk (approved):

PR #15659, opened 2026-09-17
asn1: make SetOf equality independent of element order

alex:
`SET OF` is unordered and the encoder sorts its elements, so `SetOf([3, 1, 2])` and `SetOf([1, 2, 3])` are the same ASN.1 value with the same DER. `SetOf.__eq__` compared the backing lists positionally, though, and both halves of the `DEFAULT` handling (`encode.rs`: omit the field if `value == default`; `decode.rs`: reject if `decoded == default`) rely on Python equality. As a result, with `Default(SetOf([1, 2, 3]))`, encoding `SetOf([3, 1, 2])` explicitly wrote out the DEFAULT (not valid DER, and rejected by our own `decode_der`), and with `Default(SetOf([3, 1, 2]))`, `decode_der` accepted an explicitly encoded DEFAULT.

This makes `SetOf.__eq__` compare as a multiset, pairing elements up with `==` so they needn't be hashable or orderable (quadratic, but these are small). The DEFAULT code is unchanged and becomes correct as a result, including for a `SetOf` nested in a `list[...]` or `SetOf[...]` default. The user-visible change is that `SetOf([1, 2]) == SetOf([2, 1])` is now `True`, which is noted in the reference docs.

Discovered by Anthropic's OSS Scanner, as vulnerability ANT-2026-TL4ENT1R.


tests/hazmat/asn1/test_serialization.py
```diff
@@ -1214,6 +1214,43 @@ def test_ok_decode_setof(self) -> None:
...
+        @asn1.sequence
+        class Example:
+            values: Annotated[
+                asn1.SetOf[int], asn1.Default(asn1.SetOf([3, 1, 2]))
+            ]
+            data: bytes
+
+        # A value equal to the DEFAULT is omitted, whatever order its
+        # elements are listed in.
+        for values in ([3, 1, 2], [1, 2, 3], [2, 3, 1]):
```
alex: Just use the table test helpers.

CHANGELOG.rst
```diff
@@ -16,6 +16,12 @@ Changelog
   the key loading APIs (including from X.509 certificates and certificate
   signing requests). Users should migrate to a more modern signature
   algorithm.
+* :class:`~cryptography.hazmat.asn1.SetOf` equality no longer depends on
```
alex: no changelog

tests/hazmat/asn1/test_serialization.py
```diff
@@ -1214,6 +1214,44 @@ def test_ok_decode_setof(self) -> None:
...
+                    Example(values=asn1.SetOf([1, 2, 3]), data=b"x"),
+                    b"\x30\x03\x04\x01\x78",
+                ),
+                (
+                    Example(values=asn1.SetOf([1, 2]), data=b"x"),
+                    b"\x30\x0b\x31\x06\x02\x01\x01\x02\x01\x02\x04\x01\x78",
+                ),
+            ]
+        )
+        assert (
```
alex: why is this not just in the table...

reaperhulk (approved):

PR #15657, opened 2026-09-17
Release the GIL while parsing PKCS#12 bundles

alex:
`decode_p12` called `PKCS12_parse` (via `parse2`) with the GIL held. That function runs the MAC KDF and every PBE KDF in the bundle for the iteration counts the file specifies, so a bundle with a large iteration count stalls every other Python thread in the process for the duration of the load, not just the one doing it.

Run `parse2` inside `py.detach(...)`, the same way `PBKDF2HMAC.derive` and the other KDFs already do. This only changes GIL handling; it deliberately does not add an iteration bound (the work factor is a property of the file, see the companion docs PR).

Discovered by Anthropic's OSS Scanner, as vulnerability ANT-2026-S35HH95Z.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01HVLJ4Wp5FraEfACovDdSN6

---
_Generated by [Claude Code](https://claude.ai/code/session_01HVLJ4Wp5FraEfACovDdSN6)_

PR #15660, opened 2026-09-17
Document that DoS via parsing untrusted key material is out of scope

alex:
Docs only. Adds an entry to the "things we wouldn't consider security issues" list in `docs/security.rst`: denial of service via parsing untrusted key material is out of scope. Serialized key formats such as OpenSSH, PKCS#8, and PKCS#12 embed the KDF work factor, so loading a file that specifies an enormous number of rounds or iterations takes correspondingly long.

This covers the reports about unbounded bcrypt rounds in `load_ssh_private_key` and the iteration-count half of the PKCS#12 report (the GIL-release half was #15657).

Discovered by Anthropic's OSS Scanner, as vulnerabilities ANT-2026-1AQCF1WI and ANT-2026-S35HH95Z.


docs/security.rst
```diff
@@ -52,6 +52,17 @@ Examples of things we wouldn't consider security issues:
   represent bugs that should be filed as regular issues, but they are not
   security vulnerabilities because they do not cause memory unsafety or
   undefined behavior -- the process remains in a well-defined state.
+* CPU consumption caused by the work factor embedded in serialized key
```
alex: make this half as long

docs/security.rst
```diff
@@ -52,6 +52,10 @@ Examples of things we wouldn't consider security issues:
   represent bugs that should be filed as regular issues, but they are not
   security vulnerabilities because they do not cause memory unsafety or
   undefined behavior -- the process remains in a well-defined state.
+* CPU consumption from loading serialized key material (e.g. OpenSSH,
+  PKCS#8, or PKCS#12) that specifies an enormous KDF work factor. The file
+  dictates the number of rounds or iterations, and applications that load
```
reaperhulk: This doesn’t feel like actionable advice. To apply their own limits they’d need to write their own parsers, which isn’t reasonable. We should probably just state we consider DoS attacks via parsing untrusted key material to be out of scope?

alex: Agreed; reworded in f586c5b to state that denial of service via parsing untrusted key material is out of scope, with the KDF work factor as the example, and dropped the "apply your own limits" advice.

---

reaperhulk (approved):

PR #15661, opened 2026-09-17
asn1: reject optional or DEFAULT element types for list and SetOf

alex:
With a schema like `list[int | None]`, `SetOf[int | None]` or `list[Annotated[int, Default(0)]]`, `decode_der` looped forever on a 7-byte input containing one element with a mismatched tag. The OPTIONAL and DEFAULT paths in `decode_annotated_type` return `None`/the default after only `peek_tag()`, so the `while !d.is_empty()` loops for `SEQUENCE OF` and `SET OF` examined the same unconsumed TLV on every pass, appending to the result list until `MemoryError` or the OOM killer, with the GIL held throughout.

Such element types aren't valid: the element type of a `SEQUENCE OF` / `SET OF` can't itself be OPTIONAL or DEFAULT (the collection as a whole can be, as `list[X] | None`). `_normalize_field_type` now rejects them with a `TypeError` that says so, covering `X | None`, an OPTIONAL CHOICE (`X | Y | None`), `Annotated[X | None, Explicit(n)]`, DEFAULT-annotated elements, nested collections, and top-level `decode_der(list[...], ...)`. The tutorial notes the restriction.

Discovered by Anthropic's OSS Scanner, as vulnerability ANT-2026-WW16M37X.


docs/hazmat/asn1/tutorial.rst
```diff
@@ -278,6 +278,11 @@ Use :class:`list`\[T] for ``SEQUENCE OF`` and :class:`~cryptography.hazmat.asn1.
     >>> decoded.values
     [1, 2, 3]
 
+Element types cannot be optional (``X | None``) or have a
+:class:`~cryptography.hazmat.asn1.Default` annotation, since ASN.1 does not
+define ``OPTIONAL`` or ``DEFAULT`` for ``SEQUENCE OF`` and ``SET OF``
```
alex: wat? this is total slop, OPTIONAL/DEFAULT are completely coherent for SEQUENCE OF...

alex: Dropped the schema rejection (and this doc note) in 4c89803. The loops now read each element's TLV and decode it from a sub-parser, so an OPTIONAL/DEFAULT element whose tag doesn't match raises `ValueError` instead of never consuming input.

---

src/rust/src/declarative_asn1/decode.rs
```diff
@@ -514,6 +540,76 @@ mod tests {
         });
     }
 
+    #[test]
+    fn test_decode_sequence_of_optional_element() {
```
alex: there should not be a rust test if this is reachable from the python API

tests/hazmat/asn1/test_serialization.py
```diff
@@ -1215,6 +1215,47 @@ def test_ok_decode_setof(self) -> None:
...
+        ).items == [5]
+        with pytest.raises(
+            ValueError, match="unexpected tag for SEQUENCE OF element"
+        ):
+            asn1.decode_der(Example, b"\x30\x05\x30\x03\x04\x01\x00")
+
+    def test_setof_element_tag_mismatch(self) -> None:
+        @asn1.sequence
+        class Example:
+            items: asn1.SetOf[typing.Union[int, None]]
```
alex: oh. this was _incredibly_ badly articulated. `SetOf[Optional[X]]` is not valid.

The thing you _originally_ wrote was that `Optional[SetOf[X]]` was not valid, which is wrong.

alex: Understood, sorry for the churn. fb9d28a goes back to rejecting `list[X | None]` / `SetOf[X | None]` (and DEFAULT element types) in `_normalize_field_type`, now worded as "the element type of a `list`/`SetOf` cannot be optional", with the tutorial saying the collection as a whole can still be `list[X] | None`. The per-TLV decoding and the Rust changes are gone; tests are Python-only.

---

reaperhulk (approved):

